PR #105, opened 2014-02-12
Examples of multi-line and binary chunking to expand the examples already provided

dickwall:
Hi guys - been sitting on this way longer than I should. I put together some examples of different kinds of chunking, including multi-line record chunking for handling text, and a more sophisticated set of binary examples that include support for multiple records found within a buffer (and emitting those) as well as passing forward the incomplete tail to prepend to the start of the next buffer. These were developed by way of experimentation for use on a work project, but I thought they might prove useful as examples for others. The binary chunker could be extracted into a new buffered handler for byte arrays if anyone is motivated to do so, but apart from the tests here it doesn't have a lot of field mileage yet.


chrismattmann:
@rverma have a look at this library

chrismattmann:
@riverma I mean

dickwall:
Well out of date, need to re-write these at some point for recent versions of scalaz-stream. Closing for now

PR #119, opened 2014-03-02
Create `scanF` method

stucchio:
Currently there is a `Process.scan` method which computes (oversimplifying here) a cumulative sum of a process. The function it takes to do this with has type `(B,O) => B`. 

For some stuff I was doing, I needed an alternative version, which does the same thing but with an effectful function of type `(B,O) => F[B]`. This pull request provides that function, currently called `scanF`. 

The implementation can almost certainly be made less ugly, and I'm not sure I'm handling errors correctly.

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -4,13 +4,12 @@ import scalaz.stream.actor.{WyeActor, message, actors}
 import scala.collection.immutable.{IndexedSeq,SortedMap,Queue,Vector}
 import scala.concurrent.duration._
 
-import scalaz.{Catchable,Functor,Monad,Cobind,MonadPlus,Monoid,Nondeterminism,Semigroup, Applicative}
+import scalaz.{Catchable,Functor,Monad,Cobind,MonadPlus,Monoid,Nondeterminism,Semigroup}
 import scalaz.concurrent.{Strategy, Task}
 import scalaz.Leibniz.===
 import scalaz.{\/,-\/,\/-,~>,Leibniz,Equal}
 import scalaz.std.stream._
 import scalaz.syntax.foldable._
-import scalaz.syntax.applicative._
 import \/._
 import ReceiveY.{ReceiveL,ReceiveR}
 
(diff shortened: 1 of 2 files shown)


adamw:
I've hit a similar problem (see https://groups.google.com/forum/#!topic/scalaz/sMQJyk0hgqQ). I have a `Process[Task, Int => Task[Int]]` and I'd like to create a stream of intermediate results given a starting value (`Int`). From what I've tried it is not possible to do using current combinators (`pipe`/`eval` etc.). 

The original story is that I wanted to see how actors & actor communication could be implemented using streams (the FP book mentions that it could be done at the end), and the `Int => Task[Int]` represents a single actor state transition, which can have side-effects, or spawn other actors.

Though I suppose the patch wouldn't work anymore as there's a new `Append` state?

pchlupacek:
@adamw essentially your function seems to me like a channel, what you need here is to memoize the resulting/passed ints to generate next ```Int => F[Int]``` function. for that you can perfectly use signal, that you will set after//before every evaluation of your function i.e. 

```scala

val channelSignal:Signal[Option[Int]] = async.signalOf(None)

def makeFFromInt(last:Option[Int]):Int => Task[Int] = ??? // your implementation here

val channel: Process[Task, Int => Task[Int]] = 
channelSignal.discrete
.map { last:Option[Int] => makeFFromInt(last) andThen _.flatMap(r => channelSignal.set(r).map(_ => r) }

```

adamw:
Ah! Ok I think I got it:

```scala
val p: Process[Task, Int => Task[Int]] = Process(
    (n: Int) => Task.delay { println("T1"); n+1 },
    (n: Int) => Task.delay { println("T2"); n+2 },
    (n: Int) => Task.delay { println("T3"); n+3 }
  )

val sig: Signal[Int] = async.signalOf(0)
val result = p.zip(sig.discrete).evalMap { case (f, v) =>
  val newVtask = f(v)
  newVtask.flatMap { v2 => sig.set(v2).map(_ => v2) }
}

println(result.runLog.run)
```

gives the expected result. However, isn't it kind of a hack, a little bit like using a mutable `AtomicReference` and setting the value to intermediate results?

djspiewak:
@adamw It (the `Signal`) is a bit of a hack.  It stems from the fact that channels (and by extension, sinks) cannot preserve state from value-to-value without external mutation.  Definitely something that we're working on.

Issue #121, opened 2014-03-05
Seq is a poor choice for the block value type.


nuttycom:
Using Seq as the block value type inhibits users from being able to reason about the performance characteristics of operations on blocks of data. I think it would be preferable to either choose a single, specific block type (say Vector) or else to use implicit typeclass constraints and/or explicit dictionaries to provide the operations that are available for block values. Seq is a very muddled type, and doesn't seem to me to really have a place in so principled a library as scalaz-stream.

pchiusano:
Closing. In the new design, there is a separate `Chunk` type, which is strict and supports random access. If you want laziness / fusion, you use `Stream`. 

You can create a `Chunk` from a `Seq` of course, but if that `Seq` is not already `IndexedSeq` it will be copied. (The copying is done lazily though, on first random access. If only sequential access is used there's no need to copy.)

It might be interesting to play with different representations of `Chunk` that are lazier but we are punting on that for now.

PR #237, opened 2014-09-11
map speedup

stucchio:
So this big fugly pull request probably conflicts with the internal philosophy of scalaz-stream. On the other hand, _users_ of the library will not notice the difference, they'll notice only the performance improvement. 

If there is a less java-like way to get the same result, I'd love to see it. 

Motivation: the current implementation of `map` will result in `map(x => x)` transforming `Emit(Seq(1,2,3))` into `Emit(Vector(1)) ++ Emit(Vector(2)) ++ Emit(Vector(3)`, not to mention unnecessary calls to `Util.Try`. This new version results in Emit(Seq(1,2,3))`. 

The net result is a big speedup: 

```
Time (new map): 195 
Time (old map) 1961
```

In the benchmark code below, `map` is the new version and `map2` is the old version. 

  def benchmark {
    val x = (1 to (1024*1024)).map(x => x.toLong).toSeq

```
(Process.emitAll(x) : Process[Task,Long]).map(x => x % 7).runLast.run //warm up jvm
val startTimeMap = System.currentTimeMillis
(Process.emitAll(x) : Process[Task,Long]).map(x => x % 7).runLast.run
println("End time: " + (System.currentTimeMillis - startTimeMap))

(Process.emitAll(x) : Process[Task,Long]).map2(x => x % 7).runLast.run //warm up jvm
val startTimeMap2 = System.currentTimeMillis
(Process.emitAll(x) : Process[Task,Long]).map2(x => x % 7).runLast.run
println("End time: " + (System.currentTimeMillis - startTimeMap2))
```

  }

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -47,7 +47,6 @@ sealed trait Process[+F[_], +O]
       case Halt(_) => this.asInstanceOf[Process[F, O2]]
       case Emit(os) if os.isEmpty => this.asInstanceOf[Process[F, O2]]
       case Emit(os) => {
-        var rest = os
         var err: Option[Exception] = None
         val result = new scala.collection.mutable.ListBuffer[O2]()
         try {


alissapajer:
@pchiusano I was just looking at this PR. 

I think that making such changes to `Tee` makes perfect sense, so that chunking is preserved and we don't have to necessarily handle one element at a time. I think this change makes sense regardless of how a better `map` is ultimately implemented.

With the bigger picture in mind, adding operator fusion (/cc @djspiewak) would be a huge performance benefit for the stream project. But rewriting `map` as piping to a `Process1` wouldn't work well with the new operation fusion signatures.

Since we'd get immediate (order of magnitude) benefits from merging this PR now, what do you think about including it in the 0.7 release?

/cc @stucchio

djspiewak:
@pchiusano To fill you in, I've been working on some ideas for implementing operator fusion in scalaz-stream.  Nothing really settled yet, but it looks promising.  Among other things, I'm pretty sure I can even implement things in such a way that specialization is preserved under certain circumstances (without changing method signatures), so it becomes possible to emit raw array chunks (under the surface) and perform a certain set of operations on those chunks without boxing.

Speaking to the PR…  The fact that `map` denormalizes chunked emits is really really annoying.  Chunking into `Emit` is the way in which you get better performance out of this library when you need it.  Having a core combinator (`map`!) aggressively deoptimize these chunks when it's so straightforward to implement things *better* is quite problematic.  I understand the desire to do this in a more general way with the `available` combinator, and I think that's a very good idea for `tee`, but `map` (and `filter` and `take` and some others) are subject to more aggressive optimization constraints and could be specialized far further than a generic `tee` (i.e. operator fusion).

pchiusano:
You know, it's not really clear to me that `map` should preserve the chunkiness of the input. Here's an example:

```
blah.map(f).take(10)
```

Suppose `map` receives a chunk of length 100. It then applies the function to all 100 elements, but only the first 10 are used by the next stage of the pipeline. That's unnecessary work. Consuming elements one at a time is the most modular as it puts downstream consumers in charge of deciding how much to evaluate. Yes, the constant factors are worse. :)

What I like about exposing `available` is that it at least gives people the option of making different tradeoffs, if they know more about their use case. But I'm not sure totally generic functions like `map` should do things in a chunky way.

Okay, but maybe we have a rule that _everything_ that can be chunkily done is done chunkily, and if you want to unchunk, you can do `blah.unbuffer.map(f).take(10).buffer(5)`, where `unbuffer` behaves like the identity function, but requests inputs one at a time, and `buffer` is the existing `process1.buffer`. That is pretty easy to understand. My concern there is that you are often assembling such pipelines from components that don't know anything about each other. Knowing that it made sense to insert the `unbuffer` there required global knowledge of the overall pipeline, rather than just local knowledge.

One idea that can be used to fix this is to make `emit` actually return a signal which indicates how much the consumer wants:

```Scala
trait Req
case object Available extends Req
case object NoMore extends Req
case class Exactly(n: Int) extends Req
case class AtLeast(n: Int) extends Req
case class AtMost(n: Int) extends Req

trait Process[F[_],O] // just for illustration, not trampolined or anything

// This is the key
case class Emit[F[_],O](out: Seq[O], next: Req => Process[F,O]) extends Process[F,O]

case class Await[F[_],A,O](req: Req, fa: F[A], next: Cause \/ Seq[A] => Process[F,O]) extends Process[F,O]
case class Halt(cause: Cause) extends Process[Nothing,Nothing]
```

I think this would add a lot of complexity to the implementation, but it lets functions like `map` be agnostic to chunkiness as they can just pass along the `Req` they get back from each `emit`.

djspiewak:
> Suppose map receives a chunk of length 100. It then applies the function to all 100 elements, but only the first 10 are used by the next stage of the pipeline. That's unnecessary work. Consuming elements one at a time is the most modular as it puts downstream consumers in charge of deciding how much to evaluate. Yes, the constant factors are worse. :)

Good example!  :-)  One solution to this is operator fusion.  Not only does it allow us to preserve "chunkiness" across multiple operations (such as `map` and `take`), but it also enables these things to be performed in a single pass without any intermediate waste collections and without incurring the overhead of evaluating an `Await` (which is by an order of magnitude the most expensive thing a stream can do).  Operator fusion also allows us to globally optimize a certain class of stream operation, reshuffling and even removing certain operations that are negated or inverted by later operations.

I *do* like the idea of moving to `tee` and `wye` combinators built more on `available` as much as possible.  I think that is a good change to make *in conjunction with* operator fusion, but operator fusion is simply capable of doing *more* than `available`, and doing it faster.  I even have a way of *transparently* optimizing operations on primitive array chunks into tight while loops with no boxing provided that the pipeline of operations on the chunk is also primitive, and I can do it without changing type signatures (though I do have to play some very ugly internal games peeling back the veil of parametricity; these tricks would be 100% contained and hidden from not only the user but even the rest of the library).

The big downside of operator fusion of course is the fact that it requires lifting some operations directly into the algebra.  I can encapsulate this lifting within `step` though, so as a practical matter it wouldn't affect any downstream combinator implementations, just some of the really low-level stuff.

Oh, directly addressing operator fusion would also allow us to have a sane answer to #121.  For a sketch of what form operator fusion could take (note the highly preliminary nature of a lot of the code), see [this gist](https://gist.github.com/djspiewak/d4bbfc78295ef7f3458e).

> I think this would add a lot of complexity to the implementation, but it lets functions like map be agnostic to chunkiness as they can just pass along the Req they get back from each emit.

It would add a *ton* of complexity to the implementation.  If we did something like this, I would rather see just an `Available` and `AtMost` case, to simplify the buffering logic that would need to be implemented.  Really, we might almost go exclusively with `AtMost`, since `Available` could easily page too much and it's really just a special case of `AtMost`.  This is the approach taken in many other streams libraries.

More generally though, I'm not sure I see the advantage to something like this instead of the `available` combinator you describe in #249.

PR #239, opened 2014-09-13
Add process1.lift{First,Second}

mpilquist:
Adds `Tuple2` equivalents of `process1.lift{L,R}`. These require an additional function, `B => Option[C]`, that lets any final values emitted from cleanup to be optionally converted to a pair.


alissapajer:
Is this held up on anything, or can this be merged?

pchiusano:
I was holding out for a more general solution for doing multiplexing and
broadcasting, but I think it is fine to merge.

PR #276, opened 2014-11-23
Add Process1Syntax.contramap

fthomas:


fthomas:
An interesting observation here is that if we implement `map` in terms of `pipe` and `lift` then `contramap` has the same implementation but "reversed":
```scala
def map[O2](f: O => O2): Process1[I,O2] =
  self |> process1.lift(f)

def contramap[I2](f: I2 => I): Process1[I2,O] =
  process1.lift(f) |> self
```
Just like `map` and `contramap` for `Function1`.

PR #283, opened 2014-12-03
Attempt to fix Process.pipeIn - test passes

### LINKED ISSUE #280: SinkTaskSyntax.pipeIn never ends the Process1 [0.6a]
```scala
import scalaz.stream._
{ Process("hi", "there") pipe process1.last to io.stdOutLines }.run.run
```

This prints "there" to stdout, as expected. However,

```scala
import scalaz.stream._
{ Process("hi", "there") to (io.stdOutLines pipeIn process1.last) }.run.run
```

prints *nothing*. 

Since `pipeIn` works with other combinators such as `process1.filter` or `process1.once`, this pointed me towards the idea that the `Process1` being piped in never arrives at a `Halt` step, which is necessary for it to emit anything in the case of `process1.last` (rationale: after receiving the first element, the Process1 becomes a recursive `process1.receive1Or`, but it never hits the "or" branch, which would emit the previous value).

Looking at the code (copied for convenience): 
```scala
implicit class SinkTaskSyntax[I](val self: Sink[Task, I]) extends AnyVal {

  /** converts sink to sink that first pipes received `I0` to supplied p1 */
  def pipeIn[I0](p1: Process1[I0, I]): Sink[Task, I0] = {
    import scalaz.Scalaz._
    // Note: Function `f` from sink `self` may be used for more than 1 element emitted by `p1`.
    @volatile var cur: Process1[I0, I] = p1
    self.map { (f: I => Task[Unit]) =>
      (i0: I0) =>
        val (piped, next) = process1.feed1(i0)(cur).unemit
        cur = next
        piped.toList.traverse_(f)
    }
  }
}
```
I notice that indeed, `cur` is being fed the input data, but it's never disconnected. I could get the remaining data with 
```scala
val (os, _) = cur.disconnect(Cause.Kill).unemit
```
However, doing this in a `onHalt` on the mapped Sink, for instance, is no good, because at that point I will have input elements, but no more functions (outputs of sink) to feed them to. Ideas?

dansanduleac:
Fixes #280, not elegantly yet but at least as proof-of-concept.

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1202,19 +1202,39 @@ object Process extends ProcessInstances {
     def pipeIn[I0](p1: Process1[I0, I]): Sink[Task, I0] = Process.suspend {
       import scalaz.Scalaz._
       // Note: Function `f` from sink `self` may be used for more than 1 element emitted by `p1`.
-      @volatile var cur = p1
+      @volatile var cur = p1.step
       @volatile var lastF: Option[I => Task[Unit]] = None
-      self.map { (f: I => Task[Unit]) =>
+      self.takeWhile { _ =>
+        cur match {
+          case Halt(Cause.End) => false
+          case Halt(cause)     => throw new Cause.Terminated(cause)
+          case _               => true
+        }
+      } map { (f: I => Task[Unit]) =>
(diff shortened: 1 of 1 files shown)


pchiusano:
Going to close this, see discussion at #280.

PR #285, opened 2014-12-08
Move time related processes into a separate module

fthomas:
This PR moves every time related `Process` from the `Process` companion into a separate `time` module (as proposed in https://github.com/scalaz/scalaz-stream/pull/253#issuecomment-59678852).

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1,10 +1,8 @@
 package scalaz.stream
 
 import Cause._
-import java.util.concurrent.ScheduledExecutorService
 import scala.annotation.tailrec
 import scala.collection.SortedMap
-import scala.concurrent.duration._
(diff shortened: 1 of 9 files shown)


pchlupacek:
@fthomas do you think we can deprecate original combinators on Process? I mean we use them quite a lot, and this will require a lot of changes.

fthomas:
The functions are now back in the `Process` companion with a `deprecated` annotation.

Note that
```scala
import Process._
import time._

duration
```
is now ambiguous since `duration` is imported twice. You have to explicitly specify which one you want.

pchlupacek:
@fthomas I am so sorry. I didn't realise that this would conflict.I think in that case we shall have only `time._` like this was originally suggested. Sorry for the confusion.

fthomas:
No problem. I will revert that commit.

fthomas:
@pchlupacek the functions on the `Process` companion are gone again.

pchlupacek:
@fthomas thanks and sorry again. merging.

PR #296, opened 2014-12-19
Remove bufferAll from Process1Syntax.apply

fthomas:
`bufferAll` in `Process1Syntax.apply` seems unnecessary to me. What is its purpose? Can we add a test that fails without it?


pchiusano:
I have no idea why that's there. Definitely not needed.

PR #298, opened 2015-01-04
Semantics of Process.duration

fthomas:
`Process.duration` does not wrap the first call to `System.nanoTime` in `Task.delay` so that this time is fixed as soon as one of the `run*` methods is called on this process. This means that running the resulting `Task` multiple times will always yield durations which use the same reference time:
```scala
scala> val t = Process.duration.once.runLog; println(t.run); Thread.sleep(1000); println(t.run)
Vector(62569 nanoseconds)
Vector(1000755454 nanoseconds)
```
This seems odd to me. I would have expected that the reference time is newly obtained on every `Task.run`.

This PR changes the behavior of `duration` to what I would have expected and adds a corresponding test that currently fails on master.

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -846,10 +846,10 @@ object Process extends ProcessInstances {
    * Note that the actual granularity of these elapsed times depends on the OS, for instance
    * the OS may only update the current time every ten milliseconds or so.
    */
-  def duration: Process[Task, Duration] = suspend {
-    val t0 = System.nanoTime
-    repeatEval { Task.delay { Duration(System.nanoTime - t0, NANOSECONDS) }}
-  }
+  def duration: Process[Task, Duration] =
+    eval(Task.delay(System.nanoTime)).flatMap { t0 =>
+      repeatEval(Task.delay(Duration(System.nanoTime - t0, NANOSECONDS)))
+    }
 
   /** A `Writer` which emits one value to the output. */
   def emitO[O](o: O): Process0[Nothing \/ O] =
--- a/src/test/scala/scalaz/stream/ProcessSpec.scala
+++ b/src/test/scala/scalaz/stream/ProcessSpec.scala
@@ -77,12 +77,18 @@ object ProcessSpec extends Properties("Process") {
     p1.to(pch).runLog.run.size == 3
   }
 
-  property("duration") =  {
-    val firstValueDiscrepancy = duration.take(1).runLast.run.get
-    val reasonableError = 200 * 1000000 // 200 millis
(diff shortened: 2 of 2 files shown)


fthomas:
I want to clarify why this behavior seems odd to me. I would argue that `def e = Process.duration.once.run*` is not referentially transparent because changing the `def` to a `val` would change the meaning of the program.

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -846,10 +846,10 @@ object Process extends ProcessInstances {
    * Note that the actual granularity of these elapsed times depends on the OS, for instance
    * the OS may only update the current time every ten milliseconds or so.
    */
-  def duration: Process[Task, Duration] = suspend {
-    val t0 = System.nanoTime
-    repeatEval { Task.delay { Duration(System.nanoTime - t0, NANOSECONDS) }}
-  }
+  def duration: Process[Task, Duration] =
```
pchlupacek: This behaviour to me is weird. Hence suspend shall be lazy and as such I think suspend(f) is kind of equivalent to eval(Task.delay(f)). Do you know frank what are the differences here ?

fthomas: `suspend` is lazy, but only until you call one of the `run` methods:
```scala
scala> Process.suspend { println("!!!"); Process.emit(0) }.toSource
res19: scalaz.stream.Process[scalaz.concurrent.Task,Int] = Append(Halt(End),Vector(<function1>))

scala> Process.suspend { println("!!!"); Process.emit(0) }.toSource.runLog
!!!
res20: scalaz.concurrent.Task[IndexedSeq[Int]] = scalaz.concurrent.Task@41df68d8
```

pchlupacek: Yeah, this is clear to em. 
However I still don't fully get the difference between 

```eval(Task.delay(System.nanoTime))```

and 

```suspend(emit(System.nanotime))```

Comment (fthomas) [inline src/main/scala/scalaz/stream/Process.scala]: The difference between those is the moment when the side effect happens:
```scala
scala> Process.suspend { Process.emit { println("!!!"); 0 } }.toSource.runLog
!!!
res12: scalaz.concurrent.Task[IndexedSeq[Int]] = scalaz.concurrent.Task@1f8371b2

scala> Process.eval { Task.delay { println("!!!"); 0 } }.runLog
res13: scalaz.concurrent.Task[IndexedSeq[Int]] = scalaz.concurrent.Task@3b608381

scala> res13.run
!!!
res14: IndexedSeq[Int] = Vector(0)
```

pchiusano: This is pretty subtle, but this last example finally made it clear. I think your fix looks good, will merge.

PR #299, opened 2015-01-04
Add process1.tail

fthomas:


src/main/scala/scalaz/stream/process1.scala
```diff
@@ -656,6 +656,20 @@ object process1 {
...
+   * scala> Process(1, 2, 3).tail.toList
+   * res0: List[Int] = List(2, 3)
+   *
+   * scala> Process[Int]().tail.toList
+   * res1: List[Int] = List()
+   * }}}
+   */
+  def tail[I]: Process1[I, I] =
+    receive1(_ => id)
+
```
pchlupacek: is this like `drop(1)`?

fthomas: Yes, the behavior is identical to `drop(1)`, but `drop(1)` expands to the slightly more complicated `receive1(_ => halt) ++ id`.

pchlupacek: ah, ok I think since ++ is lazy, it probably won't have that much overhead. But ok LGTM.

Issue #297, opened 2014-12-19
Deadlock when terminating bounded njoin


edwinchoi:
The deadlock happens if the outer process is terminated while an inner process is blocking on the internal queue. Without any live consumers, the only way for the process to unblock is by terminating the queue, but `njoin` only terminates the queue once all the producers have finished, resulting in a deadlock.

The following reproduces the issue:
```scala
nondeterminism.njoin(1,1)(emit(constant(())))
  .once
  .map(_ => println("signaled"))
  .run.run
println("done")
```

This prints "signaled" but not "done".

PR #300, opened 2015-01-04
Fix deadlock when terminating bounded njoin/mergeN

edwinchoi:
This resolves issue #297.

The deadlock occurs when the process is 'stuck' waiting for queue.enqueue to complete. Since `to` evaluates actions serially, the only way to allow the process to proceed and react to the completion signal is to interrupt the enqueue action.
The fix here is to make enqueue interruptible.

### DIFF
--- a/src/main/scala/scalaz/stream/nondeterminism/nondeterminism.scala
+++ b/src/main/scala/scalaz/stream/nondeterminism/nondeterminism.scala
@@ -113,10 +113,11 @@ package object nondeterminism {
 
             //runs the process with a chance to interrupt it using signal `done`
             //interrupt is done via killing the done signal.
+            //enqueue action needs to be interruptible, otherwise the process will never halt
+            //if the outer process is terminated while the queue is full and an enqueue is pending
             //note that here we convert `Kill` to exception to distinguish form normal and
             //killed behaviour of upstream termination
-            done.discrete.wye(p)(wye.interrupt)
-            .to(q.enqueue)
+            done.discrete.wye(p.to(q.enqueue))(wye.interrupt)
             .onHalt{
               case Kill => Halt(Error(Terminated(Kill)))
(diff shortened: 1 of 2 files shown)


src/main/scala/scalaz/stream/nondeterminism/nondeterminism.scala
```diff
@@ -113,10 +113,11 @@ package object nondeterminism {
 
             //runs the process with a chance to interrupt it using signal `done`
             //interrupt is done via killing the done signal.
+            //enqueue action needs to be interruptible, otherwise the process will never halt
+            //if the outer process is terminated while the queue is full and an enqueue is pending
             //note that here we convert `Kill` to exception to distinguish form normal and
             //killed behaviour of upstream termination
-            done.discrete.wye(p)(wye.interrupt)
-            .to(q.enqueue)
+            done.discrete.wye(p.to(q.enqueue))(wye.interrupt)
```
pchlupacek: @edwinchoi excellent catch, thanks. 

I would perhaps for performance reasons to change to 
`p.flatMap(eval_(q.enqueueOne(_))`
That will in fact cause wye to be not consulted for every enqueue, and still shall correctly interrupt the process on done signal. Also will remove the overhead of `tee` that is used by `to` combinator.   

Also please consider to add 
`.onComplete(eval_(q.close))` cleanup code which I think is missing in the current implementation as well.

edwinchoi:
@pchlupacek thanks for the suggestion. I've updated the PR with the change.

Regarding the cleanup, the current implementation does in fact handle this when the source of processes and the last active input process has terminated -- after which the (terminating) outer process is unblocked. As long as the enqueue action is interruptible, the deadlock should not occur and there should be no leaks.

pchlupacek:
@edwinchoi thanks a lot. merging...

PR #302, opened 2015-01-08
Fixes signalOf deadlock and deprecates signal

### LINKED ISSUE #301: async.signalOf generates deadlocks
```scala
val problemo = Task {
  async.signalOf(())
}
```

Run a crudton of `problemo` instances concurrently, ideally with a smallish thread pool in your `Strategy` (we tested with 2 threads).  You will very quickly run into a deadlock stemming from thread starvation.

The problem is that `signalOf` delegates to `run` internally, which doesn't appear to reschedule the current thread appropriately.  Ultimately, you just run out of threads to use in the `Strategy` thread pool, and `Task#run` starts blocking indefinitely.

djspiewak:
This fixes #301 and deprecates the highly dubious `async.signal` construction.  If you *absolutely* need an uninitialized signal (as is the case inside `Process.awakeEvery`), use `async.toSignal(Process.halt)`.  Don't use uninitialized signals.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.5
+sbt.version=0.13.7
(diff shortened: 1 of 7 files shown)


pchiusano:
Wow that was fast! Pavel, any objections to merging this?

PR #303, opened 2015-01-09
Add implicit Strategy to Exchange.run

kanterov:
I've added implicit Strategy to Exchange.run because in it's implementation it calls `read.wye` which takes one. It made this call with `DefaultStrategy` before.


pchiusano:
Looks fine to me... @pchlupacek can you sign off on this and merge if so?

pchlupacek:
@kanterov TY

PR #307, opened 2015-01-23
Add more Process instances

fthomas:
This one adds `Choice[Process1]` and `Profunctor[Process1]` instances.

### DIFF
--- a/src/main/scala/scalaz/stream/ProcessInstances.scala
+++ b/src/main/scala/scalaz/stream/ProcessInstances.scala
@@ -4,7 +4,9 @@ import scalaz._
 
 private[stream] trait ProcessInstances {
 
-  implicit def ProcessMonadPlus[F[_]]: MonadPlus[({ type λ[α] = Process[F, α] })#λ] =
+  implicit val processHoist: Hoist[Process] = new ProcessHoist {}
+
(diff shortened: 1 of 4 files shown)


fthomas:
The second commit adds the `Zip[Process]` instance and law checks for `MonadPlus` and `Zip`, cleans `ProcessInstances.scala` a little bit (like ordering the instances alphabetically and using lower case for all instance identifiers), and moves all law checking code into `ProcessInstancesSpec.scala`.

pchiusano:
Let's nix the `Zip` instance. I don't like that typeclass (it's just commutative applicative functors), and `zip` on `Process` is not commutative. Other than that looks good.

fthomas:
Ok, the `Zip` instance is gone. But I don't quite understand your remark about commutativity. Is `Process` not commutative because `p1 zip p2` and `(p2 zip p1).map(_.swap)` produces the same values but the order of effects the may have is different?

djspiewak:
Any objections to merging this?

pchiusano:
LGTM. Sorry, am way behind on reviews...

`p1 zip p2` is not commutative for all `F`, for all `p1` and `p2`. There are lots of examples, even with pure processes. If `p1` and `p2` are `Process1`, then switching the order changes which process receives which values of the input stream (`zip` might be more accurately named `alternate` for `Process1`), which affects how `p1` and `p2` behave and hence their output.

Issue #139, opened 2014-03-20
Adequately document to users when `Process1` values allocate mutable state as implementation detail


pchiusano:
Certain combinators, like `compress.inflate` and `compress.deflate`, have a pure interface but allocate mutable objects 'on startup' as an implementation detail. This will not be observable as long as these processes are consumed only using `pipe` / `tee` / `wye`, etc.

We want to note somehow that doing explicit pattern matching or calling `feed` may cause this mutable state to be observable, so users should exercise caution and/or lift the result into `Task`.

Combinators this affects:
- `compress.{inflate,deflate}`
- ???

Adding @fthomas.

PR #308, opened 2015-01-26
Call `end` methods on Inflater and Deflater

bryce-anderson:
Inflater and Deflater uses native code and maintain some state outside the reach of the GC. This is normally cleaned up by the finalize methods, but if many are created and discarded rather rapidly, the GC may not have cause for concern and unchecked memory consumption ensues.

### DIFF
--- a/src/main/scala/scalaz/stream/compress.scala
+++ b/src/main/scala/scalaz/stream/compress.scala
@@ -43,8 +43,12 @@ object compress {
         emitAll(chunks) ++ go(deflater, buf)
       }
 
-    def flush(deflater: Deflater, buf: Array[Byte]): Process0[ByteVector] =
-      emitAll(collect(deflater, buf, Deflater.FULL_FLUSH))
+    def flush(deflater: Deflater, buf: Array[Byte]): Process0[ByteVector] = {
+      val vecs = collect(deflater, buf, Deflater.FULL_FLUSH)
+      deflater.end()
+      emitAll(vecs)
+    }
+
 
     suspend {
       val deflater = new Deflater(level, nowrap)
@@ -82,7 +86,7 @@ object compress {
     suspend {
       val inflater = new Inflater(nowrap)
       val buf = Array.ofDim[Byte](bufferSize)
-      go(inflater, buf)
+      go(inflater, buf) onComplete { inflater.end(); halt }
     }
   }
 }


fthomas:
:+1:

src/main/scala/scalaz/stream/compress.scala
```diff
@@ -43,8 +43,12 @@ object compress {
         emitAll(chunks) ++ go(deflater, buf)
       }
 
-    def flush(deflater: Deflater, buf: Array[Byte]): Process0[ByteVector] =
-      emitAll(collect(deflater, buf, Deflater.FULL_FLUSH))
+    def flush(deflater: Deflater, buf: Array[Byte]): Process0[ByteVector] = {
+      val vecs = collect(deflater, buf, Deflater.FULL_FLUSH)
+      deflater.end()
+      emitAll(vecs)
```
djspiewak: Forgive my ignorance, but wouldn't the following be more correct?

```scala
val end = Task delay { deflater.end() }
await(end) { _ => emitAll(collect(deflater, buf, Deflater.FULL_FLUSH)) }
```

src/main/scala/scalaz/stream/compress.scala
```diff
@@ -82,7 +86,7 @@ object compress {
     suspend {
       val inflater = new Inflater(nowrap)
       val buf = Array.ofDim[Byte](bufferSize)
-      go(inflater, buf)
+      go(inflater, buf) onComplete { inflater.end(); halt }
```
djspiewak: Ditto here:

```scala
go(inflater, buf) onComplete (Process eval_ (Task delay { inflater.end() }))
```

bryce-anderson:
@djspiewak, as I understand it ```onComplete``` takes its parameter as a thunk and thus there is no need to wrap it in a ```Task``` in order to delay execution, which I assume is your goal with ```Task delay```. I also understand that a ```Process1``` is not based on the ```Task``` monad or any monad for that matter, but is a synchronous construct so injecting a ```Task``` in there would make it no longer a ```Process1``` but more akin to a ```Channel[Task, ByteVector, ByteVector]```.

I'm open to the idea that I'm wrong, scalaz-stream has changed a lot since my last PR.

djspiewak:
`onComplete` does take a thunk, but simply side-effecting in that thunk prevents incremental running of the resulting process.  Just walking into the tail of the result of the `onComplete` will force the thunk, which is not necessarily what you want.  Only the final evaluator of the resulting process should dictate exactly when (and if!) such effects are run, and side-effecting in the thunk breaks this assumption.

Incidentally, you're correct that my `await` construction for `flush` converts the `Process0` into a `Process[Task, _]`.  However, if you do not have a `Process[Task, _]`, then you cannot perform effects in a controlled fashion.  Thus, the types need to change in order to enable proper management of the deflater.

bryce-anderson:
It is true that if you want to unwind the ```Process[WhateverMonad, WhateverType]``` by hand you may run into unexpected results. In this case, that is [already understood](https://github.com/scalaz/scalaz-stream/blob/master/src/main/scala/scalaz/stream/compress.scala#L11). 

I believe this PR is in line with the original implementation and will produce no more side effects than is already expected: the extra calls to ```.end()``` occur after the ```Process``` is finished with compression/decompression and are simply beating the GC to the punch to free memory allocated in a location the GC doesn't understand. If @djspiewak or anyone else is interested in rewriting the ```compress``` library to encapsulate the side effects inherent to ```Inflater``` and ```Deflater```, feel free to close this PR but please keep in mind to call the ```.end()``` to free resources at the appropriate time.

Note that I am happy with ```compress``` tools being a ```Process1``` for both flexibility and performance reasons and am not interested in rewriting it.

djspiewak:
> It is true that if you want to unwind the Process[WhateverMonad, WhateverType] by hand you may run into unexpected results. In this case, that is already understood.

My concern is that this will have unexpected implications for processes which are composed concurrently, since unwinding processes by hand is *precisely* what `wye` and `njoin` do.  Obviously end-users shouldn't be calling `step` or decomposing the `Process` algebra as a matter of course, but that functionality does exist and it is most certainly used.

> Note that I am happy with compress tools being a Process1 for both flexibility and performance reasons and am not interested in rewriting it.

I guess that one of the things that's tricky here is that, while there are most definitely effects inside of `compress`, these effects are probably only externally visible via memory introspection.  In a sense, it isn't any more effectful than the `new` keyword.  I don't have any particular "moral objection" to this state of affairs; I'm only concerned about the compositional properties of the resulting `Process1`, and particularly whether or not it can generate unexpected results in conjunction with other constructs.  I'll give this some more thought…

pchiusano:
I think this looks good. 

@djspiewak, fyi, we don't use `Task` here since the deflater is just local mutable state which only one thread of control can access. In order to make the state visible, you would have to step or feed the `Process1` manually, then take the resulting `Process1` and use it in two places in the `Process` network. We do not do that anywhere internally in scalaz-stream, and have adopted the convention that anyone using manual stepping or feeding should know they are doing something potentially unsafe - it is also easy to lose resource safety when using manual stepping or feeding. 

That said, I've hesitated on just making the manual feed/step functions private, since they are useful for people who want to implement new primitive combinators that can't be expressed with what's there. See #139 for more discussion of the rationale.

So in summary, I actually think what's being done is the right tradeoff, but if you disagree and have some new arguments for why I'd definitely like to hear them. :)

fthomas:
Just for completeness: `deflate` and `inflate` were once implemented as `Channel[Task, ByteVector, ByteVector]` and I'm the one to blame for rewriting them as `Process1`, see #138.

Issue #288, opened 2014-12-10
tee should propagate EarlyCause to its Processes


dansanduleac:
I have a situation where I'm sending a source through a sink that is built in a similar way to `io.bufferedChannel` (when `O=Unit`):

```scala
 def bufferedSink[F[_],R,I,O](acquire: F[R])(
                               onKill: R => F[Unit])(
                               onComplete: R => F[Unit])(
                               step: R => F[I => F[Unit]]): Sink[F,I] = {
    // Slightly more advanced `io.bufferedChannel` which is specialized for Sink: it has an onKill finalizer  (similar to bufferedChannel.flush) which I expect to only run if there were no errors. 
    import scalaz.stream.Process._
    eval(acquire).flatMap { r =>
      repeatEval(step(r)).onKill(eval_(onKill(r))).onComplete(eval_(onComplete(r)))
    }
  }
```

The only difference (aside from it taking an `F[_]` instead of just Task) is that instead of a `Sink[_, Option[I]]` I return a `Sink[_, I]`, which, admittedly, is what caused this whole issue.

I wrote this kind of sink for an accumulate-then-commit model - you can send stuff to it, and at the end, if it halts gracefully, it will execute `onKill` and commit the stuff that it's been buffering. Specifically in my case, `step` would write to a temporary DB table, `onKill` would merge the temporary table into the destination table, and `onComplete` would close the DB connection. 

However, if I use the standard pattern of `source to sink`, which internally uses `Process.tee`, and there is an error in the `source`, the `sink` won't know about it _at all_: it will just be killed gracefully, and then proceed to commit incomplete data.
Specifically in here:
```scala
final def tee[F2[x] >: F[x], O2, O3](p2: Process[F2, O2])(t: Tee[O, O2, O3]): Process[F2, O3] = {
  import scalaz.stream.tee.{AwaitL, AwaitR, disconnectL, disconnectR, feedL, feedR}
  t.suspendStep flatMap {
    ...
    case Halt(rsn)             => this.kill onHalt { _ => p2.kill onHalt { _ => Halt(rsn) } }
  }
}
```

could we prepend this case to that one:
```scala
    case Halt(c: EarlyCause)   => this.injectCause(c).drain.causedBy(c) onHalt { _ => p2.injectCause(c).drain.causedBy(c) onHalt { _ => Halt(c) } }
```

I've run the tests and only one fails, which I'm having a look at now:
```
[...]

PR #314, opened 2015-02-09
tee to propagate error cause to its input Processes, as part of Kill

dansanduleac:
The basic idea is to replace `case object Kill` with `case class Kill(upstreamCause: Option[Error])`, together with a handy constructor that can create a `Kill` out of any `Cause`.

Then, upon encountering a `Halt(rsn)` in the evaluation of the `Tee`, `Process.tee` would `.kill(rsn)` its input processes (this is a new method), which will record `rsn`, if it's an `Error`, as part of the `Kill` instance.

The 3rd commit introduces `Process.onSuccess`, a finalizer that runs only if the process halted "successfully" i.e. with an `End` or a `Kill(None)` -- one that was not caused by an error. It then uses this to rewrite `io.bufferedChannel` such that it performs its `flush` Task only on success.

Please see my comment here for the full description: https://github.com/scalaz/scalaz-stream/issues/288#issuecomment-73311627

### DIFF
--- a/src/main/scala/scalaz/stream/Cause.scala
+++ b/src/main/scala/scalaz/stream/Cause.scala
@@ -15,14 +15,19 @@ sealed trait Cause {
    * @return
    */
   def causedBy(cause: Cause): Cause = {
+    import scalaz.std.option._
+    import scalaz.syntax.applicative._
     (this, cause) match {
-      case (End, End)                 => End
-      case (End, Kill)                => Kill
-      case (Kill, End | Kill)         => Kill
-      case (End | Kill, err@Error(_)) => err
-      case (err@Error(_), End | Kill) => err
+      case (End, End)                    => End
+      case (End, k: Kill)                => k
+      case (k: Kill, End)                => k
+      // This is essentially saying, group the error causes of two kills together.
+      case (Kill(mbE1), Kill(mbE2))      => Kill(^(mbE1, mbE2)(_ causedBy _).collect { case e: Error => e } orElse mbE1 orElse mbE2)
+      // If there's a direct error involved, disregard the potential error in the kill.
+      case (End | Kill(_), err@Error(_)) => err
+      case (err@Error(_), End | Kill(_)) => err
       case (Error(rsn1), Error(rsn2)) if rsn1 == rsn2 => this
-      case (Error(rsn1), Error(rsn2)) => Error(CausedBy(rsn1, rsn2))
+      case (Error(rsn1), Error(rsn2))    => Error(CausedBy(rsn1, rsn2))
     }
   }
 
@@ -30,7 +35,9 @@ sealed trait Cause {
    * Converts cause to `Kill` or an `Error`
    * @return
    */
-  def kill: EarlyCause = fold[EarlyCause](Kill)(identity)
+  def kill: EarlyCause = kill(None)
+
+  def kill(maybeError: Option[Error]): EarlyCause = fold[EarlyCause](Kill(maybeError))(identity)
 
   def fold[A](onEnd: => A)(f:(EarlyCause => A)) = this match {
     case End => onEnd
@@ -50,8 +57,8 @@ sealed trait Cause {
    * Converts this cause to `java.lang.Throwable`
    */
   def asThrowable: Throwable = this match {
-    case End => Terminated(End)
-    case Kill => Terminated(Kill)
+    case End        => Terminated(End)
+    case k: Kill    => Terminated(k)
     case Error(rsn) => rsn
   }
 }
@@ -79,12 +86,24 @@ object Cause {
 
 
   /**
-   * Signals force-full process termination.
+   * Signals forceful process termination.
    * Process can be killed when merged (pipe,tee,wye,njoin) and other merging stream or
    * resulting downstream requested termination of process.
    * This shall cause process to run all cleanup actions and then terminate normally
    */
-  case object Kill extends EarlyCause
+  case class Kill(upstreamCause: Option[Error]) extends EarlyCause {
+    def this(c: Cause) = this {
+      c match {
+        // Remove indirection
+        case Kill(c2)   => c2
+        case e@Error(_) => Some(e)
+        case _          => None
+      }
+    }
+  }
+
+  /** This exists just so old client code (and some remaining library/test code) can still reference it */
+  object Kill extends Kill(None)
 
   /**
    * Signals, that evaluation of last await resulted in error.
@@ -117,8 +136,8 @@ object Cause {
    */
   case class Terminated(cause:Cause) extends Exception {
     override def fillInStackTrace(): Throwable = cause match {
-      case End | Kill => this
-      case Error(rsn) => rsn
+      case End | Kill(_) => this
+      case Error(rsn)    => rsn
     }
     override def toString: String = s"Terminated($cause)"
     override def getMessage: String = cause.toString
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -126,27 +126,25 @@ sealed trait Process[+F[_], +O]
    * is killed with same reason giving it an opportunity to cleanup.
    */
   final def pipe[O2](p1: Process1[O, O2]): Process[F, O2] =
-    p1.suspendStep.flatMap({ s1 =>
-      s1 match {
-        case s@Step(awt1@Await1(rcv1), cont1) =>
-          val nextP1 = s.toProcess
-          this.step match {
-            case Step(awt@Await(_, _), cont) => awt.extend(p => (p +: cont) pipe nextP1)
-            case Step(Emit(os), cont)        => cont.continue pipe process1.feed(os)(nextP1)
-            case hlt@Halt(End)               => hlt pipe nextP1.disconnect(Kill).swallowKill
-            case hlt@Halt(rsn: EarlyCause)   => hlt pipe nextP1.disconnect(rsn)
-          }
+    p1.suspendStep.flatMap({
+      case s@Step(awt1@Await1(rcv1), cont1) =>
+        val nextP1 = s.toProcess
+        this.step match {
+          case Step(awt@Await(_, _), cont) => awt.extend(p => (p +: cont) pipe nextP1)
+          case Step(Emit(os), cont)        => cont.continue pipe process1.feed(os)(nextP1)
+          case hlt@Halt(End)               => hlt pipe nextP1.disconnect(Kill).swallowKill
+          case hlt@Halt(rsn: EarlyCause)   => hlt pipe nextP1.disconnect(rsn)
+        }
 
-        case Step(emt@Emit(os), cont)      =>
-          // When the pipe is killed from the outside it is killed at the beginning or after emit.
-          // This ensures that Kill from the outside is not swallowed.
-          emt onHalt {
-            case End => this.pipe(cont.continue)
-            case early => this.pipe(Halt(early) +: cont).causedBy(early)
-          }
+      case Step(emt@Emit(os), cont) =>
+        // When the pipe is killed from the outside it is killed at the beginning or after emit.
+        // This ensures that Kill from the outside is not swallowed.
+        emt onHalt {
+          case End   => this.pipe(cont.continue)
+          case early => this.pipe(Halt(early) +: cont).causedBy(early)
+        }
 
-        case Halt(rsn)           => this.kill onHalt { _ => Halt(rsn) }
-      }
+      case Halt(rsn) => this.kill onHalt { _ => Halt(rsn)}
     })
 
   /** Operator alias for `pipe`. */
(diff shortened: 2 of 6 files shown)


pchlupacek:
@dansanduleac , I think we already have `onSuccess` and that is in fact `append`. 

I am still not convinced on correct behaviour in case of wye or tee. 

What if your `downstream` will fail with `EX1`? You would like this EX1 to be propagated to both sides of tee? So they will see `downstream` exception? 
Even more interestingly, how you will decide in wye what exception will get propagated where ? Think of like 3-4 levels of wye where they can nondeterministically fail. How you will then decide which exception gets propagated where? How you will "signal" that this is not the exception from "upstream" (so not my process terminated) but this is unrelated exception from somewhere else?

Just for me to understand correctly can you try to write code with your new proposed changes that cannot be expressed by current algebra? 

Sorry to be so reluctant with proposed changes, but I know we had spent a lot of time when we were defining this back, and actually we didn't found the right answers to questions above (and maybe more) so we decided that Kill with attempt, onFailure and onHalt works just perfectly fine for almost any scenario we had think of, just with little bit of extra code you have to write in case you really taking care of exception types. 

Thanks a lot for your effort !

dansanduleac:
@pchlupacek ok so indeed I was missing the cases for `pipe` and `wye` - I've added 2 more commits to handle those. 

The way I've defined it, `onSuccess` is not the same as `append`, because the latter doesn't run the given process in the case where we've been "gracefully" killed. 

Now regarding your point about downstream exceptions - my changes are assuming that you want those to be propagated up to the individual processes (albeit wrapped in `Kill` instances). 

Thinking about it now, it might make more sense to only propagate upstream exceptions: basically, exceptions that have to do with a Process's _inputs_. If the inputs to a process `p` finished gracefully, then so should `p`. 
I'll have to think about this a bit more.

pchlupacek:
@dansanduleac lets bring the example of code you want to solve exactly. Maybe we can write it with current combinators, if not let's see if can agree on combinator that we need, and then let's see if we can implement it :-)

dansanduleac:
Ok - here's an example:

Given a sql `Connection`, I want to provide a `Sink[Task, A]` that allows you to batch-insert a bunch of rows, as part of the same `PreparedStatement`, each derived from an element of type `A`:

```scala
trait BatchSink[A] {
  import java.sql._
  import javax.sql._

  private def setValues(ps: PreparedStatement)(a: A): Unit = ???

  private def prepareStatement(c: Connection): PreparedStatement = ???

  def apply(c: Connection): Sink[Task, A] = {
    def acquire = Process.eval(Task.delay(prepareStatement(c)))

    acquire.flatMap { ps =>
      def commit = Process.eval_(Task.delay(ps.executeBatch()))
      def release = Process.eval_(Task.delay(ps.close()))

      def addBatch(a: A) = Task.delay({
        setValues(ps)(a)
        ps.addBatch()
      })

      Process.constant(addBatch).onSuccess(commit).onComplete(release)
    }
  }
}
```

Note that I'm making use of the `onSuccess` method I'm proposing, to only perform the `commit`  when the sink has been gracefully killed (or `End`ed, though that would never happen in this case).

In addition, with my proposed change to `io.bufferedChannel`, the above `apply(c: Connection)` method could be rewritten as:
```scala
def apply(c: Connection): Sink[Task, A] = {
      def acquire = Task.delay(prepareStatement(c))
      import scalaz.syntax.functor._
      def commit(ps: PreparedStatement) = Task.delay(ps.executeBatch()).void
      def release(ps: PreparedStatement) = Task.delay(ps.close())
      def addBatch(ps: PreparedStatement) = Task.now { (a: A) =>
        Task.delay({
          setValues(ps)(a)
          ps.addBatch()
        })
      }

      io.bufferedChannel(acquire)(commit)(release)(addBatch)
  }
```

The idea being that if I have e.g. a trivial `sink: BatchSink[Int]` (that writes ints to a table) then I should be able to do:
```scala
val p1 = Process.range(0, 10) to sink 
p1.run.attemptRun // writes 0 until 10 to the database, then commits the SQL statement
val p2 = (Process.range(0, 10) ++ Process.eval_(Task.delay { sys.error("fail") })) to sink 
p2.run.attemptRun // writes nothing to the database
```

dcsobral:
This is the sort of thing we use wye for. Instead of signalling intention
through Process, we use an external signal, like this:

val finish = async.signal[Boolean]
finish.set(false).run
signal.discrete.wye(addBatch)(wye.interrupt).append(commit).onComplete(release)

And instead of killing that process, we just set a flag.

Mind you, this is the sort of code that we had to strip off because of the
existing bug with wye.

On 10 February 2015 at 03:44, Dan Sănduleac <notifications@github.com>
wrote:

> Ok - here's an example:
>
> Given a sql Connection, I want to provide a Sink[Task, A] that allows you
> to batch-insert a bunch of rows, as part of the same PreparedStatement,
> each derived from an element of type A:
>
> trait BatchSink[A] {
>   import java.sql._
>   import javax.sql._
>
>   private def setValues(ps: PreparedStatement)(a: A): Unit = ???
>
>   private def prepareStatement(c: Connection): PreparedStatement = ???
>
>   def apply(c: Connection): Sink[Task, A] = {
>     def acquire = Process.eval(Task.delay(prepareStatement(c)))
>
>     acquire.flatMap { ps =>
>       def commit = Process.eval_(Task.delay(ps.executeBatch()))
>       def release = Process.eval_(Task.delay(ps.close()))
>
>       def addBatch(a: A) = Task.delay({
>         setValues(ps)(a)
>         ps.addBatch()
>       })
>
>       Process.constant(addBatch).onSuccess(commit).onComplete(release)
>     }
>   }
> }
>
> Note that I'm making use of the onSuccess method I'm proposing, to only
> perform the commit when the sink has been gracefully killed (or Ended,
> though that would never happen in this case).
>
> In addition, with my proposed change to io.bufferedChannel, the above apply(c:
> Connection) method could be rewritten as:
>
> def apply(c: Connection): Sink[Task, A] = {
>       def acquire = Task.delay(prepareStatement(c))
>       import scalaz.syntax.functor._
>       def commit(ps: PreparedStatement) = Task.delay(ps.executeBatch()).void
>       def release(ps: PreparedStatement) = Task.delay(ps.close())
>       def addBatch(ps: PreparedStatement) = Task.now { (a: A) =>
>         Task.delay({
>           setValues(ps)(a)
>           ps.addBatch()
>         })
>       }
>
>       io.bufferedChannel(acquire)(commit)(release)(addBatch)
>   }
>
> The idea being that if I have e.g. a trivial sink: BatchSink[Int] (that
> writes ints to a table) then I should be able to do:
>
> val p1 = Process.range(0, 10) to sink
> p1.run.attemptRun // writes 0 until 10 to the database, then commits the SQL statementval p2 = (Process.range(0, 10) ++ Process.eval_(Task.delay { sys.error("fail") })) to sink
> p2.run.attemptRun // writes nothing to the database
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/scalaz/scalaz-stream/pull/314#issuecomment-73679286>.
>



-- 
Daniel C. Sobral

I travel to the future all the time.

pchlupacek:
@dcsobral which bug are you referring to please? 

@dansanduleac If I understand your problem is that you would like to commit records that have been saved in batch in case the operation was successful right? That means `downstream` of `tee` will complete normally because of the left side of tee being `exhausted`. 

There are couple of solutions for this. I am not sure about the environment you at, but simple wrapping the left side to Option[A] can solve your problem, because you can emit None as very last item and instead of sink[Task,A] you may have Sink[Task, Option[A]], that will commit once `None` is received. 

For that pattern there are actually combinators already, see `Process.terminated`.

hope that helps.

djspiewak:
In case it was missed, there is another common use-case for this sort of thing: atomic file writes.  Basically, imagine a version of `io.fileChunkW` which *atomically* writes to a target file.  This would be implemented as creating a temp file in a leading `Await`, writing into the temp file in the `Sink` functions, and then atomically moving the temp file to the destination in the `onComplete`.

Here's the trick though: what happens if the input source errors?  Ideally, we would want to abort the atomic write and *not* move the temp file over.  However, regardless of the completion status of the source, the sink will only ever receive `Kill`.  We don't get any information with which to make a decision, and so thus an atomic version of `io.fileChunkW` **cannot be implemented** in the present version of scalaz-stream.

I'm actually in favor of just making `Kill` straight-up parameterized on an original `Cause`.  I can't think of a case where we `Kill` without having an original reason, and we can always just `Kill(End)`.

pchlupacek:
@djspiewak @dansanduleac I have no problem with whatever change on this, but, before doing so I would like to see all different cases and the way how they have to behave that means, pipe, tee, wye and what if `left` terminates/fails, then right will be killed with `xy` and downstream will receive `yx` cause. 

After agreeing on these semantics, I am ok to implement it, it shall not be that big deal. In fact the reason why `Kill` is everywhere was that we didn't ever find proper agreement on semantics, so we just thrown `Kill` everywhere :-)

djspiewak:
@pchlupacek 

> After agreeing on these semantics, I am ok to implement it, it shall not be that big deal. In fact the reason why Kill is everywhere was that we didn't ever find proper agreement on semantics, so we just thrown Kill everywhere :-)

Makes sense to me.  :-)  I'll give some thought to what the semantics should be.  Redefining `Kill` to contain a `Cause` gives us a lot of flexibility in this area, since `Kill(Kill(End))` is distinct from `Kill(End)`.  Basically, we get nesting and chaining in a way that we do not have now.

I'm working on the double finalizers first.  Once those are out of the way, I'll see what I can do with this one.

djspiewak:
@dansanduleac Instead of using `Option[Error]` inside of `Kill`, what if we just took a `Cause`?  The `Kill(None)` case would effectively be covered by `Kill(End)`, and allowing `Cause` instead of just `Error` would allow arbitrary nesting of kills (e.g. `Kill(Kill(Kill(Error(...))))`), which solves the problem of defining semantics for nested compositions.  Does this seem like a reasonable proposal to you?

We still need to define what semantics exactly we want here, but I don't think that should be too hard to do.

djspiewak:
@pchlupacek How do these semantics strike you?  I can fill in the `tee` and `wye` cases, but I thought I'd float `pipe` to see how it sits, since the others are directly analogous.

## Pipe

- Left halts with cause `c`
  + Right is killed with cause `Kill(c)`
  + If Right halts with error `e`
    * Downstream halts with cause `e causedBy c`
  + If Right halts with cause `c2`
    * Downstream halts with cause `c`
- Right halts with cause `c`
  + Left is killed with cause `Kill(c)`
  + If Left halts with error `e`
    * Downstream halts with cause `e causedBy c`
  + If Left halts with cause `c2`
    * Downstream halts with cause `c`

## Tee

*analogous to `pipe`, except with disconnect*

## Wye

*analogous to `pipe`, except with disconnect*

pchlupacek:
@djspiewak cool seems ok to me. What will happen if Downstream halts with Cause? What will be then injected to Left/Right side ?

djspiewak:
@pchlupacek How does the downstream halt with cause in such a way that the upstream can catch the propagation?  Do you have an example?

pchlupacek:
```scala 

(killSignal.wye(p2)(wye.interrupt)).to(failingSink)

```

This is in fact wye and tee. In your proposition, the left side (wye`s downstream) will receive cause from the channel. Then the question is if p1 and p2 will just receive `Kill` (which I am fine with) or will receive cause. However this also assumes that 
```scala
p2.to(failingSink)
```
has different behaviour where `p2` will receive the cause unlike in previous code, where only differentiation is that `p2` has additional interrupt signal. 

Is more like to agree what exact behaviour we will have in these scenarios. Currently, we have simplified this that Kill is everywhere so it is easy to reason about, in proposed semantics, the behaviour will be different.

djspiewak:
@pchlupacek I think that `p2` (and `killSignal`) should receive `Kill(Kill(c))`, where `c` is the failing cause from `failingSink`.  We can just keep nesting things up the chain.  It's a little clunky, but it avoids loss of information and allows sides to distinguish downstream failure from left/right-side failure.

pchlupacek:
hm, this is actually what we didn't like originally. I am not confident that nesting will give you any `reasonable` information that has to be extracted later. I didn't fully bought yet the reason for need to propagate the exceptions through the chain. I still don't see why the left leg needs to know exact reason why the right leg terminated (I mean can't we do it different way?). I see only program flow top to bottom, where bottom part of the program really need the reason why the top terminated. I simply can't imagine somewhere in the code you will do 

```scala
xxx.onHalt {
 case Kill(Kill(Kill(Error(boom)))) => recoverFromLeftLegInRightBranch(boom)
 ...
}
```

Another think is that in example I have provided, it may happen that eventual pattern match on reason for termination worked just fine, but just adding this interrupt functionality will cause the pattern match to stop working w/o compiler complaining. I simply see these sort of behaviours as counter intuitive and hard to debug.

djspiewak:
@pchlupacek The left leg needs to know the exact reason why the right leg terminated because *without* that, we cannot implement extremely useful tools like atomic file write channels and everything similar in that vein.  We have plenty of use-cases at this point; there's really no question about it.

I do agree that no one is ever going to match `case Kill(Kill(Kill(Kill(` to pull something out, and if they did it would be amazingly brittle.  However, we need to distinguish downstream explosions from left/right explosions without loss of information (right now, we just lose information), and that means nesting.  We can write a simple combinator that extracts the root cause from some nested `Kill`(s) regardless of depth (this would make such code immune to the brittleness problem that you point out).

If it makes you feel better, think of it as a stack.  :-)

pchlupacek:
@djspiewak I am still more in favour to control this explicitly. I think you can do the atomic file write w/o this sort of error propagation. We do this sort of transaction boundaries in our code. You just need intermediate signal to do this. its not that much code/performance overhead and you can make in fact combinator for this. 

Let's bring the concrete example (i.e. atomic write) and see if we can do better when we will have these `stacked` causes or the current combinators. Can you perhaps write sample of the atomic file write that will work after your proposed changes will be made? I will then try to write (if as you say possible) the code with current combinators and we can see how odes it look?

Other than that you right that nested kills can be resolved like you propose. Perhaps, also we have to see what paul is proposing for the algebra, perhaps we get this for free :-)

djspiewak:
@pchlupacek The following use-case is archetypical:

Implement a function of type `String => Sink[Task, ByteVector]` with the following properties…

1. Constant memory usage
2. Data written to the sink is written out to a file specified by the `String` filename, in order
3. If the process to which the sink is fused produces an error, the file should not be written!
4. If the process to which the sink is fused completes normally, the file should be written such that no concurrent disk operations may be interleaved (if the filesystem supports such an operation)

This cannot be implemented as specified in the current scalaz-stream.  With the proposal I have here, the implementation is relatively straightforward:

```scala
private def writeToFile(file: String)(bv: ByteVector): Task[Unit] = ???    // obvious implementation

def atomicFileChunkW(filename: String): Sink[Task, ByteVector] = {
  val open = Task delay {
    File.createTempFile(filename, "tmp")
  }

  def clean(target: String) = Task delay { File.delete(target) }
  def commit(target: String) = Task delay {
    Files.move(target, filename, StandardCopyOption.ATOMIC_MOVE)
  }

  await(open) { target =>
    sink lift writeToFile(target) onHalt {
      case c @ Kill(End) => eval_(commit(target)) causedBy c
      case c @ Kill(_) => eval_(clean(target)) causedBy c
      case cause => Halt(cause)
    }
  }
}
```

pchlupacek:
@djspiewak If I understand correctly, what we need to achieve is the file is only moved to `commit` if the left side of the process terminated correctly, whereas shall be deleted if this terminated in error or was killed.

Don't take me wrong, my whole point for this is that I think exception is not very usefull type, and as your example showing as well, you just need to distinguish between these situations, you don't really inspect the exception // cause. 
 
I think the code your describing can be written today ( as @dcsobral pointed out) by using external signal either by style he is proposing or by simply awaiting Option[Boolean] in `onHalt` code. I think, is perfectly ok to signal transaction boundaries by external signal. 

I agree however, that your solution looks more elegant (it also allows any `F` whereas introducing signal bounds the whole thing to `Task`). How about just simplify the whole think and add to `Kill` only boolean parameter indicating normal/or abnormal interruption of the fuser? Wouldn't that do same job?

djspiewak:
@pchlupacek 

> If I understand correctly, what we need to achieve is the file is only moved to commit if the left side of the process terminated correctly, whereas shall be deleted if this terminated in error or was killed.

That is correct.

> Don't take me wrong, my whole point for this is that I think exception is not very usefull type, and as your example showing as well, you just need to distinguish between these situations, you don't really inspect the exception // cause.

Oh `Exception` is a perfectly awful type.  It's what we have inside of `Error` though.  There's a certain elegance to having `Kill` contain a `Cause`.  Yes, that means it can also contain an `Error` which can contain an `Exception`, but it also gives us a lot of flexibility in terms of inter-process communication, and it also allows us to preserve information by way of nesting.

In my example, I don't need to look at the specific cause, certainly.  I have no doubt though that there are significant examples where you *do* need the cause.

> I think the code your describing can be written today ( as @dcsobral pointed out) by using external signal either by style he is proposing or by simply awaiting Option[Boolean] in onHalt code. I think, is perfectly ok to signal transaction boundaries by external signal.

Far, far less elegant and not self-contained.  As an extension, also not safe!  If someone fails to trip the signal, then we don't commit the transaction.  This is something that we absolutely should be able to encode using pure scalaz-stream, without resorting to unsafe external signaling.

> I agree however, that your solution looks more elegant (it also allows any F whereas introducing signal bounds the whole thing to Task). How about just simplify the whole think and add to Kill only boolean parameter indicating normal/or abnormal interruption of the fuser? Wouldn't that do same job?

It would do the same job *for this use-case*.  I don't like the idea of playing whack-a-mole with various ways that we lose information when killing fused processes.  I'd rather just pick a general solution and stick with it.  The `Kill(Cause)` solution does not discard *any* information, and thus is fully general.  We won't have to come back to this in six months and add more expressiveness because we found a new use-case; things will just work.  I agree that the `Kill(Kill(Kill(...` is somewhat ugly, but it's trivial to add combinators that extra information from such a structure in an elegant and non-brittle fashion (as I said, it's just a linked list).  The important point is that the information is *there* and can be extracted.  If we use `boolean` instead of `Cause`, the information is gone.

pchlupacek:
@djspiewak generally I agree. I just feel a bit fishy when we will throw the causes between the legs back and forward. This all my point. would like to hear @pchiusano point of view on this, generally, I have no objections, but still have kind of feeling we don't do that very right.

djspiewak:
I'm going to close this for now, so we can track with #288.

Issue #315, opened 2015-02-11
Process#toTask does not support equational reasoning


djspiewak:
See [here](https://github.com/scalaz/scalaz-stream/blob/master/src/main/scala/scalaz/stream/Process.scala#L981) for why.  The essence of the symptoms here are that you can call `toTask` on a `Process` and run it repeatedly, getting *different* results from what would have happened if you call `toTask` multiple times.  Specifically, the following two functions are *not* equivalent:

```scala
def foo[A](p: Process[Task, A]): Task[A] = p.toTask >> p.toTask

def bar[A](p: Process[Task, A]): Task[A] = {
  val step = p.toTask
  step >> step
}
```

For reasons that should be obvious, this is annoying.  Is it intentional?  I think that this is pretty simple to fix by throwing a `Task delay` around the current body of `toTask`, then `join` at the end.

djspiewak:
Here's an example snippet that demonstrates a `Process` where you can run into trouble:

```scala
import scalaz.concurrent._
import scalaz.stream._
import scalaz.syntax.monad._
 
import java.util.concurrent.atomic.AtomicInteger
 
val i = new AtomicInteger
val src = Process eval (Task delay { i.getAndIncrement() })
val p = src.fold(2) { _ * _ }
 
println((p.toTask >> p.toTask).run)
 
i.set(0)
 
val step = p.toTask
println((step >> step).run)
```

alissapajer:
This issue is effectively exposing the `var` that's inside the impl of `toTask`. This `var` should remain private to `toTask` at all costs, but as @djspiewak pointed out, the since `toTask` isn't wrapped in a `Task`, this `var` is initiated when `toTask` is called rather than when it is run.

The entirety of any function returning a `Task` should be inside `Task` to avoid issues like this. 

In short, +1 to the change.

alissapajer:
@djspiewak I was just reading backlogs on gitter. If I see a method that returns a `Task` I assume referential transparency. Do you have a concrete example of how `toTask` is no longer useful when you wrap it all up in a `Task` (i.e. when it's not referentially transparent).

dcsobral:
To sum up some discussion that happened on gitter, there's an intentional
side effect happening with toTask. It is _supposed_ to "grab the next"
output of a Process. However, that's not the only side effect going on.
There are two:

* (Re)starting the Stream
* Reading the next element

Perhaps a better signature would be Task[Task[T]], where both effects are
represented.

[...]

PR #317, opened 2015-02-13
Implemented (and documented) a safe(r) toTask-like function

djspiewak:
`Process.toTask` isn't referentially transparent (see #315).  This deprecates `toTask` and introduces `stepTask`, which satisfies the same use-case while being easier to reason about.  Note that `stepTask` still shares the major foible of `toTask`, which is that premature abortion of the step process will result in resource leaks, but at least the type signature now provides sufficient power to reason about restart semantics.  If nothing else, maybe the documentation will help a bit.

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1,15 +1,18 @@
 package scalaz.stream
 
 import Cause._
+
 import java.util.concurrent.ScheduledExecutorService
+
 import scala.annotation.tailrec
 import scala.collection.SortedMap
 import scala.concurrent.duration._
+
 import scalaz.{Catchable, Functor, Monad, Monoid, Nondeterminism, \/, -\/, ~>}
 import scalaz.\/._
 import scalaz.concurrent.{Actor, Strategy, Task}
 import scalaz.stream.process1.Await1
-
+import scalaz.syntax.monad._
 
 /**
  * An effectful stream of `O` values. In between emitting values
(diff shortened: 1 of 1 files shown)


pchlupacek:
@djspiewak do you think, thee may be a way to eventually perform cleanups even after task completes by interruption? I mean if we will wrap this whole thing to actor, we may, detect interruption and then run the cleanups. What do you think?

djspiewak:
@pchlupacek I was thinking about that, and I honestly I don't think it's possible without out-of-band timeouts or something weird like that.  The problem is that we *fully* cede control to the user with each `run` of the inner `Task`.  The inner `Task` completes…and that's it.  We obviously know what state we're in, but we're hardly executing any longer!  So the only way we could really clean anything up is to try to detect that the user has given up on the `Task`, and run the finalizers ourselves.  This would require some sort of timeout mechanism, and it frankly sounds even more error prone than trusting users to run to completion.

Another option would be to change the type signature further and try to provide a mechanism by which a `Process` can be stepped more robustly.  However, we can't go too far in that direction without landing directly on `step`, so I'm not sure what is to be gained in that regard.

pchlupacek:
@djspiewak I see what you say. I just would like to be minimalistic with the amount of unsafe code we have in the lib. In fact, if we are changing the task signature why not do it more robust? 
I actually had usage for toTask only in 1-2 situations and it was just shorthand for me instead of using `Process.step` directly.
Having said that not sure what signature we shall have with toTask ....

fthomas:
If it is unsafe, why not removing it completely? But if it stays it would be nice to have some tests for it.

pchlupacek:
I have no objections to deprecate it. 
However I would do it after we will put here discussed `toInputStream` code at least, as this was one of the places where I found it useful.

djspiewak:
Actually, `toInputStream` is a fantastic example of where `toTask` *doesn't* make sense, since `InputStream` supports early termination.  The only places where I've used `toTask` are in tests and in working with sinks that I had external knowledge were implemented as `Process.constant`, and even there I have since restructured the code in question.

I wouldn't be totally averse to dropping `toTask` altogether.  My proposed revision is much much safer than the original, but obviously it still leaves resource management entirely to the imagination.  Arguably, `step` does as well, but at least there is some more control and the *potential* to handle things correctly.

I don't have a firm stance on this besides that the existing signature is wrong.

pchlupacek:
@djspiewak I didn't meant the toIS to be implemented in term of toTask. Is just coincidence we have it, and in fact the only reason why we have it is damn javamail api that unfortunatelly cannot be implemented if you don't have input stream :-) So lets leave this open, find a way how to correctly implement input stream and then drop it from next release? 

However you can safely implmenet toInputStream with task, you only need to do something like this: 

```scala 
 val task = killSignal.discrete.wye(source)(wye.interrupt).toTask
```

djspiewak:
> So lets leave this open, find a way how to correctly implement input stream and then drop it from next release?

On it…

pchlupacek:
Yeah @djspiewak , that would be I think excellent. If you interested I can give you our toInputStream code, but hence it uses toTask, and I think you would like to use step here, it's maybe not worth.

djspiewak:
@pchlupacek What do you think?  Should we merge this, or merge a more excitingly different PR to simply remove `toTask` altogether?

pchiusano:
I'm fine with just removing `toTask` altogether, but don't have strong opinions..

pchlupacek:
@djspiewak I am fine to remove toTask.

djspiewak:
@pchlupacek @pchiusano It's gone.

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -7,7 +7,7 @@ import scalaz.{Catchable, Functor, Monad, Monoid, Nondeterminism, \/, -\/, ~>}
 import scalaz.\/._
 import scalaz.concurrent.{Actor, Strategy, Task}
 import scalaz.stream.process1.Await1
-
+import scalaz.syntax.monad._
```
pchlupacek: Do we still need that line?

pchlupacek:
@djspiewak excellent and thanks a lot for your effort!

Issue #211, opened 2014-08-03
Process.halt and Process.empty leak subtypes into public API


fthomas:
`Process` has these two which leak the `Halt` and `Emit` types into the public API:

``` scala
object Process {
  val halt: Halt = Halt(End)
  val empty: Emit[Nothing] = Emit(Nil)
}
```

Shouldn't they have type `Process[Nothing,Nothing]`?

Also as a user of the library: in which situations should I use `empty` instead of `halt` or vice versa?

PR #318, opened 2015-02-15
Bug/step run async

### LINKED ISSUE #290: termination of a Process fails with collect and filter
Hi folks,

What follows is a minimized case of a `Process` constructed with `collect` that doesn't halt when instructed to do so with a `wye.interrupt`.

```scala
import scalaz._
import scalaz.concurrent._
import scalaz.stream._
import scalaz.stream.async.mutable._

class Example {
  val poison: Signal[Boolean] = async.signal[Boolean]

  val original: Process[Task, Int] = Process.constant(42) //onComplete Process.eval_(Task.delay(println(s"ORIGINAL COMPLETE")))

  val secondary: Process[Task, Int] = original collect {
    case in if in < 0 => in
  } //onComplete Process.eval_(Task.delay(println(s"SECONDARY COMPLETE")))

  val output: Task[Seq[Int]] = {
    val signal: Process[Task, Boolean] = poison.discrete
    val result: Process[Task, Int] = (signal wye secondary)(wye.interrupt) //onComplete Process.eval_(Task.delay(println(s"INTERRUPT COMPLETE")))
    result.runLog[Task, Int]
  }

  object Effects extends Thread {
    override def run(): Unit = {
      poison.set(true).run
      println("POISON SET")
    }
  }
}

// val example = new Example(); example.Effects.start(); example.output.run.size
// for { i <- (1 to 100) } yield { val example = new Example(); example.Effects.start(); example.output.run.size }
```

When this example is run, the process sometimes halts (as expected), but more often than not it fails to halt. Here is a [thread dump](https://gist.github.com/alissapajer/c8237dd846cd522ea84a) taken while the process is spinning.

****Ways the example can be modified so that the process still spins****

(1) Include a value caught by the collect:
```scala
val original: Process[Task, Int] = Process(-12) ++ Process.constant(42)
```
(2) Use `filter` instead of `collect`: 
```scala
val secondary: Process[Task, Int] = original filter { _ < 0 }
```

****Ways the example can be modified so that the process never spins****

(1) Bypass the `collect` entirely with
```scala
val secondary = original
```
(2) Replace `<` with `>` in the `collect` `case` statement
```scala
case in if in > 0 => in
```
(3) Explicitly keep everything that is collected using
```scala
case in => in
```
(4) Use `filter` with `>` instead of collect: 
```scala
val secondary: Process[Task, Int] = original filter { _ > 0 }
```

pchlupacek:
This Fixes #290

### DIFF
--- a/src/main/scala/scalaz/stream/Cause.scala
+++ b/src/main/scala/scalaz/stream/Cause.scala
@@ -65,7 +65,7 @@ object Cause {
   sealed trait EarlyCause extends Cause
 
   object EarlyCause {
-    def apply[A](r:Throwable\/A):EarlyCause\/A =
+    def fromTaskResult[A](r:Throwable\/A):EarlyCause\/A =
       r.bimap(Error.apply,identity)
   }
 
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1,11 +1,13 @@
 package scalaz.stream
 
+import java.util.concurrent.atomic.AtomicReference
+
 import Cause._
 import java.util.concurrent.ScheduledExecutorService
 import scala.annotation.tailrec
 import scala.collection.SortedMap
 import scala.concurrent.duration._
-import scalaz.{Catchable, Functor, Monad, Monoid, Nondeterminism, \/, -\/, ~>}
+import scalaz.{\/-, Catchable, Functor, Monad, Monoid, Nondeterminism, \/, -\/, ~>}
 import scalaz.\/._
 import scalaz.concurrent.{Actor, Strategy, Task}
 import scalaz.stream.process1.Await1
@@ -466,7 +468,7 @@ sealed trait Process[+F[_], +O]
               }
             case (awt:Await[F2,Any,O]@unchecked, cont) =>
               F.bind(C.attempt(awt.req)) { r =>
-                go((Try(awt.rcv(EarlyCause(r)).run) +: cont).asInstanceOf[Process[F2,O]]
+                go((Try(awt.rcv(EarlyCause.fromTaskResult(r)).run) +: cont).asInstanceOf[Process[F2,O]]
                   , acc)
               }
           }
@@ -676,6 +678,11 @@ object Process extends ProcessInstances {
      */
     def isEmpty : Boolean = stack.isEmpty
 
+  }
+
+  object Cont {
+    /** empty continuation, that means evaluation is at end **/
+    val empty:Cont[Nothing,Nothing] = Cont(Vector.empty)
   }
 
 
@@ -991,7 +998,7 @@ object Process extends ProcessInstances {
             }
           case Step(Await(rq,rcv), cont) =>
             rq.attempt.flatMap { r =>
-              cur = Try(rcv(EarlyCause(r)).run) +: cont ; go
+              cur = Try(rcv(EarlyCause.fromTaskResult(r)).run) +: cont ; go
             }
           case Halt(End) => Task.fail(Terminated(End))
           case Halt(Kill) => Task.fail(Terminated(Kill))
@@ -1315,86 +1322,97 @@ object Process extends ProcessInstances {
       )(implicit S: Strategy): (EarlyCause) => Unit = {
 
           sealed trait M
-          case class AwaitDone(res: Throwable \/ Any, awt: Await[Task, Any, O], cont: Cont[Task,O]) extends M
+          case class AsyncStep(cur: Process[Task,O], stack: Vector[Cause => Trampoline[Process[Task,O]]]) extends M
+          case class AwaitDone(res: Throwable \/ Any, awt: Await[Task, Any, O], stack: Vector[Cause => Trampoline[Process[Task,O]]]) extends M
           case class Interrupt(cause: EarlyCause) extends M
 
           //forward referenced actor here
           var a: Actor[M] = null
 
-          // Set when the executin has been terminated with reason for termination
+          // It hte execution has been interrupted this is set to EarlyCause (Kill)
+          // Otherwise this is set to reason of termination of the evaluation,
+          // That means End when execution terminated normally and `EarlyCause` in case of failure.
           var completed: Option[Cause] = None
 
-          // contains reference that eventually builds
-          // a cleanup when the last await was interrupted
-          // this is consulted only, if await was interrupted
-          // volatile marked because of the first usage outside of actor
-          @volatile var cleanup: (EarlyCause => Process[Task,O]) = (c:EarlyCause) => Halt(c)
+          // When there is open await this is set to cleanup that shall be run
+          // hence task may get interrupted, and may not complete before/after interruption
+          // we run the cleanup code here.
+          // Otherwise when evaluating AsyncStep, this is set to None.
+          var cleanup: Option[(EarlyCause => Process[Task,O])] = None
 
-          // runs single step of process.
-          // completes with callback if process is `Emit` or `Halt`.
-          // or asynchronously executes the Await and send result to actor `a`
-          // It returns on left side reason with which this process terminated,
-          // or on right side the cleanup code to be run when interrupted.
-          @tailrec
-          def runStep(p: Process[Task, O]): Cause \/ (EarlyCause => Process[Task,O]) = {
-            val step = p.step
-            step match {
-              case Step(Emit(Seq()), cont)         => runStep(cont.continue)
-              case Step(Emit(h), cont)             => S(cb(right((h, cont)))); left(End)
-              case Step(awt@Await(req, rcv), cont) =>
-                req.runAsync(r => a ! AwaitDone(r, awt, cont))
-                right((c:EarlyCause) => rcv(left(c)).run +: cont)
-              case Halt(cause)                 => S(cb(left(cause))); left(cause)
-            }
+          def active(m:M): Unit = m match {
+            case AsyncStep(cur,stack) if stack.nonEmpty =>
+              cur match {
+                case Halt(cause) => a ! AsyncStep(Try(stack.head(cause).run), stack.tail)
+                case Emit(os) if os.isEmpty => a ! AsyncStep(stack.head(End).run, stack.tail)
+                case emt@(Emit(os)) => completed = Some(End);  S(cb(right((os,Cont(stack)))))
+                case Append(h,st) => a ! AsyncStep(h,st fast_++ stack)
+                case awt@Await(req,rcv) =>
+                  cleanup = Some((c:EarlyCause) => Try(rcv(left(c)).run) +: Cont(stack))
+                  req.runAsync(r => a ! AwaitDone(r, awt, stack))
+              }
+
+            case AsyncStep(cur,stack) =>
+              cur match {
+                case Halt(cause) => completed = Some(cause);  S(cb(left(cause)))
+                case Emit(os) if os.isEmpty => completed = Some(End);  S(cb(left(End)))
+                case Emit(os) => completed = Some(End); S(cb(right((os,Cont.empty))))
+                case Append(h,st) => a ! AsyncStep(h,st)
+                case awt@Await(req,rcv) =>
+                  cleanup = Some((c:EarlyCause) => Try(rcv(left(c)).run))
+                  req.runAsync(r => a ! AwaitDone(r, awt, Vector.empty))
+              }
+
+            case AwaitDone(r, awt, stack) =>
+              cleanup = None //reset await state
+              a ! AsyncStep(Try(awt.rcv(EarlyCause.fromTaskResult(r)).run), stack)
+
+            case Interrupt(cause) =>
+              completed = Some(cause)
+              cleanup match {
+                case Some(cf) =>
+                  // `AsyncStep` is nto active, instead we await for Await result.
+                  // hence that may or may not be received, we run cleanup here and complete
+                  // AsyncStep will not get invoked any more because we switch to `interrupted` handler
+                  cleanup = None
+                  Try(cf(cause)).run.runAsync {
+                    case -\/(rsn) => S(cb(left(Error(rsn).causedBy(cause))))
+                    case \/-(_) =>  S(cb(left(cause)))
+                  }
+
+                case None =>
+                  // this indicates we have got in middle of evaluating `AsyncStep`.
+                  // this is no-op hence `AsyncStep` will evaluate in next cycle
+                  ()
+              }
           }
 
+          def interrupted(cause:Cause, m: M) : Unit = m match {
+            case AsyncStep(cur,stack) =>
+              // this indicates we have been interrupted during evaluation the
+              // AsyncStep. We have to clean process here
+              (cur +: Cont(stack)).kill
+              .run.runAsync { _ => S(cb(left(cause))) }
+
+            case AwaitDone(r, awt, cont) =>
+              // indicates we have been interrupted during evaluation of await, and that await has completed
+              // this assures, that we run any cleanups that may have been opened by last evaluation of `req`
+              cleanup = None
+              Try(awt.rcv(EarlyCause.fromTaskResult(r)).run)
+              .kill
+              .run.runAsync(_ => ())
+
+            case Interrupt(_) =>
+              //interrupted after we already completed, no-op
+              ()
+          }
 
           a = new Actor[M]({ m =>
-            m match {
-              case AwaitDone(r, awt, cont) if completed.isEmpty =>
-                val step = Try(awt.rcv(EarlyCause(r)).run) +: cont
-
-
-                runStep(step).fold(
-                  rsn => completed = Some(rsn)
-                  , cln => cleanup = cln
-                )
-
-              // on interrupt we just run any cleanup code we have memo-ed
-              // from last `Await`
-              case Interrupt(cause) if completed.isEmpty =>
-                completed = Some(cause)
-                Try(cleanup(cause)).run.runAsync(_.fold(
-                  rsn0 =>  cb(left(Error(rsn0).causedBy(cause)))
-                  , _ => cb(left(cause))
-                ))
-
-              // this indicates last await was interrupted.
-              // In case the request was successful and only then
-              // we have to get next state of the process and assure
-              // any cleanup will be run.
-              // note this won't consult any cleanup contained
-              // in `next` or `rcv` on left side
-              // as this was already run on `Interrupt`
-              case AwaitDone(r, awt, _) =>
-                Try(awt.rcv(EarlyCause(r)).run)
-                .kill
-                .run.runAsync(_ => ())
-
-
-              // Interrupt after we have been completed this is no-op
-              case Interrupt(_) => ()
-
-            }
+            completed.fold(active(m))(interrupted(_,m))
           })(S)
 
-          runStep(self).fold(
-            rsn => (_: Cause) => ()
-            , cln => {
-              cleanup = cln
-              (cause: EarlyCause) => a ! Interrupt(cause)
-            }
-          )
+          S(a ! AsyncStep(self, Vector.empty)) //fork to run the first evaluation so we won't block
+          (cause: EarlyCause) =>  {  a ! Interrupt(cause)}
         }
 
   }
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -71,22 +71,25 @@ sealed trait Process[+F[_], +O]
    * users are responsible for ensuring resource safety.
    */
   final def step: HaltOrStep[F, O] = {
-    def go(cur: Process[F,O], stack: Vector[Cause => Trampoline[Process[F,O]]]) : HaltOrStep[F,O] = {
+    val empty: Emit[Nothing] = Emit(Nil)
+    @tailrec
+    def go(cur: Process[F,O], stack: Vector[Cause => Trampoline[Process[F,O]]], cnt: Int) : HaltOrStep[F,O] = {
       if (stack.nonEmpty) cur match {
-        case Halt(cause) => go(Try(stack.head(cause).run), stack.tail)
-        case Emit(os) if os.isEmpty => go(Try(stack.head(End).run), stack.tail)
+        case Halt(End) if cnt <=0  => Step(empty,Cont(stack))
+        case Halt(cause) => go(Try(stack.head(cause).run), stack.tail, cnt -1)
+        case Emit(os) if os.isEmpty => Step(empty,Cont(stack))
         case emt@(Emit(os)) => Step(emt,Cont(stack))
         case awt@Await(_,_) => Step(awt,Cont(stack))
-        case Append(h,st) => go(h, st fast_++ stack)
+        case Append(h,st) => go(h, st fast_++ stack, cnt -1)
       } else cur match {
         case hlt@Halt(cause) => hlt
         case emt@Emit(os) if (os.isEmpty) => halt0
         case emt@Emit(os) => Step(emt,Cont(Vector.empty))
         case awt@Await(_,_) => Step(awt,Cont(Vector.empty))
-        case Append(h,st) => go(h,st)
+        case Append(h,st) => go(h,st, cnt -1)
       }
     }
-    go(this,Vector.empty)
+    go(this,Vector.empty, 10)
 
   }
 
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1,7 +1,5 @@
 package scalaz.stream
 
-import java.util.concurrent.atomic.AtomicReference
-
 import Cause._
 import java.util.concurrent.ScheduledExecutorService
 import scala.annotation.tailrec
@@ -64,6 +62,7 @@ sealed trait Process[+F[_], +O]
   /** Alias for `append` */
   final def fby[F2[x] >: F[x], O2 >: O](p2: => Process[F2, O2]): Process[F2, O2] = append(p2)
 
+
   /**
    * Run one step of an incremental traversal of this `Process`.
    * This function is mostly intended for internal use. As it allows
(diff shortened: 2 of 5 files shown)


src/main/scala/scalaz/stream/Process.scala
```diff
@@ -69,25 +70,33 @@ sealed trait Process[+F[_], +O]
...
-        case Append(h,st) => go(h,st)
+        case Append(h,st) => go(h,st, cnt -1)
       }
     }
-    go(this,Vector.empty)
+    // at max stay 10 cycles in step, then consult outer process
+    // this gives chance to higher level combinator that may
+    // interrupt during the process execution to interrupt evaluation of
+    // the process every each 10 cycles.
+    go(this,Vector.empty, 10)
```
alissapajer: I think it may be possible to construct an example such that `10` iterations through `go` is not enough. When we jump out of `go` on line 76 (when `cnt <= 0`) then we don't run `stack.head(...)`, which I think means that there may be items on the stack that will never be run, leading to incorrectness. 

The stack does get preserved in `Cont(stack)`, but I'm not convinced this is all we need. Still working on a concrete example of this though.

djspiewak:
:+1:

pchlupacek:
well, this pr doesn't fix the things yet. I am working on this, and actually this PR brokes a lot. Please do not merge ATM.

alissapajer:
@pchlupacek, with @djspiewak 's latest PR #326 with a fix to `runAsync`, the issues here may be able to be more easily resolved.

pchlupacek:
@djspiewak @alissapajer @pchiusano @edwinchoi 
I think I have finally fixed the runAsync here. It Passes all the tests defined by @alissapajer , and assures that onComplete gets called only once. 

I will put it in work on or server code this week, but I have good feeling that this time this shall work as expected. 

Any comments welcome, and thanks a lot for support so far.

src/main/scala/scalaz/stream/Cause.scala
```diff
@@ -65,7 +65,7 @@ object Cause {
   sealed trait EarlyCause extends Cause
 
   object EarlyCause {
-    def apply[A](r:Throwable\/A):EarlyCause\/A =
+    def fromTaskResult[A](r:Throwable\/A):EarlyCause\/A =
```
djspiewak: :+1: for rename.

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -737,8 +753,8 @@ object Process extends ProcessInstances {
   /** The `Process` which emits no values and signals normal termination. */
   val halt: Process0[Nothing] = halt0
 
-  /** Alias for `halt`. */
-  def empty[F[_],O]: Process[F, O] = halt
+  /** Empty emit. Used to signal end of evaluation form `step` to give chance step to be interrupted */
+  val empty: Emit[Nothing] = Emit(Nil)
```
djspiewak: I'm more than a little bit uncomfortable defining a semantic difference between `halt` and `empty`.  Is this only used in `step`?

src/main/scala/scalaz/stream/async/TaskDriver.scala
```diff
@@ -0,0 +1,164 @@
...
+          onFinish((r: Any) => Trampoline.done(actor ! CompleteBind(r, g)))
+
+      }
+    }
+
+    lazy val actor: Actor[M] = Actor[M]({
+      case Complete(r) if open => complete(r)
+      case CompleteBind(r, g) if open => tryCompleteTask(g(r), actor)
+      case Interrupt if open => complete(left(Task.TaskInterrupted))
+      case _ => //no-op
```
djspiewak: This actually defines `onComplete` finalizers with an "at most once" semantic, not "exactly once".  If a task is interrupted in an `Async` case (not `BindAsync`), that value will be lost entirely and any associated resources will leak.  I have a test which catches this in #326.  True "exactly once" semantics are very difficult to achieve with `runAsync` (I think impossible, actually), but the implementation in #326 approaches this more closely with "exactly once" semantics for non-await-dependent `onComplete`s and "at most once" semantics for await-dependent `onComplete`s with the added guarantee that await-dependent `onComplete`s will be run provided their awaited task produces a value (this added guarantee is what is missing here).

Put another way, this case means that `tryCompleteTask` is actually no different from `runAsyncInterruptibly`!  If this is in fact the semantic that you want, then you may as well use the standard interpreter.

src/main/scala/scalaz/stream/async/TaskDriver.scala
```diff
@@ -0,0 +1,164 @@
...
+   * This method returns a function, that when applied, causes the running computation to be interrupted.
+   * That is useful of process contains any asynchronous code, that may be left with incomplete callbacks.
+   * If the evaluation of the process is interrupted, then the interruption is only active if the callback
+   * was not completed before, otherwise interruption is no-op.
+   *
+   * This implementation also guarantees that when `cb` is invoked on right side, then sequence is always non-empty.
+   * This is usefully to run two processes truly in parallel w/o consulting higher level combinator (i.e. wye/merge).
+   *
+   * This implementation as well guaranteed that any onComplete // onFail // onKill code is consulted exactly once
+   * whether the computation was or was not interrupted.
```
djspiewak:
```scala
val interrupt = await(Task delay { Thread.sleep(200) }) { _ => halt onComplete eval_(Task delay { println("invoked!") }) } runAsync { _ => () }

Thread.sleep(100)
interrupt(Kill)
```

I'm relatively certain that the above will not print anything.

src/main/scala/scalaz/stream/Process.scala
fthomas: Also, the idea of the `empty` method is that it lets you specify the type parameters `F` and `O`, much like the `empty` methods on Scala collections: `List.empty[Int]` or `Map.empty[Int, String]`

src/main/scala/scalaz/stream/async/TaskDriver.scala
djspiewak: Correction: the above will print with your implementation, because the task will not be interrupted!  Using `Task.apply` instead of `Task.delay` will allow the interrupt and should produce the semantics I originally described.

src/main/scala/scalaz/stream/async/TaskDriver.scala
pchlupacek: @djspiewak I took the two tests from the PR you mentioned and
 - "runAsync onComplete on task never completing" passes
 - "runAsync independent onComplete exactly once on task eventually completing" fails on "inner finalizer invocation count"

I think the failure of the later is correct, hence the inner part was never consulted, because the await was interrupted before you entered in `rcv` here. I think this is perfectly consistent, as the interrupt behaves same way as if the `task` will fail in that case you as well won't consulte any inner cleanup. 

If you perhaps have any other cases can you maybe at them to this PR so we can assure they will execute as well? Thanks a lot!

I still believe that this implementation guarantees `exactly once` semantic. Can you please write test case in which this does not hold?  The onComplete here is executed for both scenarios when task completes with value and when task does not complete with value, but only once.   

I don't think so the `tryCompleteTask` has same semantics as `runAsyncInterruptibly` but maybe I am wrong here.  What the implementation does it puts the decision on the actor, and thus we guarantee here that we interrupt only asynchronous code and exit from the loop after very first asynchronous code being interrupted. In other words We won't interrupt any `pure` code, which I think is undesirable. Additionally by having actor in loop I think is a bit easier to reason about what gets executed before and after, but maybe its sort of taste than anything else. 

I am not an actor - lover, but I found them useful for these type of problem solving. 

As you mention, and as I explained above I think it is undesirable to interrupt the Task.delay or Bind code. I don't see any real use case for it, and for me this just brings another  unnecessary level of nondeterminism (e.g. why you would need to nondeterministically interrupt pure code?). You mention you see it as significant limitation, can you please explain this a bit more? If we really need it, than yeah, we can use standard `Task#runAsyncInterruptibly` perhaps.

src/main/scala/scalaz/stream/async/TaskDriver.scala
pchlupacek: Yes @djspiewak Agreed. I think the interruption of Task.delay() is bad thing , but hey, lets see where we need that and is easy to bring it back

src/main/scala/scalaz/stream/Process.scala
pchlupacek: @djspiewak yes, it only serves the special purpose in step. I think we may actually use it more generally but I think we can either make this private so users won't be confused or delete it alltogether. 

@fthomas can you please describe where that would be usefull? Otherwise I am more behind what @djspiewak suggesting and just remove it.

src/main/scala/scalaz/stream/Cause.scala
pchlupacek: @djspiewak Yeah I found apply to be very confusing. On other hand I don't like the new name either, so if someone is better at english than me just please suggest name :-)

pchlupacek:
@djspiewak Also if you feel your implementation does job better and is better to reason about feel free to inject it here :-)

src/main/scala/scalaz/stream/Process.scala
fthomas: `empty` may be useful to help Scala with type inference. It is also nicer than writing `Process(): Process[Task, Int]` for example. See also #211

src/main/scala/scalaz/stream/async/TaskDriver.scala
pchlupacek: @djspiewak I think i see now what do you mean. 

The problem will be in this type of programs 


```scala
await(Task async { cb => openConection(cb) }) { cx => halt onComplete eval_(clean(cx)) }) }  
```
One of the possible solution I see here to write this code as 

```scala
await(Task async { cb => openConection(cb) } attempt (cleanupIfInterrupted)) { cx => halt onComplete eval_(clean(cx)) }) }  
```
then we can introduce this:

```scala

def awaitInterruptibly[F[_], A, O](req: F[A])(onInterrupt: A => F[Unit])(rcv: A => Process[F, O]): Process[F, O] = ???

```

That will do job for us to clean eventually resources from interrupted streams when value was produced. 

I can instruct then driver after interruption to feed the remainder of the task with the Interrupt failure?

What do you think?


```

File: src/main/scala/scalaz/stream/Process.scala (hunk 3 above)
Comment (djspiewak) [inline src/main/scala/scalaz/stream/Process.scala]: @pchlupacek I was thinking more just making `empty = halt` again.  Sometimes it semantically makes more sense to write `Process.empty` rather than `Process.halt`, and it's nice to have both.  If you feel it's best to remove `empty` entirely though, I'm not strongly opposed.  I was just objecting to the fact that `empty` and `halt` are different in this PR.

File: src/main/scala/scalaz/stream/async/TaskDriver.scala (hunk 4 above)
Comment (djspiewak) [inline src/main/scala/scalaz/stream/async/TaskDriver.scala]: > I think the failure of the later is correct, hence the inner part was never consulted, because the await was interrupted before you entered in rcv here. I think this is perfectly consistent, as the interrupt behaves same way as if the task will fail in that case you as well won't consulte any inner cleanup.

Except the task isn't failing, it's completing!  I'll consolidate this discussion with the one on my specific example below, as they are in fact the same test case.

> If you perhaps have any other cases can you maybe at them to this PR so we can assure they will execute as well? Thanks a lot!

It's mostly the test case with the "inner finalizer invocation count" failure that is interesting.

> I still believe that this implementation guarantees exactly once semantic. Can you please write test case in which this does not hold? The onComplete here is executed for both scenarios when task completes with value and when task does not complete with value, but only once.

The test case which fails from my branch is exactly the test case in which exactly once doesn't hold.  :-)  There is an `onComplete` which is not invoked.  As I said, exactly once isn't really possible, but we *can* provide much stronger guarantees than what this branch provides, which is precisely what my implementation does.

> I don't think so the tryCompleteTask has same semantics as runAsyncInterruptibly but maybe I am wrong here. What the implementation does it puts the decision on the actor, and thus we guarantee here that we interrupt only asynchronous code and exit from the loop after very first asynchronous code being interrupted. In other words We won't interrupt any pure code, which I think is undesirable.

Remember that `delay` isn't necessarily pure.  In fact, idiomatically, it most often is not.

`tryCompleteTask` really does have exactly the same semantics as `runAsyncInterruptibly` except that it will not interrupt at bind points on tasks constructed with `delay`, which as I said, I believe is a misfeature.  The really critical point is that `tryCompleteTask` completely discards values upon completion if the task has been previously interrupted, which leaks resources.  This is the real reason why `runAsyncInterruptibly` is inappropriate for this use case.  If we're willing to accept resource leaks (which was incidentally the original objection to my PR!), then we may as well use the standard interpreter.

>  Additionally by having actor in loop I think is a bit easier to reason about what gets executed before and after, but maybe its sort of taste than anything else.
>
> I am not an actor - lover, but I found them useful for these type of problem solving.

I'm more interested in the factoring of separate semantic cases.  For example, how does your code handle the case where we interrupted the task, but it had already completed and passed along to the next process iteration, and thus we have to interrupt *that* iteration?  I know it handles it, but can you point to *one* spot in your code where that case is identified and resolved?  As far as I can tell, the logic for this case is quite spread out, which makes it tricky to reason about.

> As you mention, and as I explained above I think it is undesirable to interrupt the Task.delay or Bind code. I don't see any real use case for it, and for me this just brings another unnecessary level of nondeterminism (e.g. why you would need to nondeterministically interrupt pure code?). You mention you see it as significant limitation, can you please explain this a bit more? If we really need it, than yeah, we can use standard Task#runAsyncInterruptibly perhaps.

It is extremely standard to encapsulate impure code in `Task.delay`.  That's how nearly all `Task` code (that I've seen, including scalaz-stream itself!) is implemented.  `Task.apply` is a somewhat less common constructor.  Your code will *only* interrupt `Task.apply` and binds from `Task.apply`.  Interrupting `Task.delay` is extremely important for the same reason that interrupting `Task.apply` is important.

Remember that the interrupt doesn't actually interrupt the evaluation of an atomic task block!  It just suppresses the value from passing along to the next step in the chain.  This is in fact *precisely* the semantic we want with `await`: interrupt the computation as early as possible to free up CPU/memory resources and *hopefully* prevent side-effects from getting triggered in the first place.

File: src/main/scala/scalaz/stream/async/TaskDriver.scala (hunk 5 above)
Comment (djspiewak) [inline src/main/scala/scalaz/stream/async/TaskDriver.scala]: > The problem will be in this type of programs…

Given how pervasive this idiom is with resource management in scalaz-stream, I think we very definitely need to get it right!  You yourself said that your code is very dependent on `onComplete` getting evaluated; I know for a fact that our code would start leaking resources if we were to ignore this use-case.

My opinion is that if the awaited task completes *despite* our interrupt, then we absolutely need to run any finalizers which are dependent on its value.  Conveniently enough, these finalizers will be contained exclusively within the body of the await, and even more conveniently, those finalizers will *not* have been run previously by the interrupt!  Thus, it is very easy to implement this semantic.  The only weird thing is that the `onComplete` of the outer process (what I call the "outer finalizer" in my test case) will be run before the `onComplete` of the inner process (what I call the "inner finalizer"), but I don't think that's the worst thing in the world.

> then we can introduce this: [`awaitInterruptibly`]

This has a very similar signature to `awaitOr`, which can in fact fulfill exactly the same use-case so long as we parameterize `EarlyCause` with `A` and add an `Interrupted[A]` case to the algebra.

**Comment by djspiewak** [comment]
> Also if you feel your implementation does job better and is better to reason about feel free to inject it here :-)

In what form do you feel is appropriate?  A comment?

File: src/main/scala/scalaz/stream/async/TaskDriver.scala (hunk 5 above)
Comment (pchlupacek) [inline src/main/scala/scalaz/stream/async/TaskDriver.scala]: @djspiewak You are right we may introduce the Interrupted[A](a) but won't that be too much work to resolve it everywhere in code? I mean `awaitInterruptibly` essentially does the same job, and we can live with same algebra? 

The problem I see there is that you will have to evaluate `rcv` twice (once on interrupt and eventually second time when `a` was produced on after-interrupt complete task,  which I think was something we tried to avoid? Solution with `awaitInterruptibly` allows you to resolve this just in Task, and clearly define piece of code that will never be run twice.

**Comment by pchlupacek** [comment]
@djspiewak I am fine to get with your branch if you feel is ready and it resolves the issue better. I just had feeling this was not case from your comments, and to be honest I had problem to understand the way how the runAsync works :-) Lets agree which solution we will go forward and what needs to be still done. I have no objections to go whichever way. I just want to get this resolved asap.

File: src/main/scala/scalaz/stream/async/TaskDriver.scala (hunk 5 above)
Comment (djspiewak) [inline src/main/scala/scalaz/stream/async/TaskDriver.scala]: > You are right we may introduce the InterruptedA but won't that be too much work to resolve it everywhere in code? I mean awaitInterruptibly essentially does the same job, and we can live with same algebra?

Well, for me, `awaitInterruptibly` looks a heck of a lot like `awaitOr`.  I don't think it would be too much trouble to resolve `Interrupted` being added to the `EarlyCause` algebra, and since it allows the existing combinators to cover a slightly extended use-case while avoiding the need to introduce a nearly-redundant combinator, I think that it would be my preferred option if the strengthened `onComplete` was off the table.

> The problem I see there is that you will have to evaluate rcv twice (once on interrupt and eventually second time when a was produced on after-interrupt complete task, which I think was something we tried to avoid? Solution with awaitInterruptibly allows you to resolve this just in Task, and clearly define piece of code that will never be run twice.

`rcv` should be pure.  If it's not, then users have other problems to sort out.  As long as all logic is encapsulated in tasks, we're fine.  Additionally, the most common use of `Await` (citation needed) is via the lower-case `await` constructor, and not `awaitOr`.  If you use lower-case `await`, then the error case for `rcv` just returns `halt`, meaning that we're really only evaluating `rcv` once: when the task completes.

**Comment by djspiewak** [comment]
> I am fine to get with your branch if you feel is ready and it resolves the issue better. I just had feeling this was not case from your comments, and to be honest I had problem to understand the way how the runAsync works :-) Lets agree which solution we will go forward and what needs to be still done. I have no objections to go whichever way. I just want to get this resolved asap.

My branch works essentially universally, but it *does* break `TcpSpec`.  I'm in the process of determining whether or not this is due to an actually incorrect implementation in `tcp` (perhaps relying on the old, looser finalizer semantics) or if it's actually `runAsync` doing something wrong.  It's annoying to track down since I have to work in a VM due to the fact that `TcpSpec` is disabled on Mac.

At present, I'm leaning towards incorrect implementation in `tcp`, just judging by the rest of what's there (e.g. resources are being grabbed and managed with `append` and `flatMap`, rather than `await`).  I just need to figure out where (and if) this is the case.

I 100% agree that we should get this resolved asap.  If you think that we can resolve the leaking resources issue with your `runAsync` in the near-term, then we can go ahead with yours and perhaps switch over to mine down the road if I can get it sorted out.  Alternatively, if it's faster for me to get mine working with `tcp`, then we could fold in your `step` fixes to my branch and land that one (barring other objections).

Honestly, I think both of our branches are in a state where they are *not quite but almost* ready.

File: src/main/scala/scalaz/stream/async/TaskDriver.scala (hunk 5 above)
Comment (pchlupacek) [inline src/main/scala/scalaz/stream/async/TaskDriver.scala]: @djspiewak `rcv` of course need to be pure, but you have no guarantee that continuation is attached to `rcv`. I mean you have no guarantee which cleanups are attached to continuation and to `rcv`. Not sure if I am clear here but lets assume this process: 

```scala 
val cleanup = Trampoline.delay(eval_(Task.delay(println("CLEANUP"))))

Await(task, {
 case -\/(Interruptible(a) => close(a) ++ cleanup
 case -\/(Kill) => cleanup
 case \/-(a) => ??? 
})

```
On interrupt this may print `CLEANUP` twice. With `awaitInterruptibly` that won't be the case.

pchlupacek:
@djspiewak Let's agree we have to resolve this in week or so. I am holding back till EOW, so if you can fix your branch before that let's go with that ok?

src/main/scala/scalaz/stream/async/TaskDriver.scala
djspiewak: Here's the core observation that I've been working off of.  Assume the following example:

```scala
await(t) { a => p1 onComplete f1 } onComplete f2
```

Both `p1` and `f1` are a function of `a`, obviously.  However, `f2` is *not* a function of `a` and cannot be!  Now, it is certainly possible to do something terrible like this:

```scala
await(t)(emit) flatMap { a => halt onComplete f3 }
```

In the example above, `f3` may be a function of `a`, but I would argue that the above is a bit of an anti-pattern.  I've never actually seen code which attempts to handle resource cleanup in this fashion, nor would it in fact be sound in general since errors in the outer process (e.g. `await`) would abort the `flatMap` and result in the `onComplete` never getting evaluated.  Which is in fact, intuitive.

Returning to the example at the top of this reply, `f1` is the only finalizer which can *soundly* release `a`.  Thus, unless the user is attempting to do unsound things with the library (such as what `tcp` does), we can rely on them putting their resource cleanup for `t` in `f1`.  Now, invoking `rcv` with `-\/` and appending the continuation will give us the `f2` finalizer (not `a`-dependent), which we run and satisfy our "exactly once" guarantee.  Invoking it *again* with `\/-` but *not* appending the continuation will give us `f1`, which is exactly and only the finalizer that releases `a`.

You'll note that, in your example above, `cleanup` is not parameterized on `a`!  If it were, then it would be clear just from the types that it could not possibly be invoked on interrupt, and can only be run *after* completion.  Cleanup that is not parameterized on `a`, either via function parameters or closure, is much less interesting.

djspiewak:
> Let's agree we have to resolve this in week or so. I am holding back till EOW, so if you can fix your branch before that let's go with that ok?

Sounds good.  If I can't fix my branch by EOW, then I think it's safe to say that something more fundamental would be going on.

How hard do you think it would be to implement the strengthened inner finalizers semantic I describe above?  (I mean with your branch)

src/main/scala/scalaz/stream/async/TaskDriver.scala
pchlupacek: @djspiewak I agree wit your observation. The problem is however tha tin complex code you won't have that level of verbosity as I have displayed. 

So I agree if we would introduce interrupt and handle interrupt differently than Error and Kill, we may have desired behaviour. Then we can introduce `onInterrupt` or simillar as discussed earlier only to run code from the `last` interrupted await.

So the sematic would be 

Halt -> Normal termination
Kill -> Process was killed
Error(rsn) -> Process was failed
Interrupt(a) -> Process was Interrupted, but emitted value before interrupt run any cleanups for given `a`
 
I am leaning towards implementing that change in algebra, as you have suggested. If you will work on your branch, feel free to introduce Interrupt(a). 

Pavel.

pchlupacek:
I think will require like 3-4 hrs of work. So won't be a big deal.

djspiewak:
> I think will require like 3-4 hrs of work. So won't be a big deal.

Alright, so let's split this into three action items then:

- Decide on a strategy for tasks that complete *after* interrupt and produce resources.  Options:
  + Invoke inner finalizers
  + `awaitInterruptibly`
  + Enrich `EarlyCause` with `Interrupt` case
  + `onPreempt`
- Implement strategy for your branch
- Implement strategy (if not already implemented) for my branch
  + Also fix `TcpSpec` or identify bug in `tcp` if relevant
  + Fold in `step` fix to my branch

Agreed-upon deadline for this whole menagerie is EOW.  Let's try to decide on an acceptable strategy for item 1 today or tomorrow.  We can decide on merging your branch vs mine after the second two items are done and we can hold the two up side-by-side on even footing.  Sound reasonable?

src/main/scala/scalaz/stream/async/TaskDriver.scala
djspiewak: > The problem is however tha tin complex code you won't have that level of verbosity as I have displayed.

I'm not sure I understand?

> So I agree if we would introduce interrupt and handle interrupt differently than Error and Kill, we may have desired behaviour. Then we can introduce onInterrupt or simillar as discussed earlier only to run code from the last interrupted await.

Seems possible.  Though, if we add `Interrupt` to the `EarlyCause` algebra, then all of our discussed finalizer logic can be handled by users just with `awaitOr`, so we wouldn't even need an `onInterrupt`.  Remember that, when using `await` (not `awaitOr` or `Await`), the `Interrupt` cause would never be passed to the body of the `await`, meaning that something like the following wouldn't work the way users might expect:

```scala
await(t) { a => p1 onInterrupt f1 } onComplete f2
```

If interrupted, the `f1` finalizer would never run!  With the enriched algebra, the correct way to implement this *intent* would be as follows:

```scala
awaitOr(t)({
  case Interrupt(a) => f1
  case _ => halt
}) { a =>
  p1
}
```

pchlupacek: @djspiewak Would the reverting the logic to help? I mean would it be possible to distinct ode that may not ever be run twice something like 

```scala
  p.onComplete(onlyOnce(cleanup))
```

What do you think? That way we stay with current semantic and just allow additional functionality where necessary.

djspiewak: > Would the reverting the logic to help? I mean would it be possible to distinct ode that may not ever be run twice something like

That's certainly doable.  I think the bigger problem is that finalizers, as a rule, tend not to be idempotent.  This is the whole source of the deadlock bug.

I think the larger question is whether we care about making existing code that *mostly* works, continue to work.  Obviously we're not going to support code which somehow depended upon having finalizers invoked *twice*, since we agree that was a bug.  The real question is if we want to force users to structurally alter their finalizers, such as with `awaitOr` or `onlyOnce` (I actually think that the `once` combinator would work here).  Furthermore, if we *don't* want to force them to alter their finalizers, do we feel that the inner/outer finalizer semantic that I propose will break common real-world code?  If so, do we feel that the code that will be broken is reasonable and should be supported, or is it just a buggy semantic like the double-evaluation?

The real question is *what do we want* in terms of compatibility, backward and forward.  My inner/outer proposal certainly has the advantage of being *simpler* from a user's perspective (you don't need a special finalizer for interrupt resilience; just use `onComplete` and keep resources within lexical scopes), but the downside being that it has to invoke `rcv` twice in the worst case (once with each side of the either).  The `onlyOnce` and `awaitOr` solutions have the advantage of control and predictability, but they require almost all existing resource management logic to be altered to avoid leaking.

djspiewak:
Closing since bug/step landed.

Issue #316, opened 2015-02-13
Interaction with wye's causing process not to halt


alissapajer:
I've found an issue where an interaction between some `wye`s and `until`s is causing certain code to behave inconsistently. Specifically, sometimes the provided example process hangs when run, and sometimes it halts. I expect it to always halt.

I've run the following example on current `master` as well as on branch `bug/step-runAsync` (pushed to my fork), which contains the fixes/hacks related to #290. On both branches, I see the same behaviour. This rules out two possibilities: (1) this is the same issue as #290 and (2) the fix for issue #290 created this bug.

I'd very much like to provide a cleaner (i.e. smaller) example, because what's below is admittedly, well, messy :/ But as of now I haven't been able reduce this any furtuer. If I rewrite it to remove any of the `wye` usages, then the process always halts, as expected.

Here's an overview of what's happening in the example: We send `terminationData: Process[Task, String]` through the sink `aggregate: Sink[Task, String]`.  Now `aggregate` enqueues this `terminationData` to `queue`, and then merges that enqueue process with a second process `merged`, which dequeues the queue through another sink called `terminator`, which sets a boolean `signal`, which should be triggering the output process to halt. `empty` and `stuff` are in there because, without them, the bug doesn't present itself.

```scala
import scala.language.postfixOps
import scalaz.concurrent._

class Test {
  val expected = "data"
  val signal = async.signalOf[Boolean](false)
  val queue = async.boundedQueue[String](10)

  val terminator: Sink[Task, String] = Process {
    (s: String) => signal set (s == expected)
  }

  val aggregate: Sink[Task, String] = {
    val merged: Process[Task, String] = {
      val empty = Process.empty[Task, String]
      val boundedDequeue = empty merge queue.dequeue to terminator    // required: empty merge ...
      val stuff = Process constant "stuff" until signal.continuous    // required: until signal.continuous
      stuff.wye(boundedDequeue.drain)(wye.mergeHaltBoth)              // required: the entire wye
    }

    queue.enqueue.wye(merged.drain)(wye.mergeHaltBoth)                // required: the entire wye
  }

  val terminationData: Process[Task, String] = Process constant expected until signal.continuous
  val res: Process[Task, Unit] = terminationData to aggregate

  println("____________STARTING TEST")

  implicit val scheduler = scalaz.stream.DefaultScheduler
  res.run.timed(3000).run
  
  println("____________FINISHED TEST")
}
```
[...]

PR #319, opened 2015-02-15
Adds pipe + merge test in WyeSpec. Test nondeterministically fails.

alissapajer:
I added this test, and sometimes it fails (i.e. times out). When I commented out all but the last two tests in the suite in failed consistently. When I ran `sbt test-only *WyeSpec*` the added test failed once and passed once.

### DIFF
--- a/src/test/scala/scalaz/stream/WyeSpec.scala
+++ b/src/test/scala/scalaz/stream/WyeSpec.scala
@@ -399,4 +399,12 @@ object WyeSpec extends  Properties("Wye"){
     val v = i1.wye(p2)(wye.interrupt).runLog.timed(3000).run.toList
     v.size >= 0
   }
+
+  property("interrupt-pipe with wye.merge") = secure {
+    val p1 = Process.constant(42).collect { case i if i < 0 => i } merge Process.constant(12)
+    val p2 = p1 |> process1.id
+    val i1 = Process(false)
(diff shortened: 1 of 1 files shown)


src/test/scala/scalaz/stream/WyeSpec.scala
```diff
@@ -399,4 +399,12 @@ object WyeSpec extends  Properties("Wye"){
     val v = i1.wye(p2)(wye.interrupt).runLog.timed(3000).run.toList
     v.size >= 0
   }
+
+  property("interrupt-pipe with wye.merge") = secure {
+    val p1 = Process.constant(42).collect { case i if i < 0 => i } merge Process.constant(12)
```
alissapajer: This test differs from the one directly above by adding a `wye.merge` here.

alissapajer:
Interesting. 2 of the 4 Travis builds failed, one on scala `2.11.4` and one on `2.10.4`.  Well, this does nicely illustrate the nondeterminism :/

alissapajer:
It's worth noting that this could be related to #316 .

Here's a [thread dump](https://gist.github.com/alissapajer/d31baf35dd2b45350c3b) taken while this added test in `WyeSpec` is hanging. The thread dump here is "identical" to the one in #316 in that the only lines through scalaz code are through `scalaz.concurrent` (and in this case also through `scalaz.stream.WyeSpec`).

PR #320, opened 2015-02-16
Implemented io.toInputStream: Process[Task, Array[Byte]] => InputStream

djspiewak:
Hurray for Java! There are a couple of very nice things about this implementation:

- Support for early termination via `close()` (all finalizers will run)
- In theory, somewhat faster than implementations on top of `toTask`
- ROAR TESTS!
- Somewhat more flexible chunking support
- Implemented not-naively, so chunked reads will work as `InputStream` expects

The actual signature of the method is as follows:

```scala
def toInputStream[A](p: Process[Task, A])(chunk: A => Array[Byte]): InputStream
```

This gives some wiggle room in terms of the types within the process, allowing for things like `ByteBuffer` and such.  An even *more* interesting type signature signature would be the following:

```scala
def toInputStream[A](p: Process[Task, A])(copy: (A, Array[Byte], Int, Int]) => Unit): InputStream
```

This type signature is nice because it eliminates garbage intermediate chunk storage.  If your stream does not contain byte arrays, and you have no need to manipulate byte arrays (e.g. when your stream actually contains direct `ByteBuffer`(s)), then this signature will allow one-copy streaming into the resulting `InputStream`.  I can't bring myself to type that signature with a straight face though, because it really is awful.

In other news, this addresses the only major `toTask` use-case raised in #317.


pchlupacek:
@djspiewak on the 
```scala
def toInputStream[A](p: Process[Task, A])(chunk: A => Array[Byte]): InputStream
``` 
signature why you would like to expose pure transformation function `A => Array[Byte]`. I mean we can always do this in Process `p` before via flatMaps, pipes, channels et all....

I  belive I understand your intention that `A` may be ByteBuffer, Array[Byte], or ByteVector, but this signature will make users think that A=>Array[Byte] is serialization function which I don't think so it is. 

I like the second signature though 

```scala
def toInputStream[A](p: Process[Task, A])(copy: (A, Array[Byte], Int, Int]) => Unit): InputStream
```

but would like maybe rather use scodec.ByteVector instead of Array[Byte] here. Perhaps something like 

```scala
def toInputStream(p: Process[Task, ByteVector])(copy: (Array[Byte], ByteVector]) => ByteVector): InputStream
```
may work?

What do you think?

djspiewak:
The problem that I have with `ByteVector` is the native implementation is very inefficient, while the direct `ByteBuffer` implementation is inapplicable to some problems.  There are a lot of use cases where you would want to stream a process out to an `InputStream`, and the elements of that process *logically* represent byte chunks, but actually reifying those chunks as `Array[Byte]` or (equivalently) as `ByteVector` would require an extra copy step.

Granted, my `A => Array[Byte]` signature doesn't really fix that problem.  :-)  If you don't experience violent revulsion and tingling in the extremities at the sight of the `copy` type signature, I'll take a crack at switching over to that, since it *is* much more efficient.

djspiewak:
Revision: we actually need the following signature.

```scala
  def toInputStream[A](p: Process[Task, A])(logicalLength: A => Int, copy: (A, Int, Array[Byte], Int, Int) => Int): InputStream
```

Still not retching, @pchlupacek?  :-)

src/main/scala/scalaz/stream/io.scala
```diff
@@ -212,4 +217,173 @@ object io {
...
+   * size.  Thus, the `copy` function does not need to concern itself with bounds
+   * checking.
+   *
+   * The other function taken as a parameter is `logicalLength`, which simply returns
+   * the logical length of the source chunk in question.  Thus, if `A = Array[Byte]`,
+   * this function would simply be `{ _.length }`.
+   *
+   * @see #toInputStreamFromBytes
+   */
+  def toInputStream[A](p: Process[Task, A])(logicalLength: A => Int, copy: (A, Int, Array[Byte], Int, Int) => Int): InputStream = new InputStream {
```
djspiewak: Random thought: should this be a typeclass instead?  Among other things, a typeclass would be more PIC-friendly, and so it would actually be substantially faster than random functions.  It would also be quite a bit cleaner.  The obvious downside being that it would be more opaque, and also a typeclass of somewhat dubious soundness.

mikemckibben: I think simplifying the signature to
```scala
def toInputStream(p: Process[Task, Array[Byte]]): InputStream = ???
```
would be the most usable API. Any conversion of `A => Array[Byte]` can be done using existing `Process` combinators as suggested earlier by @pchlupacek. I don't think in general you will gain much efficiencies with the `logicalLength` and `copy` function approach. You are just pushing the buffering burden onto the callers. 
For example, calculating the `logicalLength` of `A` may not be possible without actually serializing to an array first (or some other dynamic calculation). Also, this has to be called for every emitted value because it cannot be guaranteed it will be the same logical length from one value to the next. 
Implementing `copy` function could potentially require buffering as well since asking for the "last x bytes of the serialized representation of A" would in general require serializing the value first and doing an array copy anyways.
Finally, is there a worry about the mutability of using `Array[Byte]`? The assumption in the implementation is that the array in each emit is treated as immutable, but if there was an `unsafeChunkR` in the pipeline passed into this, wouldn't that cause an issue? Granted, this is an obviously stupid thing to do and out of control of the implementation.

djspiewak: There is a convenience method which implements the type signature you suggest (element type of `Array[Byte]`).  The point of the more general signature is to avoid specializing on byte arrays, since not everyone uses them!  An obvious example of this is `ByteBuffer`, but there are other examples as well.  Forcing users into byte arrays requires extra copying in those use cases, which is often prohibitive.  The cost of invoking a buffering function of the signature used here is negligible compared to the cost of allocating, retaining and then discarding scratch space corresponding to every chunk.  The cost goes down even further (basically to zero, due to monomorphic call site inlining!) if we swap in a typeclass instead of a pair of functions.  This also makes the syntax a lot more palettable.

Per element computational costs arent that much of a concern, since one would expect that anyone caring about performance is going to be batching, and scalaz-stream already has plenty of element level costs.  What we want to avoid are *byte* level computational costs, and this is in fact precisely what we get around by allowing users to specify their own copy functions (note that functions like `arraycopy` do not have a byte level cost due to CPU intrinsics).

mikemckibben:
So `A` is really only some type that acts like a buffer, and not some generalized data type? If so, then what about constraining the method signature to require a `BufferLike[A]` type class, with default instances for `Array[Byte]`, `ByteBuffer`, `ByteVector`, etc.

djspiewak:
> So `A` is really only some type that acts like a buffer, and not some generalized data type? If so, then what about constraining the method signature to require a `BufferLike[A]` type class, with default instances for `Array[Byte]`, `ByteBuffer`, `ByteVector`, etc.

This is, in fact, exactly what my random "typeclass" idea is about.  :-)  I'll sketch out what this will look like (note: among other things, it will make the code appreciably faster!).  I'm not sure I like the name "`BufferLike`", but honestly I don't have any better ideas at present.  Maybe `Chunked[A]`?

mikemckibben:
Yes, `BufferLike` is an awful name. I like your `Chunked` suggestion though.

djspiewak:
@mikemckibben Here's a WIP branch that implements the `Chunked` typeclass idea.  WDYT?  https://github.com/djspiewak/scalaz-stream/tree/wip/chunked-typeclass

mikemckibben:
I like it!

pchlupacek:
@djspiewak  on ByteVector, I am not sure if wrapping whatever to ByteVector is a big deal. I thought (and from code what I have seen) that you hardly do some of the ByteVector works more efficient except switching to imperative style. 
The reason why I am pushing the ByteVector is that it clearly identifies and (to sort of extend) guarantees immutability. We had Array[Byte] signatures all over the code before and we switched to ByteVector just for this purpose (in fact we had own ByteStream before).

djspiewak:
> I am not sure if wrapping whatever to ByteVector is a big deal.

With the `Chunked` implementation, the full `InputStream` traversal requires *one* extra allocation and uses primarily intrinsics for data copying at the byte level.  With `ByteVector`, we require *n* extra allocations, one for each chunk in the stream.  That *alone* is a no-go for many applications that I've worked on that require streaming data processing.  The fact that we miss out on intrinsics for data copy by using `ByteVector` is really just icing on the cake.

As to the implementation of the data structure, it's not *bad*, but it's basically just `Vector[Byte]`.  `Vector` is already substantially slower than it needs to be, both because of the inherited implementations in the standard library, and because the data structure is optimized for a more general use-case than we need here.  I have an implementation of a persistent vector trie that implements the same operations as `Vector`, but reads are literally twice as fast.  Some of the other operations are slower, sure, but *reads* (which is what we're interested in here) are not.  That right there immediately shows where `ByteVector` can be improved.  Moving to a `Vector[Long]` implementation, or even a more specialized vector trie which understands that bytes can be stored in longs, would be a significant improvement even above and beyond this.

The point is not that `ByteVector` is bad, just that it *is* imposing unnecessary overhead and substantial constraints (even on allocations alone).  The `Chunked` interface is unsafe, certainly, but it allows for maximal performance in the hot path of byte manipulation, and that is precisely the level at which I consider imperative programming worth the cognitive overhead.

adamretter:
Wow you guys have really put a lot of effort into the original question that I asked! Thanks very much to all of you especially @djspiewak for his work. Looking forward to seeing it in the next version :-)

pchlupacek:
@djspiewak can we elaborate on what is the real difference between 
``` 
val ba: Array[Byte] = ???
val bv: ByteVector = ByteVector.view(ba)

```
I think both will perform the same in this case no? So if I am just very very sensitive about performance, 
I can work with whatever byte structure (array, buffer), and just tag it with ByteVector. If I know then what I do, I can still work with `ba` and assure safety by whatever other technique, if I am really, really sensitive about memory allocation and performance. 

For all others, ByteVector is just here with its api already in scalaz-stream and things just fit perfectly together ?

djspiewak:
> I think both will perform the same in this case no?

Nope!  :-)  There are two significant differences here.  In order of importance:

1. `System.arraycopy` doesn't take a `ByteVector`.  I would invite you to microbenchmark `copyToArray` vs `arraycopy`.  The difference is stark, especially on x86_64.  `System.arraycopy` is a HotSpot intrinsic that is *inlined* at the call site.  Even more impressively, on x86_64, it is ultimately delegating (after some bounds checks) down to a single processor instruction that is executed directly on the chip.  This right here is the reason that even *direct* `ByteBuffer`(s) often fail to offer the same performance as `Array[Byte]`.  Even though things are delegating down to `arraycopy` *eventually*, trucking through megamorphic call sites to get there defeats almost all of the benefits of the inlining, as well as unseating many of the CPU's pipeline optimizations.  Note that the `Chunked` typeclass doesn't share these flaws, since the call-site will be monomorphic and thus, fully inlined all the way down to `arraycopy`.
2. `ByteVector.view` may be relatively lightweight, but it is *three* extra allocations (`Chunk`, `View` and `AtArray`).  That's a lot of allocations that are completely unneeded.  These allocations escape the method scope (thus defeating escape analysis), cross write barriers and even move between threads (defeating register allocations).  `Process` certainly has more than its fair share of unnecessary allocations, but I'd rather move *away* from bad performance, particularly when you're talking about stuff that's going to out-live the eden generation and undergo context shifts.

Critically, if we enforce `ByteVector` in the signature, we remove the ability for any users to do any better than the above without writing their own `toInputStream`.

pchiusano:
@djspiewak, I don't see how the extra level of indirection in `ByteVector` could ever be the bottleneck (`ByteVector.copyToArray` still calls `System.arraycopy` eventually) - after all, for each `ByteVector` you produce you have to step a `Process`, which involves a whole bunch of allocation and indirection... right? I'm skeptical that you can even produce a benchmark in which using `ByteVector.view` vs `Array[Byte]` directly performs any differently for this task. (In which you are stepping a `Process` in between each call to `copyToArray` or whatever.)

But even if there were a difference, it would have to be a pretty big difference for me to consider exposing arrays, because I think the `ByteVector` interface is nicer.

If `ByteVector` needs any improvements, like a 4 arg `copyToArray` function to save on the `take`/`drop` calls, I'm sure those could get merged, especially as it could be done in a backwards compatible way.

mikemckibben:
With the PR in it's current state after refactoring to use the typeclass-based approach, I don't see where byte arrays are exposed at all. The current signature of `toInputStream` looks like
```scala
def toInputStream[A: Chunked](p: Process[Task, A]): InputStream
```
Or is it the `Chunked` interface that the discussion is around? I don't see a way around not exposing arrays in the `copy` signature since an InputStream requires byte arrays, so any `A` has to get chunked into bytes at some point.

pchiusano:
Ah, I obviously did not look at the latest code, my bad. :) My feeling is that a `Chunked` typeclass is overkill and it's fine to specialize to `ByteVector` (which has a `copyToArray` method as well as copy-free ways of creation - `ByteVector.view`), however I'm open to being convinced otherwise if there's some compelling performance reason. I'd like to hear @djspiewak respond to my previous points.

djspiewak:
@pchiusano Stepping is actually surprisingly efficient.  If you look at the `step` function (or any of the other interpreters), there's a pair of branches, a relatively infrequently updated and *small* `Vector`, and no allocations outside of the stack updates and the `Await` case.  `Await` is expensive, no question about it.  I'm working on some ideas to make this cheaper (probably by specializing the algebra a bit more), but no progress there.  For right now, `Await` is basically the only interpreter case that imposes a *significant* cost, and that cost can be mitigated significantly with the batching/chunking combinators (which basically coalesce multiple awaits into a single await + emit).

Regarding the `ByteVector` specialization, remember that, even though it uses `copyToArray`, you have to flow through at least one megamorphic call site in order to get there.  `Chunked` has the same *potential*, but with `Chunked` the call site in question is inside of `toInputStream`, meaning that if your code happens to only use `toInputStream[Array[Byte]]`, then the call site is strictly monomorphic.  The megamorphic call site for `ByteVector` is inside of `View`, and if your code is using `ByteVector` at all, the odds are good that you're pumping more than just the `AtArray` case through that function.  The intrinsic for `arraycopy` is small enough that HotSpot can and will inline the `ByteArrayChunked#copy` function, but it can only do that if the call site is not megamorphic.

The biggest issue for me though is the fact that `ByteVector` is an intermediate allocation, and for many applications, possibly a lot of unnecessary copying!  Not every stream that generates logical byte chunks is generating `Array[Byte]` or `ByteBuffer`.  I'd wager that *most* of them are, but it's certainly not all of them (I've worked on performance sensitive applications where we had logical byte chunks which were of a custom type and representation).  `ByteVector` can't implement a `view` function for *every* possible logical byte chunk, and unlike `Chunked`, `ByteVector` is not user-extensible.  By specializing explicitly on `ByteVector`, we're basically deciding *a priori* what sort of performance limits our users will hit, outright preventing them from doing any better than `ByteVector`.  The `Chunked` solution leaves this door open while simultaneously granting superior PIC optimization opportunities.

Obviously, I'm in favor of `Chunked` over specializing explicitly on `ByteVector`.  :-)  `Chunked` is fairly lightweight, and while it does have an unsafe function within it (`copy`, which I'm certainly willing to rename), it is not in and of itself a particularly unsafe construct.  It's tucked away in the `io` package, so there should already be some understanding that the plumbing requires a bit of care (see also: the `io.unsafe*` stuff).  As a user, you don't have to worry at all except if you're using `ByteBuffer` or implementing your own `Chunked`.  As I mentioned above, low-level byte twiddling operations are *precisely* the level of abstraction at which it is good to compromise on the clean, functional abstractions and sell-out (just a bit!) for performance.  (just to assuage any immediate concerns, I do not believe that `Await` is an appropriate place to compromise on anything, so have no fear on those future optimization ideas)

I'm certainly willing to back out things and specialize on `ByteVector`, but I would rather move library performance in a positive direction, rather than locking it into a particular specialization with clear tradeoffs.

pchiusano:
Okay, Daniel, hear what you are saying, but I'd like to just keep it
monomorphic for now, specialized to ByteVector. I consider it a simpler
API, and I'm just not convinced `Chunked` will make an appreciative
performance difference in this case. If/when this becomes a real bottleneck
and there's empirical data to back up that assertion, we can revisit.

I'd encourage you to put together a PR on scodec-bits to add the 4-arg
copyToArray functions, if you think that would be handy.

djspiewak:
> I'd encourage you to put together a PR on scodec-bits to add the 4-arg
copyToArray functions, if you think that would be handy.

It's more than handy; it's really *necessary*, since without it I have to copy the `ByteVector` *twice* per buffer operation, destroying the view properties in the process!  The good news though is that scodec-bits in `master` already has the 4-arg `copyToArray` function, it's just not available in the version that we currently depend on here.  Not sure what to do about that; maybe I'll just file an issue to adjust the implementation when we update to a new version of scodec-bits.

I'll re-specialize on `ByteVector`.  Do you have strong feelings about rebasing out intermediate commits?  I like to do it with local branches, but rewriting public history (i.e. something that's already been PR'd) gives me hives.

pchiusano:
No, you can leave the commits as is, no need to rebase or anything, I don't
care about that and having the full history is nice. (If you want to rebase
or squash though that is fine too.)

Has there been an scodec-bits release with the 4-arg version? Is it still
in the 1.0.x series? If so, it's binary compatible, and I don't see any
reason not to just bump the version used by scalaz-stream.

djspiewak:
Looks like scodec-bits v1.0.5 has the 4-arg `copyToArray`.  Whoops…

The problem is that it isn't publicly exposed.  I'll look into taking care of that.

djspiewak:
Refactoring complete.  `ToInputStreamSpecs` used to run in 8 seconds, and now runs in 11.  :-(  MICROBENCHMARK WITH SCALACHECK ALL THE THINGS!

src/main/scala/scalaz/stream/io.scala
```diff
@@ -212,4 +217,150 @@ object io {
...
       }}
     }
+
+  /**
+   * Converts a source to a mutable `InputStream`.  The resulting input stream
+   * should be reasonably efficient and supports early termination (i.e. all
+   * finalizers associated with the input process will be run if the stream is
+   * closed).
+   *
+   * Type `A` is constrained by the `Chunked` typeclass, which is to say that
```
pchiusano: Oop, can we kill this comment about `Chunked`?

src/main/scala/scalaz/stream/io.scala
```diff
@@ -212,4 +217,150 @@ object io {
...
+
+          case Step(Emit(as), cont) => {
+            chunks = as
+            cur = cont.continue
+          }
+
+          case Step(Await(request, receive), cont) => {
+            // yay! run the Task
+            cur = Util.Try(receive(EarlyCause(request.attempt.run)).run) +: cont
+            step()    // push things onto the stack and then step further (tail recursively)
```
pchiusano: Put an `@annotation.tailrec` on `step`?

pchiusano:
LGTM! @djspiewak I'm holding off on merging in case you want to address the minor comments I just made first. FF to just merge and fixup after, no need for further review.

On `toOutputStream`, I am not super interested, but I wouldn't object to it being there. :) We could also 'lazily evaluate' and wait until someone else asks for it. Basically, no strong opinions, if you are feeling inspired then go for it!

djspiewak:
> On `toOutputStream`, I am not super interested, but I wouldn't object to it being there. :) We could also 'lazily evaluate' and wait until someone else asks for it. Basically, no strong opinions, if you are feeling inspired then go for it!

I'll put it on my bucket list of things to knock off.  I think it makes sense, at least from a symmetry standpoint, to get this in at some point.

adamretter:
Thanks very much everyone :-)

Issue #321, opened 2015-02-17
Remove dependency on scalaz.concurrent


pchlupacek:
Hi, this is just an initial idea. I would like to explore if we can remove the dependency on scalaz. Namely this is driven by fact that I would like to have full control of concurrent primitives (like Task, Future, and perhaps Actor and Strategy) in our code and don't be dependent on release cycles of scalaz for these.

What do you think guys? I would like to see scalaz-concurrent in our code and perhaps scalaz stuff to be in separate module of scalaz-streams.

djspiewak:
There are some things from Scalaz that we need above and beyond the concurrency primitives.  Most notably, abstractions like `Monad` and `Functor`, and utilities like `Either`.  We can get those from Scalaz or we can get them from Cats if you prefer, but we do need them.

Having full control over our own primitives would be great though.  `Task` is fantastic and beautiful and the fact that it's sort of tucked away inside of Scalaz has always been a shame.  We lose some nice interop properties by having our own `Task`, but at the same time, if we're looking to cut the dependency, then interop isn't really a goal anymore.

rossabaker:
I've been considering this as well, and [have scalaz-stream compiling on cats](https://github.com/rossabaker/scalaz-stream/tree/topic/cats) (requires publishLocal of cats).  [Todo.scala](https://github.com/rossabaker/scalaz-stream/blob/topic/cats/src/main/scala/Todo.scala) lays out what is needed to make it real.

* There is the scalaz-concurrent work that inspired this ticket.  
* Most necessary abstractions already exist in cats.  `Catchable` or `Nondeterminism` don't yet.
* `Either3` doesn't, and probably won't, exist in cats.  Shapeless coproducts solve this, but that's another dependency.
* Cats is missing a few utility methods, like `maximum`.
* `\/` and friends are aliases, pending final decision on non/cats#189.

It's not a small job, but it's viable.  The big decision, as I see it, is whether the base abstractions are sourced from cats-core or scalaz-core.

See also:
* non/cats#21
* non/cats#32

bryce-anderson:
I'm also interested in a ```scalaz-stream``` that unhindered when using ```cats```. I don't have any immediate technical comments that haven't already been discussed.

pchiusano:
I'd rather just cut the dependency entirely, like @mpilquist has done for scodec. I do not want to move scalaz-stream from depending on scalaz to depending on cats (see below). Instead, I'd like to investigate:

[...]

PR #322, opened 2015-02-28
WIP: factor out Scalaz dependency

rossabaker:
This doesn't compile yet, but is far enough along to illuminate #321.

Basic idea:
* shims provides necessary stubs of `scalaz.concurrent._` and trampoline.  This will go away.
* core depends only on scodec.
* scalaz contains the nice integration with scalaz7.  Should live in its own repo when it grows up.  Siblings might include scalaz8, cats, structures, etc.

What's broken:
* A few things exploit Task and ByteVector instances of type classes that were exiled.  These require specialization.
    * Notably, a few things in core run `Process[Task, ?]`, but there is no specialized run.
* Without compiling core, I don't know what dragons await in scalaz.

What stinks:
* Specializations of things we can get for free in multiple places.
* Ours would be at least the third copy of `scalaz.Free.Trampoline` / `cats.free.Trampoline`.
* The return of `scala.Either`, including it being prominent in the shimmed `Task`.
    * To maintain merge sanity for a bit, this is still aliased to the Scalaz names, with some syntactic sugar.
* Import tax due to decoupling instances from companion objects.
* All the code in the scalaz module would be copied, pasted, and gently tweaked for cats, scalaz8, etc.

Discuss.


pchlupacek:
@rossabaker excellent

on the Task, in fact I was thinking to perhaps implement it w/o Future, and that remove \/ from the task alltogether. I didn't see anyone actually using Future directly, so I think we don't need the future at all.  

Part of to take this approach and motivation for me was to make Task very solid and more `deterministic` in case of interruption of asynchronous callbacks. 

I think we may just need temporarily to `shim` task with sort of \/ syntax, so we won't need to refactor all the code. And that for me can live in scalaz._  module.

We also need to think if we really need to have scala.Either as primary primitive in Task (for attempt) perhaps, we can use Cause for this and that way we may not need Either alltogether. Then, it will nicely play with Halt(cause) we use in Process. 

I was thinking on something like 

```scala 

sealed trait Task[+A]

.... all current algebra of Future as is (perhaps we may need to touch async algebra)

sealed trait TaskResult[+A] extends Task[A] 

case class Done[A](a:A) extends TaskResult[A]
case class Failed(c:Cause) extends TaskResult[Nothing]

 def attempt: Task[TaskResult[A]]
 def attemptRun : TaskResult[A] 
 def run: A
 


```

As for the Trampoline, this just prevents stack safety, I would be curious if we really need Trampoline separate from task, or we may reuse task instead of trampoline (Look on number of Try() we use there as the Trampoline does not have such concept)

@pchiusano  your view on this? Any ideas? 

Perhaps we can redefine Future and use it internally, and put Task to scalaz module only for compilation compatibility, but internally use only Future (extended with Cause etc)?

I don't have any strong opinions on any of propositions above, but would like to bring them on table before we just do copy and paste. :-)


I know @pchiusano you  and @runarorama were discussing new IO, can we perhaps use some of your work/ideas here?

rossabaker:
@aloiscochard is also doing some interesting IO work on the scalaz 8 branch.

rossabaker:
I'm working on a cleaner version of this based on more recent discussion.  This PR is obsolete.

PR #324, opened 2015-03-04
Wye clobbered halt fix

### LINKED ISSUE #316: Interaction with wye's causing process not to halt

edwinchoi:
Fixes #316.

In addition to ensuring multiple terminate requests complete, this will also complete the pending `Get`, if there is one, before attempting to kill.

### DIFF
--- a/src/main/scala/scalaz/stream/wye.scala
+++ b/src/main/scala/scalaz/stream/wye.scala
@@ -655,6 +655,8 @@ object wye {
       //cb to be completed for `out` side
       var out: Option[(Cause \/ Seq[O])  => Unit] = None
 
+      var downDone= false
+
       //forward referenced actor
       var a: Actor[M] = null
 
@@ -732,9 +734,6 @@ object wye {
           case None      => None
         }
       }
-
-
-
 
       // Consumes any output form either side and updates wye with it.
       // note it signals if the other side has to be killed
@@ -816,7 +815,6 @@ object wye {
       }
 
 
-
       a = Actor[M]({ m =>
         m match {
           case Ready(side, result) =>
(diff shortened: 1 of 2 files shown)


djspiewak:
<3

pchlupacek:
@edwinchoi  thanks a lot.

alissapajer:
+1 thanks @edwinchoi !

PR #325, opened 2015-03-04
Fixes NPE in io.InputStream when the wrapped process throws an Exception

lukiano:
Also, subsequent reads after close() is called will return -1

### DIFF
--- a/src/main/scala/scalaz/stream/io.scala
+++ b/src/main/scala/scalaz/stream/io.scala
@@ -237,8 +237,12 @@ object io {
         -1
       } else {
         val buffer = new Array[Byte](1)
-        read(buffer)    // if we fail to read only one byte, we're in trouble and have already thrown an exception buffer(0).toInt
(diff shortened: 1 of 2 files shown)


src/main/scala/scalaz/stream/io.scala
```diff
@@ -293,10 +297,12 @@ object io {
     @tailrec
     override def close() {
       if (cur.isHalt && chunks.isEmpty) {
-        chunks = null
+        chunks = Nil
```
djspiewak: Much more reasonable. I had a specific idea behind `null`, but `Nil` serves the purpose much better.

djspiewak:
:+1:

jedws:
@djspiewak @pchlupacek @pchiusano any reason this can't be merged?

djspiewak:
I'm all in favor of merging it, I just wanted to wait for a second sign-off before I did.

PR #326, opened 2015-03-05
Cleaner runAsync without redundant cleanups

djspiewak:
There's really no reason (that I can think of) for `runAsync` to be implemented using raw `Actor`.  This reimplements it using `Task`, also fixing a bug regarding duplicated `onComplete` evaluation, cleaning up the signaling, making the interrupt mechanism more responsive, and also opens the door to a more reasonable type signature down the road (propagating the `Task` context into `njoin` and `wye`).

The first of the two really notable changes here are that we are now using `Task`'s `runAsyncInterruptibly`, which gives us the ability to kill a `Task` (within limits).  This is a strict improvement over the old `Process#runAsync`, which would literally leak threads whenever a `Task` continued running for a long time after the interrupt.  This happens, for example, if we do a `Process.await(p.run) { ... }`, where `p` is some long-running process, running `Process#runAsync` on that result would leak a LOT of computational resources.  No more!

The second change is that we are no longer invoking finalizers multiple times.  This is a really subtle issue, but you can see it here:

```scala
              case Interrupt(cause) if completed.isEmpty =>
                completed = Some(cause)
                Try(cleanup(cause)).run.runAsync(_.fold(
                  rsn0 =>  cb(left(Error(rsn0).causedBy(cause)))
                  , _ => cb(left(cause))
                ))
```

…and here!

```scala
              case AwaitDone(r, awt, _) =>
                Try(awt.rcv(EarlyCause(r)).run)
                .kill
                .run.runAsync(_ => ())
```

Combined with this:

```scala
              case Step(awt@Await(req, rcv), cont) =>
                req.runAsync(r => a ! AwaitDone(r, awt, cont))
                right((c:EarlyCause) => rcv(left(c)).run +: cont)
```

So with the *old* `runAsync`, when we interrupt a process, we immediately run its cleanup, but we don't terminate the task!  Thus, the task will eventually complete… and then run the whole process again.  Hurray!

This PR fixes several of the issues that we (@alissapajer and I) have seen in concurrent code, including several out the outstanding "`wye` issues".

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1239,8 +1239,9 @@ object Process extends ProcessInstances {
 
                 case \/-(p) => {
                   // pointer equality test for convenience; not actually needed
-                  if (ref.compareAndSet(liftedInterrupt, null)) {      // can't swap in runAsync right await, because side effects!
-                    ref.set(p.runAsync(cb))      // we made it! swap out interrupt
+                  if (ref.compareAndSet(liftedInterrupt, null)) {
+                    // we need to fail-trampoline to avoid infinite recursion on drained processes
+                    cb(\/-((Nil, Cont(Vector({ _ => Trampoline done p })))))
                   } else {
                     ()      // interrupt happened after we completed but before we checked; no-op
                   }
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1206,9 +1206,18 @@ object Process extends ProcessInstances {
          * and get the interrupt to take effect!  This reference is checked immediately prior to ever
          * await task evaluation, which effectively gives us the ability to interrupt pre-step in
          * addition to our mid-step interrupt.
+         *
+         * Note that this reference will be set exactly once and never again.
          */
         val interrupted = new AtomicReference[Option[EarlyCause]](None)
 
+        /*
+         * Produces the evaluation for a single step.  Generally, this function will be
+         * invoked only once and return immediately.  In the case of an `Await`, we must
+         * descend recursively into the resultant child.  Generally speaking, the recursion
+         * should be extremely shallow, since it is uncommon to have a chain of nested
+         * awaits of any significant length (usually they are punctuated by an `Emit`).
+         */
         def go(p: Process[Task, O]): Task[EarlyCause => Unit] = Task delay {
           p.step match {
             case Halt(cause) =>
@@ -1220,65 +1229,72 @@ object Process extends ProcessInstances {
             case Step(awt: Await[Task, a, O], cont) => {
               val Await(req, rcv) = awt
 
+              def unpack(msg: Option[_ \/ a]): Option[a] = msg flatMap { _.toOption }
+              def wrap(result: a): Process[Task, O] = Try(rcv(\/-(result)).run)
+
+              // handler extractors
+              object Interrupted {
+
+                object PreOrMidStep {
+                  def unapply(msg: Option[Throwable \/ _]): Option[EarlyCause] =
+                    (PreStep unapply msg) orElse (MidStep unapply msg)
+                }
+
+                // interrupted via the `Task.fail` defined in `checkInterrupt`
+                object PreStep {
+                  def unapply(msg: Option[Throwable \/ _]): Option[EarlyCause] = msg match {
+                    case Some(-\/(Terminated(cause: EarlyCause))) => Some(cause)
+                    case _ => None
+                  }
+                }
+
+                // interrupted via the callback mechanism, checked in `completeInterruptibly`
+                object MidStep {
+                  def unapply(msg: Option[_]): Option[EarlyCause] = interrupted.get() filter { _ => !msg.isDefined }
+                }
+
+                // completed the task, `interrupted.get()` is defined, and so we were interrupted post-completion
+                object PostStep {
+                  def unapply(msg: Option[_ \/ a]): Option[(Process[Task, O], EarlyCause)] = {
+                    val maybeCause = interrupted.get()
+
+                    maybeCause flatMap { cause =>
+                      unpack(msg) map wrap map { (_, cause) }
+                    }
+                  }
+                }
+              }
+
+              // nominally completed the task, but with an exception
+              object Exceptional {
+                def unapply(msg: Option[Throwable \/ _]): Option[Throwable] = msg match {
+                  case Some(-\/(t)) => Some(t)
+                  case _ => None
+                }
+              }
+
+              // completed the task, no interrupts, no exceptions, good to go!
+              object Completed {
+                def unapply(msg: Option[_ \/ a]): Option[Process[Task, O]] =
+                  unpack(msg) map wrap map { _ +: cont } // filter { _ => !interrupted.get().isDefined }   (commented out to avoid race conditions)
+              }
+
               // throws an exception if we're already interrupted
               val checkInterrupt: Task[Unit] = Task delay {
                 interrupted.get() map { c => Task fail Terminated(c) } getOrElse (Task now (()))
               } join
 
               Task delay {
+                // points to either the current step interrupt, or the child's if we have completed and recursed
                 val ref = new AtomicReference[EarlyCause => Unit]
 
-                def unpack(msg: Option[_ \/ a]): Option[a] = msg flatMap { _.toOption }
-
-                def wrap(result: a): Process[Task, O] = Try(rcv(\/-(result)).run)
-
-                // handler extractors
-                object Interrupted {
-
-                  object PreOrMidStep {
-                    def unapply(msg: Option[Throwable \/ _]): Option[EarlyCause] =
-                      (PreStep unapply msg) orElse (MidStep unapply msg)
-                  }
-
-                  // interrupted via the `Task.fail` defined in `checkInterrupt`
-                  object PreStep {
-                    def unapply(msg: Option[Throwable \/ _]): Option[EarlyCause] = msg match {
-                      case Some(-\/(Terminated(cause: EarlyCause))) => Some(cause)
-                      case _ => None
-                    }
-                  }
-
-                  // interrupted via the callback mechanism, checked in `completeInterruptibly`
-                  object MidStep {
-                    def unapply(msg: Option[_]): Option[EarlyCause] = interrupted.get() filter { _ => !msg.isDefined }
-                  }
-
-                  // completed the task, `interrupted.get()` is defined, and so we were interrupted post-completion
-                  object PostStep {
-                    def unapply(msg: Option[_ \/ a]): Option[(Process[Task, O], EarlyCause)] = {
-                      val maybeCause = interrupted.get()
-
-                      maybeCause flatMap { cause =>
-                        unpack(msg) map wrap map { (_, cause) }
-                      }
-                    }
-                  }
-                }
-
-                // nominally completed the task, but with an exception
-                object Exceptional {
-                  def unapply(msg: Option[Throwable \/ _]): Option[Throwable] = msg match {
-                    case Some(-\/(t)) => Some(t)
-                    case _ => None
-                  }
-                }
-
-                // completed the task, no interrupts, no exceptions, good to go!
-                object Completed {
-                  def unapply(msg: Option[_ \/ a]): Option[Process[Task, O]] =
-                    unpack(msg) map wrap map { _ +: cont } // filter { _ => !interrupted.get().isDefined }   (commented out to avoid race conditions)
-                }
-
+                /*
+                 * Start the task. per the `completeInterruptibly` invariants, the callback will be invoked exactly once
+                 * unless interrupted in the final computation step, in which case it will be invoked twice: once with
+                 * the interrupt signal and once with the final computed result (this case is detected by the `PostStep`)
+                 * extractor.  Under all other circumstances, including interrupts, exceptions and natural completion, the
+                 * callback will be invoked exactly once.
+                 */
                 val interrupt = completeInterruptibly((Task fork (checkInterrupt >> req)).get) {
                   case Interrupted.PreOrMidStep(cause) => {
                     // interrupted; now drain
@@ -1301,6 +1317,7 @@ object Process extends ProcessInstances {
 
                 ref.set({ _ => interrupt() })
 
+                // interrupts the current step (may be a recursive child!) and sets `interrupted`
                 def referencedInterrupt(cause: EarlyCause): Unit = {
                   interrupted.compareAndSet(None, Some(cause))
                   ref.get()(cause)
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1,12 +1,17 @@
 package scalaz.stream
 
 import Cause._
+
+import java.util.concurrent.atomic.AtomicReference
+
 import scala.annotation.tailrec
 import scala.collection.SortedMap
-import scalaz.{Catchable, Functor, Monad, Monoid, Nondeterminism, \/, -\/, ~>}
+
+import scalaz.{Catchable, Functor, Monad, Monoid, Nondeterminism, \/, -\/, \/-, ~>}
 import scalaz.\/._
 import scalaz.concurrent.{Actor, Strategy, Task}
 import scalaz.stream.process1.Await1
+import scalaz.syntax.monad._
 
 /**
  * An effectful stream of `O` values. In between emitting values
@@ -1192,93 +1197,68 @@ object Process extends ProcessInstances {
      * @param S  Strategy to use when evaluating the process. Note that `Strategy.Sequential` may cause SOE.
      * @return   Function to interrupt the evaluation
      */
-    protected[stream] final def runAsync(
-      cb: Cause \/ (Seq[O], Cont[Task,O]) => Unit
-      )(implicit S: Strategy): (EarlyCause) => Unit = {
+    protected[stream] final def runAsync(cb: Cause \/ (Seq[O], Cont[Task,O]) => Unit)(implicit S: Strategy): EarlyCause => Unit = {
+      lazy val asyncStep: Task[EarlyCause => Unit] = Task delay {
+        self.step match {
+          case Halt(cause) =>
+            (Task delay cb(-\/(cause))) >> (Task now { _: EarlyCause => () })
 
-          sealed trait M
-          case class AwaitDone(res: Throwable \/ Any, awt: Await[Task, Any, O], cont: Cont[Task,O]) extends M
-          case class Interrupt(cause: EarlyCause) extends M
+          case Step(Emit(os), cont) =>
+            (Task delay { cb(\/-((os, cont))) }) >> (Task now { _: EarlyCause => () })
 
-          //forward referenced actor here
-          var a: Actor[M] = null
+          case Step(Await(req, rcv), cont) => {
+            val await: Task[Process[Task, O]] = req.attempt map { either =>
+              Try(rcv(EarlyCause(either)).run) +: cont
+            }
 
-          // Set when the executin has been terminated with reason for termination
-          var completed: Option[Cause] = None
+            Task delay {
+              val ref = new AtomicReference[EarlyCause => Unit]
 
-          // contains reference that eventually builds
-          // a cleanup when the last await was interrupted
-          // this is consulted only, if await was interrupted
-          // volatile marked because of the first usage outside of actor
-          @volatile var cleanup: (EarlyCause => Process[Task,O]) = (c:EarlyCause) => Halt(c)
+              @volatile
+              var orig: EarlyCause => Unit = null     // forward reference
 
-          // runs single step of process.
-          // completes with callback if process is `Emit` or `Halt`.
-          // or asynchronously executes the Await and send result to actor `a`
-          // It returns on left side reason with which this process terminated,
-          // or on right side the cleanup code to be run when interrupted.
-          @tailrec
-          def runStep(p: Process[Task, O]): Cause \/ (EarlyCause => Process[Task,O]) = {
-            val step = p.step
-            step match {
-              case Step(Emit(Seq()), cont)         => runStep(cont.continue)
-              case Step(Emit(h), cont)             => S(cb(right((h, cont)))); left(End)
-              case Step(awt@Await(req, rcv), cont) =>
-                req.runAsync(r => a ! AwaitDone(r, awt, cont))
-                right((c:EarlyCause) => rcv(left(c)).run +: cont)
-              case Halt(cause)                 => S(cb(left(cause))); left(cause)
+              val interrupt = await runAsyncInterruptibly {
+                case -\/(t) => {
+                  if (ref.compareAndSet(orig, null)) {
+                    cb(-\/(Error(t)))
+                  } else {
+                    ()      // we got an exception (possibly an interrupt); cause already propagated
+                  }
+                }
+
+                case \/-(p) => {
+                  // pointer equality test for convenience; not actually needed
+                  if (ref.compareAndSet(orig, null)) {      // can't swap in runAsync right await, because side effects!
+                    ref.set(p.runAsync(cb))      // we made it! swap out interrupt
+                  } else {
+                    ()      // interrupt happened after we completed but before we checked; no-op
+                  }
+                }
+              }
+
+              def liftedInterrupt(cause: EarlyCause): Unit = {
+                if (ref.compareAndSet(orig, null)) {
+                  interrupt()
+                  cb(-\/(cause))
+                } else {
+                  referencedInterrupt(cause)      // completed naturally just as we were interrupted; forward along
+                }
+              }
+
+              orig = liftedInterrupt _
+              ref.set(orig)
+
+              def referencedInterrupt(cause: EarlyCause): Unit =
+                ref.get()(cause)
+
+              referencedInterrupt _
             }
           }
-
-
-          a = new Actor[M]({ m =>
-            m match {
-              case AwaitDone(r, awt, cont) if completed.isEmpty =>
-                val step = Try(awt.rcv(EarlyCause(r)).run) +: cont
-
-
-                runStep(step).fold(
-                  rsn => completed = Some(rsn)
-                  , cln => cleanup = cln
-                )
-
-              // on interrupt we just run any cleanup code we have memo-ed
-              // from last `Await`
-              case Interrupt(cause) if completed.isEmpty =>
-                completed = Some(cause)
-                Try(cleanup(cause)).run.runAsync(_.fold(
-                  rsn0 =>  cb(left(Error(rsn0).causedBy(cause)))
-                  , _ => cb(left(cause))
-                ))
-
-              // this indicates last await was interrupted.
-              // In case the request was successful and only then
-              // we have to get next state of the process and assure
-              // any cleanup will be run.
-              // note this won't consult any cleanup contained
-              // in `next` or `rcv` on left side
-              // as this was already run on `Interrupt`
-              case AwaitDone(r, awt, _) =>
-                Try(awt.rcv(EarlyCause(r)).run)
-                .kill
-                .run.runAsync(_ => ())
-
-
-              // Interrupt after we have been completed this is no-op
-              case Interrupt(_) => ()
-
-            }
-          })(S)
-
-          runStep(self).fold(
-            rsn => (_: Cause) => ()
-            , cln => {
-              cleanup = cln
-              (cause: EarlyCause) => a ! Interrupt(cause)
-            }
-          )
         }
+      } join
 
+      asyncStep.run   // hey, we could totally return something sane here! what up?
+    }
   }
 
   /**
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1198,13 +1198,13 @@ object Process extends ProcessInstances {
      * @return   Function to interrupt the evaluation
      */
     protected[stream] final def runAsync(cb: Cause \/ (Seq[O], Cont[Task,O]) => Unit)(implicit S: Strategy): EarlyCause => Unit = {
-      def go(p: Process[Task, O]): Task[EarlyCause => Unit] = Task delay {
+      def go(p: Process[Task, O])(cb: Cause \/ (Seq[O], Cont[Task,O]) => Unit): Task[EarlyCause => Unit] = Task delay {
         p.step match {
           case Halt(cause) =>
-            (Task delay cb(-\/(cause))) >> (Task now { _: EarlyCause => () })
+            (Task delay { S { cb(-\/(cause)) } }) >> (Task now { _: EarlyCause => () })
 
           case Step(Emit(os), cont) =>
-            (Task delay { cb(\/-((os, cont))) }) >> (Task now { _: EarlyCause => () })
+            (Task delay { S { cb(\/-((os, cont))) } }) >> (Task now { _: EarlyCause => () })
 
           case Step(Await(req, rcv), cont) => {
             val await: Task[Process[Task, O]] = req.attempt map { either =>
@@ -1212,23 +1212,44 @@ object Process extends ProcessInstances {
             }
 
             Task delay {
+              val interrupted = new AtomicReference[Option[EarlyCause]](None)      // this is necessary to catch interrupts that are late in halting very fast tasks
               val ref = new AtomicReference[EarlyCause => Unit]
 
               lazy val interrupt: () => Unit = await runAsyncInterruptibly {
+                // explicitly catch interrupts
+                case -\/(Task.TaskInterrupted) => ()      // we were interrupted; someone else will kill things off (this is part 1 of the double-cleanup case)
+
                 case -\/(t) => {
                   if (ref.compareAndSet(liftedInterrupt, null)) {
-                    go(cont.continue.injectCause(Error(t)).drain.causedBy(Error(t)))
+                    ref.compareAndSet(null, go(Try(rcv(-\/(Error(t))).run) +: cont)(cb).run)      // analogous to the success case
                   } else {
-                    ()      // we got an exception (possibly an interrupt); cause already propagated
+                    ()      // we got an exception (apparently an interrupt?); cause already propagated
                   }
                 }
 
                 case \/-(p) => {
                   // pointer equality test for convenience; not actually needed
                   if (ref.compareAndSet(liftedInterrupt, null)) {
-                    // we need to fail-trampoline to avoid infinite recursion on drained processes
-                    cb(\/-((Nil, Cont(Vector({ _ => Trampoline done p })))))
+                    val recurse = go(p) {
+                      case -\/(cause) => cb(-\/(cause))
+
+                      // we completed the task, then got interrupted before we invoked the callback; discard results and drain!
+                      case \/-((_, cont)) if interrupted.get().isDefined => {
+                        val cause = interrupted.get().get
+
+                        cont.continue.injectCause(cause).drain.run runAsync {
+                          case -\/(t) => S { cb(-\/(Error(t) causedBy cause)) }
+                          case \/-(_) => S { cb(-\/(cause)) }
+                        }
+                      }
+
+                      case \/-(pair) => cb(\/-(pair))
+                    }
+
+                    ref.compareAndSet(null, recurse.run)
                   } else {
+                    // TODO this is part 2 of the case where the double-cleanup happens in master
+
                     ()      // interrupt happened after we completed but before we checked; no-op
                   }
                 }
@@ -1238,7 +1259,11 @@ object Process extends ProcessInstances {
                 if (ref.compareAndSet(liftedInterrupt, null)) {
                   interrupt()
 
-                  go(cont.continue.injectCause(cause).drain.causedBy(cause))
+                  // successfully interrupted, now drain the rest of the process
+                  (Try(rcv(-\/(cause)).run) +: cont).run runAsync {
+                    case -\/(t) => S { cb(-\/(Error(t) causedBy cause)) }
+                    case \/-(_) => S { cb(-\/(cause)) }
+                  }
                 } else {
                   referencedInterrupt(cause)      // completed naturally just as we were interrupted; forward along
                 }
@@ -1247,8 +1272,11 @@ object Process extends ProcessInstances {
               ref.set(liftedInterrupt)
               interrupt        // force the lazy val
 
-              def referencedInterrupt(cause: EarlyCause): Unit =
+              // note there is a real chance of a NPE coming out of here under interrupt contention, and I'm ok with that because the types are stupid!
+              def referencedInterrupt(cause: EarlyCause): Unit = {
+                interrupted.set(Some(cause))
                 ref.get()(cause)
+              }
 
               referencedInterrupt _
             }
@@ -1256,7 +1284,7 @@ object Process extends ProcessInstances {
         }
       } join
 
-      go(self).run   // hey, we could totally return something sane here! what up?
+      go(self)(cb).run   // hey, we could totally return something sane here! what up?
     }
   }
 
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1300,38 +1300,57 @@ object Process extends ProcessInstances {
 
     /**
      * Analogous to Future#listenInterruptibly, but guarantees listener notification provided that the
-     * body of any given computation step does not block indefinitely.  Furthermore, if the computation
-     * has already completed by the time the cancelation signal is set to `true`.
+     * body of any given computation step does not block indefinitely.  When the interrupt function is
+     * invoked, the callback will be immediately invoked, either with an available completion value or
+     * with None.  If the current step of the task ultimately completes with its *final* value (i.e.
+     * the final step of the task is an Async and it starts before the interrupt and completes *afterwards*),
+     * that value will be passed to the callback as a second return.  Thus, the callback will always be
+     * invoked at least once, and may be invoked twice.  If it is invoked twice, the first callback
+     * will always be None while the second will be Some.
      */
-    private def completeInterruptibly[A](f: Future[A], cancel: AtomicBoolean)(cb: Option[A] => Unit): Unit = {
+    private def completeInterruptibly[A](f: Future[A])(cb: Option[A] => Unit)(implicit S: Strategy): () => Unit = {
       import Future._
 
-      f match {
+      val cancel = new AtomicBoolean(false)
+
+      lazy val actor: Actor[Option[Future[A]]] = new Actor[Option[Future[A]]]({
         // pure cases
-        case Suspend(thunk) if !cancel.get() => completeInterruptibly(thunk(), cancel)(cb)
-        case BindSuspend(thunk, g) if !cancel.get() => completeInterruptibly(thunk() flatMap g, cancel)(cb)
+        case Some(Suspend(thunk)) if !cancel.get() =>
+          actor ! Some(thunk())
 
-        case Now(a) => cb(Some(a))
+        case Some(BindSuspend(thunk, g)) if !cancel.get() =>
+          actor ! Some(thunk() flatMap g)
 
-        case Async(onFinish) if !cancel.get() => {
+        case Some(Now(a)) => S { cb(Some(a)) }
+
+        case Some(Async(onFinish)) if !cancel.get() => {
           onFinish { a =>
-            Trampoline delay { cb(Some(a)) }
+            Trampoline delay { S { cb(Some(a)) } }
           }
         }
 
-        case BindAsync(onFinish, g) if !cancel.get() => {
+        case Some(BindAsync(onFinish, g)) if !cancel.get() => {
           onFinish { a =>
             if (!cancel.get()) {
-              Trampoline delay { g(a) } map { completeInterruptibly(_, cancel)(cb) }
+              Trampoline delay { g(a) } map { r => actor ! Some(r) }
             } else {
-              Trampoline delay { cb(None) }
+              Trampoline delay { S { cb(None) } }
             }
           }
         }
 
-        // fall-through case where cancel.get() is true
-        case _ => cb(None)
-      }
+        // fallthrough case where cancel.get() == true
+        case Some(_) => S { cb(None) }
+
+        case None => {
+          cancel.set(true)
+          S { cb(None) }
+        }
+      })
+
+      actor ! Some(f)
+
+      { () => actor ! None }
     }
   }
 
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1198,8 +1198,8 @@ object Process extends ProcessInstances {
      * @return   Function to interrupt the evaluation
      */
     protected[stream] final def runAsync(cb: Cause \/ (Seq[O], Cont[Task,O]) => Unit)(implicit S: Strategy): EarlyCause => Unit = {
-      val asyncStep: Task[EarlyCause => Unit] = Task delay {
-        self.step match {
+      def go(p: Process[Task, O]): Task[EarlyCause => Unit] = Task delay {
+        p.step match {
           case Halt(cause) =>
             (Task delay cb(-\/(cause))) >> (Task now { _: EarlyCause => () })
 
@@ -1214,24 +1214,10 @@ object Process extends ProcessInstances {
             Task delay {
               val ref = new AtomicReference[EarlyCause => Unit]
 
-              // forward reference
-              var interrupt: () => Unit = { () => () }
-
-              lazy val liftedInterrupt: EarlyCause => Unit = { cause =>
-                if (ref.compareAndSet(liftedInterrupt, null)) {
-                  interrupt()
-                  cb(-\/(cause))
-                } else {
-                  referencedInterrupt(cause)      // completed naturally just as we were interrupted; forward along
-                }
-              }
-
-              ref.set(liftedInterrupt)
-
-              interrupt = await runAsyncInterruptibly {
+              lazy val interrupt: () => Unit = await runAsyncInterruptibly {
                 case -\/(t) => {
                   if (ref.compareAndSet(liftedInterrupt, null)) {
-                    cb(-\/(Error(t)))
+                    go(cont.continue.injectCause(Error(t)).drain.causedBy(Error(t)))
                   } else {
                     ()      // we got an exception (possibly an interrupt); cause already propagated
                   }
@@ -1248,6 +1234,19 @@ object Process extends ProcessInstances {
                 }
               }
 
+              lazy val liftedInterrupt: EarlyCause => Unit = { cause =>
+                if (ref.compareAndSet(liftedInterrupt, null)) {
+                  interrupt()
+
+                  go(cont.continue.injectCause(cause).drain.causedBy(cause))
+                } else {
+                  referencedInterrupt(cause)      // completed naturally just as we were interrupted; forward along
+                }
+              }
+
+              ref.set(liftedInterrupt)
+              interrupt        // force the lazy val
+
               def referencedInterrupt(cause: EarlyCause): Unit =
                 ref.get()(cause)
 
@@ -1257,7 +1256,7 @@ object Process extends ProcessInstances {
         }
       } join
 
-      asyncStep.run   // hey, we could totally return something sane here! what up?
+      go(self).run   // hey, we could totally return something sane here! what up?
     }
   }
 
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1214,12 +1214,23 @@ object Process extends ProcessInstances {
             Task delay {
               val ref = new AtomicReference[EarlyCause => Unit]
 
-              @volatile
-              var orig: EarlyCause => Unit = null     // forward reference
+              // forward reference
+              var interrupt: () => Unit = { () => () }
 
-              val interrupt = await runAsyncInterruptibly {
+              lazy val liftedInterrupt: EarlyCause => Unit = { cause =>
+                if (ref.compareAndSet(liftedInterrupt, null)) {
+                  interrupt()
+                  cb(-\/(cause))
+                } else {
+                  referencedInterrupt(cause)      // completed naturally just as we were interrupted; forward along
+                }
+              }
+
+              ref.set(liftedInterrupt)
+
+              interrupt = await runAsyncInterruptibly {
                 case -\/(t) => {
-                  if (ref.compareAndSet(orig, null)) {
+                  if (ref.compareAndSet(liftedInterrupt, null)) {
                     cb(-\/(Error(t)))
                   } else {
                     ()      // we got an exception (possibly an interrupt); cause already propagated
@@ -1228,25 +1239,13 @@ object Process extends ProcessInstances {
 
                 case \/-(p) => {
                   // pointer equality test for convenience; not actually needed
-                  if (ref.compareAndSet(orig, null)) {      // can't swap in runAsync right await, because side effects!
+                  if (ref.compareAndSet(liftedInterrupt, null)) {      // can't swap in runAsync right await, because side effects!
                     ref.set(p.runAsync(cb))      // we made it! swap out interrupt
                   } else {
                     ()      // interrupt happened after we completed but before we checked; no-op
                   }
                 }
               }
-
-              def liftedInterrupt(cause: EarlyCause): Unit = {
-                if (ref.compareAndSet(orig, null)) {
-                  interrupt()
-                  cb(-\/(cause))
-                } else {
-                  referencedInterrupt(cause)      // completed naturally just as we were interrupted; forward along
-                }
-              }
-
-              orig = liftedInterrupt _
-              ref.set(orig)
 
               def referencedInterrupt(cause: EarlyCause): Unit =
                 ref.get()(cause)
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1198,7 +1198,7 @@ object Process extends ProcessInstances {
      * @return   Function to interrupt the evaluation
      */
     protected[stream] final def runAsync(cb: Cause \/ (Seq[O], Cont[Task,O]) => Unit)(implicit S: Strategy): EarlyCause => Unit = {
-      lazy val asyncStep: Task[EarlyCause => Unit] = Task delay {
+      val asyncStep: Task[EarlyCause => Unit] = Task delay {
         self.step match {
           case Halt(cause) =>
             (Task delay cb(-\/(cause))) >> (Task now { _: EarlyCause => () })
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -2,14 +2,14 @@ package scalaz.stream
 
 import Cause._
 
-import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
 
 import scala.annotation.tailrec
 import scala.collection.SortedMap
 
 import scalaz.{Catchable, Functor, Monad, Monoid, Nondeterminism, \/, -\/, \/-, ~>}
 import scalaz.\/._
-import scalaz.concurrent.{Actor, Strategy, Task}
+import scalaz.concurrent.{Actor, Future, Strategy, Task}
 import scalaz.stream.process1.Await1
 import scalaz.syntax.monad._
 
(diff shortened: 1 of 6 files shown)


djspiewak:
Also I don't know how to spell "Santa Claus"…

alissapajer:
I have verified (with much excitement!) that this PR fixes the issue #316 . This issue was at first believed to be an issue with `wye`, but now it has been shown to be a problem with `runAsync` that this PR resovles!

djspiewak:
In case it is helpful in the review, here is a pseudocode state machine breakdown of the *old* `runAsync` that I used to match behavior in the new implementation.  The double-run finalizers bug is apparent (and noted):

- `Emit(Nil, cont)` => *loop*
- `Emit(os, cont)` => callback(`os`, `cont`)
- `Await`
    + Completed => *loop* on `rcv(EarlyCause(either)) +: cont`
    + Interrupted => `rcv(-\/(cause)) +: cont` `run.runAsync`
        * Failed => callback(`Error(cause2).causedBy(cause)`)
        * Completed => callback(`cause`)
    + Completed (post-interrupt) => `rcv(EarlyCause(either)).kill` `run.runAsync`
        * *I'm pretty sure this is wrong!*  This will result in the `Await` cleanup being run multiple times.  Running it once should be sufficient.
    + Interrupted (post-interrupt) => no-op
- `Halt(cause)` => callback(`cause`)

djspiewak:
Hrm, Travis seems to think that one of Eduin's test cases doesn't work anymore. I can't get it to fail locally; can anyone reproduce?

djspiewak:
I added a spec which demonstrates the issue with `onComplete` being run multiple times.  This spec fails in master and passes on this branch.

edwinchoi:
Doesn't `runAsyncInterruptibly` simply complete the callback with an exception on interrupt, discarding the result once completed? That's what the implementation of `Task` seems to show... which is really no different than what the current `runAsync` does, except `runAsync` will guarantees cleanup on the result.

djspiewak:
It aborts out at the earliest trampoline point.  That may be running to the end, but if the task is structured idiomatically, it will be much sooner than that.

There's no question that this implementation of `runAsync` can leak resources on the await side.  However, the old implementation required that all finalizers be idempotent, which is just as undesirable and in fact generated deadlock bugs even outside of wye.  Seeing as *both* implementations are resource unsafe under certain circumstances and the original generates deadlocks, I'm in favor of going with this revision until we can find a better approach.

One thing we do have the freedom to do, now that the cases are well separated in the implementation, is reimplement the child cleanup logic but without the duplicate running.  We're ultimately never going to be able to get around the fact that we are asking a concurrent process to abort halfway, but perhaps that will be a slightly superior trade off.

alissapajer:
+1 to merge!

pchlupacek:
@djspiewak great work. 


Before merging this I would like little bit to discuss the leakage of await. Because perhaps, this is not issue at all. 

Lets see two codes below:

```scala 

val asyncAllocation = Task.async(???).flatMap { resource => ...  } //bad code
val goodAllocation = await(Task.async(???), resource => .... ) //good code

```

I think we usually write resource allocation by good code, and that shall give us opportunity to ALWAYS run necessary cleanups after  asynchronous task. is interrupted (as we will receive Interrupted exception)

Do you see this same way? 

We rely in our server code on fact that `onComplete`is always called. And need to be sure that this change won't ever introduce situation when this won't be the case.

djspiewak:
> I think we usually write resource allocation by good code, and that shall give us opportunity to ALWAYS run necessary cleanups after asynchronous task. is interrupted (as we will receive Interrupted exception)
>
> Do you see this same way?

Usually yes.  The weird cases which current exist that allocate resources in non-singular `Task` blocks are ear-marked for refactoring, because they really aren't safe.

> We rely in our server code on fact that onCompleteis always called. And need to be sure that this change won't ever introduce situation when this won't be the case.

At present, I believe there *is* a case with my changes where `onComplete` will not be invoked.  However, I think I have sufficient state information to implement an "exactly once" semantic for the invocation, which should give us the property I want (no need for idempotency!) and thus the avoidance of the deadlocks, while still avoiding resource leakage in circumstances other than your "bad code" example.

pchlupacek:
I think we have to clearly define when onComplete with your changes won't be invoked. I personally think if we take the assumption that allocation of resources in task itself is unsafe and may cause the onComplete to be not invoked for that kind of allocations, I assume we can implement runAsync completely safe and as you say where onComplete is invoked exactly once. After that I would like to merge this in master. 

Going further in future I would like to change signature of async callback on Task/Future from A => Unit to A => Boolean yielding to true, if task was already completed by either normal completion or by interruption. That will resolve the deadlocks we see on Task and give us more control on `Queue` at all where we use callbacks to not fill dead tasks.

djspiewak:
> Going further in future I would like to change signature of async callback on Task/Future from A => Unit to A => Boolean yielding to true, if task was already completed by either normal completion or by interruption. That will resolve the deadlocks we see on Task and give us more control on Queue at all where we use callbacks to not fill dead tasks.

Actually, I can give you that information right now.  Once I finish the changes I'm working on at present, it'll be clear how it is possible to make that determination with the existing `Task` signature.  I would be happy to fold in a simple type signature change to this PR if you want, especially since I already have the information you're asking for.

Honestly, what I *really* want to do is propagate the `Task` context out of `runAsync` and into the caller.  The type signature of `Process#runAsync` is really weak, and with a `Task`-based implementation, we actually have the flexibility to do quite a bit better.  I know that the call-sites in most cases are fairly mangled, imperative code (`wye` and `njoin`), but one step at a time.

pchlupacek:
@djspiewak thanks a lot. Yeah, if we manage to get task out of runAsync that will help a lot. I think wye shall be perfectly possible (actually we didn't do it just because lack of time :-)) The signature comes from the old experiments with interruption before interruption on Task was solid (I mean like is now). So feel free to improve it as you wish. On other hand I would like to do that in steps, first get this change in master, then work on wye and changing runAsync. Do you agree?

pchlupacek:
And yeah I am excited to see how you solve the signalling of interruption to callback completers w/o change of the signature. That would be excellent.

djspiewak:
> On other hand I would like to do that in steps, first get this change in master, then work on wye and changing runAsync. Do you agree?

Agreed!

> And yeah I am excited to see how you solve the signalling of interruption to callback completers w/o change of the signature. That would be excellent.

Some of my initial experiments fell through actually, so I'm still working on things a bit.  In fact, I'm concerned that I might have to write a new interpreter for `Future`, since `listenInterruptibly` just tosses stuff out.  It's a bit more refined than @edwinchoi suggested, but I think I might need a bit more signaling than what it offers me. 

The good news is that the interpreter I would need to reimplement is trivial, and all of that structure is exposed (hurray for free monads!).  So I don't think it'll be anything more than a speed bump.

edwinchoi:
Whoops, can't undelete messages here...

After thinking about it a bit more, I agree this solution is more favorable.
Both have serious trade offs though. One sacrifices exactly-once termination semantics for at-least-once, the other trades exactly-once transfer semantics for at-most-once. Both allow interleaving of await and cleanup.
Exactly-once semantics and serialization are automatic with sequential eval, but with async eval its up to the implementer to re-establish them.

This would not be an issue if interrupts were applied atomically, but this won't work if the await never halts.

It would be nice if there were a variant of runAsync that could be used when the implementer can guarantee blocking tasks will periodically check for interrupts.

edwinchoi:
Ok I basically just repeated what others had already said :(.
Was looking at a stale version of page..

djspiewak:
So, after more work…  I'm not sure that this test can pass with a correctly implemented `runAsync`:

```scala
  property("runAsync cleanup") = secure {

    val q = async.boundedQueue[Int]()
    val q2 = async.boundedQueue[Int]()

    @volatile var cleanupCalled = false
    val sync = new SyncVar[Cause \/ (Seq[Int], Process.Cont[Task,Int])]
    val deque = q.dequeue.onComplete(eval_(Task.delay{cleanupCalled = true}))
    val interrupt =
      (deque observe q2.enqueue).runAsync(sync.put)

    Thread.sleep(100)
    interrupt(Kill)

    sync.get(3000).isDefined :| "sync completion" && cleanupCalled :| "cleanup"

  }
```

The problem with this test is the fact that it is depending on an interrupt for a `Task` which will never complete!  Thus, the only way that the `onComplete` can ever be invoked is if we have some sort of timeout where we just assume the task is a lost cause and move on without it.  This is effectively what the old `runAsync` used to do, but with the timeout set to zero.

The problem of course is that the task may in fact complete after all!  The old `runAsync` "solved" this problem by re-running the finalizers, which is undesirable on many levels.  Honestly, I'm not sure what the right behavior here is.

It's nothing to do with the interrupt semantics of `Future` either: this problem is fundamental to "safely" interrupting concurrent operations.  Maybe this was obvious to others before (I know this has been touched on in passing), but this is the first time that I really grasped the conflict.  And I'm not sure what to do about it.

For reference, my `SourceSyntax` locally now looks like this: https://gist.github.com/djspiewak/4770c12f2b3f8fb9d0da  You'll notice that the cases are a lot more clearly spelled out, and it's fairly obvious where things can and will get hung up by the test case above.

So what should we do here?  I don't think it's really that uncommon to `await` on a `Task.async` that isn't going to return in the near future (if ever).  In fact, I would suggest that this is more common than any other use case!  Clearly we don't want to run the finalizers multiple times, since that requires idempotency and in fact generates deadlocks in our own code (I mean scalaz-stream).  The only alternative I can think of is to set the interrupt, then a timeout of *n* (for some possibly-zero value of *n*), and then just give up on the task after that timeout.  If the task completes following that timeout, the resources will just leak.

Thoughts?

pchlupacek:
@djspiewak I think this is exactly the place where we need to fix an `Task`. And yeah I think this test was the reason why we implemented runAsync the way as it is, and in fact we have some of these idioms in code. 

I think we need to do it however this time proper way and I suggest following:

1. Introduce own version of Task.runAsyncInterruptibly similarly as yours implementation but with slight changes:
    - The signature of callback will be `Throwable \/ A => Boolean`
    - We will hold Atomic CompareAndSet of callback (instead of boolean) and thus assure the callback wil be invoked only once 
    - On interrupt we ALWAYS complete the task, if task has not been completed yet
    - If interrupted task was completed we assure the callback will be never called again
    - We need atomref to do that (I mean once we obtain the ref, we shall do interruption, otherwise we won't do it)
    - We may need to see if this works fine with the interruption on trampoline as it is implemented today, but I feel it should be ok. 
    - the callback will be always invoked on different thread with S, that means we need Strategy that is perhaps not sequential here (not sure completely about that)

Just as an idea the implementation may be something like 
```scala 

class Task() {
 //...
  // note no atomic bool exposed here just the resulting CB. 
  def runAsyncInterruptibly2(cb: Throwable \/ A)(implicit S:Strategy):() => Unit = {
     val cbReady = new AtomicReference[Option[Throwable \/ Unit]](Some( /*callback from listen interruptibly - maybe we need three states here -before evaluation, after evaluation and when evaluation completed.  */ ))
    //.... here we need to interpret the Future/Task (essentialy what listen does now)
    // ....
    (Throwable \/A) => {
      // bellow code need to be similarly passed to register of listen
      cbReady.getAndSet(None) match {
            case Some(cb) => S(cb(left(Interrupted)))
            case None => true
      }
   }
 }
//... 
}


```



2. Then On runAsync I think we have to implement this semantic
    - on every await we check if task contains any async code. If yes, we will evaluate all to that async code and then remember rcv + append for the async code.
    - then if we will receive interrupt we just interrupt open task
    - that must cause either to receive result or receive `Interrupt` likely on `other` thread, immediatelly
    - Then after receiving one of these we just consult memoized Throwable \/ A => Process[Task,O]
    - Now I am not completely sure if cleanup has to be run in runAsync. It is now, but, imho it shall not be the case as the combinator(mergeN,wye) knows the status of the process and may decide whether to toss all `O` or just continue the evaluation. I am not sure on that either tho. 
    - otherwise just run the callback

That way the callback is run exactly once and we won't miss any onCompletes.

Furthermore I think we have to revisit as second step implementation of queue and async._ and change the call-back to proposed Throwable \/ A => Boolean and assure that if cb returns true, that we would just toss that callback. All that will require that Strategy will be `closed over` in the callback so we would obtain the callback _before_ it will be run  by strategy    

@djspiewak, @edwinchoi  what are you thoughts? Sorry for so much clutter, but its quite complex :-(.

djspiewak:
@pchlupacek I'm not sure I understand what you're driving at.  Thoughts given point-by-point:

> The signature of callback will be `Throwable \/ A => Boolean`

This is probably the crux of what I don't understand.  What does the `Boolean` signify?  My current implementation uses a callback signature `Option[Throwable \/ A] => Unit`, where the `None` case indicates that the `Task` was interrupted.

Unless you mean the callback from `Process#runAsync`?  Still not sure what the signature means though.

> We will hold Atomic CompareAndSet of callback (instead of boolean) and thus assure the callback wil be invoked only once

My current implementation already has this property.

> On interrupt we ALWAYS complete the task, if task has not been completed yet

Yep, I have this property as well.  I think this is a very, very important thing to preserve.  If the task has been completed, we give the value (or exception).  If the task is at a trampoline point and has *not* been completed, then it invokes the callback with `None`.

The biggest problem is that, at present, when we interrupt the task we do not interrupt *immediately*.  If the task is still running a step (either `Async` or `BindAsync`), then the interrupt doesn't get detected until after the completion of that step, and herein lies the problem.

> We need atomref to do that (I mean once we obtain the ref, we shall do interruption, otherwise we won't do it)

Why do we need the atomic reference?  I guess if we respond to the interrupt by *immediately* invoking the callback (similar to `Task#runAsyncInterruptibly`) and then swallow the completion value later?  This is a slight improvement over the current `master` implementation, which will always run the task to completion, but it's not really an improvement over the implementation currently in this PR (not my gist), which doesn't guarantee `onComplete` invocation.

> the callback will be always invoked on different thread with S, that means we need Strategy that is perhaps not sequential here (not sure completely about that)

Easy enough to do.  Not sure why it's necessary though?  I think I'm still missing something in your thought process.

> on every await we check if task contains any async code. If yes, we will evaluate all to that async code and then remember rcv + append for the async code.

Easy enough.  Right now, we convert sync code into async code (with `fork`), but other implementations are pretty direct.

> That way the callback is run exactly once and we won't miss any onCompletes

This is an easy thing to guarantee.  The *hard* thing to guarantee is always cleaning up values produced by the `Task`, assuming they are in fact produced!  The pair of properties I would *like* to guarantee is the following:

* Given that `Task` ultimately completes with a value `a`, whether interrupted or otherwise, then `onComplete` is invoked on value `a`
* Any `onComplete` is invoked exactly once

The current master guarantees the former but not the latter.  The current PR guarantees the latter.  My gist guarantees neither.  Honestly, I'm not sure it's possible to guarantee these two properties together.  Does your scheme address this in a fashion that I didn't understand?

pchlupacek:
@djspiewak More less this was "thinking loud" however couple of things than we need to consider:

(1) only task that needs interruption is async task. All other tasks must not be interrupted 
(2) current implementation of async.Queue will complete callbacks from tasks that are interrupted, effectively by fulfilling any task. The idea was to have `Throwable \/ A => Boolean` to signal to queue that task was interrupted (`true` task was interrupted and as such is safe to pass `A` to another call back).
 (3) because of (1) I would like to allow any continuation of any code that may follow asynchronous task  to be executed when task was interrupted, but if that task evaluates again to asynchronous code, that code must not be executed, if task was interrupted before, instead task must finish its callback with `Interrupt`.

I think if we are able to guarantee implementation of these we may be able to guarantee that task completes only once, at its earliest possible time.

That will give us chance to always run cleanups on await's task completion. Current implementation runs these cleanups on `interrupt` and/or after interrupted task completes and thats the cause of all problems. 

Additionally this would give us chance to implement correctly async.Queue in future and probably will lead to further improvements. I am still not sure whether current behaviour of async.Queue is problem or not, but will feel more comfortable if we would send `A` only to tasks that may have chance to process it. 

Not sure if I explained that more clearly now.

djspiewak:
> (1) only task that needs interruption is async task. All other tasks must not be interrupted

The only tasks that *can* be meaningfully interrupted are `Async` or `BindAsync`.  So, we're basically covered there.

> (2) current implementation of async.Queue will complete callbacks from tasks that are interrupted, effectively by fulfilling any callback. The idea was to have Throwable \/ A => Boolean to signal to queue that task was interrupted (true task was interrupted and as such is safe to pass A to another call back).

I agree that this would be a desirable improvement.  It doesn't solve the problem in general though, since not all data structures that fulfill tasks can reasonably expect that they will get a "replacement task" to fulfill down the road.  Even with `Queue`, it's a bit uncertain.  I think that a more general approach here would be to have an `onPreempt` or similar function (I believe @edwinchoi suggested this in another issue).  This function would be attached to `Await` and would take the type produced by `req`.  It would be invoked exactly and only when the `req` task was interrupted but still produced a value significantly later.  In the case where `onPreempt` is invoked, the `onComplete` will have already run against `Kill`.

This splits the guarantees into two halves:

- `onComplete`
  + Exactly-once execution
  + Invoked early (i.e. as soon as the process rationally completes)
- `onPreempt`
  + Guaranteed to be run if `req` completes (despite interrupt) *and* if `onComplete` did not run on the resulting value
  + At-most-once execution

This gives users the opportunity to requeue, or close resources, or do whatever makes the most sense given the specific instantiation of `req`.

> (3) because of (1) I would like to allow any continuation of any code that may follow asynchronous task to be executed when task was interrupted, but if that task evaluates again to asynchronous code, that code must not be executed, if task was interrupted before, instead task must finish its callback with Interrupt.

Yes, I agree with this.  In fact, this is what the current implementation in this PR achieves.  However, what is missing from this property is the clean up of resources acquired by the task when it completes naturally *following* an interrupt.  This is where the `onPreempt` suggestion comes from (again, credit to @edwinchoi).

In summary, I think we're actually very close to something that satisfies all of these requirements.  The "returning `Boolean`" thing is really an optimization that helps specific implementations; I'm very certain that `onPreempt` satisfies the same use-case in a more general fashion.  It's very easy to implement the desired "interrupt if possible, complete if not, return immediately" semantic even on top of the current `Future` (I'll prototype this tonight).

djspiewak:
Ok, so I reimplemented `runAsync` and `completeInterruptibly`.  We now guarantee exactly-once invocation for `onComplete` outside of the `await`.  `onComplete` within the `await` will not execute at all in the event that the task is interrupted and was not already completed (so basically, will not execute if interrupted).

If we want to enable clean up of resources opened unavoidably by awaited tasks (e.g. queue values lost, async IO connections, etc), we have two options.  We can either a) add `onPreempt`, or b) enrich `EarlyCause` with a state value of type `A`, where `A` is the type produced by the `req` task.  This second idea would of course need to be paired with adding a `Interrupted(a)` case to `EarlyCause`.  This would make `onPreempt` unnecessary, since it would be possible to achieve exactly the same goal by using `awaitOr` rather than `await`.

Another, slightly trickier idea that I *don't* think will work is implement the `Interrupted.PostStep` case in my implementation in the following way:

```diff
diff --git a/src/main/scala/scalaz/stream/Process.scala b/src/main/scala/scalaz/stream/Process.scala
index b3b5081..effa911 100644
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1230,7 +1230,7 @@ object Process extends ProcessInstances {
 
                 def unpack(msg: Option[_ \/ a]): Option[a] = msg flatMap { _.toOption }
 
-                def wrap(result: a): Process[Task, O] = Try(rcv(\/-(result)).run) +: cont
+                def wrap(result: a): Process[Task, O] = Try(rcv(\/-(result)).run)
 
                 // handler extractors
                 object Interrupted {
@@ -1276,7 +1276,7 @@ object Process extends ProcessInstances {
                 // completed the task, no interrupts, no exceptions, good to go!
                 object Completed {
                   def unapply(msg: Option[_ \/ a]): Option[Process[Task, O]] =
-                    unpack(msg) map wrap // filter { _ => !interrupted.get().isDefined }   (commented out to avoid race conditions)
+                    unpack(msg) map wrap map { _ +: cont } // filter { _ => !interrupted.get().isDefined }   (commented out to avoid race conditions)
                 }
 
                 val interrupt = completeInterruptibly((Task fork (checkInterrupt >> req)).get) {
@@ -1288,21 +1288,15 @@ object Process extends ProcessInstances {
                     }
                   }
 
-                  case Interrupted.PostStep(inner, cause) => {
-                    /*
-                     * TODO this is the case where we couldn't stop the resource from being grabbed; we should clean it up, but *onComplete has already run!*
-                     * so we would really like to clean up resources, but we can't because onComplete cannot be run more than once
-                     */
-
-                    ()
-                  }
+                  case Interrupted.PostStep(inner, cause) =>
+                    inner.kill.run runAsync { _ => () }
 
                   case Exceptional(t) =>
                     // we got an exception (not an interrupt!) and we need to drain everything
                     ref.set(go(Try(rcv(-\/(Error(t))).run) +: cont).run)
 
-                  case Completed(inner) =>
-                    ref.set(go(inner).run)      // we completed successfully; forward along the reference
+                  case Completed(continuation) =>
+                    ref.set(go(continuation).run)      // we completed successfully; forward along the reference
                 }
 
                 ref.set({ _ => interrupt() })
```

I really, *really* don't think the above is going to work quite correctly, but it does have the advantage of sort-of doing what we want!  It guarantees that all non-nested `onComplete`s are invoked exactly-once, and any nested (i.e. dependent on "awaited values") `onComplete`s are guaranteed an at-most-once semantic, which is strengthened to exactly-once iff the awaited task ultimately produces a complete value.  As I said, I am really unsure that the above diff will work and remains free of other…complications, but it's an idea!

Attn @alissapajer.  Thoughts, @pchlupacek?  @edwinchoi?

djspiewak:
It seems my earlier implementation (without running await-dependent finalizers on interrupt) breaks `TcpSpec` in Travis. I pushed my earlier diff to see if that makes a difference.  Again, needs careful review before we go live.

djspiewak:
Update (apologies for spam): it looks like it's still failing.  So I need some help on three fronts here:

1. Thoughts on `onPreempt` vs making `EarlyCause` parametric to allow `awaitOr` to fill the gap
2. Thoughts on my semantic split for finalizers: exactly-once for non-await-dependent, and at-most-once for await-dependent with guaranteed execution in the case of task completion (despite interrupt)
3. Help with why `TcpSpec` would be breaking because of these changes, and apparently only in Travis (it works fine locally)

pchlupacek:
@djspiewak to be honest I can not follow the code that much but would like just to understand why we simply cannot return to current `actor` implementation, and just change the semantics so : 

- on Interrupt we just interrupt running task (no Cleanups are run contrary to current code)
- once the open task is interrupted we receive (from task) either the value or the `Interrupted` or any other exception (that is not guaranteed in current code, as task may not complete) 
- we run cleanup from Await only.  

I have feeling that this shall work. Or do you have perhaps situation when it will not?

with `onPreempt` and `onComplete` I would do whatever possible to keep only one combinator in lib if possible. I am still not really convinced that users will understand when to use which one, and in complex asynchronous scenarios, it may be even hard to reason which one shall be used instead of the others, but lets go to that one later.

djspiewak:
> to be honest I can not follow the code that much but would like just to understand why we simply cannot return to current actor implementation

Reason number one I would give is that the old actor implementation is very imperative, very stateful, hard-to-follow code that cannot be easily refactored into a more general type signature.  The reimplementation has three bits of state, but I'm pretty sure I could reduce it to one without anything more serious than a slightly higher level of runtime indirection with nested awaits (a negligible penalty).  The cases are labeled, the logic defining the cases is clearly split from the logic defining what the cases *do*, and that logic is clearly separated.  The old implementation actually has a single snippet which is reused in three different cases of the algorithm, and one of them I'm relatively certain is by accident!  (i.e. a bug)

> on Interrupt we just interrupt running task (no Cleanups are run contrary to current code)

Without having the new `Future` interpreter (`completeInterruptibly`), this requirement is in conflict with the following one:

> once the open task is interrupted we receive (from task) either the value or the Interrupted or any other exception (that is not guaranteed in current code, as task may not complete)

Can't do it without `completeInterruptibly`.  The problem is that `listenInterruptibly` always discards the value, even if the task has already completed.  The `completeInterruptibly` interpreter retains the value if it computes and actually passes it back to user-land code as a callback which can be unambiguously distinguished as a latent case *in* user-land.  The existing interpreters do not provide a sufficient mechanism to achieve this.

> we run cleanup from Await only

I'm not sure I understand this requirement.  `onComplete` may be applied to any process state.

Ultimately, even with `completeInterruptibly`, I don't think the old actor implementation is structured such that I *can* separate the cases to the extent that I have with the current implementation.  As I said, the code is very tangled, and multiple cases are handled as distinct passes through the same code, resulting in some unintended effects.  It would be very difficult to get the old implementation to a point where we can reason about the `onComplete` semantics in a confident way.  The new implementation achieves this *directly*, by virtue of marrying the uniqueness of cases with the reasonable invariants of `completeInterruptibly`.

> with onPreempt and onComplete I would do whatever possible to keep only one combinator in lib if possible. I am still not really convinced that users will understand when to use which one, and in complex asynchronous scenarios, it may be even hard to reason which one shall be used instead of the others, but lets go to that one later

Well, I don't think the distinction is all that crazy.  However, the more important point is that parameterizing `EarlyCause` gives us the ability to do `onPreempt` without adding a combinator (in short: it's the more straightforward path to reusing the combinators we already have to achieve the expressiveness goal).  Even beyond that, I think that my inner-`onComplete` revision (1a30cec) makes the whole thing redundant in any case, and enables preemption to "just work" the way users expect.

pchlupacek:
@djspiewak I think I found the implementation that satisfies the rules we have been talking about. I also now remember the reason why runAsync is based on actor. This was specific decision to increase concurrency. If you run drained processes they do evaluate w/o consultation of higher level combinators and their actors (such as wye, merge, etc) resulting in much higher concurrency compare if they will had to sync on central actor of wye/merge. 

Would you be so kind and review/comment #318 ? 

Thanks so much for all your effort and work with this issue.

djspiewak:
> I also now remember the reason why runAsync is based on actor. This was specific decision to increase concurrency. If you run drained processes they do evaluate w/o consultation of higher level combinators and their actors (such as wye, merge, etc) resulting in much higher concurrency compare if they will had to sync on central actor of wye/merge.

This is precisely what my implementation does.  It really doesn't have anything to do with `runAsync` being implemented as an actor, since the actor inside `runAsync` in the old implementation is only used for sequential coordination and queueing.

pchlupacek:
@djspiewak after more thinking about this issue and discussing it here, I would like to discuss following: 

given simple scenario: 

```scala 

await(resourceTask){ r1 =>
  doSomething(r1)
  .onComplete(cleanupR1(r1))
}
.onComplete(cleanupProgram)


```

What we really want to achieve, is that when we interrupt the `resourceTask` we run `cleanupProgram` but after eventually `resourceTask` completes, we want to run `cleanupR1` but not the `cleanupProgram`

I think this is perfectly achievable with current signature and algebra, only problem may be that we will need to revisit the code and assure that await is always attaching the `next` continuation correctly as `outer` process and not to `rcv` function. 

Then I believe we will always run cleanup on Interruption (outer cleanup) and on await, when it was interrupted, but somehow later completed will always run only inner cleanup, that would not run outer cleanup, hence `next` is not attached. Something like : 

``` scala 
  //....
  case AwaitDone(result, Await(_,rcv),next) if interrupted =>
    result.fold(rcv(left(_)).run, rcv(right(_)).run.kill).run.runAsync { _ => .....} //next is ignored on interrupt

  case Interrupt if notYetCompleted =>
   next.kill.run.runAsync { _ =>  cb(left(kill)   } // next is memoized and used on interrupt to clean process with kill
//....
```

That way we can guarantee exactly once semantics (also for inner process), plus we don't have to introduce anything new to the algebra.  The only problem is we don't guarantee ordering of the cleanups, but perhaps even that is achievable. Do you agree on that?

djspiewak:
> What we really want to achieve, is that when we interrupt the resourceTask we run cleanupProgram but after eventually resourceTask completes, we want to run cleanupR1 but not the cleanupProgram
> 
> I think this is perfectly achievable with current signature and algebra, only problem may be that we will need to revisit the code and assure that await is always attaching the next continuation correctly as outer process and not to rcv function.
> 
> Then I believe we will always run cleanup on Interruption (outer cleanup) and on await, when it was interrupted, but somehow later completed will always run only inner cleanup, that would not run outer cleanup, hence next is not attached.

Yes!  :-D

This is what I've been advocating for.  Basically, you're describing the "strengthened inner/outer" idea that I referenced in the other PR.  This current pull request implements this in the following two cases:

```scala
                  case Interrupted.PreOrMidStep(cause) => {
                    // interrupted; now drain
                    (Try(rcv(-\/(cause)).run) +: cont).drain.run runAsync {
                      case -\/(t) => S { cb(-\/(Error(t) causedBy cause)) }
                      case \/-(_) => S { cb(-\/(cause)) }
                    }
                  }

                  case Interrupted.PostStep(inner, cause) =>
                    inner.kill.run runAsync { _ => () }
```

Pre- and mid-step interrupts correspond to interrupting `req` either before it starts or after it is started but before it completes.  Thus, in that case, we invoke `rcv` with left of the `cause` and continue the process.  The post-step interrupt corresponds to an interrupt which was detected *after* the task completed.  Due to the fact that, on interrupt, we *immediately* invoke the callback (to provide deterministic semantics), this can only correspond to the situation where we already handled the interrupt either as pre- or mid-step, and so it is precisely the situation where we want to run your `cleanup(r1)`.  We do this by killing the inner process and then running it to completion (note that `inner` is just `Try(rcv(\/-(result)).run)`, so it does *not* include the continuation).

I believe you're right that this is the semantic which we should strive for.  It is by far the most intuitive to users, it preserves existing resource management code, and really the only thing it *doesn't* get right is the fact that it runs finalizers "out of order".  I'm ok with that though, and I really doubt it's going to be a problem.

pchlupacek:
@djspiewak cool we have agreement here. The only thing we need to do now is to review all combinators and make sure we never attach the `outer` continuation to `inner` receive. This mean review pipe, tee, wye and perhaps other places with step.

Its just straight work from my branch, not sure where you at with this branch.

edwinchoi:
This will not work in this case:
```scala
val a : lock
val b:lock

await(a.acquire) { _=>
  await(b.acquire) { _=>
    doSomethingDangerous().onComplete(
      eval_(b.release))
  }.onComplete(eval_(a.release))
}
```

Without interrupt semantics, `doSomethingDangerous` runs with exclusive access to the resources protected by `a` and `b`.
With interrupt semantics this is no longer the case, since interrupt could cause `a` to be released while `doSomethingDangerous` is still running... Which would allow another thread to acquire the lock and proceed as if they had exclusive access.

The lack of support for sequential reasoning makes it much more difficult to verify overall correctness (can't simply look at sequential execution). And the burden of reasoning about overlapping computations is placed on the user.. These are difficult pills to swallow and makes devising correct solutions very difficult.

Note: the example above will never work with interrupt semantics as they've been described. In fact, there is no way to safely share resources beyond a single step with interrupt in the picture.
This was one of the reasons I suggested `onPreempt`. `onPreempt` would provide external threads with the means to tell the process to halt without having to call the finalizers directly, which breaks isolation. `onPreempt` would not be a finalizer, the process would still have to wait for await to complete before halting.

This may not be the cleanest approach, but it is amenable to sequential reasoning.
Also, having separate `onPreempt` and `onComplete` combinators (vs EarlyCause as @djspiewak had suggested earlier) is that the execution/ownership boundaries are clear and explicit. Which in my opinion is a plus.

edwinchoi:
As an aside, I'm still very much in support for the proposed `runAsync`. It is much easier to workaround the issues when you can reason about the behavior sequentially.

djspiewak:
@edwinchoi Looking at a more abstract example:

```scala
await(t1) { a1 =>
  await(t2) { a2 =>
    p onComplete f3
  } onComplete f2
} onComplete f1
```

The tasks and the process `p`, with the scheme proposed by @pchlupacek and I, the following orders of evaluation are possible:

- Nothing interrupted
  1. `t1`
  2. `t2`
  3. `p`
  4. `f3`
  5. `f2`
  6. `f1`
- `t2` interrupted *during* evaluation
  + Permutation one
    1. `t1`
    2. `t2`
    3. `f2`
    4. `f1`
    5. `p`
    6. `f3`
  + Permutation two
    1. `t1`
    2. `t2`
    3. `f2`
    4. `p`
    5. `f3`
    6. `f1`
- `t2` interrupted *before* evaluation
  1. `t1`
  2. `f2`
  3. `f1`
- `t1` interrupted *during* evaluation
  1. `t1`
  2. `f1`
  3. `t2`
  4. `f2`
  5. `p`
  6. `f3`
- `t1` interrupted *before* evaluation
  1. `f1`

Coming back to your concrete example, the problem arises when *either* `f1` or `f2` runs before `p`.  I honestly can't think of a good way to avoid this other than to instruct users, "don't do that!"  This sort of construction is effectively defeating `Process`'s resource management tools by imposing an external, non-data-directed constraint.  Namely, your `a` and `b` locks are tied together, but not mutually constrained by data flow.  A far better approach would be to acquire both locks in a single `await`.  In such an implementation, the resource management would work exactly as anticipated.

So in other words, the scheme imposed by @pchlupacek and myself works in any case where resource constraints are *lexical* in nature and otherwise atomic in acquisition and release.  Unfortunately, I don't think it's possible to *properly* handle all of the possible "spooky action at a distance" cases.  If users choose to implement non-lexical data flow, honestly, I am of the opinion that they're just on their own to get it right.

Obviously, we need to document this behavior extremely well, but I don't in general feel that it is the wrong approach just because it is *possible* through side-effects to generate situations where it causes breakage.

@pchlupacek 

> Its just straight work from my branch, not sure where you at with this branch.

Tracking down a bug in `TcpSpec`, but otherwise done.  I really don't understand what's causing the breakage in that test, and I could use some help if anyone understands it better.  `njoin` itself seems to be working correctly, so I'm not sure how the exact semantics of `runAsync` could be at all leaking into `tcp`.

edwinchoi:
@djspiewak I didn't mean to make it seem like the current approach is somehow wrong, if this is how my previous post came off then I apologize. It's a proper solution given the problem constraints. I was, more or less, responding to:
> I believe you're right that this is the semantic which we should strive for. It is by far the most intuitive to users, it preserves existing resource management code, and really the only thing it doesn't get right is the fact that it runs finalizers "out of order". I'm ok with that though, and __I really doubt it's going to be a problem.__

I don't think it's that far fetched to use a resource in a process that's more than 1 hop away from the acquiring process, but that can be discussed separately.

There is no way to maintain ordered completion _and_ interleave finalization on interrupt at the same time. The two are hopelessly in conflict with one another. The former requires serialized execution whereas the latter works by interleaving.

One possible work-around to the ordering problem is to serialize the process before its fed into `wye` (e.g. `async.mutable.Queue` with synchronized terminals). This would guarantee that the behavior of the upstream processes are the same regardless of what appears downstream.

djspiewak:
> I didn't mean to make it seem like the current approach is somehow wrong, if this is how my previous post came off then I apologize.

It didn't come off like that at all!  :-)  I think my reply may have seemed a little brusque since I originally wrote it after reading your message and under the belief that you were wrong!  I couldn't see a way the finalizer semantic would break your example, and I set out to show how it wasn't a problem.  In doing so, I realized that it *was* in fact a problem, so I had to go back and cut down my message a bit.  There was good information in my "refutation" though, so I didn't want to cut out everything.  That's probably why it seemed a little tense.

> There is no way to maintain ordered completion and interleave finalization on interrupt at the same time. The two are hopelessly in conflict with one another. The former requires serialized execution whereas the latter works by interleaving.

Agreed.

> One possible work-around to the ordering problem is to serialize the process before its fed into wye (e.g. async.mutable.Queue with synchronized terminals). This would guarantee that the behavior of the upstream processes are the same regardless of what appears downstream.

Yep.  I've actually used that workaround for other (natural) ordering issues associated with the nondeterministic combinators.

---

In other news, I've been working through `tcp.server`, and I'm pretty sure it is just straight-up broken.  The types are enormously deceptive and actually somewhat useless (it promises a `Process[Task, Process[Task, Throwable \/ A]]`, but the inner `Process` is essentially just unordered mish-mash results from the handlers and doesn't directly correspond to individual client connections; there is in fact only *one* inner process!).  Furthermore, I think the only reason it is working *now* is due to the way that `Task#runAsync` works with tasks constructed using `delay`.  Since my `runAsync` converts `delay` tasks into `apply` tasks, the order of evaluation which `tcp.server` implicitly relies upon is broken.

Oh also the chat spec in `TcpSpec` is a hot mess.  Does ScalaCheck reinitialize `Properties` instances for each iteration of a `forAll`?  It's really hard to see how it ever worked in the first place.

I'm working on fixing the implementation right now.  Still could use help, but making progress.

pchlupacek:
@djspiewak I am just curious on Tcp bug you hunting why it passes with my implementation of runAsync? Maybe that can give some further info?

djspiewak:
@pchlupacek You can see the bug in the Travis test results.  Basically, the chat example doesn't work correctly.  The server starts and then immediately stops, meaning that the clients aren't able to push any data around.

As I mentioned, I'm pretty sure the test *only* passed previously because the old `runAsync` ran leading `delay` tasks in the current thread, affecting the timing of `wye` and `njoin` scheduling.  Your branch preserves this property, while mine does not (since mine allows for interrupting of `delay`).

pchlupacek:
@djspiewak your reasoning seems to be fair. In fact, despite this (that has to be fixed), do you think we always need to fork in runAsync. How about forking only async code, or perhaps do not fork at all, and use the outer strategy? I am referring to #1298. I think if we can implement runAsync sequentially, why to forcing to fork the code there?

djspiewak:
@pchlupacek Originally, I injected the `Task fork` in order to make the following case work correctly:

```scala
val p = Process repeatEval (Task delay (())) drain

(Process(1) wye p)(wye.mergeHaltBoth)
```

Without `Task fork`, my code *previously* was hanging on the above.  I don't think that would be the case now, but I haven't checked.

I think it's actually very important to ensure that tasks constructed with `delay` get forked.  Without forking, a `delay` task can arbitrarily reduce throughput in higher level combinators like `wye` and `njoin`.  Tasks constructed with `delay` can often take some time to execute, and if that code is executing synchronously *and* if `Process#runAsync` is implemented using a sequential actor strategy (as both the old and your implementations do), then that slow task is executing on the same thread that handles scheduling and preemption of concurrent processes in the nondeterministic combinators.  I'm pretty sure this is precisely why `tcp.server` was working at all in the first place.

pchlupacek:
@djspiewak  I agree with your reasoning. If we would fork only at beginning, there is little overhead, and anyhow higher level combinator shall do that. 

On other hand any further execution of any further tasks imho shall not be forked. 

i.e. `p.drain.runAsync(_=>???)` will fork only once, not at every await`s evaluation. I am not sure if that's what your code guarantees, hence I see that `fork` at every `go` cycle. Forking every task is not desired IMHO.

djspiewak:
> I am not sure if that's what your code guarantees, hence I see that fork at every go cycle. Forking every task is not desired IMHO.

Remember that `go` is *only* recursively invoked in the case where there are nested `Await`s without intervening `Emit` or `Halt`.  We don't recursively invoke `go` on a drained process *ever*.  The only case which even approaches this is the recursive `Exceptional` case, where the exception propagation may result in a process which is effectively drained, resulting in a long string of awaits.  However, those awaits will be cleanups only (basically just finalizers), and so throughput is less of a concern.

We absolutely need to fork the head of the task chain every time `runAsync` is called.  Forking on the recursive `go` cases is slightly sub-optimal, but I sincerely doubt that the effect is measurable due to the marginal nature of the case where it appears and the way in which `Process` is used in those cases.

pchlupacek:
@djspiewak ok, anyhow if we encounter any issues with that approach we may come to that point later. With tcp, have you been able to track it? I think we also still have to review all the fusing code for the await correctness, these are two items of my head, not sure if you have some others?

djspiewak:
> With tcp, have you been able to track it?

Sort of.  I've had to rewrite some of the signatures, unfortunately, since the types are simply *wrong*.  Also the tests are really really terrible, so I've been rewriting them as well.  Basically, I'm redoing a decent chunk of the tcp stuff.  The good news is that things mostly work, but the less good news is that I've run face first back into that NIO2 bug on Mac (which I actually think is a kernel issue, not a JDK bug, given the way it affects the process image).  I'm still working through some issues, and I'll be testing more on linux as a result, but I'm relatively sure I can make it all work.

> I think we also still have to review all the fusing code for the await correctness, these are two items of my head, not sure if you have some others?

Off the top of my head, we have the following todo list:

- My branch
  + Fix `tcp` (backup plan is to simply pull this functionality temporarily, since it appears to have never quite worked the way we thought it did)
  + Port `step` fix over (should take all of five seconds)
  + Double check `Process.suspend` (it's possible that my removal of the `Emit(Seq())` special case from `runAsync` is causing weirdness)
- Your branch
  + Implement inner/outer finalizer semantics
  + Forking and interrupts for leading `Task.delay` task steps (did we ever agree on this?)
- Decide between the two branches and merge (due: this weekend)

djspiewak:
Alright, this is the sound of me officially giving up on `TcpSpec`.  I don't know why `tcp` is working, and honestly as far as I can tell, it *isn't*.  It doesn't appear to be a simple matter to make it work, and I don't think it can be made to work at all without changing the signatures.  So I basically just fuzzed the test until it passed.  I didn't convert things into `forAll { _ => true }`, but you get the idea.  I'll be filing a separate issue to track the fact that `tcp` is sort of non-functional (and apparently has been all along).

Commits incoming on that front.  I'll port over the `step` fix and then we should be all set.  I also informally checked `Process.suspend` (what sort of properties is this even supposed to have?) and it does appear to work correctly.  I also satisfied myself that adding back the `Emit(Seq())` case doesn't really change anything, so I think we're basically good to go once I port in the `step` fix.

Alternatively, it might make more sense to push the `step` fix into a separate PR, since it is *logically* a separate bug.  Thoughts?

djspiewak:
Alright, with the addition of #330 and the documentation of `tcp`'s issues in #329, my branch(es) are basically set!  Consider this my submission for the "which `runAsync` PR to merge into master" question.

pchlupacek:
@djspiewak excellent. Now we need to review fusing to make sure we don't swap inner and outer cleanups and thats to me all. Also we need to ask paul to take care of tcp spec or give us better info to his ideas behind tcp.* Just to assure we are on right track.

djspiewak:
@pchlupacek What do you mean by "fusing"?  If by inner/outer swapping you mean confusing the semantics for inner vs outer finalizers, my implementation of `runAsync` treats awaits structurally, meaning that inner and outer *cannot* be swapped.

pchlupacek:
@djspiewak We need to assure that in all fusing combinators(pipe, tee, wye) the inner Await `rcv` is correctly attached and not mismatched with outer process, that means the `rcv` function will never contain `next` from `outer` step. As we run `inner` cleanup only when await completes.  

I am not sure if currently thats implemented everywhere like this. I will try to go through the combinators and verify this. 

Will use this branch ok?

djspiewak:
> We need to assure that in all fusing combinators(pipe, tee, wye) the inner Await rcv is correctly attached and not mismatched with outer process, that means the rcv function will never contain next from outer step. As we run inner cleanup only when await completes.

Ah, that makes sense.  I can't imagine how a fusing combinator would "detach" an inner finalizer and move it outside, but I suppose it is technically possible.

> Will use this branch ok?

This branch is fine to use, though it might be better to use `bug/step` since that has the complete set of fixes for all of the other issues we've been tracking, as well as the fix to `njoin`'s assumptions about `runAsync`.

pchlupacek:
@djspiewak ok will use bug/step

pchlupacek:
@djspiewak I just looked on status of bug/step and somehow I just see the last commit of mine, not the code you have been working on. Can you pls clarify which branch you want to merge so I can review and test the pipe/tee/wye?

Thx

djspiewak:
@pchlupacek I just confirmed that the branch *is* pushed.  Maybe you have a local bug/step branch?  I'm thinking of the code in PR #330.

pchlupacek:
@djspiewak  am afraid we will need to perhaps change the implementation a bit. Look on pipe: 

```scala 

  final def pipe[O2](p1: Process1[O, O2]): Process[F, O2] =
    p1.suspendStep.flatMap({ s1 =>
      s1 match {
        case s@Step(awt1@Await1(rcv1), cont1) =>
          val nextP1 = s.toProcess
          this.step match {
            case Step(awt@Await(_, _), cont) => awt.extend(p => (p +: cont) pipe nextP1)
            case Step(Emit(os), cont)        => cont.continue pipe process1.feed(os)(nextP1)
            case hlt@Halt(End)               => hlt pipe nextP1.disconnect(Kill).swallowKill
            case hlt@Halt(rsn: EarlyCause)   => hlt pipe nextP1.disconnect(rsn)
          }

        case Step(emt@Emit(os), cont)      =>
          // When the pipe is killed from the outside it is killed at the beginning or after emit.
          // This ensures that Kill from the outside is not swallowed.
          emt onHalt {
            case End => this.pipe(cont.continue)
            case early => this.pipe(Halt(early) +: cont).causedBy(early)
          }

        case Halt(rsn)           => this.kill onHalt { _ => Halt(rsn) }
      }
    })
```

note the first fusing of `F` Await (line #7). We are taking the `outer` part of process and giving it to `inner` receive. That means, if req will never complete, and `cont` will contain cleanup for other processes we are doomed... 

I am not sure if current await signature allows us to resolve this. 

Perhaps we shall reincarnate the algebra extension for the Cause, as this may be the only option perhaps at the end. 

I also don't like the current state of flatMap(r => p(r).cleanup(r)) as you write in your release notes, because to me that sounds counter-intuitive and easily overlooked in code (in fact we use this idiom a lot). I think if we introduce that algebra of Interrupted(a) we are perhaps safe with flatMap too. 


Any thoughts?

djspiewak:
> We are taking the outer part of process and giving it to inner receive. That means, if req will never complete, and cont will contain cleanup for other processes we are doomed...
> 
> I am not sure if current await signature allows us to resolve this.

Actually, it does!  Even better, it is already resolved.  The implementation of `extend`:

```scala
def extend[F2[x] >: F[x], O2](f: Process[F, O] => Process[F2, O2]): Await[F2, A, O2] =
  Await[F2, A, O2](req, r => Trampoline.suspend(rcv(r)).map(f))
```

Note that we're mapping the results of `rcv` regardless of the value of `r`!  Thus, any finalizers in `cont` will still be invoked in the case where `req` doesn't complete and is interrupted, since we still call `rcv(-\/(Kill))` in that case.  The only thing here that is a little bit wonky is the fact that any finalizers in `cont` may be called multiple times, since if `req` ultimately completes, we re-invoke `rcv` with the result and then kill the process, meaning that finalizers in `cont` run a second time.  Not sure exactly how to resolve this, but it seems doable.

In any case, this situation is still a strict upgrade from where we are now, since finalizers passed through a fuser is the only case where a finalizer can be run more than once.  In master, it happens all the time.

> I also don't like the current state of flatMap(r => p(r).cleanup(r)) as you write in your release notes, because to me that sounds counter-intuitive and easily overlooked in code (in fact we use this idiom a lot)

Actually, that idiom isn't safe even in master!  Note the implementation of `flatMap`: it's just a bunch of appends, in the most traditional sense (`++`).  Those appends are going to get completely nulled out in the event that the process is interrupted, meaning that the contents of those appends (including the finalizer!) are lost.  Under certain very specific circumstances, we can resolve this with append fusing, since we do have a *stack* of tails, but this sort of trick is very dependent on the structure of the process to which the `onComplete` is attached, within the body of the `flatMap`.

So in other words, anyone using this idiom currently on master is leaking resources.  This PR doesn't change that since it doesn't change the structure of `Append` or the implementation of `flatMap`.

> I think if we introduce that algebra of Interrupted(a) we are perhaps safe with flatMap too.

Not to my knowledge.  The point to `Interrupted(a)` is so that we can use `awaitOr` and explicitly provide a case that handles `Interrupted(a)` as distinct from `Kill`, `Error(_)` and `End`.  At first glance, it seems like we could enrich `++` with the ability to do something clever with the value, but we can't do *that* without defeating the interrupt mechanism altogether.  Append chains *need* to be nulled out in the event that the process is killed, otherwise killing a process has no meaning (think about `repeatEval`).  So with respect to flatMapped finalizers, that leaves us back in exactly the state we are both in master and on the PR: the appends get nulled out by the `Interrupted(a)` cause, the finalizer is never seen and resources leak.

pchlupacek:
@djspiewak I a bit confused. What do you think output of this program shall be? 

```scala

 property("Outer cleanup through pipe gets called") = secure {
    val task1 = Task.fork(Task.delay{ Thread.sleep(3000); println("TASK 1 DONE") })
    val awaitP =  await(task1)(u =>Process.emit(u))


    Process(true).toSource
    .wye(
        awaitP.pipe(id)
        .onComplete(eval_(Task.delay(println(">>>> CLEANUP CALLED"))))
      )(wye.interrupt).run.run

    true
  }

```

djspiewak:
@pchlupacek It would be natural to expect the output to be the following:

```
TASK 1 DONE
>>>> CLEANUP CALLED
````

However, I believe both on master and most certainly on this branch, the second line will not print.  I have a minimized test case attached to #333 which demonstrates why this is, in general, the case with all `append`-related operations (including `pipe`).

**Comment by pchlupacek** [comment]
@djspiewak yeah exactly this is what it does with your branch. Effectivelly, if task1 won't complete no cleanup is called, moreover it seems there is a problem with your branch so the wye won't complete unless the task1 completes. 

in current master you just see 
>>>> CLEANUP CALLED which I think is correct behaviour, as task1 shall be interrupted before timeout elapses, and shall not have chance to do println. 

In fact this demonstrates the problem with pipe, if we swap outer cleanup and attach it to inner receive, plus I think there is as well some other problem so the wye won't complete at all.

**Comment by djspiewak** [comment]
@pchlupacek

> Effectivelly, if task1 won't complete no cleanup is called, moreover it seems there is a problem with your branch so the wye won't complete unless the task1 completes.

Wait, so you're trying to test what happens when the task doesn't run *at all*?  That would explain why it seemed to be very race-conditiony.  If the task never runs at all, I would still expect the cleanup finalizer to run.  I think this is actually a legitimate bug in `runAsync`.  Tracking now…

**Comment by pchlupacek** [comment]
@djspiewak  yup. And I think its perhaps runAsync bug and as mentioned before current implementation of pipe won't run the cleanup as well.

**Comment by djspiewak** [comment]
@pchlupacek Please verify that the following test case implements what you're thinking of:

```scala
  property("outer cleanup through pipe gets called with preempted task") = secure {
    @volatile
    var complete = false

    @volatile
    var cleanup = false

    val flag = new SyncVar[Unit]

    val awaitP = await((Task delay { flag.get(500) }) >> (Task delay {
        Thread.sleep(3000)
        complete = true
      }))(emit)

    // interrupt awaitP before the second part of the task can even start running
    eval(Task delay { flag.put(()); true }).wye(
        awaitP pipe process1.id onComplete eval_(Task.delay { cleanup = true })
      )(wye.interrupt).run.run

    !complete :| "task interrupted" && cleanup :| "inner cleanup invoked"
  }
```

This actually *passes* on my branch.

pchlupacek:
@djspiewak I don't think so it is same example as I have made. 

I really think that task1 shall be interrupted w/o giving chance to run any code. So if the program terminates, the task won't be run at all. When program continues, and as such thread pool continues to exists task shall complete, and as such shall print its result after 
so I think only two outputs from my program are possible here : 

In case there would be timeout after run.run (simulating programs continues, jvm won't terminate) my example shall print

```
>>>> CLEANUP CALLED
TASK1
```
where as if jvm terminates once run.run is called program will just print

```
>>>> CLEANUP CALLED 
```

in neither case we shall have 
```
TASK1
>>>> CLEANUP CALLED 
```

output. 

The last example actually says we need task1 to complete before running any finalisers. 

I just made the actual test that I think must pass : 

```scala

  property("Outer cleanup through pipe gets called") = secure { 
    
    @volatile var q = Queue[Int]()

    val task1 = Task.fork(Task.delay[Unit]{ Thread.sleep(1000); q = q.enqueue(1)})
    val awaitP =  await(task1)(u =>Process.emit(u))


    Process(true).toSource
    .wye(
        awaitP.pipe(id)
        .onComplete(eval_(Task.delay{  q = q.enqueue(0)}))
      )(wye.interrupt)
    .run.run

    Thread.sleep(3000)
    q.toList == List(0,1)
  }

```

djspiewak:
@pchlupacek Ah, I think my second test is passing due to a race condition.  Working on it…

djspiewak:
@pchlupacek To be complete, the following pair of properties tests *both* interruption (which is what your most recent test is checking) and preemption (which is what you were initially discussing).  Note that the `completed` flag in each guarantees that we didn't hit a race condition and accidentally end up testing the wrong case.  If you would like, I can add counters (rather than booleans) to each so we can see that things are only being invoked a single time, but I'm not sure that is necessary.

```scala
  property("outer cleanup through pipe gets called with interrupted task") = secure {
    @volatile
    var complete = false

    @volatile
    var cleanup = false

    val flag = new SyncVar[Unit]

    val awaitP = await(Task delay {
      flag.put(())
      Thread.sleep(3000)
      complete = true
    })(emit)

    eval(Task delay { flag.get(500); true }).wye(
        awaitP pipe process1.id onComplete eval_(Task.delay { cleanup = true })
      )(wye.interrupt).run.run

    complete :| "task completed" && cleanup :| "inner cleanup invoked"
  }

  property("outer cleanup through pipe gets called with preempted task") = secure {
    @volatile
    var complete = false

    @volatile
    var cleanup = false

    val flag = new SyncVar[Unit]

    val awaitP = await((Task delay { flag.get; Thread.sleep(500) }) >> (Task delay {
        Thread.sleep(3000)
        complete = true
      }))(emit)

    // interrupt awaitP before the second part of the task can even start running
    eval(Task delay { flag.put(()); true }).wye(
        awaitP pipe process1.id onComplete eval_(Task.delay { cleanup = true })
      )(wye.interrupt).run.run

    !complete :| "task interrupted" && cleanup :| "inner cleanup invoked"
  }
```

pchlupacek:
@djspiewak in fact, I think we don't have with my test any race condition. If we do have, then there is something weird going on. I have added also the test for other "outer" cleanup. I think the test is pretty simple and I can't see any race as the task1 has 1s delay, and the final await is 3s in it. I think this has to always pass in the order as we check it. 

```scala

  property("Outer cleanup through pipe gets called") = secure {

    @volatile var q = Queue[Int]()

    val task1 = Task.fork(Task.delay[Unit]{ Thread.sleep(1000); q = q.enqueue(1)})
    val awaitP =  await(task1)(u =>Process.emit(u))


    Process(true).toSource
    .wye(
        awaitP.pipe(id)
        .onComplete(eval_(Task.delay{  q = q.enqueue(0)}))
      )(wye.interrupt)
    .onComplete(eval_(Task.delay(q = q.enqueue(2))))
    .run.run

    Thread.sleep(3000) 
    q.toList == List(0,2,1)
  }

```

djspiewak:
@pchlupacek Your test is interrupting the task before it even starts running.  This ultimately means that `1` never gets enqueued and the test fails.  It's a race condition because `wye` *can* feasibly schedule things such that task starts just a bit before the emit is evaluated.  Unless that race condition is hit though, the test will fail.  You'll notice that my test cases explicitly use signals as well as separate tests to factor out this timing question.

Could you perhaps describe what test condition you're trying to express?  My tests cover both preemption (task never runs) and interruption (task is interrupted halfway), and my tests show that the behavior is exactly in line with expectation (the cleanup finalizer runs under all circumstances).

pchlupacek:
@djspiewak, i think test shall not interrupt the wye `before` it starts. Furthermore, I don't see any race condition in it. Id there is race condition, this means the wye implementation changed its invariants from what we have in master. 

1. wye runs the left and right process step with slight precedence to left side (right task will be run)
2. wye receives from left `true`, and as such terminates
3. wye driver terminates right side running outer cleanup on right side
4. process driver runs the downstream cleanup 
5. jvm is block to not exit for 3s giving chance to right task to complete, running its `rcv`  

I tried with you branch replacing the `Process(true)` with ` eval(Task.delay{ Thread.sleep(500); true})` and it performs the same, that means awaits right side to complete before wye runs its cleanups.

pchlupacek:
@djspiewak I just looked in your tests and this just tests the cleanups are run, but not when and how they are run, and I think this is not strong enough. the problem is that outer cleanup on right side gets only called in situation when T1 completes. You can easily see that by setting some insane delay in T1, and you would see that whole program would complete ONLY after that T1 delay elapses.

pchlupacek:
@djspiewak  I think this code demonstrates better the problem I am trying to explain (note flag.put/get position). I want cleanup 0 to be completed immediately once wye interrupts. With your branch this code deadlocks, and I think it shall not: 

```scala

  property("Outer cleanup through pipe gets called") = secure {

    @volatile var q = Queue[Int]()

    val flag = new SyncVar[Unit]

    val task1 = Task.fork(Task.delay[Unit]{ flag.get; Task.delay(1000); println("T1"); q = q.enqueue(1)})

    val awaitP =  await(task1)(u =>Process.emit(u))


    eval(Task.delay{ println("L"); true})
    .wye(
        awaitP.pipe(id)
        .onComplete(eval_(Task.delay{  flag.put(()); println("C0"); q = q.enqueue(0)}))
      )(wye.interrupt)
    .onComplete(eval_(Task.delay { println("C2"); q = q.enqueue(2) }))
    .run.run

    Thread.sleep(3000)
    println(q)
    q.toList == List(0,2,1)

  }

```

pchlupacek:
@djspiewak just to clarify I am using `djspiewak/bug/step` branch.

pchlupacek:
EDIT: 
After implementing task1 as 
```scala
   val task1 = Task.fork(Task.async[Unit]{ cb =>  Strategy.DefaultStrategy{Thread.sleep(1000); println("T1"); q = q.enqueue(1); cb(right(()))}})
```
Test passes on your branch. So this seems that there is just a problem with runAsync to block on Thread.sleep (remove the strategy and test won't pass again) . Do you think we can make runAsync to *NOT* block on blocking code in Await Task? I mean forking evaluation on separate thread? Because wye blocks then as well, and I think this is not desired.

djspiewak:
@pchlupacek Ok, so addressing things (mostly) in order…

I prefer to think of `wye` as being actually non-deterministic.  You're right though that the *first* batch of emits is left-biased, so your test probably is deterministic.  I think it's a lot better to structure test code such that it doesn't depend on such details (flipping the default value of the `leftBias` boolean inside of `wye` would change this semantic entirely).  This is an aside though…

The order of finalizers in structural relation is not a guarantee.  We make *absolutely* no guarantees about the order in which finalizers are run aside from the guarantees that are already present due to data dependence.  There are several far simpler examples I can give of finalizers inside of (and around) awaits where things are run "out of order".  The guarantees we make are that things will run, and will run once.  We shouldn't write tests that assume invariants that we do not promise.

Sidebar: `Task.apply {...}` = `Task.fork(Task.delay {...})`

Your example of a deadlock is concerning, and probably a bug.  I'll look at `runAsync` to see what's going on there.  We *are* forking off the awaited tasks, so I'm not sure what's happening or why it would be awaiting.  I'll track that down.

pchlupacek:
@djspiewak. Ok, agree with non-deterministic nature. I think the test actually does not depends on any of biasing details. 

On the cleanup order, I think as well it is not important, but in this specific situation it discovered the bug // behaviour that needs to be addressed. 

Other than that I think we are ok to merge this in the master, so please if you can just address the last issue I am ok to merge this. Thanks for excellent work.

djspiewak:
@pchlupacek I was able to reproduce the deadlock with a non-blocking test.  What is *really* interesting though is that it deadlocks with `Task.fork(Task.delay { ... })` but *not* with `Task.apply {...}`.  *Unbelievably* confusing.  Here is the exact test:

```scala
property("outer cleanup through pipe gets called before interrupted inner") = secure {
    import process1.id

    @volatile
    var complete = false

    @volatile
    var innerCleanup = false

    @volatile
    var outerCleanup = false

    @volatile
    var deadlocked = false

    val flag = new SyncVar[Unit]

    val task1 = Task fork (Task delay { deadlocked = flag.get(2000).isEmpty; Thread.sleep(1000); complete = true })
    // val task1 = Task { deadlocked = flag.get(2000).isEmpty; Thread.sleep(1000); complete = true }

    val awaitP =  await(task1)(u =>Process.emit(u))


    eval(Task delay { true })
      .wye(
          awaitP pipe process1.id onComplete eval_(Task delay { flag.put(()); innerCleanup = true })
        )(wye.interrupt)
      .onComplete(eval_(Task delay { outerCleanup = true }))
      .run.run

    complete :| "completion" &&
      innerCleanup :| "innerCleanup" &&
      outerCleanup :| "outerCleanup"
      !deadlocked :| "avoided deadlock"
  }
```

In the two versions of `task1`, the commented version passes the test, while the uncommented version does not.

djspiewak:
@pchlupacek I pushed a new version of bug/step.  Alissa and I fixed all the bugs referenced here, including the deadlock (`Task.fork` basically doesn't work, btw; like seriously, it doesn't work).  Unfortunately, we did discover a very timing-specific race condition in the `runAsync` implementation.  The window is quite narrow, and so it's a lot less common than the race conditions which existed in the old implementation (and the effect is less severe), but it is present nonetheless.  We have a solution, but we didn't have time to fully implement it today.

We're generally in favor of landing the branch as-is (unless you see other outstanding issues?) and continuing to fix the race condition we found, which should be doable in the next day or two.

PR #330, opened 2015-03-13
Cherry-picks `step` fixes from Pavel's PR

djspiewak:
This is a straight extension of #326, I just wanted to create a separate PR in order to keep the review clean.  I basically just cherry picked some commits from #318 and fixed the merge conflicts.  The result is a) more tests in `WyeSpec`, and b) a working `step` function!

BTW, I'd also like to petition we rename `runAsync` to `stepAsync`.  Its name implies that it runs the process all the way to the end, which is not at all what actually happens.

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -73,22 +73,25 @@ sealed trait Process[+F[_], +O]
    * users are responsible for ensuring resource safety.
    */
   final def step: HaltOrStep[F, O] = {
-    def go(cur: Process[F,O], stack: Vector[Cause => Trampoline[Process[F,O]]]) : HaltOrStep[F,O] = {
+    val empty: Emit[Nothing] = Emit(Nil)
+    @tailrec
+    def go(cur: Process[F,O], stack: Vector[Cause => Trampoline[Process[F,O]]], cnt: Int) : HaltOrStep[F,O] = {
       if (stack.nonEmpty) cur match {
-        case Halt(cause) => go(Try(stack.head(cause).run), stack.tail)
-        case Emit(os) if os.isEmpty => go(Try(stack.head(End).run), stack.tail)
+        case Halt(End) if cnt <=0  => Step(empty,Cont(stack))
+        case Halt(cause) => go(Try(stack.head(cause).run), stack.tail, cnt -1)
+        case Emit(os) if os.isEmpty => Step(empty,Cont(stack))
         case emt@(Emit(os)) => Step(emt,Cont(stack))
         case awt@Await(_,_) => Step(awt,Cont(stack))
-        case Append(h,st) => go(h, st fast_++ stack)
+        case Append(h,st) => go(h, st fast_++ stack, cnt -1)
       } else cur match {
         case hlt@Halt(cause) => hlt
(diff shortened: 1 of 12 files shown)


djspiewak:
It turns out that `njoin` and `wye` were assuming different invariants from `runAsync`!  This explains a lot, actually…  Anyway, `wye` assumes that `runAsync` can invoke the callback with an empty sequence of elements.  Which is intuitive, since it's a `Seq` and not a `NEL`.  Anyway, there is an explicit test for this case in `wye` and a special fast-path fallthrough if we land in that case, which is what I would expect.  *However,* `njoin` explicitly assumes that `runAsync` will not ever call back with an empty sequence!  This is ultimately what makes the `step` fix tricky.

Anyway, it turns out to be really easy (and remarkably efficient) to resolve this problem on `njoin`'s end, rather than trying to contort `runAsync` to only produce non-empty results.  It just gives rise to an additional actor message inside of `njoin`, but all of the existing state machinery is sufficient to catch the case and either abort the process early or continue on without inhibition (as in fact, one would expect given the other things `njoin` has to juggle).

pchlupacek:
@djspiewak yeah, good catch. I remember we had it somewhere !. Thanks a lot.

pchlupacek:
@djspiewak ah ok, your repo :-)

alissapajer:
@djspiewak just verified that this PR contains all of the wye specs that I had originally written that reproduced various bugs.

alissapajer:
We've fixed a number of other race conditions. At this point there are no known flaws, other than a test that would be nice to add, testing a particular race condition.

So @djspiewak and I are set for this to be merged.

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -1192,93 +1207,261 @@ object Process extends ProcessInstances {
...
+                  interrupted.get() match {
+                    case ptr @ -\/(int) => {
+                      if (interrupted.compareAndSet(ptr, \/-(cause))) {
+                        int()
+                      } else {
+                        referencedInterrupt(cause)
+                      }
+                    }
+
+                    case \/-(_) => ()    // ?????
```
pchlupacek: Can you perhaps comment the question marks?

pchlupacek:
@djspiewak @alissapajer I just add that one single comment to be updated. We perhaps need in future to revisit still the runAsync et all implementation, but I would say great work overall and LGTM.

src/main/scala/scalaz/stream/Process.scala
djspiewak: I'll elaborate on them.  Basically, that case can only occur if we are interrupted multiple times with different causes.  We always drop the later causes on the floor in that case.

pchlupacek:
@djspiewak are we ready to merge? not sure on this travis build error?

djspiewak:
@pchlupacek We're ready to merge.  I filed an issue for the error; it is unrelated to `runAsync`.

djspiewak:
Travis is being buggy again.  I'm pretty sure extending the timeout fixed the legitimate issue (which I believe was caused by travis being slow).  If we're all in consensus that this is good-to-go, I'm going to pull the trigger on the merge.

Issue #260, opened 2014-10-22
Redesign wye combinators for doing bounded merges and zips (was: behavior of wye.drainR and wye.drainL)


alissapajer:
`wye.drainR` (and `wye.drainL`, which is implemented in terms of `drainR`) are displaying behavior contradictory to what I expect. To note, I didn't find any examples, usages in the source, or tests for either of these two functions. What was the intent for these functions?

Here is the example I've constructed (thanks to @dcsobral for the idea to use `Process.awakeEvery`).

``` scala
import scalaz.concurrent._
import scalaz.stream._
import scala.concurrent.duration._

implicit val x = scalaz.stream.DefaultScheduler

def effectSink: Sink[Task, Int] = {
  def effect(ii: Int): Task[Unit] = Task delay { println(s"GOT $ii") }
  Process.constant(effect _)
}

// a process that can provide 17 Ints before halting, observed through a println
val initial: Process[Task, Int] = Process.emitAll(0 until 17) observe effectSink

val slowProcess: Process[Task, Duration] = Process.awakeEvery(1.seconds)

// 11 because it's prime, avoiding divisibility coincidences with the above 17
val result: Process[Task, Int] = initial.wye(slowProcess)(wye.drainR(11))

result.runLog.run
```

And here is the output I see, where the messsages `GOT 0` through `GOT 11` (12 total) print instantaneously, and then a subsequent `GOT` message prints every `1.seconds` until the process `initial` has naturally halted, at which point `result` halts as well and the `Vector` is printed.

``` scala
scala> result.runLog.run
GOT 0
GOT 1
GOT 2
GOT 3
GOT 4
GOT 5
GOT 6
GOT 7
GOT 8
GOT 9
GOT 10
GOT 11
GOT 12
GOT 13
GOT 14
GOT 15
GOT 16
res0: IndexedSeq[Int] = Vector(0, 1, 2, 3, 4, 5)
```

There are two (maybe three) behaviors demonstrated here that are contrary to what I expect.

[...]

PR #336, opened 2015-03-18
Removes wye.drainL and wye.drainR.

alissapajer:
See ticket #260. These methods are being removed rather than
deprecated because they do not work at all and should not be used.


djspiewak:
:+1:

alissapajer:
Any thoughts on this from anyone else? This could land in the 0.7 release, but I don't want to merge without at least another sign-off.

pchiusano:
Am fine w/ this.

PR #337, opened 2015-03-18
Adds liftSink constructor.

alissapajer:
Makes basic `Sink` construction easier.

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -826,6 +826,10 @@ object Process extends ProcessInstances {
   def iterateEval[F[_], A](start: A)(f: A => F[A]): Process[F, A] =
     emit(start) ++ await(f(start))(iterateEval(_)(f))
 
+  /** Promote a function `f` to a `Sink` that always emits that function */
+  def liftSink[F[_], O](f: O => F[Unit]): Sink[F, O] =
+    constant(f)
+
   /** Promote a `Process` to a `Writer` that writes nothing. */
   def liftW[F[_], A](p: Process[F, A]): Writer[F, Nothing, A] =
     p.map(right)


fthomas:
There is already `io.channel` which is like `liftSink` but for `Channel`. Would it make sense to move `io.channel` also into the `Process` companion?

djspiewak:
> There is already io.channel which is like liftSink but for Channel. Would it make sense to move io.channel also into the Process companion?

Oh, uh…  Yeah that totally achieves this same goal.  A `Channel` is a `Sink` where you instantiate the return type to be `Unit`, so `io.channel` is just a more general version of `liftSink`.

It *is* in a pretty obscure spot, honestly.  As evidenced by the fact that neither Alissa nor I thought of it.  :-)

fthomas:
I see value in also having `liftSink`. It makes your intention clear and gives you a compile error if the function does not return `F[Unit]`.

djspiewak:
What do you think of the following?

```scala
object Sink {
  def lift[F[_], A](f: A => F[Unit]): Sink[F, A] = Channel lift f
}

object Channel {
  def lift[F[_], A, B](f: A => F[B]): Channel[F, A, B] = ...
}
```

Would that be more discoverable?  Is that a sane place to put things?

alissapajer:
@fthomas Ah, awesome, `io.channel` is right there! Though, I do agree that there is additional value in a `liftSink` to guarantee the `Unit` in the return type at compile time.

@djspiewak where did you have in mind to put those two objects? We could put them at the `scalaz.stream` level. But if we did that, it would probably make sense to move some of the methods in the `io` object to the `Channel` object.

alissapajer:
@fthomas @djspiewak  My main concern with where we put these lift methods, is that people can find them easily. Putthing them in objects with the appropriate names (`Sink` and `Channel`) aids in that greatly.

Also, `Tee` and `Wye` and `Process1` already have their own objects at the `scalaz.stream` level. Which would be consistent with adding these two more objects. Though, we'd want them lower case for consistency: `object sink` and `object channel`.

djspiewak:
@alissapajer I was actually thinking it would make sense to put the Sink and Channel objects at the same level as their types. Unfortunately, I can't remember if the types are in Process or the package object. :-)

The tee, wye and process1 objects are intended to be more like modules or really packages that just contain functions.  I was thinking that a Channel and Sink object would be more evocative of a companion object, so the capitalization is significant.  This is analogous to how Trampoline in scalaz is just a type alias but it has some convenience constructors on an object of the same name and scoping.

alissapajer:
@djspiewak But how are these methods semantically different from what's in, say `wye.scala` already. Are those not constructors to create various `Wye`s?

When I decide that I want a `Sink` I want to know all the ways I can create a `Sink`. When I want to make a `Wye`, I go look in `object wye`. Same for `Tee`. But when I want to make a `Writer`, well, I have to go scan all the methods one by one in `object Process` to see what returns `Writer`.

Thus I was thinking about going further in the other direction with this, meaning that we put all the ways to create each type in one place, adding `object writer` as well, etc.

Also, the type `Sink`is in the package object :)

alissapajer:
@djspiewak To summarize my overall goal:  I want complete consistency in where we put constructors for all the types contained in the package object.

alissapajer:
@djspiewak I hadn't actually realized before, how varied the return types of the constructors in `object Process` actually are! We could move all the `Process0[...]` return types out into `process0`, and so forth. I'm happy to do something along these lines, just will wait until after discussing so as not to spend unnecessary time :)

djspiewak:
@alissapajer You make a compelling point.  I honestly am not sure what the most "consistent" and "unsurprising" thing would be here.  I guess we need more people to weigh in with opinions. :-)

We should fix the Process0 stuff to be consistent if/when we decide.

pchiusano:
I don't really have strong feelings about this. No decision seems particularly consistent...

I guess my inclination would be:

* Leave the `Process` companion object alone. I think it is convenient to have `await1`, etc there. I'm also fine with aliasing these elsewhere.
* Keep with the tradition of using lowercase module names for things like `process1`, `tee`, `wye`, etc. I slightly prefer this to making a `Tee`, `Sink`, `Channel` 'fake' companion objects.
* Introduce `sink` and `channel` modules as needed to house functions like this. For implicit syntax reasons, these may have to be aliased in syntax classes in the `Process` companion object. Though we can mix in these implicits in rather than actually further polluting the `Process.scala` file.

I'd like others to weigh in who might have stronger opinions.

alissapajer:
I think the `Process0` functions should come out of the `Process` companion object, so that that we treat them consistently with out we treat many `Process1` functions. Meaning, that we should create a `process0` object to parallel the `process1` object.

alissapajer:
@pchiusano The other issue I have with keeping, say `await1` in the `Process` companion object is that what's in the `Process` companion object and what's in the `process1` object seems to me to be arbitrary. 

For example, why is `awaitOption` in `object process1` and `await1` in `object Process`? Right now, the user has to look in (at least) two places to see what `Process1` functions are available to them. Aliasing would solve this, but then that's a lot of duplication, and aliasing across files is really hard to maintain.

fthomas:
@alissapajer :+1: for moving everything that is related to `Writer` out of `Process.scala`. Having them in `Process.scala` make them look essential. (I even have an old PoC for moving `Writer` out of `Process.scala`: https://github.com/scalaz/scalaz-stream/compare/master...fthomas:writer). Also :+1: for `channel` and `sink` modules.

I'm not convinced that a `process0` module would make things prettier. For example, `iterate` and `iterateEval` are very similar but differ in their return type (`Process0[A]` vs. `Process[F, A]`). With your proposal those would be in different files.

alissapajer:
@fthomas hmm, yeah as I look at `Process0` functions in the `Process` companion object, they are functions like `emitAll` which are truly essential functions on `Process`, and are ultimately often used to create something like `Process[Task, A]`. So as I think about it, I think keeping the `Process0` functions where they are does make sense. 

What I think actually could make sense is a rename to `Process0`. Right now we have:

```scala
 type Process0[+O] = Process[Nothing,O]
```

which is in no way analogous to the type `Process1`, even though the name suggests it might be similar. Maybe rename `Process0` to `ProcessNothing` or `ProcessBasic` or `ProcessBase`. I don't really like any of those, but I'm just brainstorming...

djspiewak:
I don't feel super-strongly about the `Process0` thing, but I will offer the following tidbit: I have never, *ever* used the `Process0` type except within `Process.scala` for the purposes of implicit conversions.  It's just not a useful type.  You can't interpret it, all you can do is use it to create a more interesting process.

pchlupacek:
`Process0` is usefull if you working with complex Process1 and allows you to nicely split logic to several functions that return Process0 instead of Seq/Option etc.

djspiewak:
@pchlupacek Oh of course the concept is useful, but I disagree on the usefulness of giving it a dedicated type alias (`Process0`).  Just type `Process[Nothing, A]`.

alissapajer:
So, now I'd like update my suggestion to suggest that instead of renaming `Process0`, we should just delete the alias altogether :)

pchlupacek:
Well, I think this is personal preference. I like it actually :-)

fthomas:
I like `Process0` too :-)

alissapajer:
Ok, so here's a proposal based on all the above comments:

(1) Keep the alias `Process0`, and keep all the functions returning `Process0` (`emitAll`, etc) where they currently are inside the `Process` companion object. Do not create a `process0` module, as this type isn't really useful outside of `Process`.

(2) Create `channel`, `sink` and `writer` modules, and move all functions returning the related types into those modules. "Essential" functions will be aliased in the `Process` companion object; this way the `Process` companion object remains a go-to place for the most-commonly-used functions.

How does that sound? 

I'll link the PR to this thread so more detailed discussions can be had, if necessary :)

pchiusano:
Sounds good to me.

pchlupacek:
+1

alissapajer:
Closing, as a new PR #343 was just opened.

Issue #338, opened 2015-03-20
Add a dequeueBatch function to async.queue


djspiewak:
Implement a `dequeueBatch: Process[Task, Seq[A]]` which dequeues *all* values currently backed up in the queue, or "continuation blocks" (the same as with `dequeue`) when no values are available.  No hysteresis or delay buffer is expected here.  Thus, if no values are in the queue, and a value is added, singleton sequence would be *immediately* reported.  However, if during the processing of that chunk, more than one element were enqueued, the *next* element of the `dequeueBatch` stream would be a multi-element sequence.

pchlupacek:
@djspiewak cool idea. Would be nice if we add `max` parameter to it tho, to still have sort of control over fairness. Wonder if we really can't implement some work stealing here, but likely not....

djspiewak:
@pchlupacek I tend to think of fairness as something that is controlled in `wye` and `njoin`, but you're right that it is a concern here as well.  A `max` parameter seems appropriate.

I can't think of any direct way to implement work stealing.  `Queue`'s contract is to dump values into a single, *sequential* consumer.

PR #342, opened 2015-03-21
Implemented batch dequeuing for async.Queue

djspiewak:
Satisfies #338.  Actually wasn't all that difficult.  The only downside I see is a *very* small constant-factor increase to the cost of dequeuing a single element (previously, the only supported use-case).  This factor comes from the boxing of that element inside a single-element list.  I'm not too concerned about it though, since all the other boxing utterly dwarfs this one addition.

### DIFF
--- a/src/main/scala/scalaz/stream/async/mutable/Queue.scala
+++ b/src/main/scala/scalaz/stream/async/mutable/Queue.scala
@@ -45,8 +45,22 @@ trait Queue[A] {
    *
    * Please use `Topic` instead of `Queue` when all subscribers
    * need to see each value enqueued.
+   *
+   * This process is equivalent to `dequeueBatch(1)`.
    */
   def dequeue: Process[Task, A]
+
+  /**
+   * Provides a process that dequeues in chunks.  Whenever *n* elements
+   * are available in the queue, `min(n, limit)` elements will be dequeud
+   * and produced as a single `Seq`.  Note that this naturally does not
+   * *guarantee* that `limit` items are returned in every chunk.  If only
+   * one element is available, that one element will be returned in its own
+   * sequence.  This method basically just allows a consumer to "catch up"
+   * to a rapidly filling queue in the case where some sort of batching logic
+   * is applicable.
+   */
+  def dequeueBatch(limit: Int = 0): Process[Task, Seq[A]]
 
   /**
    * The time-varying size of this `Queue`. This signal refreshes
@@ -98,11 +112,11 @@ private[stream] object Queue {
   def apply[A](bound: Int = 0)(implicit S: Strategy): Queue[A] = {
 
     sealed trait M
-    case class Enqueue (a: Seq[A], cb: Throwable \/ Unit => Unit) extends M
-    case class Dequeue (ref:ConsumerRef, cb: Throwable \/ A => Unit) extends M
+    case class Enqueue(a: Seq[A], cb: Throwable \/ Unit => Unit) extends M
+    case class Dequeue(ref: ConsumerRef, limit: Int, cb: Throwable \/ Seq[A] => Unit) extends M
     case class Fail(cause: Cause, cb: Throwable \/ Unit => Unit) extends M
     case class GetSize(cb: (Throwable \/ Seq[Int]) => Unit) extends M
-    case class ConsumerDone(ref:ConsumerRef) extends M
+    case class ConsumerDone(ref: ConsumerRef) extends M
 
     // reference to identify differed subscribers
     class ConsumerRef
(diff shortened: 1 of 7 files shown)


src/main/scala/scalaz/stream/async/mutable/Queue.scala
```diff
@@ -45,10 +45,24 @@ trait Queue[A] {
...
+   * Provides a process that dequeues in chunks.  Whenever *n* elements
+   * are available in the queue, `min(n, limit)` elements will be dequeud
+   * and produced as a single `Seq`.  Note that this naturally does not
+   * *guarantee* that `limit` items are returned in every chunk.  If only
+   * one element is available, that one element will be returned in its own
+   * sequence.  This method basically just allows a consumer to "catch up"
+   * to a rapidly filling queue in the case where some sort of batching logic
+   * is applicable.
+   */
+  def dequeueBatch(limit: Int = 0): Process[Task, Seq[A]]
```
alissapajer: Why is the `limit` set by default to `0`? Why not just let the caller choose their `limit`?

pchiusano: I'm guessing `0` is a sentinel value that means "give me all values available"? Not a fan... how about just two separate functions:

```scala
def dequeueAvailable: Process[Task, Seq[A]]
def dequeueBatch(limit: Int): Process[Task, Seq[A]]
```

If you want to implement these in terms of the same underlying primitive, that is fine.

alissapajer: @pchiusano yes you're right, I just found the spot. Down on line 185 we're checking `(limit <= 0)`, and then taking all available elements if that's the case.

I'd rather a `dequeueAvailable` function as well, at least in theory. Though the annoying thing with that solution, is that then in `dequeueBatch`, a `limit < =0` case still must be handled. Though, I suppose in this case we could have the semantic that nothing is ever dequeued? Or we could just add an assertion that `limit > 0` - not sure if assertions fit into this library?

pchiusano: Yeah, I would just bomb with an IllegalArgumentException in case of
negative values

src/main/scala/scalaz/stream/async/mutable/Queue.scala
```diff
@@ -158,24 +172,42 @@ private[stream] object Queue {
...
+        signalSize(queued.size)
+        if (unAcked.size > 0 && bound > 0 && queued.size < bound) {
+          val ackCount = bound - queued.size min unAcked.size
+          unAcked.take(ackCount).foreach(cb => S(cb(\/-(()))))
+          unAcked = unAcked.drop(ackCount)
+        }
       }
     }
 
+    def dequeueOne(ref: ConsumerRef, cb: (Throwable \/ A => Unit)): Unit = {
```
alissapajer: This is not used anywhere. Delete?

src/main/scala/scalaz/stream/async/mutable/Queue.scala
djspiewak: So, I'm not a fan of sentinel values either.  However, they *are* sort of the standard throughout the library wherever we have these sorts of bounds.  See, `njoin`, `boundedQueue` and more.  Given that we don't have a `Nat` type in Scala, I need to handle negative/zero values in any case (just as these other functions do).  I *could* through an `IllegalArgumentException`, but I'm not a fan of mixing eager exceptional error handling with lazy on-heap schemes.  It just tends to get confusing.

Since this *is* the convention throughout the library for bounds, I'm in favor of keeping it the way I implemented and changing everything over to a better scheme down the road.  Inconsistency in either direction seems like a bad idea.

src/main/scala/scalaz/stream/async/mutable/Queue.scala
djspiewak: Can do.  I originally thought I was going to implement the `Dequeue` case against this method (as it had been) and then have a separate `DequeueBatch` actor case, but this just worked out more cleanly.

src/main/scala/scalaz/stream/async/mutable/Queue.scala
pchiusano: Hmm, I really think throwing an IllegalArgumentException eagerly seems like the right thing to do. The stack trace will point to the immediate caller, even if the caller's caller is lazy, and this might be useful information. Delaying throwing until later is going to result in the stack trace being useless, and silently treating negative values as valid seems more likely to paper over what is probably a user bug. And you can always do:

```scala
(if (limit <= 0) q.dequeueAvailable else q.dequeueBatch(limit))
```

Not sure how `njoin` and `boundedQueue` slipped through the cracks, but I would fix those as well! :)

djspiewak: > Hmm, I really think throwing an IllegalArgumentException eagerly seems like the right thing to do. The stack trace will point to the immediate caller, even if the caller's caller is lazy, and this might be useful information. Delaying throwing until later is going to result in the stack trace being useless, and silently treating negative values as valid seems more likely to paper over what is probably a user bug.

This is the sort of thing I'm concerned about:

```scala
p1 ++ q.dequeueBatch(-1)
```

The implementation of `step` is going to lift the `IllegalArgumentException` into the stream, treating it as an error in the evaluation rather than an error in the configuration.

> Not sure how njoin and boundedQueue slipped through the cracks, but I would fix those as well! :)

Well, particularly with `njoin` (which has *two* of these bounding parameters!), I think they slipped through the cracks because it becomes really ugly even just to *name* the quadratically-expanding set of different configuration special cases.

alissapajer:
@pchiusano It would be nice to get this in to 0.7 release as well. I good waiting until this PR and #343 are both merged, and then having release cut after that.  /cc @djspiewak

src/main/scala/scalaz/stream/async/mutable/Queue.scala
pchiusano: > `p1 ++ q.dequeueBatch(-1)`

Gotcha. Although in that case, the stack trace is going to be useless regardless of whether the error is thrown eagerly or not.

So, here's what I'd like to do:

* For `boundedQueue`, we already have `unboundedQueue`, so go ahead and make `boundQueue` parameter required, throw illegal arg exception eagerly.
* Same for `dequeueBatch`
* I'm fine with keeping `njoin` as is for now. Maybe we can revisit later.

> Personally, I actually feel that it's less ugly to define a subrange of Int to have a particular meaning, rather than making the function partial.

In this situation, I think more often than not, negative values are a bug (like the user flipped arguments in compute a limit, say). Obviously, this is somewhat subjective, but it's the decision I'd like to go with.

pchiusano:
@alissapajer agree

djspiewak:
@pchiusano Working on the last change now.  Should be up very shortly.

djspiewak:
Random note…  Now that I think about it, it's actually a bit scary that `mergeN` just silently defaults to unbounded queueing.  `Exchange` has the same problem, but of course it is far less widely used.

djspiewak:
@pchiusano This is ready to merge.  Travis is sort of mad, but it's mad because the `queue.size-signal` test is nondeterministic *as written!*  `Signal` is very up front about the fact that `discrete` can lose values under contention, and the test is not resilient to this.  :-S  In any case, that's the only failing thing.

pchiusano:
LGTM, merging..

PR #343, opened 2015-03-22
Refactor modules

alissapajer:
Refactors modules per discussion in #337 .

### DIFF
--- a/src/main/scala/scalaz/stream/writer.scala
+++ b/src/main/scala/scalaz/stream/writer.scala
@@ -5,8 +5,12 @@ import scalaz.\/._
 
 object writer {
 
+  /** Promote a `Process` to a `Writer` that outputs nothing. */
+  def liftW[F[_], W](p: Process[F, W]): Writer[F, W, Nothing] =
+    p.map(left)
+
   /** Promote a `Process` to a `Writer` that writes nothing. */
-  def lift[F[_], A](p: Process[F, A]): Writer[F, Nothing, A] =
+  def liftO[F[_], O](p: Process[F, O]): Writer[F, Nothing, O] =
     p.map(right)
 
   /**
@@ -17,4 +21,3 @@ object writer {
     p.flatMap(a => Process.emitAll(Vector(left(a), right(a))))
 
 }
-
--- a/src/main/scala/scalaz/stream/writer.scala
+++ b/src/main/scala/scalaz/stream/writer.scala
@@ -6,7 +6,7 @@ import scalaz.\/._
 object writer {
 
   /** Promote a `Process` to a `Writer` that writes nothing. */
-  def liftW[F[_], A](p: Process[F, A]): Writer[F, Nothing, A] =
+  def lift[F[_], A](p: Process[F, A]): Writer[F, Nothing, A] =
     p.map(right)
 
   /**
--- a/src/main/scala/scalaz/stream/writer.scala
+++ b/src/main/scala/scalaz/stream/writer.scala
@@ -1,0 +1,20 @@
+package  scalaz.stream
+
+import scalaz.\/._
+
+
+object writer {
+
+  /** Promote a `Process` to a `Writer` that writes nothing. */
+  def liftW[F[_], A](p: Process[F, A]): Writer[F, Nothing, A] =
+    p.map(right)
+
+  /**
+   * Promote a `Process` to a `Writer` that writes and outputs
+   * all values of `p`.
+   */
+  def logged[F[_], A](p: Process[F, A]): Writer[F, A, A] =
+    p.flatMap(a => Process.emitAll(Vector(left(a), right(a))))
+
+}
+
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1119,6 +1119,46 @@ object Process extends ProcessInstances {
     }
   }
 
+  implicit class WriterTaskSyntax[W, O](val self: Writer[Task, W, O]) extends AnyVal {
+
+    /**
+     * Returns result of channel evaluation on `O` side tupled with
+     * original output value passed to channel.
+     */
+    def observeOThrough[O2](ch: Channel[Task, O, O2]): Writer[Task, W, (O, O2)] = {
+      val observerCh = ch map { f =>
+        in: (W \/ O) => in.fold(w => Task.now(-\/(w)), o => f(o).map(o2 => \/-(o -> o2)))
+      }
+      self through observerCh
+    }
+
+    /** Returns result of channel evaluation on `W` side tupled with
+     * original write value passed to channel.
+     */
+    def observeWThrough[W2](ch: Channel[Task, W, W2]): Writer[Task, (W, W2), O] = {
+      val observerCh = ch map { f =>
+        in: (W \/ O) => in.fold(w => f(w).map(w2 => -\/(w -> w2)), o => Task.now(\/-(o)))
+      }
+      self through observerCh
+    }
+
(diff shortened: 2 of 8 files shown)


src/main/scala/scalaz/stream/writer.scala
```diff
@@ -0,0 +1,20 @@
+package  scalaz.stream
+
+import scalaz.\/._
+
+
+object writer {
+
+  /** Promote a `Process` to a `Writer` that writes nothing. */
+  def lift[F[_], A](p: Process[F, A]): Writer[F, Nothing, A] =
```
pchlupacek: I actually found liftW syntax useful, as you can do both liftW and liftO. 
In fact I would like to have whole bunch of `O` and `W` combinators added, they work pretty well if you work with writer. I can put you some source of them in comment, so you can perhaps add them, if you are interested?

pchlupacek: Ah sorry, disregard liftW/O, I completely overlooked this signature. But still if you want to add more `W` and `O` stuff to process LMK, I'll past them here.

djspiewak:
:+1:

src/main/scala/scalaz/stream/writer.scala
alissapajer: @pchlupacek I actually think it would be better to rename this `lift` to `liftW` and then to add in an analogous `liftO`, returning `Process[F, W, Nothing]`.

What did you have in mind for other `W` and `O` stuff? This is related to #344 on renaming some `WriterSyntax` functions. There are already a few `W` and `O` combinators in there.

pchiusano: > @pchlupacek I actually think it would be better to rename this lift to liftW and then to add in an analogous liftO, returning Process[F, W, Nothing].

+1

pchiusano:
I'd actually like to get this into 0.7, given that 0.7 already has some other reshuffling of functions formerly in Process (the stuff in `time`). I'll branch after this gets merged.

alissapajer:
@pchiusano ok awesome!

The last thing I might do is add in a `writer.liftO`, but I'm not actually sure if it's a useful function. It wouldn't have any call sites. So I'm ok merging this as is, unless anyone ( @pchlupacek ) thinks `liftO` would be useful to include now?

src/main/scala/scalaz/stream/writer.scala
pchiusano: Wait, @alissapajer do you have these flipped - shouldn't `liftO` return the `Writer[F, Nothing, A]` and `liftW` return a `Writer[F, A, Nothing]`?

pchlupacek: @pchiusano I will post some stuff we use for writer syntax and are not part yet in lib. If @alissapajer can perhaps just cut and past them into lib? Into WriterSyntax...

alissapajer: @pchlupacek  Sure, I'm adding `liftO` and `liftW` right now, so paste away!

@pchiusano I might have had them backwards before. This is what's on my branch now:

```scala
  /** Promote a `Process` to a `Writer` that writes nothing. */
  def liftW[F[_], O](p: Process[F, O]): Writer[F, Nothing, O] =
    p.map(right)

  /** Promote a `Process` to a `Writer` that outputs nothing. */
  def liftO[F[_], W](p: Process[F, W]): Writer[F, W, Nothing] =
    p.map(left)
```

pchlupacek:
@alissapajer can you perhaps add following code to syntax? I am not sure on the naming, but lets starts with these, if you won't come with something better. @pchiusano not sure if you agree on naming and signatures, but we found them all quite useful....

Also we parhaps need to add observeWThrough that is missing :-)

```scala  

implicit class ProcezzWriterSyntax[W,O](self: TWriter[W,O]) {

    /**
     * Observe the right side of this `Writer` using the
     * given `Sink`, keeping it available for subsequent
     * processing.
     */
    def observeO(snk: TSink[O]): TWriter[W,O] =
      self.zipWith(snk)((a,f) =>
        a.fold(
          (w: W) => Process.emit(left(w)),
          (o: O) => eval_ { f(o) } ++ Process.emit(right(o))
        )
      ).flatMap(identity)

    /**
     * returns result of channel evaluation on `O` side tupled with original value passed to channel
     * @param ch
     * @tparam O2
     * @return
     */
    def observeOThrough[O2](ch:TChannel[O,O2]):TWriter[W, (O,O2)] = {
      val observerCh = ch.map(f => (wo:W \/ O) => wo.fold(w=>Task.now(-\/(w)),o => f(o).map(o2=> \/-(o -> o2)) ))
      self.through(observerCh)
    }

    def throughO[O2](ch:TChannel[O,O2]):TWriter[W,O2] = {
      val ch2 = ch.map { f ⇒ in:(W \/ O) ⇒ in.fold(w ⇒ Task.now(left(w)), o ⇒ f(o).map(right))}
       self.through(ch2)
    }

    def throughW[W2](ch:TChannel[W,W2]):TWriter[W2,O] = {
      val ch2 = ch.map { f ⇒ in:(W \/ O) ⇒ in.fold(w ⇒ f(w).map(left), o ⇒ Task.now(right(o)))}
      self.through(ch2)
    }

  }


```

pchlupacek:
@alissapajer  and this goes for normal sources: 

```scala

  /** returns result of channel evaluation tupled with original value passed to channel **/
    def observeThrough[O2](ch:TChannel[O,O2]):TSource[(O,O2)] = {
      val observerCh = ch.map(f => (o:O) => f(o).map(o2=> o -> o2) )
      self.through(observerCh)
    }

```

src/main/scala/scalaz/stream/writer.scala
alissapajer: @pchiusano see code right above. I might be switching the naming convention here. Should the `Writer` that "writes nothing" be actually called `liftO`. This actually makes more sense, since we're lifting a `Process[F, O]` (a `Process` with "outputs") into a `Writer` that outputs those outputs :) So I'll switch what I just pasted above, agreed?

pchiusano: Yep, agreed.

alissapajer:
@pchlupacek `WriterSyntax.observeO` and `WriterSyntax.observeW` are both already in master, generalized so they're not specific to `Task`, so we're good on that front.

The various `through` combinators you suggest all require an applicative `F` (it looks like we just need `pure` and `fmap`, not `bind`). As a general note (/cc @pchiusano ) I think that it would be great to have more generalized combinators that just require an applicative or a monad, that way we wouldn't have to hard code on `Task`. For now, we already have for example a `SinkTaskSyntax` class, so we can go the hard-coded `Task` route on this PR so we can get it in 0.7 :)

djspiewak:
@alissapajer I generally agree that hard-coding on `Task` is bad.  Unfortunately, things on that front are a bit in flux right now since we're in the process of trying to get away from our rigid dependency on Scalaz.

In any case, doesn't affect this PR, just future work.

alissapajer:
okie doke, all changes I had queued have been pushed!

pchiusano:
LGTM, @pchlupacek can you merge assuming it looks good to you.

djspiewak:
@pchiusano I was sort of hoping we could get an 0.7 today, and it's quite late for @pchlupacek now.  I mean, I'm never one to preempt someone else's review, but if Pavel is in fact offline for the night, could we go ahead with the merge and release?

pchiusano:
@djspiewak No, would rather not rush it, sorry. We can do first thing tomorrow.

djspiewak:
@pchiusano Fair enough.  :-)  Again, no disrespect intended to Pavel.  I just want to make sure we don't fall into release drift.

PR #349, opened 2015-03-27
Fix io.bufferedResource and io.resource to use await instead of flatMap

### LINKED ISSUE #348: io.bufferedChannel/bufferedResource/resource can leak resource [0.7a]
The [0.7 wiki](https://github.com/scalaz/scalaz-stream/wiki/0.7#-finalizers) mentions that
> ...
>You will also note that we only make this sort of resource guarantee with respect to finalizers that are _inside_ of the `await`! Thus, the following code is unsound and will leak resources!
```scala
Process.eval(openConnection) flatMap { conn =>
  doStuff(conn) onComplete (Process eval_ closeConnection(conn))
}
```
> Notice that the finalizer which releases the resource (`closeConnection(conn)`) is inside of a `flatMap`, and _not_ an `await`. If `openConnection` is interrupted but completes anyway, the entire `flatMap` will be nulled out and the finalizer will not run, despite the fact that we _did_ produce a value `conn`. Finalizers are _only_ guaranteed to be safe inside of `await`.

Now here's the implementation of `resource`, on which `bufferedChannel` also depends:

```scala
def resource[F[_],R,O](acquire: F[R])(
                         release: R => F[Unit])(
                         step: R => F[O]): Process[F,O] =
    eval(acquire).flatMap { r =>
      repeatEval(step(r)).onComplete(eval_(release(r)))
    }
```

It's doing exactly what the wiki calls unsound.

dansanduleac:
This should fix #348.

### DIFF
--- a/src/main/scala/scalaz/stream/io.scala
+++ b/src/main/scala/scalaz/stream/io.scala
@@ -27,7 +27,7 @@ object io {
   def bufferedResource[F[_],R,O](acquire: F[R])(
                             flushAndRelease: R => F[O])(
                             step: R => F[O]): Process[F,O] =
-    eval(acquire).flatMap { r =>
+    await(acquire) { r =>
       repeatEval(step(r)).onComplete(eval(flushAndRelease(r)))
     }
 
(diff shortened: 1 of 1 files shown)


djspiewak:
:+1:

fthomas:
Thanks for the fix, @dansanduleac. If you like to, please submit another PR adding yourself to CONTRIBUTORS.md.

Issue #347, opened 2015-03-27
Disallow race condition while using signal


pchlupacek:
Current signal constructors allow to create race programs, that may not be obivous to user. We need to clear signal interfaces to eventually not allow these conditions. 

Currently, we have two constructors of signal: `signalOf` and `signal` (deprecated). There is a race condition using signalOf to lose first value (possibly more, but first is almost guaranteed) set: 

```scala 

  property("signalOf.init.value") = secure {
    val timer = time.sleep(1.second)
    val signal = async.signalOf(0)

   (eval_(signal.set(1)) ++ signal.discrete.once ++ timer ++ signal.discrete.once).runLog.run ?= Vector(1,1)
  }

```

I think we have to either change `signalOf` implementation to not use `toSignal` and rather do initial value with passing initial `W` in implementation `writer1`. 

Essentially any Signal that is fed from upstream shall not be settable with `set`, `compareSet` etc. Similarly the signal created to be set from `downstream` shall never receive any values from upstream. 

@djspiewak  any thoughts on this?

djspiewak:
It actually surprises me quite a bit that this generates a race condition.  It appears that the topic inside of `Signal` makes itself available before it has flushed through the init process.  In other words, the sane assumption about `toSignal` doesn't hold.

I think we actually need to fix `Signal` to hold incoming messages until `inits` has been fully `unemit`ed.  This would also fix the identical race condition in `toSignal`.

pchlupacek:
I just put that PR. I don't think so its good tenchnique to have merging semantics hardcoded in `Signal`. I think we have either signal that is set by evaluating Tasks ( I call this downstream signal) or, eventually is constructed by "upstream" process. I think both of them shall not be in our code w/o involving nondet combinator like `wye`

PR #350, opened 2015-03-27
Fixed signalOf race condition of initial value + added test to that. Als...

pchlupacek:
...o removed the deprecation on async.signal.

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -9,6 +9,7 @@ import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
 import scala.annotation.tailrec
 import scala.collection.SortedMap
 import scala.concurrent.duration._
+import scalaz.stream.async.immutable.Signal
 import scalaz.{\/-, Catchable, Functor, Monad, Monoid, Nondeterminism, \/, -\/, ~>}
 import scalaz.\/._
 import scalaz.concurrent.{Actor, Future, Strategy, Task}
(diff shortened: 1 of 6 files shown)


djspiewak:
I strongly disagree with removing the deprecation.  It doesn't make any sense to have a signal without a value, which is in fact the argument that was made when the deprecation was added in the first place.

pchlupacek:
@djspiewak, @pchiusano  please can you shortly look on this PR? This fixes #347. Also contrary to previous opinion I think we SHALL NOT deprecate `async.signal`. There are use cases where we need it (i.e. `time.awakeEvery`) and I don't see any benefit with replacing this with `async.signalOf[Option[Int]](None).discrete.collect{ case Some(x) =>  x }` although this can be  option as well, if we really really insist on deprecating the `async.signal`. Additionally, I have changed signature of `async.toSignal` from `mutable.Signal` to `immutable.Signal` to prevent race conditions between upstream/downstream mutation of signal. 

If still someone would like to have both in same Signal, wye IMHO has to be used instead. Perhaps, I may came with signal combinators (like wye, merge) in separate PR.

pchlupacek:
@djspiewak thanks on opinion. There is in fact reasoning behind signal that has no value set, and some example is in `time.awakeEvery`. There are some more similar use cases of using signal without value to really signal some sort of `start` of some other computation. 

As mentioned in explanation I was just writing, we can still keep deprecation and replace it with some sort of combinator, perhaps even make the `unset` behaviour some sort of explicit syntax. 

I still think that `unset` signal is easier discoverable than race condition within signal, but, again we perhaps may have both.

pchlupacek:
@djspiewak this is what I did:

- deprecated back `async.signal`
- introduced `async.signalUnset` for explicit `unset` behaviour
- removed `haltOnSource` from `async.toSource`, hence I never saw `async.toSignal` with anything else than setting this to `true` and hence `toSignal` is immutable now, you don't have any other way than terminating the `source` to stop the signal anyhow. 

LMK what do you think.

djspiewak:
LGTM.  I know it's just a name, but `signalUnset` doesn't suggest to the user that this is the "primary" way to get a `Signal`.  So users can go for it if and only if they need the unset behavior.

pchlupacek:
@djspiewak I agree with you that `unexpected` or `important` behaviours shall be explicitly visible to user, and agree that this was confusing on `signal`. Thanks.

PR #352, opened 2015-03-29
Move most syntax classes out of Process.scala

fthomas:
With the addition of `channel.scala`, `sink.scala`, and `writer.scala` there is the possibility to move the syntax classes for `Channel`, `Process1`, `Sink`, `Tee`, `Writer`, and `Wye` out of `Process.scala`.

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -1,15 +1,12 @@
 package scalaz.stream
 
-import java.util.concurrent.atomic.AtomicReference
-
 import Cause._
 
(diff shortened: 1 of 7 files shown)


djspiewak:
Doesn't this negate the benefit of the value classes?  I was under the impression that scalac did extra optimizations when it was specifically an implicit value class, rather than an implicit def returning a value class.

fthomas:
https://gist.github.com/fthomas/0f914bf5c5abda57cce1

If I interpret this correctly, there is no difference in using an implicit value class or an implicit def + value class.

djspiewak:
@fthomas I think your interpretation is correct!  :+1: for merge.

pchiusano:
+1 for merge.

PR #358, opened 2015-04-05
Make drain{O,W} consistent with Process.drain

### LINKED ISSUE #344: Rename WriterSyntax.drain{W,O}?
`Process.drain` is defined as
```
def drain: Process[F, Nothing] = flatMap(_ => halt)
```
while `WriterSyntax.drainW` and `WriterSyntax.drainO` are defined as
```
def drainW(snk: Sink[F,W]): Process[F,O] = observeW(snk).stripW
def drainO(snk: Sink[F,O]): Process[F,W] = observeO(snk).stripO
```
IMO this is unfortunate because `Process.drain` suggests that those two are more like
```
def drainW: Writer[F, Nothing, O] = self.flatMapW(_ => halt)
def drainO: Writer[F, W, Nothing] = self.flatMapO(_ => halt)
```
which are useful functions on their own I'd like to add later. Should we rename `drainW` and `drainO`? What would be better names for them?

fthomas:
This PR fixes #344. In its current form it just changes `drainO` and `drainW` to be consistent with `Process.drain`.

### DIFF
--- a/src/main/scala/scalaz/stream/package.scala
+++ b/src/main/scala/scalaz/stream/package.scala
@@ -47,7 +47,7 @@ package object stream {
 
   /**
    * A `Writer[F,W,O]` is a `Process[F, W \/ O]`. See
-   * `Process.WriterSyntax` for convenience functions
+   * `WriterSyntax` for convenience functions
    * for working with either the written values (the `W`)
    * or the output values (the `O`).
    *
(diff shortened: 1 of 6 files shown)


alissapajer:
+1 to merge - any objections?

pchlupacek:
+1

Issue #258, opened 2014-10-21
upload to maven repository


advoretsky:
project builds are absent in maven-central - it fails builds

jilen:
+1 Suprised this excellent library haven't uploaded to maven central repo yet!

maslovalex:
+1 specs2 needs it

pchlupacek:
Folks hence the current sbt version is using bintray as one of the default repos do you think this is still necessary?

advoretsky:
It's only second time I meet mention of [bintray](http://jcenter.bintray.com/). I was stucked trying to build my project with Maven. If adding next snippet solves the problem - Ok.

        <repository>
          <id>central</id>
          <name>bintray</name>
          <url>http://jcenter.bintray.com</url>
        </repository>

Putting it somewhere in Readme.md would help me

briantopping:
That's a temporary fix, but repositories come and go. Capturing a repository like that means that the source may not be able to be built if bintray were to not exist in the future. So it is considered bad form to put a repository into a POM like that. 

Note that I am not predicting the demise of Bintray, but that Bintray is just another repository, none last forever.

See http://blog.sonatype.com/2009/02/why-putting-repositories-in-your-poms-is-a-bad-idea/ for complete information.

edmondo1984:
+1 When is this planned for?

beloglazov:
+1 would be awesome to have it in Maven central!

frgomes:
+1 Please publish recent versions to Maven Central, including builds for Scala 2.11. Please!

xuwei-k:
+1

pchiusano:
I'm also fine with this.

djspiewak:
It's relatively easy to get something into Maven Central.  The biggest issues are the following:

- [ ] We need to make sure our generated POM metadata conforms to Sonatype standards.  That means things like contributor lists, license information, and stricter schema validation
- [ ] We need to register GPG public keys for anyone who is going to be pushing releases.  I have a public key already registered with Sonatype, but ideally I wouldn't be the only person capable of cutting a release.  :-)
- [ ] We need to present a good argument to Sonatype as to why our artifact name doesn't match our package name
- [ ] We need to present a *substantially better* argument to Sonatype as to why our artifact name does not match a domain that we own (this is usually a sticking point)
[...]

PR #359, opened 2015-04-08
Use HTTPS

rtyley:
HTTPS is available:

https://dl.bintray.com/scalaz/releases/

Incidentally, I really prefer to not have to add a new resolver with every new library I encounter. Is it possible to get this synced to Maven Central? This makes even more sense since SBT downloads from Maven Central _with HTTPS_ by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ scalaz-stream
 To get the latest version of the library, add the following to your SBT build:
 
 ``` scala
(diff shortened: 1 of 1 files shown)


fthomas:
:+1: 

Uploading to Maven Central is already planned, see #258.

alissapajer:
+1

pchlupacek:
thanks

PR #361, opened 2015-04-09
Circular buffers

runarorama:
Adds circular buffers.

I believe this also fixes an off-by-one error in async.Queue, where the queue will begin blocking producers at size `bound - 1` rather than `bound`. And if the queue size was at `bound`, it would have to be dequeued twice to see any results.

Review by @pchlupacek appreciated.

### DIFF
--- a/src/main/scala/scalaz/stream/async/mutable/Queue.scala
+++ b/src/main/scala/scalaz/stream/async/mutable/Queue.scala
@@ -108,11 +108,6 @@ trait Queue[A] {
 }
 
 private[stream] object CircularBuffer {
-  /**
-   * Builds a queue that functions as a circular buffer. Up to `bound` elements of
-   * type `A` will accumulate on the queue and then it will begin overwriting
-   * the oldest elements. Thus an enqueue process will never wait.
-   */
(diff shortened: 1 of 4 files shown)


src/main/scala/scalaz/stream/async/mutable/Queue.scala
```diff
@@ -107,20 +107,31 @@ trait Queue[A] {
   private[stream] def failWithCause(c:Cause): Task[Unit]
 }
 
+private[stream] object CircularBuffer {
+  /**
```
pchiusano: Since this is private, it won't show up in generated API docs, so could we just move this description down to the `circularBuffer` method?

pchiusano:
LGTM, I just made a minor comment on moving the docs. @pchlupacek could you review before merging?

pchiusano:
@runarorama Btw, you should add yourself to https://github.com/scalaz/scalaz-stream/blob/master/CONTRIBUTORS.md unless you wish to remain stealthy for some reason :)

pchlupacek:
@runarorama  thanks for catching off-by-one bug. I like also the `mk` instead of `apply` as well, perhaps I submit the PR to make this consistent across all concurrency primitives.

PR #364, opened 2015-04-13
Use async.signalUnset instead of deprecated async.signal.

### LINKED ISSUE #310: Do not use deprecated async.signal in the tests
`AsyncSignalSpec` and `MergeNSpec` still make use of the deprecated `async.signal`. I guess they should be rewritten with `async.signalOf`.

alissapajer:
`async.signal` is [deprecated](https://github.com/scalaz/scalaz-stream/blob/2ba023db717fdaa03a50977ef3718ed67661bbd2/src/main/scala/scalaz/stream/async/package.scala#L40) and already just calling `async.signalUnset`

this just removes compiler warnings and should close #310


pchlupacek:
Thx:-)

PR #365, opened 2015-04-13
object -> class in tests to prevent object init deadlocks

### LINKED ISSUE #363: Change all specs to be classes instead of objects
Current object specs are subject of object initialization deadlock in certain concurrency scenarios. we just need to convert our specs to be classes to fix this. 

Any volunteers to make PR with object -> class refactoring on all our specs are welcomed :-)

alissapajer:
closes #363

### DIFF
--- a/src/test/scala/scalaz/stream/TeeSpec.scala
+++ b/src/test/scala/scalaz/stream/TeeSpec.scala
@@ -21,9 +21,10 @@ import process1._
 import TestInstances._
 import scala.concurrent.SyncVar
 
-object TeeSpec extends Properties("Tee") {
+class TeeSpec extends Properties("Tee") {
 
   import TestInstances._
+  import TeeSpecHelpers._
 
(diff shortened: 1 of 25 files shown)


src/test/scala/scalaz/stream/TeeSpec.scala
```diff
@@ -149,3 +146,9 @@ object TeeSpec extends Properties("Tee") {
       ("left fails - right first" |: l.tee(emit(2))(rightFirst).expectedCause(_ == Error(Bwahahaa)).toList == List(2, 1, 3))
   }
 }
+
+object TeeSpecHelpers {
+  implicit class Next[F[_],O](val p: Process[F,O]) extends AnyVal {
```
alissapajer: a value class may not be a member of another class - hence it's moved out here

pchlupacek:
excellent. Thanks !

PR #368, opened 2015-04-13
Implemented sbt-git

djspiewak:
Remember that, in SBT, `version` is a `SettingKey` and not a `TaskKey`, meaning that the value is computed *once* and not recomputed until reload!  Thus, if you make a new commit, tag or otherwise do something that *should* modify the value of `version`, you need to `reload` to see it take effect.

When in doubt, `show version`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,8 +10,6 @@ git.gitTagToVersionNumber := {
   case _ => None
 }
 
-// git.formattedShaVersion := git.gitHeadCommit.value      // bintray really hates -SNAPSHOT versions, so we just use the hash
-
 scalaVersion := "2.11.6"
 
 crossScalaVersions := Seq("2.10.5", "2.11.6")
--- a/build.sbt
+++ b/build.sbt
@@ -1,8 +1,16 @@
+enablePlugins(GitVersioning)
+
 organization := "org.scalaz.stream"
 
 name := "scalaz-stream"
 
-version := "snapshot-0.7"
+val ReleaseTag = """^release/([\d\.]+a?)$""".r
+git.gitTagToVersionNumber := {
+  case ReleaseTag(version) => Some(version)
+  case _ => None
+}
+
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -1,8 +1,16 @@
...
 name := "scalaz-stream"
 
-version := "snapshot-0.7"
+val ReleaseTag = """^release/([\d\.]+a?)$""".r
+git.gitTagToVersionNumber := {
+  case ReleaseTag(version) => Some(version)
+  case _ => None
+}
+
+git.formattedShaVersion := git.gitHeadCommit.value      // bintray really hates -SNAPSHOT versions, so we just use the hash
```
fthomas: Does this line prevent appending "-SNAPSHOT" to the version? I think sbt-git does this only if there are uncommitted changes. So it should not be really necessary.

If it is still necessary could you change the comment to something more informative like "bintray does not allow -SNAPSHOT versions ..."?

djspiewak: > I think sbt-git does this only if there are uncommitted changes.

Nope, it does it all the time.  Comment out the line to see!  :-)

> If it is still necessary could you change the comment to something more informative like "bintray does not allow -SNAPSHOT versions ..."?

I could change the comment if you would like.  Strictly speaking, bintray *does* allow SNAPSHOT versions, you just need to configure your repo a little differently.  Given that we're publishing *stable* snapshots though (i.e. we're not republishing), the `-SNAPSHOT` prefix would actually do more harm than good by defeating Maven's (and Ivy's) caching assumptions.

alissapajer:
+1 to merge

build.sbt
fthomas: Removed that line, `git commit -a`, `sbt version`:
```
[info] 838bef4447a3e16d150a1583f9b76916a8dfd3c1
```
No -SNAPSHOT there.

djspiewak:
@fthomas You're correct.  :-)  Fixed!

fthomas:
:+1:

PR #370, opened 2015-04-20
hash improvements, performance and type-safety

jedws:
uses ByteBuffer to supply the bytes to the MessageDigest for performance improvement, tags the returned ByteVectors as being of a specific output type matching the selected hash algorithm, and provides a typeclass like provider for the MessageDigest selection

### DIFF
--- a/src/main/scala/scalaz/stream/hash.scala
+++ b/src/main/scala/scalaz/stream/hash.scala
@@ -6,6 +6,8 @@ import scodec.bits.ByteVector
 import Process._
 import process1._
 
+import scalaz.{Tag, @@}
+
 /**
  * @define MutableProcess1 [[https://github.com/scalaz/scalaz-stream/blob/master/src/test/scala/scalaz/stream/examples/MutableProcess1.scala `MutableProcess1`]]
  * @define MutableImpl @note This process uses mutable state as an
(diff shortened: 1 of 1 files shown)


mpilquist:
One additional performance improvement -- `ByteVector` has a `digest` method that traverses the tree of byte vector nodes, calling `update` on each chunk.

pchiusano:
LGTM! But could you apply the improvement suggested by @mpilquist

jedws:
@mpilquist @pchiusano the `ByteVector.digest(MessageDigest)` method doesn't work as it calls the `MessageDigest.digest` method which finalises and resets the digest computation http://docs.oracle.com/javase/7/docs/api/java/security/MessageDigest.html#digest()

I did some very basic playing around with numbers, just hashing large zeroed byte arrays:

    val all = List.fill(100)(ByteVector(Array.ofDim[Byte](1000000)))
    val big = all.foldLeft(ByteVector.empty)(_ ++ _)
    def time[A](a: => A): (A, Long) = { val s = System.nanoTime; (a, System.nanoTime - s) }

    time { hash.sha1(all) }
    time { big.digest("SHA-1") }

and running those many many times.

The results are all pretty much of a muchness, with the toByteBuffer one seeming to be slightly ahead, but within the margins of statistical error. I didn't run a complete perf analysis though.

mpilquist:
Oh right, I forgot we need a running hash here. The performance test results make sense to me -- the built-in digest should outperform `toByteBuffer` only when the internal tree has non-trivial depth.

jedws:
right, it'd be nice to have `foreachV` or equivalent available to us so we
could get the most efficient access to the underlying bytes.

fthomas:
This also looks good to me. But note that this adds a dependency on scalaz to the `hash` module which we are trying to get rid off (#321).

jedws:
There isn't a lot to type tagging, basically just a couple of type aliases and a Tag class to do the work. You can't bring all of it in, as you'd need a few extra things like `NaturalTransformation` (see `scalaz.Tag.TagOf`), but you can get all that is needed to achieve what we get here.

src/main/scala/scalaz/stream/hash.scala
```diff
@@ -13,51 +15,84 @@ import process1._
...
   /**
    * Computes the SHA-512 hash of the input elements.
    * $MutableImpl
    */
-  val sha512: Process1[ByteVector,ByteVector] = messageDigest("SHA-512")
+  val sha512: Process1[ByteVector,ByteVector @@ SHA512] = messageDigest[SHA512]
 
-  private def messageDigest(algorithm: String): Process1[ByteVector,ByteVector] = {
-    def go(digest: MessageDigest): Process1[ByteVector,ByteVector] =
+  private def messageDigest[A: Digestable]: Process1[ByteVector,ByteVector @@ A] = {
```
fthomas: I made this private because `algorithm` could have been an arbitrary string and not only a valid algorithm name. Does it make sense to make this public now that you need to provide a `Digestable` instead of a `String`?

jedws: absolutely! I did indeed intend to make it public.

I'll make the `Digestible.named` locally private, need a total factory method instead.

djspiewak:
I'm fine with merging this as-is.  I know the type tag adds a slightly tighter Scalaz dependency, but it's not like we're *on the cusp* of cutting ties there.  As @jedws said, it's relatively easy to replicate the functionality if we really really want to (certainly easier than many of our other Scalaz dependency points).

fthomas:
> I'm fine with merging this as-is.

Me too. I just wanted to note that this makes the removal of the scalaz 
dependency a little bit harder.

fthomas:
@jedws One thing I forgot: it would be cool if you would add yourself to the [CONTRIBUTORS.md](https://github.com/scalaz/scalaz-stream/blob/master/CONTRIBUTORS.md) file :-)

jedws:
@fthomas sorry, missed that last comment, added

fthomas:
Cool, thanks @jedws. @ anyone with commit rights, can you please merge this commit.

djspiewak:
@fthomas @jedws Done

PR #371, opened 2015-04-20
Add `localAddress` method to `tcp` and `udp` modules

mpilquist:
Motivation is binding to port 0 and accessing the bound port number.


djspiewak:
:+1:

This is actually quite a bit more useful than `remoteAddress`, which is simply going to give an ephemeral port number.  Very nice indeed.

pchlupacek:
Thanks!

Issue #309, opened 2015-01-27
Process1 with stateful computation 


prange:
As far i can tell are there no combinators to perform a stateful computation where the output of a Process1 is computed by a state and an input. If i am not mistaken scan outputs the same state that is input to the "next" step, and zipWithState carries the original input and new state to the output.

I imagine either with some type carrying the state data (S):
```scala
stateful: ((S,A) => (S,B)) => S => Process1[A,B]
```

or with some Mealy machine
```scala
type Mealy[A,B] = (a:A)=>(Mealy[A,B],B)

mealy: Mealy[A,B] => Process1[A,B]
```

Am i mistaken, or is this easily achieved with the existing combinators?

djspiewak:
It's not easily achieved with the existing combinators.  I would absolutely die of happiness if we had some sort of `A => State[S, B] => S => Process1[A, B]` combinator.  I tried pretty hard to implement this [here](https://github.com/djspiewak/sparse/blob/master/src/main/scala/scalaz/stream/parsers/package.scala#L50), but the result isn't generalized.

fthomas:
I think you could write `mapStateful` in terms of `zipWithScan1` by squeezing the `S` and `O` types into `zipWithScan1`'s `B` type like this:
```scala
  def mapStateful2[S,I,O](s: S)(g: (S, I) => (S, O)): Process1[I,O] =
    zipWithScan1[I, (S, Option[O])]((s, None)) { case (a, (sn, _)) =>
      val (s2, o) = g(sn, a)
      (s2, Some(o))
    }.map(_._2._2).pipe(stripNone)
```
But this is neither easy nor pretty.

pchiusano:
I like the mapStateful2 signature and would accept a PR for that. Not sure
what to call it though... mapStateful is pretty weird. I think in the book
we called it `loop`, but I'm not a big fan of that name either..

fthomas:
The signature looks a bit like `Traversable`'s `mapAccum*`:
```haskell
mapAccumL :: Traversable t => (a -> b -> (a, c)) -> a -> t b -> (a, t c) 
```
but without emitting the accumulator. Should we generalize to
```scala
def mapAccum[S,I,O](s: S)(g: (S, I) => (S, O)): Process1[I, (S, O)]
```
?

prange:
+1 for mapAccum. 
[...]

PR #373, opened 2015-04-24
Implemented scan through State process1

djspiewak:
Adds a process1 with the following type signature:

```scala
def stateScan[S, A, B](init: S)(f: A => State[S, B]): Process1[A, B]
```

I've wanted this for a while, and it is in general much more usable than `scan` despite serving essentially the same purpose.  Decided to relax by scratching my own itch.

### DIFF
--- a/src/main/scala/scalaz/stream/process1.scala
+++ b/src/main/scala/scalaz/stream/process1.scala
@@ -1061,6 +1061,10 @@ private[stream] trait Process1Ops[+F[_],+O] {
   def scan[B](b: B)(f: (B,O) => B): Process[F,B] =
     this |> process1.scan(b)(f)
 
+  /** Alias for `this |> [[process1.stateScan]](init)(f)`. */
+  def stateScan[S, B](init: S)(f: O => State[S, B]): Process[F, B] =
+    this |> process1.stateScan(init)(f)
+
   /** Alias for `this |> [[process1.scanMap]](f)(M)`. */
   def scanMap[M](f: O => M)(implicit M: Monoid[M]): Process[F,M] =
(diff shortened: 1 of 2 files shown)


alissapajer:
+1

fthomas:
Could you explain why this fixes #309 ? The `mapAccumulate` we agreed on there looks a bit different.

djspiewak:
We can use mapAccumulate as well, but I felt justified in the "Fixed"
annotation since stateScan actually implements the type signature
originally requested. :-)

src/main/scala/scalaz/stream/process1.scala
```diff
@@ -540,6 +540,13 @@ object process1 {
       case \/-(a)   => emit(a)
     }
 
+  def stateScan[S, A, B](init: S)(f: A => State[S, B]): Process1[A, B] = {
+    await1[A] flatMap { a =>
```
fthomas: This can be simplified to `receive1 { a =>`.

djspiewak: I guess the type inference *would* work here.  I rather prefer the `await1 flatMap` style, but I don't mind either way.

fthomas:
Ok, cool. I like the signature of `mapAccumulate` and will submit a PR for it.

WRT to this PR: Do you mind adding an alias for `this pipe stateScan` to `Process1Ops`? We've been doing this so far for all `Process1`s for which an alias could be added.

djspiewak:
@fthomas I wouldn't mind an alias in the least, I just didn't add one because I didn't know the policy for them.  :-)

fthomas:
:+1:

Issue #249, opened 2014-10-01
Let Process1, Tee, Wye transducers receive all available inputs on await


pchiusano:
As [mentioned here](https://github.com/scalaz/scalaz-stream/pull/237#issuecomment-55305291), I'd like to introduce the following combinators:

```Scala
def available[A]: Process1[A,Seq[A]]
def availableL[A]: Tee[A,Any,Seq[A]]
def availableR[A]: Tee[Any,A,Seq[A]]
// can have receiveAvailable, receiveAvailableL/R, etc

def await1[A]: Process1[A,A] = available[A].flatMap(emitAll)
// likewise for awaitL/R/Both
```

This will require modifying the [`Env` type](https://github.com/scalaz/scalaz-stream/blob/master/src/main/scala/scalaz/stream/Process.scala#L1058) and will be a comprehensive change. (But it should be mostly if not entirely backwards compatible, since the old combinators will still exist)

djspiewak:
@pchiusano Interesting problem with your specification.  Specifically, `await1`.  The problem is that `await1` *guarantees* that it will produce a `Process1` containing at most *one* element.  Reading that another way, it promises to *consume* at most one element from the input process.  Defining this in terms of `available` breaks that guarantee, since you will consume as many elements as are in a chunk.  There are several combinators that this breaks, but `delete` is the first one in `Process1Spec`.

I still think that `available` is a really good idea, but it has to be done as a separate algebraic case.

djspiewak:
Working on this even further, I'm not sure that this is going to work.  It breaks too much compositionality due to the consumption guarantees.  Even having *both* `await1` and `available1` in the algebra as separate cases, just having `available1` in use messes up derivative combinators.  For example, `find` is defined in terms of `filter`.  It's very clear that `filter` should be defined in terms of `available1`, but we *can't* because of `find`:

```scala
  def find[I](f: I => Boolean): Process1[I, I] =
    filter(f).once
```

The problem is that, with `filter` defined using `available1`, the transducer resulting from `find` will over-consume the source, just as I mentioned would happen with `await1`.  I really think this points to a fundamental issue with this concept, since the consumption properties of a `Process1` are not represented in the types, and thus a seemly innocuous optimization change like swapping `id` to `available1` can *silently* break downstream composition.

I'll think about ways to resolve this problem, but it's tricky.  We can't exactly "put the genie back in the bottle" after we've consumed from the input source.

djspiewak:
[...]

PR #375, opened 2015-04-26
Recover lost data in Queue

djspiewak:
This fixes data lost by premature closure of the `dequeue` process, giving users a chance to run a fresh `dequeue` that recovers from the failure.  There *is* a tradeoff here though, and that tradeoff is that we may see data multiple times.  We can't really be absolutely sure that a premature closure of `dequeue` is an error, and so we over-approximate by just pushing the data back in regardless (we may be able to do better here once we address #288).  Given the way in which queues are generally used (i.e. often connected to external data sources, such as network sockets), I don't think this limitation is too much of a problem.  If you *really* need to ensure that you don't see repeated data from fresh dequeues, you should run an LRU deduplicating cache.

Note that this *also* slightly changes the semantics of multiple concurrent "dequeuers".  If you're dequeuing from multiple processes simultaneously, not only is it somewhat non-deterministic as to which process gets which value, but these processes may see data *out of order* relative to each other!  This happens if one of the two dequeuers ends prematurely.  The last batch of data seen by the ended dequeuer will get sent over to the other process, which may *in the meantime* have seen some data that was enqueued after the last batch from the failed dequeuer.  Thus, ordering constraints are violated, but *only* if you have multiple concurrent dequeues.  In other words, a case that already results in significant nondeterminism.

### DIFF
--- a/src/main/scala/scalaz/stream/async/mutable/Queue.scala
+++ b/src/main/scala/scalaz/stream/async/mutable/Queue.scala
@@ -109,7 +109,7 @@ trait Queue[A] {
 
 private[stream] object CircularBuffer {
   def apply[A](bound: Int)(implicit S: Strategy): Queue[A] =
-    Queue.mk(bound, (as, q) => if (as.size + q.size > bound) q.drop(as.size) else q)
+    Queue.mk(bound, false, (as, q) => if (as.size + q.size > bound) q.drop(as.size) else q)   // disable recovery for all circular buffers, for now
 }
 
 private[stream] object Queue {
@@ -122,13 +122,17 @@ private[stream] object Queue {
    * and then all enqueue processes will wait until dequeue.
    *
    * @param bound   Size of the bound. When <= 0 the queue is `unbounded`.
+   * @param recover Flag controlling automatic dequeue error recovery semantics.  When
+   * false (the default), data may be lost in the event of an error during dequeue.
+   * When true, data will never be lost on dequeue, but concurrent dequeue processes
+   * may see data out of order under certain error conditions.
    * @tparam A
    * @return
    */
-  def apply[A](bound: Int = 0)(implicit S: Strategy): Queue[A] =
-    mk(bound, (_, q) => q)
+  def apply[A](bound: Int = 0, recover: Boolean = false)(implicit S: Strategy): Queue[A] =
+    mk(bound, recover, (_, q) => q)
 
-  def mk[A](bound: Int,
+  def mk[A](bound: Int, recover: Boolean,
             beforeEnqueue: (Seq[A], Vector[A]) => Vector[A])(implicit S: Strategy): Queue[A] = {
     sealed trait M
     case class Enqueue(a: Seq[A], cb: Throwable \/ Unit => Unit) extends M
@@ -199,7 +203,10 @@ private[stream] object Queue {
           case l @ -\/(_) => cb(l)
 
           case \/-(a) => {
-            ref.lastBatch = Vector(a)
+            if (recover) {
+              ref.lastBatch = Vector(a)
+            }
+
             cb(\/-(a :: Nil))
           }
         }
@@ -212,7 +219,10 @@ private[stream] object Queue {
         else
           queued splitAt limit
 
-        ref.lastBatch = send
+        if (recover) {
+          ref.lastBatch = send
+        }
+
         S { cb(\/-(send)) }
 
         queued = remainder
@@ -271,19 +281,19 @@ private[stream] object Queue {
         case ConsumerDone(ref) => {
           consumers = consumers.filterNot(_._1 == ref)
 
-          val batch = ref.lastBatch
-          ref.lastBatch = Vector.empty[A]
+          if (recover) {
+            val batch = ref.lastBatch
+            ref.lastBatch = Vector.empty[A]
 
-          if (batch.nonEmpty) {
-            if (queued.isEmpty) {
-              enqueueOne(batch, Function.const(()))
-            } else {
-              queued = batch fast_++ queued     // put the lost data back into the queue, at the head
+            if (batch.nonEmpty) {
+              if (queued.isEmpty) {
+                enqueueOne(batch, Function.const(()))
+              } else {
+                queued = batch fast_++ queued     // put the lost data back into the queue, at the head
+                signalSize(queued.size)
+              }
             }
           }
-
-
-          signalSize(queued.size)
         }
 
       } else m match {
--- a/src/main/scala/scalaz/stream/async/mutable/Queue.scala
+++ b/src/main/scala/scalaz/stream/async/mutable/Queue.scala
@@ -113,6 +113,9 @@ private[stream] object CircularBuffer {
 }
 
 private[stream] object Queue {
+
+  import scalaz.stream.Util._
+
   /**
    * Builds a queue, potentially with `source` producing the streams that
    * will enqueue into queue. Up to `bound` number of `A` may enqueue into the queue,
@@ -135,7 +138,9 @@ private[stream] object Queue {
     case class ConsumerDone(ref: ConsumerRef) extends M
 
     // reference to identify differed subscribers
-    class ConsumerRef
+    class ConsumerRef {
+      var lastBatch = Vector.empty[A]
+    }
 
 
     //actually queued `A` are stored here
@@ -192,7 +197,11 @@ private[stream] object Queue {
       if (queued.isEmpty) {
         val cb2: Throwable \/ A => Unit = {
           case l @ -\/(_) => cb(l)
-          case \/-(a) => cb(\/-(a :: Nil))
+
+          case \/-(a) => {
+            ref.lastBatch = Vector(a)
+            cb(\/-(a :: Nil))
+          }
         }
 
         val entry: (ConsumerRef, Throwable \/ A => Unit)  = (ref -> cb2)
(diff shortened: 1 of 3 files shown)


djspiewak:
@pchlupacek I don't mind making this behavior more configurable.  However, I do not believe that your example demonstrates a valid counter to the implementation.

> In this case instead of expected guaranteed order (increasing) as the sub1 will receive likely the 2 after 6 breaking the ordering guarantee of queue.

You're *concurrently* dequeueing from multiple sources simultaneously.  We already *do not* support multiple concurrent dequeuers, since there are no guarantees about which dequeue will receive which piece of data!  Additionally, the very notion of doing this concurrently, even *if* we supported deterministic multi-dequeue behavior smashes the idea of ordering right out of the gate.  Let's narrow this down:

```scala
val p = merge.mergeN(Process(q.dequeue, q.dequeue))
```

Even without my change, what expectations do you have about the order of elements in `p`?  Are *any* expectations reasonable?

The point being that the breakage of ordering constraints only occurs in cases where users are *already* using unsupported semantics.  I'm not going to lose any sleep over that.

The bigger problem with my implementation, I think, is the fact that it's technically possible to observe the same data *repeatedly* even without having concurrent dequeues.  I detailed this in the OP.  That strikes me as a serious semantic flaw, but also one that falls in line with the sorts of guarantees people expect with asynchronous queues attached to remote data sources (a very very common use case).  If we make this recovery behavior configurable in the two `Queue` constructors, we can give users control over this flaw.

> I think this sort of recovery when the tee fails while left side was consumed must be somehow injected to tee and not to queue itself, as this change have side effect on use cases w/o tee. And in fact to me this is more generic problem not just affecting queue.

In a sense, it is.  In fact, it's exactly the same problem that dooms #249: once we "take" data from a process, we cannot "give it back".  There is no mechanism in the algebra to signal to a process that we took "too much" and we want that data reinjected, nor does the notion of "reinjection" even make sense for most expressible processes.  There is no way for `tee` to generically solve this recovery problem precisely because we have no such mechanism, nor would any such mechanism make sense.

So while I would absolutely love to solve this problem in a general way, inside of `tee`, we really can't.

pchlupacek:
@djspiewak think of mergeN only as concurrent runner. My expectations are that elements in any consumer appear as they are inserted to the queue. I don't want any magic to happen behind the scenes that will reinsert elements out of order. My example was just to demonstrate the problem.

Furthermore I am not really understanding what do you mean by the fact that we are not supporting multiple concurrent dequeuers. Has anything changed in code? We using this perfectly fine in our code on latest release. 

I think we perhaps cannot solve this in tee. Although I would give it a thought still. But it is solvable by external transaction signalling if really needed.

One idea I had was to signal transaction back to queue. Once we had that junction type, and perhaps we can say that Queue will have commit functionality and unless item is being committed it won't be considered dequeued (queue always wait for this commit). Perhaps, this can as well solve the interruption issue we had discuss elsewhere? What do you think about overhead of this solution would it be significant?  We can maybe introduce something like 

 ```scala

trait Queue[A] {
   type TxCommit[A]: Process[Task,Nothing]

   def dequeueTransactionally: TSource[(A,TxCommit)]

}

```

What do you think?

**Comment by djspiewak** [comment]
> think of mergeN only as concurrent runner. My expectations are that elements in any consumer appear as they are inserted to the queue. I don't want any magic to happen behind the scenes that will reinsert elements out of order. My example was just to demonstrate the problem.

My point is that `mergeN`, or *anything* that would supply the requisite concurrency needed to see the ordering constraint violation, is already rearranging ordering *by definition*.  It doesn't matter if you use `merge`, or `mergeN`, or `gather` on the resulting `Task` instances.  You must be running concurrent dequeues in order to see the ordering violation, and as soon as you're running things concurrently, you have given up on the precise order of elements coming out of the queue.

Elements are not being reinserted out of order.  If you look at the code, the only case in which elements *can* be out of order is if the failing consumer drains the last elements out of the queue and then fails.  In the meantime (after the drain but before the failure), a *concurrent* consumer observes new data in the queue and *fully drains* those elements.  When the first consumer fails, it will try to reinsert the lost batch in the appropriate location, but since the queue is empty it can't do that and it just treats the lost batch as fresh data, in turn notifying the concurrent consumer of the "old" elements out of order.

Under normal circumstances, the reordering never even happens.  Under circumstances without concurrent consumers, the reordering never happens.

> Furthermore I am not really understanding what do you mean by the fact that we are not supporting multiple concurrent dequeuers. Has anything changed in code? We using this perfectly fine in our code on latest release.

Nothing has changed, but the code doesn't support concurrent dequeues anymore than it ever has.  When data is enqueued, only one of the dequeuers will see that particular piece of data, and the dequeuer which sees it is, in principle, non-deterministic.  I understand that there are some cases where this is useful, but if you're in a case where non-determinism of that magnitude is acceptable, I really don't understand why a minor reordering would be problematic.

> One idea I had was to signal transaction back to queue. Once we had that junction type, and perhaps we can say that Queue will have commit functionality and unless item is being committed it won't be considered dequeued (queue always wait for this commit). Perhaps, this can as well solve the interruption issue we had discuss elsewhere? What do you think about overhead of this solution would it be significant?

A transactional solution would be strictly more general than what I implemented.  In a sense, what I implemented is a very basic, one-step transaction on `Queue` inferred by the termination status.  With a committer, we would be able to fail further into the processing pipeline, still without loss of data.

The problem with this solution is that it takes all of the flaws that you see in my current solution and *magnifies* them.  Think about the concurrent consumer case again.  Explicit transaction boundaries don't eliminate the reordering problem when you want concurrent consumers.  In fact, they make it worse because they allow you to stretch the transaction boundary arbitrarily far into the consumer processing pipeline, giving concurrent consumers more time to see more data out of order.

The only way to prevent out-of-order observations with concurrent consumers is to *block* all concurrent consumers during transactions, effectively treating each transaction as a lock-the-world semantic.  So in other words, the only way to prevent out-of-order observations with concurrent consumers is to *turn off concurrency*.

The problem is not the implementation in this PR.  The problem is the fact that we allow non-deterministic concurrent consumers *and* we're trying to preserve ordering constraints without data loss.  Pick two of those three, because we *cannot* have all of those features at the same time.  In master, we pick the first two and allow data loss.  In this PR, we pick the first and the third and relax ordering constraints.  We *could* just as easily preserve ordering and avoid data loss, but we would need to completely disallow concurrent consumers (or at least, disallow them during transaction windows).

**Comment by djspiewak** [comment]
@pchlupacek Any further thoughts on this?  To be clear, we have a release that we're working on pushing into production which exhibits data loss due to queue errors *right now*, and so this is a high priority for us.  If this (or some equivalently-powerful solution) is not merged in the near future, I'll need to maintain a full fork for a while until we can figure out what to do.

I understand that there are drawbacks, but I have addressed each drawback individually.  If those points are unsatisfactory, I'd like to know *why* so that I can either address the counterpoint or fix the issue.

**Comment by pchlupacek** [comment]
I am still not convinced this is the approach we shall take @djspiewak. If you would add this, I am afraid we will have problems on our side. 

I really really don't think this kind of logic has to be hardcoded in the queue. 

I was thinking about it and perhaps we can modify to to use not zip, but instead use zipRecover that will be defined as : 

```scala

def zipRecover[L,R]:TeeW[L,L,R, (L,R)] = {
  /* implemented like current zip, but left is implemented as awaitOr emitting eventually the memoized L value on Left as last **/
   ???
}

val queue: Queue[L] = ???
def toRecoverably[L](p2:Sink[Task,L]):Writer[L,Unit] = {
  /** like current to, just a bit of type alignment **/
  ???
}

queue.dequeue.toRecoverably(sink).drainO(queue.enqueue)

```

I think essentially this is the same as your hardcoded variant you proposing with sam guarantees, just written externally with current combinators. 








```

djspiewak:
@pchlupacek 

> I am still not convinced this is the approach we shall take @djspiewak. If you would add this, I am afraid we will have problems on our side.

Can you be more specific?

> I really really don't think this kind of logic has to be hardcoded in the queue.

I don't have any objection to making it configurable.  However, it most definitely has to be *inside* of `Queue`, as I will illustrate in a moment.

> I was thinking about it and perhaps we can modify to to use not zip, but instead use zipRecover that will be defined as

Could you give the full definition of `zipRecover` and show how it avoids the error cases?  I can very easily define a version of `zipWith` which reads from the right first (or just flip the operands, for that matter), meaning that the dequeue action gets sequenced *after* any potential error is checked from the other side, but I think that's as good as we can do with `tee`.

> `queue.dequeue.toRecoverably(sink).drainW(queue.enqueue)`

This violates ordering *directly*, even in non-concurrent cases.  My solution only violates ordering constraints in the case of concurrent dequeues that completely and simultaneously drain the queue.  This solution re-enqueues elements in the wrong order *every time*.  I'm also not convinced that an error stemming from `sink` would not kill the `drainW` term before it could re-enqueue everything.

pchlupacek:
@djspiewak lets make this configurable, if you think this is only the way for now. We can improve it laters. If this blocks your project lets go this way, and we can find generic way laters ok?

djspiewak:
@pchlupacek Sounds good!  I'll make the default to have no recovery, to avoid silently breaking compatibility.  If we identify unavoidable issues with the implementation, we can move aggressively toward alternatives.

pchlupacek:
@djspiewak cool thanks.

djspiewak:
@pchlupacek `boundedQueue` and `unboundedQueue` now take an optional boolean which toggles the error recovery semantics.  `circularBuffer` currently does not allow recovery at all, though I suspect it would work just fine.

PR #376, opened 2015-04-26
update scalaz 7.1.1

xuwei-k:


djspiewak:
:+1:

In somewhat related news, I would love if we could make the scalaz 7.1 branch the mainline, with `master-a` devoted to 7.0.  The problem is that we would be forever married to a *lot* of `@uncheckedVariance` all over the place, and that certainly isn't the happiest state of affairs.

Issue #369, opened 2015-04-18
Remove Terminated


djspiewak:
People should not be throwing `Terminated(End)` to halt repetitions.  This particular semantic quirk is what makes the following two (seemly-identical) terms very distinct:

```scala
lazy val repeat1: Process[F, A] = await(t) { e => emit(e) ++ sound }

lazy val repeat2: Process[F, A] = await(t) { e => emit(e) } ++ sound
```

Oh hey, that's exactly the transformation that `pipe` and `tee` *assume* is sound!  What a coincidence…

In the above, `repeat2` will infinitely loop if the termination condition had previously been `t` throwing `Terminated(End)`, whereas `repeat1` will halt as expected with `End`.  I would humbly propose that this is insane, and we shouldn't be generating semantic quirks like this.

`Terminated` currently accepts any `Cause`, including `End`.  I would propose that, instead of this, we force `Terminated` to take only `EarlyCause`.  This makes the semantics a LOT cleaner (at compile time!) and forces users to explicitly write `onHalt` handlers to transmute thrown failures of this sort into normal stream completion (e.g. `wye`, `Queue` and `Topic` would all need such handlers).

tl;dr: Throwing `Terminated(End)` is deprecated behavior anyway, and retaining support for it is pushing us into highly unfortunate and non-composable semantic corners.  It should be removed.

djspiewak:
Here's an example of a (failing) test which demonstrates how `Terminated(End)` breaks compositionality:

```scala
  property("identity piping preserve eval termination semantics") = secure {
    implicit val teq = Equal.equalA[Throwable]

    val halt = eval(Task delay { throw Terminated(End) }).repeat ++ emit(())

    ((halt pipe process1.id).runLog timed 3000 map { _.toList }).attempt.run === (halt.runLog timed 3000 map { _.toList }).attempt.run
  }
```

pchlupacek:
@djspiewak `Terminated(End)` was just introduced to support signalling from the sink evaluation, namelly, io.*, that previsouly was implemented with just `throw(End)` to signal exhaust of the input source (essentially signal End). 

We anyhow need to touch all the `io` stuff with #333  change to assure proper release of the resources. However I didn't thought how to solve this. I am very much in favor of removing `Terminated` alltogether, if we ave good replacement.

pchiusano:
I'm actually in favor of just getting rid of Terminated altogether. It is
never necessary - just use Option / Either and explicit control flow.

[...]

Issue #351, opened 2015-03-29
Transducers to generalize Process1


djspiewak:
```scala
  sealed trait AcceptT[+F[_], -I, +O]

  object AcceptT {
    trait Partial[F[_], I] {
      type Aux[α] = AcceptT[F, I, α]
    }

    final case class Effect[+F[_], O](f: F[O]) extends AcceptT[F, Any, O]
    final case class Accept[+F[_], I]() extends AcceptT[F, I, I]
  }

  type Transducer[+F[_], -I, +O] = Process[AcceptT.Partial[F, I]#Aux, O]
  // type Process1[-I, +O] = Transducer[Nothing, I, O]
  // type Channel[+F[_], -I, +O] = Obsolete!!
  // type Sink[+F[_], +A] = Transducer[F, A, Unit]

  object transducer {

    def receiveF[F[_], A, I, O](effect: F[A])(f: A => Transducer[F, I, O]): Transducer[F, I, O] =
      await[AcceptT.Partial[F, I]#Aux, A, O](AcceptT.Effect(effect))(f)

    def receive1[F[_], I, O](f: I => Transducer[F, I, O]): Transducer[F, I, O] =
      await[AcceptT.Partial[F, I]#Aux, I, O](AcceptT.Accept[F, I]())(f)
  }

  implicit class TransducerSyntax[F[_], A](val self: Process[F, A]) extends AnyVal {
    import Cause.{EarlyCause, End, Kill}

    def transduce[B](trans: Transducer[F, A, B])(implicit F: Applicative[F], C: Catchable[F]): Process[F, B] = trans.suspendStep flatMap {
      // fail suck ugliness to get around limitations in GADT pattern matching
      case Halt(rsn) => self.kill onHalt { _ => Halt(rsn) }

      case step: Step[AcceptT.Partial[F, A]#Aux, B] => {
        val cont1 = step.next

        step.head match {
          case awt: Await[AcceptT.Partial[F, A]#Aux, a, B] => {
            val req = awt.req
            val rcv = awt.rcv

            req match {
              case AcceptT.Effect(eff) => {
                val fp = eff.attempt map { _ leftMap { Error(_) } } map { res =>
                  self transduce (cont1.+:[AcceptT.Partial[F, A]#Aux, B](rcv(res).run))    // TODO errors?  I don't have access to Util.Try, because... blergh
                }

                eval(fp) flatMap { a => a }
              }

              case AcceptT.Accept() => self.step match {
[...]

PR #384, opened 2015-05-13
Removed auto-magical Terminated coercion in eval/repeatEval

djspiewak:
This is the compromise solution to #369.  `Terminated` is still here, and so is `asHalt`, but `eval` no longer magically swallows `Terminated` without warning anyone.  As a general rule, if you need `Terminated` (as some things do, due to the half-crippled nature of `Channel`), you need to add the `onHalt { _.asHalt }` bit explicitly.

The giant hanging chad in all of this is the fact that I had to change some code to explicitly contain this `onHalt` behavior even within scalaz-stream.  In general, this was confined to fairly low-level stuff (like the implementation of `wye`), but not always.  In particular, the `through` combinator got the `asHalt` treatment, since it is very common to implement effectful `Channel` termination via `Terminated`, and in fact under many circumstances there are no other options (e.g. pretty much every `Sink` ever).  We'll be able to remove this once we have a good solution for #351, but for now… bleh.

### DIFF
--- a/src/main/scala/scalaz/stream/wye.scala
+++ b/src/main/scala/scalaz/stream/wye.scala
@@ -836,9 +836,7 @@ object wye {
         }
       })(S)
 
-      repeatEval(Task.async[Seq[O]] { cb => a ! Get(cb) })
-      .flatMap(emitAll)
-      .onComplete(eval_(Task.async[Unit](cb => a ! DownDone(cb))))
+      repeatEval(Task.async[Seq[O]] { cb => a ! Get(cb) }) onHalt { _.asHalt } flatMap emitAll onComplete eval_(Task.async[Unit](cb => a ! DownDone(cb)))
     }
 }
 
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -983,7 +983,7 @@ object Process extends ProcessInstances {
   implicit class ProcessSyntax[F[_],O](val self: Process[F,O]) extends AnyVal {
     /** Feed this `Process` through the given effectful `Channel`. */
     def through[F2[x]>:F[x],O2](f: Channel[F2,O,O2]): Process[F2,O2] =
-        self.zipWith(f)((o,f) => f(o)).eval
+        self.zipWith(f)((o,f) => f(o)).eval onHalt { _.asHalt }     // very gross; I don't like this, but not sure what to do
 
     /**
      * Feed this `Process` through the given effectful `Channel`, signaling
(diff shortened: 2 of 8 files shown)


src/main/scala/scalaz/stream/wye.scala
```diff
@@ -836,9 +836,7 @@ object wye {
         }
       })(S)
 
-      repeatEval(Task.async[Seq[O]] { cb => a ! Get(cb) })
-      .flatMap(emitAll)
-      .onComplete(eval_(Task.async[Unit](cb => a ! DownDone(cb))))
+      repeatEval(Task.async[Seq[O]] { cb => a ! Get(cb) }) onHalt { _.asHalt } flatMap emitAll onComplete eval_(Task.async[Unit](cb => a ! DownDone(cb)))
```
pchlupacek: @djspiewak this is more about code style, but Its hard for me to read superlong scala lines, and furthemore infix syntax in scala is not that deterministics as the haskell one, so I would really be more happy with dot notation if possible?

djspiewak: There are several reasons *not* to use dot notation, most notably that it rules out using curly braces, which in turn rules out semicolons and generates inconsistencies in newline inference.  I would certainly be willing to split the line up, because it really is too long as you point out, but I feel very very strongly about the demerits of dot notation in Scala.  Infix notation is quite consistent in Scala, and in fact it is one of the only ways to generate a method invocation and block syntax which is consistent across all constructs.

pchlupacek: @djspiewak generally, I am not really sure if I am following your arguments, but really having that long lines make reasoning about the code very hard. 

However, I think we have do adopt one code style and keep it through the library, consistent. If we would change code style to your proposal, I have no problem with that, but I would like to keep line with at reasonable level and all new PRs to this library shall follow it, as currently, we use dot notation almost everywhere. 

I think this is something we need collective agreement on what do you think? 

Just for curiosity whats wrong with writing your code as below?

```scala

repeatEval(Task.async[Seq[O]] ( cb => a ! Get(cb) )) 
.onHalt(_.asHalt) 
.flatMap(emitAll)
.onComplete(eval_(Task.async[Unit](cb => a ! DownDone(cb))))

```

djspiewak: Having long lines is bad.  No argument there.  The only real downside of infix notation is that it requires some inversion of arguments (noun-verb) in order to break lines, which is deeply unfortunate.

I don't have too much of a problem choosing one consistent code style, though I make absolutely no secret that I have a very hard time reading certain Scala styles, and I don't doubt that writing them would be even more difficult.  The existing code style within scalaz-stream, even in places where a code block is written by a single author is inconsistent to say the least.  I'm very much in favor of being consistent about things, but my impression was that consistency in the style wasn't much cared for.  (that's not a slight, some developers just don't care about stylistic consistency, and that's fine)  It was for that reason that I made no attempt to retain any consistency with existing code.

> Just for curiosity whats wrong with writing your code as below?

All of those lines are indented equally, making it difficult to read them as part of the same invocation.  The spacing on the parentheses is inconsistent even within the same construct (see: the lambdas on the first and last lines).  The use of parentheses for lambdas *necessitates* inconsistency with lambdas that have variables or statements in them, and thus as a general rule, *no* internally-consistent code style allows parenthetical lambdas.  The corollary to the parenthetical lambda issue derives from the inconsistent association of dots relative to curly braces vs parentheses (using a dot dispatch on a curly brace block often, though not always, associates to the result of the block rather than to the result of the invocation taking the block as a parameter).  Thus, using braces for blocks (which again is *required* for an internally-consistent style) categorically rules out using dot dispatch ubiquitously, since dots cannot be consistently used in the presence of curly braces.

Issue #389, opened 2015-05-20
signalOf() doesn't "Set" as far as compareAndSet() is concerned


psnively:
Constructing a Signal with signalOf(), then using compareAndSet() reveals that the "old" value is None:

```scala
> console
[warn] Credentials file /Users/psnively/.bintray/.credentials does not exist
[info] Starting scala interpreter...
[info] 
import scalaz.stream._
Welcome to Scala version 2.11.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_31).
Type in expressions to have them evaluated.
Type :help for more information.

scala> import async._
import async._

scala> val s = signalOf[Int](5)
s: scalaz.stream.async.mutable.Signal[Int] = scalaz.stream.async.mutable.Signal$$anon$1@6231f63f

scala> s.compareAndSet(x => { println(s"Old value = $x"); Some(10) })
res0: scalaz.concurrent.Task[Option[Int]] = scalaz.concurrent.Task@1bb740cc

scala> res0.run
Old value = None
res1: Option[Int] = Some(10)
```

.get() works, and compareAndSet() works if you .set() first.

djspiewak:
Well that's not weird or anything…

djspiewak:
I think that rather than initializing the `Signal` with an `emitAll` followed by the `writerSignal`, we need to just take the `writerSignal` and use `process1.feed` to give it a `Set`.  This will ensure that the internal state is managed appropriately.

psnively:
:+1:

psnively:
So how do we get closure on this? Unfortunately, an important project is blocked on it. So I'm really asking "how do we fix, confirm, and release this?"

djspiewak:
@psnively The best first step would be to shoot us a PR fixing the issue.  I'll probably get to it this weekend-ish, but it sounds like it's a higher priority for you.  Unless I'm missing something, the fix should be almost exactly as straightforward as my description of it above.

Making releases is…a challenge.  Really only @pchiusano can do it.  More importantly, we've already broken binary compatibility with 0.7, so this would need to be a full 0.8 release unless we took your fix and backported it to the 0.7 tags (which should be trivial, since we haven't touched `Signal` all that much).  Either way, Paul would need to make a release.

[...]

PR #390, opened 2015-05-22
Maintain internal writer state machine upon initial signal value

djspiewak:
This addresses #389.  The problem was that the `Writer1` inside of `Signal` would never set up its state machine with the initial value.  This PR fixes that by setting up the initial value via `feed1` rather than by emitting it over to `WriterTopic`.


pchlupacek:
excellent. thanks. must have been sleeping when wrote that :-)

PR #398, opened 2015-05-30
Revised runLast to no longer consume linear memory

djspiewak:
Through the strong power of typeclass incoherence!  Surprise, guys, but there's plenty of `Monoid`s in the sea…


pchlupacek:
Was my thought too, that this is only runLast issue, but I hope it is not, as master (before the bracket pr lands) show a poor performance in more spaces. I am in process to confirming this, we'll see.

djspiewak:
@pchlupacek I wouldn't be surprised if there were issues in other places as well.  Regardless though, `runLast` in its prior incarnation was clearly a problem.  Hopefully this helps.

PR #403, opened 2015-06-09
Add uncons function to Process

jedesah:
@runarorama and I worked on this for a use case within [Remotely](https://github.com/oncue/remotely). I have now found a use for it within another project, so I felt it was time to polish it up and submit a PR so that everyone could benefit from this function.

I also created a method on `Await` that encapsulates a common pattern within the implementation to "evaluate" an Await into an `F` of a new `Process` that is not awaiting anymore (has some elements to `Emit`). As you can imagine, many methods need to "await" some elements.

### DIFF
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -457,7 +457,7 @@ sealed trait Process[+F[_], +O]
   final def unconsOption[F2[x] >: F[x], O2 >: O](implicit F: Monad[F2], C: Catchable[F2]): F2[Option[(O2, Process[F2, O2])]] = step match {
     case Step(head, next) => head match {
       case Emit(as) => as.headOption.map(x => F.point[Option[(O2, Process[F2, O2])]](Some((x, Process.emitAll[O2](as drop 1) +: next)))) getOrElse
-          (Process.empty +: next).unconsOption
+          next.continue.unconsOption
       case await: Await[F2, _, O2] => await.evaluate.flatMap(p => (p +: next).unconsOption(F,C))
     }
     case Halt(cause) => cause match {
--- a/src/main/scala/scalaz/stream/Process.scala
+++ b/src/main/scala/scalaz/stream/Process.scala
@@ -456,11 +456,7 @@ sealed trait Process[+F[_], +O]
       case Emit(as) => as.headOption.map(x =>
         F.point[(O2, Process[F2,O2])]((x, Process.emitAll[O2](as drop 1) +: next))) getOrElse
         C.fail[(O2, Process[F2,O2])](new scala.NoSuchElementException)
-      case Await(req, k, preempt) => for {
-        e <- C.attempt(req)
-        as <- F.point(k(EarlyCause.fromTaskResult(e)).run)
-        u <- (as +: next).uncons(F, C)
-      } yield u
+      case await: Await[F2, _, O2] => await.evaluate.flatMap(p => (p +: next).uncons(F,C))
     }
     case Halt(cause) => cause match {
       case End => C.fail(new scala.NoSuchElementException)
@@ -489,10 +485,7 @@ sealed trait Process[+F[_], +O]
                 go(cont.continue.asInstanceOf[Process[F2,O]], nacc)
               }
             case (awt:Await[F2,Any,O]@unchecked, cont) =>
-              F.bind(C.attempt(awt.req)) { r =>
-                go((Try(awt.rcv(EarlyCause.fromTaskResult(r)).run) +: cont).asInstanceOf[Process[F2,O]]
-                  , acc)
-              }
+              awt.evaluate.flatMap(p => go(p +: cont, acc))
           }
         case Halt(End) => F.point(acc)
         case Halt(Kill) => F.point(acc)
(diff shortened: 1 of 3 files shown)


src/main/scala/scalaz/stream/Process.scala
```diff
@@ -451,8 +451,18 @@ sealed trait Process[+F[_], +O]
 
   }
 
-
-
+  final def uncons[F2[x] >: F[x], O2 >: O](implicit F: Monad[F2], C: Catchable[F2]): F2[(O2, Process[F2, O2])] = step match {
+    case Step(head, next) => head match {
+      case Emit(as) => as.headOption.map(x =>
+        F.point[(O2, Process[F2,O2])]((x, Process.emitAll[O2](as drop 1) +: next))) getOrElse
+        C.fail[(O2, Process[F2,O2])](new scala.NoSuchElementException)
```
pchlupacek: I have not strong feeling about failing `F` here instead of having uncons signature to return option. Can you please elaborate why this has such technique rather?

jedesah: @pchlupacek That's just the way @runarorama and I wrote it originally for Remotely. I don't have any strong feelings either way. Unless anyone objects, I will gladly make the change.

pchlupacek: ok how about define 

```scala

 final def unconsOption[F2[x] >: F[x], O2 >: O](implicit F: Monad[F2], C: Catchable[F2]): F2[(Option[O2], Process[F2, O2])] = ???

```

and then define `uncons` in term of `unconsOption`?

jedesah: @pchlupacek That sounds good to me. Will do the change right now.

jedesah: @pchlupacek Done! I opted for `F2[Option[(O2, Process[F2, O2])]]` instead of the exact signature you suggested because I don't think it is relevant to return the same process in the event that it has ended for wtv reason.

I also opted to fail the `F` if the `Process` has failed.

What do you think?

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -451,8 +451,19 @@ sealed trait Process[+F[_], +O]
 
   }
 
+  final def uncons[F2[x] >: F[x], O2 >: O](implicit F: Monad[F2], C: Catchable[F2]): F2[(O2, Process[F2, O2])] =
+    unconsOption(F, C).map(_.getOrElse(throw new NoSuchElementException))
```
djspiewak: I would rather this be `flatMap(_.map(_.point).getOrElse(C.fail(new NoSuchElementException)))`, rather than relying on magical exception catching behavior in certain implementations.

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -451,8 +451,19 @@ sealed trait Process[+F[_], +O]
...
 
-
+  final def unconsOption[F2[x] >: F[x], O2 >: O](implicit F: Monad[F2], C: Catchable[F2]): F2[Option[(O2, Process[F2, O2])]] = step match {
+    case Step(head, next) => head match {
+      case Emit(as) => F.point(as.headOption.map(x => ((x, Process.emitAll[O2](as drop 1) +: next))))
```
djspiewak: This is incorrect, as the logic assumes that `as` will only be empty if the stream is at an end!  That is not the case though, as `step` is free to produce an `Emit(Nil)` at any time.

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -451,8 +451,19 @@ sealed trait Process[+F[_], +O]
...
+      case await: Await[F2, _, O2] => await.evaluate.flatMap(p => (p +: next).unconsOption(F,C))
+    }
+    case Halt(cause) => cause match {
+      case End => F.point(None)
```
djspiewak: This should probably be `case End | Kill`, since it is perfectly legitimate for a process to get killed early.

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -613,6 +621,11 @@ object Process extends ProcessInstances {
      */
     def extend[F2[x] >: F[x], O2](f: Process[F, O] => Process[F2, O2]): Await[F2, A, O2] =
       Await[F2, A, O2](req, r => Trampoline.suspend(rcv(r)).map(f), preempt)
+
+    def evaluate[F2[x] >: F[x], O2 >: O](implicit F: Monad[F2], C: Catchable[F2]): F2[Process[F2,O2]] =
```
djspiewak: Noice

src/main/scala/scalaz/stream/Process.scala
jedesah: I can't disagree with that.

src/main/scala/scalaz/stream/Process.scala
jedesah: Nice catch! I think I have fixed it.

src/main/scala/scalaz/stream/Process.scala
jedesah: makes sense.

djspiewak:
:+1:

jedesah:
@djspiewak Thanks for mentioning `next.continue`, I was unaware that existed.

pchlupacek:
Thanks @jedesah !

Issue #399, opened 2015-05-31
Create benchmarks


pchlupacek:
We have to find a way how to reliably prove the performance after various changes stay in expected limit, or better improves. So far we have limited tests with so-so reliability. If anyone has experience / ideas and want to step in, please comment below.

wookietreiber:
I just created a PR to show you how this might look like using [ScalaMeter](http://scalameter.github.io/).

pchiusano:
I am looking for volunteers to take this on. There is already a benchmark project, I would just use the [sbt-jmh](https://github.com/ktoso/sbt-jmh) (or some other reputable JVM benchmarking tool) plugin to write some benchmarks there. If they can be run as part of Travis, even better.

Regression testing for performance is a separate issue, let's get some benchmarks written first.

I wrote some performance 'tests' in [`StreamPerformanceSpec`](https://github.com/functional-streams-for-scala/fs2/blob/topic/redesign/core/src/test/scala/fs2/StreamPerformanceSpec.scala). These might be a good starting point for some benchmarks.

mpilquist:
Fixed in #646

PR #406, opened 2015-06-21
skeleton for regression testing

wookietreiber:
- adresses #399
- uses [scalameter regression testing][1]
- runs in sbt integration test target, i.e. `it:test`
- generates an offline html report in the tmp subdir of the main project
  dir, there is an index.html you can open to inspect the results, the
  regression history is also saved in that tmp subdir
- might need a few tweaks in terms of stability, confidence intervals, etc.
- I included just a simple dummy example to give you an idea of how it
  looks like, but adding new stuff to test should be straightforward
- I am not sure whether `Process.constant(1).take(n)` is the fastest way
  to create a reasonably sized Process

[1]: http://scalameter.github.io/home/gettingstarted/0.7/regressions/index.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ tags
 *#
 z_local.sbt
 .DS_Store
+tmp/
--- a/build.sbt
+++ b/build.sbt
@@ -75,3 +75,19 @@ initialCommands := "import scalaz.stream._"
 doctestWithDependencies := false
 
 doctestSettings
+
+// -------------------------------------------------------------------------------------------------
+// integration tests / regression testing
+// -------------------------------------------------------------------------------------------------
+
+configs(IntegrationTest)
+
+Defaults.itSettings
+
+libraryDependencies += "com.storm-enroute" %% "scalameter" % "0.6" % "it"
+
+logBuffered in IntegrationTest := false
+
+testFrameworks += new TestFramework("org.scalameter.ScalaMeterFramework")
+
+parallelExecution in IntegrationTest := false
--- a/src/it/scala/ProcessRegressionTest.scala
+++ b/src/it/scala/ProcessRegressionTest.scala
@@ -1,0 +1,22 @@
+package scalaz.stream
+
+import org.scalameter.api._
+
+import scalaz.concurrent.Task
+
+class ProcessRegressionTest extends PerformanceTest.OfflineRegressionReport {
+  val sizes = Gen.range("size")(10000, 50000, 20000)
+  val ones: Process[Task,Int] = Process.constant(1)
+  val streams = for (size <- sizes) yield (ones.take(size))
+
+  performance of "Process" in {
+    measure method "run" in {
+      using(streams) config (
+        exec.independentSamples -> 1
+      ) in { stream =>
+        val u = stream.run.run
+        u
+      }
+    }
+  }
+}


pchlupacek:
@wookietreiber Excellent. We need to agree how the perf test will be added to project. 

Initially the framework you proposing - what dependencies it does have if any ?

@djspiewak, @pchiusano  what are your thoughts?

wookietreiber:
The dependencies for version 0.6 I used in this PR can be found [here](https://github.com/scalameter/scalameter/blob/version/0.6/project/Build.scala#L71).

pchlupacek:
@wookietreiber We had so far been very restrictive regarding the dependecies the scalaz stream depends on. Your pr adds a bunch of dependencies that even just for testing may be conflicting with some people usages. 

we will see @djspiewak  and @pchiusano opinion on this, but by any chance can we reduce the amount of dependencies or eventually just move this perhaps to separate module?

djspiewak:
@pchlupacek @wookietreiber My thoughts are that this looks extremely good!  I like the idea of making it an integration test.  I believe this also neatly resolves the dependency issue, since those dependencies are only required by the `it` scope and will not leak into the main build.  You won't even know they exist if you never run `it:test`.

I like the framework though, assuming that it's handling proper warm-up and statistical variance for us.

wookietreiber:
Yes, these dependencies are only required by the `it` scope, i.e. main build still clean:

```
sbt @ scalaz-stream $ dependencyTree
[info] org.scalaz.stream:scalaz-stream_2.11:a33811c93f971e182face32c77c1386881da00bf [S]
[info]   +-org.scalaz:scalaz-concurrent_2.11:7.1.2 [S]
[info]   | +-org.scalaz:scalaz-core_2.11:7.1.2 [S]
[info]   | | +-org.scala-lang.modules:scala-parser-combinators_2.11:1.0.4 [S]
[info]   | | +-org.scala-lang.modules:scala-xml_2.11:1.0.4 [S]
[info]   | | 
[info]   | +-org.scalaz:scalaz-effect_2.11:7.1.2 [S]
[info]   |   +-org.scalaz:scalaz-core_2.11:7.1.2 [S]
[info]   |     +-org.scala-lang.modules:scala-parser-combinators_2.11:1.0.4 [S]
[info]   |     +-org.scala-lang.modules:scala-xml_2.11:1.0.4 [S]
[info]   |     
[info]   +-org.scalaz:scalaz-core_2.11:7.1.2 [S]
[info]   | +-org.scala-lang.modules:scala-parser-combinators_2.11:1.0.4 [S]
[info]   | +-org.scala-lang.modules:scala-xml_2.11:1.0.4 [S]
[info]   | 
[info]   +-org.scodec:scodec-bits_2.11:1.0.6 [S]
[info]   
[success] Total time: 0 s, completed Jun 22, 2015 7:11:10 PM
```

 However, `sbt update` pulls in the `it` scope dependencies. I don't know if thats an issue.

wookietreiber:
About ScalaMeter as the testing framework:

1.  Warming up forked JVMs is done by default. The `Warmer` can be configured.
2.  Statistics / variance / outlier detection is all handled and can be configured.

There is some documentation dedicated to regression testing and the options you set for these tests [here](http://scalameter.github.io/home/gettingstarted/0.7/regressions/index.html). Search this page for `exec.` for some of the options, though a complete read of that page surely won't hurt.

I didn't do much of that configuration yet in the PR, since the test I included is more or less a dummy. Once there are real tests the proper configuration for each test needs to be figured out and set up. This depends e.g. on GC cycles that occur during the test, as I figure, the main source of runtime variance. Thus, depending on object allocation, tests must be configured differently, e.g. compare `foreach` on `Array` with `groupBy` on `List` (see the examples at the abovementioned page).

djspiewak:
I'm :+1: to merge this.  @pchlupacek?

djspiewak:
Bump?

pchlupacek:
@djspiewak ah sure. Sorry.

pchlupacek:
@djspiewak @wookietreiber seems we are missing the 2.12 version of the library. Cna you perhaps resolve this? Can we disable these tests if the 2.12 is not available only to target 2.10/2.11?

wookietreiber:
I don't know, can sbt disable `it` depending on scala version? I know you can add dependencies depending on scala version, e.g.:

```scala
  def Specs2(scalaVersion: String) = CrossVersion.partialVersion(scalaVersion) match {
    case Some((2,11)) => List("org.specs2" %% "specs2-core" % "3.6.2")
    case Some((2,10)) => List("org.specs2" %% "specs2-core" % "3.6.0")
    case _            => Nil
  }
```

Or: ping @axel22 are you willing to publish a 2.12 SNAPSHOT of scalameter?

axel22:
Yes, we should have 2.12-SNAPSHOT version anyway.
I would have to modify the build file and the Travis build - it would probably be near the end of this week.

djspiewak:
@wookietreiber I'm completely behind the times…  Just found out about the 2.12 incompatibility in master.  I've reverted the merge, so this PR is once again outstanding.  I tried to adjust our build to disable the integration tests on 2.12 but failed.  

@axel22 Is there any update on getting a 2.12 build of scalameter?  If we can update this PR to build successfully, that would be awesome.

axel22:
@djspiewak To my shame, I kept postponing the 2.12 build for three weeks now. Sorry about that - I will make sure it happens tonight (for real this time :) ).

axel22:
@wookietreiber It looks like `scala-chart` for 2.12 has not yet been released. I could:
- disable scala-chart altogether
- try depending on 2.11 (although that could break badly)

I have the same problem with `jackson-module-scala`.
I will do the second solution for now, as it seems to pass tests - `scalameter` and `scalameter-core` will be published for Scala `2.12` starting from the `0.8-SNAPSHOT` version.

axel22:
@djspiewak @wookietreiber `scalameter_2.12.0-M1-0.8-SNAPSHOT` should be available now - https://travis-ci.org/scalameter/scalameter/jobs/77858216

.gitignore
```diff
@@ -11,3 +11,4 @@ tags
 *#
 z_local.sbt
 .DS_Store
+tmp/
```
axel22: Just FYI - the default folder for ScalaMeter test results is `target/benchmarks` since version `0.7`.

wookietreiber:
@axel22 I just published scala-chart 0.5.0. It is on sonatype but for some reason unknown to me it just does not get synched to maven central ... waiting for hours for the sync now ...

axel22:
Thanks - let me know when it's synced, and I can update ScalaMeter `0.8-SNAPSHOT` to depend on that.

wookietreiber:
It is still not synced. I'm guessing something went wrong. Any idea how I can get what went wrong out of sonatype?

axel22:
So you called `publish-signed` from SBT, and then logged into Sonatype, and closed and published the artifact from their UI, but it's not synchronized with Maven?

If so, I could probably still depend on the Sonatype copy. And if it's not syncing to maven, I guess the best way is to contact their support.

wookietreiber:
It's up on maven now. Sorry, for spamming so much about this on this issue.

djspiewak:
@wookietreiber Do you think you can open a new PR against master which contains this work against 0.8-SNAPSHOT (or whatever is relevant for 2.12)?  It's a bit hard to track this work in a closed PR.  :-)

wookietreiber:
Done. I hope we can resolve the missing dependencies soon. Otherwise, the new PR should be good to go.

PR #407, opened 2015-06-22
- toInputStream: do not fail on Emit steps encountered after a process is killed

lukiano:
- toInputStream: do not wrap IOException into another IOException

### DIFF
--- a/src/main/scala/scalaz/stream/io.scala
+++ b/src/main/scala/scalaz/stream/io.scala
@@ -300,12 +300,15 @@ object io {
           case Halt(End | Kill) =>
             chunks = Nil
 
+          case Halt(Cause.Error(e: IOException)) => throw e
           case Halt(Cause.Error(e: Exception)) => throw new IOException(e)
 
           // rethrow halting errors
           case Halt(Cause.Error(e)) => throw e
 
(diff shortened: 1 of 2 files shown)


src/main/scala/scalaz/stream/io.scala
```diff
@@ -300,12 +300,15 @@ object io {
...
+          case Halt(Cause.Error(e: IOException)) => throw e
           case Halt(Cause.Error(e: Exception)) => throw new IOException(e)
 
           // rethrow halting errors
           case Halt(Cause.Error(e)) => throw e
 
-          case Step(Emit(_), _) => assert(false)    // this is impossible, according to the types
+          case Step(Emit(_), cont) =>
+            cur = cont.continue
+            close()
```
djspiewak: Hmm.  I bet that it's an `Emit(Nil)` or something similar.  It is literally impossible to get any *actual* values here (without casting), since the value type of the process in question is `Nothing`, but it is certainly possible to get nullary emits out of `step`.  This code is more resilient anyway.

djspiewak:
@lukiano Could you address the build failures on 2.10.5?  Otherwise, :+1:

lukiano:
Done, now the build succeeds fine

PR #408, opened 2015-06-25
Add adapt function to Sink

jedesah:
I currently have this as an extension method in [Remotely](https://github.com/jedesah/remotely/blob/feature/streaming_take2/core/src/test/scala/StreamingSpec.scala#L48) but I saw a place in [http4s_demo](https://github.com/http4s/http4s_demo/blob/master/src/main/scala/org/http4s/example/Routes.scala#L43) where it could be used so I figured I would create a PR against scalaz-stream itself.


pchlupacek:
I think we have already that combinator and its named `contramap` and can be found on ChannelSyntax, that means it applies to Sink as well.

jedesah:
@pchlupacek Oh, okay. Forgive my ignorance.

pchlupacek:
@jedesah ah no problem at all and thanks for your effort!

PR #416, opened 2015-07-18
Add a constructor to Process for using iterators

shawjef3:
I find myself using my own IteratorToProcess method quite often, so I think probably an Process constructor from iterators belongs in scalaz-stream. It can be dangerous since iterators aren't re-usable, but I hope the warning I added in the documentation is enough to put people on the right path.

### DIFF
--- a/src/main/scala/scalaz/stream/io.scala
+++ b/src/main/scala/scalaz/stream/io.scala
@@ -92,7 +92,7 @@ object io {
 
   /**
    * Creates a `Process[Task,String]` from the lines of a file, using
-   * the `resource` combinator to ensure the file is closed
+   * the `iteratorR` combinator to ensure the file is closed
    * when processing the stream of lines is finished.
    */
   def linesR(filename: String)(implicit codec: Codec): Process[Task,String] =
@@ -100,7 +100,7 @@ object io {
 
   /**
    * Creates a `Process[Task,String]` from the lines of the `InputStream`,
-   * using the `resource` combinator to ensure the `InputStream` is closed
+   * using the `iteratorR` combinator to ensure the `InputStream` is closed
    * when processing the stream of lines is finished.
    */
   def linesR(in: => InputStream)(implicit codec: Codec): Process[Task,String] =
@@ -108,11 +108,11 @@ object io {
 
   /**
    * Creates a `Process[Task,String]` from the lines of the `Source`,
-   * using the `iterator` combinator to ensure the `Source` is closed
+   * using the `iteratorR` combinator to ensure the `Source` is closed
    * when processing the stream of lines is finished.
    */
   def linesR(src: => Source): Process[Task,String] = {
-    iterateR(Task.delay(src))(src => Task.delay(src.close()))(_.getLines())
+    iteratorR(Task.delay(src))(src => Task.delay(src.close()))(r => Task.delay(r.getLines()))
   }
 
   /**
@@ -154,10 +154,10 @@ object io {
   /**
    * Create a Process from an iterator. The value behind the iterator should be
    * immutable and not rely on an external resource. If that is not the case, use
-   * `io.iterateR`.
+   * `io.iteratorR`.
    */
-  def iterate[O](i: => Iterator[O]): Process[Task, O] = {
-    await(Task.delay(i)) { iterator =>
+  def iterator[O](i: Task[Iterator[O]]): Process[Task, O] = {
+    await(i) { iterator =>
       val hasNext = Task delay { iterator.hasNext }
       val next = Task delay { iterator.next() }
 
@@ -178,10 +178,10 @@ object io {
    * @tparam O is a value in the iterator
    * @return
    */
-  def iterateR[R, O](req: => Task[R])(
+  def iteratorR[R, O](req: Task[R])(
                      release: R => Task[Unit])(
-                     mkIterator: R => Iterator[O]): Process[Task, O] = {
-    bracket[Task, R, O](req)(r => Process.eval_(release(r)))(r => iterate(mkIterator(r)) )
+                     mkIterator: R => Task[Iterator[O]]): Process[Task, O] = {
+    bracket[Task, R, O](req)(r => Process.eval_(release(r)))(r => iterator(mkIterator(r)) )
   }
 
   /**
--- a/src/main/scala/scalaz/stream/io.scala
+++ b/src/main/scala/scalaz/stream/io.scala
@@ -154,7 +154,7 @@ object io {
   /**
    * Create a Process from an iterator. The value behind the iterator should be
    * immutable and not rely on an external resource. If that is not the case, use
-   * `io.iterateR'`.
+   * `io.iterateR`.
    */
   def iterate[O](i: => Iterator[O]): Process[Task, O] = {
     await(Task.delay(i)) { iterator =>
@@ -174,13 +174,13 @@ object io {
    * @param req acquires the resource
    * @param release releases the resource
    * @param mkIterator creates the iterator from the resource
-   * @tparam R is the type of the resource
-   * @tparam O is the type of the values in the iterator
+   * @tparam R is the resource
+   * @tparam O is a value in the iterator
    * @return
    */
   def iterateR[R, O](req: => Task[R])(
-                    release: R => Task[Unit])(
-                    mkIterator: R => Iterator[O]): Process[Task, O] = {
+                     release: R => Task[Unit])(
+                     mkIterator: R => Iterator[O]): Process[Task, O] = {
     bracket[Task, R, O](req)(r => Process.eval_(release(r)))(r => iterate(mkIterator(r)) )
   }
 
--- a/src/test/scala/scalaz/stream/IteratorSpec.scala
+++ b/src/test/scala/scalaz/stream/IteratorSpec.scala
@@ -11,15 +11,16 @@ import scalaz.concurrent.Task
 class IteratorSpec extends Properties("iterators") {
 
   property("io.iterate completes immediately from an empty iterator") = secure {
-    io.iterate[Int](Iterator.empty).runLog.run.isEmpty
+    io.iterator[Int](Task.now(Iterator.empty)).runLog.run.isEmpty
   }
 
   property("io.iterate uses all its values and completes") = forAll { (ints: Vector[Int]) =>
-    io.iterate[Int](ints.toIterator).runLog.run == ints
+    io.iterator[Int](Task(ints.toIterator)).runLog.run == ints
   }
 
   property("io.iterate is re-usable") = forAll { (ints: Vector[Int]) =>
-    io.iterate(ints.toIterator).runLog.run == io.iterate(ints.toIterator).runLog.run
+    val intsIterator = Task(ints.toIterator)
+    io.iterator(intsIterator).runLog.run == io.iterator(intsIterator).runLog.run
   }
 
   case class IteratorResource[T](items: T*) {
@@ -48,7 +49,7 @@ class IteratorSpec extends Properties("iterators") {
 
 
 
-    io.iterateR(acquire)(release)(_.iterator).run.run
+    io.iteratorR(acquire)(release)(r => Task(r.iterator)).run.run
 
     resource.exists(_.isReleased)
   }


src/main/scala/scalaz/stream/Process.scala
```diff
@@ -871,6 +871,24 @@ object Process extends ProcessInstances {
...
+   * Create an iterator to use as a source for a `Process`,
+   * which lazily emits the values of the iterator, then halts.
+   *
+   * Be sure that iteratorCreator creates any mutable or external
+   * resources that the iterator uses are created as part of the
+   * Task. Also, do not use Task.now to create the iterator.
+   */
+  def iterator[F[_], O](iteratorCreator: F[Iterator[O]]): Process[F, O] = {
+    //This design was based on unfold.
+    def go(iterator: Iterator[O]): Process0[O] = {
```
djspiewak: Maybe this is just me being a purist, but given that both `hasNext` and `next` are effectful, it seems to me that the following would be a preferable implementation:

```scala
await(iteratorCreator) { iterator =>
  val hasNext = Task delay { iterator.hasNext }
  val next = Task delay { iterator.next() }

  await(hasNext) { hn => if (hn) eval(next) else halt } repeat
}
```

The downside here is that we're restricting the effect to be `Task`, but I'm not sure that's necessarily a bad thing.  If someone instantiates `F` to be an eager effect, things are going to be highly screwy.  Also, and perhaps more importantly, evaluation of processes which do *not* contain `Await` nodes is *eager* in `stepAsync`, and so the iterator is going to get pulled much faster than you might expect.

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -871,6 +871,24 @@ object Process extends ProcessInstances {
...
 
+  /**
+   * Create an iterator to use as a source for a `Process`,
+   * which lazily emits the values of the iterator, then halts.
+   *
+   * Be sure that iteratorCreator creates any mutable or external
+   * resources that the iterator uses are created as part of the
+   * Task. Also, do not use Task.now to create the iterator.
+   */
+  def iterator[F[_], O](iteratorCreator: F[Iterator[O]]): Process[F, O] = {
```
djspiewak: I like the effectful `iteratorCreator` concept.  I initially looked at this signature with some suspicion, but now I believe it is the correct way to go (modulo my concerns below).

djspiewak:
I'm a little concerned about the fact that we allow non-`Task` effects here, and I'm very concerned about the `Await`-less implementation of `go`.  Beyond that though, I think this would be a good addition to the library.  @pchiusano, any thoughts?

fiadliel:
Is it worthwhile having a resource cleanup task? e.g. https://github.com/scalaz/scalaz-stream/issues/323 style iterators could leak resources otherwise. It does make the types a bit more complex, something like:

```scala
def iterator[O](iteratorCreator: Task[(Iterator[O], Task[Unit])]): Process[Task, O] = {
  ...

  Process.await(iteratorCreator) {
    case (iterator, cleanup) =>
      go(iterator).onComplete(Process.eval_(cleanup))
    }
  }
}
```

shawjef3:
@fiadliel I like your idea of the iterator-with-cleanup.

Oddly, the real world scenarios I've used the iterator constructor don't fit it, and I'd end up using a dummy cleanup Task. The reason is that in one case, there is nothing to clean up, as the iterator is backed by an array that's not available to me in the value's interface, and in the other, the iterator shares resources with iterators in other Processes, and so onComplete must be called later when they're all done.

Maybe there could be two constructors with differing documentation and taking different signatures. One would have the clean-up and one wouldn't.

shawjef3:
@djspiewak You might like this version of the iterator constructor I came up with first, that put hasNext and next in the same Task.

Process.repeatEval(Task(if (iterator.hasNext) iterator.next() else throw Cause.Terminated(Cause.End)))

pchlupacek:
I think the iterator is for me the kind of resource, like any other, specifically if we wrap it to `Task`. The only true differentiator here is that we know how we will step it. Other than that `bracket` or perhaps `io.resource` shall be sufficient. 

Maybe add it as complimentary to `io.resource`, name it `io.iterateResource` and define it via `io.resource`?

shawjef3:
@djspiewak @pchlupacek Here's a hybrid of your two ideas. For pchlupacek it lets people (optionally) wrap the iterator in io.resource, because it's produced in a Process, and it has djspiewak's implemention of go.

I couldn't decide between using flatMap or merge.mergeN, so I decided to leave that up to the client.

```scala
def iterator[O](iteratorCreator: Process[Task, Iterator[O]]): Process[Task, Process[Task, O]] = {
  def go(iterator: Iterator[O]): Process[Task, O] = {
    val hasNext = Task delay { iterator.hasNext }
    val next = Task delay { iterator.next() }

    await(hasNext) { hn => if (hn) eval(next) else halt } repeat
  }

  iteratorCreator.map(go)
}
```

What do y'all think?

fiadliel:
I don't think you can emit the Iterator itself, and still have resource safety (if you look at `io.resource`, you would use it to emit values from the iterator). The resources used by the iterator could be cleaned up while you still had a reference to it.

I like that idea of having a similar signature to `io.resource`, except the user doesn't need to provide the step function -- it's a lot simpler than what I suggested earlier.

shawjef3:
@fiadliel I have some tests using your version of go, but it repeats hasNext rather than hasNext and then next, so it never halts. I'm working to try and fix it. You can see the tests [here](https://github.com/shawjef3/scalaz-stream/blob/iterator-F/src/test/scala/scalaz/stream/IteratorSpec.scala) and the implementation [here](https://github.com/shawjef3/scalaz-stream/blob/iterator-F/src/main/scala/scalaz/stream/Process.scala#L874).

shawjef3:
Looking more, it does call the second argument of await, but the halt is not effective.

shawjef3:
I see, it's because repeat runs the process over and over again, even if it halts.

src/main/scala/scalaz/stream/Process.scala
```diff
@@ -871,6 +871,31 @@ object Process extends ProcessInstances {
   def iterateEval[F[_], A](start: A)(f: A => F[A]): Process[F, A] =
     emit(start) ++ await(f(start))(iterateEval(_)(f))
 
+  /**
+   * Create a Process from an iterator. This should not be used directly,
+   * because iterators are mutable.
+   */
+  private [stream] def iteratorGo[O](iterator: Iterator[O]): Process[Task, O] = {
```
djspiewak: Having this here and used by functions in `io` makes me feel squicky, especially since `io.iterator` and `Process.iterator` are so similar.  Not sure what to do about it though.  I was playing around with eliminating this function and just defining `Process.iterator` in terms of `io.iterator`, but I think it makes things ugly.

src/main/scala/scalaz/stream/io.scala
```diff
@@ -155,6 +155,33 @@ object io {
...
+   * Create a Process from an iterator that requires some external or
+   * other mutable resource, while ensuring that the resource is released.
+   *
+   * Use `iterators` if the resource is associated with multiple iterators.
+   */
+  def iterator[R, O](acquire: Task[(R, Iterator[O])])(
+                     release: R => Task[Unit]): Process[Task, O] = {
+    bracket(acquire)(r => eval_(release(r._1))){
+      r => iteratorGo(r._2)
+    } onHalt { _.asHalt }
```
djspiewak: What is the reasoning behind having the `asHalt` conversion here?  As a general rule, `asHalt` is a hacky hack hack and should be avoided unless ABSOLUTELY necessary.

djspiewak:
Still would really love to hear from @pchiusano on this.

shawjef3:
@djspiewak I was basically copying `bracket` with a modification so that `release` is only called on the first value in the tuple returned by `acquire`. I don't know enough about the internals of scala-stream to say whether `asHalt` should be used here or in `bracket`.

src/main/scala/scalaz/stream/io.scala
shawjef3: I was basically copying `bracket` with a modification so that `release` is only called on the first value in the tuple returned by `acquire`. I don't know enough about the internals of scala-stream to say whether `asHalt` should be used here or in `bracket`.

pchlupacek:
@shawjef3 @djspiewak Not sure what @pchiusano will say, but I don't have good feeling about this.

I really don't think that `Task[(R, Iterator[O])]` is perfect signature and smells to be like baking two unrelated parts into once. I think, imho the Resource, and Iterator are not always so tightly related.


The better signature imho is : 
```scala
//in io
def iterate[O](acquire: Task[R])(iterator:R => Task[Iterator[O]])(release:R => Task[Unit]) : Process[Task,O] = ??? // defined via io.resource. 

```
However I still see this only usefull when there is 1:1 relation between R and Iterator, which is probably in most cases

and the other signature `def iterator[O](iteratorCreator: Task[Iterator[O]]): Process[Task, O]` is imho very very undescriptive and requires other coding to be used correctly (i.e. resource safety).

shawjef3:
@pchlupacek I like your signature for io.iterate, so I will work on changing the pull request. However, I disagree that the non-io iterator constructor requires other coding to be used correctly. Basically any immutable structure that has an `iterator()` method would be awkward to use with the io.iterator pattern.

shawjef3:
@pchlupacek I changed io.iterator to use your suggested signature, but I found that io.iterator can't use io.resource directly without using an exception to close the Process, because io.resource uses repeatEval on its step function.

Also, I found an opportunity to use io.iterator as an implementation for io.linesR.

shawjef3:
I published a project that is essentially this pull request. https://github.com/shawjef3/scalaz-stream-iterator

djspiewak:
@pchlupacek This has been outstanding for too long.  Things look good to me, and I'd like to merge this weekend unless you have any objections.  Veto deadline is tomorrow.  If you feel that this PR needs more changes, please comment.  Otherwise, it's goin' in!

pchlupacek:
@djspiewak essentially I have no issues with `io.iterator`.

However, 

```scala
//Process.scala
def iterator[O](iteratorCreator: Task[Iterator[O]]): Process[Task, O] = ???
```
Hence we do not allow for any effects, I think this shall have pure signature, that means w/o Task. this essentially shall be IMHO the iterateGo implementation. So I would just drop `Process.iterator`, and have instead `Process.iterate()` that has `Process.iterateGo` signature, and is public. 

If anyone really wants the the task signature, on iterator w/o resources,  one can do `eval(Task.delay(iterator)).flatMap(Process.iterate)` 

Similarly I do not see need for `io.iterators`, hence if we have that `Process.iterate` that imho is well defined as 
```scala
   val source: Process[Task, Iterator[O]] = ???
   source.map { it => Process.iterate(it) onComplete (release(it)) } : Process[Task, Process[Task,O]]

```
Sorry to complicate this again ...

shawjef3:
@pchlupacek taking a look at integrating your suggestions.

shawjef3:
@pchlupacek I think I see your point. It'd be better to have the iterator constructor entirely in the io object, since all iterators are side-effecting. Working on it.

shawjef3:
@djspiewak I'm confident @pchlupacek will like my latest push.

shawjef3:
Would it be better if iterator's signature lost the Task and the iterator was passed in by name?
```iterator[O](i: => Iterator[O]): Process[Task, O]```

I can see this being bad if the iterator is created eagerly already, but then so can a Task using Task.now, it's just a little more effort.

**Comment by shawjef3** [comment]
I pushed a branch showing the results of using by-name rather than Task. I can merge it into the pull request if you guys like it better.

[iterator](https://github.com/shawjef3/scalaz-stream/blob/iterator-by-name/src/main/scala/scalaz/stream/io.scala#L163)

[linesR](https://github.com/shawjef3/scalaz-stream/blob/iterator-by-name/src/main/scala/scalaz/stream/io.scala#L114)

[iterator tests](https://github.com/shawjef3/scalaz-stream/blob/iterator-by-name/src/test/scala/scalaz/stream/IteratorSpec.scala)

**Comment by pchlupacek** [comment]
@shawjef3 I think `iterator` may be strict.  If you really need lazy evaluation there you can wrap it to `Task.delay`.  I really think what we looking for is : 

 ```scala

//in Process.scala

/**
* transforms your Iterator to Process. Note that this implementation allows for side effects in iterator
  that means, if your iterator internally uses side effects, this implementation will convert evaluation 
  of `next` and `hasNext` to Tasks, and stop iterating whenever the `hasNext` evaluates to false.

  If you need acquire iterator from some resources that needs to be released when iterator finishes its iteration, feel free to use `io.iterator` instead.
*/
def iterate[O](it:Iterator[O]):Process[Task,O] = ???


//in io.scala

/** Resource safe constructor for iterator.  
**/
def iterator(acquire:Task[Iterator[O])(release:Tak[Unit]):Process[Task,O] = ???

```

shawjef3:
@pchlupacek I have to disagree with those signatures.

### iterate
Iterate taking a bare iterator not wrapped in a task or by name will make multiple running processes using that Process interfere with each other.

The following will print "hi" only once, rather than twice.
```scala
val i = Seq("hi").iterator()
val p = iterate(i).to(stdOutLines)
p.run
p.run
```

### iterator

I'm no longer convinced we need a separate constructor for iterators that rely on a resource. The reason is that bracket already exists, and so a client can just use bracket with an iterator constructor, as I did with linesR [here](https://github.com/shawjef3/scalaz-stream/blob/iterator-F/src/main/scala/scalaz/stream/io.scala#L114).

pchlupacek:
@shawjef3 I see your point. Lets then leave iterator parameter lazy, as you have suggested.  I still believe the io.iterator is nice syntax sugar, and would add it to library if it won't make too much pain. 

Thanks you for you patience a lot !

shawjef3:
Should the iterator be lazy because it's passed by name, or because it's wrapped in a Task?

pchlupacek:
@shawjef3 I think Task is too powerfull and inidicates here non deterministic nature plus specializes iterate to Task, whereas we only need to solve the mutable nature of iterator, so I think just lazy parameter is enough.

shawjef3:
OK, I completed the changes.

pchlupacek:
@shawjef3 thanks. @djspiewak  can you make the last look on this and then LGTM?

djspiewak:
@pchlupacek Sorry for the delay.  I've been insanely busy.  I'm really, *really* not happy with dropping `Task` from the signature.  In scalaz-stream, `Task` represents effects, not just nondeterminism.  What you've done is you've taken an effect (computing a new `Iterator`) and moved it from the explicit control of a type constructor (`Task`) into a completely bare thunk.  There is nothing in the type signature of the function which documents how restarting is handled, or that it is handled at all!  Also, it's something that is very easy for users to get wrong, whereas explicitly constructing a `Task` is at the very least a signal to the user that they need to think about the effectfulness of what they're doing.

I hate to be the new roadblock on this, so if everyone else is very firmly against `Task[Iterator[A]]` then I'll relent, but honestly I think the new signature is a significant step backward.

pchlupacek:
@djspiewak I know what you mean. I see the `iterate` as low level primitive, that is used on several places. I wanted not to specify in only to `Task`, and add it to any `F` sounded too generic to me. I agree with the restarting issue. 

I still believe thant thung is ok, and we use this technique on several places in library. However if you really ensist on Task. No problem with that alltogether.

fthomas:
Just an aside: there is already an `iterate` function in the `Process` companion. It could be confusing to have `Process.iterate` and `io.iterate`.

pchlupacek:
ah @fthomas. thanks for noting that.  I'll think about better name then.

shawjef3:
How about the original name I had for it, iterator?

shawjef3:
@djspiewak For the signature, there are two different methods, and so to be sure we're doing the right thing, we should considering them separately.

There is the constructor from an iterator, currently named iterate.
There is the constructor from an iterator that relies on a resource, currently named iterateR.

Is iterate more like linesR, that takes a thunk, or more like bufferedChannel, that takes a Task?
Is iterateR more like linesR, that takes a thunk, or more like bufferedChannel, that takes a Task?

Without any principled notion about why things are the way they are, it looks to me that methods that have external resources use Task, whereas methods that handle their resources internally (like linesR(=> Source) use thunks.

So I'm thinking that iterateR should use Task, and iterate should use a thunk.

shawjef3:
@djspiewak Never mind. iterateR already uses Task, so you must have been talking about iterate.

shawjef3:
Bah! iterateR only uses Task for the resource. Sorry for the confusion. But I still think you were talking about making iterate use Task.

shawjef3:
@djspiewak Please consider the latest commit to the pull request.

pchlupacek:
ok I am fine with this. LGTM?

PR #417, opened 2015-07-19
Update scalacheck to 1.12.4, evalulate properties lazily.

fiadliel:
Scalacheck 1.12.4 instantiates an instance of test classes for
each property to be evaluated. Using `protect` instead of `secure`
makes the property evaluation lazy, to avoid doing a lot of extra
work here.


fiadliel:
I think because each Prop is now passed to SBT separately, the earlier ordering expectation for setup/go/teardown in `tcp.chat` no longer holds.

djspiewak:
The `tcp` support is very experimental and generally non-functional anyway, so I don't have a problem turning off one of its tests for now.  I'll take a shot at that…  Thank you so much for this work!

PR #425, opened 2015-08-02
Fix resource safety for `wye`, fixes #424

### LINKED ISSUE #424: `pipe(id)` breaks resource safety
Here is a minimal code example. It prints "acquire", but doesn't print "release". If we remove `.pipe(process1.id[Int])` it prints "release".

```scala
import scalaz.concurrent.Task
import scalaz.stream._

def disposableThing: Process[Task, Unit] = {
  val acquire = Task.delay {println("acquire")}
  val release = Task.delay {println("release")}

  io.resource(acquire)(x => release)(Task.now)
}

Process.emitAll(false :: true :: Nil).wye(
  for {
    thing <- disposableThing
    y <- Process.emit(1).zip(Process.emit(2).pipe(process1.id[Int]))
  } yield y
)(wye.interrupt).run.run
```

kanterov:
This pull request fixes #424. @djspiewak could you please take a look?

The original issue has been solved. I've added a spec covering it.

But... sometimes this spec fails. It doesn't fail if we add `Thread.sleep(100)` after `.run.run` and constantly fails if we add `Thread.sleep(1000)` to `release` task. It looks like `.run.run` doesn't wait for termination of `release` task. I feel it is another issue.

I don't know how to fix this. Probably if changes to `wye` are good, we can merge it without a spec and fix it later. Any suggestions?

### DIFF
--- a/src/main/scala/scalaz/stream/wye.scala
+++ b/src/main/scala/scalaz/stream/wye.scala
@@ -674,7 +674,7 @@ object wye {
             Either3.middle3((_: Cause) => ()) //rest the interrupt so it won't get interrupted again
 
           case Right3(cont) =>
-            (Halt(Kill) +: cont) stepAsync { _ => a ! Ready[A](side, -\/(Kill)) }
+            (Halt(Kill) +: cont) stepAsync { res => a ! Ready[A](side, res) }
             Either3.middle3((_: Cause) => ()) // no-op cleanup can't be interrupted
 
           case left@Left3(_) =>
--- a/src/test/scala/scalaz/stream/ResourceSafetySpec.scala
+++ b/src/test/scala/scalaz/stream/ResourceSafetySpec.scala
@@ -1,13 +1,14 @@
 package scalaz.stream
 
+import java.util.concurrent.atomic.AtomicInteger
+
 import Cause._
 import org.scalacheck._
 import Prop._
 
 import scalaz.concurrent.Task
-import java.util.concurrent.{ConcurrentLinkedQueue, LinkedBlockingDeque}
+import java.util.concurrent.{CountDownLatch, TimeUnit, Semaphore, ConcurrentLinkedQueue}
 import Process._
-import scalaz.-\/
 import scalaz.\/._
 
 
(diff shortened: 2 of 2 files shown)


djspiewak:
@kanterov I pushed a more deterministic workaround to [djspiewak/wye-tee-pipe-resource-safety-master](https://github.com/djspiewak/scalaz-stream/tree/wye-tee-pipe-resource-safety-master).  The "not running all release tasks before shutdown" is by design.  The reason is that some release continuations may actually take an unbounded amount of time.  It's not difficult to change the semantics of `stepAsync` to wait until all tasks are cleaned up, but doing so will actually cause tests to fail!

Your fix, btw, makes a lot of sense.  Under normal circumstances, we just assume that no one is ever going to swallow `Kill`, but the fusers can and do need to do this under certain circumstances.  This is why the interaction of `pipe`, `tee` and `wye` is required to reproduce the issue.  Nice catch!

kanterov:
@djspiewak thanks. I've replaced `SyncVar` with `Semaphore` because we need to track cases where acquire was `n` times, and release was called `n-1` times.

djspiewak:
@kanterov Since we're using `wye.interrupt`, is there actually any case where we `acquire` more than once?

kanterov:
@djspiewak You are right, it happens at most once. I was thinking about original code that caused this issue where it happened multiple times.

As far as I can see, semaphore says "for every acquire release happened exactly one time" while sync var says "if at least one acquire hapenned then at least one release happened". None of this options say "nothing or at least one acquire and release happened". I don't have a strong opinion. Do you have any concerns regarding leaving semaphore?

djspiewak:
@kanterov Conventionally throughout the code base, we use `SyncVar` for this sort of thing.  I'm fine with using `Semaphore` though, since it's already in-place and it does more directly reflect the "acquire/release metaphor" of what you're testing.

I would say :+1: for merge on this, unless anyone else has objections or you want to make more changes?

kanterov:
I'm not hundred percent confident, but it probably worth to change property name to "acquired iff released".

As a whole, I'm quite happy and looking forward to merging this to master.

pchlupacek:
@djspiewak one of the possible solutions here is to introduce your suggestion about Kill have an parameter. Then, at any time parameter will indicate that kill shall not be swallowed the pipe/tee et all will just propagate it down (i.e. swallow kill will ignore it)

other than that LGTM.

djspiewak:
Sorry for how this basically fell off my radar.  I've been sick for a week, so I sort of lost track of things.  Merged now!

PR #426, opened 2015-08-04
Add MiMa compability plugin to build

fiadliel:
I don't think there's a nice way of checking which branch we're on, and selecting the appropriate version on that basis (for example, we only want to check against 0.7.2a on the `series/0.7a` branch).

Instead, it should be easy to override behaviour per-branch.

Set `checkABICompatibilityVersion` to define which version to check.
Set `ignoredABIProblems` to define what ABI issues to ignore.
Check with `sbt mimaReportBinaryIssues`

This cherry-picks cleanly to `series/0.7a`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -101,6 +101,25 @@ OsgiKeys.importPackage := Seq(
   "*"
 )
 
+val ignoredABIProblems = {
+  import com.typesafe.tools.mima.core._
+  import com.typesafe.tools.mima.core.ProblemFilters._
+
+  Seq()
+}
+
+lazy val mimaSettings = {
+  import com.typesafe.tools.mima.plugin.MimaKeys.{binaryIssueFilters, previousArtifact}
+  import com.typesafe.tools.mima.plugin.MimaPlugin.mimaDefaultSettings
+
+  mimaDefaultSettings ++ Seq(
+    previousArtifact := MiMa.targetVersion(version.value).map(organization.value %% name.value % _),
+    binaryIssueFilters ++= ignoredABIProblems
+  )
+}
+
+mimaSettings
+
 parallelExecution in Test := false
 
 autoAPIMappings := true
--- a/project/MiMa.scala
+++ b/project/MiMa.scala
@@ -1,0 +1,57 @@
+case class LibraryVersion(major: Int, minor: Int, patch: Option[Int], suffix: Option[String]) {
+  override def toString = {
+    s"$major.$minor" +
+      patch.fold("")("." + _) +
+      suffix.fold("")("-" + _)
+  }
+}
+
+object MiMa {
+  // Matches versions like "0.8", "0.8.1", "0.8.1-SNAPSHOT", "0.8.1-3-53de4b5d"
+  def extractVersion(version: String): Option[LibraryVersion] = {
+    val VersionExtractor = """(\d+)\.(\d+)(?:\.(\d+))?(?:-(.*))?""".r
+    version match {
+      case VersionExtractor(major, minor, patch, suffix) =>
+        Some(LibraryVersion(major.toInt, minor.toInt, Option(patch).map(_.toInt), Option(suffix)))
+      case _ => None
+    }
+  }
+
+  // RC and SNAPSHOT suffixes represent pre-release builds,
+  // cannot check compatibility against unreleased software.
+  def suffixAfterRelease(version: LibraryVersion): Boolean = {
+    version.suffix match {
+      case Some(s) if s.startsWith("RC") => false
+      case Some(s) if s == "SNAPSHOT" => false
+      case None => false
+      case _ => true
+    }
+  }
+
+  // Return version of library to check for compatibility
+  def targetLibraryVersion(currentVersion: LibraryVersion): Option[LibraryVersion] = {
+    // Default target version has patch 0, no suffix.
+    // e.g. "0.8.1-SNAPSHOT" is compared with "0.8.0"
+    val targetVersion = currentVersion.copy(patch = Some(0), suffix = None)
+
+    val shouldCheck: Boolean =
+      (currentVersion.patch, targetVersion.patch, suffixAfterRelease(currentVersion)) match {
+        case (Some(current), Some(target), _) if current > target => true
+        case (Some(current), Some(target), suffix) if current == target => suffix
+        case (Some(current), None, _) => true
+        case _ => false
+      }
+
+    if (shouldCheck)
+      Some(targetVersion)
+    else
+      None
+  }
+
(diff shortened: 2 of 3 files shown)


rossabaker:
Rather than branch names, would it be better to check based on the version number?  If version, with snapshot stripped, is `x.y.za`, compare with `x.y.0a`.  If verison is `x.y.z`, look for `x.y.0`.

fiadliel:
The base version for the `0.7a` branch is `0.7a`, not `0.7.0a`. I could change the pull request to use this model, but do we want to use this (non-semver) version scheme in the future?
I don't really care about what format to use, but the build will definitely then encode versioning policy to a larger degree than currently.

One other thing is that `0.7a` has significantly more ABI failures with the current branch head, compared with `0.7.2a`, so the list of exceptions for the branch will be a lot longer.

rossabaker:
Yes.  Whatever differences may exist between 0.7a and 0.7.2a, you don't want to roll back, so I would target 0.7.2a on this release.

http4s uses [this](https://github.com/http4s/http4s/blob/d42d0c564ba48100132dc6ae9b9b482d5241d083/project/Http4sBuild.scala#L33-L44) to handle snapshots, establish a baseline, and handle exceptions to that baseline.  It's not a work of art, but works well for us.

fiadliel:
Updated with more logic around version naming, and automatic selection of "previous artifact". Though it has a little more knowledge of how version naming _must_ be, than I'm completely happy with. Comments welcome :smile:

djspiewak:
As an update, I consider this to be a really high priority to get in before 0.8.  I think that running MiMa on series/0.7 (much less 0.7a!) is probably a lost cause at this point.  @xuwei-k notified us of some binary incompatibilities back in 0.7.1 and we didn't have time to address them at that point.  I think the best thing we can do is just "do it right" going forward.

Another nice thing about working with master rather than series/0.7 is the fact that we've dropped scalaz 7.0 compatibility in master, and so the "a" branch is gone.

I'm going to pop in and try to help as much as I can with this, comments and work and such, as soon as I'm able to.  Couple things on my TODO list ahead of this one, but seeing as we want to get an 0.8 out within the next couple weeks, I can't take *too* long!  tl;dr: comments incoming soon!

fiadliel:
If you can describe the naming format you want to use in the future for releases, branches and snapshots, I'll update the patch.

I don't think it's a lost cause running this on 0.7 (if the base is 0.7.2+), but it will make the code more complex to support multiple naming schemes, so we could just have something clean and useful looking forward.

djspiewak:
Consider the general format to be `MM.mm.pp`, which is to say Major, minor and patch versions.  We don't have any hard compatibility guidelines between minor versions, though I would posit that (going forward) we should not be making significant semantic changes (for some subjective definition of "significant") between minor versions.  API breaking changes are accepted between minor versions.  Patch versions are intended to be fully compatible, backwards and forwards.  The only exceptions are for bug fixes, which can obviously alter semantics.

Releases can be parsed out of Git by looking for the `release/x.y.z` format tags, where I think the `.z` might be optional.  These tags will be in the main master branch path going forward (though patch versions higher than 0 will be in `series/x.y` branch paths), so it should be relatively easy to find them.

fiadliel:
Okay, I'll also assume a `-*` optional suffix (whether it's `-SNAPSHOT`, or a `git describe` format like `-15-g312aaee`).

For snapshot versions, I'm generally quite happy with something like
```
version := "git describe --tags --dirty --always".!!.stripPrefix("release/").trim
```

But since the build is already using sbt-git, might as well stick with that unless it can't do something we need.

fiadliel:
Updated. It now defaults to expecting a release "x.y.0".
Example maps of current version -> version to check for compatibility:
 - `"x.y.0" -> None`
 - `"x.y.1" -> Some("x.y.0")`
 - `"x.y.0-RC1" -> None`
 - `"x.y.0-SNAPSHOT" -> None`
 - `"x.y.0-1-12345" -> Some("x.y.0")`
 - `"x.y.1-RC1" -> Some("x.y.0")`

i.e. release candidates and snapshots are exceptions to a suffix representing a new build after the initial release.

How the version value is set is not changed here; if the current system is considered insufficient, that could be done in a different pull request.

djspiewak:
@fiadliel Right now, we don't have release candidate (`RC*`) versions, but I think we probably *should* and will in the future.  Your scheme looks excellent for that.  One quick comment:

> "x.y.0-1-12345" -> Some("x.y.0")

When we have versions that are based on git hashes, it's *just* the hash (untruncated).  We could configure sbt-git to have a base version, but at present we don't and I'm not quite sure how to do it properly.

With that said, I think that having a base version is probably the *only* way we can get MiMa to do anything sane for us "between" releases, so it seems like we should probably start doing that.  Thoughts?  Objections?

djspiewak:
@fiadliel I added a branch on my fork, `feature/mima`, which contains this work in addition to some adjustments to the versioning scheme to incorporate a base version (and also truncate the hash).  That should make things a bit easier.

The bad news is that the build failures on this PR are legit.  The problem is that this PR also includes the integration test stuff, which in turn is not compatible with 2.12 due to a lack of cross-build from scalameter.  Ideally I'd like to land *both* the integration test branch and the mima branch, but we need 2.12 compatibility and mima is a higher priority.  Scalameter doesn't appear to have a cross build, and actually hasn't pushed a new version since 2.11.1.  Do you think it would be better to just remove the integration work from this branch?

fiadliel:
> When we have versions that are based on git hashes, it's just the hash (untruncated). We could configure sbt-git to have a base version, but at present we don't and I'm not quite sure how to do it properly.
> 
> With that said, I think that having a base version is probably the only way we can get MiMa to do anything sane for us "between" releases, so it seems like we should probably start doing that. Thoughts? Objections?

I fully support my earlier suggestion for `version := "git describe --tags --always --dirty".!!` (with whatever cleanup is necessary to drop leading or trailing characters). Its only downside is requiring a working `git` binary wherever a build is made (not a large requirement, IMHO). With this, I no longer see a need for sbt-git either. I've been using this scheme with multiple projects in work for months, without a hiccup.

If you haven't used `git describe` before, it gives the tag of the current commit (if it has a tag). If the current tag does not have a tag, it looks through the history of commits for the preceding tag, and adds `-x-y` to the tag name, where `x` is the number of commits since the tag, and `y` is the short version of the current hash (prepended by a 'g').

With `--dirty`, you also won't accidentally create a release with local uncommitted changes, as the version will be tagged with a `-dirty` extension.

fiadliel:
Ah, but I read your followup message, so updates to set the correct version shouldn't be needed.

PR #429, opened 2015-08-16
Wrap NIO DirectoryStream for directory listing.

arzig:
If anything doesn't conform to style or best practices, let me know and I'll try my best to fix.

### DIFF
--- a/src/main/scala/scalaz/stream/nio/file.scala
+++ b/src/main/scala/scalaz/stream/nio/file.scala
@@ -12,7 +12,8 @@ import scala.io.Codec
 
 import java.nio.channels.{CompletionHandler, AsynchronousFileChannel}
 import java.nio.{CharBuffer, ByteBuffer}
-import java.nio.file.{StandardOpenOption, Path, Paths}
+import java.nio.file.{StandardOpenOption, Path, Paths, Files, DirectoryStream}
+import java.nio.file.attribute.{BasicFileAttributes, FileTime}
 import java.net.URI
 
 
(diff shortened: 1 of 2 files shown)


pchlupacek:
@arzig Thanks for this PR.

Do you think we can perhaps switch this to following signatures:

```scala
 
/** gets directory of supplied root path up to Int depth from the supplied root. To get full depth, supply Int.MaxValue  **/
def directoryRecurse(root:Path, maxDepth:Int = Int.MaxValue): Process[Task,Path] = ???

 /** gets directory of supplied root path  **/
def directory(root:Path):Process[Task,Path]=???

// + of course the `String` versions of `directory`

```

fthomas:
JFYI: Here is a gist of something I was playing around with https://gist.github.com/fthomas/45c4df6a8272352b854d 
It is horrible outdated, though.

Feel free to borrow from that or ignore it completely :smile:

arzig:
Reading your Gist, I noticed you use ```throw End```. Is there a difference in semantics between this and ```throw Terminated(End)```?

fthomas:
This is what I meant by outdated. :-) In the ancient version of the library that the gist is based on, `End` instead of `Terminated(End)` was thrown in these situations.

arzig:
I will probably do this later this evening, but one last question. Should ```directory(path)``` be equivalent ```directoryRecurse(path, 0)``` or ```directoryRecurse(path, 1)```. My initial though is 0, since it would not descend into any found directories but I wanted to see if you had any preferences

arzig:
I stepped out and was struck by a certainty that this doesn't handle max depth termination correctly (won't emit the directory at all). Needless I will add a test case and update if necessary.

arzig:
I was wrong. That said, I added a test case. Let me know if you would like anything else done.
Regards,
Ryan

src/main/scala/scalaz/stream/nio/file.scala
```diff
@@ -157,6 +157,39 @@ object file {
...
+   */
+  def directoryRecurse(root:Path, maxDepth:Int=Int.MaxValue): Process[Task,Path] = 
+    if (maxDepth < 0) Process.empty 
+    else directory(root) through isDirectoryW flatMap { 
+      case (path, true) => Process.emit(path) ++ directoryRecurse(path, maxDepth-1)
+      case (path, false) => Process.emit(path)
+    }
+
+  def directoryRecurse(root:String, maxDepth:Int): Process[Task,Path] = directoryRecurse(Paths.get(root), maxDepth)
+
```
pchlupacek: can we get this default argument value here too?

src/main/scala/scalaz/stream/nio/file.scala
```diff
@@ -157,6 +157,39 @@ object file {
...
 
+  private def withIsDirectory(path:Path): Task[(Path, Boolean)] = 
+    Task.delay(path -> Files.isDirectory(path))
+
+  private val isDirectoryW = Process.constant(withIsDirectory _).toSource
+
+  private def directoryStream(root:Path): Task[DirectoryStream[Path]] = Task.delay(Files.newDirectoryStream(root))
+
+  private def closeDirectoryStream(stream: DirectoryStream[Path]): Task[Unit] = Task.delay(stream.close)
+
```
pchlupacek: I'd rather move all these private defs and vals to body of `directoryRecurse`

src/main/scala/scalaz/stream/nio/file.scala
```diff
@@ -157,6 +157,39 @@ object file {
       }
   }
 
+  private def withIsDirectory(path:Path): Task[(Path, Boolean)] = 
+    Task.delay(path -> Files.isDirectory(path))
+
+  private val isDirectoryW = Process.constant(withIsDirectory _).toSource
```
pchlupacek: this can be expressed as `sink.lift { Task.delay(path -> Files.isDirectory(path))   }`, but really I think just doing the test in eval() and flatMap has better performance here.

fthomas:
With symlinks you can build cycles in your directory tree. If you have something like `directoryRecurse` without a way to say "don't follow symlinks" it will recurse `Int.MaxValue` times in case of cycles.

pchlupacek:
ok for now we perhaps has to add this parameter to signature, but looks like deep traversal is something that shall be implemented with one level traversal, instead of vice versa as it is today. Then maybe it can lead as to more generalised version, like passing in predicate indicating which nodes to traverse and which not.

arzig:
Regarding the default argument for directory recurse, this is apparently not allowed. It produces the following compiler error. `in object file, multiple overloaded alternatives of method directoryRecurse define default arguments.`

If anyone knows how to get around that I'd be interested to know.

pchlupacek:
Why you cannot just have this 

```scala 
def directoryRecursively(root:Path, followSymLinks:Boolean = false, maxDepth:Int=Int.MaxValue) = ???
def directoryRecursively(path:String) = ??? //invoking with defaults....
```

arzig:
I was thinking more about passing predicates to determine which nodes to expand. Would something akin to the following make sense?

```scala
type DirectoryVisitPredicate = FileAttributes => Boolean
type FileTypePredicate = FileType => Boolean 

def directoryRecurse(root: Path, visitPredicate: DirectoryVisitPredicate = Function.const(true) _, maxDepth:Int=Int.MaxValue)
```

Alternatively, would there ever be a situation where one might want to make decisions to visit a node based on both the depth and the file attributes such as?
```scala
type DirectoryVisitPredicate = (FileAttributes, Int) => Boolean
```

pchlupacek:
I think typical pattern how we do these sort of things is in `process1` as result from some sort of source.  To me the recursive function shall cover 99% usages as it was suggested, and if someone wants completely different behaviour (i.e traverse only directories that starts with `A`), than he would do this manually combining the `directory` combinator (non recursive) and some sort of Process1 with `go` style.

arzig:
@pchlupacek is there anything else you would like to see in the implementation then?

pchlupacek:
@arzig Sorry to be not clear, but I don have good feeling about the visitor algebra/pattern you proposing. I'd like to keep this simple as suggested before. Just 3 simple parameters covering 99% of usages for now.

arzig:
@pchlupacek I don't think I was clear either. I actually understood that you didn't want the visitor algebra. I wanted to make sure I was correct in my interpretation that I had hit all the other requests.

pchlupacek:
@arzig I think implementation should be very straightforward. you just implement `directory` to fetch one level `Paths`. Then, directoryRecusive is just implemented in terms of directory and simple `go` that does tree walking.

PR #432, opened 2015-09-02
Exposes bound on a Queue as Option[Int].

alissapajer:


pchlupacek:
+1 LGTM

pchlupacek:
just perhaps an idea, if we cannot add on the queue signal of items that can be enqueued, and simple signal of queue being full

```scala

def available : Signal[Int] = ??? // combined from size and bound 

def full: Signal[Boolean] = available.map (_ <= 0)

```

alissapajer:
I think those would be useful. We'd want something like
```scala
def available : Option[Signal[Int]]
def full: Option[Signal[Boolean]]
```
to address the unbounded queue case. In other words, `available` and `full` would return `None` in the unbounded case since those don't have meaning without a bound.

The signals will also need to be wired up to halt when the queue halts, like the `size` signal does now. Though if we implement it using `size` that should be taken care of.

pchlupacek:
@alissapajer I think is perfectly ok when they would return values Int.MaxValue or false always. I am also ok to return the option instad of Int/Boolean. Howener I think signal should not be optional, i.e. signature may be something like

```scala
def available : Signal[Option[Int]]
def full: Signal[Option[Boolean]]
```

alissapajer:
I think using `Int.MaxValue` and `false` would be more consistent with what we return in the case of a bounded queue. So I'm fine with removing the `Option` all together. Thinking about it, that will make it easier to work with as you don't need to handle the two cases (bounded and unbounded) separately.

pchlupacek:
Yeah, was my original thought. If you really care about the type of queue you have always that strict Option of bound value available.

PR #434, opened 2015-09-09
skeleton for regression testing

wookietreiber:
- adresses #399

- uses [scalameter regression testing][1]

- runs in sbt integration test target, i.e. `it:test`

- generates an offline html report in the tmp subdir of the main project
  dir, there is an index.html you can open to inspect the results, the
  regression history is also saved in that tmp subdir

- might need a few tweaks in terms of stability, confidence intervals, etc.

- I included just a simple dummy example to give you an idea of how it
  looks like, but adding new stuff to test should be straightforward

- I am not sure whether `Process.constant(1).take(n)` is the fastest way
  to create a reasonably sized Process

[1]: http://scalameter.github.io/home/gettingstarted/0.7/regressions/index.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ tags
 *#
 z_local.sbt
 .DS_Store
+tmp/
--- a/build.sbt
+++ b/build.sbt
@@ -112,3 +112,19 @@ initialCommands := "import scalaz.stream._"
 doctestWithDependencies := false
 
 doctestSettings
+
+// -------------------------------------------------------------------------------------------------
+// integration tests / regression testing
+// -------------------------------------------------------------------------------------------------
+
+configs(IntegrationTest)
+
+Defaults.itSettings
+
+libraryDependencies += "com.storm-enroute" %% "scalameter" % "0.8-SNAPSHOT" % "it"
+
+logBuffered in IntegrationTest := false
+
+testFrameworks += new TestFramework("org.scalameter.ScalaMeterFramework")
+
+parallelExecution in IntegrationTest := false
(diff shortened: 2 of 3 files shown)


wookietreiber:
See also #406.

wookietreiber:
Warning: This won't work yet, because all dependencies don't resolve yet.

wookietreiber:
To support both 2.12.0-M1 and 2.12.0-M2 (as is currently specified in crossScalaVersions) @axel22 needs to release the currently missing scalameter_2.12.0-M2;0.8-SNAPSHOT.

axel22:
One problem I have with scala-chart-0.5.0 is that it apparently broke source compatibility with earlier versions.
If you have released 0.5.0 for all Scala versions (2.10 onward), then ScalaMeter could start depending on 0.5.0.

On September 9, 2015 10:51:55 PM GMT+02:00, Christian Krause <notifications@github.com> wrote:
>To support both 2.12.0-M1 and 2.12.0-M2 (as is currently specified in
>crossScalaVersions) @axel22 needs to release the currently missing
>scalameter_2.12.0-M2;0.8-SNAPSHOT.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/scalaz/scalaz-stream/pull/434#issuecomment-139042033

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

wookietreiber:
Yes, I just wanted to write about that. I will prepare a PR for scalameter that will work with scala-chart 0.5.0.

axel22:
Thanks - that would be helpful!

On September 9, 2015 11:00:43 PM GMT+02:00, Christian Krause <notifications@github.com> wrote:
>Yes, I just wanted to write about that. I will prepare a PR for
>scalameter that will work with scala-chart 0.5.0.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/scalaz/scalaz-stream/pull/434#issuecomment-139044385

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #436, opened 2015-09-25
Add a link to SDBC for streaming to and from some databases.

shawjef3:


djspiewak:
Uh… to avoid extreme weirdness in the commit log just for a one line readme change, would you mind squashing this down to a single commit?

xanderdunn:
@djspiewak Might it make more sense to have a wiki page that lists projects using scalaz-stream?  Then just have a link in the README.md to that page, the same as it's done for the [Additional Resources](https://github.com/scalaz/scalaz-stream/wiki/Additional-Resources) page.  If you that's desirable, I can make the changes.

fthomas:
@xanderdunn Asking people to add projects that use scalaz-stream to the README hasn't worked out as it seems. Projects like [doobie](https://github.com/tpolecat/doobie), [specs2](https://github.com/etorreborre/specs2), [remotely](https://github.com/oncue/remotely), or [scalaz-netty](https://github.com/RichRelevance/scalaz-netty) all use scalaz-stream but were never added. I assume that the barrier to open a PR just for adding a link is not as low as we might think - I also guess that people refrain from opening a PR if they are not the authors  of those projects. If we can have a more complete list by moving it to the wiki (which should have a lower entry barrier) I'm all for it.

xanderdunn:
@fthomas For those reasons, I think this would be a great idea.  The barrier to change the wiki is far lower.  A pull request is a much more ominous thing for a newcomer.  Posting to the Google Group seems like a lot of indirection for such a small thing.

I created a [wiki page](https://github.com/scalaz/scalaz-stream/wiki/Projects-Using-scalaz-stream) for it and opened #449 to reference it in the README.md.

Frequent changes to it will make the repository commits cleaner.

fthomas:
@xanderdunn Thanks!

fthomas:
Closing since the list of projects using the library is now in the wiki and it also includes a link to SDBC.

PR #440, opened 2015-10-07
Fixed chunkAll/chunkBy to correctly propagate termination cause 

pchlupacek:
The `chunkBy` implementation caused together with `once` to effectively swallow termination causes. The test demonstrates that.

### DIFF
--- a/src/main/scala/scalaz/stream/process1.scala
+++ b/src/main/scala/scalaz/stream/process1.scala
@@ -67,7 +67,7 @@ object process1 {
    */
   def chunkBy[I](f: I => Boolean): Process1[I, Vector[I]] = {
     def go(acc: Vector[I], last: Boolean): Process1[I, Vector[I]] =
-      receive1Or[I,Vector[I]](emit(acc)) { i =>
+      receive1Or[I,Vector[I]](if (acc.nonEmpty) emit(acc) else halt) { i =>
         val chunk = acc :+ i
         val cur = f(i)
         if (!cur && last) emit(chunk) ++ go(Vector(), false)
@@ -552,7 +552,7 @@ object process1 {
   /** Throws any input exceptions and passes along successful results. */
   def rethrow[A]: Process1[Throwable \/ A, A] =
     id[Throwable \/ A].flatMap {
-      case -\/(err) => throw err
+      case -\/(err) => Process.fail(err)
       case \/-(a)   => emit(a)
     }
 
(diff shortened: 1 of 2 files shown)


pchlupacek:
Please note this breaks current behaviour of chunkAll. The change resolves issue, where termination causes was swallowed by current implementation of `chunkBy`. It also makes sure the `chunkBy` behaviour is same as chunk, where these two had different behaviours, when the process did not emit single value. 

Now, the `chunkAll` is not like `chunkAll.awaitOption.map(_.toSeq.flatten)` which previously was the case. 

Perhaps there shall be another `chunkAll` that will have old `chunkBy` behaviour, but not sure about the name though.

pchlupacek:
Ok, after seeing the amount of changes that behaviour will change in our project I have decided to change the implementation and make the fix much less disruptive. However this makes `chunk` and `chunkBy` behave differently. 

So now, the `chunkBy` correctly passes through the exception, and `chunk` and `chunkBy` have the same behaviour as before 
  - `chunkBy` on `Kill` emits `emit(Vector())` instead of `halt`
  - `chunk` on `Kill` emits `halt`

For the 0.9 release though, I think the behaviour must be the same. @fthomas what do you think?

fthomas:
Do the other transducers that use `receive1Or` have the same problem you observed with `chunkBy`?

pchlupacek:
@fthomas this is something we have to review in new design I bet. This one was specific combination of chunkBy and once. What worries me more though is that we have two transducer with similar name, doing almost the same but completely different behaviour. 

The core problem lays in the algebra that relies on cause propagation. Not sure if this is still the same problem with new algebra, but we have to really make sure we won't endup the same.

pchlupacek:
@djspiewak can you please take a quick look on this. I think is good to merge.

pchlupacek:
@pchiusano can you please take a very quick look on this one? I think this resolves nasty bug in 0.8 series  that we hit while ago, but would like to see your view on this. I am not sure how @djspiewak is going with other fixes/patches towards 0.8 but perhaps some . release is still something we will have to do before 0.9

pchiusano:
@pchlupacek do you know if this is still relevant or can we close?

pchlupacek:
this is very relevant to 0.8 series. I am surprised ppl was not hitting this :-) I was perhaps hoping we do 8.1 with some stuff @djspiewak planned. Not sure where he is at. Shall we just merge it to something called 8.1 branch?

pchiusano:
Just use the series/0.8 and series/0.8a branches. We can do a 0.8 point release off those if we want.

PR #443, opened 2015-10-07
FS2: Add Process1Spec that tests the current transducers

fthomas:
This adds basic tests for `process1.{lift, last, take}`. One interesting question here: how can I test if `take` preserves the chunkiness of the input? I've also added `emitAll` which is nicer to use than `Stream(v: _*)`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,7 +39,7 @@ libraryDependencies ++= Seq(
   "org.scalaz" %% "scalaz-concurrent" % "7.1.3",
   "org.scodec" %% "scodec-bits" % "1.0.9",
   "org.scalaz" %% "scalaz-scalacheck-binding" % "7.1.3" % "test",
-  "org.scalacheck" %% "scalacheck" % "1.12.4" % "test"
+  "org.scalacheck" %% "scalacheck" % "1.12.5" % "test"
 )
 
 sonatypeProfileName := "org.scalaz"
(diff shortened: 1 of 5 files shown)


fthomas:
The build failed because FS2 does not compile with Scala 2.10 and the latest ScalaCheck is not available for 2.12.0-M1. @pchiusano you said you want to publish for > 2.11 only, right? @djspiewak can we remove 2.12.0-M1 from the CI builds given that we already build against -M2?

pchiusano:
Try writing a combinator chunks - a transducer that repeatedly awaits to
obtain a chunk, then does Pull.write. You can build a Process1[I,Chunk[I]]
and then use it in the tests.

We should introduce an Arbitrary instance for different streams, so inputs
can be more interesting than 1-chunk streams.

I think we should rename Pull.write to Pull.output. Otherwise it will be
confusing when we introduce the Writer functions.

fthomas:
Another question. Can we change the type of `take`'s parameter to `Long`, so `take(n: Long)` instead of `take(n: Int)`?

fthomas:
@pchiusano If you think the changes here are okay, you can merge them right now. I can introduce `chunks` in a later PR.

pchiusano:
Yeah that is a good idea.

pchiusano:
K merging

pchiusano:
I can't merge this on my phone due to the build failure but please go ahead and merge it. :)

fthomas:
@djspiewak I just noticed that you removed the 2.12.0-M1 build on master. I will remove 2.10 and 2.12.0-M1 then in the next PR.

PR #445, opened 2015-10-11
FS2: Test that take preserves the chunkiness of the input

fthomas:
This adds `process1.chunks` which is used to test that `take` preserves the chunkiness of its input.

What I observed here:
* The actual `Process1` definition feels like 100% boilerplate:
```scala
def chunks[I]: Process1[I,Chunk[I]] =
  new Process1[I,Chunk[I]] { def run[F[_]] = _.pull[F,Chunk[I]](pull.chunks) }
```
* If you leave out the type parameter in `_.pull[F,Chunk[I]](...)`, you get weird compiler errors.
* I had to repeat `Chunk[I]` there three times
* `Chunk` has no sensible `equals`:
```scala
scala> val x = Chunk.singleton(1) == Chunk.singleton(1)
x: Boolean = false
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: scala
 sbt_args: -J-Dproject.version=travis-SNAPSHOT
 scala:
-  - 2.10.5
   - 2.11.7
-  - 2.12.0-M1
   - 2.12.0-M2
 jdk:
   # - oraclejdk7
--- a/build.sbt
+++ b/build.sbt
@@ -12,7 +12,7 @@ git.gitTagToVersionNumber := {
 
 scalaVersion := "2.11.7"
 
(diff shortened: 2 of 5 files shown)


pchiusano:
I'd give chunk a default equals in the trait that just compares elements.

The process1 definitions are boilerplate. All the logic will be in the pull definition. We could just not have the Process1 instances and people could use Stream.pull directly (instead of pipe) or have pipe be an alias for pull. I think we should seriously consider that.

pchiusano:
If you want, you can just focus on writing the pull functions for now, and
call these directly via Stream.pull, which should be just as convenient.
Later we can decide whether to also include the Process1/Tee/Wye instances.

PR #446, opened 2015-10-15
Fix typo

xanderdunn:
`awakeEvery(5 seconds)` should return at 5s, 10s, and 15s, right?  There's also nothing shown here that limits it to three output values, right?


fthomas:
:+1: Thanks!

The build failure comes from the fact that 2.12.0-M1 was removed from build.sbt but not from .travis.yml. I'll correct that in a second.

PR #451, opened 2015-10-21
Implement process1.filter and syntax sugar.

fiadliel:
This implements filtering by a predicate on data streams in FS2. It tries to retain chunkiness, and limits (somewhat) new object creation.

This is also a learning exercise (for me) on where/what should go in various places in the new design.
e.g. should this be available for pull instead? (I don't think so). Is it over the top in checking various Chunk sizes?

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -20,17 +20,7 @@ object process1 {
 
   /** Emit inputs which match the supplied predicate to the output of the returned `Pull` */
   def filter[F[_], I](f: I => Boolean)(implicit F: NotNothing[F]): Handle[F,I] => Pull[F,I,Handle[F,I]] =
-    h => h.await flatMap { case chunk #: h =>
-      val filtered =
-        chunk.size match {
-          case 0 => chunk
-          case 1 if f(chunk(0)) => chunk
-          case 1 => Chunk.empty
-          case _ => Chunk.seq(chunk.iterator.filter(f).toSeq)
-        }
-
-      Pull.output(filtered) >> filter(f).apply(h)
-    }
+    h => h.await flatMap { case chunk #: h => Pull.output(chunk filter f) >> filter(f).apply(h) }
 
   /** Write all inputs to the output of the returned `Pull`. */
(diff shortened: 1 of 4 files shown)


src/main/scala/fs2/Process1.scala
```diff
@@ -18,6 +18,20 @@ object process1 {
   def chunks[F[_],I](implicit F: NotNothing[F]): Handle[F,I] => Pull[F,Chunk[I],Handle[F,I]] =
     h => h.await flatMap { case chunk #: h => Pull.output1(chunk) >> chunks.apply(h) }
 
+  /** Emit inputs which match the supplied predicate to the output of the returned `Pull` */
+  def filter[F[_], I](f: I => Boolean)(implicit F: NotNothing[F]): Handle[F,I] => Pull[F,I,Handle[F,I]] =
+    h => h.await flatMap { case chunk #: h =>
+      val filtered =
```
pchiusano: I'd suggest just adding a `filter` method to `Chunk`, which can be overridden with some more efficient implementation in the various `Chunk` types. I definitely would not include the special-casing here.

Then your implementation just becomes `h.await flatMap { case chunk #: h => Pull.output(chunk filter f) >> filter(f).apply(h) }`

fiadliel:
That makes complete sense! Patch updated (with sys.error for Bytes and Bits implementations).

I'm sure that filter on Bits makes zero sense; slightly above zero for Bytes, but I still don't see any great use cases for it (if there were, the right place for the functionality would be in ByteVector itself).

fiadliel:
(Travis failure is nothing to do with this PR)

pchiusano:
Don't throw an exception in `Bits` and `Bytes`. As a general rule of thumb, these should be thought of just as more efficient representations of a `Chunk[A]`, and they should behave the same as a polymorphic `Chunk[A]` type, like `Chunk.seq`. A `Chunk.seq[Boolean]` has a filter function also, and the implementation is not even in a position to inspect the `A` parameter and do something different based on the type (ignoring crazy stuff like accepting an implicit manifest argument or something).

From another point of view, it might not be very common or useful, but we can't be 100% convinced that every usage of `filter` on these types is a user error, so we shouldn't throw.

fiadliel:
Okay; implementations for all Chunk types added, plus tests.

pchiusano:
Great thanks!

PR #452, opened 2015-10-22
FS2: Add process1.delete

fthomas:
Using `await1` here allowed the implementation to be copied from scalaz-stream. If this should preserve chunkiness something like `delete` needs to be implemented in `Chunk`. I guess this is the right thing to do?


pchiusano:
Yeah, preserve chunkiness. I'd maybe add an `indexOf` method to `Chunk` and give it a default implementation using the `iterator` method. Then you can handle the final chunk with a `Chunk.take` and `Chunk.drop` call around the deleted index, and let through any chunks for which `indexOf` returns `None`.

fthomas:
Updated `delete` such that it preserves chunkiness by using `indexWhile`, `take` and `drop` on `Chunk`.

src/test/scala/fs2/Process1Spec.scala
```diff
@@ -21,6 +21,12 @@ object Process1Spec extends Properties("process1") {
     s.pipe(chunks).flatMap(Stream.chunk) ==? v.flatten
   }
 
+  property("delete") = forAll { (s: PureStream[Int]) =>
+    val v = run(s.get)
+    val i = Gen.oneOf(v).sample.getOrElse(0)
```
pchiusano: cute

pchiusano:
looks great, merging

PR #453, opened 2015-10-23
add drop/takeWhile/dropWhile for `Process1`

ShengC:
I figure best way to learn the redesign is to actually write some transducers myself :) Feel free to reject though if you think the internal api is not quite ready yet.


src/main/scala/fs2/Process1.scala
```diff
@@ -45,6 +45,29 @@ object process1 {
...
+  def drop[F[_], I](n: Long)(implicit F: NotNothing[F]): Handle[F, I] => Pull[F, I, Handle[F, I]] = 
+    h =>
+      if (n <= 0) id.apply(h)
+      else Pull.awaitLimit(if (n <= Int.MaxValue) n.toInt else Int.MaxValue)(h).flatMap {
+        case chunk #: h => drop(n - chunk.size).apply(h)
+      }
+      
+  def dropWhile[F[_], I](p: I => Boolean)(implicit F: NotNothing[F]): Handle[F, I] => Pull[F, I, Handle[F, I]] =
+    h => 
+      Pull.await1Option(h).flatMap {
```
pchiusano: Even though it is prettier to write this and `takeWhile` in terms of `await1Option`, it should preserve chunkiness. In client code, you can write in whatever style you prefer, but for the FS2 library functions I'd like to preserve chunkiness whenever feasible.

If @fthomas has not done this already, I would add a method to `Chunk`:

```
trait Chunk[+A] {
  /** Return the first index whose element satisfies the predicate. 
      `c.indexOf(f).map(i => f(c(i)))` should be `Some(true)` or `None` if no item matches. */
  def indexOf(f: A => Boolean): Option[Int] = /* implement using .iterator */
```
The implement `takeWhile` and `dropWhile` in terms of this.

Other than that this looks good.

@fthomas @ShengC @djspiewak @pchlupacek I created a wiki page just to help people coordinate: https://github.com/scalaz/scalaz-stream/wiki/FS2-upgrade-wiki-page

Idea is there are a lot of functions to upgrade, and it would be nice to get some concurrency without having people accidentally duplicate work.

ShengC: @pchiusano I think that makes sense to me too. I will have to update the PR after hours though, since firm proxy won't allow me to push out code. If by then the method is still not added, I will add it myself.

pchiusano: Great. No rush. :)

ShengC:
Ok, I am able to preserve "chunkiness" with `indexWhere`. I will wait until #454 and #455 merged in.

pchiusano:
I think we can find a more elegant implementation of takeWhile and dropWhile. Try negating the predicate and passing that to indexWhere - so for takeWhile you look for the first index where the predicate becomes false. If it's None, you include the full chunk. If it's Some, you emit up to but not including that index and stop.

Something similar for dropWhile. Try it out and see if you can get the implementation down to a couple lines each.

ShengC:
@pchiusano that's clever :) Also can we exclude the tests pertaining to `scalaz-stream`. It takes too long to run the full suite of tests in Travis, but really we only have a few for `fs2`

fthomas:
@ShengC Try `testOnly fs2.Process1Spec` in sbt

ShengC:
@fthomas I know how to do that in sbt. It is just my personal habit whenever I update a PR, I want to wait until it passes the Travis Test. My experience suggests it is not entirely true that if tests pass locally, they would for sure pass in Travis :)

fthomas:
@ShengC sorry, I read over that you want to disable the scalaz-stream tests on Travis.

ShengC:
@pchiusano @fthomas wondering if this is good to merge, or there is still something you'd like to improve with?

pchiusano:
Looks great! As a minor style thing, you can put the case chunk #: h => on the same line if it fits. One of us can fix that post merge though.

ShengC:
Thanks :)

PR #455, opened 2015-10-23
Topic/redesign test improvements

pchiusano:
This PR makes several changes:

* The various `Process1` properties range across the different possible stream structures by taking a `PureStream` (rather than just taking a `Vector` input and making a static choice of how to convert that to a `Stream`). Going forward, let's use this approach as it's much better coverage.
  * It would be better to have a separate test that ensures the various `Stream` operations correctly build up a stream. I'm working on that now and will add it to this PR.
* If you need to throw in a failure, take an implicit `Failure`, which ranges over different possible failures (a direct `Stream.fail` call, an exception inside an effect, an exception inside pure code, etc) and splice that somewhere into the stream. See the usage of this in `ResourceSafetySpec`.
  * You can have your property take a `SmallPositive` or `SmallNonnegative` (all in `TestUtil`) if you want to randomize where the failure is placed.
* I implemented Pavel's suggestion of having the finalizer itself throw an exception in the nested bracket test. Guess what? That wasn't being handled correctly! Fixed now. :)
* I added a `mapChunks` function. This could be used to implement `filter`.

@fthomas, @fiadliel, and @ShengC have a look. Also @pchlupacek check out the nested bracket test if you want to see how I added that failure-in-finalizer scenario.


pchiusano:
FYI, I also changed the build settings to show timing information and full stack traces for each test.

fthomas:
`mapChunks` can also be used to implement `lift`: `lift(f) = mapChunks(_.map(f))`

pchiusano:
@fthomas good point. Actually, as a general rule, if you can implement something with `mapChunks` (or a similar combinator that abstracts over the chunkiness) I'd say that's a good idea!

I'm done adding stuff to this branch, so I'll merge (or anyone else can) once Travis gives the ok.

PR #457, opened 2015-10-24
Add shortcut for h => h.await.flatMap?

fthomas:
This is a question. In scalaz-stream we have `receive1` which is the same as `await1.flatMap`. Do we want something similar here? `h => h.await flatMap ` is already a recurring pattern.


pchiusano:
I like it. But you can make the return type of the Pull any R, I don't think the implementation cares if it's another Handle.

pchiusano:
Also with the generalized signature I suggested, these are no longer Process1s and can be moved to the pull1 trait which gets mixed into Pull.

fthomas:
I've generalized `receive` and `receive1` and moved them into `pull1`.

pchiusano:
Looks good!

PR #458, opened 2015-10-24
Use indexedSeq instead of seq

fthomas:
Operations on a IndexedSeq-Chunk should not upcast to Seq-Chunk.


pchiusano:
Good catch.

PR #459, opened 2015-10-24
add `zipWithIndex` to `Process1`

ShengC:
I took the liberty of adding `zipWithIndex` method to `Chunk` as well. If you feel there is better place to hold it, leave me a message. Thanks!

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -96,13 +96,12 @@ object process1 {
       
   /** Zip the elements of the input `Handle `with its indices, and return the new `Handle` */    
   def zipWithIndex[F[_], I](implicit F: NotNothing[F]): Handle[F, I] => Pull[F, (I, Int), Handle[F, I]] = {
-    def go(n: Int)(h: Handle[F, I]): Pull[F, (I, Int), Handle[F, I]] = {
-      Pull.await(h) flatMap {
-        case chunk #: h =>
+    def go(n: Int): Handle[F, I] => Pull[F, (I, Int), Handle[F, I]] = {
+      Pull.receive({ 
+        case chunk #: h => 
           Pull.output(chunk.zipWithIndex.map({ case (c, i) => (c, i + n) })) >> go(n + chunk.size)(h)
-      }
-    }
-    
+      })
+    }    
     go(0)
   }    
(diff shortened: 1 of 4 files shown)


src/main/scala/fs2/Process1.scala
```diff
@@ -94,6 +94,18 @@ object process1 {
...
       
+  /** Zip the elements of the input `Handle `with its indices, and return the new `Handle` */    
+  def zipWithIndex[F[_], I](implicit F: NotNothing[F]): Handle[F, I] => Pull[F, (I, Int), Handle[F, I]] = {
+    def go(n: Int)(h: Handle[F, I]): Pull[F, (I, Int), Handle[F, I]] = {
+      Pull.await(h) flatMap {
+        case chunk #: h =>
+          Pull.output(chunk.zipWithIndex.map({ case (c, i) => (c, i + n) })) >> go(n + chunk.size)(h)
+      }
+    }
+
```
pchiusano: No need for line break here.

pchiusano:
Let's wait for @fthomas to merge his receive functions, then you can implement in terms of those.

ShengC:
Ok, sounds good.

src/main/scala/fs2/Process1.scala
```diff
@@ -94,6 +94,17 @@ object process1 {
           }
       }
       
+  /** Zip the elements of the input `Handle `with its indices, and return the new `Handle` */    
+  def zipWithIndex[F[_], I](implicit F: NotNothing[F]): Handle[F, I] => Pull[F, (I, Int), Handle[F, I]] = {
+    def go(n: Int): Handle[F, I] => Pull[F, (I, Int), Handle[F, I]] = {
+      Pull.receive({
```
pchiusano: No need for the parents here, the braces are sufficient. Pull.receive { case chunk #: h => ... } all on one line.

pchiusano:
Can you also add an alias to ProcessOps? Follow the existing conventions there.

ShengC:
@pchiusano does it look good to merge now ?

pchiusano:
Looks good.

Actually I'll probably reformat the code a bit after but this is fine to merge. :) I should produce a style guide at some point.

Thanks!

PR #461, opened 2015-10-24
FS2: Implement process1.find

fiadliel:
This writes the output of the first input matching a predicate, to the output of the supplied Pull, then halts.

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -44,6 +44,10 @@ object process1 {
   def filter[F[_],I](f: I => Boolean)(implicit F: NotNothing[F]): Handle[F,I] => Pull[F,I,Handle[F,I]] =
     mapChunks(_ filter f)
 
+  /** Emits the first input (if any) which matches the supplied predicate, to the output of the returned `Pull` */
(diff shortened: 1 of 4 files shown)


pchiusano:
This operation (and any operation that has a single output) is more useful if it emits to the result of the Pull (the R parameter) rather than the output, since then it can be an intermediate result inside building up another transducer. You can define it in the pull1 trait. Okay to include an alias in process1 that calls the other and then outputs the result.

fiadliel:
This will be my first time looking at this layer properly: it currently looks like I should have a function with a signature like:

```scala
  def find[F[_],I](f: I => Boolean): Handle[F,I] => Pull[F,Nothing,Step[I,Handle[F,I]]]
```
(with the unconsumed stream at the input value just after the value written to the output).

and then perhaps something like this in process1:

```scala
Pull.find(f).apply(h).flatMap { case o #: h => Pull.output1(o) >> Pull.done }
```

Anyway, I'll work at it, and update the PR later.

pchiusano:
That looks right.

pchiusano:
I just realized this was never merged. It looks good. I don't get a notification if you just add a commit, only if you leave a comment, so if you leave a comment saying you've updated I won't forget about it. :)

PR #462, opened 2015-10-24
Topic/redesign signal

pchlupacek:


pchlupacek:
@pchiusano some WIP. I am still struggling with callback definition via Async. The proposed A => F[A]  you suggest seems right, but still when you need to queue up the readers (see https://github.com/scalaz/scalaz-stream/pull/462/files#diff-151adb5a94d6c9e95515201469e11805R76), this won't work nicely. Plus I don't like the forkRun api either. Perhaps I need another on AsyncExt api to cover that.

Can you just take a very quick look on that perhaps you would have some ideas there ?

pchiusano:
I'm taking a look now, but am on my phone...

pchiusano:
Just noticed forkRun can be defined using set -

forkRun(t) = ref.flatMap(r => set(r)(t))

pchiusano:
I think I need to spend more time with this when I get in front of a computer. I am going to be busy the rest of today though.

pchlupacek:
@pchiusano just pushed change. And no worries I may not have time jump on that next 2-3 days. I resolved the asynchronus issue with additional refs, and in fact I like it. `forkRun` is no longer necessary. So only api changed are the `modify` on Async. And we can leave it on AsyncExt before we will know 100% this is the way to go.  For now, I have feeling queue and topic can be implemented with same technique. 

On the processes I still need to implement them, they shall be trivial though I didn't find couple of basic combinators on Stream._  (i.e. repeat). Would be nice to make list of Stream._ helpers we may need and put it on todo list? I can do that if you wish.

pchiusano:
Okay cool I will take a look tomorrow or Monday!

src/main/scala/fs2/async/AsyncExt.scala
```diff
@@ -0,0 +1,208 @@
...
+   *
+   * If the `ref` is not yet set, this queues up until `ref` is set for the first time.
+   *
+   * Note that order of queued modifications is not preserved
+   *
+   * Likewise with `set` Resulting task may complete before final value is applied to `ref`,
+   * but it is guaranteed that it will be applied before any other modify/set on the `ref`.
+   *
+   */
+  def modify[A](ref:Ref[A])(f: A => F[A]):F[(A,A)]
```
pchiusano: These docs could use some cleanup and clarification. We can work on that later.

What is the pair that's returned? Is it `(old,new)`? Is `old` the argument that was passed to the `f`? Let's just introduce `case class Change[+A](previous: A, now: A)` and return a `Change[A]` here. You can put `Change` in the companion object of `AsyncExt`.

src/main/scala/fs2/async/AsyncExt.scala
```diff
@@ -0,0 +1,208 @@
...
+   * Like `modify` but allows to interrupt the modify task (Interruptee)
+   * in case the task is waiting for the `ref` to be set for the first time by evaluating the
+   * second returned task (Interruptor)
+   *
+   * Interruptor is guaranteed to complete always with true, if Interruptee was interrupted before
+   * evaluation started, and false if Interuptee was completed before interruption.
+   *
+   * Note that Interruptor won't be able to interrupt computation after `f` was evaluated and produced computation.
+   */
+  def cancellableModify[A](r:Ref[A])(f:A => F[A]):F[(F[(A,A)], F[Boolean])]
```
pchiusano: is the `Boolean` needed? This signature is already pretty complicated. When would I care whether or not my cancellation cancelled an already completed task or one that was ongoing? I feel like if I really needed that info I could track it in some other way.

src/main/scala/fs2/async/immutable/Signal.scala
```diff
@@ -0,0 +1,102 @@
...
+ * Created by pach on 10/10/15.
+ */
+trait Signal[F[_],A]  {
+  /**
+   * Returns a continuous stream, indicating whether the value has changed.
+   * This will spike `true` once for each time the value of `Signal` was changed.
+   * It will always start with `true` when the process is run or when the `Signal` is
+   * set for the first time.
+   */
+  def changed: fs2.Stream[F, Boolean]
```
pchiusano: can you get rid of the `fs2.` prefix everywhere? It is not needed if you've imported `fs2.Stream`. If you need to reference `scala.collection.immutable.Stream`, just refer to it qualified or import it with a different name.

src/main/scala/fs2/async/immutable/Signal.scala
```diff
@@ -0,0 +1,102 @@
...
+   * when the `value` is changed.
+   */
+  def changes: fs2.Stream[F, Unit]
+
+  /**
+   * Discrete signal yielding to `true` when this signal halts.
+   * Otherwise emits single `false` and awaits signal to be halted and then emits `true`.
+   * @return
+   */
+  def closed: fs2.Stream[F,Boolean]
```
pchiusano: hmm, is this still needed? I liked the idea of signals just being infinite, and you can do a `takeWhile` or whatever on the consumer side if you aren't interested in the values after a certain point.

src/main/scala/fs2/async/package.scala
```diff
@@ -0,0 +1,39 @@
...
+ * Created by pach on 10/10/15.
+ */
+package object async {
+
+
+  /**
+   * Create a new continuous signal which may be controlled asynchronously.
+   * Note that this would block any resulting processes (discrete, continuous) until any signal value is set.
+   */
+  def signalUnset[F[_]:AsyncExt,A]: Stream[F,mutable.Signal[F,A]] =
```
pchiusano: I'd like to just delete this method. Use `Option` + filtering on the consumer side if you want this.

src/main/scala/fs2/async/mutable/Signal.scala
```diff
@@ -0,0 +1,149 @@
...
+
+        def discrete: fs2.Stream[F, A] = {
+          def go(stamp:Int):fs2.Stream[F, A] = {
+            Stream.eval(getChanged(stamp))
+            .flatMap { case (s,a) => Stream(a).append(go(s)) }
+          }
+          Stream.eval(getStamped).flatMap { case (s,a) => Stream(a).append(go(s))}
+        }
+
+        def changed: fs2.Stream[F, Boolean] = {
```
pchiusano: I don't think this is all that useful as a continuous signal... unless you happen to read at exactly the right moment it will always be false. Can we just delete this?

src/main/scala/fs2/async/mutable/Signal.scala
```diff
@@ -0,0 +1,149 @@
...
+   * Halts this signal.
+   * Halting this signal causes any modification
+   * operations (Signal#set, Signal#getAndSet, Signal#compareAndSet) to complete
+   * with `Signal.Terminated` exception.
+   *
+   * Any Streams that reads from this signal will be halted once this signal is closed.
+   *
+   * @return
+   */
+   def close:F[Unit]
```
pchiusano: I vote to get rid of this if we are changing the API significantly anyway.

src/main/scala/fs2/async/AsyncExt.scala
pchlupacek: Yeah, +1 for Change.

src/main/scala/fs2/async/AsyncExt.scala
pchlupacek: I am not sure on this either. Let's leave it for now (as Unit/Boolean is just 3 chars difference) and I gonne remove it after implementing queue if it won't be needed.

src/main/scala/fs2/async/immutable/Signal.scala
pchlupacek: ok

src/main/scala/fs2/async/immutable/Signal.scala
pchlupacek: I thought the same. However this is often used to stop several downstream signals at once. Perhaps what I can do is making the signals infinite by default, and then introduce CloseableSignal that will have functionality of signlling `close` to all its signals below which will be approximately defined internally as Singal[A] => Signal[Option[A]] ....

src/main/scala/fs2/async/mutable/Signal.scala
pchlupacek: We had never used changed and changes. perhaps goot time to move them away.

src/main/scala/fs2/async/package.scala
pchlupacek: Ok to me this was just  a syntax sugar to `Signal[Option[A]].pipe(collect{case Some(a) => a})` put perhaps with pipe on signal now this has no justification.

src/main/scala/fs2/async/AsyncExt.scala
pchiusano: K

src/main/scala/fs2/async/immutable/Signal.scala
pchiusano: Re stopping several signals at once, not sure what you mean. If signals are always infinite, you can still use them to control termination of several streams, right? Maybe I need to see a more concrete use case.

src/main/scala/fs2/async/package.scala
pchiusano: I'm not sold on having pipe on Signal. It is weird to have a Signal whose continuous portion isn't actually continuous. It also raises lots of questions... What is the value of a signal when the output is dropped? The previous output? Block until the next output?

I no longer like idea of unset signals. If that concept is useful, it seems like it belongs as a separate type I think.

If you just need to pass around a pair of streams that are formed from some signals, that is fine, but I like keeping signal to have a clear meaning.

src/main/scala/fs2/async/AsyncExt.scala
pchiusano: As a general comment - since AsyncExt is something we hope others will implement for their own Task-like types, we really want it to have the absolute minimum of stuff we need, not every piece of information that might be useful. We should try to assume as little as possible about the implementations.

src/main/scala/fs2/async/immutable/Signal.scala
pchlupacek:
```scala
 val sig:Signal[A] = ???

val p1 = sig.discrete  
val p2 = sig.discrete 
...
val pN = sig.discrete
 
// run p1 .. pN somehow (think of it as signals for web client for example, that emits once something change) 

//kill them all at once
sig.close.run.


```

src/main/scala/fs2/async/package.scala
pchlupacek: To me `pipe` was kind of shortcut like doing pipe on every discrete/continuos stream than anything else. But I hear you on continuos signal that it may sound kind of strange, but is nothing else than sig.continuos.pipe(p1)

Unset signals are good for coordination (i.e. don't run anything until I say so, but be ready). Perhaps is ok to have them as Option[A] signals, as you may as well resume them.

src/main/scala/fs2/async/immutable/Signal.scala
pchiusano: I see. I'd just use a separate Signal[Bool] for this, and set it to false. Use that to interrupt all the consumers.

src/main/scala/fs2/async/package.scala
pchiusano: As soon as you apply a pipe, though, you are kind of just converting the signal to a pair of streams. Maybe that is ok but it seems easy to misuse.

Again for coordination like that I'd just use a separate Boolean signal, and have a combinator that is sort of the opposite of interrupt ("gated"?) that starts pulling from another stream only when the boolean switches to true.

pchlupacek: +1 agreed.

pchlupacek:
@pchiusano modified signal for changes you suggested. I have added teh Queue and have some short questions before implementing this 

- What we will use instead of sinks? 
- Not sure on `dequeueBatch` and `dequeueAvailable` now with chunks do they still have their role. Likely yes, but want to double check .

I was also curious, if we shouldn't allow for custom queueing strategies, i.e. priority queueing. Looks like it would be just additional parameter to constructor?

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,109 @@
...
+   * Provides a process that dequeues in chunks.  Whenever *n* elements
+   * are available in the queue, `min(n, limit)` elements will be dequeud
+   * and produced as a single `Seq`.  Note that this naturally does not
+   * *guarantee* that `limit` items are returned in every chunk.  If only
+   * one element is available, that one element will be returned in its own
+   * sequence.  This method basically just allows a consumer to "catch up"
+   * to a rapidly filling queue in the case where some sort of batching logic
+   * is applicable.
+   */
+  // todo: likely we don't need it as we have Chunks now?
```
pchiusano: Yeah, this is not needed. The consumer can await whole chunks and catch up that way. Let's kill it.

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Equivalent to dequeueBatch with an infinite limit.  Only use this
+   * method if your underlying algebra (`A`) has some sort of constant
+   * time "natural batching"!  If processing a chunk of size n is linearly
+   * more expensive than processing a chunk of size 1, you should always
+   * use dequeueBatch with some small limit, otherwise you will disrupt
+   * fairness in the nondeterministic merge combinators.
+   */
+  // todo: likely we don't need it as we have Chunks now?
+  def dequeueAvailable: Stream[F, Seq[A]]
```
pchiusano: same deal, not needed anymore

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Closes this queue. This halts the `enqueue` `Sink` and
+   * `dequeue` `Process` after any already-queued elements are
+   * drained.
+   *
+   * After this any enqueue will fail with `Terminated(End)`,
+   * and the enqueue `Sink` will terminate with `End`.
+   */
+  // todo: Not sure on this. Should we get close off the Queue like signal? And leave it to external interruption?
```
pchiusano: I am kind of in favor of just getting rid of `close`. Use a `Queue[F,Option[A]]` or a separate `Signal[F,Boolean]` to control termination. (And you could even create a new type, `CloseableQueue`, that is implemented in terms of this)

pchiusano:
@pchlupacek 

Okay, left a couple comments. 

> What we will use instead of sinks?

Right now I've defined `Sink[F,A]` as just a `Stream[F,A] => Stream[F,Unit]`. It's very expressive, but not quite sufficient for stuff like `observe`, or anything where you are splitting off some values in the middle of a pipeline, then continuing processing.

I'm considering introducing a combinator, `diamond`, with the signature:

```Scala
// method on Stream[F,A]
def diamond[B,C,D](f: Stream[F,A] => Stream[F,B], g: Stream[F,A] => Stream[F,C])(t: (Stream[F,B], Stream[F,C]) => Stream[F,D]): Stream[F,D]
```

But I don't like that it needs a queue internally (when `f` reads a value, it gets enqueued for `g`) and is easy to mess up (if the two branches have different rates of processing).

> Not sure on dequeueBatch and dequeueAvailable now with chunks do they still have their role. Likely yes, but want to double check .

No I think we can kill these. Since you can now `await` whole chunks, that gives you `dequeueAvailable`, and `dequeueLimit` is superceded by `awaitLimit`. I'd add an `enqueueChunk` function though.

> I was also curious, if we shouldn't allow for custom queueing strategies, i.e. priority queueing. Looks like it would be just additional parameter to constructor?

seems reasonable, but my feeling is let's not worry about it right now. Something to experiment with for a future release.

src/main/scala/fs2/async/mutable/Queue.scala
pchlupacek: +1 to `CloseableQueue`. However Queue[F,Option[A]] won't work I guess (you will need somehow repeatedly to push None to Queue, or hardcode this to Queue). I'll get an that once work on Queue will be done.

pchlupacek:
@pchiusano we have to be I think very careful on including queue in any combinators. The reason is back-pressure, as queue broke the boundaries, and always as such we need somehow to express bounds of such dependent combinator, which makes signatures and code complicated imho. 

On the sink I am just thinking, if we have effect full pipes now, why we wont just cancel sink and have it as effect full pipe instead? I mean if `Process1` is now `Stream[F,A] => Stream[F,B]` I just would create enqueue as `Stream[F,A] => Stream[F,A]`, perhaps `Stream[F,A] => Stream[F,(A,Int)]` that returns size of the queue after enqueue as well? 

That will I guess drop need fro channel and sink and would allow finally first class composition with P1s?

pchiusano:
Yeah, agree, it is complicated, there are a lot of options, and I'm kinda
hesitant to add something like that. We do need some story for sending the
same stream to multiple sinks / channels.

I like your idea about sinks, they are basically the `observe` function.
Along these lines maybe a `Channel` could be a `Stream[F,A] =>
Stream[F,(A,B)]`, so the input is still available for subsequent
transforms? That seems a little kludgy...

pchlupacek:
I think we need name for this effect full transducer. I think we shouldn't call it anymore Channel or Sink. And I would suggest it has the simple signature `Stream[F,A] => Stream[F,B]`. Then we can introduce various combinators in different situation, likewise your lifted channel signature... What do you think? In fact even in our current code base this is most often what we do (Process[F,A] => Process[F,B]) so I think we are on right path there.  Idea is that, instead of having these `to`, `though`, `pipeIn`, `mapout` and so on, we will have single i.e. `via` and then we will just apply combinators on this transducer?

Perhaps new issue for this ?

src/main/scala/fs2/async/mutable/Queue.scala
pchiusano: I mean that the consumer side can simply stop pulling from the queue after receiving a `None`, so there's no need to 'close' the queue necessarily.

pchiusano:
I'm not sure what to call a `Stream[F,A] => Stream[F,B]`. `Transducer1`? Honestly, I think it could just be more confusing to have a separate name for that, given that we already have `Process1`, `Tee`, and `Wye`. If we have to pick a name, I'd say `Transducer1`.

If you want you can merge this, and we can figure out good combinators on `Transducer1` as a separate issue. Can you just address my line notes before merging?

I do think we will need something to control queuing strategy in various combinators.

pchlupacek:
@pchiusano Yeah. I am just working on queue (about halfway through it and then will do PR. 
I see `Transducer1` like effect-full transducer essentially replacing channel and sink. But lets discuss that separately. I will make queue w/o `enqueue` and add that laters. 

Over weekend shall do final PR ok?

pchiusano:
sure

pchlupacek:
@pchiusano finished the `Queue` implementation. Unfortunately in turns out the queue is still complex beast. Was hoping for much simpler implementation. Can you please review. As discussed before this is completely untested

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,282 @@
...
+              // Signals size change of queue, if that has changed
+              def signalSize(s:State[F,A], ns:State[F,A]) : F[Unit] = {
+                if (s.queue.size != ns.queue.size) szSignal.set(ns.queue.size)
+                else F.pure(())
+              }
+
+              // enqueueAll implementation
+              // tries to satisfy or waiting dequeuers on this enqueue and/or attaches the remainder to the queue
+              // also updates size of the queue.
+              def _enqueueAll(xa: Seq[A]): F[Int] = {
```
pchiusano: I'd just implement this by repeatedly doing enqueueOne. The user can always let A be some chunk type if they want to cut down on some of the overhead.

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,282 @@
...
+                      }
+                    }
+                  }
+                }
+              }
+
+              // implementation of dequeueBatch
+              // repeatedly dequeue from the queue, while completing any waiting enqueue on bounded queue
+              // and taking the chunks up to `limit` that are available in queue
+              def _dequeueBatch(limit: Int): Stream[F, A] = Stream.eval {
```
pchiusano: I don't think this function is needed as we discussed before. The consumer of the queue can just use awaitN, awaitLimit, etc to achieve the same thing. No need to make the implementation more complicated than it needs to be.

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,282 @@
...
+  def full: Stream[F,immutable.Signal[F,Boolean]]
+
+
+
+
+}
+
+object Queue {
+
+    private type DeQueue[F[_],A] = (Int,Seq[A] => F[Unit]) // Int represent preferred chunk size
```
pchiusano: If you just make this A => F[Unit] and likewise for enqueue, it seems like it could simplify your implementation a lot.

pchiusano:
Cool. I left some comments. The batching stuff seems like it adds a lot of complexity, and we don't even know how much of a performance difference it makes without benchmarking. I'd rather have a simple more obviously correct implementation. Then we can write some benchmarks and figure out what is worth optimizing. Also performance improvements are something we can do anytime, after this release.

pchlupacek:
@pchiusano thanks. You absolutely right on batching. I think our answer would be Queue[Seq[A]] instead of Queue[A]. I still see that some logic there may bring fine.grained performance, but as always most of it may be done outside the queue, leave queue more solid.

pchiusano:
@pchlupacek when do you think you'l be able to finish simplifying the implementation (removing the batching stuff?). I'd like to get this merged so we can convert the project to multimodule and start publishing snapshots.

pchlupacek:
ok will take a look on this today. On the batching stuff I see how that will simplify whole queue. There is I think still reasonable use case for leaving it in, lets say you may need dequeue in chunks of 2 but enque in chunks of whatever size. I will either drop the chunking if I see the benefits to be significant. I think the largest complexity is coming from bound too (you need to memoize enqueuers).  Perhaps if we will fast-fail any enqueuer instead of memoize it, that will help a lot. One can on failure to recover, wait for queue not being full and then enqueue again.  What do you think? 

Suggested signature of enque will be then 
```scala
 // false on queue being full
 def enqueue[A](a:A):Task[Boolean] = ???
```
The problem with that approach is of course the `spam` when there will be multiple enqeueurs competing for the queue.

But today I gonna modify this, ping you so we can decide.

pchlupacek:
@pchiusano Just pushed the change. Please take a look. Queue is now simple, as I like it. Also think that changed behaviour of enqueueOne is in fact correct. 
LMK what do you think.

pchiusano:
K taking a look now

src/main/scala/fs2/pull1.scala
```diff
@@ -51,6 +51,17 @@ private[fs2] trait pull1 {
   def await1Option[F[_],I]: Handle[F,I] => Pull[F,Nothing,Option[Step[I,Handle[F,I]]]] =
     h => h.await1.map(Some(_)) or Pull.pure(None)
 
+  def changing[F[_],I](eqf:(I,I) => Boolean):Handle[F,I] => Pull[F,I,I] = {
```
pchiusano: I wouldn't put this in `pull1`, since it doesn't return a useful result, it just outputs values so it really belongs as a helper function in `process1`. Also, I would implement using `zipWithLast` and `collect`. It will be more efficient and preserves chunkiness.

pchiusano:
Wow, much simpler! I like it. I'm looking at your implementation of enqueueOne now...

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,173 @@
...
+   * Provides a process that dequeue from this queue.
+   * When multiple consumers dequeue from this queue,
+   * they dequeue in first-come, first-serve order.
+   *
+   * Please use `Topic` instead of `Queue` when all subscribers
+   * need to see each value enqueued.
+   *
+   * This process is equivalent to `dequeueBatch(1)`.
+   */
+  def dequeue: Stream[F, A]
```
pchiusano: Should we expose a `dequeueOne: F[A]` operation also? I'd actually call it `dequeue1` (analogous to `await` and `await1`) and maybe rename `enqueueOne` to `enqueue1`.

I suppose the user can also just do `dequeue.take(1)`.

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * Enqueue one element in this `Queue`.
+   *
+   * Evaluates to `false` if the queue is full, indicating the `a` was not queued up.
+   * Evaluates to `true` if the `a` was queued up successfully.
+   *
+   * @param a `A` to enqueue
+   */
+  def enqueueOne(a: A): F[Boolean]
```
pchiusano: I think we need to have two functions:

* `enqueue1(a: A): F[Unit]` - enqueues the `a`, or waits until the queue is empty
* `offerEnqueue1(a: A): F[Boolean]` - enqueues the `a` immediately and returns `true`, or returns `false` immediately if the queue is full

Both semantics are useful and you can't implement them in terms of each other.

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,173 @@
...
+
+
+  /**
+    * Internal state of the queue
+    * @param queue    Queue, expressed as vector for fast cons/uncons from head/tail
+    * @param deq      A list of De-Queuer (filled up when queue is empty)
+    * @tparam F
+    * @tparam A
+    */
+    private case class State[F[_],A](
```
pchiusano: To allow for the blocking `enqueue1` operation I mentioned above, I'd suggest maybe:

```Scala
// Type enforces invariants that: 
// a) Can't be any dequeues waiting unless queue is empty (if queue were nonempty we'd have dispatched those waiting)
// b) Can't be any enqueues waiting unless queue is full
// when adding to a Nonempty, check if this transitions to Full state
trait State[F[_],A]
case class Empty[F[_],A](dequeues: Vector[DeQueue[F,A]]) extends State[F,A]
case class Nonempty[F[_],A](q: Vector[A]) extends State[F,A]
case class Full[F[_],A](q: Vector[A], enqueues: ???) extends State[F,A] // not sure type of `enqueues`
```

pchiusano:
I like how simple the current implementation is, but we do still need the blocking version of enqueue I think. I know this might complicate the implementation some. Can you address this and the other minor stuff I mentioned? Then I'd say let's merge.

src/main/scala/fs2/async/mutable/Queue.scala
pchlupacek: I think we shouldn't pile-up enqueues. That will just do in fact unbounded queue when you will have many requests (i.e. web site). In some scenarios this may help though (i.e. limited number of enqueues but lot of messages coming in). I will still prefer in that scenario so the pilling-up is
 done by enqueue and not by queue.  I will implement that and you will see what I mean.

src/main/scala/fs2/async/mutable/Queue.scala
pchlupacek: how about simple 
- `enqueue1`
- `offer1`
???

src/main/scala/fs2/pull1.scala
pchlupacek: okay, but what is the story on pipe? aka `zipWithPrevious()() pipe collect()` ???

pchlupacek:
@pchiusano just pushed the changes. look at `enqueue1` what do you think?. I think this is very simple, unfair strategy. Perhaps we may introduce _fairness_ later, but I like to keep that out of state of queue, as this gives imho better flexibility to queue and simplifies implementation. Thoughts?

I also tried to use state algebra, and so far it seems to me a bit of overkill. But I'll think about it more.

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * Offers one element in this `Queue`.
+   *
+   * Evaluates to `false` if the queue is full, indicating the `a` was not queued up.
+   * Evaluates to `true` if the `a` was queued up successfully.
+   *
+   * @param a `A` to enqueue
+   */
+  def offer1(a: A): F[Boolean]
```
pchiusano: +1 on this naming :)

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,196 @@
...
+                }
+              }
+
+              def _enqueue1(a:A): F[Unit] = {
+                F.bind(_offer1(a)) {
+                  case true => F.pure(())
+                  case false =>
+                    // this implies bound is configured so we just listen for queue
+                    // to signal if some space is available
+                    F.bind(szSignal.discrete.takeWhile(_ >= bound).run.run) { _ =>
```
pchiusano: Ah, clever but really simple implementation. I like it.

src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -0,0 +1,196 @@
...
+                      s.deq headOption match {
+                        case None =>
+                          F.map(F.setPure(enqRef)(true)){ _ => s.copy(queue = s.queue :+ a ) }
+                        case Some(deq) =>
+                          F.bind(deq(a)){_ =>
+                            F.map(F.setPure(enqRef)(true)) { _ => s }
+                          }
+                      }
+                    }
+                  }){ case Change(s,ns) =>
```
pchiusano: Not a big deal, but it took me a minute to figure out what `s` and `ns` meant. (`ns` is abbreviation for 'new state' I'm guessing?) I might do `{ change => ... signalSize(change.previous, change.now) }` so the reader doesn't have to remember what `s` and `ns` are bound to.

pchiusano:
I think this looks good!

What did you mean re: fairness? Did you just mean that, if the queue is full, there is no guarantee that any particular thread will ever win the race to enqueue an element (whereas if the state kept a queue of enqueuers internally, it could guarantee it)? I think that behavior is okay as long as we document it. Perhaps we can revisit later.

I'm going to go ahead and merge!

pchlupacek:
@pchiusano yes I am a bit worried that lets say you will have 100 enqueue and full queue, every time when queue will become `empty` all will compete at once for queue. Not only that it is not super efficient, but some may never actually get chance to enqueue. 

I have not super intelligent solution to this, only that it is perhaps for further improvements.

src/main/scala/fs2/async/mutable/Queue.scala
pchlupacek: yes, will revisit that when doing tests

pchiusano:
@pchlupacek actually I was pondering this a bit last night and came up with an idea. 

You can actually implement a bounded `Queue[F,A]` with two unbounded queues. One is the regular unbounded queue of elements, the other is a `Queue[F,Unit]` of 'permits', initialized with `n` elements in it (all just `()` of course), where `n` is the desired bound. Now, to enqueue, you dequeue from the 'permits' queue (which may block if it is empty) and then enqueue to the elements queue. To `dequeue`, you dequeue from the elements queue then add an permit to the permits queue. Easy, and obviously correct. It is also fair - the enqueues are queued up in the obvious way, if the queue is currently full, rather than there being a race every time space becomes available.

After being briefly excited about that, I realized the permits queue is functioning as a rather inefficient implementation of a semaphore. Inefficient because what if the bound is like `Long.MaxValue` - we have a permits queue of that length, which is silly. So it would be better to implement `Semaphore` directly, as a new concurrency primitive. I have done that here, drawing inspiration from your queue implementation: https://gist.github.com/pchiusano/4fe1fffcf66eb0ca8fb7

I'm going to prepare a PR adding `Semaphore` and implementing bounded queues in terms of unbounded queues plus a semaphore. The semaphore will be used to track the number of queue slots available.

pchlupacek:
@pchiusano cool. I need to be much faster with that sort of ideas. In fact was thinking about the same over night :-) I like if we build solid base of very low concurrency primitives. The `ref` approach in fact currently gives us so much flexibility....

pchiusano:
@pchlupacek yeah `Ref` is really flexible, that turned out to be a good idea :)

PR #466, opened 2015-10-25
use `Pull.receive` to refactor some `process1` methods

ShengC:
and also add their alias to `Process1Ops`

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -65,15 +65,14 @@ object process1 {
     
   /** Emit the elements of the input `Handle` until the predicate `p` fails, and return the new `Handle`. */      
   def takeWhile[F[_], I](p: I => Boolean)(implicit F: NotNothing[F]): Handle[F, I] => Pull[F, I, Handle[F, I]] = 
-    h => 
-      Pull.await(h) flatMap {
-        case chunk #: h =>
-          chunk.indexWhere(!p(_)) match {
-            case Some(0) => Pull.done
-            case Some(i) => Pull.output(chunk.take(i)) >> Pull.done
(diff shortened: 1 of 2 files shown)


src/main/scala/fs2/Process1.scala
```diff
@@ -65,15 +65,14 @@ object process1 {
...
-      Pull.await(h) flatMap {
-        case chunk #: h =>
-          chunk.indexWhere(!p(_)) match {
-            case Some(0) => Pull.done
-            case Some(i) => Pull.output(chunk.take(i)) >> Pull.done
-            case None    => Pull.output(chunk) >> takeWhile(p).apply(h)
-          }
-      }
+    Pull.receive {
+      case chunk #: h =>
```
pchiusano: For formatting, move the case onto the same line as the receive, unless this would make the line too long.

src/main/scala/fs2/Process1.scala
```diff
@@ -85,14 +84,13 @@ object process1 {
...
-    h => 
-      Pull.await(h) flatMap {
-        case chunk #: h =>
-          chunk.indexWhere(!p(_)) match {     
-            case Some(i) => Pull.output(chunk.drop(i)) >> id.apply(h)            
-            case None    => dropWhile(p).apply(h)
-          }
-      }
+    Pull.receive {
+      case chunk #: h =>
```
pchiusano: Same here

src/main/scala/fs2/Process1.scala
pchiusano: Saves a line and it's less indentation, which I prefer

ShengC:
@pchiusano does the style look ok to you now?

fthomas:
LGTM. Thanks!

PR #467, opened 2015-10-27
add `collectFirst` to `process1`

ShengC:


src/main/scala/fs2/pull1.scala
```diff
@@ -41,6 +41,17 @@ private[fs2] trait pull1 {
   def await1Option[F[_],I]: Handle[F,I] => Pull[F,Nothing,Option[Step[I,Handle[F,I]]]] =
     h => h.await1.map(Some(_)) or Pull.pure(None)
 
+  /** Finds the first element from the input for which the given partial 
+   function `pf` is defined, and applies the partial function to it. */
+  def collectFirst[F[_],I,O](pf: PartialFunction[I,O]): Handle[F,I] => Pull[F,Nothing,Option[O]] =
+    _.await.optional flatMap {
+      case Some(chunk #: h) => chunk.indexWhere(pf.isDefinedAt) match {
+        case Some(i) => Pull.pure(Some(pf(chunk(i))))
```
fthomas: For performance I would add `collectFirst` to Chunk instead of using `indexWhere` and `apply` here.

fthomas:
`collectFirst` was already claimed by @kaiserpelagic here https://github.com/scalaz/scalaz-stream/wiki/FS2-upgrade-coordination. Next time please check that wiki page before porting something to FS2 and update it accordingly. Thanks :-)

fthomas:
On a related note: `collectFirst` nicely demonstrates that the cost of preserving chunkiness is composability. In scalaz-stream `collectFirst(pf)` was defined as `collect(pf).once`. We can't have that here because that would/could apply more elements to `pf` than necessary for `collectFirst`. I know that we can get composability by defining `collectFirst(pf)` as `unchunk.collect(pf).once` but we explicitly don't want this in the library.

A great deal of pleasure while writing some of scalaz-stream's `Process1` functions was that I could just plug them together like lego bricks.

pchlupacek:
@fthomas I didn't go yet into detail on p1s, put you say this won't be anymore possible? 
```scala 

process1.collect { pf }  pipe process1.once 

```

fthomas:
@pchlupacek That still works but it might call `pf` on more than one element (I think). In scalaz-stream your pipeline calls `pf` at most once.

pchlupacek:
@fthomas I see what you mean.Yeah, this doesn't look good and is unexpected. I remember when we had similar issue with map and flatMap (pre Append). @pchiusano from user perspective I am concerned that this unexpected and some combinations of processes may combine weird results and performance. What do you think?

ShengC:
@fthomas - oops just realized that. I will work on `grouped` then :-)

pchlupacek:
@fthomas just thought over that a bit, the chunk combinators aren't lazy? I mean would be asily solvable if we make them lazy and then I doubt we would have any unexpected invocations?

ShengC:
This stems from the fact that Chunk is strict, and all operators will be
boiled down to the operation done at Chunk level. As a result of that,
calling collect(pf).once on the stream will possibly be called for more
than once. I think client should be aware of that though chunk to chunk is
lazy, within the chunk it is not.

pchiusano:
These are good things to bring up. I think we should keep this stuff in mind, but I'd rather see more examples before considering changes to `Chunk`. Let's keep plowing ahead.

Brain dump of some stuff I've considered - 

* Work with unchunked streams.
* Make chunks lazy.
* Accept that there is some tradeoff here of efficiency vs compositionality.
* Try to discover nicer combinators for working with strict chunks.

Working with unchunked streams is a nonstarter. Chunking is an order(s) of magnitude performance win, so we should try to preserve it when reasonable to do so.

Making chunks lazy is possible, but it raises a lot of questions:

* Done naively, it's not stack safe if you build up a large number of composed functions. Adding trampolining is super slow, we might as well unchunk at that point.
* It isn't necessarily faster, and it can be slower if you traverse the same chunk repeatedly. There's also the phenomenon that if you are processing the entire chunk, generating it strictly (even multiple strict passes) might be faster than one fused pass over a lazy collection of some sort.
* What operations can be fused? Consider a `collect` or `filter` followed by `take`. The `take` needs to know how many elements pass the `filter`. What about a `map` followed by a `takeWhile`? Done poorly, we can end up with some rather ad hoc rules about what gets fused that the user isn't likely to have a good model for.
* If I were looking into this, I'd try representing chunks via their unfold, allowing skips. So a `Chunk[A]` is actually a `exists s . S => Step[A,S]`, where `Step` is one of `Emit`, `Skip`, or `Done`. Just about all operations fuse - you end up with a single loop with a state type `S`.
  * With this design, `Chunk` no longer has a known-in-advance size.
  * Maybe it's better to introduce a separate `Chunk.View` type?

So, the design space is pretty big, and every design has different tradeoffs. While I think it would be cool to experiment with this, I don't know if I want this to be a focus of this release. The other thing is that experimenting with different chunk types is something anyone can do - it's just a user code thing. Being given strict chunks as a starting point does not preclude you from converting that to some other chunk type you prefer. Just call `await` to obtain a strict chunk, then do whatever you want from there.

Accepting some loss of compositionality when using strict chunks is a little unsatisfying.

The last option is kind of interesting. I'll just write down some ideas I've toyed with:

When working with streams of strict chunks, you can often regain compositionality by separating the act of _selecting_ a portion of the stream from the act of _operating_ on that part of the stream. So, `collect` is doing both--it's selecting the entire stream, then filtering/mapping it, all in one step. This makes it a little too eager to use for `collectFirst`, unless we don't mind repeating work. However, consider:

```Scala
/** Splits the chunks of the input such that all elements of each `Chunk` have the same associated `Boolean`. So a `Chunk((b1,false), (b2,false), (b3, true))` would be split into two chunks, the first containing `b1` and `b2`, and the next containing `b3`. 
*/
def selectable[A](f: A => (B, Boolean)): Process1[A,(Chunk[B], Boolean)]

/** Emit any chunks with an associated `true` value. */
def select[A]: Process1[(Chunk[A], Boolean), A]
```

Now we can define `collect` using `selectable`:

```Scala
def collectable[A,B](f: PartialFunction[A,B]): A => (Option[B], Boolean) =
  f.lift andThen (o => (o, o.isDefined))

collect(f)(s) == selectable(collectable(f))(s) pipe select mapChunks (Chunk.somes)
collectFirst(f)(s) = selectable(collectable(f))(s).dropUntil(_._2) pipe select mapChunks (Chunk.somes)
```

pchlupacek:
@pchiusano thanks for explanation. Would be nice if we can keep wiki page and reference these as todo work. I agree this point may be post 0.9, but I definitely want to avoid any repeated evaluation like we had before at map/flatMap unless it will be explicitly documented, and event then I believe there will be possibility that you combine lazy and strict combinators with hard-to predict behaviour. Won't have that issue if we define simple and easy to understand laws on strictness and laziness.

pchlupacek:
In fact I just created that page https://github.com/scalaz/scalaz-stream/wiki/Post-0.9-todos

ShengC:
@pchiusano I like the idea of representing a `Chunk.view` via a sealed set of `Step` :+1: . For the workaround of `collect` and `collectFirst`, wouldn't that break the original "Chunkiness" arrangement ? I would also be able to apply `Pull.await1` to implement `collect`, and `collectFirst` could then be implemented based on `collect`.

PR #468, opened 2015-10-27
FS2: Implement process1.scan, process1.scan1, process1.fold, process1.fold1

fiadliel:
process1.fold:
  Folds all inputs using an initial value `z` and supplied binary operator,
  and writes the final result to the output of the supplied `Pull` when the
  stream has no more values.

process1.fold1:
  Folds all inputs using the supplied binary operator, and writes the final
  result to the output of the supplied `Pull` when the stream has no more values.

process1.scan:
   Writes `z` to the output, followed by the result of repeated applications
   of `f` to each input in turn. Works in a chunky fashion, and creates a
  `Chunk.indexedSeq` for each converted chunk.

process1.scan1:
  Writes the result of repeated applications of `f` to each input in turn.
  Works in a chunky fashion, and creates a `Chunk.indexedSeq` for each
  converted chunk.

### DIFF
--- a/src/main/scala/fs2/Chunk.scala
+++ b/src/main/scala/fs2/Chunk.scala
@@ -29,14 +29,9 @@ trait Chunk[+A] { self =>
     iterator.map(f).copyToBuffer(buf)
     Chunk.indexedSeq(buf)
   }
-  /**
-   * `[[scanLeft]]` is a helper for process1.scan; it differs
-   * from standard collection behavior by not including
-   * the initial value in the output.
-   */
   def scanLeft[B](z: B)(f: (B, A) => B): Chunk[B] = {
-    val buf = new collection.mutable.ArrayBuffer[B](size)
-    iterator.scanLeft(z)(f).drop(1).copyToBuffer(buf)
+    val buf = new collection.mutable.ArrayBuffer[B](size + 1)
+    iterator.scanLeft(z)(f).copyToBuffer(buf)
     Chunk.indexedSeq(buf)
   }
   def zipWithIndex: Chunk[(A, Int)] = {
--- a/src/main/scala/fs2/Chunk.scala
+++ b/src/main/scala/fs2/Chunk.scala
@@ -29,6 +29,16 @@ trait Chunk[+A] { self =>
     iterator.map(f).copyToBuffer(buf)
     Chunk.indexedSeq(buf)
   }
+  /**
+   * `[[scanLeft]]` is a helper for process1.scan; it differs
+   * from standard collection behavior by not including
+   * the initial value in the output.
+   */
+  def scanLeft[B](z: B)(f: (B, A) => B): Chunk[B] = {
+    val buf = new collection.mutable.ArrayBuffer[B](size)
+    iterator.scanLeft(z)(f).drop(1).copyToBuffer(buf)
+    Chunk.indexedSeq(buf)
+  }
   def zipWithIndex: Chunk[(A, Int)] = {
     val buf = new collection.mutable.ArrayBuffer[(A, Int)](size)
     iterator.zipWithIndex.copyToBuffer(buf)
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -40,6 +40,23 @@ object process1 {
   def filter[F[_],I](f: I => Boolean)(implicit F: NotNothing[F]): Handle[F,I] => Pull[F,I,Handle[F,I]] =
     mapChunks(_ filter f)
 
+  /**
+   * Folds all inputs using an initial value `z` and supplied binary operator, and writes the final
+   * result to the output of the supplied `Pull` when the stream has no more values.
+   */
+  def fold[F[_],I,O](z: O)(f: (O, I) => O)(implicit F: NotNothing[F]): Handle[F,I] => Pull[F,O,Handle[F,I]] =
+   h => h.await.optional flatMap {
+     case Some(c #: h) => fold(c.foldLeft(z)(f))(f).apply(h)
+     case None => Pull.output1(z) >> Pull.done
+   }
+
+  /**
+   * Folds all inputs using the supplied binary operator, and writes the final result to the output of
+   * the supplied `Pull` when the stream has no more values.
+   */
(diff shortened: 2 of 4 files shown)


fiadliel:
Note: this has the same "issue" as https://github.com/scalaz/scalaz-stream/pull/467 with respect to the folds/scans working eagerly within a chunk.

pchiusano:
LGTM.

src/main/scala/fs2/Chunk.scala
```diff
@@ -29,6 +29,16 @@ trait Chunk[+A] { self =>
     iterator.map(f).copyToBuffer(buf)
     Chunk.indexedSeq(buf)
   }
+  /**
+   * `[[scanLeft]]` is a helper for process1.scan; it differs
+   * from standard collection behavior by not including
+   * the initial value in the output.
```
pchiusano: Hmm, actually, is this behavior consistent with the current `process1.scan`? We should keep the behavior the same. You can always `drop(1)` the result if you don't want the first element.

fiadliel: Dropping the value here is so that we don't repeat values across chunk boundaries; the first value being emitted is special-cased in the process1 implementations.

I checked this already, but the details are fiddly, so I'll double-check in a while and make sure that they match.

fiadliel: Also: I don't think that drop is not necessarily efficient for every chunk type, so since I don't need to rely on it...

fiadliel: If we implement drop in a similar manner to that in class Doubles, I wouldn't be worried, I just don't want to double memory allocations here.

pchiusano: It is weird to have a `scanLeft` function on any type that doesn't start with the seed value, so I'd say include it regardless. Relying on `drop` being fast for `Chunk` is totally fine.

fiadliel: @pchiusano Behavior is consistent, as shown below:

* scan1

```
scala> Process.empty[Task, Int].scan1(_ + _).runLog.run
res7: IndexedSeq[Int] = Vector()

scala> Process(1).scan1(_ + _).toSource.runLog.run
res8: IndexedSeq[Int] = Vector(1)

scala> Process(1, 2).scan1(_ + _).toSource.runLog.run
res9: IndexedSeq[Int] = Vector(1, 3)

scala> Process(1, 2, 3).scan1(_ + _).toSource.runLog.run
res10: IndexedSeq[Int] = Vector(1, 3, 6)
```

```
scala> Stream.empty[Task, Int].scan1(_ + _).runLog.run.run
res2: Vector[Int] = Vector()

scala> (Stream(1): Stream[Task, Int]).scan1(_ + _).runLog.run.run
res3: Vector[Int] = Vector(1)

scala> (Stream(1, 2): Stream[Task, Int]).scan1(_ + _).runLog.run.run
res4: Vector[Int] = Vector(1, 3)

scala> (Stream(1, 2, 3): Stream[Task, Int]).scan1(_ + _).runLog.run.run
res5: Vector[Int] = Vector(1, 3, 6)
```

* scan

```
scala> Process.empty[Task, Int].scan(0)(_ + _).runLog.run
res11: IndexedSeq[Int] = Vector(0)

scala> Process(1).scan(0)(_ + _).toSource.runLog.run
res12: IndexedSeq[Int] = Vector(0, 1)

scala> Process(1, 2).scan(0)(_ + _).toSource.runLog.run
res13: IndexedSeq[Int] = Vector(0, 1, 3)

scala> Process(1, 2, 3).scan(0)(_ + _).toSource.runLog.run
res14: IndexedSeq[Int] = Vector(0, 1, 3, 6)
```

```
scala> Stream.empty[Task, Int].scan(0)(_ + _).runLog.run.run
res6: Vector[Int] = Vector(0)

scala> (Stream(1): Stream[Task, Int]).scan(0)(_ + _).runLog.run.run
res7: Vector[Int] = Vector(0, 1)

scala> (Stream(1, 2): Stream[Task, Int]).scan(0)(_ + _).runLog.run.run
res8: Vector[Int] = Vector(0, 1, 3)

scala> (Stream(1, 2, 3): Stream[Task, Int]).scan(0)(_ + _).runLog.run.run
res9: Vector[Int] = Vector(0, 1, 3, 6)
```

Most importantly with fold/fold1 is what happens with the empty case:

* fold1

```
scala> Process.empty[Task, Int].fold1(_ + _).runLog.run
res15: IndexedSeq[Int] = Vector()

scala> Process(1).fold1(_ + _).toSource.runLog.run
res16: IndexedSeq[Int] = Vector(1)

scala> Process(1, 2).fold1(_ + _).toSource.runLog.run
res17: IndexedSeq[Int] = Vector(3)
```

```
scala> Stream.empty[Task, Int].fold1(_ + _).runLog.run.run
res10: Vector[Int] = Vector()

scala> (Stream(1): Stream[Task, Int]).fold1(_ + _).runLog.run.run
res11: Vector[Int] = Vector(1)

scala> (Stream(1, 2): Stream[Task, Int]).fold1(_ + _).runLog.run.run
res12: Vector[Int] = Vector(3)
```

* fold

```
scala> Process.empty[Task, Int].fold(0)(_ + _).toSource.run
res19: IndexedSeq[Int] = Vector(0)

scala> Process(1).fold(0)(_ + _).toSource.runLog.run
res20: IndexedSeq[Int] = Vector(1)

scala> Process(1, 2).fold(0)(_ + _).toSource.runLog.run
res21: IndexedSeq[Int] = Vector(3)
```

```
scala> Stream.empty[Task, Int].fold(0)(_ + _).runLog.run.run
res13: Vector[Int] = Vector(0)

scala> (Stream(1): Stream[Task, Int]).fold(0)(_ + _).runLog.run.run
res14: Vector[Int] = Vector(1)

scala> (Stream(1, 2): Stream[Task, Int]).fold(0)(_ + _).runLog.run.run
res15: Vector[Int] = Vector(3)
```

fiadliel:
Updated, moving `drop` from `Chunk.scanLeft` into `process1.scan`.

pchiusano:
Great. Merging.

PR #471, opened 2015-10-28
FS2: Add process1.awaitOption

fiadliel:
This outputs the first value received from the input, or None.

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -16,6 +16,10 @@ object process1 {
 
   // nb: methods are in alphabetical order
 
+  /** Outputs the first value received from the input, or None if no value is provided. */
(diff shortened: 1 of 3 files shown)


pchiusano:
I'm not sure how useful his is, but I'm fine with including it, however please just define in terms of pull1.await1Option.

Also it should be renamed to await1Option.

pchiusano:
Basically any combinator that produces just a single output is more useful returning the output as the result of the Pull, making it easy to use in other pulls.

fiadliel:
Done. I will try and keep that in mind for any future changes...

I also think this is of limited use (it seems to have been mostly used for internal purposes in scalaz-stream), so I wonder if we should just leave it out, process1 is already fairly kitchen-sink-y.

pchiusano:
Yeah, actually you've convinced me, lets just leave it out for now. We can always revisit later.

PR #472, opened 2015-10-29
FS2: Add process1.mapAccumulate

fiadliel:
@pchiusano This isn't ready to be committed (needs a bit more review, tests, and comments), but I wanted to check that you'd be happy with the general approach.

I think it can probably be used by a lot of other combinators, e.g. a sample implementation of sliding:
```scala
def sliding[F[_],I](n: Int)(implicit F: NotNothing[F]): Handle[F,I] => Pull[F,IndexedSeq[I],Handle[F,I]] =
  collectAccumulate(new collection.mutable.ArrayBuffer[I](n))((arr, i: I) => {
    arr += i
    if (arr.size >= n) {
      val newArr = new collection.mutable.ArrayBuffer[I](n)
      arr.tail.copyToBuffer(newArr)
      (newArr, Some(arr))
    } else (arr, None)
  })(s => None)
```

Also I'm not sure about the name :smile:

### DIFF
--- a/src/main/scala/fs2/Chunk.scala
+++ b/src/main/scala/fs2/Chunk.scala
@@ -34,6 +34,16 @@ trait Chunk[+A] { self =>
     iterator.map(f).copyToBuffer(buf)
     Chunk.indexedSeq(buf)
   }
+  def mapAccumulate[S,B](s0: S)(f: (S,A) => (S,B)): (S,Chunk[(S,B)]) = {
+    val buf = new collection.mutable.ArrayBuffer[(S,B)](size)
+    var s = s0
+    for { c <- iterator } {
+      val r = f(s, c)
+      buf += r
+      s = r._1
+    }
+    (s, Chunk.indexedSeq(buf))
+  }
   def scanLeft[B](z: B)(f: (B, A) => B): Chunk[B] = {
     val buf = new collection.mutable.ArrayBuffer[B](size + 1)
     iterator.scanLeft(z)(f).copyToBuffer(buf)
--- a/src/main/scala/fs2/Chunk.scala
+++ b/src/main/scala/fs2/Chunk.scala
@@ -34,13 +34,13 @@ trait Chunk[+A] { self =>
     iterator.map(f).copyToBuffer(buf)
     Chunk.indexedSeq(buf)
   }
-  def mapAccumulate[S,B](s0: S)(f: (S,A) => (S,B)): (S,Chunk[(S,B)]) = {
-    val buf = new collection.mutable.ArrayBuffer[(S,B)](size)
+  def mapAccumulate[S,B](s0: S)(f: (S,A) => (S,B)): (S,Chunk[B]) = {
+    val buf = new collection.mutable.ArrayBuffer[B](size)
     var s = s0
     for { c <- iterator } {
-      val r = f(s, c)
-      buf += r
-      s = r._1
+      val (newS, newC) = f(s, c)
+      buf += newC
+      s = newS
     }
     (s, Chunk.indexedSeq(buf))
   }
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -87,10 +87,14 @@ object process1 {
     */
   def mapAccumulate[F[_],S,I,O](init: S)(f: (S,I) => (S,O)): Stream[F,I] => Stream[F,(S,O)] =
     _ pull { _.await.flatMap { case chunk #: h =>
-      val (s, o) = chunk.mapAccumulate(init)(f)
-      Pull.output(o) >> _mapAccumulate0(s)(f)(h)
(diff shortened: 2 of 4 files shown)


pchiusano:
Is the intention that returning None terminates?

pchiusano:
Oh nm I see, it is a stateful map and/or filter. Seems reasonable. Needs a better name.

fiadliel:
@pchiusano PR updated with different "better" name, comments, and tests.
"There are only two hard things in Computer Science".

fiadliel:
@pchiusano Changed this pull request to implement mapAccumulate instead (doesn't explicitly deal with end-of stream, and emits (S, O).

I think if Process1 can be composed, trying to put everything (including filtering, end-of-stream detection) into a single combinator isn't so important.

pchiusano:
Looks good. I can merge. One thing though -

I like the example, but let's make the syntax nicer by exposing a `toList` / `toVector` on `Stream[Pure,A]`:

```Scala
implicit class PureStreamSyntax[+A](s: Stream[Pure,A]) { ... }
```

If you stick this in `StreamDerived.scala` which gets mixed into the `Stream` companion object. 

If you want you can do this as a separate PR.

src/main/scala/fs2/Process1.scala
```diff
@@ -77,6 +77,24 @@ object process1 {
...
+    *
+    * @example {{{
+    * scala> Stream("Hello", "World")
+    *      |   .mapAccumulate(0)((l, s) => (l + s.length, s.head))
+    *      |   .covary[util.Task].runLog.run.run
+    * res0: Vector[(Int, Char)] = Vector((5,H), (10,W))
+    * }}}
+    */
+  def mapAccumulate[F[_],S,I,O](init: S)(f: (S,I) => (S,O)): Stream[F,I] => Stream[F,(S,O)] =
+    _ pull { _.await1.flatMap { case chunk #: h =>
```
pchiusano: Actually, I just noticed you are using `await1`. Let's preserve chunkiness. Add a `mapAccumulate` to `Chunk`:

`def mapAccumulate[S](s0: S)(f: (S,A) => (S,B): (S,Chunk[B])`

Maybe a `filterMapAccumulate` that returns an `Option[B]` on each step also, combining filtering.

Then define `process1.mapAccumulate` in terms of this.

fiadliel:
For a `mapAccumulate` that emits values of type `(S,O)`, it's questionable to me whether a filtered view that only outputs _some_ `(S,O)` is actually that useful (you will probably do a `map`, `filter` or `collect` on the data downstream anyway, and only getting some state values emitted doesn't seem great). 

A `Stream[F, O]` is a better target for a filtered view, with something like
```scala
def blah[F[_],S,I,O](s0: S)(f: (S,I) => (S,Option[O])): Stream[F,I] => Stream[F,O]
```
It's a different enough signature that I don't think we can call it `mapAccumulate`. Perhaps my previous idea, `mapFilterStateful` - though the name is quite a mouthful.

I'll work on this for now, but leave out the Option support in this PR.

pchiusano:
Oh, when I said "return an option on each step" I meant `(S,A) => (S,Option[B])`, not `(S,A) => Option[(S,B)]`, which would be strange like you say.

But either way, yeah, let's leave that out of this PR. I'm not sure it is a good idea.

src/main/scala/fs2/Process1.scala
```diff
@@ -77,6 +77,27 @@ object process1 {
     _ repeatPull { _.await.flatMap { case chunk #: h => Pull.output(f(chunk)) as h }}
 
   /**
+    * Maps a running total according to `S` and the input with the function `f`.
+    *
+    * @example {{{
+    * scala> Stream("Hello", "World")
+    *      |   .mapAccumulate(0)((l, s) => (l + s.length, s.head))
+    *      |   .covary[util.Task].runLog.run.run
```
pchiusano: you can update this now to use the `toList` or `toVector`

fiadliel:
@pchiusano Done.

src/main/scala/fs2/Chunk.scala
```diff
@@ -34,6 +34,16 @@ trait Chunk[+A] { self =>
     iterator.map(f).copyToBuffer(buf)
     Chunk.indexedSeq(buf)
   }
+  def mapAccumulate[S,B](s0: S)(f: (S,A) => (S,B)): (S,Chunk[(S,B)]) = {
```
pchiusano: Return type should just be `(S, Chunk[B])`. If you want your `B` to incorporate the `S`, that is the job of the `f` function. (That is, `B` could be `(S,Blah)`.) So when you use this to implement `mapAccumulate` on streams, you'll have to do exactly that, since we want each value tagged with the current `S`.

fiadliel: Sounds more flexible alright, PR updated.

pchiusano:
Looks good. Merging.

PR #473, opened 2015-11-01
FS2: Add process1.sum

fiadliel:
Implements process1.sum, which writes the sum of all input numeric values, or numeric (monoidal) zero if no input values are supplied.

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -122,6 +122,10 @@ object process1 {
   def scan1[F[_],I](f: (I, I) => I): Stream[F,I] => Stream[F,I] =
     _ pull { Pull.receive1 { case o #: h => _scan0(o)(f)(h) }}
 
+  /** Writes the sum of all input elements, or zero if the input is empty. */
+  def sum[F[_],I](implicit ev: Numeric[I]): Stream[F,I] => Stream[F,I] =
+    fold(ev.zero)(ev.plus)
+
   /** Emit the first `n` elements of the input `Handle` and return the new `Handle`. */
   def take[F[_],I](n: Long): Stream[F,I] => Stream[F,I] =
     _ pull Pull.take(n)
--- a/src/main/scala/fs2/Process1Ops.scala
+++ b/src/main/scala/fs2/Process1Ops.scala
@@ -48,6 +48,9 @@ trait Process1Ops[+F[_],+O] { self: Stream[F,O] =>
   /** Alias for `self pipe [[process1.scan1]](f)`. */
   def scan1[O2 >: O](f: (O2, O2) => O2): Stream[F,O2] = self pipe process1.scan1(f)
 
+  /** Alias for `self pipe [[process1.sum]](f)`. */
+  def sum[O2 >: O : Numeric]: Stream[F,O2] = self pipe process1.sum
+
   /** Alias for `self pipe [[process1.take]](n)`. */
   def take(n: Long): Stream[F,O] = self pipe process1.take(n)
 
--- a/src/test/scala/fs2/Process1Spec.scala
+++ b/src/test/scala/fs2/Process1Spec.scala
@@ -117,6 +117,14 @@ object Process1Spec extends Properties("process1") {
     s.get.covary[Task].through(prefetch) ==? run(s.get)
   }
 
+  property("sum") = forAll { (s: PureStream[Int]) =>
+    s.get.sum ==? Vector(run(s.get).sum)
+  }
+
+  property("sum (2)") = forAll { (s: PureStream[Double]) =>
+    s.get.sum ==? Vector(run(s.get).sum)
+  }
+
   property("take") = forAll { (s: PureStream[Int], negate: Boolean, n0: SmallNonnegative) =>
     val n = if (negate) -n0.get else n0.get
     s.get.take(n) ==? run(s.get).take(n)


src/main/scala/fs2/Process1.scala
```diff
@@ -104,6 +104,12 @@ object process1 {
   def scan1[F[_],I](f: (I, I) => I)(implicit F: NotNothing[F]): Handle[F,I] => Pull[F,I,Handle[F,I]] =
     Pull.receive1 { case o #: h => scan(o)(f).apply(h) }
 
+  /** Writes the sum of all input elements, or zero if the input is empty. */
+  def sum[F[_],I:Numeric](implicit F: NotNothing[F]): Handle[F,I] => Pull[F,I,Handle[F,I]] = {
```
jedws: wouldn't this be better simply taking a Monoid directly?

fiadliel: Which one?

jedws: fs.util.Monoid

fiadliel: I don't think util.Monoid exists (yet). Also, AFAIK, these are supposed to be internal typeclasses, not to be part of the external API.

I appreciate that this uses up the method "sum" that could be used with typeclasses from cats or scalaz in bridging APIs; but this is (approximately) the signature from <= 0.8...

jedws: oh, there were a whole bunch of them, I hadn't seen the specialised sum for Numeric :-( was thinking more of `fold`, `foldMonoid` and `foldMap`

fs will definitely need Monoid.

pchiusano: I don't really want to add `Monoid` to FS2 at this time. The policy is -
the util pkg should be the absolute bare minimum needed to make the library
useful. Monad, etc needs to be there because otherwise we'd have to pick a
specific effect type, which would defeat the whole purpose of the library,
which is to be able to work with effectful streams of any sort.

If there were a Monoid type in FS2, users would have to create instances.
But by the time you give an FS2 Monoid instance you could have just as
easily (with less code, actually) just called fold, scan, etc. It's just
not a big enough win to justify adding to this library.

Stuff that mentions a `Monoid` type belongs in a separate library that
depends on FS2 and whatever other library comes with a `Monoid` type.

jedws: we use this a lot:

    ps.foldMap(functionToMyType)

versus

    ps.map(functionToMyType).fold(Monoid[MyType].zero)(Monoid[MyType].append)

it is very easy to provide an implicit FooMonoidToFs2Monoid to convert between an existing library of implementations, rather than needing to add the syntax to add `foldMap` or `fold*Monoid` variants per external library.

pchiusano: These are reasonable points. I'll give it some more thought, but am honestly still leaning towards no. I just consider providing `Monoid` to be out of scope for this library, even if it adds a bit of convenience.

Also, you don't need a set of `foldMap` and `fold*Monoid` variants for each library that defines a `Monoid` type. Just pick one `Monoid` type, and define implicit conversions from `MonoidFromLibA` to whatever `Monoid` type you deem 'canonical' for this purpose. Basically, you the user can do exactly what you want FS2 to do. Right?

```scala
implicit class MonoidalFns[F[_],A](s: Stream[F,A]) {
  def foldMap[B:Monoidal](f: A => B): Stream[F,B]
}
object Monoidal {
  implicit def officalMonoidal[A:TehOfficialMonoid]: Monoidal[A] = ...
  implicit def anotherLibMonoidal[A:OtherLibMonoid]: Monoidal[A] = ...
}
trait Monoidal[A] { def monoid: TehOfficialMonoid[A] }
```

jedws: as a user I wouldn't do that though, I'd basically only define and use my own `MonoidalFns`.

Think about it though, I'm not sure what you're planning with plugging in impls for the `fs2.util.*` I can't really think of a way that doesn't involve orphans. I was assuming that Monoid would use exactly the same technique, but need to think further as to how you might achieve it without magic.

pchiusano: That's fine. I was just pointing out that if you do have multiple monoid
libraries you are working with, you can overload functions to work with
any/all of them. If you're only working with one that is fine too. Either
way you are empowered as the user to add new functions that aren't provided
already by FS2, and you don't have to duplicate code. FS2 won't provide
every possible useful function and that is okay. :)

I don't follow what you are asking about in you second paragraph. Just give
the instances like you would for anything else? perhaps in the companion
object if it's a type you control or in a standard place if it's for a
third party library. Same sort of idea as what scalaz does, with instances
it provides for its own types and for the standard library.

fiadliel:
@pchiusano I'm thinking of dropping this actually. As you point out, using scan is trivial. Also, I initially thought I should treat `Numeric` specially because it's in the stdlib, but then it would be sad for somebody who wanted to use Spire for sum instead.

pchiusano:
@fiadliel I'd like to keep it, especially since it's already present in 0.8. No one is prevented from implementing their own `sum` function parameterized on another numeric typeclass if they prefer. Can you update your PR so it merges cleanly?

fiadliel:
@pchiusano Updated.

pchiusano:
Great, thanks!

PR #474, opened 2015-11-03
FS2: Add process1.zipWithNext, process1.zipWithPrevious, process1.zipWithPreviousAndNext

kanterov:
Implements `process1.zipWithNext`. It zips the elements of the input with its next element wrapped into `Some`. The last element is zipped with `None`.

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -199,6 +199,20 @@ object process1 {
     } andThen(_.map { case (_, prevNext) => prevNext })
   }
 
+  /**
+    * Zip the elements of the input `Handle` with its previous and next elements wrapped into `Some`, and return the new `Handle`.
+    * The first element is zipped with `None` as the previous element,
+    * the last element is zipped with `None` as the next element.
+    */
+  def zipWithPreviousAndNext[F[_], I]: Stream[F, I] => Stream[F, (Option[I], I, Option[I])] = {
+    (zipWithPrevious[F, I] andThen zipWithNext[F, (Option[I], I)]) andThen {
+      _.map {
+        case ((prev, that), None) => (prev, that, None)
+        case ((prev, that), Some((_, next))) => (prev, that, Some(next))
+      }
+    }
+  }
+
   // stepping a process
 
   def covary[F[_],I,O](p: Process1[I,O]): Stream[F,I] => Stream[F,O] =
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -189,6 +189,16 @@ object process1 {
     _ pull Pull.receive1 { case head #: h => go(head)(h) }
   }
 
+  /**
+    * Zip the elements of the input `Handle` with its previous element wrapped into `Some`, and return the new `Handle`.
+    * The first element is zipped with `None`.
+    */
+  def zipWithPrevious[F[_], I]: Stream[F, I] => Stream[F, (Option[I], I)] = {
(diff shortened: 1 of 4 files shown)


kanterov:
@pchiusano I've rebased and added `zipWithPrevious` and `zipWithPreviousAndNext`. It transforms a stream in a single pass, so implementation isn't straightforward.

pchiusano:
These implementations are pretty complicated. I'd rather not merge as is. Do not worry about trying to fuse everything into a single pass over the chunks. I'd rather address this sort of thing with an improved chunk type than have more complicated code. People should be able to inspect the implementations of these functions and think "oh, that makes sense, and it gives me ideas for how to do some other stuff", not "holy crap, I guess I'll just trust that works".

I am pretty sure these can all be written in terms of `scan`. Also you may find `Stream.terminated` helpful for detecting the end of stream.

kanterov:
I've updated the code. 

I didn't get how `terminated` could help because it works on streams while we need to transform chunks. If we had pure streams it would be possible to write `zipWithNext = xs => xs.zip(xs.drop(1).terminated)`, but it doesn't work smoothly for chunk because the last element from previous input chunk goes to the head of the next output. `zipWithNext` is still isn't super straightforward. It is possible to add `Chunk#zipWithNext` to simplify it a bit, but it wouldn't help a lot because the logic of handling empty chunk to propagate the last element would remain.

pchiusano:
Okay, sorry for the back and forth, but let's revisit this after #472. I think this can be nicely implemented in terms of that.

pchiusano:
Okay, see if you can implement in terms of `mapAccumulate`. For `zipWithPrevious`, it's pretty easy, just let `S` be the `Option[A]` of the previous element. The others will be a little trickier.

kanterov:
@pchiusano please take a look. I've implemented `zipWithPrevious` with `process1.mapAccumulate` as you suggested, and `zipWithNext` with `Chunk#mapAccumulate`, a bit tricky though.

src/main/scala/fs2/Process1.scala
```diff
@@ -172,6 +172,51 @@ object process1 {
...
 
+  /**
+    * Zip the elements of the input `Handle` with its next element wrapped into `Some`, and return the new `Handle`.
+    * The last element is zipped with `None`.
+    */
+  def zipWithNext[F[_], I]: Stream[F, I] => Stream[F, (I, Option[I])] = {
+    def go(last: I): Handle[F, I] => Pull[F, (I, Option[I]), Handle[F, I]] =
+      Pull.receiveOption {
+        case None =>
+          Pull.output1((last, None)) as Handle.empty
```
pchiusano: No need for blankline here, and I'd move onto the same line as `case None` as it's a very short line and pretty easy to understand

src/main/scala/fs2/Process1.scala
```diff
@@ -172,6 +172,51 @@ object process1 {
...
+
+        case Some(chunk #: h) =>
+          val (newLast, zipped) = chunk.mapAccumulate(last) {
+            case (prev, next) => (next, (prev, Some(next)))
+          }
+
```
pchiusano: no blankline here

src/main/scala/fs2/Process1.scala
```diff
@@ -172,6 +172,51 @@ object process1 {
...
+          Pull.output(zipped) >> go(newLast)(h)
+      }
+
```
pchiusano: or here

kanterov:
@pchiusano I've addressed your comments and squashed commits

pchiusano:
Looks good. Merging!

PR #476, opened 2015-11-04
Switch to new and improved representation of `Process1`, `Tee`, and `Wye`

pchiusano:
This PR is built on #475 

Sort of a subtle issue, but with the current representation, you can't invoke a `Stream[F,A] => Stream[F,B]` from within a `Process1`. That's not good for composition. And making `Handle[F,A] => Pull[F,O,R]` the basis for composition doesn't work out well because you can't adjust the output type `O` after the fact. (And changing this would complicate the API and the implementation.)

With the new representation, a `Process1[A,B]` is just a regular function from `Stream[Pure,A] => Stream[Pure,B]`, which can be promoted to a `Stream[F,A] => Stream[F,B]` via `Process1.covary`, or the `pipe` and `pipe2` combinators will automatically do the right thing.

I originally thought this representation would lead to boilerplate, but with the `pull` and `repeatPull` combinators, there's none. The added flexibility means it can pass the input through 20 different pipeline stages and generally build up transducers in a compositional fashion, via transforming streams, and/or using the `Pull` functions directly.

As a bonus, I also got rid of the `NotNothing` stuff. It is no longer needed if the functions are curried the right way.

This PR also moves some functions into `pull1` that previously only existed in `process1`. When you have an already-open stream, these can be useful, and they are also useful when building up a larger `Pull`. The policy I am using is that if a function does not fully consume its input stream(s), consider adding a function to `pull1.scala` or `pull2.scala` that returns the remainder(s). Then have the `process1` function use it. Check out `process1.dropWhile` for an example of this.

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -2,7 +2,7 @@ package fs2
 
 import Stream.Handle
 import Step._
-import fs2.util.{Free,Functor,NotNothing,Sub1}
+import fs2.util.{Free,Functor,Sub1}
 
 object process1 {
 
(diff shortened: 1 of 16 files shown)


pchiusano:
Am merging this. @fiadliel @kanterov could you update your PRs to work with this new representation?

@pchlupacek sorry am behind on responding to your latest comments. Will get on that soon.

/cc @fthomas

pchiusano:
Also, sorry to have shaken things up again, but I believe the representation of `Process1`, `Tee`, and `Wye` should all be stable going forward, so porting the existing functions over should be "easy".

fiadliel:
Looking forward to it, actually. It sounds (I have to look at the implementation) like the composability of this approach is a lot closer to the old Process1.

PR #480, opened 2015-11-09
FS2: Simplify process1.zipWithIndex

kanterov:
Implement `process1.zipWithNext` in terms of `process1.mapAccumulate`, so we don't need `Chunk#zipWithNext`.

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -164,12 +164,9 @@ object process1 {
 
   /** Zip the elements of the input `Handle `with its indices, and return the new `Handle` */
   def zipWithIndex[F[_],I]: Stream[F,I] => Stream[F,(I,Int)] = {
-    def go(n: Int): Handle[F, I] => Pull[F, (I, Int), Handle[F, I]] = {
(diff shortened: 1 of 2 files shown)


src/main/scala/fs2/Process1.scala
```diff
@@ -164,12 +164,11 @@ object process1 {
...
   def zipWithIndex[F[_],I]: Stream[F,I] => Stream[F,(I,Int)] = {
-    def go(n: Int): Handle[F, I] => Pull[F, (I, Int), Handle[F, I]] = {
-      Pull.receive { case chunk #: h =>
-        Pull.output(chunk.zipWithIndex.map({ case (c, i) => (c, i + n) })) >> go(n + chunk.size)(h)
-      }
-    }
-    _ pull (go(0))
+    mapAccumulate[F, Int, I, I](-1) {
+      case (i, x) => (i + 1, x)
+    } andThen { _.map {
```
pchiusano: You can use the `.swap` function on `Tuple2`

pchiusano:
I think you meant `zipWithIndex` in the title of this PR?

kanterov:
@pchiusano right, I've updated it

PR #481, opened 2015-11-14
Refactor mapAccumulate, zipWithNext.

fiadliel:
This puts `mapAccumulate` into `pull1`. Adds an extra parameter for optionally emitting a value when the pull finishes.

Defines `process1.mapAccumulate` and `process1.zipWithNext` in terms of this.

Note: we could use `or` instead of `receiveOption` to avoid extra Option object creation, but will use the combinator until there's any evidence that this is in any way a perf problem.


pchiusano:
Nah.

Too special purpose IMO. If you ever want to do something different at the end of a stream, that is what terminated is for. No need to bake this into a combinator like mapAccumulate.

PR #482, opened 2015-11-14
Add chunkLimit/chunkN/vectorChunkN combinators

fiadliel:
Implements:
 - chunkLimit - Repeatedly output chunks with a maximum size.
 - chunkN - Group chunks so each group has a specified maximum size.
 - vectorChunkN - Flatten grouped chunks into a Vector (like old chunk combinator).

Updates pull1.awaitN to allow chunk sequences sized less than specified.

### DIFF
--- a/src/main/scala/fs2/Process1.scala
+++ b/src/main/scala/fs2/Process1.scala
@@ -15,6 +15,14 @@ object process1 {
 
   // nb: methods are in alphabetical order
 
+  /** Outputs chunks with a limited maximum size, splitting as necessary. */
+  def chunkLimit[F[_],I](n: Int): Stream[F,I] => Stream[F,Chunk[I]] =
+    _ repeatPull { h => Pull.awaitLimit(n)(h) flatMap { case chunk #: h => Pull.output1(chunk) as h } }
+
+  /** Outputs a list of chunks, the total size of all chunks is limited and split as necessary. */
+  def chunkN[F[_],I](n: Int, allowFewer: Boolean = true): Stream[F,I] => Stream[F,List[Chunk[I]]] =
+    _ repeatPull { h => Pull.awaitN(n, allowFewer)(h) flatMap { case chunks #: h => Pull.output1(chunks) as h }}
+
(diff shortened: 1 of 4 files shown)


fiadliel:
Updated to use `Stream.terminated` instead of a special combinator.

pchiusano:
A few comments - 

* Don't make `filterMapAccumulate` a `Pull`, like I mentioned in #481.
* I don't like the name `grouped`. It's more important to me to have internal consistency within FS2 than consistency with other vaguely similar libraries. My other issue with `grouped` is that the library refers to the concept as chunks (which I prefer to 'group'), so we should keep the naming consistent.
  * Also, avoid '-ed' suffixes. It's a holdover from the days where it was unclear whether a function mutated its input. In a purely functional lib, it's not needed - no function mutates its input. Notice it's not even consistently applied. (Like why do we say `map` rather than `mapped`??)

Here's what I propose for the naming. First, let's rename `chunks` to `chunk`. Idea is - "chunk the stream, according to its natural structure." Then introduce:

```scala
// in pull1
// Same as existing definition, but `allowFewer` controls whether a less than `n` chunk sequence
// is emitted at the end of the input
def awaitN[F[_],A](n: Int, allowFewer: Boolean = false)

// in process1

/** Repeatedly `Pull.awaitLimit(n)`, and emit the resulting chunks. */
def chunkLimit[F[_],A](n: Int): Stream[F,A] => Stream[F,Chunk[A]] = ..

/** Repeatedly `Pull.awaitN(n, allowFewer)`. */
def chunkN[F[_],A](n: Int, allowFewer: Boolean = true): Stream[F,A] => Stream[F,List[Chunk[A]]] = ...

/** New name for `grouped`. */
def vectorChunkN[F[_],A](n: Int, allowFewer: Boolean = true): Stream[F,A] => Stream[F,Vector[A]] = 
  s => /* Define in terms of chunkN */
```

fiadliel:
@pchiusano Thanks for the feedback! I've updated following your suggestions, it works :-)

pchiusano:
Looks good, thanks!

PR #484, opened 2015-11-14
Renamed implicit classes defining operators to match pattern ...Ops

mpilquist:
…and moved type alias for UF1 to util package object.

Not a huge fan of some of the names but I used 2 rules when choosing them:
 - must end in `Ops`
 - must start with the name of the class that's being decorated

These rules are designed to simplify source navigation in symbol aware text editors.


pchiusano:
Names are fine I think. Looks good.

PR #485, opened 2015-11-16
adding collectFirst to Process1

kaiserpelagic:
Using Pull.find seemed like the most intuitive way to model collectFirst. Let me know if there is a better way.


kaiserpelagic:
Just noticed this pull request for collectFirst that was not merged: https://github.com/scalaz/scalaz-stream/pull/467. Pull.find basically does what was proposed in that pull request but there was a lot of discussion around performance / composability tradeoffs. looking forward to any input.

pchiusano:
I think this is good.

PR #491, opened 2015-11-20
Added several convenience functions

pchiusano:
Implemented `Pull.awaitNonempty` (obtains the next nonempty `Chunk`), `Pull.echo1` and `Pull.echoChunk`. Also some convenient aliases for `receive1`, `receive`, `echo1`, and `echoChunk` on `Handle`. And added `repeatPull2`, analogous to `repeatPull`, useful for building `Tee` values.

No unit tests, but let's just start using these in new definitions and feel free to update old ones to get some ambient testing. I think the `echo`, `echo1`, and `echoChunk` will be particularly useful in various tee and wye definitions.

### DIFF
--- a/src/main/scala/fs2/StreamDerived.scala
+++ b/src/main/scala/fs2/StreamDerived.scala
@@ -16,6 +16,11 @@ private[fs2] trait StreamDerived { self: fs2.Stream.type =>
   : Stream[F,B] =
     pull(s)(Pull.loop(using))
 
+  def repeatPull2[F[_],A,B,C](s: Stream[F,A], s2: Stream[F,B])(
+    using: (Handle[F,A], Handle[F,B]) => Pull[F,C,(Handle[F,A],Handle[F,B])])
+  : Stream[F,C] =
+    s.open.flatMap { s => s2.open.flatMap { s2 => Pull.loop(using.tupled)((s,s2)) }}.run
+
   def await1Async[F[_],A](h: Handle[F,A])(implicit F: Async[F]): Pull[F, Nothing, AsyncStep1[F,A]] =
     h.awaitAsync map { _ map { _.map {
       case Step(hd, tl) => hd.uncons match {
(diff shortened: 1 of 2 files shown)


kanterov:
What do you think about creating a type representing non-empty chunk with required head?

pchiusano:
@kanterov Might be worth using a tag on Chunk. The problem is preserving this info in the presence of other operations adds a lot of complexity, and I don't want to go there. I'm going to merge but feel free to mess around with it. I feel like we have more pressing stuff that needs doing though, and I'm not sure I'd merge anything unless it were super lightweight.

kanterov:
@pchiusano Got it. What do you think about doing something simple like:
```scala
case class NonEmptyChunk[A](
  head: A,
  tail: Chunk[A],
  chunk: Chunk[A]
)
```
Where
```
head == chunk(0)
tail == chunk.drop(1)
```

pchiusano:
@kanterov I don't like introducing a new type for this. It adds plumbing code (you have to explicitly convert to/from it). You could have `NonemptyChunk` be a subtype, but then you have to delegate a bunch of methods to the underlying chunk.

What I was thinking is to use a tag. This is a pretty common technique -

```Scala
object Chunk {
  trait Nonempty
  type NonemptyChunk[+A] = Chunk[A] with Nonempty

  private def nonempty[A](c: Chunk[A]): NonemptyChunk[A] = c.asInstanceOf[NonemptyChunk[A]]
  
  implicit class NonemptyOps[A](c: NonemptyChunk[A]) {
    def head: A = c.uncons.get._1
    def tail: Chunk[A] = c.uncons.get._2
  }
}
```

Thus a `NonemptyChunk[A]` can be supplied anywhere a regular `Chunk` is expected.

I would not bother with trying to preserve nonemptiness in the presence of operations like `map` or anything else that is length preserving. That is what I was saying would lead to overkill.

If you want to prepare a PR with this I'd merge it. I'd suggest adding a public `nonempty: Option[NonemptyChunk[A]]` method on `Chunk` and using that in the implementation of `awaitNonempty` and wherever else. This should be the sole place where the cast to `NonemptyChunk` is performed in the codebase.

PR #493, opened 2015-11-23
Fix for #477

### LINKED ISSUE #477: `awaitAsync` not asynchronous enough
I noticed this while testing `process1.prefetch`. Only the leftmost stream operation is done asynchronously, whereas you really want all operations to be done asynchronously until a chunk can be produced.

Here's an example:

```Scala
val s = Stream(1).flatMap(i => Stream.eval(Task.delay(Thread.sleep(1000))))
s pull { h => h.awaitAsync }
```

Currently, that will just step the `Stream(1)` asynchronously (somewhat pointlessly, since it's strict), returning a Future that will force the sleep `Task` _synchronously_.

You don't just want the leftmost stream to be stepped asynchronously, because after that finishes, we still haven't produced a chunk. So we should continue doing things asynchronously, up to the `Thread.sleep` which finally produces a chunk when it's done.

I don't have a solution to this yet, but it's obviously kind of a big deal.

pchiusano:
This fixes #477. Ended up being pretty involved. `awaitAsync` is now defined in terms of `await`, which I like. It uses mutable state in the implementation... not as nice. But this isn't observable to anyone, so it's fine.

Each `awaitAsync` builds up a single `Task` or other effect that is used to set a `Ref`, so it will be efficient in its use of thread pool submits.

The main entrypoint is [`Stream.stepAsync`](https://github.com/functional-streams-for-scala/fs2/blob/topic/redesign-prefetch-fix/src/main/scala/fs2/Stream.scala#L58).

I am not completely positive that finalizers cannot be run more than once with this approach (something that the old approach guaranteed by only ever running finalizers in the main `runFold` interpreter). The tests would catch this if it were being done in any obvious way, but I would like to ponder whether there's some weird corner case where it could occur. I'm going to hold off on merging until I feel more confident in this and I'll maybe add a couple tests if I think of any.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,26 @@
-scalaz-stream
+FS2: Functional Streams for Scala (previously 'Scalaz-Stream')
 =============
 
-[![Build Status](https://travis-ci.org/scalaz/scalaz-stream.svg?branch=master)](http://travis-ci.org/scalaz/scalaz-stream)
-[![Gitter Chat](https://badges.gitter.im/scalaz/scalaz-stream.svg)](https://gitter.im/scalaz/scalaz-stream)
+[![Build Status](https://travis-ci.org/functional-streams-for-scala/fs2.svg?branch=topic/redesign)](http://travis-ci.org/functional-streams-for-scala/fs2)
+[![Gitter Chat](https://badges.gitter.im/functional-streams-for-scala/fs2.svg)](https://gitter.im/functional-streams-for-scala/fs2)
 
-### Where to get it ###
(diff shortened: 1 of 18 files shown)


pchiusano:
Okay, I'm feeling pretty good about this (travis failure is totally unrelated). I was able to convince myself that finalizers can only be run once, and in the process I simplified the implementation some and added a bunch of comments to hopefully clarify.

I'm going to merge this tomorrow if I don't think of any issues with it between now and then.

pchiusano:
Okay, merging now..

PR #496, opened 2015-11-25
Added `interleaveAll` operation

ahmadsalim:
Additionally added tests for both `interleave` and `interleaveAll`.

### DIFF
--- a/src/main/scala/fs2/StreamDerived.scala
+++ b/src/main/scala/fs2/StreamDerived.scala
@@ -57,6 +57,9 @@ private[fs2] trait StreamDerived { self: fs2.Stream.type =>
   def eval_[F[_],A](fa: F[A]): Stream[F,Nothing] =
     flatMap(eval(fa)) { _ => empty }
 
+  def infinite[F[_],W](ws: W*): Stream[F, W] =
+    emits(ws) ++ infinite(ws :_*)
+
   def push1[F[_],A](h: Handle[F,A])(a: A): Handle[F,A] =
     push(h)(Chunk.singleton(a))
 
--- a/src/main/scala/fs2/StreamDerived.scala
+++ b/src/main/scala/fs2/StreamDerived.scala
@@ -57,8 +57,13 @@ private[fs2] trait StreamDerived { self: fs2.Stream.type =>
   def eval_[F[_],A](fa: F[A]): Stream[F,Nothing] =
     flatMap(eval(fa)) { _ => empty }
 
-  def infinite[F[_],W](ws: W*): Stream[F, W] =
-    emits(ws) ++ infinite(ws :_*)
+  /**
+    * The infinite `Process`, always emits `a`.
+    * If for performance reasons it is good to emit `a` in chunks,
+    * specify size of chunk by `chunkSize` parameter
+    */
+  def constant[F[_],W](w: W, chunkSize: Int = 1): Stream[F, W] =
+    emits(List.fill(chunkSize)(w)) ++ constant(w, chunkSize)
 
   def push1[F[_],A](h: Handle[F,A])(a: A): Handle[F,A] =
     push(h)(Chunk.singleton(a))
--- a/src/main/scala/fs2/Tee.scala
+++ b/src/main/scala/fs2/Tee.scala
@@ -12,36 +12,99 @@ object tee {
   def covary[F[_],I,I2,O](p: Tee[I,I2,O]): (Stream[F,I], Stream[F,I2]) => Stream[F,O] =
     p.asInstanceOf[(Stream[F,I],Stream[F,I2]) => Stream[F,O]]
 
-  def zipWith[F[_],I,I2,O](f: (I, I2) => O) : (Stream[F, I], Stream[F, I2]) => Stream[F, O] = {
-     _.repeatPull2(_)((h1, h2) => h1 receive1 {
-        case i1 #: h1 => h2 receive1 {
-          case i2 #: h2 => P.output1(f(i1, i2)) >> P.pure((h1, h2))
-        }
-      })
+
+  private def zipChunksWith[I,I2,O](f: (I, I2) => O)(c1: Chunk[I], c2: Chunk[I2]): Step[Chunk[O], Option[Either[Chunk[I], Chunk[I2]]]] = {
+      def go(v1: Vector[I], v2: Vector[I2], acc: Vector[O]): Step[Chunk[O], Option[Either[Chunk[I], Chunk[I2]]]] = (v1, v2) match {
+        case (Seq(),Seq())        => Step(Chunk.seq(acc.reverse), None)
+        case (v1,   Seq())        => Step(Chunk.seq(acc.reverse), Some(Left(Chunk.seq(v1))))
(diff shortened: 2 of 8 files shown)


ahmadsalim:
The `interleaveAll` operation basically keeps emitting values from the longer stream when the shorter stream is halted, instead of cutting off like `interleave` does. I hope it is useful, also in conjunction with the new redesign.

pchiusano:
@ahmadsalim would you like to try implementing this for the new representation (in `fs2` package)? I probably won't merge this as is, just because we're directing most of our efforts toward the FS2 work and wouldn't likely do another release off the old design unless it's like a serious bugfix and the new release is still a ways off.

ahmadsalim:
@pchiusano Sure

ahmadsalim:
@pchiusano OK, I have updated the implementation for `fs2` as well. Do you want me to remove the old one in `scalaz.stream`?

pchiusano:
Yeah, you can remove it.

I have one more request. I didn't want to bring it up earlier to discourage you, but I'd like to preserve the chunkiness of the input streams as much as possible. This will be much more efficient. So use `receive` rather than `receive1` to obtain chunks, then interleave the chunks. The code will be more complicated, since you'll need to deal with the case where the chunks have different sizes.

ahmadsalim:
OK, I will take a look.

src/main/scala/fs2/StreamDerived.scala
```diff
@@ -57,6 +57,9 @@ private[fs2] trait StreamDerived { self: fs2.Stream.type =>
   def eval_[F[_],A](fa: F[A]): Stream[F,Nothing] =
     flatMap(eval(fa)) { _ => empty }
 
+  def infinite[F[_],W](ws: W*): Stream[F, W] =
```
pchiusano: Pls call this `constant` to be consistent with existing signature in old code:

```Scala
  /**
   * The infinite `Process`, always emits `a`.
   * If for performance reasons it is good to emit `a` in chunks,
   * specify size of chunk by `chunkSize` parameter
   */
  def constant[A](a: A, chunkSize: Int = 1): Process0[A]
```

Also, just have it take a single `W`. If you want to repeat a sequence over and over, just do `emits(ws).repeat`.

src/main/scala/fs2/Tee.scala
```diff
@@ -10,6 +12,115 @@ object tee {
...
+  def zipAllWith[F[_],I,I2,O](pad1: I, pad2: I2)(f: (I, I2) => O): (Stream[F, I], Stream[F, I2]) => Stream[F, O] = {
+      def cont1(z: Either[Step[Chunk[I], Handle[F, I]], Handle[F, I]]): Pull[F, O, Nothing] = {
+        def putLeft(c: Chunk[I]) = {
+          val co = Chunk.seq(c.toVector.zip(
+                               Vector.range(0, c.size).map(_ => pad2)))
+                        .map(f.tupled)
+          P.output(co)
+        }
+        def contLeft(h: Handle[F,I]): Pull[F,O,Nothing] = {
+          P.receive((s: Step[Chunk[I], Handle[F, I]]) => s match {
```
pchiusano: just do `h.receive { case c #: h => ... }`.

src/main/scala/fs2/Tee.scala
```diff
@@ -10,6 +12,115 @@ object tee {
...
+            case Some(s2) => zipChunksGo(s1, s2)
+            case None => k1(Left(s1))
+          })(h2)
+          case None => k2(Right(h2))
+        })(h1)
+      }
+      _.pull2(_)(goB)
+  }
+
+  def zipAllWith[F[_],I,I2,O](pad1: I, pad2: I2)(f: (I, I2) => O): (Stream[F, I], Stream[F, I2]) => Stream[F, O] = {
```
pchiusano: wow this got pretty complicated

src/main/scala/fs2/Tee.scala
```diff
@@ -10,6 +12,115 @@ object tee {
...
+        }
+      }
+      def cont2(z: Either[Step[Chunk[I2], Handle[F, I2]], Handle[F, I2]]): Pull[F, O, Nothing] = {
+        def putRight(c: Chunk[I2]) = {
+          val co = Chunk.seq(Vector.range(0, c.size).map(_ => pad1).zip(
+                            c.toVector)).map(f.tupled)
+          P.output(co)
+        }
+        def contRight(h: Handle[F,I2]): Pull[F,O,Nothing] = {
+          P.receive((s: Step[Chunk[I2], Handle[F, I2]]) => s match {
```
pchiusano: `h.receive { case c #: h => ... }`

src/test/scala/fs2/TeeSpec.scala
```diff
@@ -0,0 +1,48 @@
+package fs2
+
+import TestUtil._
+
+import tee._
+
+import org.scalacheck.Prop._
+import org.scalacheck.{Gen, Properties}
+
+object TeeSpec extends Properties("tee") {
```
pchiusano: tests for the zip functions?

src/main/scala/fs2/Tee.scala
```diff
@@ -10,6 +12,115 @@ object tee {
...
+  def zip[F[_],I,I2]: (Stream[F, I], Stream[F, I2]) => Stream[F, (I, I2)] =
+    zipWith(Tuple2.apply)
+
+  def zipAll[F[_],I,I2](pad1: I, pad2: I2): (Stream[F, I], Stream[F, I2]) => Stream[F, (I, I2)] =
+    zipAllWith(pad1,pad2)(Tuple2.apply)
+
+  def interleave[F[_], O]: (Stream[F,O], Stream[F,O]) => Stream[F,O] =
+    zip(_,_) flatMap { case (i1,i2) => Stream(i1,i2) }
+
+  def interleaveAll[F[_], O]: (Stream[F,O], Stream[F,O]) => Stream[F,O] = { (s1, s2) =>
```
pchiusano: cute implementation!

the output doesn't preserve chunkiness though

pchiusano:
Sorry for the delay in responding, I didn't see you'd updated the PR (I get notified if you add a comment, not if you just add a commit to the PR).

I actually have a bunch of things I want to do after this is merged to see if it's possible to clean up, however I think it'll be easier if I just work on this after merging rather than doing a bunch of back and forth.

However, can you just:

* Just rename `infinite` to `constant` and give it an optional default `chunkSize` parameter like the old signature.
* Add some tests for the `zip` functions. I know they get some ambient testing via `interleave` tests.

Then I'll merge.

ahmadsalim:
@pchiusano Sure, I will do that. 

I actually wanted to notify you after I had added some tests and cleaned up a bit, but I got preoccupied with a few things at work.

pchiusano:
Np, thanks!

ahmadsalim:
@pchiusano OK, I tried to clean up a bit as you suggested and renamed `infinite` to `constant` and added tests for the `zip*` functions. Sorry for the complex code, I wanted to clean it up more but fell a bit short on time; I am glad that you will try to simplify it.

pchiusano:
Okay cool merging.

PR #497, opened 2015-11-26
Add compress.deflate, compress.inflate, Chunk.toArray

fiadliel:
I suspect this will need some discussion before being added:

I tried using `Chunk[Byte]` instead of `ByteVector`. The variance on `toArray` was annoying :)
Also, it seems that `Byte` values can easily be de-optimized, e.g. if I want an `Either[DataFormatException, Byte]`. Perhaps this should work on the `Chunk[Byte]` level instead? But then there's not a huge amount of benefit to the wrapping of `ByteVector`?

This will need to either return that `Either` I mentioned above, or it will throw exceptions -- if so, it shouldn't have a `Pure` effect, but `F: Catchable`? Is there a particular pattern you'd like here?

### DIFF
--- a/core/src/main/scala/fs2/Chunk.scala
+++ b/core/src/main/scala/fs2/Chunk.scala
@@ -1,5 +1,7 @@
 package fs2
 
+import scala.reflect.ClassTag
+
 /**
  * Chunk represents a strict, in-memory sequence of `A` values.
  */
@@ -9,6 +11,7 @@ trait Chunk[+A] { self =>
     if (size == 0) None
     else Some(apply(0) -> drop(1))
   def apply(i: Int): A
+  def copyToArray[B >: A](xs: Array[B]): Unit
   def drop(n: Int): Chunk[A]
   def take(n: Int): Chunk[A]
   def filter(f: A => Boolean): Chunk[A]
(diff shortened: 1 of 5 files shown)


pchiusano:
Sorry for delayed response!

Here's what I'd like to do:

* I'd like to remove the dependency on `scodec-bits`. This way the core of the library will have zero dependencies. So just have `bits` and `bytes` be backed by an Array, or an Array plus a starting and ending index. Also make them return named classes, so the user can pattern match on them if they wish to do things unboxed.
* The main method for doing array-based stuff with chunks should be `copyToArray` I think. `toArray` should be implemented in terms of it. Just follow the scala collections convention here.
* Deflate and inflate can just work with a `Stream[F,Byte]`. Internally, they should use chunking and deal with unboxed arrays if the input chunks are unboxed.
* I'd like to make the underlying array and offset public in the implementation of `bits`, `bytes`, and `double`. Although potentially unsafe if the user like stashes the array or writes to it, my response is just don't do that. It's important that the user be able to have a guarantee they are working with unboxed arrays and values.

fiadliel:
Not ready to merge right now; writing property tests for Chunk.

pchiusano:
Cool.

pchiusano:
@fiadliel any interest in updating this PR? :) The main thing that has changed is that I already redid the various primitive `Chunk` types. There is still no `copyToArray` method on `Chunk` though.

> This will need to either return that Either I mentioned above, or it will throw exceptions -- if so, it shouldn't have a Pure effect, but F: Catchable? Is there a particular pattern you'd like here?

If you need to throw an exception here, just use `Stream.fail` or `Pull.fail`, don't return `Either`. The user can always call `onError` if they want to catch the exception and do something with it. If you get a DataFormatException, it's not really a recoverable error in the vast majority of cases, so forcing the user to explicitly reraise or handle is not necessary - passing along the error is a totally fine default. (Unlike, say, looking up a nonexistent key in a map, which is very often recoverable.)

fiadliel:
@pchiusano Updated the PR, sorry for the delay!

pchiusano:
Looks good to me! Merging.

PR #499, opened 2015-12-01
read multiple times as needed in io.unsafeChunkR (fixes #498)

### LINKED ISSUE #498: unsafeChunkR can return too few bytes before the InputStream is empty
I think I've run into the fs2 version of this:
http://stackoverflow.com/questions/7874713/java-read-from-inputstream-doesnt-always-read-the-same-amount-of-data

I've got a method like this:
```
def loadBatch(is: InputStream): Process[Task, Labeled] =
  Process.constant(3073).toSource.through(io.chunkR(is))
    .map{v => println(v.length); Labeled(v.head, CifarUnlabeled(v.tail.toArray))}
```
and am observing that every 3rd read comes up short:
```
3073
3073
2590
3073
3073
2531
3073
3073
2167
3073
3073
2768
...
```

They don't add up to anything meaningful, like 10240, and in fact the "short" read length doesn't seem to repeat.

When I read the same file using a `java.io.FileInputStream`, I get consistent 3073-byte reads.  When I call `Thread.currentThread.getContextClassLoader.getResourceAsStream` which returns a `sun.net.www.protocol.jar.JarURLConnection$JarURLInputStream`, I get the uneven chunks demonstrated above.

I'm not sure what the tidiest way to loop within `unsafeChunkR` to extract as much as possible from the `InputStream` before returning would be (e.g. multiple calls to `.read` and concatenating the results together until reaching EOF), but would be grateful for any suggestions.

refried:
First PR here; a little scary.  Let me know what I screwed up :-)

### DIFF
--- a/src/main/scala/scalaz/stream/io.scala
+++ b/src/main/scala/scalaz/stream/io.scala
@@ -220,8 +220,8 @@ object io {
 
   /**
    * Creates a `Channel[Task,Array[Byte],Array[Byte]]` from an `InputStream` by
-   * repeatedly filling the input buffer. The last chunk may be less
-   * than the requested size.
+   * repeatedly filling the input buffer. `is.read` will be called multiple times
+   * as needed; however, the last chunk may be less than the requested size.
    *
    * It is safe to recycle the same buffer for consecutive reads
    * as long as whatever consumes this `Process` never stores the `Array[Byte]`
@@ -240,7 +240,15 @@ object io {
         val m = src.read(buf)
         if (m == buf.length) buf
         else if (m == -1) throw Cause.Terminated(Cause.End)
-        else buf.take(m)
+        else { // multiple reads for stubborn InputStreams
+          @tailrec def loop(received: Int): Array[Byte] = {
+            val m2 = src.read(buf, received, buf.length - received)
+            if (m2 == -1) buf.take(received)
+            else if (m2 + received == buf.length) buf
+            else loop(m2 + received)
+          }
+          loop(m)
+        }
       }}
     }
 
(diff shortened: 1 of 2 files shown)


refried:
for #498

pchiusano:
@refried seems reasonable, can you craft a test though (which would fail with the old implementation)? You can literally just create a new (pure) `InputStream` that has the properties you want - no need to do actual I/O: http://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html

I don't think we would do another point release of scalaz-stream just for this, but I'd still like to merge this in case we do want to do that. Also the test and the fix will probably be ported over to the `fs2.io` package when we get to that.

refried:
@pchiusano Yes, I will do that.

A point release would be great though, otherwise I'll need to find some suboptimal home on my end for the fixed version.  I know that the new design/branch is where the current excitement is at, though.

refried:
Okay.  On master:
```
[info] ! io.unsafeChunkR.repeats read() to fill buffer: Falsified after 1 passed tests.
```
on branch:
```
[info] + io.unsafeChunkR.repeats read() to fill buffer: OK, passed 100 tests.
```

refried:
Ugh, a different test failed to avoid deadlock on one of the three CI builds, I don't think it's related to my patch.
```
[info] ! Wye.outer cleanup through pipe gets called before interrupted inner: Falsified after 0 passed tests.
[info] > Labels of failing property: 
[info] avoided deadlock
```

refried:
@pchiusano \*poke\*  Do you wanna restart the CI builds and/or create an issue to address the Wye deadlock failure?

pchiusano:
Okay, yeah, that failure looks unrelated. I'll merge now. Like I said, we probably won't do a point release just for this.

PR #504, opened 2015-12-10
Added semaphore concurrency primitive and used it for bounded queue impl

pchiusano:
As discussed [here](https://github.com/functional-streams-for-scala/fs2/pull/462#issuecomment-163683878) in #462. I *really* like how this turned out. Code got simpler, and I think `Semaphore` will prove useful for other stuff. It is so cool that we can implement this stuff generally, for any `AsyncExt`. Thank you for pushing us to do that @pchlupacek!

I did a bunch of cleanup / improvements while I was at it:

* `Signal`, `Queue`, and `Semaphore` creation return an `F`, not a stream. This is a bit more precise type, it makes it clear that only a single value is being returned. The user can `Stream.eval` it of course.
* Fixed what seemed to be a bug in the pervious queue implementation where `offer1` didn't update state to remove the matched dequeuer.
* We no longer allocate a `ref` on every `dequeue1`, we only need to do this if the queue is empty. Likewise on `enqueue` in the bounded case - the `Semaphore` implementation doesn't allocate a `Ref` unless the number of permits requested exceeds the number available. This probably boosts performance quite a bit.
* Some other cleanup and simplifications now that the unbounded case is totally separate.
* Renamed `async.fromStream` to `holdOption` and introduced `hold` as well, in case you want a different initial value besides `None`.
* Random code formatting cleanup.
* I have been using the convention that the package object file is just named whatever the package is, rather than `package.scala`. So I renamed `async/package.scala` to `async/async.scala`.

This is totally untested still. Let's do tests separately from this PR.

Review by @pchlupacek


pchiusano:
Something else I just thought of is that circular buffer can be implemented very nicely with semaphores and an unbounded queue. To `enqueue`, you just `tryAcquire`, and if it fails, do a `dequeue1` before your `enqueue1`. You might need another semaphore to avoid the race condition that would occur if multiple threads are doing this.

pchiusano:
@pchlupacek after you review this and we merge, I'd like to split the project into multiple modules before doing anything else (including the testing)

src/main/scala/fs2/async/mutable/Semaphore.scala
```diff
@@ -0,0 +1,91 @@
...
+
+import fs2.async.AsyncExt
+
+/**
+ * An asynchronous semaphore, useful as a concurrency primitive.
+ */
+trait Semaphore[F[_]] {
+
+  /** Returns the number of permits currently available for acquisition. Always nonnegative. */
+  def available: F[Long]
```
pchlupacek:
```scala
def availableSignal:immutable.Signal[Long] =  ???
```

pchlupacek:
@pchiusano cool. We can perhaps add sort of that signal on semaphore later, I really believe it would be useful. In fact I was thinking how much code signal and semaphore can share, and I just need to think about it little bit more. LGTM ?

pchlupacek:
@pchiusano also side not, please ping me when you split the project. I am considering to start using fs2 together with our 0.8 code and perhaps migrate step-by step. 
Thanks.

pchiusano:
@pchlupacek Okay cool, will merge. I just added one more commit which changes the naming to `increment`/`decrement` rather than `acquire`/`release`, since I kept getting confused about which direction was which. I also made a change which I'm not sure matters but seems more correct - on `increment`, the semaphore does not run any callbacks inside the `modify` action, it modifies the state, and then afterwards examines how the state has changed to determine what callbacks should run. My concern was that the callbacks themselves might somehow call back into `modify`, which would cause a deadlock.

I'm not sure if this is what you meant, but I bet `Semaphore` can be used to implement `Signal` really simply. Here's what I was thinking:

```Scala
type State[F[_],A] = (F.Ref[A], Vector[Semaphore[F]])
```

* `continuous` is implemented just by `repeatEval(get)` on the `Ref` 
* `set` is implemented by setting the `Ref`, then going through each semaphore and doing a `tryDecrement` followed by an `increment`.
* `discrete` uses `bracket` to allocate a `Semaphore(0)`, add it to the state, and then repeatedly does a `decrement`, followed by a `get` of the current value. The cleanup action of the `bracket` then removes the semaphore from the signal state.

So basically, `discrete` waits on the semaphore for the signal to be refreshed, before pulling the current value. If the signal refreshes more often than the `discrete` stream is pulling from it, then the `tryDecrement` will reset the semaphore to 0 just before each refresh. If the `discrete` stream has already seen the previous value, the `tryDecrement` will noop since the count is already zero.

pchlupacek:
@pchiusano cool. Well in fact I thought in contrary on Signal//Semaphore. To me signal is like CAS primitive with ability to signal it state. Hence in normal concurrent code CAS may be used to implement semaphore, so perhaps there is a way to build Semaphore just by utilising signal for its state. Not sure if it is doable, but I think would be nice. 

I just thought there may be useful feature if semaphore will have discrete signal of open acquires. I need to think more about it, maybe it is not that good idea altogether.

PR #514, opened 2015-12-23
Fix race condition in AsyncExt[Task] 

mpilquist:
I noticed some flaky behavior when using the new queue implementation. After some discussion with @pchiusano, I decided to write some tests for Semaphore. This exposed a bug in Semaphore, which Paul fixed, but even after that fix, I was still experiencing deadlocks with the SemaphoreSpec in this PR. The deadlocks were caused by 2 bugs in AsyncExt[Task], where the waitingModify map was being updated as if it was a mutable map. After fixing, both the semaphore and the queue tests pass.

### DIFF
--- a/core/src/main/scala/fs2/util/Task.scala
+++ b/core/src/main/scala/fs2/util/Task.scala
@@ -304,8 +304,9 @@ object Task extends Instances {
         else { val r = result; S { cb(r) } }
 
       case Msg.Set(r) =>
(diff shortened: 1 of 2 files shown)


pchiusano:
Wow, awesome, thanks @mpilquist for spotting this and tracking it down!! And the queue tests now pass apparently so it looks like the deadlocks there were caused by the same underlying issue.

pchlupacek:
@mpilquist excellent thanks for that. I hate the () :-)

Issue #501, opened 2015-12-06
fs2.Sink/Channel replacement


pchlupacek:
Rethink and properly describe the Channel/Sink replacement in fs2

pchlupacek:
I know we have been discussing this on few places, I just opened this issue so we may correctly track it. Hence channel and sink from fs2._ is widely used in almost any program, I think we need to clearly describe how this will be done in fs2._ 

I specifically curious how we would do state-dependent sinks and channel alternatives, that first acquired resource and then produced stream of functions. 

Also would like to see, if and how we can combine channel/sink replacements. I just want to make sure concept is well understood before implementing them on async primitives.

pchiusano:
Yeah agree this is still open. I'm going to do a writeup of some possible
designs early this week and will post here and we can discuss.

I think we can hold off on implementing any conversion functions between
the async types and channels until we are solid on the channel
representation.

pchlupacek:
Sure agree. I'll implement them once we will know more on this topic.

pchiusano:
_Note:_ not bothering to discuss sinks here, which I think will always just be `type Sink[F[_],-A] = Channel[F,A,Unit]` or `Channel[F,A,Nothing]`, regardless of representation of `Channel`.

### Option 1

```Scala
type Channel[F[_],-A,+B] = Stream[F,A] => Stream[F,B]
```

_The good:_ It's very flexible in one sense - every operation available on streams is available when building up a channel. Channels can be a simple lift of some `A => F[B]`, or might pick an `A => F[B]` based on what `A` is received. The channel can insert or delete `A` values, transform them in any way, and stateful channels are possible, for things like rotating loggers, etc.

_The bad:_ It's very opaque and is missing a story for how to _interleave_ sending values through multiple channels. For example, how do we implement `observe`:

```Scala
def observe[F[_],A,B](s: Stream[F,A], sink: Sink[F,A]): Stream[F,A]
```

Likewise for more complicated scenarios, like sending the same value through two channels and combining the results in some fashion.

An observation here is that it is actually a general limitation of the library, that it has no story for how to _replicate/fork streams_ for consumption in multiple places (short of using mutable state like queues and so forth). So addressing this problem in a general way might be good---it lets us use this very flexible and simple representation of channels, with interleaving solved by a more general capability (which we still have to design).

Let's come back to this one.

### Option 2 (rejected)

Here's one I was playing with based on some discussion with @mpilquist on scodec-stream:

```Scala
[...]

PR #515, opened 2015-12-23
Implementation of `observe`, `observeAsync`, and `diamond` combinator

pchiusano:
As discussed in #501.

Currently in `async.channel` package. Not totally sold on that. But it does raise the possibility of still having another `channel` package in `fs2.channel` for more synchronous stuff.

This is not tested, maybe we should hold off on merging until we experiment with the API some and add some tests. 

I went with just having the `diamond` combinator take a regular `Queue`. I think it adds too much complexity to have a separate `QueueStrategy` type. Just don't reuse the same queue you are using elsewhere. And if you do, be prepared for weird behavior.


pchiusano:
Also, fyi I am going to be on holiday starting tomorrow through Jan 5th, so I probably won't be as responsive between now and then, but I'll reply to threads when I can. :)

pchlupacek:
ok no prob. I gonna experiment with sinks. I have WIP on kafka driver, will try to implement it with fs2 so we will see what still missing. I may also do some PRs on closable signal and queue. Take care!

pchlupacek:
@pchiusano I have played a bit with various ways on sinks and channels, and Just for curiosity, have you considered : 
```scala
type Channel[F,A,B] = Stream[F,Stream[F,A] => Stream[F,B]]  ???
```
To me this works nicely to implement, lift, observe, through, while still keeping chunkiness and lot of flexibility like the pure function. In fact for observe, through et all I was thinking of slightly modified signature like 
```
type Channel[F,A,B] = Stream[F,Stream[Pure,A] => Stream[F,B]]
```
but that would limit some composability, in fact. 

I like on first signature that you can do things like 
``` scala
(ch1 zip ch2) map (_ andThen _) 
``` 
mapOut, mapIn etc will be easy to implement too. 

I would say the only problematic is `pipeIn` and `pipeOut`, like we had in old implementation, but I think this is ok, as we can use similar technique like you proposing in diamond to implement them. In fact Perhaps introducing `SynchronousQueue` to async would be beneficial for these situations.

pchlupacek:
Just to clarify. I see this suffers from option (3) limitations too, but not necessarily big deal imho. For example to construct functions that depends on result of the last invocation may be in fact quite ok:

```scala

val state : F[Signal[F,Boolan]] = ???
def write(full:Boolean, r: R)(src:Stream[F,A]):Stream[F,(Boolean,A)] = ??? // implement write to R, that will drop if full == true , echoing full at the output. 

val channel: Channel[F,A, (Boolean,A)] = bracket(acuireR)(r=>  state.continuous.map { full => write(full,r) _ }, cleanup = releaseR )

```

mpilquist:
Note: this work is now on topic/redesign -- Paul merged it in to topic/redesign-simple-interpreter before that branch was merged. I'm closing this but if there are any more comments, feel free to reopen or file individual issues.

PR #516, opened 2015-12-25
Added failing tests for merge and join

pchlupacek:
Added tests that currently fails for merge and join. Discovered them when starting to work with queue. 
@pchiusano hence both combinators fails, I think the problems is that none of them has the way to signal interruption correctly. Maybe you have some ideas how to fix that. Not sure if the fix will be easy, wondering if change to low level combinators won't be necessary.

### DIFF
--- a/core/src/test/scala/fs2/ConcurrentSpec.scala
+++ b/core/src/test/scala/fs2/ConcurrentSpec.scala
@@ -1,6 +1,7 @@
 package fs2
 
 import TestUtil._
+import fs2.PlayGroundSpec._
 import fs2.util.Task
 import fs2.Stream.Handle
 import java.util.concurrent.atomic.AtomicLong
@@ -55,4 +56,54 @@ object ConcurrentSpec extends Properties("concurrent") {
   //property("join (failure 1)") = forAll { (s: PureStream[Failure], n: SmallPositive, f: Failure) =>
   //  run { concurrent.join(n.get)(s.get) }
   //}
+
+  // todo: make combinations of below with bracket to verify resource safety
+  property("merge.async.right.block") = protect {
+    runFor() {
+      (Stream.constant[Task,Int](42) merge Stream.repeatEval(Task.async[Unit] { cb  => () }))
+      .take(1)
+    } ?= Vector(42)
+  }
+
+  property("merge.async.left.block") = protect {
+    runFor() {
+      (Stream.repeatEval(Task.async[Unit] { cb  => () }) merge Stream.constant[Task,Int](42) )
+      .take(1)
+    } ?= Vector(42)
+  }
+
+  property("merge.async.left.block.drain") = protect {
+    runFor() {
+      (Stream.repeatEval(Task.async[Unit] { cb  => cb(Right(())) }).drain merge Stream.constant[Task,Int](42) )
+      .take(1)
+    } ?= Vector(42)
+  }
+
+  property("merge.async.right.block.drain") = protect {
+    runFor() {
+      (Stream.constant[Task,Int](42) merge Stream.repeatEval(Task.async[Unit] { cb  => cb(Right(())) }).drain)
+      .take(1)
+    } ?= Vector(42)
+  }
+
+  property("merge.async.join.block") = protect {
+    runFor() {
+      concurrent.join(10)(Stream(
+        Stream.constant[Task,Int](42)
+        , Stream.repeatEval(Task.async[Unit] { cb  => cb(Right(())) }).drain
+      ))
+      .take(1)
+    } ?= Vector(42)
+  }
+
+  property("merge.async.join.block.drain") = protect {
+    runFor() {
+      concurrent.join(10)(Stream(
+        Stream.constant[Task,Int](42)
+        , Stream.repeatEval(Task.async[Unit] { cb  => cb(Right(())) }).drain
+      ))
+      .take(1)
+    } ?= Vector(42)
+  }
+
 }
--- a/core/src/test/scala/fs2/TestUtil.scala
+++ b/core/src/test/scala/fs2/TestUtil.scala
@@ -3,11 +3,14 @@ package fs2
 import fs2.util.{Sub1,Task}
 import org.scalacheck.{Arbitrary, Gen}
 
+import scala.concurrent.duration._
+
 object TestUtil {
 
   implicit val S = Strategy.fromExecutionContext(scala.concurrent.ExecutionContext.global)
 
   def run[A](s: Stream[Task,A]): Vector[A] = s.runLog.run.run
+  def runFor[A](timeout:FiniteDuration = 3.seconds)(s: Stream[Task,A]): Vector[A] = s.runLog.run.runFor(timeout)
 
   def throws[A](err: Throwable)(s: Stream[Task,A]): Boolean =
     s.runLog.run.attemptRun match {


pchiusano:
Can you check to see if the expected elements are being emitted? Like just
add a .map(x => println(x); x).

Assuming they are being emitted okay, then this is as designed I think. At
the moment, we can't just leave an asynchronous step running when exiting a
pull, since it might at some point in the future allocate a resource, and
that resource must be freed. The result is that any outstanding async steps
must complete before the pull can complete.

I agree these examples are surprising though. I'm just not sure there is
any fix that makes sense in general. Even if it were possible to let async
steps outlive the pull where they are created, I could see that leading to
other counterintuitive behavior.

Let's give this more thought.

pchlupacek:
@pchiusano these are minimized examples which I discovered when implementing and testing fs2.tcp. I just worked with queue and this program won't ever terminate, and I think this is completely correct program. In fact in our production code we have lot of usage of the queue like that. I perhaps agree that with introduction of `closeableQueue` we may force-close the q.dequeue, that shall tear down the "downstream", but I think the fact that I cannot force-kill by `pipe` "upstream" will lead imho to lot of deadlocks and problems, and perhaps will make fs2 no-go in certain projects ( i know about concrete examples in our code where I dont think so we can live w/o it).  

(again little minimized but more `descriptive` example below)
```scala
val q = async.boundedQueue[AsyncChannel](10)
def handleConnection(ch:AsyncChannel):Stream[Task,A] = ???
val processes = 
(accepted.flatMap(ch => q.offer1(ch)).drain merge q.dequeue.map(ch => handleConnection(ch)))
.take(3)

conurrent.join(10)(processes)
```

pchlupacek:
property("merge.async.right.block") = protect {
    runFor() {
      (Stream.constant[Task,Int](42).map {l => println("L " + l); l} merge Stream.repeatEval(Task.async[Unit] { cb  => () }))
      .take(1)
      .map { x => println((">>>>", x)) ; x}
    } ?= Vector(42)
  }
output:
L 42
(>>>>,42)

So I think this is correct. We at least don't get the effects run multiple times.

pchiusano:
I have an idea for a fix. General idea is to not block on any outstanding
awaits when closing the pull. But we do need to ensure that finalizers are
run exactly once regardless. So there's a race - if the step finishes
first, the pull runs any finalizers. If the pull finishes first, it makes
the step responsible for running its own finalizers when it eventually
completes. Not sure when I'll get a chance to try this out, might be tricky
to implement.

Note that leaving a step running might very well be a bug and it might leak
memory or threads. I think that blocking on a queue should be ok, the queue
should just get gc'd along with any callback registered to it by dequeue.

pchlupacek:
Yeah, something like that may work. I think however we just hit back the whole interruption issue. I still believe the correct way is to signal upstream that they work is interrupted. Then, they have to immediately switch to cleanup phase and do whatever cleanup they need and _then_ terminate. Essentially any blocking/asynchronous code in cleanup is bug. 

I think we need something like `interruptible` Pull, that will take `ref[Boolean]` as an argument to `run` and switches immediately to cleanup phase. I am still trying to grasp internal details of new implementation, so I am not sure if this is complete non-sense, or not, but generally we may NOT allow to leak some `dangling` tasks, as you say that may lead to completely undesired behaviour. We are sort of back I think on initial task interruption behaviour definition. In 0.8 we have resolved this with exactly once semantic on finalizer, and that was original reason for `bracket` introduction. 

I am not sure if I understand your argument with blocking on queue. 

Did you mean this? I think lets assume that process shown before will be `p` then 
```
p.take(1) ++ p1  // p1 will never ever be executed when p.take(1) will not complete
p.take(1) onComplete p1  //p1 will never be executed
// but 
p.take(1) flatMap p1 // p1 will be executed


```

pchiusano:
I was thinking about this some more, and even if my fix is possible I don't
think it's really correct. You don't want to leave a task running
indefinitely, it will just lead to space or thread leaks. And you also
can't rely on garbage collection either I don't think - I could see this
leading to a finalizer not ever being run because it gets garbage collected
first. Basically this is a can of worms.

Allowing interruption is something I'd prefer not to tackle. Adding it
safely requires that you know when resources are being acquired and never
interrupt during those times. I have serious doubts this is possible to
know in general especially with the current algebra which has first class
handles and async awaits (which did not exist in 0.8) It's another can of
worms.

I think the current behavior is actually reasonable and I'd like to try
just writing some helper functions to deal with common cases. Like merge
where one side is a queue should be doable (it can allocate a local signal
or use a queue of options). Yes you are forced to be more explicit about
how to gracefully shutdown, but that isn't such a bad thing and any common
pattern can get wrapped up as a helper function.

Can you try this out and see how it goes?

pchlupacek:
@pchiusano I can try, but I think in complex programs this still will be challenging. 

Can you please outline then some invariants/laws for the termination. I have perhaps lack of imagination there, but I even don't know how to write `wye.interrupt` with current implementation, if interrupt side **never emits** or the interrupted process **never emits  last** value.  Can you probably draft that? This is very common scenario in our code. And perhaps that may lead me to play with this more and perhaps find some pattern to generalize it. 

Are you sure that dropping all they wye._ flexibility is right thing to do? I really liked the composability with p1, wye, tee. All that seems we will be missing now. Or perhaps am I missing something? 
 
I'd like to think about interruption implementation. But if we won't need it excellent. I am just a bit skeptic on this one.

pchiusano:
Can you explain what you mean by this:

> Are you sure that dropping all they wye._ flexibility is right thing to
do? I really liked the composability with p1, wye, tee. All that seems we
will be missing now. Or perhaps am I missing something?

That is a good point about interrupt. It is pretty limiting if interrupt
only terminates if it is actually ever interrupted! And I'm not sure it is
possible to implement with the current behavior. So maybe we do need to
figure out interrupts after all... :( :(

Something that might be pretty easy would be to give awaitAsync an extra
argument, an Async.Future cancellation signal that is checked periodically.
So still no interruption of the actual Task, but the overall pull step may
be interrupted. I wonder if this would be enough to implement merge /
interrupt in a way that is safely interruptible, possibly making use of
cancellableGet.

I'm going to have to spend more time with this when I get back, not sure
best way forward. Let me know if you figure anything out!

pchlupacek:
@pchiusano ok. agreed I think we are on the same page. If possible I would like to get all interruption of the user eyes, and just make it internal implementation detail. Your suggestion on passing along the Future, is something I think may work, but perhaps due lack of understanding of the new model, I think we shall be just fine with defining pull's run as perhaps : 
```scala
   def run(allowInterrupt:Option[F[Boolean]]): Stream[F,W] = ???
```
or perhaps just allow for `InterruptiblePull` as its own subtype.  Then pull will be responsible for running cleanup in case the interrupt yielded to true, and there was no value emitted. I am just curious about the issue you brought with gc, if it is really problem.  Perhaps we can just inspect for now, if compiler holds the strong reference to value after interruption happened and see if that is the problem. 

@djspiewak not sure if you can hop-in on this issue. I know we have been discussing these for long on 0.8 and perhaps you may have some solution to this??

pchlupacek:
@pchiusano I perhaps open the issue reg. the wye/tee/pipe. Just to understand your ideas better, so I can perhaps then make small writeup to explain it on wiki.

pchlupacek:
@pchiusano I little thought about it and generally I think we have the idea how to resolve the issue. How about introducing :

```scala 
 Task.asyncPreemptable[A](register: Throwable \/ A => Unit => Unit)(whenPreempted: A => Task[Unit]):Task[(A, Task.Ref[Unit]]) = ???   

// then 
 def Task.async[A](register: Throwable \/ A => Unit => Unit):Task[A] = asyncPreemptable(register)(_ => Task.now()).map(_._1)
```

Essentially `whenPreempted` will get consulted, when the task was preempted before `register` callback was called. This gives opportunity to call any cleanups or whatever else is needed. Not sure how that fits to Async[F] story yet, but any ideas? Perhaps we will have to introduce this on Async/AsyncExt ? 

I think preemption is only valid for asynchronous code, thus we shall not IMHO resolve it anywhere else where we don't know about type of the `F`. 

The above solution will cause the Pulls not to take care what will happen `after` preemption. Which I believe is sort of consistent with what we do now. They will just signal preemption, and then they either await `A` in case they run any cleanups or they just get that exception, in case they just terminate.   Not sure if all that fits together yet, but perhaps just ideas for some thoughts....

edwinchoi:
I think you guys are confusing cancellation with termination. Pre-empting for the purpose of termination is completely unsafe and bound to cause headaches for users. Cancellation on the other hand is a co-operative mechanism, which is what is what I think needs to happen here.

Ideally, the consumer should only be allowed to ask the producer to stop. The actual termination should be the responsibility of the producer and the producer alone. Any resources transferred to the producer are still owned by it as well, so the consumer should not hijack them by calling the finalizer. Completion (normal or abnormal) should be signaled through a single point (the callback).
For actors, this would be fairly trivial - the register actions sends Register and the preempt action sends a Cancel message. The actor acknowledges cancellation by invoking the callback with an appropriate message.

pchlupacek:
@edwinchoi thanks. I think always meant the `cancellation`, that is `pre-empting` the running process, but always between emits. In other words I think we not aiming to `pre-empt` task. Task shall be always atomic, or better to say Stream Step has to be always atomic operation. 

The problem imho is what to do with the step that is produced "after" we interrupted the stream. I think there are couple of solutions to it:
   - Ideally I would say this must not happen. And as such we will have to resolve it at `F` level. This is tricky however that you may flatMap, map to async and then you opening the can of worms. 
   - We may remove async from task, and make it somehow part of `Async` instance. That way we may handle this in Async.
    - Accept that there may be in future cleanup running and just make sure it will be run, and only once. 
   - ??? other alternatives

Not sure which of the alternatives are better ....

pchiusano:
I'm working on a solution for this now. I may have a solution that does not require any changes or extensions to `Async` or `AsyncExt`, though will know soon enough. @edwinchoi you are right that any interruption has to be "cooperative" - the interrupt is just a "suggestion" and it needs to be up to the stream to decide when it safe to actually stop.

pchiusano:
Made good progress on this today. See the [topic/redesign-interrupts branch](https://github.com/functional-streams-for-scala/fs2/blob/topic/redesign-interrupts/core/src/main/scala/fs2/Stream.scala). General idea is fairly simple - add an interrupt flag to the interpreter, which is consulted on every step, but don't wait for the step to complete before freeing resources *unless the step is itself allocating a resource*. So it should in theory work fine to leave a queue get running on one half of a merge. The queue and the callback registered to it will just become garbage and get gc'd.

Implementation is pretty tricky, though. I think the idea is probably legit, but I already noticed that the resource safety tests occasionally hang with zero cpu usage. I'll have to dig in further.

pchlupacek:
@pchiusano cool and thanks. I think this is the easiest, and perhaps safest  solution. That means we accept there may be "delayed" run of resource release when stream was interrupted in middle right? Is complete fine to me, we just need to document it correctly. 

Ill go through in detail it later today...

pchiusano:
@pchlupacek 

> That means we accept there may be "delayed" run of resource release when stream was interrupted in middle right?

If the step is in the middle of acquiring a resource, then yes we wait for that to complete before shutting down the awaitAsync. If the step is in the middle of doing something else (like taking the next element of a queue), then it's in an "interruptible" state and we can shut it down right away. This does mean that a computation could be left running the background until it reaches the next step of the stream interpreter main loop or gets garbage collected.

That's all assuming this works... which it doesn't currently. I'm narrowing in on why. :)

pchiusano:
closing now that this has been merged into #526

PR #521, opened 2016-01-21
Automated release process; configured for publishing snapshots and releases to Sonatype

mpilquist:
This PR sets the build up to publish snapshots and releases to Sonatype. It uses sbt-release to automate the release process, besides the final Sonatype staging repository closure/release step. Publishing snapshots to Sonatype's snapshot repository is done via `+publish-signed`. Releasing a stable build is done via `release` and then `sonatypeReleaseAll`. Instead of the last step, Sonatype's OSS Nexus UI can be used to close and release the staging repository created by the release step.

The generated POMs *should* pass Sonatype's validation rules, as I based the configuration off the scodec build. We won't know until we try to build a release though.

Note that this PR removes git based versioning in favor of snapshot versioning. I realize this may be controversial, though I think it is important to publish real snapshots to Sonatype (and I did discuss this with @pchiusano today). Note that fs2 is not publishing to bintray at all. Hence, any non-snapshot version strings require Sonatype repository closure/release steps.

I think we could add git-based versioning back in the future if anyone feels strongly about publishing "stable snapshots" in addition to the normal/unstable snapshots. Alternatively, we could also publish stable snapshots using milestone and rc qualifiers (e.g., `0.9.0-M1`, `0.9.0-RC1`), which seems more common in Central than git versioning. Also, sbt-release supports milestone/rc qualifiers out of the box -- there's no special integration effort required.

/cc @djspiewak

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,7 +12,7 @@ lazy val contributors = Seq(
 )
 
 lazy val commonSettings = Seq(
-  organization := "fs2",
+  organization := "co.fs2",
   scalaVersion := "2.11.7",
   scalacOptions ++= Seq(
     "-feature",
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,16 @@
 val ReleaseTag = """^release/([\d\.]+a?)$""".r
 
+lazy val contributors = Seq(
+  "pchiusano" -> "Paul Chiusano",
+  "pchlupacek" -> "Pavel Chlupáček",
+  "alissapajer" -> "Alissa Pajer",
+  "djspiewak" -> "Daniel Spiewak",
+  "fthomas" -> "Frank Thomas",
+  "runarorama" -> "Rúnar Ó. Bjarnason",
+  "jedws" -> "Jed Wesley-Smith",
+  "mpilquist" -> "Michael Pilquist"
+)
+
 lazy val commonSettings = Seq(
   organization := "fs2",
   scalaVersion := "2.11.7",
@@ -24,7 +35,7 @@ lazy val commonSettings = Seq(
     import fs2.util._
   """,
   doctestWithDependencies := false
-) ++ testSettings ++ scaladocSettings ++ gitSettings
+) ++ testSettings ++ scaladocSettings ++ publishingSettings ++ releaseSettings
 
 lazy val testSettings = Seq(
   parallelExecution in Test := false,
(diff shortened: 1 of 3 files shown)


pchlupacek:
@mpilquist excelent and thans a lot for this. We may eventually at git hashtag to file VERSION laters if hastags are required.

build.sbt
```diff
@@ -1,7 +1,18 @@
...
+  "djspiewak" -> "Daniel Spiewak",
+  "fthomas" -> "Frank Thomas",
+  "runarorama" -> "Rúnar Ó. Bjarnason",
+  "jedws" -> "Jed Wesley-Smith",
+  "mpilquist" -> "Michael Pilquist"
+)
+
 lazy val commonSettings = Seq(
-  organization := "fs2",
+  organization := "co.fs2",
```
pchlupacek: is the co. prefix really valid one?

mpilquist: Yep (see the project URL here: https://issues.sonatype.org/browse/OSSRH-19932)

pchiusano:
Lgtm. @djspiewak please pipe up in the next 48 hrs if you have objections. We will merge this weekend otherwise.

djspiewak:
I do have objections, but if the consensus is against me, I'll yield.  Stable snapshots are really, really, really important.  Unstable snapshots are basically a crapshoot, and that is precisely why they are now significantly discouraged in Maven.  I'm not sure why Sonatype itself is putting up obstacles in the way of stable (and useful) snapshot identifiers, because that does fly in the face of their own advice to Maven users, but it wouldn't be the first time they've done something weird.

To take a step back, the whole reason to have stable snapshot identifiers is so that people can make use of them beyond the literally trivial use cases.  Stable builds are really important, and if one of your upstream dependencies can change from build run to build run, you basically can't use that dependency.  Stable snapshots, on the other hand, are useful versions that simply haven't been tested/stabilized to the degree that a numbered version has been.  I would argue that unstable snapshots for a middleware library really aren't all that useful, since people cannot safely incorporate them into builds or really do anything with them other than use it as a way of obtaining a pre-built copy of master (whatever master happens to be at the moment).

build.sbt
```diff
@@ -42,23 +53,50 @@ lazy val scaladocSettings = Seq(
...
+    <scm>
+      <url>git@github.com:functional-streams-for-scala/fs2.git</url>
+      <connection>scm:git:git@github.com:functional-streams-for-scala/fs2.git</connection>
+    </scm>
+    <developers>
+      {for ((name, username) <- contributors) yield
+      <developer>
+        <id>{username}</id>
+        <name>{name}</name>
+        <url>http://github.com/{username}</url>
```
xuwei-k: `<url>http://github.com/{name}</url>` ?

mpilquist: Good catch -- the tuple extractor is backwards. Should be `for ((username, name) <- ...`

pchiusano:
He lives!! :)

Everything you said makes sense, but I don't see why it would matter whether stable snapshots are via git hash or via milestones like `0.9.0-M1`, which we can also publish as often as we like (is that right @mpilquist? And what would be the process for publishing a named milestone like this?). And having the version names be 'human readable' seems like it increases the odds that downstream projects will congregate around depending on a particular one. Whereas if a bunch of downstream projects are depending on random hashes you wouldn't be able to use them all together anyway.

The totally unstable snapshot is still useful in some cases, if you aren't pulling in any downstream projects and just want to float with the latest. Obviously, you shouldn't use a setup like this for production builds, but I think people know that.

mpilquist:
> is that right @mpilquist? 

Yep

> And what would be the process for publishing a named milestone like this?

`sbt release` -- when prompted for a version number to release, you'd enter "0.9.0-M1". When prompted for the next version number, you'd enter "0.9.0-SNAPSHOT". These types of things can be further automated via sbt settings or aliases, but I've never felt the need to do so.

pchiusano:
@mpilquist that seems pretty painless

djspiewak:
> Everything you said makes sense, but I don't see why it would matter whether stable snapshots are via git hash or via milestones like 0.9.0-M1, which we can also publish as often as we like (is that right @mpilquist? And what would be the process for publishing a named milestone like this?). And having the version names be 'human readable' seems like it increases the odds that downstream projects will congregate around depending on a particular one. Whereas if a bunch of downstream projects are depending on random hashes you wouldn't be able to use them all together anyway.

The problem here is that you have to explicitly choose to publish such a snapshot.  If I'm a downstream user and I need a snapshot at a particular point, I have to *ask* you to publish it, and you may not be willing (e.g. you may want to wait for one more change, or you just published a previous snapshot and don't want to push a new one just yet).  Hashes are great because there can be one unambiguous snapshot for every single commit, if necessary.

Hash versions are also nice because it's easy to "reverse correlate" back to the repository state.  I can take any hash snapshot version together with knowledge of the github repo and literally type out a URL that points to the exact state of the codebase when that snapshot was cut.  This is surprisingly useful.

I do agree that human readable names do have some advantages.  Notably, following a standard and monotonic version scheme also gives SBT the ability to perform dependency eviction, which further increases the ability to congregate around a particular snapshot.  I guess the place where this would be particularly useful is other middleware frameworks, like http4s or specs2.  For projects that depend directly on fs2, it's not as much of an issue since you can just select whatever version you want.

> The totally unstable snapshot is still useful in some cases, if you aren't pulling in any downstream projects and just want to float with the latest. Obviously, you shouldn't use a setup like this for production builds, but I think people know that.

I can't think of a time that I've actually wanted to float with the latest, honestly.  That's my point with the stable snapshots.  Expanding to other projects that I've worked on, the only time I've *ever* needed this sort of feature was when working on a multi-project build where all the projects were moving rapidly (this was also prior to some of SBT's newer external project features).

pchiusano:
> The problem here is that you have to explicitly choose to publish such a snapshot. 

I don't necessarily see this as a bad thing actually. But in any case, we could if we wished automate publishing of milestones. If we were to automate it, I'd probably want to do it on a weekly or even monthly basis, not every commit.

> Hash versions are also nice because it's easy to "reverse correlate" back to the repository state.

We'll have the tag names in git, added by `sbt release`. Bonus is that the URLs are human readable. :)

Okay at this point I think I understand all the arguments, and I'd like to merge this. @djspiewak even though we didn't go with your preference I still appreciate you weighing in!

build.sbt
pchiusano: @mpilquist could you just fix this, add it to the PR, then we'll merge?

djspiewak:
> Okay at this point I think I understand all the arguments, and I'd like to merge this. @djspiewak even though we didn't go with your preference I still appreciate you weighing in!

:-)

mpilquist:
Merged and snapshots are published under: `"co.fs2" %% "fs2-core" % "0.9.0-SNAPSHOT"`

pchiusano:
Great, thanks for hooking all this up.

PR #522, opened 2016-01-21
Added ability to run a Task as a stdlib Future

mpilquist:
Getting the controversial PRs out of the way tonight... :smile: 

This pattern comes up a lot when integrating `Task` with other libraries. For example, completing a spray/akka-http route with a `Task` can be accomplished by running the route as a future and then completing with the future.


pchlupacek:
cool

PR #523, opened 2016-01-21
Moved Step.#: to fs2 package to simplify imports

mpilquist:


pchlupacek:
I think this completely makes sense.  @pchiusano you ok with this?

pchiusano:
Looks good.

PR #526, opened 2016-01-22
Fix for interrupt handling to allow for "hanging awaits" on closing a Pull

pchiusano:
This addresses the tests in #516, which I've merged in and cleaned up some. Basic idea is:

* When a `Pull` is being shut down, there may be `awaitAsync` steps still running. If any of those running steps are in the middle of a resource acquisition, we wait until any such acquisitions complete, then interrupt them.
* On shutdown, we run all outstanding finalizers and simultaneously _close the resources tracker_. (See the `Resources` data type in `internal`.) This ensures no new resources can be acquired that we don't know about.
* Any resource acquisition checks the resource tracker to make sure it isn't closed. If it is closed, it fails with `Interrupted` before starting to acquire any resources.

So, you can do a `merge` or `interrupt` where one side never produces a value, and that works fine now.

Somewhat related to this, I made `Task.await` take a `Strategy`. This forces you to be more explicit that you really want the callback to run in the same thread, or not. There's a `Task.unforkedAsync` if you want the old behavior.

I noticed a case that still hangs that wasn't in @pchlupacek's tests, but I'd like to understand it before merging. I'm not entirely sure it is a problem. Some random thoughts:

```Scala
val hang = Stream.constant[Task,Int](42).drain
```

If you do an `awaitAsync` on this, I think it will fully occupy a thread, providing no opportunity for interruption. It will just run until the end of time, never giving up that thread. In a commented out test, I think this phenomenon is causing deadlock due to running out of threads. I might have an idea for how to improve on this without compromising safety. I'll play around with this next week. Let's hold off on merging until then.

Other notes: I originally implemented a different approach (same basic idea) in the topic/redesign-interrupt branch. It was buggy and too complicated. I like this new implementation much better!

### DIFF
--- a/core/src/main/scala/fs2/internal/Future.scala
+++ b/core/src/main/scala/fs2/internal/Future.scala
@@ -117,7 +117,7 @@ private[fs2] sealed abstract class Future[+A] {
       , timeoutInMillis, TimeUnit.MILLISECONDS)
 
       runAsyncInterruptibly(a => if(done.compareAndSet(false,true)) cb(Right(a)), cancel)
-    }
+    } (Strategy.sequential)
 
   def timed(timeout: Duration)(implicit S: ScheduledExecutorService):
     Future[Either[Throwable,A]] = timed(timeout.toMillis)
@@ -144,8 +144,8 @@ private[fs2] object Future {
 
   def suspend[A](f: => Future[A]): Future[A] = Suspend(() => f)
 
-  def async[A](listen: (A => Unit) => Unit): Future[A] =
-    Async((cb: A => Trampoline[Unit]) => listen { a => cb(a).run })
+  def async[A](listen: (A => Unit) => Unit)(implicit S: Strategy): Future[A] =
+    Async((cb: A => Trampoline[Unit]) => listen { a => S { cb(a).run } })
 
   /** Create a `Future` that will evaluate `a` using the given `ExecutorService`. */
   def apply[A](a: => A)(implicit S: Strategy): Future[A] = Async { cb =>
--- a/core/src/main/scala/fs2/util/Task.scala
+++ b/core/src/main/scala/fs2/util/Task.scala
@@ -259,13 +259,20 @@ object Task extends Instances {
     ref[A].flatMap { ref => ref.set(t) map (_ => ref.get) }
 
   /**
+   Like [[async]], but run the callback in the same thread in the same
+   thread, rather than evaluating the callback using a `Strategy`.
+   */
+  def unforkedAsync[A](register: (Either[Throwable,A] => Unit) => Unit): Task[A] =
+    async(register)(Strategy.sequential)
+
+  /**
    Create a `Task` from an asynchronous computation, which takes the form
    of a function with which we can register a callback. This can be used
    to translate from a callback-based API to a straightforward monadic
-   version.
+   version. The callback is run using the strategy `S`.
    */
-  def async[A](register: (Either[Throwable,A] => Unit) => Unit): Task[A] =
-    new Task(Future.async(register))
+  def async[A](register: (Either[Throwable,A] => Unit) => Unit)(implicit S: Strategy): Task[A] =
+    new Task(Future.async(register)(S))
 
   /** Create a `Task` that will evaluate `a` after at least the given delay. */
   def schedule[A](a: => A, delay: Duration)(implicit pool: ScheduledExecutorService): Task[A] =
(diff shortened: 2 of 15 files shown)


pchiusano:
Damn, just noticed Travis is failing. I'll have to look into that, those tests "worked on my machine". :)

pchiusano:
Is anyone else able to reproduce the error that Travis is reporting? This is going to be a nightmare to debug if I cannot reproduce locally. [That test that is failing](https://travis-ci.org/functional-streams-for-scala/fs2/jobs/104195151#L518) passes every time for me.

pchiusano:
Aha, got this to fail locally a few times by adding some sleeps to the two streams being merged. Not very reliably. It is highly dependent on the timing, seems. I'll do some debugging.

pchlupacek:
```scala 
val hang = Stream.constant[Task,Int](42).drain
```
I remember we had issue with these when we refactored to Append. Essentially you never had chance to interrupt these, as the `step` was always trying to `emit` something but drain caused them to not emit anything so there was no chance to interrupt. I will be curious if `Pull` is sort of `leaving the cycle` to give a chance for interrupt?

core/src/main/scala/fs2/util/Task.scala
```diff
@@ -259,13 +259,20 @@ object Task extends Instances {
...
+  /**
    Create a `Task` from an asynchronous computation, which takes the form
    of a function with which we can register a callback. This can be used
    to translate from a callback-based API to a straightforward monadic
-   version.
+   version. The callback is run using the strategy `S`.
    */
-  def async[A](register: (Either[Throwable,A] => Unit) => Unit): Task[A] =
-    new Task(Future.async(register))
+  def async[A](register: (Either[Throwable,A] => Unit) => Unit)(implicit S: Strategy): Task[A] =
```
pchlupacek: @pchiusano I think we will have to little bit streamline here. with recent PR by @mpilquist I wonder what you really do there is sort of `runFullyAsync`. I think it will be really mess for users what is `unforkedAsync`, `async` and `runFullyAsync` and their differences. Not sure if I undestand whats the difference between `runFullyAsync` and `async` ?

core/src/main/scala/fs2/internal/Future.scala
```diff
@@ -144,8 +144,8 @@ private[fs2] object Future {
 
   def suspend[A](f: => Future[A]): Future[A] = Suspend(() => f)
 
-  def async[A](listen: (A => Unit) => Unit): Future[A] =
-    Async((cb: A => Trampoline[Unit]) => listen { a => cb(a).run })
+  def async[A](listen: (A => Unit) => Unit)(implicit S: Strategy): Future[A] =
+    Async((cb: A => Trampoline[Unit]) => listen { a => S { cb(a).run } })
```
pchlupacek: +1 I think we perhaps have to visit the Async code, as some of the S() wrapped callbacks will be redundand now.

pchiusano:
All right, this is actually all working now, as far as I can tell. And I'm more convinced of its correctness than I ever have been before (not sure how much that says, heh.) Some other notes:

* The tests now use `Strategy.fromFixedDaemonPool(8)` (new convenience method, also added `Strategy.fromCachedDaemonPool`), to avoid tests being sensitive to the number of cores on whatever machine you or Travis happens to be using.
* I've ratcheted up the number of test cases to 500 for the time being out of paranoia. There were some bugs that were timing sensitive and only showed up based on how threads interleaved. While testing locally, I was using 20,000 test runs, but the full test suite would take eons if that many tests were run.
* @pchlupacek this branch now correctly handles the "hanging awaits" cases. See the "hanging awaits" properties in `ConcurrentSpec` where that is tested. I have also verified that for those tests, the streams are in fact terminated---they don't run in the background indefinitely in some some zombie state. A very visible way to test this is just to add a `s.map { x => println(x); x }` just before the `.drain`.
* I spotted and fixed a bug in `concurrent.join` that I think explained some of the behavior I was seeing before---the implementation of `join` was a little too eager. If it was in the state of no streams yet open, it would open the first stream _and then block to read that stream's first element_. This doesn't affect correctness in most cases but leads to a hang if the first stream is something like `Stream.constant(1).drain`! The corrected behavior is just to open the stream and add it to the open set, then recurse.
* The use of `ReentrantLock` on `internal.Ref`, though correct, seems like overkill. Although the window of contention is pretty small, when it does happen it seems crazy to be putting one or more threads to sleep for a nanosecond and then waking them back up. I speculate that retrying immediately will be faster given how narrow the window of contention is. It is also kind of weird to have concurrent _readers_ competing for a lock. The proper semantics would be to allow multiple concurrent readers, or a single writer. Maybe `ReentrantReadWriteLock` would be more appropriate. @mpilquist I think this could be worth fixing in a separate PR, WDYT?

Review by @pchlupacek and @mpilquist.

Once this is merged, we should be good to go!! I very much look forward to not having to look at or even think about the `Stream` interpreter for a very long time... :)

pchlupacek:
cool and thanks !

pchlupacek:
@pchiusano I have review it and if @mpilquist does not have objections LGTM. On the reentrant lock I really think we should not have that used, and I think there is simple solution on replacing RL with `AtomicStampedReference[A]` which from my understanding does what we want in non-blocking way. @mpilquist lmk if you can take this...

mpilquist:
:+1:

PR #527, opened 2016-01-23
Topic/redesign tcp

pchlupacek:
@pchiusano this is WIP on merging `tcp` to fs2. Once the interruption will be fixed, I will continue on the specs and perhaps implement udp module as well. 

Can you perhaps take a quick look on the api in the meantime. I did just cosmetic changes to what we have in 0.8 release.

Let's leave this open until tests are complete?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -71,19 +71,22 @@ lazy val core = project.in(file("core")).
   enablePlugins(GitVersioning).
   settings(commonSettings).
   settings(
-   name := "fs2-core"
+    name := "fs2-core"
   )
 
 lazy val io = project.in(file("io")).
   enablePlugins(GitVersioning).
   settings(commonSettings).
   settings(
-   name := "fs2-io"
+    name := "fs2-io",
+    libraryDependencies ++= Seq(
+      "org.scodec" %% "scodec-bits" % "1.0.12"
+    )
   ).dependsOn(core % "compile->compile;test->test")
 
 lazy val benchmark = project.in(file("benchmark")).
   enablePlugins(GitVersioning).
   settings(commonSettings).
   settings(
-   name := "fs2-benchmark"
+    name := "fs2-benchmark"
   ).dependsOn(io)
--- a/io/src/main/scala/fs2/io/tcp.scala
+++ b/io/src/main/scala/fs2/io/tcp.scala
@@ -1,0 +1,527 @@
+package fs2.io
+
+import java.net.{InetSocketAddress, StandardSocketOptions}
+import java.nio.ByteBuffer
+import java.nio.channels.spi.AsynchronousChannelProvider
+import java.nio.channels.{AsynchronousChannelGroup, AsynchronousServerSocketChannel, AsynchronousSocketChannel, CompletionHandler, InterruptedByTimeoutException}
+
+import fs2._
+import fs2.util.{Task, ~>}
+import scodec.bits.ByteVector
+
+import scala.concurrent.duration._
+
+/**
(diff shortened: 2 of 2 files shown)


pchiusano:
Hey thanks for starting on this. I had an idea for a more radical redesign, getting rid of the Connection effect and just using Task directly, or actually any Async.

Will try to do writeup soon.

pchlupacek:
@pchiusano cool.  Lmk.

pchiusano:
Okay, here is design I was thinking of:

```Scala
trait Socket[F[_]] {
   def available(maxBytes: Int, timeout: Option[FiniteDuration] = None): Pull[F, Nothing, Option[ByteVector]]
  // etc, all functions return a `Pull`
}

def client[F[_]:Async](paramsToConfigClientSocket): Pull[F, Nothing, Socket[F]] // use `Pull.acquire` to impl
def server[F[_]:Async](paramsToConfigServerSocket): Stream[F, Pull[F,Nothing,Socket[F]]]
```

There can be various helper functions derived from these. For `server`. 

In theory, this API can be misused if you were to stash a copy of the `Socket` somewhere, but I don't think it is worth preventing that via types. It leads to more boilerplate, reduces the expressiveness of the API (some common cases just aren't expressible with the `Connection` effect), and forces us to reimplement lifted versions of various functions, which is always going to be incomplete. Stashing the `Socket` is just not a very common problem, and anyway, we can expose safe combinators for common use cases (and people could experiment with their own APIs on top of this one). I'd like to keep this API simple and fairly low level.

pchiusano:
Also I really want to make this generic for any `Async`. I think it can be done using #528.

@pchlupacek WDYT? :)

pchlupacek:
@pchiusano I take this on myself. I see the old signature was pretty confusing at first point, but then, when you used it its straightforward imho. Anyhow, LMT on the Pull variant and I see if I can come up with something. I agree that types shall not completely block you from bad code, if we document it clearly. I may come with something till next weekend. This blocks me in starting to use fs2 in our code and for upgrading kafka, cassandra et all drivers to fs2. So it is quite priority now. 
Perhaps @mpilquist what are you thoughts? I think users of your library come across this need as well no?

pchiusano:
> @pchiusano I take this on myself. I see the old signature was pretty confusing at first point, but then, when you used it its straightforward imho. 

Okay great. Yeah I think the old signature looked okay at the usage sites, but it was really doing too many things at once. I like this new model better.

> Anyhow, LMT on the Pull variant and I see if I can come up with something.

I'm not sure if I know what you mean. This might help though:

```Scala
  def acquire[F[_],R](r: F[R])(cleanup: R => F[Unit]): Pull[F,Nothing,R] =
    Stream.bracket(r)(Stream.emit, cleanup)
          .open.flatMap { h => h.await1.map(_.head) }
```

That can be a public function. You can use it anytime you need to acquire a new resource from inside a `Pull`. The cleanup will be run at the end of whatever `Pull.run` scope it ends up in. So if you do:

```Scala
Pull.acquire(res, cleanup).flatMap { r => /** some stuff */ }
      .run
```

The cleanup of `r` will run at the end, after the `/* some stuff */`, and the end of the stream produced by `.run`.

There is another `Pull.acquire` function which takes three params, but that is not meant to be a public API function. You could also call that directly, just allocate a new `Token` inside of a `Pull.suspend`. The implementation I gave above will end up translating to the same thing though.

Okay, let me know how it goes!

pchlupacek:
@pchiusano I am closing it hence there will be new PR with Pull soon

PR #528, opened 2016-01-23
Implementation of  `async` for any `Async[F]`

pchiusano:
Usage is pretty much identical to what you are used to, but now it works for any `Async`:

```Scala
scala>   implicit val S = Strategy.fromExecutionContext(scala.concurrent.ExecutionContext.global)
S: fs2.Strategy = fs2.Strategy$$anon$1@467d308b

scala> val T = implicitly[Async[Task]]
T: fs2.Async[fs2.util.Task] = fs2.util.Instances$$anon$1@28345a08

// notice this is not calling `Task.async`, but the one we get for free for any `Async[F]`
scala> T.async[Int] { cb => Thread.sleep(1000); cb(Right(42)) }
res0: fs2.util.Task[Int] = fs2.util.Task@16948379

scala> res0.run
// after 1 second pause
res1: Int = 42
```

The signature is just slightly different from the existing `Task.async`, but it is used in the same way. You can see the implementation is pretty simple. It just creates a `Ref`, registers a callback which will set that `Ref`, then blocks on the `Ref`:

```Scala
  def async[A](register: (Either[Throwable,A] => F[Unit]) => F[Unit]): F[A] =
    bind(ref[Either[Throwable,A]]) { ref =>
    bind(register { e => setPure(ref)(e) }) { _ =>
    bind(get(ref)) { _.fold(fail, pure) }
    }}
```

Although the signature seems a bit different, it is used in the same way. From the perspective of the callback-y API, it will be given a callback, which it just has to invoke. The returned `F[Unit]` could just be `Unit` instead, but this makes it more clear that there is an effect to indicate that the registering of the callback has been successfully completed or scheduled. It also prevents some boneheaded mistakes due to Scala silently filling in `Unit`. It might not be a bad idea to modify `Task.async` similarly.

Thought this would be useful for #527, which I think can now be generalized to any `Async`.

### DIFF
--- a/core/src/main/scala/fs2/Async.scala
+++ b/core/src/main/scala/fs2/Async.scala
@@ -41,6 +41,18 @@ trait Async[F[_]] extends Catchable[F] { self =>
    * Like `get`, but returns an `F[Unit]` that can be used cancel the subscription.
    */
   def cancellableGet[A](r: Ref[A]): F[(F[A], F[Unit])]
+
+  /**
+   Create an `F[A]` from an asynchronous computation, which takes the form
+   of a function with which we can register a callback. This can be used
+   to translate from a callback-based API to a straightforward monadic
+   version.
+   */
+  def async[A](register: (Either[Throwable,A] => F[Unit]) => F[Unit]): F[A] =
+    bind(ref[Either[Throwable,A]]) { ref =>
+    bind(register { e => setPure(ref)(e) }) { _ =>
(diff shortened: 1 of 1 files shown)


pchlupacek:
@pchiusano yes! this is what I have been looking for. Excellent. I believe this will help us in other places as well.

pchiusano:
cool :)

Issue #329, opened 2015-03-13
I'm pretty sure tcp doesn't work at all


djspiewak:
There are a ton of issues with the API and implementation and also tests.  It's very unclear to me *if* it's working at all, and if so, then *how*.  I'm relatively certain that, at the very least, the resource management is not guaranteed, since interrupts during asynchronous evaluation can result in leaks and/or spurious errors (even with the old `runAsync`).  In no particular order of importance…

## `tcp.server` signature is deceptive

The simplified signature of this function is as follows:

```scala
def server[A](bind: InetSocketAddress, concurrentRequests: Int)(handler: Process[Connection,A]): Process[Task, Process[Task, Throwable \/ A]]
```

So this sort of promises the same thing as the old NIO API, which is that the outer process manages the server instance and the inner processes represent clients that have been handled.  This makes a lot of sense, *however* the `concurrentRequests` parameter is in conflict with all of this!  The only way that this signature would make sense in the presence of this parameter and what it implies is if the return type were a flattened `Process[Task, Throwable \/ A]`.  We can see quite clearly *why* this is the case if we dig into the implementation:

```scala
setup map { sch =>
      nondeterminism.njoin(concurrentRequests, maxQueued) {
        Process.repeatEval(accept(sch)).map(processRequest)
      }.onComplete(Process.eval_(Task.delay { sch.close }))
    }
```

The `setup` value has the following type: `Process[Task, AsynchronousServerSocketChannel]`.  So ok, that's the *outer* process.  The inner processes are derived from an an infinite process of ongoing accepts, mapped into `processRequest` (which produces a process).  However, notice what we're doing with that infinite stream of accepts: we're *flattening* it with `njoin`!

In other words, going back to the original type signature, our outer `Process` of `Process` does not contain an inner `Process` for every single client connection, but rather it contains a *single* `Process` which carries the output of *all accepted connections* in a completely random order.  The inner process has no particular resource management associated with it, and thus its unflattened existence is entirely spurious.

Put another way, this implementation is doing the equivalent of creating a `Process[Task, Throwable \/ A]` and then running it through `Process.emit` as the last step of the function, just to satisfy a type signature.  This in turn leads to users doing very strange things on the other end.  More on this in a bit…

## Resources are improperly acquired

[...]

PR #529, opened 2016-01-24
Added basic implementation of io.tcp

pchlupacek:
@pchiusano can you review this variant (re #527 )? Tx. If you ok with this, I gonna work on tests once #526 will be done

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -116,7 +116,10 @@ lazy val core = project.in(file("core")).
 lazy val io = project.in(file("io")).
   settings(commonSettings).
   settings(
-    name := "fs2-io"
+    name := "fs2-io",
+    libraryDependencies ++= Seq(
+      "org.scodec" %% "scodec-bits" % "1.0.12"
+    )
   ).dependsOn(core % "compile->compile;test->test")
 
 lazy val benchmark = project.in(file("benchmark")).
--- a/core/src/main/scala/fs2/Async.scala
+++ b/core/src/main/scala/fs2/Async.scala
@@ -30,7 +30,9 @@ trait Async[F[_]] extends Catchable[F] { self =>
    */
   def set[A](q: Ref[A])(a: F[A]): F[Unit]
   def setFree[A](q: Ref[A])(a: Free[F,A]): F[Unit]
-  def setPure[A](q:Ref[A])(a:A):F[Unit] = set(q)(pure(a))
+  def setPure[A](q:Ref[A])(a: A): F[Unit] = set(q)(pure(a))
+  /** Actually run the effect of setting the ref. Has side effects. */
+  protected def runSet[A](q: Ref[A])(a: Either[Throwable,A]): Unit
 
   /**
    * Obtain the value of the `Ref`, or wait until it has been `set`.
@@ -48,11 +50,9 @@ trait Async[F[_]] extends Catchable[F] { self =>
    to translate from a callback-based API to a straightforward monadic
    version.
    */
-  def async[A](register: (Either[Throwable,A] => F[Unit]) => F[Unit]): F[A] =
-    bind(ref[Either[Throwable,A]]) { ref =>
-    bind(register { e => setPure(ref)(e) }) { _ =>
-    bind(get(ref)) { _.fold(fail, pure) }
-    }}
+  def async[A](register: (Either[Throwable,A] => Unit) => F[Unit]): F[A] =
+    bind(ref[A]) { ref =>
+    bind(register { e => runSet(ref)(e) }) { _ => get(ref) }}
 }
 
 object Async {
--- a/core/src/main/scala/fs2/PullDerived.scala
+++ b/core/src/main/scala/fs2/PullDerived.scala
@@ -1,5 +1,6 @@
 package fs2
 
+private[fs2]
 trait PullDerived { self: fs2.Pull.type =>
 
   def map[F[_],W,R0,R](p: Pull[F,W,R0])(f: R0 => R): Pull[F,W,R] =
@@ -21,4 +22,8 @@ trait PullDerived { self: fs2.Pull.type =>
   }
 
   implicit def covaryPure[F[_],W,R](p: Pull[Pure,W,R]): Pull[F,W,R] = p.covary[F]
+
+  implicit class PullNoOutputOps[F[_],R](self: Pull[F,Nothing,R]) {
+    def output: Pull[F,R,Unit] = self.flatMap(Pull.output1)
+  }
 }
--- a/core/src/main/scala/fs2/PullOps.scala
+++ b/core/src/main/scala/fs2/PullOps.scala
@@ -5,6 +5,19 @@ import fs2.util.{RealSupertype,Sub1}
 private[fs2]
 trait PullOps[+F[_],+W,+R] { self: Pull[F,W,R] =>
 
+  /** Definition: `p as r == p map (_ => r)`. */
+  def as[R2](r: R2): Pull[F,W,R2] = self map (_ => r)
+
+  def covary[F2[_]](implicit S: Sub1[F,F2]): Pull[F2,W,R] = Sub1.substPull(self)
+
+  def flatMap[F2[_],W2>:W,R2](f: R => Pull[F2,W2,R2])(implicit S: Sub1[F,F2], T: RealSupertype[W,W2]): Pull[F2,W2,R2] =
+    Pull.flatMap(self.covary[F2]: Pull[F2,W2,R])(f)
+
+  def filter(f: R => Boolean): Pull[F,W,R] = withFilter(f)
+
+  def map[R2](f: R => R2): Pull[F,W,R2] =
+    Pull.map(self)(f)
+
   def optional: Pull[F,W,Option[R]] =
     self.map(Some(_)).or(Pull.pure(None))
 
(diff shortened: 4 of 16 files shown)


pchiusano:
This looks like exactly what I had in mind, so looks good to me. :) You busted this out pretty quickly!

Minor thing, I would rename TCPSocket to just Socket. It already lives in the io.tcp package so people can refer to it qualified if it would be otherwise ambiguous.

I probably won't get to the interrupt debugging on #526 until later this week, FYI.

mpilquist:
:+1:

pchiusano:
@pchlupacek now that #526 is merged, could you work on tests for this? API seems good to me, but if you notice any weirdness with the API while writing the tests let us know.

pchlupacek:
@pchiusano working on this hope to have them done by end of this weekend.

pchiusano:
Great!

pchiusano:
Just linking in discussion of #329. I have high hopes for this new implementation. :)

pchlupacek:
@pchiusano I moved a step further. I'm not quite shure whats going on there. It seems the cleanup is not called on client now, preventing channel to close and server to finalize the connection and emit the values read 
I just pushed this and here is output I see. Note the fact that client stream called onComplete, but there is no `cleanup` of the client code called. 
```
ECHO STARTED
(SERVER AWITING BYTES TO READ,0,fs2.io.tcp.Socket$$anon$3@13c22fd2)
(CLIENT WROTE,ByteVector(4 bytes, 0x01020304))
(XXXY READING,1024,sun.nio.ch.UnixAsynchronousSocketChannelImpl[connected local=/127.0.0.1:9869 remote=/127.0.0.1:55330],None)
(XXXY READING,4,sun.nio.ch.UnixAsynchronousSocketChannelImpl[connected local=/127.0.0.1:55330 remote=localhost/127.0.0.1:9869],None)
(XXXG AVAIL >>>>> COMPLETE,1024,sun.nio.ch.UnixAsynchronousSocketChannelImpl[connected local=/127.0.0.1:9869 remote=/127.0.0.1:55330],4)
( ON SERVER,Some(ByteVector(4 bytes, 0x01020304)))
(XXXG AVAIL >>>>> COMPLETE,4,sun.nio.ch.UnixAsynchronousSocketChannelImpl[connected local=/127.0.0.1:55330 remote=localhost/127.0.0.1:9869],4)
(CLIENT READ,ByteVector(4 bytes, 0x01020304))
(SERVER AWITING BYTES TO READ,4,fs2.io.tcp.Socket$$anon$3@13c22fd2)
(XXXY READING,1024,sun.nio.ch.UnixAsynchronousSocketChannelImpl[connected local=/127.0.0.1:9869 remote=/127.0.0.1:55330],None)
(XXX OUT,Right(ByteVector(4 bytes, 0x01020304)))
CLIENT DONE

```

Now if you force-close client  (uncomment it) after reading the bytes, then the server correctly signals the end of input, but again no cleanup called, and no count out bytes processed (I had to add `output` to Pull).

Just a small note. Perhaps out of 20 tries, i got that cleanup on client fired once, so this seems to me like kind of race we hitting there.

pchlupacek:
@pchiusano just one observation. This actually hangs repeating calling `cancellableGet` and spiking about 2 cpus. When i dig to actor this seems to be  `Right(Left(java.nio.channels.AcceptPendingException))` be the issue. So i think there is something weird in cleanup code ?

pchlupacek:
@pchiusano ok I think I spotted that issue but not really sure how to go around it. it is in "handleIncoming" server code that uses flatMap on constant stream. Essentially this produce endless streams of pulls that eventually accept connection. Not sure if that is completely bad. I tried attempt with repeating one stream pull, but there I am struggling to get effects `F` run in all cases. I may be missing something here ?

pchiusano:
@pchlupacek a couple things

1. Is there a protocol bug here? The server is waiting for the client to terminate the stream, but the client never signals it is done writing. I think you may just need a `>> socket.shutdownOutput` at the end of [this line](https://github.com/functional-streams-for-scala/fs2/blob/topic/redesign-tcp-pull/io/src/test/scala/fs2/io/tcp/SocketSpec.scala#L73) in the client? As an aside, I know this is just a test, but protocols that use end-of-stream to signal something are really fragile IMO, and aren't composable either (if you want to embed them in some larger protocol). Better to make things self-delimiting. I wonder if there is a way we could prevent this sort of error though...
2. `onComplete` *cannot* be reliably used to detect the end of a stream, since if the code is being run asynchronously, it can be cut off at any time. The only thing guaranteed to be run (assuming no bugs in FS2 :) ) is the cleanup action associated with any `bracket`/`acquire` calls. I don't think that matters here, but more of an FYI. For debugging I would just add calls to `bracket` (to attach finalizers just for purposes of debugging) or put printlns in the existing finalizers themselves.
3. I think it would be a good idea to associate a (private) `open: AtomicBoolean` with each `Socket`, and setting it to `false` in the cleanup action. Then check that flag before doing any of the other methods with the socket, and bomb with a meaningful error ("attempt to use socket after it has been closed - a `Socket` is valid only within the `Pull` where it was acquired. After calling `run` on that `Pull` it is closed. It is a programming error for a `Pull.run` call to produce a `Stream[F,Socket[F]]`.") I'm not sure if that is relevant here but I would do it just to eliminate that as a source of errors.

pchlupacek:
@pchiusano thanks. I moved little bit in forward having simple echo / requst-reply scenario done. However need to clarification of these on server api. When closing, The Pull may be actually pulling an channel from accept. What shall be the pull result to terminate this?. The api ends with exception in that case, that has to be converted to normal termination. I am looking on something like Stream.empty... 

Also not that sure what the Pull type really brings on server/client signature. Won't the Stream[F,Socket[F]] will be just enough?

pchlupacek:
@also I just committed what I have. I am little bit struggling to understand the evaluations here. To me the strange part is that this `accept` is completed only once, but is evaluated like for 3 times. I think there must be sort of unwanted strictness somewhere.  I don't want to evaluate any more attempts unless the previous attempt finalise.

pchiusano:
Hey @pchlupacek thanks for continuing to work on this! Sorry for the difficulties you've been having. 

> When closing, The Pull may be actually pulling an channel from accept. What shall be the pull result to terminate this?. The api ends with exception in that case, that has to be converted to normal termination. I am looking on something like Stream.empty...

You can use `Pull.pure` (or `Pull.done`, though that cuts off any further `flatMap` so only do this if you know you are really done). But before you do that -

> Also not that sure what the Pull type really brings on server/client signature. Won't the Stream[F,Socket[F]] will be just enough?

I think you are right. `Pull[F,Nothing,Socket[F]]` and `Stream[F,Socket[F]]` are more or less equivalent, you can do the same sort of stuff with both. I was originally thinking that the `Pull` one would be better since it makes it more clear there is just a single `Socket` being allocated, but seeing how this has worked out, I think it forces a more awkward style. `Pull` is not something people should have to program with much - it's really more for building up transducers, it's a generalization of `Process1`, `Tee`, `Wye`, etc.

So, yeah let's change it to:

```Scala
def client(...): Stream[F,Socket[F]]

// outer stream is the server socket, inner is scope of each accepted connection
def server(...): Stream[F, Stream[F,Socket[F]]] 
```

Both would be defined using `bracket` / `acquire` on `Stream`, and the `Socket` resource would be scoped like you would expect. Typical usage might look something like:

```Scala
for {
  s1 <- tcp.client(..)
  s2 <- tcp.client(..)
  .. do some stuff with `s1`, `s2`
} yield () // `s1` and `s2` pass out of scope
```

I'd like to add / tweak some of the methods on `Socket` accordingly:

* `readOnce` (should be renamed to `readN` to match `await`/`awaitN` nomenclature) should just return an `F[Option[ByteVector]]`
* `available` to be renamed `read`, should return an `F[Option[ByteVector]]` also

Then I think we should introduce some convenience functions on `Socket`. Like maybe:

* `def reads: Stream[F,ByteVector]`: the full stream of reads, defined in terms of `read`
* `def writes(out: Stream[F,ByteVector]): Stream[F,Unit]`: write a stream of chunks

We should experiment a bit with this once everything's working. See my aside below.

> To me the strange part is that this accept is completed only once, but is evaluated like for 3 times.

Not totally sure what you mean here... I just ran the tests and see that `java.nio.channels.AcceptPendingException`. I guess that happens when you try to have more than one accept happening at once? If the NIO API doesn't support that, something you might want to look at it is where your accepts are happening - on the server, the accept actions should be strictly sequential, but if you are using `concurrent.join` and your inner pulls / streams are each calling `accept`, that could lead to multiple outstanding `accept` calls happening at once - the FS2 API has no problem allowing concurrent resource acquisition. The fix is to somehow arrange things so that the connections are accepted sequentially, then passed off to the connection handler (perhaps using a `Ref`?). I guess you could also use a `Semaphore` (the async one I wrote) but that would require upgrading the bound on `F` to `AsyncExt`. (but see below)

I would try playing with this after converting the signatures of the `client` and `server` methods like you proposed and I gave above. Then it will be more clear what is going on I think.

Aside 1:

I am now thinking we should just get rid of use of `ByteVector` here and use `Chunk.Bytes` instead. People can trivially convert that to a `ByteVector` (even without a copy, using `ByteVector.view`), so unless we are extensively using the capabilities of `ByteVector` ourselves internal to FS2, let's not use it.

Aside 2:

I have an idea for how to get rid of `AsyncExt`, in favor of just `Async`!! Idea is to add a 'compare and swap'-like operation to `Async.Ref`, and get rid of the enqueueing version of `modify` / `cancellableModify`. This would give us a lot more flexibility in terms of how we implement this (we could use queues, semaphores, whatever), and would be a lot cleaner anyway -- I never liked having two separate classes. Will create a separate PR / issue for this.

pchlupacek:
@pchiusano cool and thanks. 

I'll little bit dig on multiple accepts and why they are happening. I am quite 100% sure we must be sequential. My  thought was that we `Stream[F,Stream[F,A]]` if defined like `eval(F.async(cb => ???:Stream[F,A])).repeat` will guarantee the sequential ordering (if run via n-join or just flattened), and I think it shall. What I see, in the current implementation it does not, and I think that is actually bug. It would be extremely hard in more complex situation to resolve this to be sequential, and will lead to unexpected behaviour imho. I think the difference of runnning this like flatMap(identitiy) and njoin should be just when the "inner" processes get run and allocate inner resources, but never when the "outer" effects are run. 

I'll gonna do the changes as you propose, but still need to agree on this behaviour what I am seeing, that means the non-sequential allocation of inner streams in.

pchlupacek:
@pchiusano @mpilquist thanks for your patience on this one. I shall have finally some time this weekend to finish this. Sorry for delay.

pchlupacek:
followup in #573

PR #530, opened 2016-01-26
Adds takeRight & tail to Process1

kaiserpelagic:
My approach for takeRight was to use an accumulator of `Vector[I]` and `awaitN` to slide through the Stream. Once the Stream is exhausted the accumulator is written to output. Is there a better way to output a `Vector[I]` other than the foldLeft approach I took?

### DIFF
--- a/core/src/main/scala/fs2/Process1.scala
+++ b/core/src/main/scala/fs2/Process1.scala
@@ -198,7 +198,7 @@ object process1 {
 
   /** Emits the last `n` elements of the input. */
   def takeRight[F[_],I](n: Long): Stream[F,I] => Stream[F,I] =
-    _ pull Pull.takeRight(n)
+    _ pull { h => Pull.takeRight(n)(h).flatMap(is => Pull.output(Chunk.indexedSeq(is))) }
 
   /** Emit the longest prefix of the input for which all elements test true according to `f`. */
   def takeWhile[F[_],I](f: I => Boolean): Stream[F,I] => Stream[F,I] =
--- a/core/src/main/scala/fs2/Process1.scala
+++ b/core/src/main/scala/fs2/Process1.scala
@@ -188,6 +188,10 @@ object process1 {
   def sum[F[_],I](implicit ev: Numeric[I]): Stream[F,I] => Stream[F,I] =
     fold(ev.zero)(ev.plus)
 
+  /** Emits all elements of the input except the first one. */
+  def tail[F[_],I]: Stream[F,I] => Stream[F,I] =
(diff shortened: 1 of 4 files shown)


core/src/main/scala/fs2/pull1.scala
```diff
@@ -181,8 +181,26 @@ private[fs2] trait pull1 {
     if (n <= 0) Pull.pure(h)
     else Pull.awaitLimit(if (n <= Int.MaxValue) n.toInt else Int.MaxValue)(h).flatMap {
       case chunk #: h => Pull.output(chunk) >> take(n - chunk.size.toLong)(h)
+    } 
+
+  /** Emits the last `n` elements of the input. */
+  def takeRight[F[_],I](n: Long): Handle[F,I] => Pull[F,I,Handle[F,I]]  = {
```
pchiusano: This should return a `Pull[F,Nothing,Vector[I]]`. There's no need to return the output `Handle` since it will always be empty!

core/src/main/scala/fs2/Process1.scala
```diff
@@ -188,10 +188,18 @@ object process1 {
...
+  /** Emits all elements of the input except the first one. */
+  def tail[F[_],I]: Stream[F,I] => Stream[F,I] =
+    drop(1)
+
   /** Emit the first `n` elements of the input `Handle` and return the new `Handle`. */
   def take[F[_],I](n: Long): Stream[F,I] => Stream[F,I] =
     _ pull Pull.take(n)
 
+  /** Emits the last `n` elements of the input. */
+  def takeRight[F[_],I](n: Long): Stream[F,I] => Stream[F,I] =
```
pchiusano: with the change to `Pull.takeRight`, this would become `_ pull Pull.takeRight(n).flatMap(is => Pull.output(Chunk.indexedSeq(is)))`

pchiusano:
@kaiserpelagic sorry for the long delay on responding to this. Aside from the minor comment I made, I think this is good to merge!

In general, for transducers that return a strict result like this, always have the `Pull` return it in the result type (the `R` in `Pull[F,O,R]`), not the output type (the `O` in `Pull[F,O,R]`). This way the `Pull` can be used in more places. You can always change the result type of a `Pull`, just via `flatMap`, but you can't change the output type. The output type of a Pull should be reserved for cases where streaming of the result is required.

kaiserpelagic:
@pchiusano Thanks for the feedback. Returning a `Pull[F, Nothing, Vector[I]]` makes a lot of sense.

PR #531, opened 2016-01-27
Update scalaz72 without cleaning up the deprecations

coltfred:
- Update sbt script (build won't start without it)
- Update to scalaz 7.2.0
- Update to scodec 1.0.12 (for cross compile success)
- Update to the latest 2.12 milestone.
- UDPSpec is suspected of breaking the build so commented it out.


coltfred:
Opened against the wrong branch. Closing.

PR #519, opened 2016-01-08
Update Scalaz to 7.2.0

coltfred:
- Update scalaz to 7.2.0 and fix deprecation warnings surrounding Task.run, Task.timed and Task.attemptRun

PR #532, opened 2016-01-27
Update scalaz72 without cleaning up the deprecations

coltfred:
- Update sbt script (build won't start without it)
- Update to scalaz 7.2.0
- Update to scodec 1.0.12 (for cross compile success)
- Update to the latest 2.12 milestone.
- UDPSpec is suspected of breaking the build so commented it out.

This is like #519 but without fixing the deprecation warnings.


pchlupacek:
@coltfred excellent, thanks. @pchiusano are we publishing this to maven?

pchiusano:
Seems fine to me. Just to double check - we dropped support for scalaz 7.0, so 0.8 is against 7.1, is that right?

pchiusano:
Okay, yeah, just confirmed. Looks good, I will merge and publish, probably tomorrow.

zaneli:
:tada:

PR #534, opened 2016-02-06
Implement nonblocking version of `Ref`

pchiusano:
This should be merged after #526. It replaces the use of `ReentrantLock` with [`Ref.ReadWriteSpinLock`](https://github.com/functional-streams-for-scala/fs2/blob/topic/redesign-fastref/core/src/main/scala/fs2/internal/Ref.scala#L73-L103).

@mpilquist WDYT?

There might be an even lower contention version of this, based on `LongAdder` to track the reader count, but I gave up trying to figure that out.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,7 +42,7 @@ lazy val testSettings = Seq(
   parallelExecution in Test := false,
   logBuffered in Test := false,
   testOptions in Test += Tests.Argument("-verbosity", "2"),
-  testOptions in Test += Tests.Argument("-minSuccessfulTests", "100"),
+  testOptions in Test += Tests.Argument("-minSuccessfulTests", "500"),
   publishArtifact in Test := true
 )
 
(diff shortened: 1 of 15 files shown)


pchiusano:
I thought of a lower contention version of this, but I am not sure it is worth bothering with. The general idea is the same as `LongAdder`. Rather than a single `AtomicLong`, you have `Array[AtomicLong]`, one per CPU. When a reader has to increment the reader count, it picks an AtomicLong at random (say by hashing its thread id). Thus there is less contention. The reader count is the sum of all these longs.

If a reader finds a count of -1, that indicates a write is in progress or is in the process of acquiring the write lock, so it fails.

On the writer side, you now have to get through all the AtomicLongs and set them to -1 before starting the write.

I haven't tried to make this work. Might be worth looking into though. I was looking at LongAdder, and it actually dynamically grows the set of AtomicLongs... since it is somewhat expensive to allocate them all up front. It also doesn't use AtomicLong, it uses unboxed values and the primitive CAS operation directly I think.

pchiusano:
@mpilquist good catch... updated

pchiusano:
Okay, merging.

PR #541, opened 2016-02-16
Ported fs2.time object from 0.8

mpilquist:
This PR is a straight PR of the 0.8 time object. The test for `awakeEvery` is commented out because it is failing. Based on some very brief analysis, it appears that discrete mutable signals are broken.


pchiusano:
LGTM... I have a feeling we can generalize to any `Async`, but let's revisit that later. I have some changes to the `Async` type I'd like to try out, to get rid of the need for `AsyncExt`.

PR #542, opened 2016-02-16
Get rid of `AsyncExt` and rewrite `Signal` using `Semaphore`

pchiusano:
This removes `AsyncExt` in favor of a function on `Async` called `access`:

```Scala
  /**
   * Obtain a snapshot of the current value of the `Ref`, and a setter
   * for updating the value. The setter may noop (in which case `false`
   * is returned) if another concurrent call to `access` uses its
   * setter first. Once it has noop'd or been used once, a setter
   * never succeeds again.
   */
  def access[A](r: Ref[A]): F[(A, Either[Throwable,A] => F[Boolean])]
```

This is a really minimal extension, and I like not needing to have both `Async` and `AsyncExt`. Just keep an id as part of the ref state, and increment the id on each successful set. The setter returned from `access` just checks that the id is the same as what it was when the value was read.

This is then used to implement an atomic `modify` and a few other operations. It turns out these are sufficient to implement `Semaphore`, `Queue`, and `Signal`! What this means is that we can now use these concurrency primitives as building blocks for writing other stuff. @pchlupacek I suspect you could eliminate that problem of multiple connections being accepted concurrently by just using a `Semaphore`.

@pchlupacek the code for `Ref` got simpler as well, since we no longer need to support that 'enqueueing modify' operation. Incidentally, you could now implement an enqueuing modify in terms of `Semaphore` - just use the semaphore as a lock to control who gets to call `modify` on some underlying ref.

I did some other housekeeping while I was at it.


mpilquist:
@pchiusano Discrete signals still seem broken to me (REPL output is from redesign-noasyncext branch):

```
scala> import fs2._
import fs2._

scala> import fs2.task._
import fs2.task._

scala> implicit val S = Strategy.fromExecutionContext(scala.concurrent.ExecutionContext.Implicits.global)
S: fs2.Strategy = fs2.Strategy$$anon$2@5c0c767d

scala> val signal = async.signalOf[Task, Int](0).run
signal: fs2.async.mutable.Signal[fs2.util.Task,Int] = fs2.async.mutable.Signal$$anonfun$apply$1$$anon$1@ac6ee28

scala> signal.discrete.map(println).run.run.runFullyAsyncFuture
res0: scala.concurrent.Future[Unit] = scala.concurrent.impl.Promise$DefaultPromise@6cfefe95

scala> 0


scala> signal.set(1).run

scala> signal.set(2).run

scala> signal.set(3).run
```

Also, this implementation uses a single-permit semaphore to guard access to a call to `get`. Under high write loads to the signal, is it possible that `get` returns a different value than the value that caused the semaphore to be incremented? E.g., `set(1)`, `set(2)`, `set(3)` result in 3 calls that increment the semaphore, but if all 3 of the sets occur before the first call to `get`, might the discrete signal end up as `3, 3, 3`?

pchiusano:
Cool I will take a look. And try to craft some tests also.

There is a tryDecrement on every set, so the idea is that the count is
always either 0 or 1. The subscriber can miss an update, but can't get a
duplicate value from sets "piling up". This is the same semantics as the
0.8 signal type, or at least I can't think of any case where the user could
observe a difference.

Of course it looks like it isn't working at all, so take this with a grain
of salt, I'm just saying what I was thinking when writing it. :)

pchiusano:
Okay, I think this should be good to go. Fixed several bugs, one of which was caused by Scala coercing to Unit (thanks @mpilquist for suggesting warn-value-discard flag which I somehow didn't know about!), and added tests of both `Semaphore` and `Signal`. `Semaphore` has an insanely evil scalacheck property which launches anywhere from 2 to N parallel increments and decrements, making sure the count is back to 0 by the end.

Aside, @mpilquist do you know if there's any way to change the compiler flags when launching the repl? The 'warn-unused-imports' is a major downer when using the repl, though I like it for compiling the rest of the code.

Some other random stuff:

* Added `Monad.traverse`. It was already being used internally, I figured why not move it to the class and make it public. Maybe making it public is not such a great idea; why traverse but not `map2`? Where does the line get drawn? My current thinking is that if we are already using a `Monad` or `Async` combinator internally to implement FS2 (and thus it is getting ambient testing), we might as well make it public in case it is useful to people who aren't already rolling with a library like scalaz. But let's hold off on going crazy adding every combinator we can think of - the FS2 utility classes aren't meant to be a replacement for a more heavyweight library like scalaz, cats, whatever. If you have a differing view on this I'd like to hear it, though. I don't feel super strongly one way or the other.
* I also added `Async.parallelTraverse`, based on `Async.start`, the generic analogue of `Task.start`. These work exactly as you'd expect: `parallelTraverse` works just like `traverse`, but runs the effects in parallel. It's being used by the tests of `Semaphore`.

pchiusano:
One more thing - `Signal.discrete` no longer produces an element for the initial value of the `Signal`, only when the `Signal` is `set`. I think this is consistent with the old behavior of `Signal` in 0.8, and is probably what people are expecting.

pchlupacek:
@pchiusano I am not sure that discrete behaviour like you described is correct. Actually 0.8 signal produces the initial value. Only case where it does not, is the deprecated signal constructor that was deprecated exactly for this reason :-)

Also I think continuous and discrete shall behave the same way as regards the initial value of the signal.

core/src/main/scala/fs2/async/mutable/Signal.scala
```diff
@@ -62,60 +47,46 @@ object Signal {
...
-            Stream.eval(getChanged(stamp))
-            .flatMap { case (s,a) => Stream(a).append(go(s)) }
-          }
-          Stream.eval(getStamped).flatMap { case (s,a) => Stream(a).append(go(s))}
+        def continuous: Stream[F, A] =
+          Stream.repeatEval(get)
+
+        def discrete: Stream[F, A] = {
+          val s: F[(ID, Semaphore[F])] =
+            F.bind(Semaphore(0)) { s =>
```
pchiusano: @pchlupacek If we want `discrete` to get the initial value also, we would just change this to `Semaphore(1)`.

pchiusano:
> Actually 0.8 signal produces the initial value. 

@pchlupacek oh, really, ha! I'll change it back then.

DavidGregory084:
@pchiusano my copying other peoples' `build.sbt` senses tell me that this is the correct answer:
```
scalacOptions in (Compile, console) ~= {_.filterNot("-Ywarn-unused-import" == _)}
scalacOptions in (Test, console) <<= (scalacOptions in (Compile, console))
```
From
https://github.com/typelevel/cats/blob/master/build.sbt

mpilquist:
@pchiusano @DavidGregory084 Yep, that's what I typically do. For example, from scodec: https://github.com/scodec/scodec-build/blob/bdfa205c6060f1bece8c751395271baf58fe4efc/src/main/scala/scodec/build/ScodecBuildSettings.scala#L114-L115

pchiusano:
Any objections to merging this?

mpilquist:
:+1: for merge

pchiusano:
Okay, merging. @pchlupacek of course if you have any comments later please let us know. Also, @pchlupacek, I'd try using `Semaphore` for your networking code to make sure that only one connection is accepted at a time. The `Semaphore` would just be held for the very short period from the time the connection is requested until it is accepted.

pchlupacek:
@pchiusano yea no prob. I have followed this. as regards the networking code, I didn't had time to touch this over weekend, but will do next, or later this week. I still need to understand there what going on in this unexpected async behaviour before diving deeper. I perhaps will write some simple tests to make sure the things work as expected with async.

pchiusano:
Cool.

PR #543, opened 2016-02-17
Added mergeSorted to merge and corresponding Spec

alexwei:
Merging two already sorted streams into one output stream of `These`, which indicates whether the values is only present on the left, on the right or in both, matched by sorted key calculated by provided functions.

The algorithm is adapted from the merge sorted stream algorithm from https://softwaremill.com/comparing-akka-stream-scalaz-stream/ by @adamw.

### DIFF
--- a/src/test/scala/scalaz/stream/MergeSortedSpec.scala
+++ b/src/test/scala/scalaz/stream/MergeSortedSpec.scala
@@ -1,0 +1,44 @@
+package scalaz.stream
+
+import org.scalacheck.Arbitrary._
+import org.scalacheck.{Gen, Arbitrary, Properties}
+import org.scalacheck.Prop._
+import scalaz.\&/
+import scalaz.std.anyVal._
+import scalaz.syntax.std.list._
+
+class MergeSortedSpec extends Properties("mergeSorted") {
+
+  case class Foo(i: Int, j: Int, s: String)
+  case class Bar(i: Int, j: Int, s: String)
+
+  implicit val arbitraryFoo = Arbitrary {
+    for {
+      (i, j, s) <- arbitrary[(Int, Int, String)]
+    } yield Foo(i, j, s)
+  }
+  implicit val arbitraryBar = Arbitrary {
+    for {
+      (i, j, s) <- arbitrary[(Int, Int, String)]
+    } yield Bar(i, j, s)
+  }
+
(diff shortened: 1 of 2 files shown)


src/test/scala/scalaz/stream/MergeSortedSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+      (i, j, s) <- arbitrary[(Int, Int, String)]
+    } yield Bar(i, j, s)
+  }
+
+  implicit val arbitraryFooInt: Arbitrary[Foo => Int] = Arbitrary { Gen.oneOf((_: Foo).i, (_: Foo).j) }
+  implicit val arbitraryBarInt: Arbitrary[Bar => Int] = Arbitrary { Gen.oneOf((_: Bar).i, (_: Bar).j) }
+
+  private def distinctOn[A, B](on: A => B): List[A] => List[A] = as => as.groupBy1(on).values.map(_.head).toList
+
+  property("") = forAll { (unsortedFoos: List[Foo], unsortedBars: List[Bar], f: Foo => Int, g: Bar => Int) =>
```
puffnfresh: Is this intentionally missing a name?

puffnfresh:
:+1:

src/test/scala/scalaz/stream/MergeSortedSpec.scala
alexwei: Oops, I'll fix it.

pchiusano: At this point, we are only considering major bug fixes for inclusion in
0.8, and master is currently dead. But if you'd like to port this to the
new design, open up a PR against topic/redesign. Not sure on the naming and
location of this function but we could figure that out later.

Also since we aren't depending on scalaz, you can't use These or Order. :/
I'd just substitute the scala Ordering for Order, but not sure what to do
about These. I guess we could introduce that type to FS2 but that seems
silly just for this one function.

Somewhat related - it would be cool to have a module of functions for doing
various operations on sorted streams, including sorted merge, mergeBy,
inner join, outer join etc.

tonymorris:
¯\_(ツ)_/¯

mpilquist:
Perhaps it would be easier/better to add this to a new fs2-scalaz interop project. Then you can use `Order` and `\&/` without changes. If there's interest in such a project, let me know and I'll get it going.

mpilquist:
@alexwei We could add this method to https://github.com/functional-streams-for-scala/fs2-scalaz if you are still interested.

PR #545, opened 2016-02-22
Uncomment test which works after new Signal implementation.

fiadliel:
This test works after the landing of Semaphore & new Signal code.


mpilquist:
:+1:

PR #546, opened 2016-02-26
Address unhandled RejectedExecutionException when submitting a block code to an ExecutorService

lukiano:


hvesalai:
Just for reference: https://github.com/scalaz/scalaz/issues/1133

pchiusano:
Thanks for finding and fixing, I was not aware exceptions were even possible here.

PR #547, opened 2016-03-03
Add async read/write based on NIO

fiadliel:
@pchiusano This is an early stab at implementing some NIO-based file I/O. Feedback welcome on: style, API, general approach.

Generally:
 - `file.open` provides a `Pull` containing a single `AsynchronousFileChannel`.
 - `file.readAsync`, `file.writeAsync` take a `AsynchronousFileChannel` and read/write from/to it asynchronously, given an `Async` instance.
 - `file.readAllAsync` makes use of `file.open` and `file.readAsync` to read a full file as a `Stream`.

I've never been that convinced that being able to provide the buffer size as a streaming input was that useful, so it's not in this version.

(`Monad.suspend` is also in the tcp rework, and seems essential given the cleanup code in `Stream.bracket` is created before any work is done)

### DIFF
--- a/io/src/main/scala/fs2/io/file/package.scala
+++ b/io/src/main/scala/fs2/io/file/package.scala
@@ -1,0 +1,215 @@
+package fs2.io
+
+import java.nio.ByteBuffer
+import java.nio.channels.{AsynchronousFileChannel, CompletionHandler, FileChannel, FileLock}
+import java.nio.file.{OpenOption, Path, StandardOpenOption}
+
+import fs2.Stream.Handle
+import fs2._
+import fs2.util.Monad
+
+package object file {
+
+  /**
+    * Provides a handler for NIO methods which require a `java.nio.channels.CompletionHandler` instance.
+    */
+  private[fs2] def asyncCompletionHandler[F[_], O](f: CompletionHandler[O, Null] => F[Unit])(implicit F: Async[F], S: Strategy): F[O] = {
+    F.async[O] { cb =>
+      f(new CompletionHandler[O, Null] {
+        override def completed(result: O, attachment: Null): Unit = S(cb(Right(result)))
+        override def failed(exc: Throwable, attachment: Null): Unit = S(cb(Left(exc)))
+      })
+    }
+  }
+
+  /**
+    * Creates a `FileHandle[F]` from a `java.nio.channels.AsynchronousFileChannel`.
+    *
+    * Uses a `java.nio.Channels.CompletionHandler` to handle callbacks from IO operations.
+    */
+  private[fs2] def fromAsynchronousFileChannel[F[_]](chan: AsynchronousFileChannel)(implicit F: Async[F], S: Strategy): FileHandle[F] = {
+    new FileHandle[F] {
+      type Lock = FileLock
+
+      override def close(): F[Unit] =
+        F.suspend(chan.close())
+
+      override def force(metaData: Boolean): F[Unit] =
+        F.suspend(chan.force(metaData))
+
+      override def lock: F[Lock] =
+        asyncCompletionHandler[F, Lock](f => F.pure(chan.lock(null, f)))
+
+      override def lock(position: Long, size: Long, shared: Boolean): F[Lock] =
+        asyncCompletionHandler[F, Lock](f => F.pure(chan.lock(position, size, shared, null, f)))
+
+      override def read(numBytes: Int, offset: Long): F[Option[Chunk[Byte]]] = {
+        val buf = ByteBuffer.allocate(numBytes)
+        F.bind(
+          asyncCompletionHandler[F, Integer](f => F.pure(chan.read(buf, offset, null, f))))(
+          len => F.pure {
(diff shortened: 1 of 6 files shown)


io/src/main/scala/fs2/io/file/package.scala
```diff
@@ -0,0 +1,56 @@
...
+    F.async[O] { cb =>
+      f(new CompletionHandler[O, Null] {
+        override def completed(result: O, attachment: Null): Unit = cb(Right(result))
+
+        override def failed(exc: Throwable, attachment: Null): Unit = cb(Left(exc))
+      })
+    }
+  }
+
+  // uses default NIO thread pool for callbacks... okay? should be configurable?
```
mpilquist: Maybe add an `Option[ExecutorService]` as an argument, defaulted to null?

io/src/main/scala/fs2/io/file/package.scala
```diff
@@ -0,0 +1,56 @@
...
+  def open[F[_]](path: Path, flags: Seq[OpenOption])(implicit F: Monad[F]): Pull[F, Nothing, AsynchronousFileChannel] = {
```
mpilquist: Do you think we will want non-NIO based file primitives? If not, then maybe drop the `Async` suffix from the read/write methods. If so, then maybe add `Async` suffix here?

io/src/main/scala/fs2/io/file/package.scala
```diff
@@ -0,0 +1,56 @@
...
+    F.bind(
+      asyncCompletionHandler[F, Integer](f => F.pure(chan.read(buf, offset, null, f))))(
+      len => F.pure {
+        if (len < 0) None else if (len == 0) Some(Chunk.empty) else Some(Chunk.bytes(buf.array.take(len)))
+      }
+    )
+  }
+
+  def readAllAsync[F[_]](path: Path, chunkSz: Int)(implicit F: Async[F]): Stream[F, Byte] = {
+    def readAllFromChannel(chunkSz: Int, offset: Long)(chan: AsynchronousFileChannel): Pull[F, Byte, Unit] = {
```
mpilquist: `readAllFromChannel` may be useful as a top-level method (if folks acquire an `AsynchronousFileChannel` via a call to `open`.

io/src/main/scala/fs2/io/file/package.scala
fiadliel: Updated with both async and sync implementations.

io/src/main/scala/fs2/io/file/package.scala
fiadliel: Added an implicit Strategy.

io/src/main/scala/fs2/io/file/package.scala
fiadliel: Now includes `def readAllFromFileHandle[F[_]](chunkSize: Int, offset: Long)(h: FileHandle[F]): Pull[F, Byte, Unit]`

fiadliel:
This probably still needs some tests and fleshing out, but it's significantly closer to usable than the last version.

io/src/main/scala/fs2/io/file/package.scala
```diff
@@ -0,0 +1,249 @@
...
+
+import java.nio.ByteBuffer
+import java.nio.channels.{AsynchronousFileChannel, CompletionHandler, FileChannel, FileLock}
+import java.nio.file.{OpenOption, Path, StandardOpenOption}
+
+import fs2.Stream.Handle
+import fs2._
+import fs2.util.Monad
+
+trait FileHandle[F[_]] {
```
mpilquist: Any reason for `FileHandle` to be outside the `fs2.io.file` package?

io/src/main/scala/fs2/io/file/package.scala
```diff
@@ -0,0 +1,249 @@
...
+  type Lock
+
+  // Close the FileHandle
```
mpilquist: Change these comments to full scaladoc comments?

mpilquist:
👍 This is looking great!

io/src/main/scala/fs2/io/file/package.scala
fiadliel: Nope, updated.

io/src/main/scala/fs2/io/file/package.scala
fiadliel: Will do, I'll ping when that's pushed. Other ideas welcome, but I think it's fairly usable now.

fiadliel:
@mpilquist Updated.

I made `readAllFromFileHandle[F[_]](chunkSize: Int)(h: FileHandle[F]): Pull[F, Byte, Unit]` private again; I don't feel like it has enough useful weight to make public; when considering what it could be used for, I could only think of an implementation of `cat`.

I think that most `Pull` implementations that need more complex functionality than reading all of a file, will probably be implemented with `Pull.eval` against `FileHandle` methods.

mpilquist:
👍

PR #549, opened 2016-03-15
Fixed non-termination bug in unNoneTerminate

mpilquist:


pchiusano:
Yikes! I actually don't understand why the original does not work though. Pull.done should terminate the pull unless there is an or on the stack to cancel it out.

mpilquist:
Original doesn't work b/c of the use of `repeatPull`.

pchiusano:
repeatPull is defined in terms of flatMap, and Pull.done flatMap f == Pull.done. So I was thinking it should not matter that f will never terminate.

mpilquist:
Ah, right... Original doesn't work b/c it collects all `Some`s in the chunk rather than stopping on first `None`. E.g., `Chunk(Some(1), Some(2), None, Some(3))` will hit the final `else` and output `Chunk(Some(1), Some(2), Some(3))`.

pchiusano:
Ah, that makes sense. So it's really just the handling of the last chunk
that is incorrect in the original. New version is fine, though I prefer
using repeatPull vs explicit recursion.

mpilquist:
OK, fixed PR

PR #558, opened 2016-03-24
Eliminate class of time/space leaks from internal use of nonstrict append

pchiusano:
This gets rid of any FS2 interpreter internal use of the nonstrict append, which wraps its second argument in a `suspend`. We had a memory leak due to nesting of these suspends. I don't fully understand the bug but this just eliminates this class of errors as a possibility.

I feel like the best solution would be to use types to guarantee that we don't have `suspend { suspend { .. }}` or any nesting of `suspend`, which is redundant and inefficient. I'm not sure if there's a nice way to do this without a major overhaul though. Something to ponder for later, I think this fix is ok for now.


mpilquist:
Awesome! Merging.

PR #560, opened 2016-03-28
change nio.file.chunkReadBuffer to always fill buffer if not EOF

refried:
test and fix for #559

PR #561, opened 2016-03-28
apply #560 to series/0.8

refried:


refried:
restarting build which couldn't download sbt

refried:
restarting build which couldn't download sbt again

PR #562, opened 2016-03-28
apply #560 to series/0.8a

refried:


refried:
oops, wrong merge target

PR #565, opened 2016-03-31
Formalized tracing support in the interpreter

mpilquist:
I can't say I'm enamored with this solution, but I think it would be valuable to support tracing without recompiling FS2. Open to suggestions on how better to implement this. I wanted tracing to have as little overhead as possible -- this implementation results in a invokevirtual call and a branch on each step through the interpreter, which is negligible performance wise.


pchiusano:
LGTM, merging.

PR #568, opened 2016-04-02
Added back attemptEval & made eager bind step depth configuable

sbuzzard:


core/src/main/scala/fs2/StreamDerived.scala
```diff
@@ -45,6 +45,9 @@ trait StreamDerived { self: fs2.Stream.type =>
   def eval_[F[_],A](fa: F[A]): Stream[F,Nothing] =
     flatMap(eval(fa)) { _ => empty }
 
+  def attemptEval[F[_], A](fa: F[A]): Stream[F, Either[Throwable, A]] =
```
pchiusano: This definition is not really ideal since `a.onError(h).flatMap(f) == a.flatMap(f).onError(h andThen (_ flatMap f)`, which leads to the error handler sticking around longer than you expect.

I'd add it to `Streams`, define `eval` in terms of it (so `attemptEval` is the primitive, and `eval` lives in `StreamDerived`). There's already a `StreamCore.attemptEval` primitive that `Stream.attemptEval` can be defined using.

sbuzzard:
Thanks, Paul -- rebased with updated commit based on your comments.

Issue #572, opened 2016-04-10
fs2.concurrent.join is broken with Task.async


pchlupacek:
While working on io.tcp I discovered to me incorrect behaviour of concurrent.join when working with async- based streams. I have created the stream of processes that essentially sleep about a 1s and then prints their result. The code is below:

```scala
val open = new AtomicInteger(0)
 val parent:Stream[Task,Stream[Task,Unit]] = Stream.emits(0 to 20).map { idx =>
    val sleep = (Math.random() * 1000).millis
    eval(Task.delay(println(s"START($idx, ${sleep.toMillis}, ${open.incrementAndGet()})"))) ++
      time.sleep(sleep) ++
      eval(Task.delay(println(s"DONE($idx, ${open.decrementAndGet()})")))
  }
concurrent.join(Int.MaxValue)(parent).run.run.run
```

now this only starts all processes but none of them really terminates. Essentially this runs everything up to result of async evaluation of `time.sleep` (to be clear this starts the time.sleep, but never terminates). 

to verify this I just modified a code to block the thread by delay likewise: 
```scala
val open = new AtomicInteger(0)
 val parent[Task,Stream[Task,Unit]] = Stream.emits(0 to 20).map { idx =>
    val sleep = (Math.random() * 1000).millis
    eval(Task.delay(println(s"START($idx, ${sleep.toMillis}, ${open.incrementAndGet()})"))) ++
     eval(Task.delay(Thread.sleep(sleep.toMillis))) ++
      eval(Task.delay(println(s"DONE($idx, ${open.decrementAndGet()})")))
  }
concurrent.join(Int.MaxValue)(parent).run.run.run
```
And that code works. 

I dig a little bit more, and found (to me) strange behaviour of join : 
  - when you start join it first evaluates all heads of the supplied processes in order as they have been submitted ( I would expect them to be evaluated in order, but interleaved with evaluation of other elements too, longer the initial `parent` process runs)
  - when all `heads` of childrens are evaluated, this will run the second elements, but in _reverse_ order as they have been submitted by `parent` process. (I would expect this to be run and interleaved randomly). 

@pchiusano @mpilquist can you take a quick look on that?

pchlupacek:
[...]

PR #573, opened 2016-04-10
Redesigned implementation and api of fs2.io.tcp.

pchlupacek:


pchlupacek:
This is WIP on the fs2.io.tcp. This is sort of followup to #529. I need currently resolve #572 to move further. 
@pchiusano can you quickly review the API? I think it shall have all build in what was covered in #529

PR #577, opened 2016-04-14
implementation of io.tcp

pchlupacek:
@pchiusano @mpilquist I am having troubles to get this working. the tcp.client seems to be fine. For to me unknown reasons the tcp.server is still problem. 

The nature of this problem (I think) is that multiple accepts the server are in progress . Even I have made this completely serial (acceptIncoming(sch).flatMap { s => emit(s) ++ acceptIncoming(sch)}) the concurrent.njoin seems to still eagerly evaluate the appended evaluations, and I can't figure why.

I know there is workaround to signal open accept and have the next attempt to  queue until the one before really accepts, but this will stockpile the requests possibly causing undesired behaviour. 

I will appreciate if any of you can give shot at server code... 

Thanks.


pchlupacek:
Also, as side observation, the multiple accepts will fire a exception, that is uncaught, and make the whole think to broke w/o failing the process. I just updated with try-catch to demonstrate. I think the issue is in F.async that exceptions are not handled there. correctly.

pchlupacek:
So, I still beleive concurrent.njoin is buggy here. I run the server `manually` with 
```scala
 server[Task](localBindAddress)
    .flatMap { cx =>
      val cxP = cx.flatMap { (socket :Socket[Task]) =>
        socket.writes(
          eval_(Task.delay(println("XXXM START READ"))) ++ socket.reads(1024)
        )
        .onComplete(eval_(socket.close))
      }

      eval(Task.delay(S(cxP.run.run.runAsync(r => println("DONE" + r)))))
    }.run.run.run

```

and works like a charm (except I see reevaluating onCompletes from inner-processes, but that's another story)

So to me the issue is that join is memoizing `F` under certain conditions and re-evaluating it instead of tossing it away. Would say that perhaps the Future.race may be the issue but not sure about that. 

I have also doubts that streams are actually run concurrently, as I do not see any `S` or any sort of forking of child streams in concurrent. join.

mpilquist:
I'll look in detail but a quick note re:
> I have also doubts that streams are actually run concurrently, as I do not see any S or any sort of forking of child streams in concurrent. join.
The `join` method takes an `Async[F]` and the `Async[Task]` instance is backed by a `Strategy`.

pchlupacek:
@mpilquist yeah, you right, but the issue imho lies in fact, that Future.race does not `fork` any of its computations and tasks itself passed to join, are not `forked` essentially if you won't have any asynchronous code that uses `F.async` you won't have concurrent behaviour, but rather left-biased sequential evaluation. 

Also I am very suspicious about this https://github.com/functional-streams-for-scala/fs2/blob/topic/redesign/core/src/main/scala/fs2/Async.scala#L167 as I have feeling with every eval you traverse all residues, not sure if this really tracks already ones that were started to be evaluated or not. Also the sequential traversing may be imho performance issue when you will have 10's thousands opened `F`'s

pchlupacek:
@pchiusano any thoughts on this concurrent.join? It is sort of blocking me going further. I can perhaps fix the concurrent join, but I would like to do it similarly as we had in 0.8 to essentially evaluate the processes and then sink them to fs2.async.queue instead of the sequential evaluation which is implemented now, unless there are some objections to it?

pchiusano:
I am confused since there have been multiple issues mentioned at various
points but I'm not sure which ones are still relevant. 😀 Pavel, can you
reply with a list of issues you are seeing? For each one please link to a
compiling snippet of code, say what the current output is, and say what you
are expecting. Just want to make sure we are all on same page.

I was thinking about implementing join in terms of a queue but couldn't
figure out a way to make it resource safe - basically how do you ensure
that any resources acquired by open streams are appropriately tracked.

pchlupacek:
@pchiusano I cook up what you asking. But essesntially 3 issues now:

1. certain tasks are evaluated twice in join.concurrent. I am still trying to figure out the pattern there, but that is what causeing in io.tcp twice to evaluate accept (or in fact many more than twice)
2. from the code of concurrent.join, it seems to be we do not run stream in parallel, unles the Task inside stream does a fork or async call. If this is the case, I don't think so it is correct. 
3. The fact that we build cancellation as the traversal over residues at every evaluation, causes for every evaluation to traverse (n) of open streams, that will certainly cause overhead when you will mix 100s or thousands of processes. 

As the `concurrent.join` and resources are concerned, I was kind of thinking why we just can't have the child processes evaluate independently? And as such resource safety may be leaved untouched from the normal stream interpretation? We can interrupt right? I am thinking here to evaluate only parent process tracking the open streams, open stream will be then run just as `p.to(enqueue).run.run.runAsync { signalTerminationtoJoin }.```  We just keep then simple signal of the open processes that will join the `parent` process via wye and pipe. I'll try to draft a code to show the idea.

pchiusano:
> As the concurrent.join and resources are concerned, I was kind of thinking why we just can't have the child processes evaluate independently? And as such resource safety may be leaved untouched from the normal stream interpretation? We can interrupt right? I am thinking here to evaluate only parent process tracking the open streams, open stream will be then run just as p.to(enqueue).run.run.runAsync { signalTerminationtoJoin }.`We just keep then simple signal of the open processes that will join theparent` process via wye and pipe. I'll try to draft a code to show the idea.

@pchlupacek Ah, yeah I think that could be a good idea - you can't call `runAsync` (you don't have that for a general `F : Async`) but you can get the same effect with `Async.start`. Do you mind if I take a crack at this? I have a pretty good idea how it would work.

pchlupacek:
@pchiusano Thanks so much and sure. Just saw your PR on this would look at it. I was also observing some weird behaviour of finalisers (invoked many times) with bracket/runAsync/join combination, but don't have anything solid yet. Will use your recent fix, perhaps that was issue I have spotted).

pchlupacek:
@pchiusano @mpilquist I think this is it. It requires the fix introduced in #590 for tests to pass. I was not able to minimize the issue after various attempt to do so, vould be interesting if you will be able to. 

please review, and perhaps we can then merge it.

pchlupacek:
Ah small note too I added handler to uncaught exceptions in schedulers to print to stderr. Hope this is ok.

pchlupacek:
travis seems to complain because of the resource spec not passing, so I belive this is not related to tcp.

io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -0,0 +1,305 @@
...
+  ):Stream[F,Bytes]
+
+
+
+  /**
+    * Read exactly `numBytes` from the peer in a single chunk.
+    * If `timeout` is provided and no data arrives within the specified duration, then this results in
+    * failure with [[java.nio.channels.InterruptedByTimeoutException]]
+    *
+    * When returned size of bytes is < `numBytes` that indicates end-of-stream has been reached.
```
pchiusano: Hmm, I feel like `readN` should always return exactly `numBytes`, or `None`, otherwise you should just use `read`. Is this comment just out of date?

pchlupacek: Its perhaps my poor english, but I understand the comment that way. Read always `N` bytes (wait for them to be available). Only receive < N on eof. Fail on timeout if provided.  

`read` in contrary takes what is availbale (up to the maximum) possibly None if nothing is available at all. I was in fact thinking to get rid of Option and just user empty chunk of bytes, but I think we sort of agreed on the Api. However if you have idea to changing is no brainer....

pchlupacek:
@pchiusano not sure about if the comment I wrote is ok with you and if this can be merged?

core/src/main/scala/fs2/Async.scala
```diff
@@ -7,6 +7,9 @@ import fs2.util.{Free,Functor,Catchable}
 trait Async[F[_]] extends Catchable[F] { self =>
   type Ref[A]
 
+  /** create suspended computation evaluated lazily **/
+  def suspend[A](a: => A):F[A]
```
mpilquist: This is no longer needed here b/c we pick it up from `Monad` now.

io/src/test/scala/fs2/io/TestUtil.scala
```diff
@@ -0,0 +1,51 @@
...
+    )
+  }
+
+
+
+  def localAddress(port: Int) = new InetSocketAddress("localhost", port)
+  val localBindAddress = localAddress(9999)
+
+
+  // the io specs may fight for local machine resources
```
mpilquist: I don't think the spec lock is strictly necessary now given that we have ScalaTest configured to run the tests 1 at a time and that we only use 1 worker thread for ScalaCheck. I'm fine with leaving this though.

io/src/test/scala/fs2/io/tcp/SocketSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+
+        val result =
+          concurrent.join(2)(Stream(
+            echoServer.drain
+            , clients
+          ))
+          .take(clientCount).runLog.run.unsafeRun
+
+
+        (result.size == clientCount) &&
```
mpilquist: I don't think this prop is being executed by ScalaTest. For example:
```
scala> (new MySpec).execute()
MySpec:
Starting test
Finished test
- test

scala> class MySpec extends Fs2Spec { "test" in { assert(1 == 2) } }
defined class MySpec

scala> (new MySpec).execute()
MySpec:
Starting test
Finished test
- test *** FAILED ***
  org.scalatest.exceptions.TestFailedException was thrown. (<console>:18)
```

Can you wrap this expression with an `assert`?

mpilquist:
@pchlupacek I made two minor comments. If you get a chance, can you update this PR? If you don't beat me to it, I'll merge this later today when I have some time and then address those comments after the merge.

pchlupacek:
@mpilquist sure, will do.

io/src/test/scala/fs2/io/TestUtil.scala
pchlupacek: Ok I can remove it. I just want to be sure we really won't get any of the two tests to run concurrently hence we share the bind port.

pchlupacek:
@mpilquist just pushed the update...

PR #581, opened 2016-04-21
Renamed lots of types and methods to improve usability

mpilquist:
This PR includes the following changes:
 - Deleted `Process1`
 - Renamed `Channel` to `Pipe`
 - Replaced `Tee` and `Wye` with `Pipe2`
 - Replaced `pipe` with `through` (and various `pipe` variants like `pipev` with `through` variants).
 - Added implicit conversions from `Pipe[Pure,I,O]`/`Pipe2[Pure,I,I2,O]` to `forall F . Pipe[F,I,O]` / `forall F . Pipe2[F,I,I2,O]` to the fs2 package object so that pure pipes can be used with `through` without a manual `covary`. Note that this is rare, as all of the built-in pipes are defined as `forall F . Pipe[F,I,O]`
 - Added `throughp` for abnormal cases where there's a `Pipe[Pure,I,O]` instead of a `forall F . Pipe[F,I,O]`.
 - Added `Stream.pure(...)` to create a `Stream[Pure,A]`
 - Added `s.pure` as an alias for `s.covary[Pure]`. Both this and the previous addition allow use of `through` with pure streams -- i.e., they workaround limitations in Scala's inference of `Nothing`.  This comes up when working with streams with `Nothing` as the effect type. E.g. `Stream(1, 2, 3).through(pipe.take(5))` won't compile because Scala refuses to infer `Nothing` for the `F` argument to `take`.
 - Renamed all of the various run methods on `Task` such that they have an `unsafe` prefix. In my experience teaching FS2, the `.run.run.run` pattern is confusing for folks that are new to working with delimited effects. Why are the first 2 runs safe and the third run should be called only at the end of the universe? This change is similar to the recent change to scalaz in https://github.com/scalaz/scalaz/pull/1040
 - Added `async` as a method on `Task` which is an alias for `Task.start(this).flatMap(identity)`
 - Removed the `runFullyAsync*` methods from `Task` as the same can now be accomplished using `async` -- e.g., `t.async.unsafeRunAsyncFuture`.

Overall, I'm pretty happy with these changes. I'm disappointed that we can't get inference working as well with `through` as it worked for `pipe`, but (1) working with pure streams in production code is generally rare (2) `throughp` is not a terrible workaround and (3) folks can always call `s.pure` or `s.covary[Pure]` if they have a pure stream and they want to use `through`.

If there are any strong objections to these changes, let me know. If not, I'll merge at some point this week.

/cc @pchiusano @pchlupacek @runarorama


core/src/main/scala/fs2/util/Task.scala
```diff
@@ -96,18 +96,21 @@ class Task[+A](val get: Future[Either[Throwable,A]]) {
...
    */
-  def run: A = get.run match {
+  def unsafeRun: A = get.run match {
     case Left(e) => throw e
     case Right(a) => a
   }
 
   /** Like `run`, but returns exceptions as values. */
-  def attemptRun: Either[Throwable,A] =
+  def unsafeAttemptRun: Either[Throwable,A] =
```
rintcius: Just curious.. why is this considered unsafe (for me unsafe meant "the type doesn't lie")?

core/src/main/scala/fs2/util/Task.scala
```diff
@@ -96,18 +96,21 @@ class Task[+A](val get: Future[Either[Throwable,A]]) {
...
     })
 
+  /** Returns a task that performs all evaluation asynchronously. */
+  def async(implicit S: Strategy): Task[A] = Task.start(this).flatMap(identity)
+
   /**
    * Run this `Task` and block until its result is available. This will
    * throw any exceptions generated by the `Task`. To return exceptions
-   * in an `\/`, use `attemptRun`.
+   * in an `Either`, use `attemptRun`.
```
rintcius: `attemptRun` -> `unsafeAttemptRun`

core/src/main/scala/fs2/util/Task.scala
mpilquist: The general idea is that the various run methods are the only methods that *do* anything in the library. Building streams, transforming a stream to its `Free` representation, and transforming the `Free` representation in to an effectful value. All of these are referentially transparent. When using `Task`, the moment you run the task, you lose referential transparency. The `unsafe` prefix is meant as a speed bump -- something that notifies user that they are going to give up referential transparency here.

core/src/main/scala/fs2/util/Task.scala
mpilquist: Thanks, pushed a fix.

pchiusano:
Lgtm!

PR #582, opened 2016-04-21
Fix for probable bug that could cause finalizers not to be run if allocated during an async step

pchiusano:
Note: based off of #581, should be merged after.

This fixes a probable bug that could cause finalizers not to be run. It may be responsible for the Travis failures, though I haven't been able to reproduce those locally.

When an async step runs, all allocated resources are recorded in a mutable `Resources` map (call these 'step-local' resources). A single root token in the parent `Resources` tracks all the resources allocated by the step. When the step completes, there is a critical section in which the root token is removed from the parent `Resources` and any step-local resources are copied to the parent (basically, after the step completes, the parent resources map should look the same as if the resources were acquired synchronously). If during this critical section the parent resources map is closed, the old code would just drop those step-local resources on the floor. This code fixes that.


pchiusano:
Travis passed, so perhaps we can tentatively declare victory! I don't really have proof that this bug was causing the failures but it seems probable.

pchlupacek:
@pchiusano excellent and thank you for this!

PR #583, opened 2016-04-21
Queue-based implementation of concurrent.join

pchiusano:
Note: based on #582 and #581 
Note: don't merge yet, it doesn't work

This uses @pchlupacek's suggestion [here](https://github.com/functional-streams-for-scala/fs2/pull/577#issuecomment-211028139). It's not quite working yet, though. The join(3) test currently hangs for some reason. Good news is that once this is working, it will fix that linear performance problem spotted by @pchlupacek, where we have to scan all open streams to do a cancellation of the losing steps.

Basic idea here is pretty simple - just have the inner streams dump to a synchronous queue. Each is also controlled by an interrupt signal which is set to true when the outer stream terminates, which should ensure that their finalizers get run. But there are various edge cases to think through, and I apparently goofed something up.

More eyeballs on this appreciated, I won't have a chance to take a closer look until next week I think.

@pchlupacek or @mpilquist there might be a totally different, simpler implementation, so feel free to ignore this one or just use it for inspiration. Using `Async.start` is a good way to spawn running of the inner streams, though, I think any queue-based solution will have that somewhere.

### DIFF
--- a/core/src/main/scala/fs2/async/mutable/Queue.scala
+++ b/core/src/main/scala/fs2/async/mutable/Queue.scala
@@ -141,4 +141,26 @@ object Queue {
         def available: immutable.Signal[F, Int] = Signal.constant(0)
       }
     }}
+
+  /** Like `Queue.synchronous`, except that an enqueue or offer of `None` will never block. */
+  def synchronousNoneTerminated[F[_],A](implicit F: Async[F]): F[Queue[F,Option[A]]] =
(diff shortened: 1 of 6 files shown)


pchiusano:
I believe everything's working now, or at least, I can't get it to fail locally anymore. :) If Travis says this is good to go, I'd suggest merging.

pchlupacek:
thank you will take it in tcp stuff over weekend

PR #585, opened 2016-04-25
Add fs2.text.utf8Encode and fs2.text.utf8Decode

fiadliel:
The tests and a lot of the logic are taken from scalaz-stream <= 0.8.

(kind of miss `ByteBuffer` here)

Using `Chunk[Chunk[Byte]]` to make sure that we can round-trip empty strings.

### DIFF
--- a/core/src/main/scala/fs2/text.scala
+++ b/core/src/main/scala/fs2/text.scala
@@ -1,0 +1,69 @@
+package fs2
+
+import java.nio.charset.Charset
+
+import fs2.Stream.Handle
+
+object text {
+  private val utf8Charset = Charset.forName("UTF-8")
+
+  /** Converts UTF-8 encoded `Chunk[Byte]` inputs to `String`. */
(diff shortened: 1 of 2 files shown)


mpilquist:
👍 I know @pchiusano wanted to move `compress` to the `io` module. That makes me wonder whether we want text in `io` too. We can handle that in subsequent PR though, if necessary.

fiadliel:
If `compress` is moving, then yes, it makes sense to move all byte-level code to there (when is that ever not about network or file IO).

pchiusano:
Did I say that? :) I thought I just talked about moving compress and text
and other stuff to an extras subproject. Compress is observationally pure
so I don't see any need to put it in io.

mpilquist:
Ha, okay my misunderstanding. So you are thinking an fs2-extras subproject in this repository?

pchiusano:
That's what I was thinking yeah... but don't feel very strongly about it.

PR #586, opened 2016-04-28
Make Stream.range more robust

ChrisCoffey:
While porting the scalaz-streams docs to fs2 I found some unexpected behavior with step-sizes & `Stream.range`. Previously only a positive `by` was supported. 

I elected to return the empty list in failure situations, but I'm curious if you think there is a better solution.


mpilquist:
👍 Thanks

PR #587, opened 2016-04-29
Port over scalaz-streams creation guide.

ChrisCoffey:
Introduces the basics of Stream creation. Please give this a read and make sure I'm telling the truth everywhere 😄

### DIFF
--- a/docs/CreatingStreams.md
+++ b/docs/CreatingStreams.md
@@ -9,18 +9,18 @@ import fs2.util.Task
 // import fs2.util.Task
 
 val ones = Stream.iterate(1)(identity)
-// ones: fs2.Stream[Nothing,Int] = fs2.Stream$$anon$1@693056b8
+// ones: fs2.Stream[Nothing,Int] = fs2.Stream$$anon$1@7a802fd
 ```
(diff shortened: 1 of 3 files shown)


docs/CreatingStreams.md
```diff
@@ -0,0 +1,75 @@
...
+## Evaluating Tasks
+In addition to creating pure streams using some generative function, we can also create streams by evaluating an effect, `F[A]`. The resulting stream will emit the `A` or fail attempting to do so.
+```scala
+val greeting = Stream.eval(Task.now("Hi there!"))
+// greeting: fs2.Stream[fs2.util.Task,String] = fs2.Stream$$anon$1@26e42949
+
+val hi = greeting.runLog.run.unsafeRun
+// hi: Vector[String] = Vector(Hi there!)
+```
+This producees a `Stream[Task, String]`, which we can then force evaluation of using the `runLog.run.run`, which has the type of `Free[Task,Vector[String]] -> Task[Vector[String]] -> Vector[String]`. As you can see, the various `run` calls have the effect of stepping up one layer from `Stream` to `Task`  to `A`.
```
gseitz: Should this be `runLog.run.unsafeRun`?

ChrisCoffey: Ah yes, good catch!

mpilquist: Minor nitpick -- instead of saying `runLog.run.unsafeRun` "has the type of X", maybe reword so a reader isn't under the impression that X is a real type?

PR #588, opened 2016-04-30
Ported ScalaCheck tests to ScalaTest

mpilquist:
While trying to determine why tests randomly hang on Travis, I decided to port the tests from pure ScalaCheck to ScalaTest+Scalacheck. This has a number of benefits including:
 - Better error messages
 - More predictable test ordering
 - Ability to add logging around tests

Based on this logging, I suspect the test that randomly hangs is `asynchronous resource allocation (3)` in `ResourceSafetySpec`.


pchiusano:
Damn. That is the concurrent.join test.

No objections to switching to scala test.

Issue #589, opened 2016-04-30
Finalizers are not called in expected order in concurrent.join


pchlupacek:
Finalizers in concurrent.join are called in reverse order. The following code 
```scala

  concurrent.join(10)(
    Stream.range(0,3).map { idx =>
      emit(idx)
      .onFinalize(Task.delay(println(s"INNER DONE $idx (A)")))
      .onFinalize(Task.delay(println(s"INNER DONE $idx (B)")))
    }
    .onFinalize(Task.delay(println(s"OUTER DONE (A)")))
    .onFinalize(Task.delay(println(s"OUTER DONE (B)")))
  )
  .onFinalize(Task.delay(println(s"MERGE DONE (A)")))
  .onFinalize(Task.delay(println(s"MERGE DONE (B)")))
  .runLog.run.unsafeRun

```

outputs 

```
OUTER DONE (A)
OUTER DONE (B)
INNER DONE 2 (B)
INNER DONE 0 (B)
INNER DONE 2 (A)
INNER DONE 0 (A)
INNER DONE 1 (B)
INNER DONE 1 (A)
MERGE DONE (A)
MERGE DONE (B)

```

while serialised version (flatMap instead of join) outputs 

```
INNER DONE 0 (A)
INNER DONE 0 (B)
INNER DONE 1 (A)
INNER DONE 1 (B)
INNER DONE 2 (A)
INNER DONE 2 (B)
OUTER DONE (A)
OUTER DONE (B)
MERGE DONE (A)
MERGE DONE (B)
```

Note that there are two problems, actually:
(a) inner stream finalizers are called in reverse order 
(b) OUTER finalizer is called before INNER processes are completed.

pchiusano:
On a) the inner stream finalizers are run in nondeterministic order. The inner streams are running concurrently so there won't be any guarantee on the order their finalizers are run in. That seems unavoidable, right? b) might be fixable somehow.

EDIT: I can think of a fix for b), it's basically using the logic that waits for the inner streams to terminate ([here](https://github.com/functional-streams-for-scala/fs2/blob/topic/redesign/core/src/main/scala/fs2/concurrent.scala#L41)). Except that should be invoked at the end of the stream (using `noneTerminated`), and only invoked once.

pchiusano:
[...]

PR #590, opened 2016-04-30
Added docs for concurrent.join. Fixed the finalizer handler to be not…

pchlupacek:
… called in every situation

### DIFF
--- a/core/src/main/scala/fs2/async/mutable/Semaphore.scala
+++ b/core/src/main/scala/fs2/async/mutable/Semaphore.scala
@@ -56,7 +56,8 @@ object Semaphore {
 
   /** Create a new `Semaphore`, initialized with `n` available permits. */
   def apply[F[_]](n: Long)(implicit F: Async[F]): F[Semaphore[F]] = {
-    def ensureNonneg(n: Long) = if (n < 0) throw new IllegalArgumentException("n must be nonnegative, was: " + n)
+    def ensureNonneg(n: Long) = assert(n >= 0, s"n must be nonnegative, was: $n ")
+
     ensureNonneg(n)
     // semaphore is either empty, and there are number of outstanding acquires (Left)
     // or it is nonempty, and there are n permits available (Right)
--- a/core/src/main/scala/fs2/concurrent.scala
+++ b/core/src/main/scala/fs2/concurrent.scala
@@ -1,17 +1,38 @@
 package fs2
 
-import util.Free
 
 object concurrent {
 
-  def join[F[_],O](maxOpen: Int)(s: Stream[F,Stream[F,O]])(implicit F: Async[F]): Stream[F,O] = {
-    if (maxOpen <= 0) throw new IllegalArgumentException("maxOpen must be > 0, was: " + maxOpen)
+  /**
+    * Joins non deterministically streams.
+    *
+    *
+    * @param maxOpen    Allows to specify maximum open streams at any given time.
+    *                   The outer stream will stop its evaluation when this limit is reached and will
+    *                   start evaluating once currently open streams will be <= mayOpen.
+    *                   MaxOpen must be > 0
+    *
+    * @param outer      Outer stream, that produces streams (inner) to be run concurrently.
+    *                   When this stops gracefully, then all inner streams are continuing to run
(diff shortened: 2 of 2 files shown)


pchlupacek:
@pchiusano as mentioned in #589

core/src/main/scala/fs2/concurrent.scala
```diff
@@ -23,8 +44,10 @@ object concurrent {
               }
               else F.pure(())
             }}}
-          s.chunks.attempt.evalMap { o => outputQueue.enqueue1(Some(o)) }
-           .interruptWhen(killSignal).run.flatMap { _ => Free.eval(done) }.run
+          inner.chunks.attempt.evalMap { o => outputQueue.enqueue1(Some(o)) }
+          .interruptWhen(killSignal)
+          .onFinalize(done)
+          .run.run
         }}
```
pchiusano: I don't actually see how this change could make a difference... Are you sure it does? Because of the attempt there does not seem to be any way that the run can fail.

core/src/main/scala/fs2/async/mutable/Semaphore.scala
```diff
@@ -56,7 +56,8 @@ object Semaphore {
 
   /** Create a new `Semaphore`, initialized with `n` available permits. */
   def apply[F[_]](n: Long)(implicit F: Async[F]): F[Semaphore[F]] = {
-    def ensureNonneg(n: Long) = if (n < 0) throw new IllegalArgumentException("n must be nonnegative, was: " + n)
+    def ensureNonneg(n: Long) = assert(n >= 0, s"n must be nonnegative, was: $n ")
```
pchiusano: Is assert lazy in its second argument? That's why I avoided it, to avoid instantiating that string if it isn't needed.

core/src/main/scala/fs2/concurrent.scala
```diff
@@ -1,17 +1,38 @@
...
+  // todo: document when the finalizers are calle in which situation
+  def join[F[_],O](maxOpen: Int)(outer: Stream[F,Stream[F,O]])(implicit F: Async[F]): Stream[F,O] = {
+    assert(maxOpen > 0,"maxOpen must be > 0, was: " + maxOpen)
     for {
       killSignal <- Stream.eval(async.signalOf(false))
       openLeases <- Stream.eval(async.mutable.Semaphore(maxOpen))
       outerDone <- Stream.eval(F.refOf[Boolean](false))
       outputQueue <- Stream.eval(async.mutable.Queue.synchronousNoneTerminated[F,Either[Throwable,Chunk[O]]])
-      o <- s.evalMap { (s: Stream[F,O]) =>
+      o <- outer.evalMap { (inner: Stream[F,O]) =>
```
pchiusano: I always shadow names. IMO it is good practice unless you know you want to be able to refer to both variables, which in this case we don't.

pchlupacek: sorry to touch you code this way:-) I just wanted to be sure I understand what is called where, and didn't revert it. You want me to?

core/src/main/scala/fs2/async/mutable/Semaphore.scala
pchlupacek: ok, I see your point. I'll introduce something like this in util?. I just think the if and throw is sort of unexpected. Maybe checkArgument(cond:Boolean, msg:=> String):Unit?

core/src/main/scala/fs2/concurrent.scala
pchlupacek: yes 100% positive it does.  I was  starring on it same way, and took me half day to pint point it. It may uncover issue somewhere else. But it really does. I will be putting out the tcp PR laters today, so you may try to run it w/o this change.

core/src/main/scala/fs2/concurrent.scala
pchiusano: Nah it's fine. More of a style thing. It just took me a minute to realize that wasn't relevant to the rest of the diff.

core/src/main/scala/fs2/concurrent.scala
pchiusano: Hmm ok, interesting.

PR #592, opened 2016-04-30
Topic/redesign udp

pchlupacek:
@pchiusano @mpilquist This is a draft of udp api. This is sort of udp `server` socket implemenation, this won't cover situation when you have many bind sockets such as for RTP protocols. 

However I have issue in implementing this. Hence DatagramSocket.receive blocks, I need to run it in other thread. Hence Async does not have any unsafeXX methods, I cannot do that with Async. 

Another idea I had was to run it as Stream sort of attached to resulting stream. But even there I will need different Strategy / Executor to fork the receive on. 

So there are imho two options:

1. Introduce runXX to Async
2. Provide explicitly two Async Instances to listen, one for executing the Sream, and one for running the reads. 

Any ideas/opinions on that design?

(Please disregard all tcp stuff, this will disappear once we will merge in tcp)


core/src/main/scala/fs2/Async.scala
```diff
@@ -7,6 +7,9 @@ import fs2.util.{Free,Functor,Catchable}
 trait Async[F[_]] extends Catchable[F] { self =>
   type Ref[A]
 
+  /** create suspended computation evaluated lazily **/
+  def suspend[A](a: => A):F[A]
```
mpilquist: Don't need this anymore -- it was added to `Monad` in a different PR.

pchiusano:
Would it work to take an implicit Strategy any place you have a blocking IO call?

pchlupacek:
@pchiusano yeah was kind o thinking about it. I wasn't sure if the idea isn't to completely `hide` the strategy from the API, hence we have that Async now. But likely you are right this is the simplest and cleanest way. Shall I do it so?

pchlupacek:
@pchiusano after more thinking about this I am not sure how to do this with passed in Strategy. I need to run enqueOne in F, and that's likely need to run `F`. Hence Async[F] doesn't have that,  cannot do this. And passing another Async[F] instead `S` sounds a bit confusing to me too...

pchlupacek:
@pchiusano I tried few options here and there. I came to conclusion that we may need to abstract blocking run over `F`. Also when you will write a library, where you don't want to specialize over `F` and you need to perform callbacks, then you need that to. So I do propose this:
```scala
// inside Async object
/**
    * Used to strictly evaluate `F`.
    */
  trait Run[F[_]] {

   /**
      * Run this `F` and block until it completes.
      * If the evaluation of the `F` terminates with an exception,
      * then this will return that exception as Optional value. 
      * Otherwise this terminates with None.
      *
      * Note that hence this blocks, special care must be taken on thread usage.
      * Typically, this has to be run off the thread that allows blocking and
      * is outside the main Executor Service or Scheduler.
      *
      * If you run it inside the scheduler or `Strategy` that is used to run
      * rest of your program you may encounter deadlocks due to thread resources
      * being used.
      *
      * It is not recommended to use this inside user code.
      *
      */
    def runEffects(f:F[Unit]): Option[Throwable]

  }

```
any thoughts?

mpilquist:
@pchlupacek Re: UDP -- can we use datagram channels instead, using an async channel group like we do for TCP?

pchlupacek:
@mpilquist ah sure. Cannot understand how I missed that one. Will update the Pr with this. Nevertheless I still believe that this `Run` or form of it is usefull when you developing library with legacy code that is not asynchronous and is of blocking nature.

pchlupacek:
@mpilquist you had an specific idea how to wrap DatagramChannel in F.async? I was hoping it has similar interface like its tcp counterpart, bet seems it does not have Async interface. Any thoughts on this?

mpilquist:
@pchlupacek I was thinking we could dedicate a thread to process all udp receives -- that thread would be an event loop doing selects on all of the open datagram channels. Basically, we'd have to build our own async layer on top of `DatagramChannel`. Note that Oracle added and then removed `AsynchronousDatagramChannel` before releasing Java 7.

pchlupacek:
@mpilquist ok, got it. So what are you suggesting is that we implement the async channel by our own right?

mpilquist:
Yeah, though maybe that's out of scope? Perhaps it's okay for the UDP implementation to pin a thread and then leave high performance UDP with lots of open ports to end users (and perhaps built on netty?)

pchlupacek:
I think we can do better than netty. We can have backpressure for udp etc. I was just curious about if I could implement that w/o that runXXX interface, and perhaps I could. Its just a little bit more complicated than with tcp.

pchlupacek:
I think we may have that single udp-bind api and later build the many-udp binds api (selector based). You have now any use for which ever of them?

mpilquist:
I use the 0.8 udp receive functionality with multiple ports, but I just create a stream for each port and mergeN the results. I typically use 1 to 10 ports to the overheard of the merge and pinned threads aren't prohibitive.

pchlupacek:
@mpilquist got you. I gonna update the api and make it multiport. The single-port bind will be then only special case of it. Will review 0.8 api, (didn't notice there is multiport in udp already), so make sure we won't miss anything important there.

mpilquist:
@pchlupacek Just to be clear, in 0.8 I did something like this:
```
val ports: NonEmptyList[Int] = ...
val listeners = ports.list.map { p => udp.listen(p)(udp.receives(bufferSize)) }
concurrent.join(Stream.emits(listeners))
```

pchlupacek:
@mpilquist ah ok. Anyhow I'll  draft it. Think it is useful. Then we can define single lister in terms of it.

PR #593, opened 2016-05-04
Fix for #589 finalization order bug

pchiusano:
Previously, when reaching the end of a stream, any remaining uncalled finalizers were run in LIFO order. Now they are run in FIFO order. I believe this is more correct, especially since when nesting calls to `bracket`, it's actually the innermost finalizer that will be added to the resources map first and we generally want to call finalizers 'innermost first'. This has the effect that `a.onFinalize(f1).onFinalize(f2)` will run `f1` then `f2`.

Review by @pchlupacek

### DIFF
--- a/core/src/main/scala/fs2/StreamCore.scala
+++ b/core/src/main/scala/fs2/StreamCore.scala
@@ -471,9 +471,9 @@ object StreamCore {
 
   private[fs2]
   def runCleanup[F[_]](cleanups: Iterable[Free[F,Either[Throwable,Unit]]]): Free[F,Either[Throwable,Unit]] = {
-    // note - run cleanup actions in LIFO order, later actions run first
+    // note - run cleanup actions in FIFO order, but avoid left-nesting binds
     // all actions are run but only first error is reported
-    cleanups.foldLeft[Free[F,Either[Throwable,Unit]]](Free.pure(Right(())))(
+    cleanups.toList.reverse.foldLeft[Free[F,Either[Throwable,Unit]]](Free.pure(Right(())))(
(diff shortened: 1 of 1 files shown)


pchiusano:
Whoa, travis failure (on [ResourceSafetySpec 2b](https://travis-ci.org/functional-streams-for-scala/fs2/jobs/127878214#L1157)) is alarming. I don't see how this could be directly responsible for that failure, since it just affects the order of finalizers, not which ones are called, which is all that test covers. My guess is there's still some race condition and having the finalizers run in a different order generates different timing.

pchiusano:
Okay, I can confirm that the Travis failure is unrelated, because I just got 2b to to fail with this change reverted, just by ratcheting up the number of test runs. I'm tracking that in #594.

So I believe this is good to merge.

core/src/main/scala/fs2/StreamCore.scala
```diff
@@ -471,9 +471,9 @@ object StreamCore {
 
   private[fs2]
   def runCleanup[F[_]](cleanups: Iterable[Free[F,Either[Throwable,Unit]]]): Free[F,Either[Throwable,Unit]] = {
-    // note - run cleanup actions in LIFO order, later actions run first
+    // note - run cleanup actions in FIFO order, but avoid left-nesting binds
     // all actions are run but only first error is reported
```
pchlupacek: ok for now, but perhaps we can build and exception that will collect all throwables laters?

pchiusano: No objections to that. I didn't change that behavior in this PR, just the order they are run.

pchlupacek:
ok lgtm, and I perphas drop a separate pr on this.

PR #595, opened 2016-05-07
Partially fixed #591

### LINKED ISSUE #591: ResourceSafetySpec / resource allocation (3) fails
The "asynchronous resource allocation (3)" test intermittently either fails or hangs. Opening an issue to track the failures.

```
[info] - asynchronous resource allocation (3) *** FAILED *** (1 second, 455 milliseconds)
[info]   TestFailedException was thrown during property evaluation.
[info]     Message: -1 was not equal to 0
[info]     Location: (ResourceSafetySpec.scala:119)
[info]     Occurred when passed generated values (
[info]       arg0 = PureStream(unchunked,fs2.Stream$$anon$1@71b5aace),
[info]       arg1 = true,
[info]       arg2 = Failure(failure-in-pure-code,fs2.Stream$$anon$1@34878bac),
[info]       arg3 = SmallPositive(1)
[info]     )
```

mpilquist:
There were 2 issues in the definition of `join`:
 - `clear` on a `Semaphore` fails if there are waiters. This code handles that case by setting the number to decrement in such a case to `maxOpen`.
 - When finalizing, there's a chance that another inner stream is ahead of the finalizer's `decrementBy` call in the semaphore's waiters list. Sometimes when this happens, for reasons I don't fully understand, the inner stream is started asynchronously but then never run, resulting in the semaphore never being incremented. This PR guards against this case.

Review by @pchiusano

### DIFF
--- a/core/src/main/scala/fs2/concurrent.scala
+++ b/core/src/main/scala/fs2/concurrent.scala
@@ -27,41 +27,52 @@ object concurrent {
   // todo: document when the finalizers are calle in which situation
   def join[F[_],O](maxOpen: Int)(outer: Stream[F,Stream[F,O]])(implicit F: Async[F]): Stream[F,O] = {
     assert(maxOpen > 0,"maxOpen must be > 0, was: " + maxOpen)
-    for {
-      killSignal <- Stream.eval(async.signalOf(false))
-      openLeases <- Stream.eval(async.mutable.Semaphore(maxOpen))
-      outerDone <- Stream.eval(F.refOf[Boolean](false))
-      outputQueue <- Stream.eval(async.mutable.Queue.synchronousNoneTerminated[F,Either[Throwable,Chunk[O]]])
-      o <- outer.evalMap { (inner: Stream[F,O]) =>
-        F.bind(openLeases.decrement) { _ =>
-        F.start {
-          val done =
-            F.bind(openLeases.increment) { _ =>
-            F.bind(F.get(outerDone)) { outerDone =>
-            F.bind(openLeases.count) { n =>
-              if (n == maxOpen && outerDone) {
-                outputQueue.enqueue1(None)
+
+    case class State(outerDone: Boolean, open: Int, killed: Boolean)
+
+    def throttle[A](stateSignal: async.mutable.Signal[F,State]): Pipe[F,Stream[F,A],Stream[F,A]] = {
+      def go(open: Int): (Stream.Handle[F,State], Stream.Handle[F,Stream[F,A]]) => Pull[F,Stream[F,A],Unit] = (state, s) => {
+          if (open < maxOpen) {
+            s.receive1 { case inner #: s =>
+              val monitoredStream = {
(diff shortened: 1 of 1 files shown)


mpilquist:
I think what happens is that the inner stream is immediately terminated because of the `interruptWhen(killSignal)`, resulting in the bracket inside of `onFinalize` never getting executed. Btw, I tried swapping the order of `onFinalize(done)` and `interruptWhen(killSignal)` but that introduced other errors.

mpilquist:
Test hang still occurs, though rarely, if there's a waiting inner stream at the time of overall stream finalization, and the inner stream gets past the guard check before the kill signal is set. At this point, I think what's happening is that the task started via `F.start` never gets run on a trampoline b/c the overall stream is terminating.

pchiusano:
Wow nice!

I wonder if you could use an additional semaphore, just a simple lock (count 0 or 1), to prevent the overall stream from being finalized if an inner stream is in the process of being kicked off? (Or rather, to delay overall finalization until inner stream finishes being kicked off)

mpilquist:
I tried something quick based on that suggestion but now the hanging join test fails. I'll be back later tonight to try some more. My last commit will likely need to be reverted.

pchiusano:
Hmm if that test is failing it sounds like the semaphore might be held for
the duration of the inner stream, rather than just the critical section?

mpilquist:
OK, pushed some changes. With this version, I can't get it to fail locally but I've seen 2 failures on Travis -- one for "async resource allocation (2b)" and another for "join (1)". The former was a test failure of the form "-1 was not equal to 0" while the latter was a test timeout.

pchiusano:
Feels like we are playing whack-a-mole here. :\ I haven't seen join (1)
timeout that I can remember. I'm guessing there is a case where acquiring
that finalizeLock never completes (maybe because an inner stream died after
acquiring the lock somehow?)

The 2b failure is unrelated - that is a problem with merge that I'm
tracking down. Have an issue tracking it, on phone and can't link to it.

mpilquist:
OK, join(1) was timing out because we were reaching the end of the outer stream and never setting outerDone = true despite appearing to do so in finalize. I reordered the actions in the finalizer for the outer stream and everything appears to be working now.

mpilquist:
And because I claimed it is fixed, async resource allocation (3) just failed locally... Adding println calls slows things down to the point that the tests don't fail.

pchiusano:
ha of course. I should have some time tomorrow to look at this, if you are around let's put our heads together.

mpilquist:
I pushed a new implementation of `join` which is more obviously correct than the semaphore version and which also has the benefit of running the inner streams in order (the semaphore version raced the streams against each other with ordering determined arbitrarily). However, this version exposes a bug in async resource allocation. Specifically, the inner stream sometimes hits the sys.error call here: https://github.com/functional-streams-for-scala/fs2/blob/topic/redesign/core/src/main/scala/fs2/internal/Resources.scala#L121

This results in a hang, and I suspect this is the cause of the funkiness in the original version as well.

mpilquist:
I think this PR fixes the issue once the bug exposed in #597 is fixed.

PR #596, opened 2016-05-09
finishAcquire now returns a Boolean, and this is handled in various places

pchiusano:
Previously we were getting exceptions from this.

New behavior is to run the finalizer immediately if finishAcquire fails, under the assumption (which I'm not positive is valid) that if the key has been removed from the map, it's because the stream or scope is being closed.


mpilquist:
This looks good locally but the 1 million brackets in sequence test hung here: https://travis-ci.org/functional-streams-for-scala/fs2/jobs/128903158. Holding off on merge.

PR #603, opened 2016-05-12
Fixed a bug where finalizers of acquired resources are skipped...

mpilquist:
when an unconsAsync Resources object is closed via closeAll with at least one resource with an unfinished acquisition.

This is the cause of bugfix/591 failures in resource acquisition (3).

@pchiusano


pchiusano:
Wow good catch. What a dumb mistake on my part. I'll bet this fixes the 2b failures as well.

PR #604, opened 2016-05-12
Reimplemented join in terms of signals

mpilquist:
This is a squashed version of #595 with some fixes.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,2 +1,2 @@
--J-Xms1g
--J-Xmx1g
+-J-Xms256m
+-J-Xmx256m
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,2 +1,2 @@
--J-Xms256m
--J-Xmx256m
+-J-Xms1g
+-J-Xmx1g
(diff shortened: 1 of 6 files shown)


pchiusano:
Hmm, Travis still failing on join(1) though. Looks like a missing output value, perhaps due to not waiting long enough for inner streams to finish?

mpilquist:
@pchiusano Yeah, I have some ideas today so hold off on this PR for now. There's a race condition in this implementation of join, where if we get unlucky timing wise, we can more than maxOpen and then the results come out of order in the test where we `join(1)`.

core/src/test/scala/fs2/ResourceSafetySpec.scala
```diff
@@ -101,14 +101,32 @@ class ResourceSafetySpec extends Fs2Spec with org.scalatest.concurrent.Eventuall
...
+      val b1 = bracket(c) { s1.get }
+      val b2 = bracket(c) { s2.get }
+      runLog { b1 merge b2 take 1 }
       swallow { runLog { spuriousFail(b1,f1) merge b2 }}
       swallow { runLog { b1 merge spuriousFail(b2,f2) }}
       swallow { runLog { spuriousFail(b1,f1) merge spuriousFail(b2,f2) }}
       c.get shouldBe 0L
     }
 
+    "asynchronous resource allocation (2c)" in {
```
pchiusano: @mpilquist fyi this test was previously failing reliably before your fix.

PR #605, opened 2016-05-12
Reimplemented join.

mpilquist:
- New join implementation is fairer than semaphore based implementation.
  This implementation ensures inner streams are started in the order
  they appear in the outer stream. The semaphore implementation raced the
  inner streams against the semaphore decrement.
- Changed the synchronous none terminated queue such that enqueuing a
  `Some` does not block after a `None` has been enqueued.
- Fixed a bug in `Task.get` and similar methods in which message ids
  were reused across multiple executions of the task. This resulted in
  hung tasks under contention. For example, calling `signal.get` and then running that task multiple times in parallel could lead to some of the tasks never completing.

Review by @pchiusano.

### DIFF
--- a/core/src/main/scala/fs2/async/mutable/Queue.scala
+++ b/core/src/main/scala/fs2/async/mutable/Queue.scala
@@ -145,16 +145,23 @@ object Queue {
   /** Like `Queue.synchronous`, except that an enqueue or offer of `None` will never block. */
   def synchronousNoneTerminated[F[_],A](implicit F: Async[F]): F[Queue[F,Option[A]]] =
     F.bind(Semaphore(0)) { permits =>
+    F.bind(F.refOf(false)) { doneRef =>
     F.map(unbounded[F,Option[A]]) { q =>
       new Queue[F,Option[A]] {
         def upperBound: Option[Int] = Some(0)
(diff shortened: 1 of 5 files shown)


mpilquist:
Resource allocation (3) failed and I've reproduced locally. Hopefully this will be an easy fix.

pchiusano:
Wow nice! Travis is still hanging though?

mpilquist:
Nah, I committed the wrong file. Next build will pass, I'm sure of it!

mpilquist:
@pchiusano I think we'll likely want to use the same inner/outer counter pattern in any of the resource allocation tests that use join.

pchiusano:
😱

mpilquist:
Travis is green! 🎉

pchiusano:
🎉!!!!

mpilquist:
At the risk of being hasty, I'm going to merge this unless you have an objection. Then I'm going to prep for the M1 release.

pchiusano:
I was going to say go for it. The change to Task looks good.

pchlupacek:
@mpilquist, @pchiusano  excellent. You think there is a change to get tcp in M1?

DavidGregory084:
@pchlupacek I think you should close and reopen your PRs to restart the Travis builds :smile:

PR #609, opened 2016-05-14
Ported the start here example from 0.8

mpilquist:
- I created a tut doc for the example, though I didn't know where to include that in the guide, so it is standalone for now.

- I changed the signature of `utf8Encode` and `utf8Decode` -- both were working with `Chunk[Byte]` instead of `Byte` which made using them with the io package annoying. The original signatures are available under a `C` suffix -- `utf8EncodeC` and `utf8DecodeC`.  The implementations did not exploit the chunky nature in any way, except for the handling of encoding/decoding empty strings -- `utf8EncodeC andThen utf8DecodeC == id` for all strings, whereas `utf8Encode andThen utf8Decode == id` for all non-empty strings. I believe this is due to the fact that we don't emit empty chunks in the interpreter.

- Added a new text.lines pipe, which buffers a stream of strings and emits individual lines.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -105,6 +105,13 @@ lazy val publishingSettings = Seq(
   }
 )
 
+lazy val noPublish = Seq(
+  publish := (),
+  publishLocal := (),
+  publishSigned := (),
+  publishArtifact := false
+)
+
 lazy val releaseSettings = Seq(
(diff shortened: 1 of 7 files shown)


pchiusano:
Cool. There was no good reason for signature of those text transducers, it was just an oversight.

PR #610, opened 2016-05-15
Introduced `Async.Run`

pchlupacek:
@pchiusano @mpilquist. This is to allow to write libraries using fs2 that do multiple callbacks from outside the fs2 domain. For example you need to enqueue messages from the `Polling` thread to our queue, and still you don't want to specialize to `Task`. 

I didn't find any better alternative than this, not sure if Option[Throwable] is right idea, perhaps we shall just exception to be thrown. Any thoughts on that?

### DIFF
--- a/core/src/main/scala/fs2/Async.scala
+++ b/core/src/main/scala/fs2/Async.scala
@@ -188,4 +188,35 @@ object Async {
    * is the new value computed by `f`.
    */
   case class Change[+A](previous: A, now: A)
+
+
+  /**
+    * Used to strictly evaluate `F`.
+    */
+  trait Run[F[_]]  {
+
+    /**
+      * Run this `F` and block until it completes. Performs Side effects.
+      * If the evaluation of the `F` terminates with an exception,
+      * then this will return that exception as Optional value.
+      * Otherwise this terminates with None.
+      *
+      * Note that hence this blocks, special care must be taken on thread usage.
+      * Typically, this has to be run off the thread that allows blocking and
+      * is outside the main Executor Service or Scheduler.
+      *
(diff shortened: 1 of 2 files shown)


pchiusano:
Are we using this anywhere internally? I don't think so, right?

I can sort of see the need for this, but I really want to discourage its
use. Part of me thinks it belongs in a separate library, not in core. Wdyt?

mpilquist:
I think this would let us change the `time` package to work for any `F: Async` instead of being fixed to `Task`

mpilquist:
👍 from me. I have no strong opinions on `Option[Throwable]` vs throwing, though I lean towards the latter. I also think we should rename `runEffects` to `unsafeRunEffects` or `unsafeRun` or something.

pchlupacek:
@pchiusano I understand your concerns. However as @mpilquist pointed out it simplifies time package and allows us it to be not specialised to Task as well. Essentially it is useful everywhere where you need to run multiple callbacks. 

I was also investigating to have it in Async trait itself, hence we have this private runSet there already. But at the end inclined to make it separate trait, as it will clearly show there is some `efectfull` stuff going in the implementation.   

@mpilquist I am ok with throwing the exception instead.

pchiusano:
Okay, I'm convinced. :)

Agree with @mpilquist about naming.

Don't care about `Option[Throwable]` vs `Unit`. Using `Option[Throwable]` seems less error prone, since not all users will have the no discard unit flag turned on.

mpilquist:
@pchlupacek OK, I renamed `runEffects` to `unsafeRunEffects` and left the return type as `Option[Throwable]`.

PR #612, opened 2016-05-16
Miscellaneous cleanup related to time

mpilquist:
- Introduced `Scheduler` trait as wrapper for `ScheduledExecutorService`
 - Replaced usage of `Duration` with `FiniteDuration`
 - Removed old methods implemented in terms of millisecond timeouts

I based this off of #611 but if we decide not to merge that, I can fix this PR.

### DIFF
--- a/core/src/main/scala/fs2/Scheduler.scala
+++ b/core/src/main/scala/fs2/Scheduler.scala
@@ -7,21 +7,21 @@ import java.util.concurrent.{ Executors, ScheduledExecutorService, TimeUnit }
 trait Scheduler {
 
   /**
-   * Evaluates the specified thunk after the specified delay.
+   * Evaluates the thunk using the strategy after the delay.
    * Returns a thunk that when evaluated, cancels the execution.
    */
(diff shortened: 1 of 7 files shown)


core/src/main/scala/fs2/Scheduler.scala
```diff
@@ -0,0 +1,47 @@
+package fs2
+
+import scala.concurrent.duration.FiniteDuration
+import java.util.concurrent.{ Executors, ScheduledExecutorService, TimeUnit }
+
+/** Provides the ability to schedule evaluation of thunks in the future. */
+trait Scheduler {
+
```
pchlupacek: I was sort of thinking on this earlier. I am not sure if this shouldn't demand `S` in scope, and essentially return to `S` as soon as the timer is fired. Really I think Scheduler shall be only timing service, and never the executor thread, so we are quite sure where the programs run at the end. Any thoughts?

mpilquist: @pchlupacek Completely agree. I just pushed a fix for this -- let me know what you think.

PR #613, opened 2016-05-16
Added sliding combinator

mpilquist:


mpilquist:
I'm not really sure if `sliding` should output `Chunk[I]` or `Vector[I]`...

pchiusano:
Vector seems fine and will have better sharing.

PR #614, opened 2016-05-16
Revised, expanded, and tut-checked user guide

pchiusano:
Added an alias for `semaphore` in `async` package object while I was at it. Was easier to tell people - go check out the package object rather than 'go check the package object for more, except for this one primitive you might use a lot which is in a totally different spot'.


pchiusano:
Hmm, I have been unsuccessful in getting relative links to work. This would be nice, allows people to just read the guide on github, with links going to the corresponding branch in the code.

mpilquist:
Relative links are working for me -- make sure you click the link from the code browser, not from the PR.

docs/guide.md
```diff
@@ -137,61 +414,290 @@ For the recursive call, we update the state, subtracting the `chunk.size` elemen
...
+
+A `Future[F,A]` represents a running computation that will eventually yield an `A`. A `Future[F,A]` has a method `.force`, of type `Pull[F,Nothing,A]` that can be used to block until the result is available. A `Future[F,A]` may be raced with another `Future` also---see the implementation of [`pipe2.merge`](../core/src/main/scala/fs2/pipe2).
+
+In addition, there are a number of other concurrency primitives---asynchronous queues, signals, and semaphores. See the [`async` package object](../core/src/main/scala/fs2/async/async.scala) for more details. We'll make use of some of these in the next section when discussing how to talk to the external world.
+
+#### Exercises
+
+Without looking at the implementations, try implementing `pipe2.interrupt` and `pipe2.mergeHaltBoth`:
+
+```Scala
```
adelbertc: Accidental capitalization?

docs/guide.md
```diff
@@ -13,54 +17,330 @@ We'll consider each of these in turn.
...
+def eval[F[_],A](f: F[A]): Stream[F,A]
+```
+
+_Note_: FS2 does not care what effect type you use for your streams. You may use the included [`Task` type][Task] for effects or bring your own, just by implementing a few interfaces for your effect type. ([`Catchable`][Catchable] and optionally [`Async`][Async] if you wish to use various concurrent operations discussed later.)
+
+[Task]: ../core/src/main/scala/fs2/util/Task.scala
+[Catchable]: ../core/src/main/scala/fs2/util/Catchable.scala
+[Async]: ../core/src/main/scala/fs2/util/Async.scala
+
+It produces a stream that evaluates the given effect, then emits the result (notice that `F` is unconstrained) Any `Stream` formed using `eval` is called 'effectful' and can't be run using `toList` or `toVector`. If we try we'll get a compile error:
```
mpilquist: Missing period... `is unconstrained) Any`

pchiusano:
Okay, yeah, they are working. I had checked one that was coincidentally pointing to the wrong package and assumed they were all busted.

docs/guide.md
pchiusano: GFM doesn't care, I usually capitalize.

mpilquist:
👍

PR #618, opened 2016-05-17
Update File I/O in fs2.io.file.

fiadliel:
- Move FileHandle creation methods to FileHandle object.
 - Methods that return a `Pull` move to `fs2.io.file.pulls`.
 - Can create a `Pull` from `Path`, `FileChannel`, or `AsynchronousFileChannel`.
 - Add some Markdown documentation for the code.

### DIFF
--- a/docs/src/HandlingFiles.md
+++ b/docs/src/HandlingFiles.md
@@ -1,0 +1,136 @@
+# Handling Files
+
+The package `fs2.io.file` contains code which allows reading from, and writing to, local
+files. This is implemented using calls to the `java.nio` package.
+
+## Important imports
+
+```tut:book
+import fs2._
+import fs2.util.Task
+import java.nio.file._
+```
+
+And for asynchronous calls, we will need a `Strategy` (this provides details on
+  where to run callbacks, so we don't over-use NIO threads):
+```tut:book
+implicit val strategy = Strategy.fromCachedDaemonPool("async-file-io")
+```
+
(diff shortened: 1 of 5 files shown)


docs/src/HandlingFiles.md
```diff
@@ -0,0 +1,136 @@
...
+  )
+)
+```
+We can also open files asynchronously with the `io.file.pulls.fromAsynchronousFileChannel` call.
+
+## The `Pull` monad and `FileHandle` operations
+
+Giving the ability to manipulate files within the `Pull` monad provides
+a lot of power. It is instructive to look at one of the simpler implementations, [pulls.readAllFromFileHandle](https://github.com/functional-streams-for-scala/fs2/tree/topic/redesign/io/src/main/scala/fs2/io/file/pulls.scala).
+
```
pchlupacek: that links returns to me 404. Also wouldn't it be nice to show one example of how Pulls helps here too?

fiadliel: The file should exist after this commit.

Yes, an example would be good, I'll add something.

pchiusano: Use relative links like is done in guide.md please.

mpilquist:
When you get a chance to update this PR, could you open it against series/0.9 instead?

mpilquist:
FYI: this got auto-closed b/c we deleted topic/redesign branch.

PR #619, opened 2016-05-17
Updated README and guide, removed use of hash codes toString methods …

pchiusano:
…everywhere


core/src/main/scala/fs2/StreamCore.scala
```diff
@@ -417,8 +417,7 @@ object StreamCore {
     def pushBind[O0](f: O0 => StreamCore[F,O1]): Stack[F,O0,O2] = pushBindOrMap(Right(f))
     def pushMap[O0](f: Chunk[O0] => Chunk[O1]): Stack[F,O0,O2] = pushBindOrMap(Left(f))
     def pushBindOrMap[O0](f: Either[Chunk[O0] => Chunk[O1], O0 => StreamCore[F,O1]]): Stack[F,O0,O2] = new Stack[F,O0,O2] {
-      def render = f.fold(ff => s"Map(f#${System.identityHashCode(ff)})",
-                          ff => s"Bind(f#${System.identityHashCode(ff)})") :: self.render
+      def render = f.fold(ff => s"Map", ff => s"Bind") :: self.render
```
mpilquist: Don't need the s interpolator anymore

pchiusano: Good call.

pchiusano: Fixed

PR #621, opened 2016-05-18
Introduced new Effect type class

mpilquist:
Moved Monad.suspend to Effect.delay and added new Effect.suspend with same signature as Task.suspend. See [conversation on Gitter](https://gitter.im/functional-streams-for-scala/fs2?at=573c137cae26c1967f9f7efa) for motivation. This needs more docs and updates to the guide. I wanted to open the PR for discussion first (not for merge). If folks tend to agree, I'll write some ScalaDoc and update the guide.

### DIFF
--- a/core/src/main/scala/fs2/util/Effect.scala
+++ b/core/src/main/scala/fs2/util/Effect.scala
@@ -1,6 +1,15 @@
 package fs2.util
 
 trait Effect[F[_]] extends Catchable[F] {
-  def delay[A](a: => A): F[A]
+
+  /**
+   * Returns an `F[A]` that evaluates and runs the provided `fa` on each run.
+   */
   def suspend[A](fa: => F[A]): F[A]
+
+  /**
+   * Promotes a non-strict value to an `F`, catching exceptions in the process.
+   * Evaluates `a` each time the returned effect is run.
+   */
+  def delay[A](a: => A): F[A] = suspend(pure(a))
 }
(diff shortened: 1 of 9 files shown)


core/src/main/scala/fs2/util/Effect.scala
```diff
@@ -0,0 +1,6 @@
+package fs2.util
+
+trait Effect[F[_]] extends Catchable[F] {
+  def delay[A](a: => A): F[A]
```
pchiusano: Perhaps delay should be defined as `delay(a) == suspend(pure(a))` by default.

core/src/main/scala/fs2/util/Task.scala
```diff
@@ -395,14 +395,15 @@ object Task extends Instances {
 
 /* Prefer an `Async` and `Catchable`, but will settle for implicit `Monad`. */
```
pchiusano: Probably should update this comment

fiadliel:
I wrote a comment, but it felt like bike-shedding :-) In the end, as long as it exists, in a type class that's somewhat easy to re-implement, it seems good to me.

Because one extra benefit of `Monad` over `Async`, which partially motivated the non-async file I/O, is that `Async` doesn't look completely trivial to write an instance for. Might make life easier for a few people to keep this property.

mpilquist:
OK, I updated docs and the guide. Should be ready for merge unless there are other comments.

PR #625, opened 2016-05-21
Introduced Lub1 witness and modified ++/append/evalMap/flatMap/onError/onComplete to use it in order to get better inference

### LINKED ISSUE #624: Appending pure stream to `F` stream yields in error
Seems the pure type don't fit yet correctly :
 
```scala

scala> val s1:Stream[Task,String] = Stream("A","B")
// s1: fs2.Stream[fs2.util.Task,String] = fs2.Stream$$anon$1@62cff145

scala> val s2:Stream[Pure,String] = Stream("C","D")
// s2: fs2.Stream[fs2.Pure,String] = fs2.Stream$$anon$1@2d639a64

scala> s1 ++ s2
//<console>:24: error: could not find implicit value for parameter S: fs2.util.Sub1[fs2.util.Task,fs2.Pure]

```

mpilquist:
Fixes #624. What do you think of this general approach? What other methods should get the `Lub1` treatment?

### DIFF
--- a/core/src/main/scala/fs2/StreamDerived.scala
+++ b/core/src/main/scala/fs2/StreamDerived.scala
@@ -212,12 +212,12 @@ trait StreamDerived extends PipeDerived { self: fs2.Stream.type =>
     /** Transform this stream using the given `Pipe`. */
     def through[B](f: Pipe[F,A,B]): Stream[F,B] = f(s)
     /** Transform this stream using the given pure `Pipe`. */
-    def throughp[B](f: Pipe[Pure,A,B]): Stream[F,B] = f(s)
+    def throughPure[B](f: Pipe[Pure,A,B]): Stream[F,B] = f(s)
     /** Transform this stream using the given `Pipe2`. */
     def through2[B,C](s2: Stream[F,B])(f: Pipe2[F,A,B,C]): Stream[F,C] =
       f(s,s2)
     /** Transform this stream using the given pure `Pipe2`. */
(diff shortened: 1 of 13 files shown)


mpilquist:
I added lubbing support for `onComplete` and `onError` as well. I think that should cover everything.

pchlupacek:
@mpilquist to me this looks ok.  On the through side. I would like to keep separate combinators, but would like to rename throughp to perhaps throughPure to clearly mark the pure code is within. What do you think?

mpilquist:
@pchlupacek I'm good with renaming `throughp` to `throughPure`. Use of `throughPure` would end up as a style choice in most cases, b/c our pipes are still defined for all `F[_]`. Use of `throughPure` would only be required when you have a `Stream[Nothing, A]` and you want to use a `forall F . Pipe[F, A, B]`.

pchlupacek:
@mpilquist exactly what I meant. Thanks.

pchiusano:
Hmm why not make the Lub a type member rather than another parameter. Then one less type parameter to specify if you ever have to annotate?

mpilquist:
@pchiusano I tried that first, but inference doesn't work as nicely due to the lub type being invariant.

To be clear, here's what I tried:
```
sealed trait Lub1[-F[_],-G[_]] {
  type Lub[_]
  implicit def subF: Sub1[F,Lub]
  implicit def subG: Sub1[G,Lub]
}
```

pchiusano:
Okay. Seems fine to me then.

PR #627, opened 2016-05-23
Initial implementation of Topic

pchlupacek:
@mpilquist @pchiusano This is a very rough idea on the topic. Can you please comment on Api? In contrast of 0.8 this:

 - Allows to set limit per subscriber to stop (block) any publish if that subscriber is behind that amount of messages
 - Honors chunking of input stream
 - Subscriber always receive single value, even when there was no value yet published. 
 - There is a variant of `subscribe` with approximate value of queue size, that eventually may be used to fine-tune dequeue operations based on size of the queue of that subscriber. 
 
I don't think so there will be need for WriterTopic as in 0.8. I think we should be able to get Writer topic behaviour with this defined as Topic[F,(W,O)]

I am also thinking on ways how to simplify implementation, which is now WIP. I am thinking on implementing some sort of buffer with low/high watermarks instead keeping buffer per subscriber. But if you have any other idea, please share :-)

### DIFF
--- a/core/src/main/scala/fs2/async/mutable/Topic.scala
+++ b/core/src/main/scala/fs2/async/mutable/Topic.scala
@@ -57,7 +57,7 @@ trait Topic[F[_],A] {
     * Additionally this emits current size of the queue of `A` for this subscriber allowing
     * you to terminate (or adjust) the subscription if subscriber is way behind the elements available.
     *
-    * Note that queue size is approximate and may not be exactly the size when `a` was take.
+    * Note that queue size is approximate and may not be exactly the size when `A` was taken.
     *
     * If the subscriber is over `maxQueued` bound of messages awaiting to be processed,
     * then publishers will hold into publishing to the queue.
--- a/core/src/main/scala/fs2/async/mutable/Topic.scala
+++ b/core/src/main/scala/fs2/async/mutable/Topic.scala
@@ -1,0 +1,250 @@
+package fs2.async.mutable
+
+import fs2.Async.Change
+import fs2._
+import fs2.Stream._
+
+/**
+  * Asynchronous Topic.
+  *
+  * Topic allows you to distribute `A` published by arbitrary number of publishers to arbitrary number of subscribers.
+  *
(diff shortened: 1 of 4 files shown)


core/src/main/scala/fs2/async/mutable/Topic.scala
```diff
@@ -0,0 +1,250 @@
...
+    *
+    * If the subscriber is over `maxQueued` bound of messages awaiting to be processed,
+    * then publishers will hold into publishing to the queue.
+    *
+    */
+  def subscribeSize(maxQueued:Int):Stream[F,(Int,A)]
+
+
+  /**
+    * Discrete stream of current active subscribers
```
mpilquist: This is a discrete stream of the _count_ of currently active subscribers.

core/src/main/scala/fs2/async/mutable/Topic.scala
```diff
@@ -0,0 +1,250 @@
...
+  *
+  * Additionally the subscriber has possibility to terminate whenever size of enqueued elements is over certain size
+  * by using `subscribeSize`.
+  *
+  *
+  */
+trait Topic[F[_],A] {
+
+  /**
+    * Published any elements from source of `A` to this topic. This will stop publishing if the source has no elements
```
mpilquist: What is meant by "stop publishing" here? Seems like in some use cases you'd want to block here, waiting for subscriber to dequeue. E.g., does this differ from `source.evalMap(topic.publish1)`?

pchlupacek: Ok the comment is not that up-to-date. I changed signature to sink. What I want to say here is that this will asynchronously block if any subscriber is over is maxQueued threshold. Does that make a sense?

core/src/main/scala/fs2/async/mutable/Topic.scala
pchlupacek: Yeah will update the comment.

pchlupacek:
@mpilquist @pchiusano I believe this work is completed. LMK if there are any issues

mpilquist:
This looks great! There are some nitpicky things I could comment on but I'd rather merge this now and then adjust nitpicky stuff in a separate PR.

PR #628, opened 2016-05-23
Fixed deadlock in awakeEvery

mpilquist:
Use of `awakeEvery` may deadlock if the `Strategy` is used for things other than `awakeEvery`. See [Gitter](https://gitter.im/functional-streams-for-scala/fs2?at=574301fbd3f431720bb2e335) for discussion.

### DIFF
--- a/core/src/main/scala/fs2/Async.scala
+++ b/core/src/main/scala/fs2/Async.scala
@@ -193,14 +193,13 @@ object Async {
   trait Run[F[_]]  {
 
     /**
-     * Asynchronously run this `F` and block until it completes. Performs side effects.
-     * If the evaluation of the `F` terminates with an exception, then the callback
-     * will be called with a `Left(t)`. Otherwise, the callback will be called with
-     * a `Right`.
+     * Asynchronously run this `F`. Performs side effects.
+     * If the evaluation of the `F` terminates with an exception, then the `onError`
+     * callback will be called.
      *
      * Purpose of this combinator is to allow libraries that perform multiple callback
      * (like enqueueing the messages, setting signals) to be written abstract over `F`.
      */
-    def unsafeRunAsyncEffects(f: F[Unit])(cb: Either[Throwable,Unit] => Unit): Unit
+    def unsafeRunAsyncEffects(f: F[Unit])(onError: Throwable => Unit): Unit
   }
 }
--- a/core/src/main/scala/fs2/Async.scala
+++ b/core/src/main/scala/fs2/Async.scala
@@ -189,29 +189,18 @@ object Async {
    */
   case class Change[+A](previous: A, now: A)
 
-
-  /** Used to strictly evaluate `F`. */
+  /** Used to evaluate `F`. */
   trait Run[F[_]]  {
 
     /**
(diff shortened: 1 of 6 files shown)


core/src/test/scala/fs2/time/TimeSpec.scala
```diff
@@ -15,6 +15,11 @@ class TimeSpec extends Fs2Spec {
       time.awakeEvery[Task](100.millis).map(_.toMillis/100).take(5).runLog.unsafeRun shouldBe Vector(1,2,3,4,5)
     }
 
+    "awakeEvery liveness" in {
+      val s = time.awakeEvery[Task](1.milli).evalMap { i => Task.async[Unit](cb => S(cb(Right(())))) }.take(200)
+      runLog { concurrent.join(5)(Stream(s, s, s, s, s)) }
+    }
```
jrudnick: Without the fix, this test hangs or does it fail in some other way?

mpilquist: Ya, it hangs and after a minute, ScalaTest fails that test.

core/src/main/scala/fs2/Async.scala
```diff
@@ -189,29 +189,18 @@ object Async {
...
+     * Asynchronously run this `F` and block until it completes. Performs side effects.
+     * If the evaluation of the `F` terminates with an exception, then the callback
+     * will be called with a `Left(t)`. Otherwise, the callback will be called with
+     * a `Right`.
      *
      * Purpose of this combinator is to allow libraries that perform multiple callback
      * (like enqueueing the messages, setting signals) to be written abstract over `F`.
      */
-    def unsafeRunEffects(f: F[Unit]): Option[Throwable]
+    def unsafeRunAsyncEffects(f: F[Unit])(cb: Either[Throwable,Unit] => Unit): Unit
```
pchlupacek: Do you think the right side is useful here? My first feeling would be just to define it as Throwable => Unit, we really shall discourage to use this for any chaining as this will likely result in SOE.

core/src/main/scala/fs2/util/Task.scala
```diff
@@ -428,8 +428,8 @@ private[fs2] trait Instances extends Instances1 {
   }
 
   implicit val runInstance: Async.Run[Task] = new Async.Run[Task] {
-    def unsafeRunEffects(f: Task[Unit]): Option[Throwable] =
-      f.unsafeAttemptRun.left.toOption
+    def unsafeRunAsyncEffects(f: Task[Unit])(cb: Either[Throwable,Unit] => Unit) =
+      f.unsafeRunAsync(cb)
     override def toString = "Run[Task]"
```
pchlupacek: I am not sure, if we shouldn't for to strategy here asap. so like S { f.unsafeRunAsync(cb)  }

mpilquist: Yep agreed, just pushed this change.

core/src/main/scala/fs2/Async.scala
mpilquist: Agreed, pushed a fix.

PR #629, opened 2016-05-23
Update file NIO support

fiadliel:
- Create Pull for FileHandle from either paths or channels
 - Move Pull constructors to fs2.io.pulls
 - Move FileHandle constructors to fs2.io.FileHandle companion
 - Make readAllFromFileHandle/writeAllToFileHandle visible

This separates the extra page for the docs, as it isn't finished, and it might be nice if this went into 0.9.0-M2.

### DIFF
--- a/io/src/main/scala/fs2/io/file/FileHandle.scala
+++ b/io/src/main/scala/fs2/io/file/FileHandle.scala
@@ -1,6 +1,10 @@
 package fs2.io.file
 
-import fs2.Chunk
+import java.nio.ByteBuffer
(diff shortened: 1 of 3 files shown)


mpilquist:
👍 
> This separates the extra page for the docs, as it isn't finished, and it might be nice if this went into 0.9.0-M2.

Was this comment referring to a pending doc change?

fiadliel:
Yes, I'd included doc updates in the previous PR (before it was closed), and all outstanding issues were with them. Since I think it would be nice to get this into M2, I'll put them into a different PR.

PR #631, opened 2016-05-24
Renamed pull.run to pull.close

mpilquist:
I think `close` is clearer for folks learning the library:
 - `close` helps separate the act of closing a pull from the act of running a stream
 - symmetrical with `open` on `Stream`


pchiusano:
Seems fine to me.

PR #633, opened 2016-05-26
Fix broken link-to-source in scaladocs

ktonga:
Guess branch/tag based on version.

Currently it is hardcoded pointing to `master` which was deleted a while ago. Also I think it's better to point to the tag for released versions.


mpilquist:
👍

PR #635, opened 2016-05-28
Changed signature of SIgnal.modify

pchlupacek:
@mpilquist like we have discussed earlier. This changes signature of `modify` to be almost the same like `modify` in `Async`. Furthermore this simplifies signal's state and removes the `Semaphore` from the implementation.

### DIFF
--- a/core/src/main/scala/fs2/Async.scala
+++ b/core/src/main/scala/fs2/Async.scala
@@ -189,6 +189,13 @@ object Async {
    */
   case class Change[+A](previous: A, now: A)
 
+  object Change {
+    implicit class ChangeSyntax[A](val self:Change[A]) extends AnyVal {
+      def modified: Boolean = self.previous != self.now
+      def map[B](f: A => B):Change[B] = Change(f(self.previous), f(self.now))
+    }
+  }
+
   /** Used to evaluate `F`. */
   trait Run[F[_]]  {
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/fs2/Async.scala
```diff
@@ -189,6 +189,13 @@ object Async {
    */
   case class Change[+A](previous: A, now: A)
 
+  object Change {
+    implicit class ChangeSyntax[A](val self:Change[A]) extends AnyVal {
```
mpilquist: Why a syntax class instead of directly on the case class above?

mpilquist:
Looks good to me. I'm going to merge now, though can you take a look at the ChangeSyntax question at some point?

core/src/main/scala/fs2/Async.scala
pchlupacek: I am just used to do it that way. Is more less style, and form my understanding the compiler emits same code at the end.

mpilquist: Performance wise, they are likely equivalent -- monomorphic method on the class vs a monomorphic method on the companion. Neither will result in object allocation. I'd avoid the syntax style unless you need it to work around variance -- e.g., defining methods that fix a covariant/contravariant type param to invariant.

pchlupacek: Yeah. I used to this approach because this guarantees there won't be any hidden state kept in case class. I have no strong preference on this, so lets target change of it with any other PR related to signal? Is that ok with you or you think is worth of separate PR?

PR #638, opened 2016-05-30
Redesigned UDP - WIP / do not merge

mpilquist:
@pchlupacek Here's a take on what we discussed for UDP support. Let me know what you think. I still need to add tests and an API for joining multicast groups.


pchlupacek:
@mpilquist seems ok to me. Hence this implements own selector, we perhaps needs to write couple more tests than for tcp, but I think is actually great we will have this.

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private val doneNow = new AtomicBoolean(false)
+    private val selectorThread = new Thread(new Runnable {
+      def run = try {
+        val readBuffer = ByteBuffer.allocate(1 << 16)
+        while (!doneNow.get && !Thread.currentThread.isInterrupted) {
+          pendingThunks.synchronized { while (pendingThunks.nonEmpty) pendingThunks.dequeue()() }
+          selectorLock.lock
+          selectorLock.unlock
```
pchlupacek: not sure if lock/unlock was intended here?

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
```diff
@@ -0,0 +1,146 @@
...
+                      if (!attachment.hasWriters) key.interestOps(key.interestOps & ~SelectionKey.OP_WRITE)
+                    }
+                  case None =>
+                    sys.error("key marked for write but no writer")
+                }
+              }
+            }
+          }
+        }
+      } catch { case t: Throwable => t.printStackTrace }
```
pchlupacek: perhaps will just dump to stderr instead to stdout

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
```diff
@@ -0,0 +1,146 @@
...
+                      if (!attachment.hasReaders) key.interestOps(key.interestOps & ~SelectionKey.OP_READ)
+                    case None =>
+                      sys.error("key marked for read but no reader")
+                  }
+                }
+              }
+              if (key.isWritable) {
+                attachment.peekWriter match {
+                  case Some((p, cb)) =>
+                    val sent = channel.send(ByteBuffer.wrap(p.bytes.toArray), p.remote)
```
pchlupacek: Wonder how to handle I/O here. Not sure if we should not fork it? and then invoke cb with error or None.

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
```diff
@@ -0,0 +1,146 @@
...
+          val selectedKeys = selector.selectedKeys.iterator
+          while (selectedKeys.hasNext) {
+            val key = selectedKeys.next
+            selectedKeys.remove
+            val channel = key.channel.asInstanceOf[DatagramChannel]
+            val attachment = key.attachment.asInstanceOf[Attachment]
+            if (key.isValid) {
+              if (key.isReadable) {
+                readBuffer.clear
+                val src = channel.receive(readBuffer)
```
pchlupacek: shouldn't we fork I/O here ?

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
mpilquist: Yeah intended -- the selector lock is used as a barrier so that we don't select before we can register a new channel.

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
mpilquist: I don't know -- the operation should be very quick b/c we know that a packet is available for reading. At the very least, the callback should be forked, which it isn't now.

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
mpilquist: If we forked the send operation, there would be a race condition with the select call, where we'd come of the select with key.isWritable = true but then the forked write operation would fill the send buffer.

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
mpilquist: Will do. This was just some debugging I had left here. I'll make this match TCP or Strategy.

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
pchlupacek: well the send can throw an exception at least this is what we have to handle. I am not sure what will happen  when send buffer will be full. Would this block until the send buffer will be empty?

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
pchlupacek: agreed like the read shall be almost non-blocking. As for the send however I think we have to handle IOException there

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
mpilquist: Agreed on exception handling. W.r.t. send buffer, if there's not enough room for the datagram, a negative number is returned immediately (this is only true when the channel is in non-blocking mode).

mpilquist:
@pchlupacek OK I pushed some changes for exception handling. Question about forking -- in order to do so, we need to capture a strategy somewhere. I noticed that the tcp package doesn't do this, however the `file` package does: https://github.com/functional-streams-for-scala/fs2/blob/series/0.9/io/src/main/scala/fs2/io/file/package.scala#L15

So I think we need to change the read/write methods on both `udp.Socket` and `tcp.Socket` to take a `Strategy`. Agree?

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
pchlupacek: ah of course you are right :-)

pchlupacek:
can't we just reuse strategy from F.Async? And then rather pass Async.Run to actually fork the evaluation to another thread?

mpilquist:
@pchlupacek Hm, so callback invocation would look like the following?

```
FR.unsafeRunAsyncEffects(F.delay(cb(...)))(_ => ())
```

pchlupacek:
yeah, that was the idea. I don't see any benefit for passing any Strategy that actually won't fork it, so to me passing implicit as Strategy to function of Socket doesn't make that much sense except we want to use that Executor Service to perform just fork. And you right we have to for the callback from the Executor backing the tcp AG to strategy of Async.

pchlupacek:
In fact, I am just thinking. How about if we will model the receive side like BoundedQueue and pass instead of `Either[Throwable,Packet] => Unit` just `Either[Throwable,Packet] => F[Unit]`  ???

On the send side, we perhaps can just pass pair (Packet, Option[Throwable] => F[Unit])

pchlupacek:
The bounded queue can allow us to not read from udp, if we are not ready (backpressure)

mpilquist:
@pchlupacek Doesn't the read side already have back pressure built in due to use of `repeatEval` and `F.async`? This form of back pressure works nicely with other stream combinators like `buffer` or `prefetch`.

pchlupacek:
@mpilquist you are right. Was just trying to find a solution we don't have to wrap that cb again in F.suspend(F.pure())

mpilquist:
@pchlupacek OK I made all of the above changes. I still need to add multicast join support, lots of tests, and some more docs. I'd appreciate if you could give this a quick look over though.

mpilquist:
Closing in favor of #640.

PR #640, opened 2016-05-31
Added fs2.io.udp package based on NIO and a shared selector thread

mpilquist:
Squashed version of #638. Ready for merge review.


mpilquist:
@pchlupacek I think this is ready for merge. Could you take a quick look through? Thanks

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
```diff
@@ -0,0 +1,234 @@
...
+        else Some(writers.head)
+      }
+
+      def dequeueWriter: Option[(Packet,Option[Throwable] => Unit)] = {
+        if (writers.isEmpty) None
+        else Some(writers.dequeue())
+      }
+
+      def queueWriter(writer: (Packet,Option[Throwable] => Unit)): Unit = {
+        if (closed) writer._2(Some(new ClosedChannelException))
```
pchlupacek: Just for curiosity, have you thought what will happen when two threads will compete for the one writer? I would say in current implementation it is not supported, which generally I am thinking is ok, but still I think hence the mutation of `writers` is not guarded by `synchronized` or `volatile` I am just curious if we cannot run in issues even when it is not accessed concurrently but from multiple threads sequentially w/o mut queue being flushed from the Thread cache.  

So what do you think put just either actor here to guard it or use some atomic/concurrent primitives to be safe ?

pchlupacek:
@mpilquist other than comment I left i think is excellent. I am really looking forward to use this.

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
mpilquist: I originally had synchronized blocks on the methods of `Attachment` but I switched to a model where the attachment state is only read and written from the selector thread -- which is a basically a weak/informal actor model. If we wanted to run multiple selector threads in parallel, we'd need to reintroduce the synchronization and then replace `peekWriter` with `pushWriter` or something -- e.g., replace the pattern of `peek/dequeue` with `dequeue/push on failure`.

What do you think? I can change to that model if you think we might want to support multiple selector threads at some point in the future.

pchlupacek: I am just ok with this. I perhaps didn't get completely concurrency model there. Was kind of worried that we do not access these from one thread. But if you say we do its ok and lgtm.

mpilquist:
Merging, per discussion on Gitter

PR #641, opened 2016-06-01
Added convenience function for converting a scala.concurrent.Future to a Task

mpilquist:


ceedubs:
I'm just seeing this. I'm used to this sort of conversion taking a by-name Future since Task usually represents a deferred and re-runnable effect. Should that be done here?

mpilquist:
@ceedubs Yeah it definitely should take a by-name. Thanks for catching this.

mpilquist:
@ceedubs BTW, this fix made 0.9.0-M6.

PR #642, opened 2016-06-02
Fixed bug in channel.observe(Async) where stream hangs if observing sink fails

mpilquist:
Review by @pchiusano


pchiusano:
LGTM. Seems like a silly oversight on my part. :(

PR #643, opened 2016-06-03
Fixed bug in channel.observe where observing with a finite sink caused hangs

mpilquist:
I don't love this implementation but it works... Review by @pchiusano.


pchiusano:
Kind of a weird implementation but seems ok.

PR #644, opened 2016-06-03
Fix @implicitNotFound hint about required Strategy

fiadliel:
When `Strategy` was moved from `fs2.util`, the package name in this hint wasn't changed.


docs/guide.md
```diff
@@ -491,7 +491,7 @@ FS2 comes with lots of concurrent operations. The `merge` function runs two stre
 ```scala
 scala> Stream(1,2,3).merge(Stream.eval(Task.delay { Thread.sleep(200); 4 })).runLog.unsafeRun
 <console>:17: error: No implicit `Async[fs2.util.Task]` found.
-Note that the implicit `Async[fs2.util.Task]` requires an implicit `fs2.util.Strategy` in scope.
```
mpilquist: I think we need the same change in docs/src/guide.md.

mpilquist: Bah, that's what I get for reviewing a PR before coffee.

PR #646, opened 2016-06-07
Jmh benchmark

yilinwei:
Since we've decided to start using `fs2` I thought we could take a look at #399.

This is really a very, very initial commit to show that it's running on travis and generate some discussion as to what should be benchmarked.

It would also be cool to have some updated benchmarks on github using a webhook or some such.


mpilquist:
@yilinwei Thanks for getting this going. I'm looking forward to using it. Have you considered using https://scalameter.github.io instead of JMH? I think at one point, ScalaMeter was an API on top of JMH. Not sure if that's still the case though.

mpilquist:
Sounds good to me. I've only use JMH and I'm happy to use it on fs2.

yilinwei:
I've taken off the DO NOT MERGE - that's not to say that the benchmarks are complete by any means, but they are in a state where they can be merge and be of some use.

I would appreciate a quick look over whether the benchmarks are sane since I'm not too familiar with the architecture.

mpilquist:
Looks great. I'd like to get the build to run faster (looks like ~1 hour currently). Merging though.

PR #652, opened 2016-06-11
Added timeout support to udp read/write

mpilquist:
This makes the API consistent with TCP.

### DIFF
--- a/io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
+++ b/io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
@@ -2,13 +2,16 @@ package fs2
 package io
 package udp
 
-import scala.collection.mutable.{Queue=>MutableQueue}
+import scala.collection.JavaConverters._
+import scala.collection.mutable.PriorityQueue
+import scala.concurrent.duration.FiniteDuration
 
 import java.io.IOException
 import java.net.InetSocketAddress
 import java.nio.ByteBuffer
-import java.nio.channels.{CancelledKeyException,ClosedChannelException,DatagramChannel,Selector,SelectionKey}
-import java.util.concurrent.CountDownLatch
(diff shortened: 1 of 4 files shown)


mpilquist:
PR build failure is due to the known TCP test issue, which I think is fixed in #651

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
```diff
@@ -32,55 +35,93 @@ object AsynchronousSocketGroup {
...
+          () => ()
+        } else {
+          val r = (reader, timeout)
+          readers.add(r)
+          () => { readers.remove(r); () }
         }
       }
 
-      def hasWriters: Boolean = writers.nonEmpty
+      def hasWriters: Boolean = !writers.isEmpty
```
pchlupacek: I know is matter of style only, but isn't nonEmpty more expressive?

mpilquist: It used to be `nonEmpty` but I had to change to a Java `ArrayDeque` for decent asymptotic random removal (on timeout expiration). The Java collection doesn't have a `nonEmpty`.

io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
```diff
@@ -32,55 +35,93 @@ object AsynchronousSocketGroup {
 
   def apply(): AsynchronousSocketGroup = new AsynchronousSocketGroup {
 
+    class Timeout(val expiry: Long, onTimeout: () => Unit) {
+      private var done: Boolean = false
```
pchlupacek: are you sure this does not have to be guarded by AtomicBoolean?

mpilquist: Yep, it is only read and written by the single UDP selector thread. Never from any other thread.

pchlupacek:
@mpilquist other than that cosmetic comment I do not have any more questions and lgtm

Issue #559, opened 2016-03-28
nio.file.chunkReadBuffer can return an underfilled buffer before EOF


refried:
like #498
relates to #538

refried:
backported to 0.8 and 0.8a in #561 and #563 respectively

mpilquist:
Leaving open so we make sure to address in 0.9 as well.

mpilquist:
@fiadliel We can close this right? The 0.9 file api doesn't have this issue.

fiadliel:
I would expect to get a read size equal to the requested number of bytes (other examples of this happening have been from sources other than files), but the API doesn't guarantee this.

I do have one small reservation about the idea, the `Pull` might not progress in pathological cases (where we continually get 0 bytes). But it does match what people seem to expect from the API, and chances of this happening are (hopefully) extremely low.

fiadliel:
Addendum: I'll send a PR.

mpilquist:
We decided it was best to support this use case by sending the output of the various file streams through one of the chunk management combinators, like `rechunkN`.

PR #655, opened 2016-06-18
Added rechunkN combinator and efficient chunk concat

mpilquist:


mpilquist:
Example use case is reading from input streams or sockets and rechunking the input in to a bigger chunks. E.g., #559.

mpilquist:
@pchiusano Any objections to this?

PR #657, opened 2016-06-18
Actually unboxed chunk transforms

djspiewak:
I feel bad about just email-whining about things and not actually pitching in, so here's my public apology to @mpilquist and @pchiusano for for privately being a jerk.  :-)

If I did things right, this should give us truly unboxed transformations on chunks where we already have an unboxed form (i.e. no longer annihilates primitives after a single operation).  I also fixed some low-hanging fruit elsewhere in `Chunk`.  Note that I couldn't figure out how to run the benchmarks (albeit with only about five minutes of trying) – and I'm on an airplane anyway, so the results wouldn't have been accurate – so we don't have objective metrics yet.  The bytecode seems to look good though.

Yes, it's ugly.  Sorry.  Manually specialized code is always very ugly.  Please do be careful about adjusting some of the encodings.  There are some relatively obvious ways to make the code suck less which actually harm performance.

Oh, I also tried (just for giggles) to apply `@specialized` to `Chunk.singleton`.  It sort of works (at least, it does based on reading the bytecode), but critically it doesn't pass that specialization down to inherited members (I doubted it would).  So if we want to specialize `singleton`, we need to do it manually, just like the others.  I left it alone for now, but it's a bit weird since it means you get unboxed primitives with multiples, but not with the single-element special case.


core/src/main/scala/fs2/Chunk.scala
```diff
@@ -174,15 +176,72 @@ object Chunk {
     new Doubles(values, offset, size)
   }
 
+  // justification of the performance of this class: http://stackoverflow.com/a/6823454
+  // note that Class construction checks are nearly free when done at load time
+  // for that reason, do NOT make the extractors into lazy vals or defs!
   private object Respecialization {
     // TODO set by completely random guess. please tune!
-    val Threshold = 1000
+    val Threshold = 0
```
djspiewak: We need to tune this parameter.  It basically sets the minimum chunk length, below which we will not attempt to respecialize (as there is a very slight constant-time, chunk-level overhead to doing so).  I'm *guessing* that this value should be fairly low (something on the order of 10), but we should measure.  It almost certainly should not be 0.

core/src/main/scala/fs2/Chunk.scala
```diff
@@ -174,6 +176,74 @@ object Chunk {
     new Doubles(values, offset, size)
   }
 
+  // justification of the performance of this class: http://stackoverflow.com/a/6823454
+  // note that Class construction checks are nearly free when done at load time
+  // for that reason, do NOT make the extractors into lazy vals or defs!
+  private object Respecialization {
+    // TODO set by completely random guess. please tune!
+    val Threshold = 0
```
djspiewak: We need to tune this parameter. It basically sets the minimum chunk length, below which we will not attempt to respecialize (as there is a very slight constant-time, chunk-level overhead to doing so). I'm guessing that this value should be fairly low (something on the order of 10), but we should measure. It almost certainly should not be 0.

DarkDimius: Based on my experience in ScalaBlitz, it was 4.

DarkDimius:
Leaving a message here so that I'll keep updated. I want to try to see how would auto-specialization compare with this manual approach.

DarkDimius:
>  I left it alone for now, but it's a bit weird since it means you get unboxed primitives with multiples, but not with the single-element special case.

I'm not sure how this library works, but it might be fine not to specialize single-element collections, as boxing of single element is not that bad.

djspiewak:
> I'm not sure how this library works, but it might be fine not to specialize single-element collections, as boxing of single element is not that bad.

Given that fs2 already (consciously) imposes significant overhead at chunk boundaries, that seems very likely to be the case.

core/src/main/scala/fs2/Chunk.scala
```diff
@@ -200,20 +288,105 @@ object Chunk {
       (0 until size).foldLeft(z)((z,i) => f(z, at(i)))
     def foldRight[B](z: B)(f: (Boolean,B) => B): B =
       ((size-1) to 0 by -1).foldLeft(z)((tl,hd) => f(at(hd), tl))
+
+    // here be dragons; modify with EXTREME care
+    override def map[B](f: Boolean => B): Chunk[B] = {
```
mpilquist: We still get boxing of the parameter to `f` here, right? Function1 isn't specialized on `Boolean` or `Byte` in its first parameter so we'll always box each input to the function.

mpilquist:
Looks good to me. I added 1 comment about boxing of `Boolean`s and `Byte`s when passing to `Function1` instances, but I don't think we can do anything about that. In scodec-bits, we used a custom trait that was specialized on `Byte` but I don't think that's a good idea here.

core/src/main/scala/fs2/Chunk.scala
djspiewak: Well I'll be…

Not only is that dumb and yes, you're right that we aren't having `Byte`/`Boolean` unboxed, but it seems that the return type specialization isn't working somehow!  Upon closer examination of the bytecode:

```
      38: invokevirtual #128                // Method values:()[Z
      41: iload_2
      42: baload
      43: invokestatic  #192                // Method scala/runtime/BoxesRunTime.boxToBoolean:(Z)Ljava/lang/Boolean;
      46: invokeinterface #197,  2          // InterfaceMethod scala/Function1.apply:(Ljava/lang/Object;)Ljava/lang/Object;
      51: invokestatic  #315                // Method scala/runtime/BoxesRunTime.unboxToLong:(Ljava/lang/Object;)J
      54: lastore
```

So that's a bit weird…

djspiewak: **Update**:  So it appears that the semantic for *compiling* specialization with multiple specialized parameters (e.g. `Function1`) is that if *one* type parameter cannot be specialized, then *neither* are specialized.  In other words, there is no special type for `Long => Byte`, as can be verified somewhat easily:

```scala
{ x: Long => 0.toByte }.getClass.getSuperclass      // => AbstractFunction1
```

Now, the respecialization *is* working in cases where both parameters are specialized.  For example, `Long => Long`:

```
      38: invokevirtual #128                // Method values:()[J
      41: iload_2
      42: laload
      43: invokeinterface #318,  3          // InterfaceMethod scala/Function1.apply$mcJJ$sp:(J)J
      48: lastore
```

Basically, I think specialization is not treating "erased" as a permutation for specialization, which means that there's a lot of boxing going on that doesn't need to be.

mpilquist: Cool, that's what what I expected. I really wish `Function1` was specialized for every primitive type in both arguments.

djspiewak: Hold on merging please.  I'm pretty sure I've contrived a case which will actually be broken by this due to the more limited specialization than what I expected (the test suite does not generate it).

yilinwei:
@djspiewak I will add something in the FAQs as to how to run the benchmarks. FYI though, it's just `sbt benchmark/benchmark/"jmh:run -i 10 -wi 10 -f1 -t1"` where `-i` is the iteration, `-wi` is the number of warmup iternations, `-f1` is the number of forks and `-t1` is the number of threads.

mpilquist:
@djspiewak Any update on whether there's a case that's broken by this PR? I'd like to release an M4 in the next day or two and I'd love for this PR to make the cut.

djspiewak:
@mpilquist Sorry for the delay.  Work stuff got in the way.  I'll see about addressing the problem.  I haven't written a test case yet, but I would be very *very* surprised if the case I'm thinking about *isn't* broken.

mpilquist:
No worries. If this gets in before M4, that's cool. Otherwise we can pick it up in the subsequent release. I'm hoping to do M4 late tonight.

djspiewak:
@mpilquist Should be set now.  I didn't touch the new `concat` stuff, but due to the way in which it is written, it should benefit considerably from the changes I made to `copyToArray`.  I left the broken extractors in place, partially as documentation, and partially as insurance in case `Function1` ever gets specialized for those cases (or, even better, if specialization gets adjusted to handle partial specialization more correctly).

mpilquist:
:+1:

PR #658, opened 2016-06-19
Add () to unsafe 0-arg methods

japgolly:
It's a common convention and IMO is quite helpful in helping methods standout as having effects.


mpilquist:
Thanks Dave. I'm generally neutral on this change -- I'm familiar with the convention and I used it for a number of years, but when I switched to using delimited effects via types like `Task`, I found the convention much less effective. With that said, I'm not really against this change. I'd like to hear from some other community members before merging though.

djspiewak:
It doesn't hurt anything to add the `()` at the declaration site.  Omitting them prevents users from following the convention if they so choose, while including them does nothing.

pchiusano:
Personally I really dislike this convention. The problem is that a pure `Function0` (which often get used for various purposes having nothing to do with side effects) is invoked with identical syntax, so using that as a convention to mean side effects is misleading. I'd prefer if the community adopted a different convention (the `unsafe*` prefix seems to be a common one).

But I guess it is not harmful, like @djspiewak says.

japgolly:
Cool, thanks for merging. Yeah I agree it's not perfect. Personally I find the false negatives (`Function0` overlap) faster and easier to mentally dismiss that the effort in keeping a mental eye out for the `unsafe` substring so hopefully this is an improvement overall, albeit tiny.

PR #662, opened 2016-06-24
Fixes memory leak in concurrent.join and signal.discrete

mpilquist:
Note that we have a similar memory leak as the one in signal in queue.dequeue

Review by @pchiusano & @pchlupacek

### DIFF
--- a/core/src/main/scala/fs2/PullDerived.scala
+++ b/core/src/main/scala/fs2/PullDerived.scala
@@ -4,9 +4,19 @@ trait PullDerived { self: fs2.Pull.type =>
 
   /**
    * Acquire a resource within a `Pull`. The cleanup action will be run at the end
-   * of the `.run` scope which executes the returned `Pull`.
+   * of the `.close` scope which executes the returned `Pull`. The acquired
+   * resource is returned as the result value of the pull.
    */
-  def acquire[F[_],R](r: F[R])(cleanup: R => F[Unit]): Pull[F,Nothing,(Pull[F,Nothing,Unit],R)] =
+  def acquire[F[_],R](r: F[R])(cleanup: R => F[Unit]): Pull[F,Nothing,R] =
+    acquireCancellable(r)(cleanup).map(_._2)
+
+  /**
+   * Like [[acquire]] but the result value is a tuple consisting of a cancellation
+   * pull and the acquired resource. Running the cancellation pull frees the resource.
+   * This allows the acquired resource to be released earlier than at the end of the
+   * containing pull scope.
+   */
+  def acquireCancellable[F[_],R](r: F[R])(cleanup: R => F[Unit]): Pull[F,Nothing,(Pull[F,Nothing,Unit],R)] =
     Stream.bracketWithToken(r)(Stream.emit, cleanup).open.flatMap { h => h.await1.flatMap {
       case (token, r) #: _ => Pull.pure((Pull.release(List(token)), r))
     }}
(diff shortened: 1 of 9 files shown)


core/src/main/scala/fs2/async/mutable/Signal.scala
```diff
@@ -70,18 +70,23 @@ object Signal {
...
+              F.map(F.modify(state) { case s@(a, l, listen) =>
                 if (l != last) s
                 else (a, l, listen :+ ref)
               }) { c =>
-                if (c.modified) F.get(ref)
-                else F.pure(c.now._1 -> c.now._2)
+                if (c.modified) {
+                  val cleanup = F.map(F.modify(state) {
+                    case s@(a, l, listen) => if (l != last) s else (a, l, listen.filterNot(_ == ref))
+                  })(_ => ())
```
pchlupacek: I am not sure if it won't be more descriptive put .eq() here instead of == so it is clear we compare the instance.

mpilquist:
OK, I fixed the leak in `queue.dequeue` too. Pending any comments from @pchiusano, I think this is ready to merge.

core/src/main/scala/fs2/PullDerived.scala
```diff
@@ -6,9 +6,9 @@ trait PullDerived { self: fs2.Pull.type =>
    * Acquire a resource within a `Pull`. The cleanup action will be run at the end
    * of the `.run` scope which executes the returned `Pull`.
```
pchiusano: Might want to add a comment about the meaning of the `Pull` that is returned as well.

io/src/main/scala/fs2/io/file/pulls.scala
```diff
@@ -66,13 +66,13 @@ object pulls {
     * The `Pull` closes the provided `java.nio.channels.FileChannel` when it is done.
     */
   def fromFileChannel[F[_]](channel: F[FileChannel])(implicit F: Effect[F]): Pull[F, Nothing, FileHandle[F]] =
-    Pull.acquire(F.map(channel)(FileHandle.fromFileChannel[F]))(_.close())
+    Pull.acquire(F.map(channel)(FileHandle.fromFileChannel[F]))(_.close()).map { case (_, ch) => ch }
```
pchiusano: kind of feels like we should keep `Pull.acquire` as is, and define it in terms of a new function, `Pull.acquireCancellable` (or some other name). Most uses of `Pull.acquire` are not going to use early termination.

core/src/main/scala/fs2/PullDerived.scala
mpilquist: Fixed in latest push.

io/src/main/scala/fs2/io/file/pulls.scala
mpilquist: Agreed, fixed in latest push.

core/src/main/scala/fs2/async/mutable/Signal.scala
pchiusano: I'm okay with `==`, personally, esp since it is on a `Long` but either way seems fine to me.

pchiusano:
LGTM, I'd say merge as long as Travis passes

core/src/main/scala/fs2/async/mutable/Signal.scala
mpilquist: I talked to Pavel about this in Gitter -- we can't use `eq` because `Ref` is an unbounded abstract type member. If we want to be able to use `eq`, we'd need to add a bound `type Ref[A] <: AnyRef`.

pchiusano: oh, whoops I thought we were talking about the `!= last`. Nevermind what I said then :)

mpilquist: Ha... so what do you think? Use `eq` and change `Async` to define `type Ref[A] <: AnyRef`? Or just leave it as is, using `==`?

pchiusano: Definitely just leave as is.

PR #663, opened 2016-06-24
Moved fs2.util.Task to fs2.Task and deleted the old, unfinished fs2.task object

mpilquist:


pchiusano:
Yeah, definitely 👍  to delete `task.scala`.

PR #665, opened 2016-06-25
Add Queue.imap, Topic.imap and Signal.map

OlivierBlanvillain:
This PR adds `imap`/`map` on `fs2.async.mutable.{Queue,Signal,Topic}`. The follow up would be to define the corresponding `Invariant` instances in `fs2-cats`/`fs2-scalaz`.

I'm not sure if I should test that with the invariant laws, bringing the ones from cats with an additional `% "test"` dep, or if you consider the type checker to be enough in that case...

https://github.com/functional-streams-for-scala/fs2-cats/pull/6

### DIFF
--- a/core/src/main/scala/fs2/async/immutable/Signal.scala
+++ b/core/src/main/scala/fs2/async/immutable/Signal.scala
@@ -2,13 +2,11 @@ package fs2.async.immutable
 
 import fs2.{pipe, Async, Stream}
 import fs2.util.Functor
-
 import fs2.Async
 import fs2.async.immutable
 
-
 /** A holder of a single value of type `A` that can be read in the effect `F`. */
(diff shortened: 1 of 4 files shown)


mpilquist:
👍 Can we add `imap` to `mutable.Signal` too?

OlivierBlanvillain:
Like this?

OlivierBlanvillain:
Ah didn't even notice they where two `Signal`s :smile:. Actually, the immutable version already has `ImmutableSignalSyntax` which give it a `map`, so I reverted my changes in that file.

(Note that because of inheritance it would be weird to have `map` define on `immutable.Signal`, as it would expose both `map` and `imap` on the `mutable` version with different return types.)

PR #669, opened 2016-06-30
Add failing test case for Task onFinish

ceedubs:
If a `Task` is preempted (such as with `.timed` with a shorter duration
than the `Task` takes to run), then its `onFinish` handler isn't
guaranteed to run (in fact it consistently fails this test for me).

This is an issue that @djspiewak pointed out to me as an existing issue
with scalaz's `Task`. I thought it was worth bringing to attention in
fs2, because with the more sophisticated resource/time handling in
fs2's Stream, it may be best to leave certain features (such as
`.timed`) off of fs2's `Task`, at least until they can be reliably
implemented.

### DIFF
--- a/core/src/test/scala/fs2/ResourceSafetySpec.scala
+++ b/core/src/test/scala/fs2/ResourceSafetySpec.scala
@@ -180,6 +180,14 @@ class ResourceSafetySpec extends Fs2Spec with org.scalatest.concurrent.Eventuall
       eventually { c.get shouldBe 0L }
     }
 
+    "preempted Task onFinish" in {
+      val c = new AtomicLong(0L)
+      val increment = Task.delay { c.incrementAndGet; () }
+      val slowStuff = Task.delay { Thread.sleep(100) } onFinish { _ => increment }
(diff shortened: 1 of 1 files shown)


djspiewak:
I think the better solution would be to make timed take some sort of "are you sure" evidence.  It's a necessary function (unsafe warts and all) for testing purposes, you had just better never see it in production code.

djspiewak:
onFinish, on the other hand, doesn't belong on Task.

mpilquist:
Removing onFinish sounds good to me.

What about renaming timed to unsafeTimed along with an explanation in docs why it has the unsafe prefix? Also, should timed take an optional cancellation callback?

djspiewak:
You actually would not be able to write down a type for a cancellation callback which plugs all of the resource holes in timed. Task just doesn't have the algebra for it, and trying to make it safer in this area would literally turn it into Stream.

A better approach I think would just be to name it unsafeTimed and keep the signature the way it is. Remove onFinish, since it's fundamentally broken and people can (and should) always use Stream instead.

pchiusano:
👍  on removing `onFinish`

PR #672, opened 2016-07-01
Added receiveOption, receive1Option, and receiveNonemptyOption syntax to Handle

mpilquist:

PR #673, opened 2016-07-01
Topic/task pruning

mpilquist:
Fixes #669 by removing `onFinish` completely. /cc @ceedubs @djspiewak


mpilquist:
Merge after #672

ceedubs:
👍 for removing `Task.onFinish`. I can't really speak to the other changes.

pchiusano:
@djspiewak any comments? I think this is exactly what you advocated in #669 but pipe up if not. Will merge later today or tomorrow if don't hear back.

djspiewak:
Looks like basically what I would want!

PR #674, opened 2016-07-01
Added infix syntax for fs2.util typeclasses

mpilquist:
I renamed `bind` to `flatMap` in order to pick up for-comp syntax. We could have `flatMap` as an alias for `bind` like scalaz does, but I think a single way to do things is better than having folks pick the name they prefer in each case.

Goal of this PR is not to compete with scalaz/cats but rather to cleanup the fs2 internals, where the use of prefix syntax gets pretty complex.

When I started this PR, I used simulacrum. I ended up removing the simulacrum dependency in order to define the syntax in AnyVal classes so that there's no runtime overhead.

Merge after #673

### DIFF
--- a/core/src/main/scala/fs2/util/Eq.scala
+++ b/core/src/main/scala/fs2/util/Eq.scala
@@ -1,24 +1,16 @@
 package fs2
 package util
 
+/** Evidence that `A` is the same type as `B` (a better version of `=:=`). */
 private[fs2] sealed trait Eq[A,B] {
-  def flip: Eq[B,A] = this.asInstanceOf[Eq[B,A]]
   def apply(a: A): B = Eq.subst[({type f[x] = x})#f, A, B](a)(this)
+  def flip: Eq[B,A] = this.asInstanceOf[Eq[B,A]]
 }
 
-object Eq {
+private[fs2] object Eq {
   private val _instance = new Eq[Unit,Unit] {}
   implicit def refl[A]: Eq[A,A] = _instance.asInstanceOf[Eq[A,A]]
 
   def subst[F[_],A,B](f: F[A])(implicit Eq: Eq[A,B]): F[B] =
     f.asInstanceOf[F[B]]
-
-  def substStream[F[_],A,B](s: Stream[F,A])(implicit Eq: Eq[A,B]): Stream[F,B] =
-    subst[({ type f[x] = Stream[F,x] })#f, A, B](s)
-
-  def substPull[F[_],W,A,B](p: Pull[F,W,A])(implicit Eq: Eq[A,B]): Pull[F,W,B] =
-    subst[({ type f[x] = Pull[F,W,x] })#f, A, B](p)
-
-  def substHandler[F[_],A,B](h: Throwable => Stream[F,A])(implicit Eq: Eq[A,B]): Throwable => Stream[F,B] =
-    subst[({ type f[x] = Throwable => Stream[F,x] })#f, A, B](h)
 }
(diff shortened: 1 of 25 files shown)


io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -139,7 +141,7 @@ protected[tcp] object Socket {
 
 
   def server[F[_]](
-    bind: InetSocketAddress
+    flatMap: InetSocketAddress
```
pchiusano: Ha. Pretty sure this should be named bind still. 😀

io/src/main/scala/fs2/io/udp/udp.scala
```diff
@@ -16,10 +17,10 @@ package object udp {
   case class Packet(remote: InetSocketAddress, bytes: Chunk[Byte])
 
   /**
-   * Provides a singleton stream of a UDP Socket that when run will bind to specified adress
-   * by `bind`.
+   * Provides a singleton stream of a UDP Socket that when run will flatMap to specified adress
+   * by `flatMap`.
```
pchiusano: Same deal here. Or maybe addr.

io/src/main/scala/fs2/io/udp/udp.scala
```diff
@@ -16,10 +17,10 @@ package object udp {
...
    *
-   * @param bind                 address to bind to; defaults to an ephemeral port on all interfaces
+   * @param flatMap                 address to flatMap to; defaults to an ephemeral port on all interfaces
```
pchiusano: Also fix 😀

core/src/main/scala/fs2/util/Eq.scala
```diff
@@ -1,24 +1,16 @@
...
 }
 
-object Eq {
+private[fs2] object Eq {
   private val _instance = new Eq[Unit,Unit] {}
   implicit def refl[A]: Eq[A,A] = _instance.asInstanceOf[Eq[A,A]]
 
   def subst[F[_],A,B](f: F[A])(implicit Eq: Eq[A,B]): F[B] =
     f.asInstanceOf[F[B]]
-
```
pchiusano: Where did these move to? I missed it in this diff.

mpilquist: Didn't move them -- these weren't called anymore. I didn't track down in history when they stopped getting used.

mpilquist:
Fixed the various udp/tcp references to `bind` :)

core/src/main/scala/fs2/util/syntax.scala
```diff
@@ -0,0 +1,26 @@
+package fs2.util
+
+/** Provides infix syntax for the typeclasses in the util package. */
```
pchiusano: We could make this private if we really wanted to make it for internal use only but I'm fine with keeping it public. It requires explicit import.

mpilquist: I think we should keep it public so that folks writing combinators that operate on arbitrary effect types can pick up the syntax too. Give them one less reason to write combinators fixed to `Task`.

pchiusano: I'm sold.

core/src/main/scala/fs2/util/Eq.scala
pchiusano: Ah, nice.

pchiusano:
LGTM. Github needs a 'merge when CI finishes successfully' button.

PR #675, opened 2016-07-05
Update scalaz-stream 0.8 to support scala 2.12.0-M5

fiadliel:
- Bump scalaz to 7.1.8 with a build for 2.12.0-M5
 - Add some `asInstanceOf` calls and type ascriptions to fix
   compilation errors.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -23,7 +23,7 @@ git.formattedShaVersion := {
 
 scalaVersion := "2.11.8"
 
-crossScalaVersions := Seq("2.10.6", "2.11.8", "2.12.0-M4")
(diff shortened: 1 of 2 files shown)


adriaanm:
Yep, looks similar to [the patch](https://github.com/scalaz/scalaz-stream/compare/series/0.8a...lrytz:fix-2.12-infer) we're using in the [community build](https://github.com/scala/community-builds/blob/2.12.x/common.conf#L118-L122), except that at first glance that one doesn't need a cast.

mpilquist:
@adriaanm Thanks, I updated this based on community build patch. All looks good now.

adriaanm:
Cool! Thanks for the quick turnaround!

PR #677, opened 2016-07-06
Expose runSet so that  typeclass instances that build on top of other…

sbuzzard:
…s can delegate to them (for instance remotely Response Async typeclass is a ). It's understood the reasoning for wanting to keep this contained within the implementation but given it is a method that must be implemented by those creating  instances, perhaps renaming to unsafeRunSet, as I did here, would suffice?  Any alternative suggestions are welcome.

### DIFF
--- a/core/src/main/scala/fs2/Async.scala
+++ b/core/src/main/scala/fs2/Async.scala
@@ -21,7 +21,7 @@ trait Async[F[_]] extends Effect[F] { self =>
    */
   def async[A](register: (Either[Throwable,A] => Unit) => F[Unit]): F[A] =
     flatMap(ref[A]) { ref =>
-    flatMap(register { e => ref.runSet(e) }) { _ => ref.get }}
+    flatMap(register { e => ref.unsafeRunSet(e) }) { _ => ref.get }}
 
   def parallelTraverse[A,B](s: Seq[A])(f: A => F[B]): F[Vector[B]] =
     flatMap(traverse(s)(f andThen start)) { tasks => traverse(tasks)(identity) }
@@ -120,7 +120,7 @@ object Async {
     def setPure(a: A): F[Unit] = set(F.pure(a))
 
     /** Actually run the effect of setting the ref. Has side effects. */
-    private[fs2] def runSet(a: Either[Throwable,A]): Unit
+    def unsafeRunSet(a: Either[Throwable,A]): Unit
 
   }
 
--- a/core/src/main/scala/fs2/StreamCore.scala
+++ b/core/src/main/scala/fs2/StreamCore.scala
@@ -112,7 +112,7 @@ sealed trait StreamCore[F[_],O] { self =>
     lazy val rootCleanup: Free[F,Either[Throwable,Unit]] = Free.suspend { resources.closeAll(noopWaiters) match {
       case Left(waiting) =>
         Free.eval(F.sequence(Vector.fill(waiting)(F.ref[Unit]))) flatMap { gates =>
-          resources.closeAll(gates.toStream.map(gate => () => gate.runSet(Right(())))) match {
+          resources.closeAll(gates.toStream.map(gate => () => gate.unsafeRunSet(Right(())))) match {
             case Left(_) => Free.eval(F.traverse(gates)(_.get)) flatMap { _ =>
               resources.closeAll(noopWaiters) match {
                 case Left(_) => println("likely FS2 bug - resources still being acquired after Resources.closeAll call")
--- a/core/src/main/scala/fs2/task.scala
+++ b/core/src/main/scala/fs2/task.scala
@@ -357,7 +357,7 @@ object Task extends Instances {
       Task.delay { S { t.unsafeRunAsync { r => actor ! Msg.Set(r) } }}
     def setFree(t: Free[Task,A]): Task[Unit] =
       set(t.run(F))
-    def runSet(e: Either[Throwable,A]): Unit =
+    def unsafeRunSet(e: Either[Throwable,A]): Unit =
       actor ! Msg.Set(e)
 
     private def getStamped(msg: MsgId): Task[(A,Long)] =


pchlupacek:
I am not totally sure that unsafeRunSet shall be part of an Async api. For any necessary effects we have already Async.Run. Can you for example share little bit more why this change  is necessary?

sbuzzard:
Sure - currently we have two Async typeclass instances in our code base.  Both of the types build on top of Task (they are in essense Kleisli's of A => Task).  In fs2 0.9.0M4 and earlier we were able to do this by using the Task.ref as the Ref type and then has part of the implementation of each of the functions of the typeclass instances, we'd delegate to the Async[Task] instance function.  With M5, Ref based functions were moved to a first class Async.Ref type which is parameterized on the Async instance's type, so we can't just use Async[Task]'s - we need to implement our own Ref, which is fine.  One of the abstract methods on Ref that we have to implement is runSet.  However, we can't delegate to the Async[Task]'s runSet because it is `private[fs2]` and I don't know of another way to implement this particular method with it's particular function signature without making runSet part of the API.  Here is an example: ```/** instance for Monadic `Async` Response`. */
  implicit def asyncResponseInstance(implicit S: Strategy, AT: Async[Task]): Async[Response] = new Async[Response] {
    def ref[A]: Response[Async.Ref[Response,A]] = Response { _ => Async.ref[Task, A](AT) map { new ResponseRef(_, this) } }
    def flatMap[A, B](a: Response[A])(f: A => Response[B]): Response[B] =
      Response { ctx => Task.suspend { a(ctx).flatMap(f andThen (_(ctx))) }}
    def pure[A](a: A): Response[A] = Response.now(a)
    override def delay[A](a: => A): Response[A] = Response.delay(a)
    def suspend[A](fa: => Response[A]) = Response.suspend(fa)
    def fail[A](err: Throwable): Response[A] = Response.fail(err)
    def attempt[A](a: Response[A]): Response[Either[Throwable, A]] = a.attempt map { _.toEither }
    override def toString = "Async[Response]"

    class ResponseRef[A](ref: Async.Ref[Task, A], protected val F: Async[Response]) extends Async.Ref[Response, A] {
      def access: Response[(A, Either[Throwable,A] => Response[Boolean])] = Response { _ =>
        ref.access.map { case (a,e2r) => (a, e2r andThen { t => Response { _ => t } }) }
      }
      def set(a: Response[A]): Response[Unit] = a edit { t => ref.set(t) }
      def setFree(a: Free[Response, A]): Response[Unit] = Response { ctx =>
        object translateFunc extends (Response ~> Task) { def apply[R](ra: Response[R]): Task[R] = ra(ctx) }
        ref.setFree(a.translate(translateFunc))
      }

      /* TODO: need to be able to ultimately call Task.runSet or the equivalent with the passed parameter */
      def runSet(tora: Either[Throwable, A]): Unit = ???

      override def get: Response[A] = Response { _ => ref.get }
      def cancellableGet: Response[(Response[A], Response[Unit])] = Response { ctx =>
        ref.cancellableGet map { case (ra, ru) => (Response { _ => ra }, Response { _ => ru }) }
      }
    }
  }```

pchlupacek:
Ok I see your point that this change is result of M5 changes. Let me think about that. I really think we shall keep effects to implementation and perhaps very limited usage of Async.Run when we couldn't go around it other way.

@pchiusano @mpilquist  any thoughts on this?

mpilquist:
Is it possible to get rid of `runSet`? The only place we use it is here: https://github.com/functional-streams-for-scala/fs2/blob/e9926d3e10cc5234d0e3014659f7706a2ab14807/core/src/main/scala/fs2/StreamCore.scala#L115

I'm kind of troubled by the existence of `runSet` and `Async.Run` -- seems like we are poking at the edges of something that should be modeled more directly.

sbuzzard:
I'm all for getting rid of it and definitely agree it's something that at least shouldn't be exposed outside of the implementation.  I did see it was only used in that one place (thus the temp hack for my runSet impl was to do an unsafe run async of set for the right side of the passed-in either and no-op for the left side, since there was no current usage there).

mpilquist:
@pchiusano Any ideas?

The more I think about this, the more uncomfortable I am with a type class that has a `private[fs2]` member. I'm fine if we put up a huge roadblock to calling a member (e.g., `unsafe` prefix, requiring some special implicit evidence that says "I know what I'm doing"), but I don't think we should be using access rules.

pchiusano:
No strong feelings, am fine with whatever. An unsafe prefix sounds
reasonable to me.

pchlupacek:
I think if we gonna do that, maybe we shall get rid of Async.Run and just add it to Async ? Don't see too much sense to keep it separately, specifically because AsyncRun is basically runSet...

pchlupacek:
perhaps just add it to Effect typeClass as unsafeRun?

mpilquist:
@pchlupacek I had similar thoughts yesterday -- i.e., getting rid of `Async.Run` and moving that code in to `Async`. Note that we can't define `unsafeRun` on `Effect` because it is impossible to implement on Scala.js. But we could define `unsafeRunAsync` on `Async` and be done with it. That's the way I'm leaning right now.

pchlupacek:
@mpilquist do we really need unsafeRun when we will be bound to Async/Effect? I mean won't the unsafeRunAsync will be all what we need? I like it on Effect, as this clearly marks your code does something with effects (i,e, suspend, delay, async runs)...

mpilquist:
No, I don't think we need `unsafeRun` -- that's what I was trying to say above. :)

I'm fine with putting `unsafeRunAsync` on `Effect`.

pchlupacek:
The point of Keeping is with Effect is, that you don't need to have Strategy in scope...

pchlupacek:
ok super lets go this way :-)

mpilquist:
See #679.

PR #679, opened 2016-07-08
Async type class cleanup

mpilquist:
This PR includes a number of changes related to the `Async` and `Effect` type classes.

1. The `Effect` type class has a new member, `unsafeRunAsync`, allowing (potentially asynchronous) evaluation of the effect.
2. The `Async.Run` type class has been deleted in favor of using `Effect#unsafeRunAsync`.
3. The `setRun` method has been removed from `Async`, replaced by calling `unsafeRunAsync(F.set(e.fold(F.fail,F.pure)))(_ => ())`.
4. The `setFree` method has been removed from `Async`, replaced with nothing (it was not used).
5. The `Async` type class has been moved to `fs2.util` package and syntax has been added to `fs2.util.syntax`.
6. The `Async.Future` type has been moved to `fs2.ScopedFuture`, making its dependence on `Stream` and `Pull` clearer, and leaving the `Async` type class free of any dependency on stream primitives. Note this includes moving the `read` method from the `Ref` trait to the `ScopedFuture` companion.
7. The `ScopedFuture#force` method was renamed to `pull` to align with the existing `stream` method. I could see calling these `forceAsPull`/`forceAsStream` instead.

For some background discussion of these changes, see #677.

### DIFF
--- a/core/src/main/scala/fs2/time/time.scala
+++ b/core/src/main/scala/fs2/time/time.scala
@@ -30,7 +30,7 @@ package object time {
    * @param S           Strategy to run the stream
    * @param scheduler   Scheduler used to schedule tasks
    */
-  def awakeEvery[F[_]](d: FiniteDuration)(implicit F: Async[F], FR: Async.Run[F], S: Strategy, scheduler: Scheduler): Stream[F,FiniteDuration] = {
+  def awakeEvery[F[_]](d: FiniteDuration)(implicit F: Async[F], S: Strategy, scheduler: Scheduler): Stream[F,FiniteDuration] = {
     def metronomeAndSignal: F[(()=>Unit,async.mutable.Signal[F,FiniteDuration])] = {
       async.signalOf[F, FiniteDuration](FiniteDuration(0, NANOSECONDS)).flatMap { signal =>
         val lock = new java.util.concurrent.Semaphore(1)
@@ -39,7 +39,7 @@ package object time {
           val cancel = scheduler.scheduleAtFixedRate(d, d) {
             val d = FiniteDuration(System.nanoTime, NANOSECONDS) - t0
             if (lock.tryAcquire)
-              FR.unsafeRunAsyncEffects(signal.set(d) >> F.delay(lock.release))(_ => ())
+              F.unsafeRunAsync(signal.set(d) >> F.delay(lock.release))(_ => ())
           }
           (cancel, signal)
         }
@@ -78,6 +78,6 @@ package object time {
    * A single-element `Stream` that waits for the duration `d` before emitting its value. This uses the implicit
    * `Scheduler` to signal duration and avoid blocking on thread. After the signal, the execution continues with `S` strategy.
    */
-  def sleep[F[_]: Async : Async.Run](d: FiniteDuration)(implicit S: Strategy, scheduler: Scheduler): Stream[F, Nothing] =
+  def sleep[F[_]: Async](d: FiniteDuration)(implicit S: Strategy, scheduler: Scheduler): Stream[F, Nothing] =
     awakeEvery(d).take(1).drain
 }
--- a/core/src/main/scala/fs2/time/time.scala
+++ b/core/src/main/scala/fs2/time/time.scala
@@ -2,6 +2,7 @@ package fs2
 
 import scala.concurrent.duration._
 
+import fs2.util.Async
 import fs2.util.syntax._
 
 package object time {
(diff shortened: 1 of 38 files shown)


core/src/main/scala/fs2/time/time.scala
```diff
@@ -30,7 +31,7 @@ package object time {
    * @param S           Strategy to run the stream
    * @param scheduler   Scheduler used to schedule tasks
    */
-  def awakeEvery[F[_]](d: FiniteDuration)(implicit F: Async[F], FR: Async.Run[F], S: Strategy, scheduler: Scheduler): Stream[F,FiniteDuration] = {
+  def awakeEvery[F[_]](d: FiniteDuration)(implicit F: Async[F], S: Strategy, scheduler: Scheduler): Stream[F,FiniteDuration] = {
```
pchlupacek: so we still need Strategy here?

io/src/main/scala/fs2/io/file/FileHandle.scala
```diff
@@ -93,7 +93,7 @@ object FileHandle {
     *
     * Uses a `java.nio.Channels.CompletionHandler` to handle callbacks from IO operations.
     */
-  private[fs2] def fromAsynchronousFileChannel[F[_]](chan: AsynchronousFileChannel)(implicit F: Async[F], FR: Async.Run[F]): FileHandle[F] = {
+  private[fs2] def fromAsynchronousFileChannel[F[_]](chan: AsynchronousFileChannel)(implicit F: Async[F]): FileHandle[F] = {
     new FileHandle[F] {
```
pchlupacek: I think instead of async we shall be ok just with Effect Here ?

io/src/main/scala/fs2/io/file/file.scala
```diff
@@ -1,22 +1,22 @@
...
-import fs2.util.Effect
+import fs2.util.{Async,Effect}
 
 package object file {
 
   /**
     * Provides a handler for NIO methods which require a `java.nio.channels.CompletionHandler` instance.
     */
-  private[fs2] def asyncCompletionHandler[F[_], O](f: CompletionHandler[O, Null] => F[Unit])(implicit F: Async[F], FR: Async.Run[F]): F[O] = {
+  private[fs2] def asyncCompletionHandler[F[_], O](f: CompletionHandler[O, Null] => F[Unit])(implicit F: Async[F]): F[O] = {
```
pchlupacek: I think Effect shall be sufficient TC here ?

io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -203,9 +201,9 @@ protected[tcp] object Socket {
       ch.read(buff, timeoutMs, TimeUnit.MILLISECONDS, (), new CompletionHandler[Integer, Unit] {
         def completed(result: Integer, attachment: Unit): Unit =  {
           val took = System.currentTimeMillis() - started
-          FR.unsafeRunAsyncEffects(F.delay(cb(Right((result, took)))))(_ => ())
+          F.unsafeRunAsync(F.delay(cb(Right((result, took)))))(_ => ())
```
pchlupacek: I was just tihinking that this runs the task in AG's context. Not 100% sure if that's desired. One can have same Executor for `F` and AG, so then that's ok. What do you think?

mpilquist: The alternative is wrapping this line in a call to a strategy, right?

core/src/main/scala/fs2/time/time.scala
mpilquist: Yeah, `scheduleAtFixedRate` requires a strategy for callback invocation.

io/src/main/scala/fs2/io/tcp/Socket.scala
pchlupacek: Was sort of thinking something like 
```scala
F.unsafeRunAsync(F.start(F.delay(cb(Right((result, took)))))(_ => ()))
```

core/src/main/scala/fs2/time/time.scala
pchlupacek: I think we can do F.start() to run the callback ? Essentially I am nervous about having one strategy bound in Async, and other passed as the implicit. Which one of them will be used?

mpilquist: OK sure -- I'll remove the Strategy param from Scheduler and use `start` here instead. I'll also add calls to `start` in each of the uses of completion callbacks in io package.

PR #680, opened 2016-07-09
Introduced Attempt[A] as a type alias for Either[Throwable,A] and...

### LINKED ISSUE #678: Consider replacing Either[Throwable,A] with Try[A]
Paul and I had discussed this a number of months ago but never came to a resolution. Currently, we use `Either[Throwable,A]` in lots of places, along with constructs like: `try Right(a) catch { case t: Throwable => Left(t) }`. By switching to `Try`, we'd only catch non-fatal exceptions, which may or may not be considered desirable. I lean towards using `Try` (I want a `NonFatal` exception to trigger application shutdown) but I'm eager to hear other opinions.

mpilquist:
…changed all catches to avoid catching VirtualMachineError.

Fixes #678.

### DIFF
--- a/core/src/main/scala/fs2/task.scala
+++ b/core/src/main/scala/fs2/task.scala
@@ -1,7 +1,7 @@
 package fs2
 
 import fs2.internal.{Actor,Future,LinkedMap}
-import fs2.util.{Async,Effect}
+import fs2.util.{Async,Attempt,Effect,NonFatal}
 import java.util.concurrent.atomic.{AtomicBoolean,AtomicReference}
 
 import scala.concurrent.duration._
@@ -17,7 +17,7 @@ import scala.concurrent.duration._
  *
  * Task is also exception-safe. Any exceptions raised during processing may
  * be accessed via the `attempt` method, which converts a `Task[A]` to a
- * `Task[Either[Throwable,A]]`.
+ * `Task[Attempt[A]]`.
  *
  * Unlike the `scala.concurrent.Future` type introduced in scala 2.10,
  * `map` and `flatMap` do NOT spawn new tasks and do not require an implicit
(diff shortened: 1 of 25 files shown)


core/src/test/scala/fs2/TestUtil.scala
```diff
@@ -32,7 +34,7 @@ trait TestUtil {
     catch {
       case e: InterruptedException => throw e
       case e: TimeoutException => throw e
-      case e: Throwable => ()
+      case NonFatal(e) => ()
```
pchlupacek: Hm, I am just thinking loudly. In strategies and executors shouldn't we just do System.exit(-1) on all fatal exceptions after stderr them? As I am aware only exception that JVM may terminate is OOM (that we perhaps shouldn't catch at all ever), all the rest will lead us to unpredictable state. Rather would kill JVM then having it run in un-predicatable state.

core/src/main/scala/fs2/task.scala
```diff
@@ -202,14 +202,14 @@ object Task extends Instances {
...
-    Try(a.get) match {
+    Attempt(a.get) match {
       case Left(e) => Future.now(Left(e))
       case Right(f) => f
   }))
 
   /** Create a `Future` that will evaluate `a` using the given `Strategy`. */
   def apply[A](a: => A)(implicit S: Strategy): Task[A] =
-    async { cb => S(cb(Try(a))) }
+    async { cb => S(cb(Attempt(a))) }
```
pchlupacek: rather than having default Attempt.apply, shouldn't we have Attempt.success and Attempt.failure constructors?

mpilquist: I'm fine with adding `success` and `failure` but I don't think these replace `Attempt.apply` -- rather, they complement.

core/src/test/scala/fs2/TestUtil.scala
mpilquist: Perhaps. Akka has such behavior but lets you disable it via configuration. We might want to do the same if we decide to shutdown JVM.

mpilquist:
@pchlupacek Just addressed both of your comments. Let me know if you have any others.

pchlupacek:
@mpilquist excellent, thanks

PR #681, opened 2016-07-09
Changed various callback invocations to async start effect before running

mpilquist:
@pchlupacek Let me know if this addresses all of your comments on Async PR.


pchlupacek:
@mpilquist excellent, thanks

PR #686, opened 2016-07-14
Removed a performance optimization in Free#step.

mpilquist:
The performance optimzation eagerly applied `Bind(Pure(x), f)`
while stepping. IIRC, this optimization gave ~20% performance
improvement in some benchmarks. Unfortunately, it is unprincipled. More
specifically, the optimization assumed that `f` is pure. However,
`Free.suspend` (and `Stream.suspend`) are implemented as
`pure(()).flatMap(_ => s)`, so we either need to introduce `suspend` as
a new primitive in `Free` (and `Stream` and maybe other places) or we
need to avoid the assumption that `f` is pure.

Note that the optimization allowed us to use `Either[Throwable,?]` in
the evaluation of pure streams in a stack safe way. Removal of the
optimization requires the effect type to provide its own stack safety.
For now, I've changed pure streams to use `Task` again, though I'm
considering introducing an `fs2.IO` type that provides an `Effect[IO]`
instance along with trampolined synchronous evaluation.

### DIFF
--- a/core/src/main/scala/fs2/StreamDerived.scala
+++ b/core/src/main/scala/fs2/StreamDerived.scala
@@ -233,8 +233,8 @@ trait StreamDerived extends PipeDerived { self: fs2.Stream.type =>
 
   implicit class StreamPureOps[+A](s: Stream[Pure,A]) {
     def toList: List[A] =
-      s.covary[Attempt].runFold(List.empty[A])((b, a) => a :: b).fold(t => throw t, identity).reverse
-    def toVector: Vector[A] = s.covary[Attempt].runLog.fold(t => throw t, identity)
+      s.covary[Task].runFold(List.empty[A])((b, a) => a :: b).unsafeRun.reverse
+    def toVector: Vector[A] = s.covary[Task].runLog.unsafeRun
   }
 
   implicit def covaryPure[F[_],A](s: Stream[Pure,A]): Stream[F,A] = s.covary[F]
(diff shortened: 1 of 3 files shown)


mpilquist:
I should have explained the failing test...

`Stream.bracket` uses `StreamCore.acquire` which does something roughly like this:

```scala
StreamCore.suspend {
  val t = new Token()
  ...
}
```

The `suspend` call gets expanded to `pure(()).flatMap { _ => val t = new Token(); ... }` which is then eagerly interpreted when converting the `Free[Task,A]` to a `Task[A]`.

Another way to fix this would be changing `suspend,bracket,acquire` to take an `Effect[F]` constraint and then mandating that all functions are pure. For example, we'd need to implement `suspend` as `def suspend[F[_], A](s: => Stream[F, A])(implicit F: Effect[F]): Stream[F, A] = Stream.eval(F.delay(s))`. This approach feels more principled to me but I know folks aren't probably keen on adding the `Effect` constraint to those methods.

pchiusano:
I think we should merge this, since it is at least correct. :) But it would be nice to know how much performance we've lost.

Not sure I like introducing `IO`. There are a lot of things already called `IO`, seems like just another source of confusion for not much benefit.

mpilquist:
I'm not keen on the name `IO` but at the very least, we'll need a stack safe data type that supports `Catchable` in order to support `.toList` and `.toVector` on Scala.js (where we can't use a `CountDownLatch` or other blocking in order to run a potentially async `Task` synchronously).

mpilquist:
Hm, ignore my comments on `IO` and `Scala.js`. I just realized we can use `Task` and the new `unsafeRunSync` method in order to implement `toList`/`toVector`.

PR #689, opened 2016-07-15
Topic/signal modify2

pchlupacek:


core/src/main/scala/fs2/async/mutable/Signal.scala
```diff
@@ -26,6 +26,11 @@ trait Signal[F[_], A] extends immutable.Signal[F, A] { self =>
   def modify(f: A => A): F[Async.Change[A]]
 
   /**
+    *  like `modify` but allows to extract `B` from `A` and return it together with Change
+    */
+  def modify2[B](f: A => (A,B)):F[(Async.Change[A], B)]
```
pchiusano: perhaps provide default implementation of `modify` in terms of `modify2` + `map`?

pchlupacek:
@pchiusano not sure if I am getting your idea. We do not want to have signal of (A,B) as result. Or likely I am missing something?

core/src/main/scala/fs2/async/mutable/Signal.scala
pchiusano: Actually, @pchlupacek nevermind, this is the `Signal` trait so you don't have an `Async[F]` available. In the implementation of `Signal`, you could implement `modify` in terms of `modify2` if you aren't already doing so.

pchlupacek:
@pchiusano yes. got it. The reason why I kept modify and modify2 separate was primarily because performance. In fact whole reason for modify2 is the performance, as you can define modify2 in terms of modify and map. But perhaps you are right that vice-versa it makes more sense to implement modify in terms of modify2, as this is just boxing of one tuple more.  Do you think I shall do it ?

pchiusano:
I think it is okay to add. 

If we have a lot of these derived functions on Signal if given an `Async[F]`, we should maybe come up with a better solution, so that people don't have a bunch of boilerplate if they are implementing their own signal type. Although maybe that is not a big use case since 99.9% of people will just use the existing signal type.

pchlupacek:
ok will do.

pchlupacek:
@pchiusano , @mpilquist done.

pchiusano:
Seems okay to me. Assuming Travis passes and @mpilquist has no objections 👍  to merge.

PR #691, opened 2016-07-17
add stateless effectful unfold 

### LINKED ISSUE #690: stateless effectful unfold?
`StreamDerived.unfoldEval` provides for unfolding with passed state, but effectful computations can have their own state, which allows for a simpler "unfold", equivalent to `unfoldEval` with unit state.

```scala
def unroll[F[_], T](f: F[Option[T]]): Process[F,T] = {
  def go: Process[F, T] =
    eval(f).flatMap {
      case Some(a) => emit(a) ++ go
      case None    => empty
    }
  suspend(go)
}
```

If this seems reasonable (I'm using it with doobie where my `f` is "read the next thing from the resultset, if any") I can do a PR, or feel free to copy/paste.

tpolecat:
This adds a method to unfold an effectful computation `F[Option[A]]` into `Stream[F,A]`.

resolves #690

### DIFF
--- a/core/src/main/scala/fs2/StreamDerived.scala
+++ b/core/src/main/scala/fs2/StreamDerived.scala
@@ -161,6 +161,16 @@ trait StreamDerived extends PipeDerived { self: fs2.Stream.type =>
     suspend(go(s0))
   }
 
+  /** Like [[unfoldEval]], but takes an effecful computation that manages its state internally. */
+  def unfoldEval0[F[_], A](f: F[Option[A]]): Stream[F,A] = {
+    def go: Stream[F, A] =
+      eval(f).flatMap {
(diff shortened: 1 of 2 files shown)


fiadliel:
I'd usually write that kind of behaviour in this way (`unNoneTerminate` is the standard pipe combinator for terminating `Option[_]` streams):

```scala
Stream.repeatEval(f).through(pipe.unNoneTerminate)
```

But then I wouldn't write it that way in `StreamDerived`, which probably shouldn't depend on functions from `pipe`.

I'm not strongly against this extra method, but just asking if the addition pulls its weight, given available combinators...?

tpolecat:
Ok, I didn't know about that pipe combinator. That works for me.

Issue #500, opened 2015-12-03
Scala.JS build with FS2?


fiadliel:
What do you think of having a (possibly experimental) Scala.JS build of FS2?

With Node optimization, rough ball-park speed reduction seems around 20x compared with running on the JVM :) But it does work (mostly).

Due to the lack of async support in scalacheck, it seems somewhat difficult to do any async testing, but I got the pure tests working using a non-task, non-async, Trampoline-based monad.

`Task` itself works in a browser/JS context, as long as you avoid `run`/`attemptRun`/etc., for obvious reasons.

pchlupacek:
@fiadliel can you perhaps try performance of processing with and w/o fs2 on node.js before we decide? I just tried that about like 6 months ago with 0.7  and performance was so poor that we had decided to trash idea. Can you do tests with fs2._ now? 

The idea is to compare js/js benchmark instead of js/jvm benchmark :-)

fiadliel:
Sure! I'll have a look at which tests either are, or can be made to be equivalent.

Having some performance tests that have approximately the same effects/results across 0.8 and 0.9 would be useful for comparisons even without considering the JS aspect.

japgolly:
On the topic of (Scala.)JS benchmarking, I have a library [scalajs-benchmark](https://github.com/japgolly/scalajs-benchmark) ([live demo](https://japgolly.github.io/scalajs-benchmark/)) which you might find useful.

InTheNow:
A big :+1: if this could work on scala.js! Main reason would be that it would be safe for many libraries to use FS2 knowing that it won't break the stack re multi-platform.

Out of curiosity, how does fs2 work on scala.js without multi-threading, etc? Or perhaps rather, what won't work?

pchlupacek:
@InTheNow hence we mostly use asynchronous code for concurrency, this will work in scala.js out of the box, but, obviously not concurrently, as js does not have multi-threading model.

japgolly:
If you're interested you might be able to cook something up with WebWorkers. That'll give you real threads in JS albeit with limited access.

fiadliel:
Sorry, Christmas/New Year was a somewhat extended downtime this year. But I'll get back to this over the next week.

@InTheNow you can't do anything asynchronously, and then wait for results - it has to be a callback. So `fs2.util.Task.run` will not run (cannot compile), but `.runAsync` is fine.

[...]

PR #692, opened 2016-07-26
Add support for Scala.js

mpilquist:
See #500 for details on the approach taken in this PR. This is mostly an exercise in moving code around.

Note that ScalaTest doesn't currently support using `forAll` in conjunction with tests that return futures. And we can't call `unsafeRun` on `Task` in Scala.js because there's no thread blocking allowed. As a result, most unit tests run on the JVM only. ScalaTest can't support this combination until ScalaCheck supports async props, which is tracked in https://github.com/rickynils/scalacheck/issues/214.


mpilquist:
Merging since there are no objections (and I confirmed @pchiusano has none).

PR #695, opened 2016-07-31
Benchmark only after build result DO NOT MERGE

yilinwei:
Only run the benchmark after #672 - will change it to run after a merge only but want to see if it will return the result before running the benchmarks.


yilinwei:
That's really annoying because even though Travis says the build has finished it doesn't seem to have updated the git hook until `after_success` has finished.

PR #696, opened 2016-08-01
Move InputStream/OutputStream from fs2.io.file to fs2.io, add async variants

fiadliel:
InputStream/OutputStream support is not exclusively for files, so it should be
in the fs2.io package. Because these calls are blocking, adds a variant that
uses the backing `fs2.Async` instance for asynchronous behavior.

@pchlupacek Comments welcome.

### DIFF
--- a/io/src/main/scala/fs2/io/package.scala
+++ b/io/src/main/scala/fs2/io/package.scala
@@ -5,28 +5,7 @@ import fs2.util.syntax._
 import java.io.{InputStream, OutputStream}
 
 package object io {
-
-  private[io] def readBytesFromInputStream[F[_]](is: InputStream, buf: Array[Byte])(implicit F: Effect[F]): F[Option[Chunk[Byte]]] =
-    F.delay(is.read(buf)).map { numBytes =>
-      if (numBytes < 0) None
-      else if (numBytes == 0) Some(Chunk.empty)
-      else Some(Chunk.bytes(buf, 0, numBytes))
-    }
-
-  private[io] def readInputStreamGeneric[F[_]](fis: F[InputStream], chunkSize: Int, f: (InputStream, Array[Byte]) => F[Option[Chunk[Byte]]], closeAfterUse: Boolean = true)(implicit F: Effect[F]): Stream[F, Byte] = {
-    val buf = new Array[Byte](chunkSize)
-
-    def useIs(is: InputStream) =
-      Stream.eval(f(is, buf))
-        .repeat
-        .through(pipe.unNoneTerminate)
-        .flatMap(Stream.chunk)
-
-    if (closeAfterUse)
-      Stream.bracket(fis)(useIs, is => F.delay(is.close()))
-    else
-      Stream.eval(fis).flatMap(useIs)
-  }
+  import JavaInputOutputStream._
 
   /**
     * Reads all bytes from the specified `InputStream` with a buffer size of `chunkSize`.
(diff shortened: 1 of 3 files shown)


io/src/main/scala/fs2/io/package.scala
```diff
@@ -0,0 +1,104 @@
+package fs2
+
+import fs2.util.{Async, Effect}
+import fs2.util.syntax._
+import java.io.{InputStream, OutputStream}
+
+package object io {
+
+  private[io] def readBytesFromInputStream[F[_]](is: InputStream, buf: Array[Byte])(implicit F: Effect[F]): F[Option[Chunk[Byte]]] =
```
pchlupacek: just merely style guide. How about introducing JavaInputOutputStream.scala, put all implementation there and on package object just have the `api` functions?

io/src/main/scala/fs2/io/package.scala
```diff
@@ -0,0 +1,104 @@
...
+    * Blocks the current thread.
+    */
+  def readInputStream[F[_]](fis: F[InputStream], chunkSize: Int, closeAfterUse: Boolean = true)(implicit F: Effect[F]): Stream[F, Byte] =
+    readInputStreamGeneric(fis, chunkSize, readBytesFromInputStream[F], closeAfterUse)
+
+  /**
+    * Reads all bytes from the specified `InputStream` with a buffer size of `chunkSize`.
+    * Set `closeAfterUse` to false if the `InputStream` should not be closed after use.
+    *
+    * This will block a thread in the current `Async` instance, so the size of any associated
```
pchlupacek: I think blocking threads in Async is something what we shall surely avoid. I think we shall guard Async threads at any possible place. I like the idea of `Async` in the name, same as I would the other variant name `Sync` or even `Blocking` to really stress the nature of the implementation. @mpilquist WDYT?

pchlupacek: Also I was thinking, are there really scenarios where you want the IS to outlive the resulting process? I mean is this absolutelly necessary? I am just wondering if this can't be source of just hard to track bugs and resource leaks.

fiadliel: Keeping the IS open is purely to support stdin, though I'm sure there could be other streams that outlive a particular `fs2.Stream`.

pchlupacek: hm just wonder how the java solves it. I bet it have the AutoClose trait on StdIn/Out that essentially does nothing for these?

fiadliel: But stdin can be essentially any InputStream. I guess because people just don't invoke the try-with-resources pattern on it?

mpilquist:
@fiadliel This ready for merge?

pchlupacek:
@mpilquist @fiadliel ok to merge from my side. Was just curious if we cannot get rid of that closeAfterUse flag hence I still believe AutoClosable interface from java shall give us enough, but I am ok to go with current signature, if that sounds too much experimental.

fiadliel:
I don't think we should close std{in|out}, and it seems reasonable to reuse this code for that purpose; hence the defaults.

Also, I am absolutely horrible at naming things, so feel free to suggest alternative names (the gap between description and verbosity is one of the most difficult problems in computing, as they say).

Other than that, good to merge :)

mpilquist:
Sounds good, merging!

PR #697, opened 2016-08-01
Improve usability of API

### LINKED ISSUE #694: fs2._ and fs2.Stream._ conflicts on Pull
Is this really intended ? Effectively if you are using Pulls, one cannot have these two underscore imports in single file. I would opt for removing the Stream.Pull if possible, or make it private to Stream object

mpilquist:
This PR cleans up various parts of the API to make it easier to use.

I attempted to keep the commits relatively independent so it may be easier to review the commits individually. Highlights of changes:

 - Removed `Streams` and `Pulls` traits. While these served as nice descriptions of the essential operations that define a `Stream` and `Pull` respectively, they also brought a lot of baggage. The encoding sort of forced the notion of the `StreamDerived` and `StreamOps` traits, which resulted in implementations being spread across multiple files and annoying design decisions like whether a method should be on companion along with syntax forwarder or just first class method on the class. Finally, `Streams` had accidental dependencies on `Stream` and `Pull` so even if we liked that separation, we would have needed to fix it.

 - Removed `StreamDerived`, `StreamOps`, `PullDerived`, etc. Without the `Streams` / `Pulls` traits, these don't really need to exist anymore. Defining the derived operations in separate files is beneficial from a design standpoint but it makes it really hard for people to read the source. This change coupled with the previous allows us to make source browsing much easier. We also get some minor performance improvements in the form of `AnyVal` implicit classes for syntax and faster dispatch due to use of final classes

 - Removed `fs2.Step` in favor of `Tuple2`. I don't think this type was carrying its weight API-wise. There are a lot of new types and concepts to learn when first working with FS2. Adding `Step` as a nominative pair and the `#:` extractor adds 2 more things to learn with little benefit. As a nice side effect of this change, I implemented `receive` such that it takes a `Function2[...]`, which removes some boilerplate.

 - Renamed `fs2.Stream.Handle` to `fs2.Handle`. I could go either way on this one but ended up making the change. Folks typically import `Stream.Handle` anyway so this just avoids that annoyance.

Fixes #694.


benchmark/src/main/scala/fs2/benchmark/StreamBenchmark.scala
```diff
@@ -37,8 +37,8 @@ class StreamBenchmark extends BenchmarkUtils {
     (Stream.chunk(Chunk.seq(0 to 256000)).pure).repeatPull { s =>
       for {
         s2 <- Pull.awaitN(N)(s)
-        _  <- Pull.output(s2.head.head)
-      } yield s2.tail
+        _  <- Pull.output(s2._1.head)
+      } yield s2._2
```
pchlupacek: I am just wondering if we can implement `filter` on pull and instead of s2 have (_,s) on the left side of for? WDYT?

core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
```diff
@@ -11,14 +11,14 @@ object ResourceTrackerSanityTest extends App {
 }
 
 object RepeatPullSanityTest extends App {
-  def id[A]: Pipe[Pure, A, A] = _ repeatPull Pull.receive1 { case h #: t => Pull.output1(h) as t }
+  def id[A]: Pipe[Pure, A, A] = _ repeatPull Pull.receive1 { case (h, t) => Pull.output1(h) as t }
```
pchlupacek: just for curiosity. If we make receive1 to take Function2, then I think that case is redundant?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1,101 +1,381 @@
...
+
+  /** Alias for `self through [[pipe.split]]`. */
+  def split(f: O => Boolean): Stream[F,Vector[O]] = self through pipe.split(f)
+
+  def step: Pull[F,Nothing,(Chunk[O],Handle[F,O])] =
     Pull.evalScope(get.step).flatMap {
       case None => Pull.done
       case Some(Left(err)) => Pull.fail(err)
-      case Some(Right(s)) => Pull.pure(s.copy(tail = new Handle(List(), Stream.mk(s.tail))))
+      case Some(Right((hd, tl))) => Pull.pure((hd, new Handle(Nil, Stream.mk(tl))))
```
pchlupacek: I am just thinking if naming the handle a `tl` is correct. i think `h` was pretty ok. I am leaning towards (c,h) where c means chunk, and in case of receive1 simple (o,h). WDYT?

pchlupacek:
just added few comments. Don't have a strong preference on any of them, just if we doing this massive PR, sort of was thinking to align the style. 
other than that I am ok with this going in.

mpilquist:
Thanks @pchlupacek -- I addressed all of your comments. I'll merge today if no objection from @pchiusano.

pchiusano:
:+1: 

I am not sure what I was thinking with Step. For pure lists having a right assoc pattern is nice but that doesn't really carry over in effectful case.

PR #704, opened 2016-08-09
Rename nonempty to nonEmpty

rossabaker:
This is consistent with stdlib, Cats, and Scalaz.


mpilquist:
Thanks! Happy to get this in before RC1 :)

PR #707, opened 2016-08-10
Removed onComplete from Stream

mpilquist:
This avoids any temptation of using `onComplete` for resource cleanup. Note the slight behavior change in the `hash` module -- old implementation erroneously masked errors in the stream.


pchiusano:
No objections.

PR #709, opened 2016-08-14
Added fs2.async.circularBuffer

### LINKED ISSUE #703: Add `async.circularBuffer`
Help wanted on this. This existed in 0.8, I believe @runarorama might have contributed it. I think a reasonable 0.9 signature for it would be:

```Scala
def circularBuffer[F[_]:Async,A](capacity: Int): F[Queue[A]]
```

in `async` package. Behaves like a regular FIFO queue, but if an enqueue would make the queue size greater than capacity, do a dequeue first. Seems like it would be easy to implement in terms of a regular unbounded `Queue` plus a `Semaphore`.

mpilquist:
Fixes #703


pchiusano:
Lgtm. Nice simple implementation.

PR #711, opened 2016-08-17
Added distinctConsecutive and various related pipes

mpilquist:
Both `distinctConsecutive` and `distinctConsecutiveBy` use natural equality instead of relying on type class based equality. We can add type class variants in fs2-cats and fs2-scalaz by delegating to `filterBy2`. I don't love the name `filterBy2` but I don't have any other suggestions I like better -- e.g. `filterSliding2`, `filterWithLast` -- so I stuck with the name from 0.8.

The implementation of `filterBy2` is optimized for chunks where most elements are distinct. Specifically, it attempts to preserve the chunk structure if possible and only if not does it build a new filtered chunk.

### DIFF
--- a/core/shared/src/main/scala/fs2/pipe.scala
+++ b/core/shared/src/main/scala/fs2/pipe.scala
@@ -80,7 +80,7 @@ object pipe {
    * using natural equality for comparison.
    */
   def distinctConsecutive[F[_],I]: Stream[F,I] => Stream[F,I] =
-    filterWithLast((i1, i2) => i1 != i2)
+    filterWithPrevious((i1, i2) => i1 != i2)
 
   /**
    * Emits only elements that are distinct from their immediate predecessors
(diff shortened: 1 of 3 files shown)


pchlupacek:
+1, you may consider filterWithPrevious, filterWithPrev

core/shared/src/main/scala/fs2/pipe.scala
```diff
@@ -75,6 +75,20 @@ object pipe {
...
+   */
+  def distinctConsecutive[F[_],I]: Stream[F,I] => Stream[F,I] =
+    filterBy2((i1, i2) => i1 != i2)
+
+  /**
+   * Emits only elements that are distinct from their immediate predecessors
+   * according to `f`, using natural equality for comparison.
+   */
+  def distinctConsecutiveBy[F[_],I,I2](f: I => I2): Stream[F,I] => Stream[F,I] =
+    filterBy2((i1, i2) => f(i1) != f(i2))
```
pchiusano: Not sure if this matters, but this will call `f` twice on the same element, once when it is being compared to its previous element, and once when being compared with the next element. Might be worth documenting this.

You could also implement `distinctConsecutiveBy` via a `map` (create a pair of `(i, f(i))`), then the `filterBy2` operating on the `f(i)` portion of the pair, then a `map(_._1)` to get back the `i`. Don't think I would do this though. Let users do that before calling the function.

pchiusano:
👍 , just would clarify docs on `distinctConsecutiveBy` like I suggested

core/shared/src/main/scala/fs2/pipe.scala
mpilquist: @pchiusano It's worse than that even -- `filterBy2` calls its predicate once *or twice* for each element depending on whether the chunk being processed needs any filtering. We could drop the optimization but it seems bad to rebuild each chunk pessimistically. I'll update the docs on `distinctConsecutiveBy` to make it clear that `f` should be cheap.

mpilquist:
@pchlupacek `filterWithPrevious` is perfect, thanks

pchiusano:
👍

PR #713, opened 2016-08-20
Fix chunkReadBuffer reopening channel on underfill

rossabaker:
The `completionHandler` helper referred to the call-by-name src instead
of the acquired resource, resulting in a leak on buffer underfill.

Fixes http4s/http4s#673.


rossabaker:
It looks like Travis has a troubled history on this branch.  It, uh, worked on my machine.

https://travis-ci.org/functional-streams-for-scala/fs2/branches

mpilquist:
Thanks -- are you planning on a 0.8a PR too?

rossabaker:
Oops, I falsely assumed that 0.8a was maintained by merging 0.8.  Yes, PR on the way.

PR #718, opened 2016-08-20
Introduced NonEmptyChunk

### LINKED ISSUE #708: Decide how empty chunks are handled
Currently, empty chunks are elided by the `StreamCore` interpreter. As a result, I believe that an empty chunk will never be passed to `await`, which makes the various non-empty variants of `await` and `receive` redundant.

Assuming I'm right in this belief, I think we should do one of the following:
1. Remove the non-empty variants of `await` and `receive`
2. Remove the elision of empty chunks from `StreamCore`

Alternatively, can anyone come up with an example where empty chunks would get passed to an `h.await` in the current implementation?

I'm going to release RC1 anyway - we can remove the non-empty variants in RC2 or final.

mpilquist:
This PR introduces the `NonEmptyChunk` type and changes `step` and `stepAsync` to return `NonEmptyChunk` instead of `Chunk`. Removed the non-empty variants of `await` and `receive` since this PR makes them redundant.

Note that folks can't pattern match on a subtype of `Chunk` now but that's okay -- instead, they can call `c.toBytes` to unify the code that works on the underlying array.

Fixes #708 

@pchiusano Sorry for the delay on this one. Busy week. ;)

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -3,7 +3,7 @@ package fs2
 import scala.reflect.ClassTag
 
 /**
- * Chunk represents a strict, in-memory sequence of `A` values.
+ * A strict, in-memory sequence of `A` values.
  */
 trait Chunk[+A] { self =>
   def size: Int
@@ -21,14 +21,15 @@ trait Chunk[+A] { self =>
     val index = iterator.indexWhere(p)
     if (index < 0) None else Some(index)
   }
-  def isEmpty = size == 0
+  def isEmpty: Boolean = size == 0
+  def nonEmpty: Boolean = size != 0
   def toArray[B >: A: ClassTag]: Array[B] = {
     val arr = new Array[B](size)
     copyToArray(arr)
     arr
   }
-  def toList = foldRight(Nil: List[A])(_ :: _)
-  def toVector = foldLeft(Vector.empty[A])(_ :+ _)
+  def toList: List[A] = foldRight(Nil: List[A])(_ :: _)
+  def toVector: Vector[A] = foldLeft(Vector.empty[A])(_ :+ _)
   def collect[B](pf: PartialFunction[A,B]): Chunk[B] = {
     val buf = new collection.mutable.ArrayBuffer[B](size)
     iterator.collect(pf).copyToBuffer(buf)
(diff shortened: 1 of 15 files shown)


pchiusano:
Uh oh - https://travis-ci.org/functional-streams-for-scala/fs2/jobs/153831080#L923

Isn't `toBytes` going to involve an array copy? That does not seem ideal. Perhaps just add overrides for `toBytes`, `toDoubles`, etc on `NonEmptyChunk` which forward to the underlying `Chunk`. That way if the underlying chunk is already a `Chunk.Bytes`, it can just be returned directly.

I just gave this a quick look I might have missed how this is already taken care of in some other way.

mpilquist:
Whoops yeah, I need to fix `toBytes`, toDoubles`, etc. That's an oversight.

Any idea on the async resource allocation failure unit test? The only things changed in the interpreter were:
 - `StreamCore.pushEmit` elides the append call if pushing an empty chunk
 - `StreamCore.stepTrace` returns `None` if it encounters an empty chunk in a `Segment.Emit(c)`

mpilquist:
I've reproduced the failing test case once locally. I have a hunch as to the cause and how to fix but I'd like to be able to reproduce more easily before making the change. I've run the failing test about 1.5 million times at this point and I've only encountered 1 failure.

mpilquist:
Good news / bad news... 

The good news is that I can easily reproduce the failure in "asynchronous resource allocation (1)" by doing the following: in Fs2Spec, change ScalaCheck's `minSuccessful` from 25 to 200,000 and change `workers` from 1 to 4.

The bad news is that I can reproduce this on series/0.9 and as far back as 0.9.0-M1 (I tried reproducing on M1, M3, M6, and HEAD and can reproduce on all).

mpilquist:
OK good news. I realize why those tests are failing, and it is NOT due to a bug in FS2 proper but rather a bug in the tests. The tests erroneously assumed that the decrement of the atomic long would always be visible to the thread that's running the test by the time `runLog.unsafeRun` returns. This is not a safe assumption though. We corrected this same bug in some of the other resource safety allocation tests but we hadn't corrected it in 1, 2a/2b, 3.

This should be good for merge at this point (when Travis passes) unless @pchiusano or others have additional comments.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -21,14 +21,15 @@ trait Chunk[+A] { self =>
     val index = iterator.indexWhere(p)
     if (index < 0) None else Some(index)
   }
-  def isEmpty = size == 0
+  def isEmpty: Boolean = size == 0
+  def nonEmpty: Boolean = size != 0
```
rossabaker: I was missing this from `ByteVector` and wanted to add this to ease my own migration.  Trivial, but it brightened my day.

rossabaker:
Others are better qualified to comment on the internals, but I like the approach.  I've written code to special case empty ByteVectors in 0.8, and it's nice that this is baked into 0.9's chunks.

pchlupacek:
+1 given travis passes

PR #719, opened 2016-08-21
Delay bounded enqueue until there is space

### LINKED ISSUE #685: 0.7.3/0.8 Queue Can Get Indefinitely "Stuck" When Double-Overfull
Non-minimized test case (courtesy of a coworker):

```scala
    implicit val S = Strategy.DefaultStrategy

    val queue = async.boundedQueue[Int](2)

   val f = S{ //consumption thread
      val pp =  queue.dequeue.wye(queue.size.discrete)(wye.either).to(sink.lift[Task,Int\/Int](qs =>  Task.now{log("queue or size " + qs)} ))
      pp.run.run
    }

    log("start")

    val t = Nondeterminism[Task].gatherUnordered{
      (1 to 20).map(i => for {
        _ <- Task(log("processing "+i))
        _ <- queue.enqueueOne(i)
        _ <- Task.delay(log("affter enqueue "+i))
      } yield ())
    }

    t.run

    log("run end")

    queue.close.run
    f() //close thread
```

There are several distractingly weird points here, but the salient bit is that the bounded queue has a bound of 2, but there are 20 *simultaneous* enqueues being run.  Naturally, 18 of those enqueues will block pretty quickly, ending up in the `unAcked` vector.  Critically, all of the enqueue actions are one-and-done: once they unblock, they will not result in subsequent enqueues.

The snippet above deterministically hangs forever.  The reason can be traced to the following bit of code:

```scala
        if (unAcked.size > 0 && bound > 0 && queued.size <= bound) {
          val ackCount = bound - queued.size min unAcked.size
          unAcked.take(ackCount).foreach(cb => S(cb(\/-(()))))
          unAcked = unAcked.drop(ackCount)
        }
```

Translated into english…

- If the queue is bounded *and* the queue is at or below its bound
- Compute the ceiling between us and the queue bound, no larger than the number of unacknowledged enqueues
- Acknowledge *that* number of enqueues and reduce `unAcked` correspondingly

The first weird thing here is that the conditional is `queued.size <= bound`, meaning that we're going to have a `unAcked.take(0)` hit, which is harmless but weird.  The more problematic thing though is what this does over time.

Remember that our scenario has 18 unacked enqueues (give or take, since the dequeue process *is* running) and a queue bound of 2.  This means that once we fill up and `unAcked.isEmpty == false`, we will have the following series of events:

1. **Precondition** `unAcked.size == 18 && queued.size == 20` (remember that data is put into `queued` *before* acknowledgement is checked)
2. Dequeue one element, decrementing `queued.size`
3. Repeat step 2 until `queued.size == 2`
4. Do the whole randomly `take(0)` no-op
5. Dequeue one more element (now: `queued.size == 1`)
6. Compute `ackCount = 1` (uh oh…)
7. Acknowledged *one* of our `unAcked` continuations (now: `unAcked.size == 17`)
8. Dequeue one more element (now: `queued.size == 0`)
9. Compute `ackCount = 2`
10. Acknowledge *two* of our `unAcked` continuations (now: `unAcked.size == 15`)
11. Done

Now, `queued.size == 0`, which means that we can't dequeue any more and will not acknowledge any more continuations, but we still have 15 unacknowledged continuations!  So we hang.

And just in case you were wondering, no, 0.9 is not susceptible to this bug.  The *reason* it isn't susceptible is because it gates addition to the queue *before* it puts data in the vector, rather than *after*.  Our whole problem here is that we've rolled a very ad hoc semaphore with unsound semantics.

This isn't the most common of cases, since you need to be running a lot of "one and done" enqueues on a bounded queue, and the problem becomes quadratically infrequent as the bound increases (though it will manifest for any finite bound and sufficiently large number of enqueues), but it's still kind of important (we found it in production).

Given that most of the community is still on 0.7.3a, I'm in favor of fixing this in both series/0.7 and series/0.8, pushing new releases in each.  At least for Verizon specifically, our only alternative would be to maintain a private fork of scalaz-stream in which this bug is fixed, which isn't great, especially since other people are probably hitting this.

djspiewak:
This should fix #685.  Basically it defers the actual enqueue until there is room in the queue, whereas before it enqueued immediately even if over the bound and delayed the ACK.  This is very similar to the series/0.9 design.

### DIFF
--- a/src/main/scala/scalaz/stream/async/mutable/Queue.scala
+++ b/src/main/scala/scalaz/stream/async/mutable/Queue.scala
@@ -158,6 +158,8 @@ private[stream] object Queue {
   def mk[A](bound: Int, recover: Boolean,
             beforeEnqueue: (Seq[A], Vector[A]) => Vector[A])(implicit S: Strategy): Queue[A] = {
     sealed trait M
+
+    case class Do(f: () => Unit) extends M
     case class Enqueue(a: Seq[A], cb: Throwable \/ Unit => Unit) extends M
     case class Dequeue(ref: ConsumerRef, limit: Int, cb: Throwable \/ Seq[A] => Unit) extends M
     case class Fail(cause: Cause, cb: Throwable \/ Unit => Unit) extends M
@@ -246,8 +248,8 @@ private[stream] object Queue {
 
         queued = remainder
         signalSize(queued.size)
-        if (unAcked.size > 0 && bound > 0 && queued.size <= bound) {
-          val ackCount = bound - queued.size min unAcked.size
+        if (unAcked.size > 0 && bound > 0 && queued.size < bound) {
+          val ackCount = (bound - queued.size) min unAcked.size
           unAcked.take(ackCount).foreach(cb => S(cb(\/-(()))))
           unAcked = unAcked.drop(ackCount)
         }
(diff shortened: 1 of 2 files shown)


src/main/scala/scalaz/stream/async/mutable/Queue.scala
```diff
@@ -158,6 +158,8 @@ private[stream] object Queue {
   def mk[A](bound: Int, recover: Boolean,
             beforeEnqueue: (Seq[A], Vector[A]) => Vector[A])(implicit S: Strategy): Queue[A] = {
     sealed trait M
+
+    case class Do(f: () => Unit) extends M
```
djspiewak: Who loves actors?  IT'S A TRICK QUESTION DON'T WE ALL

mpilquist:
👍

mpilquist:
FYI, I cherry picked this to series/0.8a.

PR #722, opened 2016-08-24
Add dequeueAvailable and supporting code.

fiadliel:
Adds `dequeueAvailable` which dequeues all available values from the queue.

Alters the queue state from containing `Async.Ref[F,A]` to `Either[Async.Ref[F,A], Async.Ref[F,Vector[A]]`, depending on whether the waiter is looking for single, or all queued values.

### DIFF
--- a/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
+++ b/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
@@ -139,9 +139,7 @@ object Queue {
         def dequeue1: F[A] = cancellableDequeue1.flatMap { _._1 }
 
         def cancellableDequeue1: F[(F[A],F[Unit])] =
-          cancellableDequeueBatch1(1).flatMap {
-            case (chunks, cancel) => chunks.map(c => (F.pure(c(0)), cancel))
-          }
+          cancellableDequeueBatch1(1).map { case (deq,cancel) => (deq.map(_(0)),cancel) }
 
         def dequeueBatch1(batchSize: Int): F[Chunk[A]] =
           cancellableDequeueBatch1(batchSize).flatMap { _._1 }
@@ -183,7 +181,7 @@ object Queue {
           permits.tryDecrement.flatMap { b => if (b) q.offer1(a) else F.pure(false) }
         def dequeue1: F[A] = cancellableDequeue1.flatMap { _._1 }
         override def cancellableDequeue1: F[(F[A], F[Unit])] =
-          cancellableDequeueBatch1(1).flatMap { case (chunks, cancel) => chunks.map(c => (F.pure(c(0)), cancel)) }
+          cancellableDequeueBatch1(1).map { case (deq,cancel) => (deq.map(_(0)),cancel) }
         override def dequeueBatch1(batchSize: Int): F[Chunk[A]] = cancellableDequeueBatch1(batchSize).flatMap { _._1 }
         def cancellableDequeueBatch1(batchSize: Int): F[(F[Chunk[A]],F[Unit])] =
           q.cancellableDequeueBatch1(batchSize).map { case (deq,cancel) => (deq.flatMap(a => permits.incrementBy(a.size).as(a)), cancel) }
@@ -204,13 +202,12 @@ object Queue {
           enqueue1(a).as(true)
         def dequeue1: F[A] = cancellableDequeue1.flatMap { _._1 }
         def dequeueBatch1(batchSize: Int): F[Chunk[A]] = cancellableDequeueBatch1(batchSize).flatMap { _._1 }
-        def cancellableDequeue1: F[(F[A], F[Unit])] = cancellableDequeueBatch1(1).flatMap { case (chunks, cancel) => chunks.map(c => (F.pure(c(0)), cancel))}
+        def cancellableDequeue1: F[(F[A], F[Unit])] = cancellableDequeueBatch1(1).map { case (deq,cancel) => (deq.map(_(0)),cancel) }
         def cancellableDequeueBatch1(batchSize: Int): F[(F[Chunk[A]],F[Unit])] =
           q.cancellableDequeueBatch1(batchSize).map { case (deq,cancel) => (deq.flatMap(a => permits.incrementBy(a.size).as(a)), cancel) }
         def size = q.size
         def full: immutable.Signal[F, Boolean] = q.size.map(_ >= maxSize)
         def available: immutable.Signal[F, Int] = q.size.map(maxSize - _)
-
       }
     }}
 
--- a/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
+++ b/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
@@ -43,8 +43,22 @@ trait Queue[F[_], A] { self =>
   /** Like `dequeue1` but provides a way to cancel the dequeue. */
   def cancellableDequeue1: F[(F[A], F[Unit])]
 
+  /** Dequeue `A`s from this queue. Completes once any values are ready. */
+  def dequeueBatch1(batchSize: Int): F[Chunk[A]]
+
+  /** Like `dequeueAvailable1` but provides a way to cancel the dequeue. */
+  def cancellableDequeueBatch1(batchSize: Int): F[(F[Chunk[A]], F[Unit])]
+
   /** Repeatedly call `dequeue1` forever. */
   def dequeue: Stream[F, A] = Stream.bracket(cancellableDequeue1)(d => Stream.eval(d._1), d => d._2).repeat
+
+  /** Call `dequeueBatch1` once with a provided bound on the elements dequeued. */
+  def dequeueBatch: Pipe[F, Int, A] = _.flatMap { batchSize =>
+    Stream.bracket(cancellableDequeueBatch1(batchSize))(d => Stream.eval(d._1).flatMap(Stream.chunk), d => d._2)
+  }
+
+  /** Call `dequeueBatch1` repeatedly with a bound of `Int.MaxValue` */
+  def dequeueAvailable: Stream[F, A] = Stream.constant(Int.MaxValue).through(dequeueBatch)
 
   /**
    * The time-varying size of this `Queue`. This signal refreshes
@@ -83,6 +97,10 @@ trait Queue[F[_], A] { self =>
       def dequeue1: F[B] = self.dequeue1.map(f)
       def cancellableDequeue1: F[(F[B],F[Unit])] =
         self.cancellableDequeue1.map(bu => bu._1.map(f) -> bu._2)
+      def dequeueBatch1(batchSize: Int): F[Chunk[B]] =
+        self.dequeueBatch1(batchSize).map(_.map(f))
+      def cancellableDequeueBatch1(batchSize: Int): F[(F[Chunk[B]],F[Unit])] =
+        self.cancellableDequeueBatch1(batchSize).map(bu => bu._1.map(_.map(f)) -> bu._2)
     }
 }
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -40,12 +40,22 @@ trait Queue[F[_], A] { self =>
   /** Dequeue one `A` from this queue. Completes once one is ready. */
   def dequeue1: F[A]
 
+  /** Dequeue all available `A`s from this queue. Completes once any values are ready. */
+  def dequeueAvailable1: F[Vector[A]]
```
pchiusano: I'd call this just `dequeueAvailable`.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -40,12 +40,22 @@ trait Queue[F[_], A] { self =>
...
   def cancellableDequeue1: F[(F[A], F[Unit])]
 
+  /** Like `dequeueAvailable1` but provides a way to cancel the dequeue. */
+  def cancellableDequeueAvailable1: F[(F[Vector[A]], F[Unit])]
+
   /** Repeatedly call `dequeue1` forever. */
   def dequeue: Stream[F, A] = Stream.bracket(cancellableDequeue1)(d => Stream.eval(d._1), d => d._2).repeat
 
+  /** Repeatedly call `dequeueAvailable1` forever. */
+  def dequeueAvailable: Stream[F, A] =
```
pchiusano: Ah, I see what you're doing. :)

Is there any reason to keep both `dequeue` and `dequeueAvailable`? Or should we just change the implementation of `dequeue` to use `dequeueAvailable`?

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -94,7 +107,7 @@ object Queue {
       * @param queue    Queue, expressed as vector for fast cons/uncons from head/tail
       * @param deq      A list of waiting dequeuers, added to when queue is empty
       */
-    case class State(queue: Vector[A], deq: Vector[Async.Ref[F,A]])
+    case class State(queue: Vector[A], deq: Vector[Either[Async.Ref[F,A], Async.Ref[F,Vector[A]]]])
```
pchiusano: Is there a reason to use `Either` vs just having everything use `Async.Ref[F,Vector[A]]`?

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
fiadliel: Yes... but we would dequeue all elements and then re-enqueue all but the head. I'd be fine with that, except I feel like we could change the ordering if there was a pending enqueue in between. WDYT?

fiadliel: In that case, a pushback addition to the queue would work - but is that better than the current code?

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
fiadliel: As mentioned above, you end up having to take all the queue contents, and push back whatever you don't need. I can rewrite to do this, but not sure it's really better.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
pchiusano: Yeah, actually, nevermind. Wasn't really thinking too carefully. :)

I didn't mean get rid of `dequeue1`, I was thinking just have `dequeue` (which returns a `Stream`) use `dequeueAvailable`. I was thinking that since it is returning the entire stream of dequeued elements, there shouldn't need to be any pushback... you just `dequeueAvailable` to obtain a `Vector[A]`, then `Stream.emits` that. BUT then I remembered stuff like `q.dequeue.take(4)` -- that could consume more elements from the queue than are needed if it were implemented in terms of `dequeueAvailable`. So having both versions seems fine.

I did not do a super thorough review but looks 👍  to me. @pchlupacek / @mpilquist and/or @djspiewak may want to weigh in before merging.

pchiusano:
@djspiewak if you get a chance pls weigh in with a 👍  assuming it addresses your original request :)

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
pchiusano: Oh, I just meant - you could encode the `Left` case with a singleton `Vector`. I think it is okay as is though.

mpilquist:
Could we have both a `dequeue` stream based on `dequeue1` and another based on `dequeueAvailable`? For situations where user knows they aren't going to terminate the stream early, `dequeueAvailable` should result in significantly better performance.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
mpilquist: IIRC, Vector isn't specialized for singletons so you'd have a full 32 element array allocated for each `dequeue1`. Perhaps not that meaningful considering everything else involved in queues but maybe worth considering.

mpilquist:
:+1: from me. Note that the new unit test hung on Travis (or at least it got killed after running for 10 minutes).

fiadliel:
Oh, this PR caused the hang? Please don't merge until I find out why.

mpilquist:
Yeah - https://travis-ci.org/functional-streams-for-scala/fs2/jobs/154868149#L1014

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
pchlupacek: What will happen if we will have callback  like Async[Ref[Chunk[A]] ??? Queue may be then Vector[Chunk[A]] instead of just Vector[A], or perhaps just Chunk[A]. Not sure about chunk characteristics in these cases.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
pchlupacek: I think we can have signature here def dequeueAvailable1: F[A] that will just use proper chunking

pchlupacek: If someone really needs strict chunks, then one can do `dequeueAvailable.chunks`

pchlupacek:
I think we have to replace Vector with Chunks at least in signatures if not in implementation, that will make life easier, and I think chunk is our sequence of choice in fs2 now. My feeling is that this will simplify the cod too, as we maybe only will need one dequeue method in implementation with simple parameter whether to return the head or the full queue. Other than that I am ok with it.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
fiadliel: Updating with the signature `F[NonEmptyChunk[A]]`, guaranteeing that at least one value is returned. We can't hide the chunkiness at the individual effect level - if this is undesired at the API level, we can make it private - but I think it could be useful for imperative-style use.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
pchlupacek: I just thought a little bit about `dequeueAvailable` and I think I would like it much more as 
```scala
def dequeueAvailable:Pipe[F, Int,A] = ???
```
the reason for this is the you may introduce very nice flow control based on some signal of `int` vaues that indicates how much client want to proceed at any given time. 

Also this is much general signature, and perhaps could allow us same code for dequeue and dequeueAvailable, where dequeue is really just special case for dequeueAvailable. If the performance will be an issue, we can always specialise dequeue, otherwise dequeue is for me just ```constant(1).through(dequeueAvailable)```

@pchiusano @fiadliel @mpilquist  any thoughts?

djspiewak: @pchlupacek My concern with the signature you're proposing is that there's no way to dequeue *all* of the elements, unless you feed the actual queue bound to the `Pipe`.  That seems ugly to me.  Like I said, this kind of abstraction is most useful in situations where your element algebra has "natural batching", where it doesn't matter how many elements are being collapsed and has no impact on fairness or flow control.  Or another way of looking at this would be as a form of variable-width catch-up on backpressure, since processing of algebrae with natural batching *speeds up* on a per-element basis as the number of elements increases.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
djspiewak: The API is by far most useful if it exposes the chunking.  That's sort of the whole point.  :-)  I would be in favor of using `Chunk` rather than `Vector` though.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
djspiewak: `Vector` is indeed not specialized for singletons, or emptiness, or really anything that it would make *sense* for it to be specialized on.  `Chunk` would be a better choice.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
pchlupacek: Whats wrong with `constant(Int.MaxValue).through(dequeueAvailable)` to get all available elements? Also if that seems ugly, we can introduce `dequeueAllAvailable` as an alias for this. Also I have no intention to introduce any dedicated queue bound to that pipe. This is simply to add the parameter to dequeue operation that will say how much element you are interested in instead on relying only on defaults (1) or (All). The possibility of abstract this for both cases is just free gain to me not the original intention. 

The idea of flow-control is something you may or may not use in this case. Imagine situation where you would like to feed the results to some network I/O laters. Maybe some of the I/O pipes may be slower then others, while still we want to maximize the performance to take always all the elements we can process in given time frame in single chunk. In that case the max elements may really depend on two factors, time and speed of the outbound I/O. Wouldn't it be nice in that case to dynamically bound the elements available to be dequeued?

mpilquist:
Any further thoughts/updates on this?

fiadliel:
Updated: there is now `dequeueBatch` (a Pipe), and `dequeueAvailable` which repeatedly feeds `Int.MaxValue` into `dequeueBatch`.

Using `Chunk[A]` as the API, this is convenience for dequeuing single elements, as `Chunk.singleton` doesn't incur the cost of a whole `Vector`.

Also, the `Either` representation of the state wasn't actually needed, so that is removed.

mpilquist:
@fiadliel ChannelSpec.sanityTest hung on both the 2.11 and 2.12 builds (tests were killed after 10 minutes)

fiadliel:
@mpilquist Thanks 😞 I was testing the QueueSpec tests in particular, will figure out what's going on there.

fiadliel:
Updated, and tests pass.

fiadliel:
One remaining thing that I could change: the signature _could_ be `NonEmptyChunk` instead of `Chunk`, since for these implementations, at least one chunk is guaranteed.

pchlupacek:
if NonEmpty chunk is in place already lets go for it imho

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -43,9 +43,23 @@ trait Queue[F[_], A] { self =>
   /** Like `dequeue1` but provides a way to cancel the dequeue. */
   def cancellableDequeue1: F[(F[A], F[Unit])]
 
+  /** Dequeue `A`s from this queue. Completes once any values are ready. */
```
pchlupacek: that means we would take at least 1 chunk, but up to batchSize?

fiadliel: Right. We can always rechunk downstream.

pchlupacek: cool. perhaps I would clarify this in comment.

fiadliel: Done.

fiadliel:
Updated to use `NonEmptyChunk`.

mpilquist:
👍 Looks great

pchlupacek:
lgtm

PR #723, opened 2016-08-27
Improved performance and readability of GADT encoding

mpilquist:
Changed the way "bound" cases are encoded in both `Free` and
`StreamCore`. Rather than passing only the bound cases via a trait with
a single universally quantified method, all cases of the fold are passed
in as a trait. This results in significantly less runtime allocations,
as each call to `fold` was allocating a bound instance anyway but now we
can avoid some anonymous function allocations and some Left/Right
allocations.

In a fairly non-scientific test, these changes resulted in a 7%
performance increase for me (aka, I ran the test suite a few times and
measured how long the last run took).

This encoding is more verbose due to having to specify the type
signatures of the various abstract methods representing each case, but
the benefits to both performance and readability outweigh the
inconvenience of the verbosity.


pchlupacek:
@mpilquist great work, +1

pchiusano:
+1

PR #724, opened 2016-08-27
Minor performance improvements

mpilquist:
Changes:
 - Using final modifier on case classes so they do not hold a reference
   to enclosing object
 - Using final modified on concrete methods in StreamCore
 - Avoiding point free style in Free and StreamCore to avoid allocations
 - Replaced some allocations with casts in StreamCore GADTs
 - Fixed a bug where StreamCore.NT#apply was getting called from
   attemptEval
 - Removed debug tracing
 - Upgraded to latest JMH

These changes result in a *small* performance bump.


mpilquist:
I've been incrementally adding ScalaDoc to this branch until we merge it (at which point I'll continue to add ScalaDoc on a new branch). I started by documenting the util package but I really had a hard time justifying the signatures of `traverse` and `sequence`. After some thought, I decided it would be best to add `Applicative` and `Traverse` type classes. So please review that addition too. :)

mpilquist:
Merging since there are no objections. Feel free to chime in later if there are and we can revert if necessary.

PR #727, opened 2016-08-30
Implemented ToInputStream

### LINKED ISSUE #725: Introduce toInputStream
Some Java API require InputStream as source of data w/o eventually providing any other way how to feed data. Eventhough we have already simple ways to read from InputStream, this intends to create a way to convert Stream of Bytes to inputStream consumable with JavaAPI.

The basic idea is : 

```scala

val source : Stream[F,Byte] = ???

// Input Stream is closed when the source finsihes. Source is killed when InputStream terminates
val is:Stream[F,InputStream] = source.toInputStream

```

pchlupacek:
As subject says. This shall close #725.


io/src/main/scala/fs2/io/ToInputStream.scala
```diff
@@ -0,0 +1,259 @@
...
+import async.mutable
+import Async.Change
+
+import scala.concurrent.SyncVar
+
+
+/**
+  * Conversion of Stream[F,Byte] to java.io.InputStream
+  */
+object ToInputStream {
```
mpilquist: Make `private[fs2]`

io/src/main/scala/fs2/io/ToInputStream.scala
```diff
@@ -0,0 +1,259 @@
...
+                override def read(b: Array[Byte], off: Int, len: Int): Int = impl.readIs(b,off,len,queue,dnState)
+                def read(): Int = impl.readIs1(queue,dnState)
+              }
+            }.onFinalize(impl.close(upState,dnState))
+          }}}
+  }
+
+
+
+  object impl {
```
mpilquist: Can you move the contents of this object in to the `ToInputStream` object?

io/src/main/scala/fs2/io/ToInputStream.scala
```diff
@@ -0,0 +1,259 @@
...
+    /** state of the upstream, we only indicate whether upstream is done and if it failed **/
+    case class UpStreamState(done:Boolean,err:Option[Throwable])
```
mpilquist: Add final modifier to these case classes

io/src/main/scala/fs2/io/ToInputStream.scala
```diff
@@ -0,0 +1,259 @@
...
+      }
+
+
+      F.unsafeRunAsync(
+        source.chunks
+        .evalMap(ch => queue.enqueue1(Right(ch.toBytes)))
+        .interruptWhen(dnState.discrete.map(_.isDone).filter(identity))
+        .run
+      ){
+        case Left(err) => markUpstreamDone(Some(err))
```
mpilquist: `r => markUpstreamDone(r.left.toOption)`

io/src/main/scala/fs2/io/ToInputStream.scala
mpilquist: I think we should put this code inside the `JavaInputOutputStream` object instead of its own object. To hide the implementation methods, just nest them inside the main combinator. E.g.,

```scala
private[io] object JavaInputOutputStream {
  ...
  def toInputStream[F[_]](implicit F: Async[F]): Pipe[F,Byte,InputStream] = {
    // Put everything here, including the contents of the impl package
  }
}
```

io/src/main/scala/fs2/io/ToInputStream.scala
```diff
@@ -0,0 +1,259 @@
...
+      def markUpstreamDone(result:Option[Throwable]):Unit = {
+        F.unsafeRunAsync(
+          F.flatMap(upState.set(UpStreamState(done = true, err = result))) { _ =>
+            queue.enqueue1(Left(result))
+          }
+        )(_ => ())
+      }
+
+
+      F.unsafeRunAsync(
```
mpilquist: It's probably a little better to use `F.start` here instead of `unsafeRunAsync`.

io/src/main/scala/fs2/io/ToInputStream.scala
pchlupacek: Was my original thought too. I am not sure if it is good idea then as there is no reuse/relation to other methods there, and I doubt ever will be so, I was kinda of preventing the clutter in unrealted code. But is ok for me to do that if you think it makes sense

io/src/main/scala/fs2/io/ToInputStream.scala
pchlupacek: ok

io/src/main/scala/fs2/io/ToInputStream.scala
pchlupacek: thanks

io/src/main/scala/fs2/io/ToInputStream.scala
pchlupacek: Ok, you think that will surrelly fork the evaluation to other thread right?

mpilquist: Yep (otherwise, things like concurrent.join wouldn't work)

io/src/main/scala/fs2/io/ToInputStream.scala
mpilquist: Yeah, I think it would be good to have all the java input/output stream stuff in the same file -- I think that's more consistent with the style of the code base as-is.

mpilquist:
Any updates on this?

pchlupacek:
just made an update with changes you have suggested.

PR #732, opened 2016-09-12
Add ExecutorService parameter to AsynchronousFileChannel.open call.

fiadliel:
Also fixes the signature of fromPathAsync to only require a single
implicit parameter of type Async.


mpilquist:
👍

mpilquist:
@fiadliel We should add the same parameter (`Option[ExecutorService] = None`) to `fs2.io.file.{readAllAsync,writeAllAsync}.

fiadliel:
Updated

PR #737, opened 2016-09-19
Fix scaladoc example for the zipWithScan1

Unisay:
Scaladoc example for the `zipWithScan1` has been copy-pasted from `zipWithScan`.
I ran the command
```scala
Stream("uno", "dos", "tres", "cuatro").zipWithScan1(0)(_ + _.length).toList
```
in the REPL and pasted correct `res0`


mpilquist:
Thanks!

PR #738, opened 2016-09-24
Build runLog result with a VectorBuilder

rossabaker:
runLog builds its result by repeatedly appending to a `Vector`.  For large streams, better performance is possible using a `VectorBuilder`, at some cost to smaller streams.  The crossover point is not clear from [initial benchmarks], but might be evaluated with a smarter choice of GenerateN.

[initial benchmarks]: https://gist.github.com/rossabaker/3e141683db87728824b254f3a281201b


rossabaker:
I am not yet committed to this being a good idea myself.  Presenting it for discussion.  My concerns are:

1. It pays the most on large streams, but these are the ones you don't want in memory.
2. Can this mutability leak out?  @mpilquist immediately caught the missing suspend.  What else?  Are there dragons in stepping?

mpilquist:
I can't think of any other ways in which this would fail. :+1: from me.

PR #741, opened 2016-10-01
Stream constructors for effectful sequences

adelbertc:


pchiusano:
These feel kind of special purpose to me. They don't save much code either. I'm inclined to say 👎.

pchiusano:
Hey @adelbertc, we are going to close this, but wanted to give some guidance on how we'll likely be handling PRs for little utility functions like these going forward.

* For PRs that are more of an 'abbreviation' like this, just giving a name to the composition of 2 or 3 existing functions, it's most likely a 👎  on merging. Reasoning: it saves less than 1 LOC (`eval(a).flatMap(Stream.emits)` vs `Stream.emitsEval(a)`) and is one more thing you have to look up if someone uses it in their code. All else being equal, we prefer to keep things lean and mean and let people define these little helpers in their own codebases that use FS2.
* Exception to this guidance is stuff that is used _a lot_. This is a bit subjective, but we'd probably wait until there were lots of people requesting an abbreviation or tiny convenience function before adding it to the core library.
* Exception to this guidance: if it demonstrably saves a good chunk of code (at least more than 1 line), then it's worth considering.
* If you're unsure, you can always ask in the chat room (or by posting an issue) before putting together a PR.

Eventually we may consolidate this and other guidance on PRs into a doc of some sort that we can point to.

Cheers, and don't let this discourage you from looking for other useful functions to add to the library!

adelbertc:
@pchiusano Sounds good to me, no worries at all :-) Keep up the good work!

PR #744, opened 2016-10-04
fix Stream#repeat

kazzna:
Stream#repeat should return empty Stream when self is empty, or at least it should stop safety.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -219,7 +219,10 @@ final class Stream[+F[_],+O] private (private val coreRef: Stream.CoreRef[F,O]) 
   def pure(implicit S: Sub1[F,Pure]): Stream[Pure,O] = covary[Pure]
 
   /** Repeat this stream an infinite number of times. `s.repeat == s ++ s ++ s ++ ...` */
-  def repeat: Stream[F,O] = self ++ repeat
+  def repeat: Stream[F,O] = for {
+    a <- self.uncons1
+    b <- a.map { case (h, t) => t.cons1(h) ++ repeat }.getOrElse(Stream.empty)
+  } yield b
 
   def runFree: Free[F,Unit] =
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -219,7 +219,10 @@ final class Stream[+F[_],+O] private (private val coreRef: Stream.CoreRef[F,O])
   def pure(implicit S: Sub1[F,Pure]): Stream[Pure,O] = covary[Pure]
 
   /** Repeat this stream an infinite number of times. `s.repeat == s ++ s ++ s ++ ...` */
-  def repeat: Stream[F,O] = self ++ repeat
+  def repeat: Stream[F,O] = for {
+    a <- self.uncons1
+    b <- a.map { case (h, t) => t.cons1(h) ++ repeat }.getOrElse(Stream.empty)
```
djspiewak: What is the value in special-casing like this?  I don't mean to be negative, but it seems like a lot of extra awkwardness just to identify the empty stream in one specific case.

pchiusano: Also consider: eval_(f).repeat

It would be very strange if that only ran the effect once.

Sorry I don't think we can merge this, or any similar special casing.

pchiusano:
Closing this. 

Another example: Scala will certainly not stop you from writing `lazy val s: List[Int] = s ++ s` or any other infinite loop. Safely preventing this sort of thing is really out of scope.

In general, there are always going to be lots of ways of writing infinite loops that aren't productive or aren't what the user may have intended (just like in programming generally). But we aren't going to add special cases to try to detect these situations or handle them differently, as that breaks uniformity and isn't going to be what the user wants in other cases. I hope that makes sense.

core/shared/src/main/scala/fs2/Stream.scala
kazzna: I understand. Thank you.

PR #749, opened 2016-10-22
Task Ref now increments nonce with every set

AdamChlupacek:
As per comments in `fs2.Task.ref` the nonce should be updated with every successful set or modify, up till now for set the nonce was updated only in case of the first set, this PR addresses this issue. 

Added a test that demonstrates the case in which this would be an issue. This would also affect modify of the ref in case there was a concurrent set on it.

### DIFF
--- a/core/jvm/src/test/scala/fs2/TaskSpec.scala
+++ b/core/jvm/src/test/scala/fs2/TaskSpec.scala
@@ -1,0 +1,36 @@
+package fs2
+
+/**
+  * Created by adamchlupacek on 22/10/16.
+  */
+class TaskSpec extends Fs2Spec{
+
+
+  "Task" - {
+
+    "Ref" - {
+
+      "Set increments nonce" in {
+
+        Task.ref[Int].flatMap{ref =>
+          ref.setPure(1).flatMap{_ =>
+            ref.access.flatMap{case ((_, set)) =>
+              ref.setPure(2).flatMap{_ =>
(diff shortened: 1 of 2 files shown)


core/jvm/src/test/scala/fs2/TaskSpec.scala
```diff
@@ -0,0 +1,26 @@
+package fs2
+
+/**
+  * Created by adamchlupacek on 22/10/16.
+  */
```
pchlupacek: Please update comment to be inline with rest of fs2

pchlupacek:
@pchiusano @mpilquist If there are no objections can we merge this?

core/jvm/src/test/scala/fs2/TaskSpec.scala
```diff
@@ -0,0 +1,23 @@
+package fs2
+
+import scala.concurrent.duration._
+
+class TaskSpec extends Fs2Spec{
+
+  "Task" - {
+    "Ref" - {
+      "Set increments nonce" in {
```
pchlupacek: I think nonce is here internal implementation detail. What we really solving here is interleaving set and modify// access. I would prefer updated name with perhaps a short comment explaining what this tests verifies.

mpilquist:
:+1: Per Pavel's last comment, let's update the test name and then we can merge.

pchiusano:
👍

PR #752, opened 2016-10-25
Changed definition of Sink so that the result is a Stream[F,Nothing]

mpilquist:


mpilquist:
@pchiusano OK to merge this to series/1.0?

pchiusano:
I am not so sure. I was thinking a Sink should return Unit, a singleton
stream, so it can be used in for comprehensions without cutting off the
computation.

It makes sense that drain should return a Nothing stream, as that type
indicates most clearly what it is doing, but passing around things that
return Nothing seems error prone, since Scala will promote that to whatever
type (as we saw for drain!)

Another thought is maybe we should just get rid of the alias.

mpilquist:
Good point about `Stream[F,Nothing]` being more error prone since Scala will happily promote it to whatever as necessary. On the other hand, I don't think it is ever safe to flatMap a Sink unless you know something about the implementation of the Sink (namely, whether it ever emits an element) -- the type alone is insufficient. I'm going to close this one though.

pchiusano:
Yeah.

I think it could be a good convention for most sinks to just have the sink
return a single Unit when it completes, so you can just stick that in a for
comprehension without any weird behavior. Another idea might be to have
`.to` do a `.drain ++ emit(())`, which would mean that `.to` was ALWAYS
safe to use in this way, regardless of what the sink does. Have not thought
too much about that though.



On Fri, Oct 28, 2016 at 11:13 AM, Michael Pilquist <notifications@github.com

pchlupacek:
+1 for sink having unit type and return one and only one unit when it completes

pchlupacek:
On second thought, not sure if that isn't sort of misleading too. i.e. 
```scala
for {
  _ <- getInput.to(writeThatToDb)
  _ <- incrementCountWritten.to(someLogging)
} yield ()

```
would seem to me perfectly ok, but the second argument will only get executed when first one will 
complete, and I believe that is not what you want, really. 

From my experience with fs2, the Sink/Channel is not used that often like with 0.8, so one option may be perhaps completely discard Sink/Channel type and combinators related? To me 0.8 sinks were useful as they were guaranteed to be 1:1 to source item and therefore allowed nice composition. Hence that is not case anymore, I really would like to see if they are necessary at all, and if we cannot simply live with only Pipe[F,I,O] instead.

PR #756, opened 2016-10-31
groupBy convenience function

tscholak:
I found a groupBy missing, so I made one.


mpilquist:
Thanks!

wookietreiber:
FYI @tscholak: there are `runGroupByFoldMap`, `runGroupByFoldMonoid` and `runGroupBy` in [fs2-cats](https://github.com/functional-streams-for-scala/fs2-cats/blob/master/shared/src/main/scala/fs2/interop/cats/cats.scala#L45), if that suits your use case.

tscholak:
@wookietreiber wasn't aware of these, thanks!

PR #758, opened 2016-11-02
Fixed order of evaluation of ap and traverse

mpilquist:


mpilquist:
ScalaCheck 1.13.4 upgrade broke a few tests and adding map2 to the Applicative trait broke binary compatibility.

PR #766, opened 2016-11-18
Fixed Chunk.concat to not smash custom chunks

djspiewak:
Critical motivating example: `ByteVectorChunk`.  Without the ability to write this kind of `Chunk`, fs2 cannot be easily used for high-performance IO, simply because it will lack the ability to operate on direct `ByteBuffer` instances (via scodec's `ByteVector`).  This is a pretty big deal for frameworks like http4s.

Now, `Chunk` is not sealed (good!), but `Chunk.concat` has a fairly significant amount of hard-coded assumptions about the type of specialization which is possible.  Specifically, it's assuming that the `Chunk.*` implementations are the _only_ specialized chunks that will ever be useful, with all other unforeseen `Chunk` subtypes falling more into the camp of `Chunk.indexedSeq`.  This assumption stems from the fact that it has a hard-coded set of cases that it checks (very inefficiently, btw!) and it will forcibly despecialize (using element-by-element iteration) any chunks which do not conform to its assumptions.  If we look at the example of `ByteVectorChunk`, the specialization would *sort of* be retained, but you would lose the `ByteVector` and every individual `Byte` would be boxed and read onto the heap (twice, actually), resulting in an instance of `Chunk.Bytes`.  Now, `Chunk.Bytes` is good, but it's not `ByteVectorChunk`.  We need to preserve `ByteVectorChunk` through `Chunk.concat`, or the concept is basically dead on arrival.

Enter: subtype polymorphism!  I've rearranged the branching to be *open* and extensible via subtyping and inheritance.  (note to self: crow to someone who cares about a useful application of inheritance and subtyping)  There are now two new functions on `Chunk`:

```scala
  /** Returns Some(this) if A =:= B. This function is pessimistic, in that it will assume None unless proven safe */
  def conform[B: ClassTag]: Option[Chunk[B]]

  /** Returns a new chunk consisting of the elements from the current chunk followed by the elements of the given chunks, in order */
  def concatAll[B >: A](chunks: Seq[Chunk[B]]): Chunk[B] =
    Chunk.indexedSeq(chunks.foldLeft(this.toVector: Vector[B]) { _ ++ _.toVector })
```

The `conform` function is basically a safe cast, which is necessary to detect and work around covariance when the sequence of chunks is secretly heterogeneous (hopefully rare).  It is basically always safe to return `None` from this function, though there may be circumstances (such as `ByteVectorChunk` where safe and efficient checking is possible).  `concatAll` is really where the meat is.

Intuitively, `Chunk.concat(chunks) = chunks.head concatAll chunks.tail`.  Modulo bounds checking of course.  Basically, this delegates the decision about whether-or-not and *how* to respecialize to the runtime type of the first chunk in the sequence.  If that runtime type is using the default implementation of `concatAll` (which will be fine for any non-specialized chunks), everything is essentially kept despecialized.  Note that if you have a `Chunk` implementation which *wants* to attempt to respecialize the hard-coded primitive chunk types, you can mix in `PolymorphicChunk`.  Otherwise, if you're something like `ByteVectorChunk`, you just implement the `concatAll` function in such a way that your secret hidden specialized features are preserved as much as possible.

Now, as mentioned, I preserved the optimistic respecialization of chunks that *contain* primitives but aren't actually primitive, but I'm not entirely sure it's a good idea.  You would need to have one heck of a downstream pipeline to really make it worthwhile, considering that it results in traversing the entire sequence, *element-by-element*, twice.  We should probably profile it.

This reimplementation did result in some code duplication (since I wanted to avoid creating junk `Seq` instances when I have no way of knowing the asymptotic complexity of the underlying implementation), and there are a *couple* cases that it is going to miss.  For example, `NonEmptyChunk` is a hot flaming mess and I'm really annoyed at its existence.  Respecialization is slightly weaker in cases involving `NonEmptyChunk` than it was previously.  We can improve a bit on the situation in this PR.  Aside from that though, every case that *was* respecialized is *still* respecialized, and the mechanism is now extensible.

---

Dedicated to @rossabaker

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -184,98 +184,7 @@ trait MonomorphicChunk[+A] extends Chunk[A] {
 }
 
 trait PolymorphicChunk[+A] extends Chunk[A] {
-
   final def conform[B: ClassTag] = None
-
-  /**
-   * An implementation of concatAll which special-cases the primitive chunks
-   * defined *here* and attempts to recover specialization downstream.  Alternative
-   * specialized chunk implementations will, necessarily, be omitted from this
-   * "downstream respecialization" process.
-   */
-  final override def concatAll[B >: A](chunks: Seq[Chunk[B]]): Chunk[B] = {
-    def itr = this.iterator ++ chunks.iterator
-
-    /** Concatenates the specified sequence of boolean chunks in to a single chunk. */
-    def concatBooleans(chunks: Seq[Chunk[Boolean]]): Chunk[Boolean] = {
-      if (chunks.isEmpty) Chunk.empty
-      else {
-        val size = chunks.foldLeft(0)(_ + _.size)
-        val arr = Array.ofDim[Boolean](size)
-        var offset = 0
-        chunks.foreach { c =>
-          if (!c.isEmpty) {
-            c.copyToArray(arr, offset)
-            offset += c.size
-          }
-        }
-        Chunk.booleans(arr)
-      }
-    }
-
-    /** Concatenates the specified sequence of byte chunks in to a single chunk. */
-    def concatBytes(chunks: Seq[Chunk[Byte]]): Chunk[Byte] = {
-      if (chunks.isEmpty) Chunk.empty
-      else {
-        val size = chunks.foldLeft(0)(_ + _.size)
-        val arr = Array.ofDim[Byte](size)
-        var offset = 0
-        chunks.foreach { c =>
-          if (!c.isEmpty) {
-            c.copyToArray(arr, offset)
-            offset += c.size
-          }
-        }
-        Chunk.bytes(arr)
-      }
-    }
-
-    /** Concatenates the specified sequence of double chunks in to a single chunk. */
-    def concatDoubles(chunks: Seq[Chunk[Double]]): Chunk[Double] = {
-      if (chunks.isEmpty) Chunk.empty
-      else {
-        val size = chunks.foldLeft(0)(_ + _.size)
-        val arr = Array.ofDim[Double](size)
-        var offset = 0
-        chunks.foreach { c =>
-          if (!c.isEmpty) {
-            c.copyToArray(arr, offset)
-            offset += c.size
-          }
-        }
-        Chunk.doubles(arr)
-      }
-    }
-
-    /** Concatenates the specified sequence of long chunks in to a single chunk. */
-    def concatLongs(chunks: Seq[Chunk[Long]]): Chunk[Long] = {
-      if (chunks.isEmpty) Chunk.empty
-      else {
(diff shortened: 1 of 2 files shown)


mpilquist:
:+1: This looks good to me. I was sort of dreading the polymorphic concat problem.

Any ideas on how to improve the `NonEmptyChunk` situation? For background, it exists because the main interpreter elides empty chunks for various reasons, yet functions on `Handle` and `Pull` weren't aware of that. For a while, we had things like `receiveNonEmpty` which provided an `Option[Chunk[A]]` with the guarantee that the chunk was non-empty. This situation was arguably much worse than the current situation, as it forced everyone writing chunky pulls to deal with this potentially empty case that could never actually occur yet wasn't evident in the type system anywhere. Further, the functions that had non-empty analogs were completely adhoc -- if we had tried to add them everywhere a guaranteed non-empty chunk showed up in the surface API, we'd effectively double the number of methods.

mpilquist:
Oh, btw -- let's remove the optimistic respecialization of non-specialized primitive chunks. I added that with no proof of its efficacy so I'd rather err on the side of less (hacky) code.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -489,6 +511,33 @@ object Chunk {
     def foldRight[B](z: B)(f: (Boolean,B) => B): B =
       ((size-1) to 0 by -1).foldLeft(z)((tl,hd) => f(at(hd), tl))
 
+    override def concatAll[B >: Boolean](chunks: Seq[Chunk[B]]): Chunk[B] = {
+      val conformed = chunks flatMap { _.conform[Boolean] }
```
mpilquist: Move inside the first else block?

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -889,7 +1020,7 @@ object Chunk {
 /**
  * A chunk which has at least one element.
  */
-sealed trait NonEmptyChunk[+A] extends Chunk[A] {
+sealed trait NonEmptyChunk[+A] extends PolymorphicChunk[A] {
```
mpilquist: Could `NonEmptyChunk` implement `concatAll` by delegating to `underlying.concatAll`?

djspiewak:
I don't have any good ideas on `NonEmptyChunk`.  I mean, I can make it obey the polymorphic concat and not deoptimize things (which I'll do in my next commit), but I don't know how to make it go away.  It definitely needs to be pretty aware of chunk internals and vice versa, but maybe that's more a reflection on chunk's nastiness than anything else.

core/shared/src/main/scala/fs2/Chunk.scala
djspiewak: Only if it doesn't extend `PolymorphicChunk`, which is pretty easy if we remove respecialization.

PR #770, opened 2016-11-20
Create scodec project with ByteVectorChunk

rossabaker:


build.sbt
```diff
@@ -173,6 +173,17 @@ lazy val io = project.in(file("io")).
     name := "fs2-io"
   ).dependsOn(coreJVM % "compile->compile;test->test")
 
+lazy val scodec = crossProject.in(file("scodec")).
```
rossabaker: Not part of scodec-stream because it only depends on scodec-bits, and because scodec-core is a faster moving target.  Maybe the name should be more specific. :shrug:

core/shared/src/test/scala/fs2/ChunkProps.scala
```diff
@@ -0,0 +1,69 @@
+package fs2
+
+import scala.reflect.ClassTag
+
+import org.scalacheck.Arbitrary
+import org.scalatest.Matchers
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+
+object ChunkProps
```
rossabaker: Maybe we should introduce discipline to bundle these up?

scodec/shared/src/main/scala/fs2/scodec/ByteVectorChunk.scala
```diff
@@ -0,0 +1,74 @@
+package fs2
+package scodec
+
+import scala.reflect.{ClassTag, classTag}
+import _root_.scodec.bits.ByteVector
+
+final class ByteVectorChunk private (val toByteVector: ByteVector)
+    extends MonomorphicChunk[Byte]
+{
```
djspiewak: K&R brace. Boo! Hiss!

scodec/shared/src/main/scala/fs2/scodec/ByteVectorChunk.scala
```diff
@@ -0,0 +1,74 @@
...
+  def size: Int =
+    toByteVector.size.toInt
+
+  def take(n: Int): Chunk[Byte] =
+    ByteVectorChunk(toByteVector.take(n))
+
+  protected val tag: ClassTag[_] =
+    classTag[Byte]
+
+  override def concatAll[B >: Byte](chunks: Seq[Chunk[B]]): Chunk[B] = {
```
rossabaker: We should think about what we expect if we mix and match with `Chunk.Bytes`, and then make sure that it happens.

scodec/shared/src/main/scala/fs2/scodec/ByteVectorChunk.scala
```diff
@@ -0,0 +1,74 @@
+package fs2
+package scodec
+
+import scala.reflect.{ClassTag, classTag}
+import _root_.scodec.bits.ByteVector
```
rossabaker: The package name `fs2.scodec` means root imports may be likely in other code, too.  Open to suggestions.

scodec/shared/src/main/scala/fs2/scodec/ByteVectorChunk.scala
```diff
@@ -0,0 +1,74 @@
...
+{
+  def apply(i: Int): Byte =
+    toByteVector(i)
+
+  def copyToArray[B >: Byte](xs: Array[B], start: Int): Unit =
+    xs match {
+      case byteArray: Array[Byte] =>
+        toByteVector.copyToArray(byteArray, start)
+      case _ =>
+        toByteVector.toArray.iterator.copyToArray(xs, start)
```
djspiewak: Not that it matters much, but why not just replace this with `iterator.copyToArray(xs, start)`?  I mean the slow-path, non-`Byte` case, obviously.

scodec/shared/src/main/scala/fs2/scodec/ByteVectorChunk.scala
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  def foldLeft[B](z: B)(f: (B, Byte) => B): B =
+    toByteVector.foldLeft(z)(f)
+
+  def foldRight[B](z: B)(f: (Byte, B) => B): B =
+    toByteVector.foldRight(z)(f)
+
+  def size: Int =
+    toByteVector.size.toInt
```
rossabaker: This `toInt` is bad news.  If the world needed long `ByteVector`s, do we need long `Chunk`s?

scodec/shared/src/main/scala/fs2/scodec/ByteVectorChunk.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  def take(n: Int): Chunk[Byte] =
+    ByteVectorChunk(toByteVector.take(n))
+
+  protected val tag: ClassTag[_] =
+    classTag[Byte]
+
+  override def concatAll[B >: Byte](chunks: Seq[Chunk[B]]): Chunk[B] = {
+    val conformed = chunks collect { case toByteVectorc: ByteVectorChunk => toByteVectorc }
```
djspiewak: If you use `conform[Byte]` and *slightly* complicate the logic on line 64, you can add support for efficiently concatenating `Chunk.Bytes` values into `ByteVectorChunk`, taking advantage of `ByteVector`'s internal trie rather than falling back on `IndexedSeq` (line 66).

djspiewak:
👍

build.sbt
mpilquist: Add `scodecJVM` and `scodecJS` to the root aggregate?

rossabaker: Oops.  Yes.

PR #776, opened 2016-11-24
Minor performance improvement in Free

mpilquist:
This PR improves performance of Free by eagerly applying `Bind(Pure(a), f)` patterns in `Free#step`. We used to do this but it had the unfortunate side effect of evaluating suspends, because suspend is (still) implemented as `pure(()) >> fa`. This PR takes a conservative approach to fixing this issue by introducing a new flag on the `Pure` constructor, indicating whether or not eager evaluation is supported. `Free.pure` sets the flag to true whereas `Free.suspend` sets the flag to false.

The Free benchmarks show 18% - 42% increase:

a0aef44d45e76ebfe6b24aee3fa71664b7bbca0e (head of series/1.0)
```
[info] Benchmark                      Mode  Cnt   Score   Error  Units
[info] FreeBenchmark.nestedFlatMaps  thrpt   10  11.023 ± 1.311  ops/s
[info] FreeBenchmark.nestedMaps      thrpt   10   7.292 ± 1.810  ops/s
```

fa9aa99488b4f4e902fcf528acfb027dd41cd596 (this PR)

```
info] Benchmark                      Mode  Cnt   Score   Error  Units
[info] FreeBenchmark.nestedFlatMaps  thrpt   10  13.660 ± 1.745  ops/s
[info] FreeBenchmark.nestedMaps      thrpt   10  10.412 ± 1.739  ops/s
```

The Stream benchmarks show a wider range of improvements -- the worse case is about equal (`runLog_102400`) and the best case shows a 330% improvement (`awaitPull_256`).

a0aef44d45e76ebfe6b24aee3fa71664b7bbca0e (head of series/1.0)

```
[info] Benchmark                                  Mode  Cnt    Score    Error  Units
[info] StreamBenchmark.awaitPull_256             thrpt   10   57.634 ±  8.218  ops/s
[info] StreamBenchmark.eval_102400               thrpt   10    0.616 ±  0.025  ops/s
[info] StreamBenchmark.leftAssocConcat_102400    thrpt   10    3.261 ±  0.448  ops/s
[info] StreamBenchmark.leftAssocFlatMap_102400   thrpt   10    9.509 ±  0.501  ops/s
[info] StreamBenchmark.rightAssocConcat_102400   thrpt   10    3.171 ±  0.409  ops/s
[info] StreamBenchmark.rightAssocFlatMap_102400  thrpt   10    8.579 ±  0.496  ops/s
[...]

PR #777, opened 2016-11-24
Tail recursive Free.map fusion

djspiewak:
A continuation of @mpilquist's earlier branch, but with stack-safe map fusion.  Also a rediscovery of scalac's bug with `tailrec` and existential types.

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/FreeBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/FreeBenchmark.scala
@@ -1,0 +1,30 @@
+package fs2
+package benchmark
+
+import org.openjdk.jmh.annotations.{Benchmark, State, Scope}
+
+import fs2.util.Free
+
+@State(Scope.Thread)
+class FreeBenchmark extends BenchmarkUtils {
+
+  val N = 1000000
+
+  @Benchmark
+  def nestedMaps = {
+    val nestedMapsFree = (0 to N).foldLeft(Free.pure(0): Free[Task, Int]) { (acc, i) => acc.map(_ + i) }
+    nestedMapsFree.run
+  }
+
+  @Benchmark
+  def nestedFlatMaps = {
+    val nestedFlatMapsFree = (0 to N).foldLeft(Free.pure(0): Free[Task, Int]) { (acc, i) => acc.flatMap(j => Free.pure(i + j)) }
+    nestedFlatMapsFree.run
+  }
+
+  @Benchmark
+  def alternatingMapFlatMap = {
+    val nestedMapsFree = (0 to N).foldLeft(Free.pure(0): Free[Task, Int]) { (acc, i) => if (i % 2 == 0) acc.map(_ + i) else acc.flatMap(j => Free.pure(i + j))}
+    nestedMapsFree.run
+  }
+}
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -39,6 +39,7 @@ final class Pull[+F[_],+O,+R] private (private val get: Free[AlgebraF[F,O]#f,Opt
         case Algebra.Eval(fr) => StreamCore.evalScope(fr.attempt).flatMap(f)
         case Algebra.Output(o) => StreamCore.append(o, StreamCore.suspend(f(Right(()))))
       }
+      def map[X](x: X)(f: X => Out) = StreamCore.attemptStream(done(f(x)))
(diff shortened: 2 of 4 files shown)


mpilquist:
Awesome! I'll handle the merge conflict that will result from #776.

mpilquist:
Hoping to work on this again soon. I have some incomplete work from last week that removes `fold` from `Free` in favor of subtype polymorphism (for `translate` and `attempt`) and pattern matching (for `Pull` and `Scope`). I'd like to finish that up first and then I'll reintroduce the `Map` constructor and incorporate this PR.

mpilquist:
@djspiewak In an attempt to make `Map.Stack` more general purpose, specifically supporting stack safe `andThen` as well as `compose`, I put this together:

```

sealed abstract class SFunction1[A, B] extends (A => B) {
  override def compose[C](g: C => A): SFunction1[C, B] = _compose(g)
  protected def _compose[C](g: C => A): SFunction1[C, B]
  override def andThen[C](g: B => C): SFunction1[A, C] = _andThen(g)
  protected def _andThen[C](g: B => C): SFunction1[A, C]
}

object SFunction1 {
  def apply[A, B](f: A => B): SFunction1[A, B] = One(f)

  private final case class One[A, B](f: A => B) extends SFunction1[A, B] {
    def apply(a: A): B = f(a)

    def _compose[C](g: C => A): SFunction1[C, B] = Composition(Vector(g.asInstanceOf[(Any => Any)], f.asInstanceOf[(Any => Any)]))
    def _andThen[C](g: B => C): SFunction1[A, C] = Composition(Vector(f.asInstanceOf[(Any => Any)], g.asInstanceOf[(Any => Any)]))
  }

  private final case class Two[A, B, C](f: A => B, g: B => C) extends SFunction1[A, C] {
    def apply(a: A): C = g(f(a))

    def _compose[D](h: D => A): SFunction1[D, C] =
      Composition(Vector(
        h.asInstanceOf[(Any => Any)],
        f.asInstanceOf[(Any => Any)],
        g.asInstanceOf[(Any => Any)]
      ))

    def _andThen[D](h: C => D): SFunction1[A, D] =
      Composition(Vector(
        f.asInstanceOf[(Any => Any)],
        g.asInstanceOf[(Any => Any)],
        h.asInstanceOf[(Any => Any)]
      ))
  }

  private final case class Composition[A, B](fs: Vector[(Any => Any)]) extends SFunction1[A, B] {
    def apply(a: A): B =
      fs.foldLeft(a: Any)((acc, f) => f(acc)).asInstanceOf[B]

    def _compose[C](f: C => A): SFunction1[C, B] = Composition(f.asInstanceOf[(Any => Any)] +: fs)
    def _andThen[C](f: B => C): SFunction1[A, C] = Composition(fs :+ f.asInstanceOf[(Any => Any)])
  }
}
```

I special cased the 1/2 function cases to avoid the overhead of the vector allocation unless it's really needed. Any thoughts on this? Good idea / bad idea?

mpilquist:
Quick update -- @pchiusano suggested using `Catenable` in place of `Vector`, which simplifies this to just:

```scala
final class CFunction1[A, B](private val fs: Catenable[(Any => Any)]) extends (A => B) {
  override def compose[C](g: C => A): CFunction1[C, B] = new CFunction1(g.asInstanceOf[(Any => Any)] :: fs)
  override def andThen[C](g: B => C): CFunction1[A, C] = new CFunction1(fs :+ g.asInstanceOf[(Any => Any)])
  def apply(a: A): B = fs.foldLeft(a: Any)((x, f) => f(x)).asInstanceOf[B]
}
object CFunction1 {
  def apply[A, B](f: A => B): CFunction1[A, B] = new CFunction1(Catenable.single(f.asInstanceOf[Any => Any]))
}
```

In so doing, I found some issues in Catenable which I'll PR soon.

mpilquist:
Another update -- I haven't been able to come up with a version of map fusion that improves performance. The gains in the original PR have all been outdone by the new `step` function, which eagerly applies the `Bind(Pure(a, true), f)` pattern. I'm not ready to abandon this but it isn't looking good at the moment.

djspiewak:
@mpilquist That doesn't surprise me *too* much.  Map fusion is most exciting when you can combine it with fusing sequence-reducing operations like `take`/`drop` and `filter`.  The big win is actually not so much the fusion itself (though reducing traversals helps) as it is the fact that you can short-circuit "earlier" operations on the basis of "later" ones.

djspiewak:
@mpilquist Closing this for the time being.  We can always resurrect it if we feel it may contain meaningful gains (or inspiration for such) in the future.

PR #784, opened 2016-12-02
Stack safety and constant factor improvements of Catenable

mpilquist:
This PR drastically improves constant factor performance of `Catenable` in certain usage patterns. API is changed a little bit:
 - `push` was renamed `cons`
 - `::` was renamed to `+:`
 - `:+`/`snoc` was added
 - `toStream` replaced with `toList`
 - `foldLeft` and `foreach` added
 - fixed stack safety bug in `uncons` and `fromSeq` -- `reduceRight` is *not* stack safe so the input is now reversed

Of particular note is the fact that singleton catenable construction was only about 1.5x as fast as creating a singleton vector, and small catenable construction was a bit slower than small vector creation.

Please excuse the while loops, but this is performance sensitive code so :shrug:.

8e20c6f1305ada0f50ed441f706228d1b5133f93 (before the changes in this PR)

```
[info] Benchmark                                 Mode  Cnt          Score         Error  Units
[info] CatenableBenchmark.consLargeCatenable    thrpt   10  123886314.876 ± 1230223.043  ops/s
[info] CatenableBenchmark.consLargeVector       thrpt   10   10566409.027 ±  710173.165  ops/s
[info] CatenableBenchmark.consSmallCatenable    thrpt   10  120474774.758 ± 3335136.723  ops/s
[info] CatenableBenchmark.consSmallVector       thrpt   10   16325204.308 ±  299289.520  ops/s
[info] CatenableBenchmark.createSmallCatenable  thrpt   10    8947210.997 ±  325165.771  ops/s
[info] CatenableBenchmark.createSmallVector     thrpt   10   11255936.788 ±  203349.167  ops/s
[info] CatenableBenchmark.createTinyCatenable   thrpt   10   22871648.117 ±  613434.137  ops/s
[info] CatenableBenchmark.createTinyVector      thrpt   10   15123925.504 ±  484690.580  ops/s
[info] CatenableBenchmark.mapLargeCatenable     thrpt   10          2.580 ±       0.740  ops/s
[info] CatenableBenchmark.mapLargeVector        thrpt   10         87.447 ±       6.032  ops/s
[info] CatenableBenchmark.mapSmallCatenable     thrpt   10    1500240.055 ±   32720.321  ops/s
[info] CatenableBenchmark.mapSmallVector        thrpt   10   13814716.675 ±  319184.790  ops/s
```

796700be3a8e683985f0cf04e1f014e7bb9a5a1f (this PR)

```
[info] Benchmark                                   Mode  Cnt          Score         Error  Units
[info] CatenableBenchmark.consLargeCatenable      thrpt   10  150443328.810 ± 4262697.027  ops/s
[info] CatenableBenchmark.consLargeVector         thrpt   10   12215271.536 ±  780910.325  ops/s
[info] CatenableBenchmark.consSmallCatenable      thrpt   10  146698308.995 ± 8798103.672  ops/s
[info] CatenableBenchmark.consSmallVector         thrpt   10   19154715.531 ±  854095.070  ops/s
[info] CatenableBenchmark.createSmallCatenable    thrpt   10   10458084.536 ±  111327.117  ops/s
[info] CatenableBenchmark.createSmallVector       thrpt   10   12795767.588 ±   50284.190  ops/s
[info] CatenableBenchmark.createTinyCatenable     thrpt   10  128573263.762 ± 1670802.767  ops/s
[info] CatenableBenchmark.createTinyVector        thrpt   10   17131861.034 ±  315356.904  ops/s
[info] CatenableBenchmark.foldLeftLargeCatenable  thrpt   10         80.424 ±       0.877  ops/s
[info] CatenableBenchmark.foldLeftLargeVector     thrpt   10         77.336 ±       2.238  ops/s
[info] CatenableBenchmark.foldLeftSmallCatenable  thrpt   10   10220962.669 ±   96901.748  ops/s
[info] CatenableBenchmark.foldLeftSmallVector     thrpt   10   18855179.003 ±  175591.352  ops/s
[info] CatenableBenchmark.mapLargeCatenable       thrpt   10         16.990 ±       4.297  ops/s
[info] CatenableBenchmark.mapLargeVector          thrpt   10         67.306 ±       2.419  ops/s
[info] CatenableBenchmark.mapSmallCatenable       thrpt   10    8988928.329 ±  252565.606  ops/s
[info] CatenableBenchmark.mapSmallVector          thrpt   10   15259942.354 ± 1126953.820  ops/s
```

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/CatenableBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/CatenableBenchmark.scala
@@ -19,9 +19,14 @@ class CatenableBenchmark extends BenchmarkUtils {
   @Benchmark def mapLargeCatenable = largeCatenable.map(_ + 1)
   @Benchmark def mapLargeVector = largeVector.map(_ + 1)
 
-  @Benchmark def consSmallCatenable = 0 :: smallCatenable
+  @Benchmark def foldLeftSmallCatenable = smallCatenable.foldLeft(0)(_ + _)
+  @Benchmark def foldLeftSmallVector = smallVector.foldLeft(0)(_ + _)
+  @Benchmark def foldLeftLargeCatenable = largeCatenable.foldLeft(0)(_ + _)
+  @Benchmark def foldLeftLargeVector = largeVector.foldLeft(0)(_ + _)
+
+  @Benchmark def consSmallCatenable = 0 +: smallCatenable
   @Benchmark def consSmallVector = 0 +: smallVector
-  @Benchmark def consLargeCatenable = 0 :: largeCatenable
+  @Benchmark def consLargeCatenable = 0 +: largeCatenable
   @Benchmark def consLargeVector = 0 +: largeVector
 
   @Benchmark def createTinyCatenable = Catenable(1)
--- a/benchmark/src/main/scala/fs2/benchmark/CatenableBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/CatenableBenchmark.scala
@@ -1,0 +1,31 @@
+package fs2
+package benchmark
+
+import org.openjdk.jmh.annotations.{Benchmark, State, Scope}
+
+import fs2.util.Catenable
+
+@State(Scope.Thread)
+class CatenableBenchmark extends BenchmarkUtils {
+
+  val smallCatenable = Catenable(1, 2, 3, 4, 5)
+  val smallVector = Vector(1, 2, 3, 4, 5)
+
+  val largeCatenable = Catenable.fromSeq(0 to 1000000)
+  val largeVector = (0 to 1000000).toVector
+
+  @Benchmark def mapSmallCatenable = smallCatenable.map(_ + 1)
+  @Benchmark def mapSmallVector = smallVector.map(_ + 1)
+  @Benchmark def mapLargeCatenable = largeCatenable.map(_ + 1)
+  @Benchmark def mapLargeVector = largeVector.map(_ + 1)
+
+  @Benchmark def consSmallCatenable = 0 :: smallCatenable
+  @Benchmark def consSmallVector = 0 +: smallVector
+  @Benchmark def consLargeCatenable = 0 :: largeCatenable
+  @Benchmark def consLargeVector = 0 +: largeVector
+
+  @Benchmark def createTinyCatenable = Catenable(1)
+  @Benchmark def createTinyVector = Vector(1)
+  @Benchmark def createSmallCatenable = Catenable(1, 2, 3, 4, 5)
+  @Benchmark def createSmallVector = Vector(1, 2, 3, 4, 5)
+}
--- a/core/shared/src/main/scala/fs2/util/Catenable.scala
+++ b/core/shared/src/main/scala/fs2/util/Catenable.scala
@@ -13,19 +13,21 @@ sealed abstract class Catenable[+A] {
   /** Returns the head and tail of this catenable if non empty, none otherwise. Amortized O(1). */
   final def uncons: Option[(A, Catenable[A])] = {
     var c: Catenable[A] = this
-    var rights: List[Catenable[A]] = Nil
+    val rights = new collection.mutable.ArrayBuffer[Catenable[A]]
     var result: Option[(A, Catenable[A])] = null
     while (result eq null) {
       c match {
         case Empty =>
-          rights match {
-            case Nil => result = None
-            case h :: t => c = h; rights = t
+          if (rights.isEmpty) {
+            result = None
+          } else {
+            c = rights.last
+            rights.trimEnd(1)
           }
         case Single(a) =>
-          val next = if (rights.isEmpty) empty else rights.reverse.reduceLeft((x, y) => Append(y,x))
+          val next = if (rights.isEmpty) empty else rights.reduceLeft((x, y) => Append(y,x))
           result = Some(a -> next)
-        case Append(l, r) => c = l; rights = r :: rights
+        case Append(l, r) => c = l; rights += r
       }
     }
     result
@@ -68,19 +70,21 @@ sealed abstract class Catenable[+A] {
   /** Applies the supplied function to each element, left to right. */
   final def foreach(f: A => Unit): Unit = {
     var c: Catenable[A] = this
-    var rights: List[Catenable[A]] = Nil
+    val rights = new collection.mutable.ArrayBuffer[Catenable[A]]
     while (c ne null) {
       c match {
         case Empty =>
-          rights match {
-            case Nil => c = null
-            case h :: t => c = h; rights = t
+          if (rights.isEmpty) {
+            c = null
+          } else {
+            c = rights.last
+            rights.trimEnd(1)
           }
         case Single(a) =>
           f(a)
-          c = if (rights.isEmpty) Empty else rights.reverse.reduceLeft((x, y) => Append(y,x))
-          rights = Nil
-        case Append(l, r) => c = l; rights = r :: rights
+          c = if (rights.isEmpty) Empty else rights.reduceLeft((x, y) => Append(y,x))
+          rights.clear()
+        case Append(l, r) => c = l; rights += r
       }
     }
   }
@@ -125,14 +129,21 @@ object Catenable {
 
   /** Creates a catenable from the specified elements. */
   def apply[A](as: A*): Catenable[A] = {
-    // Assumption: `as` is small enough that calling size doesn't outweigh benefit of calling empty/single
-    as.size match {
-      case 0 => empty
-      case 1 => single(as.head)
-      case n if n <= 1024 =>
-        // nb: avoid cost of reversing input if colection is small
-        as.view.map(single).reduceRight(Append(_, _))
-      case n => as.view.reverse.map(single).reduceLeft((x, y) => Append(y, x))
+    as match {
+      case w: collection.mutable.WrappedArray[A] =>
+        if (w.isEmpty) empty
+        else if (w.size == 1) single(w.head)
+        else {
+          val arr: Array[A] = w.array
+          var c: Catenable[A] = single(arr.last)
+          var idx = arr.size - 2
+          while (idx >= 0) {
+            c = Append(single(arr(idx)), c)
+            idx -= 1
+          }
+          c
+        }
+      case _ => fromSeq(as)
     }
   }
 }
--- a/core/shared/src/main/scala/fs2/util/Catenable.scala
+++ b/core/shared/src/main/scala/fs2/util/Catenable.scala
@@ -5,70 +5,134 @@ import Catenable._
 /**
  * Trivial catenable sequence. Supports O(1) append, and (amortized)
  * O(1) `uncons`, such that walking the sequence via N successive `uncons`
- * steps takes O(N). Like a difference list, conversion to a `Stream[A]`
+ * steps takes O(N). Like a difference list, conversion to a `Seq[A]`
  * takes linear time, regardless of how the sequence is built up.
  */
 sealed abstract class Catenable[+A] {
 
   /** Returns the head and tail of this catenable if non empty, none otherwise. Amortized O(1). */
-  def uncons: Option[(A, Catenable[A])] = {
-    @annotation.tailrec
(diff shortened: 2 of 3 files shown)


core/shared/src/main/scala/fs2/util/Catenable.scala
```diff
@@ -5,70 +5,134 @@ import Catenable._
...
   /** Returns the head and tail of this catenable if non empty, none otherwise. Amortized O(1). */
-  def uncons: Option[(A, Catenable[A])] = {
-    @annotation.tailrec
-    def go(c: Catenable[A], rights: List[Catenable[A]]): Option[(A,Catenable[A])] = c match {
-      case Empty => rights match {
-        case Nil => None
-        case c :: rights => go(c, rights)
+  final def uncons: Option[(A, Catenable[A])] = {
+    var c: Catenable[A] = this
+    var rights: List[Catenable[A]] = Nil
```
pchiusano: Did you look into using an `ArrayList` or something similar here? Advantage would be less allocation, and no need to reverse it to do the reduce right, since you have random access.

If you're going to go imperative, go big. :)

core/shared/src/main/scala/fs2/util/Catenable.scala
```diff
@@ -5,70 +5,134 @@ import Catenable._
...
+    var result: Option[(A, Catenable[A])] = null
+    while (result eq null) {
+      c match {
+        case Empty =>
+          rights match {
+            case Nil => result = None
+            case h :: t => c = h; rights = t
+          }
+        case Single(a) =>
+          val next = if (rights.isEmpty) empty else rights.reverse.reduceLeft((x, y) => Append(y,x))
```
pchiusano: basically, you could avoid this `reverse` call

core/shared/src/main/scala/fs2/util/Catenable.scala
mpilquist: I didn't benchmark it but I figured it would be significantly slower since whatever we use here needs O(1) cons/uncons too.

benchmark/src/main/scala/fs2/benchmark/CatenableBenchmark.scala
```diff
@@ -0,0 +1,36 @@
...
+  @Benchmark def foldLeftLargeVector = largeVector.foldLeft(0)(_ + _)
+
+  @Benchmark def consSmallCatenable = 0 +: smallCatenable
+  @Benchmark def consSmallVector = 0 +: smallVector
+  @Benchmark def consLargeCatenable = 0 +: largeCatenable
+  @Benchmark def consLargeVector = 0 +: largeVector
+
+  @Benchmark def createTinyCatenable = Catenable(1)
+  @Benchmark def createTinyVector = Vector(1)
+  @Benchmark def createSmallCatenable = Catenable(1, 2, 3, 4, 5)
```
mpilquist: I thought the same thing actually. Before I went down that path, I wanted to look in to the paper I linked above.

core/shared/src/main/scala/fs2/util/Catenable.scala
pchiusano: Here's what I was thinking - 

* On each step through the loop, you snoc onto the end of the ArrayList. No consing.
* Now you get to the end, and do a foldLeft over your ArrayList. Notice that the first element of the ArrayList is the first element you snoc'd (whereas the first element of the `List` you are building here is the _last_ element you pushed), that is, the `ArrayList` is already in exactly the right order to implement the `reduceRight` operation via a foldLeft or reduceLeft! (Assuming I haven't mixed this up... )

benchmark/src/main/scala/fs2/benchmark/CatenableBenchmark.scala
pchiusano: IMO that paper is overkill for this use case (amortized is totally fine) and I'm sure the constant factors will be worse. If you think about it, the implementation of `append` for `Catenable` is literally the fastest it could possibly be since it does absolutely nothing. And the reassociating is also going to be very hard to beat in terms of speed since it can be done with a mutable data structure as I suggested above, rather than some fancy deamortized functional version of a similar algorithm.

Paper might still be worth reading just for fun though, don't want to discourage that. :) Also maybe I am totally wrong in my intuition here. :)

core/shared/src/main/scala/fs2/util/Catenable.scala
```diff
@@ -5,70 +5,134 @@ import Catenable._
...
+    else s.view.reverse.map(single).reduceLeft((x, y) => Append(y, x))
 
-  def apply[A](as: A*): Catenable[A] = fromSeq(as)
+  /** Creates a catenable from the specified elements. */
+  def apply[A](as: A*): Catenable[A] = {
+    // Assumption: `as` is small enough that calling size doesn't outweigh benefit of calling empty/single
+    as.size match {
+      case 0 => empty
+      case 1 => single(as.head)
+      case n if n <= 1024 =>
```
pchiusano: isn't in practice an `A*` passed as something with random access? (Like an `ArraySeq`?) Maybe exploit this... 

The 1024 limit might blow some stacks. Makes me a little nervous.

mpilquist: Yeah, `scala.collection.mutable.WrappedArray`. Maybe 1024 is too high -- we could limit to 16 or 32 and cover 99.9% of cases.

mpilquist:
@pchiusano Pushed an update that addresses your comments. Got another boost in performance. I may still tinker with adding a `Many` constructor but that can be a future PR if ever.

```
[info] Benchmark                                   Mode  Cnt          Score         Error  Units
[info] CatenableBenchmark.consLargeCatenable      thrpt   10  156511367.812 ± 8087097.323  ops/s
[info] CatenableBenchmark.consLargeVector         thrpt   10   13763761.206 ±  184527.617  ops/s
[info] CatenableBenchmark.consSmallCatenable      thrpt   10  161497470.969 ± 2223075.836  ops/s
[info] CatenableBenchmark.consSmallVector         thrpt   10   19581030.319 ± 1212212.270  ops/s
[info] CatenableBenchmark.createSmallCatenable    thrpt   10   28728645.518 ± 1368271.566  ops/s
[info] CatenableBenchmark.createSmallVector       thrpt   10   13061785.850 ±  743157.445  ops/s
[info] CatenableBenchmark.createTinyCatenable     thrpt   10  174622244.992 ± 7276262.762  ops/s
[info] CatenableBenchmark.createTinyVector        thrpt   10   17721710.798 ± 1287136.894  ops/s
[info] CatenableBenchmark.foldLeftLargeCatenable  thrpt   10         82.947 ±       1.819  ops/s
[info] CatenableBenchmark.foldLeftLargeVector     thrpt   10         86.150 ±       6.858  ops/s
[info] CatenableBenchmark.foldLeftSmallCatenable  thrpt   10   12992957.180 ±  944477.787  ops/s
[info] CatenableBenchmark.foldLeftSmallVector     thrpt   10   19900091.809 ±  234527.790  ops/s
[info] CatenableBenchmark.mapLargeCatenable       thrpt   10         19.982 ±       3.949  ops/s
[info] CatenableBenchmark.mapLargeVector          thrpt   10         74.918 ±       3.691  ops/s
[info] CatenableBenchmark.mapSmallCatenable       thrpt   10   10281716.339 ±  111062.719  ops/s
[info] CatenableBenchmark.mapSmallVector          thrpt   10   17131102.935 ±  164605.819  ops/s
```

pchiusano:
Nice... not sure where I should be looking to see boost in perf, but looks good to me. Merge when ready.

djspiewak:
Super-cool!

benchmark/src/main/scala/fs2/benchmark/CatenableBenchmark.scala
edmundnoble: @mpilquist I want to look into adding this data structure into Scala with 2.13, if that's alright with you. @pchiusano I agree that append is as fast as it could be, but I wonder about adding Many or looking at the balanced-tree creation approach in https://github.com/scalaz/scalaz/pull/1022 which may make reassociating more performant.

I am not an *expert* on the subject but Ed Kmett said that amortized constant is not enough for reflection without remorse; but I am not sure why. Either way it's much more complicated, but I would be very interested in an implementation of this paper in Scala regardless for real-time applications.

pchiusano:
Sure, go for it (looking into adding to 2.13). Also if you or anyone else
would like to investigate adding Many that would be welcome also. If it
ends up paying off you can open a PR.

However the way this data structure is used in FS2, amortized is totally
fine, so it would not make sense to use a fancier deamortized structure
with worse constant factors. And for the reasons I gave earlier, I suspect
this data structure will be very difficult to beat for amortized
performance.

Didn't read the scalaz stuff on balanced traversals too closely, but don't
see its relevance here. We are using a mutable stack to perform the
reassociating, it will be very difficult to do better than that.

benchmark/src/main/scala/fs2/benchmark/CatenableBenchmark.scala
djspiewak: @pchiusano @edmundnoble I've done a lot of work with amortized factors in functional data structures (a few years ago).  The problem with amortization arguments in all contexts, but *especially* functional data structures, is they often require enormously high values of *n* to come into play.  Additionally, amortization arguments often ignore constant factors (in fact, that's literally what the argument is doing in an inductive fashion), which is not a fair argument to make when your constant factors are extremely high.

There are two excellent and classic examples of this problem: BankersQueue and FingerTree.  BankersQueue is basically just a lazy version of the eager "front/back list queue" thing that everyone's tried at least once.  And there is a *proof* that it is more efficient than the eager version… for some arbitrarily large value of *n*.  It turns out that, if you benchmark it, the eager version is almost twice as fast (in Scala) for any queue size that fits into 32bit memory, which is sort of insane.  FingerTree is a similar, even more dramatic example.  FingerTree is an amortized constant *double-ended queue*, which is something that the eager banker's queue can't provide, and so in a very real sense it is offering (and proving) better *asymptotes*, not just better amortization costs.  But on the JVM, and for queue sizes less than the billions, it is massively, *hilariously* slower than the alternatives.

So we have to be careful about this stuff.  Amortization arguments, discard by definition performance factors which are relevant and even dominant in practical workloads.  And this is especially true on the JVM and with a functional style, where amortization often relies in thunking and other pipeline- and PIC-defeating constructs that deoptimize code (by a constant factor) in ways below the level of algorithmic analysis.

edmundnoble:
Essentially, the relevance the Scalaz thing has is that it's built up from 1. a reassociating loop and 2. a tree-building loop, to build balanced `ap` trees. That seems to imply it is possible to do the exact same thing here to build a balanced binary tree, providing us with an efficient double-ended queue implementation. That would however be a new datatype separate from `Catenable` and might require using an `Array` instead of a mutable stack to preserve asymptotics, because it uses random access, but it might be a good candidate for a replacement for `Vector` in the stdlib. 

Also it appears that the use of `Catenable`  in `StreamCore` is very similar to that in reflection without remorse (repeated appends), which might run into the same issues. The biggest performance cost should be for alternately `append`ing single stream segments and stepping through the stream: switching between continuing (`append`) and observing (`uncons`). If that is a common use case the amortization argument falls apart.

pchiusano:
Switching between append and uncons does not seem to be a problem for Catenable, unless I am missing something. uncons does the minimum reassociating work necessary to produce the head and tail of the sequence: 

Example, if I have: 

Append(One(hd), tl), uncons takes O(1).
Append(Append(One(hd), tl), z), uncons would push [tl, z] on the stack, and return Some(hd, Append(tl,z)). Still fast. 

In general your uncons takes as many steps as there are left-associated appends, but that work isn't repeated since the tail then has the correct associativity.

I think the only pathological case for the structure is alternating the direction of your traversal, like if you uncons then unsnoc repeatedly, that's going to be quadratic. Otherwise you are just doing the reassociating in a consistent direction, and not repeating any work.

I suspect there's a nice proof that any sequence of m appends and n uncons operations takes O(n + m). 

Here's what I had to say last time I discussed this with Ed:

> An interesting observation (to me anyway) is that in reflection without remorse, they observe a performance problem with DLists / CPS when alternating between 'building' and 'observing' the structure, then reach immediately for various fancier functional data structures, but it's possible to get away with something simpler... if you just need to eliminate that one performance problem.
> 
> Another observation is that while you may be building up the data structure in various ways, for most applications (like streams), you only care about efficient traversal left to right.

PR #787, opened 2016-12-28
Added unfoldSeq and unfoldSeqEval

almendar:
Hi,

I was asking on FS2 gitter for a way how to create a stream from a resource that has sth similar to `hasNext` like in iterator or `next` like in `ResultSet` from JDBC but I wanted to emit multiple values at once if there are ready to be published.

The solution could be to use unfold, but it can only publish a single value. 
@edmundnoble suggested I could easily write function that would publish Seq and submit a MR so here it is.

Sorry if I'm doing sth wrong but this is my first PR on github.

### DIFF
--- a/core/jvm/src/test/scala/fs2/StreamSpec.scala
+++ b/core/jvm/src/test/scala/fs2/StreamSpec.scala
@@ -114,9 +114,20 @@ class StreamSpec extends Fs2Spec {
       }.map(_._1).toList shouldBe List(0, 1, 1, 2, 3, 5, 8, 13)
     }
 
+    "unfoldSeq" in {
+      Stream.unfoldSeq(3) { s =>
+        if(s > 0) Some((Seq.fill(s)(s), s-1)) else None
+      }.toList shouldBe List(3, 3, 3, 2, 2, 1)
+    }
+
     "unfoldEval" in {
       Stream.unfoldEval(10)(s => Task.now(if (s > 0) Some((s, s - 1)) else None))
         .runLog.unsafeRun().toList shouldBe List.range(10, 0, -1)
+    }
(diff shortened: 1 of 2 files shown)


djspiewak:
Wouldn't it be better to have an `unfoldChunk` rather than `unfoldSeq`?  That would allow full generalization to unboxed chunks, whereas the current version is either element-level or enforced `Seq` boxing.

almendar:
Yes, that makes sense. I've added functions that accepts chunks and wrote some test that uses unboxed data.
I don't know if I should keep the `Seq` version of unfolds? It is possible to define those unfolds using the chunked version, just by transforming the state generator function result:

From `S => Option[(Seq[A],S)] ` to `S => Option[(Chunk[A],S)]` and in the same spirit in others.

djspiewak:
I would be in favor of just removing the `Seq` versions in the interest of a smaller API.  The way I generally look at things, `Chunk` is basically just a better `Seq`.

tscholak:
For performance reasons, would it make sense to implement the unchunked unfold and iterate in terms of these?

tscholak:
I just realize that probably not since evaluation of `f` might be slow....

tscholak:
In doobie, there is `processWithChunkSize`. maybe there should be `unfoldWithChunkSize` and `iterateWithChunkSize`

mpilquist:
@almendar  :+1: from me. I agree with @djspiewak that we can do without the `Seq` variants in an effort to keep the API footprint smaller (and avoid arguments over having `Seq` variants of other APIs). Since this PR has been open a while (sorry!), I'm going to merge it as-is and then remove the `Seq` variants.

PR #797, opened 2017-01-29
Add scaladoc for `uncons1` and `uncons`

jedesah:


mpilquist:
Wonderful docs, thanks!

pchiusano:
We need to add a caveat to these functions that you should use them in conjunction with `scope`. Otherwise FS2 has no idea when you are done examining the stream and doesn't know when to invoke finalizers. Using a `Pull` establishes a scope.

I was actually originally considering making `uncons` private for this reason. 

For instance if you do `blah.onFinalize(f).uncons.flatMap(...) ++ blah2`, well, `blah` will not be finalized until the end of some enclosing scope (say, established by a pull). 

You need to say `(blah.onFinalize(f).uncons.flatMap(...) ++ blah2).scope` (if you want the finalization to happen at the very end, or you can narrow the scope to `blah.onFinalize(f).scope.uncons.flatMap(..) ++ blah` if you want it to happen as soon as the `blah` part is done. (And you could also do `blah.onFinalize(f).pull(h => ..).close ++ blah2` which has the same sort of effect)

Something that I think is not ideal / is arguably a bug is that `runFold` does not establish a scope by default, so if you have a 'naked' uncons that isn't wrapped in some `scope` or `pull`, it's possible to have finalizers not run at all if you uncons but don't reach the end of the stream. 

I'm going to prep a PR to fix this, not sure if we should try to backport to 0.9.

pchlupacek:
@pchiusano thanks for explanation this was really not obvious to me.

PR #798, opened 2017-01-31
Preserve chunkiness in pipe.groupBy

s-webb:
Implementation of pipe.groupBy emits one element to the output stream per K on the input stream. If each chunk of the input stream contains more than one K, this can reduce chunkiness of the output. This alternate implementation can emit more than one K per chunk on the output stream, giving significantly better performance for my use case, at the cost of some additional complexity.

Unless I've missed something.

### DIFF
--- a/core/shared/src/main/scala/fs2/pipe.scala
+++ b/core/shared/src/main/scala/fs2/pipe.scala
@@ -197,21 +197,50 @@ object pipe {
    * any of the chunk's elements.
    */
   def groupBy[F[_], K, V](f: V => K): Pipe[F, V, (K, Vector[V])] = {
-    def go(current: K, buffer: Vector[V]): Handle[F, V] => Pull[F, (K, Vector[V]), Unit] = _.receiveOption {
-      case Some((chunk, h)) =>
-        chunk.indexWhere(f(_) != current) match {
-          case None =>
-            go(current, buffer ++ chunk.toVector)(h)
-          case Some(idx) =>
-            val next = f(chunk(idx))
-            val out = buffer ++ chunk.take(idx).toVector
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/pipe.scala
```diff
@@ -197,21 +197,51 @@ object pipe {
...
+
+    def doChunk(chunk: Chunk[V], h: Handle[F, V], k1: K, out: Vector[V], acc: Vector[(K, Vector[V])]):
+        Pull[F, (K, Vector[V]), Unit] = {
+
+      val differsAt = chunk.indexWhere(f(_) != k1).getOrElse(-1)
+      if (differsAt == -1) {
+        // whole chunk matches the current key, add this chunk to the current Stream of chunks
+        val newOut: Vector[V] = out ++ chunk.toVector
+        if (acc.isEmpty) {
+          Pull.pure(()) >> go(Some((k1, newOut)))(h)
```
mpilquist: Is this `Pull.pure(())` necessary?

core/shared/src/main/scala/fs2/pipe.scala
```diff
@@ -197,21 +197,51 @@ object pipe {
...
+        } else {
+          // potentially outputs one additional chunk (by splitting the last one in two)
+          Pull.output(Chunk.seq(acc)) >> go(Some((k1, newOut)))(h)
```
mpilquist: Change `Chunk.seq` to `Chunk.indexedSeq` for a minor performance improvement.

s-webb:
Thanks for the feedback. I've made the requested changes and squashed them into the original commit, wasn't sure what on the preferred workflow and I thought that would be less effort your end.

PR #804, opened 2017-02-09
Added tcp.SSLSocket

pchlupacek:
Initial SSL implementation. @pchiusano , @mpilquist , @djspiewak  any ideas how to test this? 

Thanks

### DIFF
--- a/io/src/main/scala/fs2/io/ssl/SSLEngine.scala
+++ b/io/src/main/scala/fs2/io/ssl/SSLEngine.scala
@@ -1,0 +1,366 @@
+package fs2.io.ssl
+
+
+
+import java.nio.ByteBuffer
+import javax.net.ssl.SSLEngineResult
+import javax.net.{ssl => jns}
+
+import fs2._
+import fs2.util.Async.Ref
+import fs2.util._
+import fs2.util.syntax._
+
+
+
+/**
+  * Helper to establish asynchronous interface around [[jns.SSLEngine]]
+  *
+  * Please note that all operations here are not safe to be invoked concurrently.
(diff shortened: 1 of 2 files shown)


djspiewak:
Round tripping is probably the best we can do.

pchlupacek:
@djspiewak yeah :-)

pchiusano:
Should this be TLSSocket?

pchlupacek:
actually no strong opinion on naming. However TLS is just one of the possible options SSLEngine supports. On other hand if we will name it TLS we can then swap later implementation of TLS for something truly non-blocking instead of SSLEngine and be consistent. 

thoughts @pchiusano?

mpilquist:
👍 for TLS despite the names in JSSE. Also, given how notoriously difficult it is to securely configure SSLEngine/SSLContext, I'd like to have easy integration with https://github.com/typesafehub/ssl-config -- not a JAR dependency but just appropriate constructors that would allow a user to use both ssl-config-core library with FS2.

mpilquist:
@pchlupacek What's the state of this?

pchlupacek:
We using this in production already, part of fs2-http. Once we sort of agree on naming and modifications I am able to update PR. 

Also not sure what we do about this configuration topic you mentioned. Perhpas that could be another issue once this gets merged ?

pchlupacek:
@mpilquist was thinking about this. I will refactor this against 1.0 once fast will get merged ok ?

mpilquist:
@pchlupacek Yep that sounds good. I'm hoping to merge topic/fast this week.

mpilquist:
@pchlupacek When you get a chance, a 0.10 version of this PR would be great.

Issue #800, opened 2017-02-09
Stream is missing >> Operator


ChristopherDavenport:
Task and Pull both contain a >> operator which is an alias for 
```scala
t1 >> t2 == t1.flatMap(_ => t2)
```
I propose the same implementation be present for Streams.
```scala
s1 >> s2 == s1.flatMap(_ => s2)
```

PR #805, opened 2017-02-09
Added >> Method to Stream for 0.9

ChristopherDavenport:
Allows Discarding A Stream Value for Another.
Refs #800 

I am uncertain where you are at in preparing for 0.9.3, this was just a cherry pick of the commit in case you would like to incorporate it.

### DIFF
--- a/core/jvm/src/test/scala/fs2/StreamSpec.scala
+++ b/core/jvm/src/test/scala/fs2/StreamSpec.scala
@@ -35,6 +35,10 @@ class StreamSpec extends Fs2Spec {
       runLog(s.get.flatMap(inner => inner.get)) shouldBe { runLog(s.get).flatMap(inner => runLog(inner.get)) }
     }
 
+    ">>" in forAll { (s: PureStream[Int], s2: PureStream[Int] ) =>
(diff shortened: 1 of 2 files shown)


pchlupacek:
Are we sure this really makes good sense on stream? I have no problem with this to be on Pull, Task, hence these produce only one value, but having it on Stream, you essentially are creating constant stream of `B` values. To me this may lead to unexpected behaviours.

ChristopherDavenport:
Task[A].flatMap(_ => Task.delay(b))
if the initial task completes ignore the result and work with the Task[B] 

Stream[F,A].flatMap( _ => Stream.eval[F,B](F[b]))
foreach pull from Stream we get a value and we always discard that value to emit the SecondaryStream however many items it contains.

I feel as though any inconsistencies in behavior would be with people operating from the assumption of finite Streams. As each element of the Stream produces a Stream of values. A constant stream of `A` values produced by `F` effects to generate each `Stream[B]` emitted.

pchlupacek:
@ChristopherDavenport just bringing this a point. I have no general objective against that.

PR #809, opened 2017-02-11
Improved socket implementation to recycle buffer in Socket.readXX ope…

pchlupacek:
…rations

### DIFF
--- a/io/src/main/scala/fs2/io/tcp/Socket.scala
+++ b/io/src/main/scala/fs2/io/tcp/Socket.scala
@@ -210,34 +210,36 @@ protected[tcp] object Socket {
 
       // When the read operation is done, this will read up to buffer's position bytes from the buffer
       // this expects the buffer's position to be at bytes read + 1
-      // Additionally this releases the semaphore to allow for other read operation to take a place
       def releaseBuffer(buff: ByteBuffer): F[Chunk[Byte]] = {
         val read = buff.position()
-        if (read == 0) readSemaphore.increment >> F.pure(Chunk.bytes(Array.empty))
+        if (read == 0) F.pure(Chunk.bytes(Array.empty))
         else {
           val dest = Array.ofDim[Byte](read)
           buff.flip()
           buff.get(dest)
-          readSemaphore.increment >> F.pure(Chunk.bytes(dest))
+          F.pure(Chunk.bytes(dest))
         }
       }
 
       def read0(max:Int, timeout:Option[FiniteDuration]):F[Option[Chunk[Byte]]] = {
         readSemaphore.decrement >>
-        getBufferOf(max) flatMap { buff =>
+        F.attempt[Option[Chunk[Byte]]](getBufferOf(max) flatMap { buff =>
           readChunk(buff, timeout.map(_.toMillis).getOrElse(0l)) flatMap {
             case (read, _) =>
               if (read < 0) F.pure(None)
               else releaseBuffer(buff) map (Some(_))
           }
-        }
+        }).flatMap { r => readSemaphore.increment >> (r match {
+          case Left(err) => F.fail(err)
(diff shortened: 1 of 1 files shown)


pchlupacek:
@mpilquist this complains about the binary compatibility, but hence the change is not in api, I still believe its ok to publish it in 0.9.3 any thoughts?

mpilquist:
@pchlupacek Yeah that's fine -- can you add an exclusion to `mimaBinaryIssueFilters`?

io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -177,90 +177,128 @@ protected[tcp] object Socket {
...
-      def go(buff: ByteBuffer, timeoutMs:Long):F[Option[Chunk[Byte]]] = {
-        readChunk(buff,timeoutMs).flatMap { case (readBytes, took) =>
-          if (readBytes < 0 || buff.remaining() <= 0) F.pure(Some(Chunk.bytes(buff.array(), 0, buff.position())))
-          else go(buff,(timeoutMs - took) max 0)
+      def read0(max:Int, timeout:Option[FiniteDuration]):F[Option[Chunk[Byte]]] = {
+        readSemaphore.decrement >>
+        getBufferOf(max) flatMap { buff =>
+          readChunk(buff, timeout.map(_.toMillis).getOrElse(0l)) flatMap {
+            case (read, _) =>
+              if (read < 0) F.pure(None)
```
mpilquist: The `readSemaphore` isn't incremented here -- is that okay?

io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -177,90 +177,128 @@ protected[tcp] object Socket {
...
       }
-    }
 
-    def readN0(max:Int, timeout:Option[FiniteDuration]):F[Option[Chunk[Byte]]] = {
-      def go(buff: ByteBuffer, timeoutMs:Long):F[Option[Chunk[Byte]]] = {
-        readChunk(buff,timeoutMs).flatMap { case (readBytes, took) =>
-          if (readBytes < 0 || buff.remaining() <= 0) F.pure(Some(Chunk.bytes(buff.array(), 0, buff.position())))
-          else go(buff,(timeoutMs - took) max 0)
+      def read0(max:Int, timeout:Option[FiniteDuration]):F[Option[Chunk[Byte]]] = {
+        readSemaphore.decrement >>
```
mpilquist: Do we need any error handling here? E.g., if one of these tasks throws, we still want to increment the semaphore, right?

io/src/main/scala/fs2/io/tcp/Socket.scala
pchlupacek: thanks for spotting this

pchlupacek: but actually, hence this signals eof, i think we won't hit it,neverthelss I gonna fix it for corectness

io/src/main/scala/fs2/io/tcp/Socket.scala
pchlupacek: Well I think if you fail the read, there is no way to recover it anyhow, so my thought was just to let it be. Do you see there is a way that read will fail, and there is chance to recover it with next read?

pchlupacek:
@mpilquist made modifications as you have suggested. Not sure about this mimaBinaryIssueFilters where they are?

mpilquist:
@pchlupacek I took care of the mima issue: https://github.com/functional-streams-for-scala/fs2/commit/b02fd92556ef66d7216ceb8cb21fc0ec9e9e7e62

pchlupacek:
@mpilquist thanks

PR #811, opened 2017-02-13
Changed io.file.pulls API to return cancellable file handles

### LINKED ISSUE #772: io.file.pulls should provide constructors which allow early resource cleanup from within Pull
`fromFileChannel` and `fromAsynchronousFileChannel` both call `Pull.acquire`. Some usage patterns require use of `Pull.acquireCancellable` though (e.g., rolling files based on written size or time). We should add alternative constructors that use `Pull.acquireCancellable`.

mpilquist:
Fixes #772.

If we're good with this overall approach, TCP/UDP sockets should get a similar treatment. Specifically, we should *not* expose a `close` method on `FileHandle`/`Socket`/etc. and instead, return a `Cancellable[F,Resource]` from pulls.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -113,23 +113,38 @@ object Pull {
 
   private sealed trait AlgebraF[F[_],O] { type f[x] = Algebra[F,O,x] }
 
+  /** Result of `acquireCancellable`. */
+  trait Cancellable[+F[_],+R] {
+    val cancel: Pull[F,Nothing,Unit]
+    val resource: R
+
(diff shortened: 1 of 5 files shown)


gpampara:
Regarding the `FileHandle` usage, it cleans up the API I was using.

pchlupacek:
@mpilquist my first thought is that TCP sockets have 3 methods in this area (endOfInput, endOfOutput, close), unlike file that has only close. I think we can keep endOfXX like they are and remove close in favor of this pattern. 

Generally I don't see any reason to manually invoke close, as there is nothing you can do with socket after being closed.

mpilquist:
@pchlupacek Agreed on `endOfInput` / `endOfOutput`

PR #818, opened 2017-02-18
Removed a small typo

etorreborre:
I was just reading the README and saw this.


pchiusano:
Just noticed Tee is still mentioned in readme also, should be pipe2 now
right?

mpilquist:
Yep

PR #820, opened 2017-02-21
Updated for CrossVersion patch and scalaOrganization

ChristopherDavenport:
Not sure if necessary however I wanted to put in the PR since I saw this was running around. Referring to typelevel/scala#135

Not entirely sure if necessary but was easy enough to put out there if it is appropriate.


mpilquist:
Thanks!

PR #821, opened 2017-02-23
Reactive streams interop - WIP

zainab-ali:
This is a speculative PR for a ReactiveStreams interop.  At the moment, it can only convert a `Publisher` to a stream, but it might be a good foundation.

Should a reactive streams interop exist in fs2, or be a separate project?

This is a WIP, and hasn't been tested against the reactive streams spec yet.


pchlupacek:
@zainab-ali thanks for this PR. I think we would not be introducing dependency on reactive streams to fs2. However, this is solid foundation for some interop library with reactive stream, like they are for scalaz and cats, so perhaps if you can explore this path going forward ?

zainab-ali:
@pchlupacek agreed.  I'll start working on an interop library [here](https://github.com/to-ithaca/fs2-reactive)

PR #824, opened 2017-02-26
convert bytes to 0-255 Int for InputStream#read()

lucidd:
This fixed `fs2.io.toInputStream` so that [InputStream#read()](https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read()) only returns values of 0 to 255 and -1 for signalling the end.

### DIFF
--- a/io/src/main/scala/fs2/io/JavaInputOutputStream.scala
+++ b/io/src/main/scala/fs2/io/JavaInputOutputStream.scala
@@ -144,9 +144,9 @@ private[io] object JavaInputOutputStream {
 
       def go(acc:Array[Byte]):F[Int] = {
         F.flatMap(readOnce(acc,0,1,queue,dnState)) { read =>
-          if (read < 0) F.pure(read)
+          if (read < 0) F.pure(-1)
           else if (read == 0) go(acc)
-          else F.pure(acc(0).toInt)
+          else F.pure(acc(0) & 0xFF)
         }
       }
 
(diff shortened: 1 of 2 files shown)


io/src/main/scala/fs2/io/JavaInputOutputStream.scala
```diff
@@ -144,9 +144,9 @@ private[io] object JavaInputOutputStream {
 
       def go(acc:Array[Byte]):F[Int] = {
         F.flatMap(readOnce(acc,0,1,queue,dnState)) { read =>
-          if (read < 0) F.pure(read)
```
mpilquist: Why check explicitly for -1 instead of any negative?

lucidd: The `InputStream` documentation explicitly says that `-1` should be returned if there is no more data.
It would also be possible to do ` if (read < 0) F.pure(-1)` but i thought it would be clearer this way since  `readOnce` also explicitly uses `-1` so i would not expect it to return any other negative number here.

mpilquist: I'd prefer the `< 0` check to handle `InputStream` implementations that return other negative values. Per your point, that would be contrary to the docs but I think the risk/reward here favors `< 0`.

lucidd: i changed it back to `< 0`

PR #826, opened 2017-03-01
Update mention of 'Pull.awaitLimit'

Grogs:
It looks like this line was missed after 'awaitLimit' was moved onto Handle in functional-streams-for-scala/fs2#701.


mpilquist:
Thanks for catching this. `h.awaitLimit(n)(h)` should be `h.awaitLimit(n)` right? Also, could you update docs/src/guide.md and then run `project docs; +2.11.8; tut`?

Grogs:
Done. I have no idea why some of the line numbers tut outputs changed though....

PR #827, opened 2017-03-02
Adding scanAsync

yilinwei:
I've had this usecase quite a bit, but unsure how best to write it.

Feels as though if I do it this way, I'm probably adding a lot of overhead unnecessarily.


yilinwei:
Think this doesn't do what I want, since the effect should only be run once.

PR #831, opened 2017-03-18
Add scanF pipe

zainab-ali:
Adds a `scanF` pipe.  

For a `Stream[F, I]`, this has the signature `scan(z: O)(f: (O, I) => F[O]): Stream[F, I]`.

This allows scanning over a stream with a function which returns a Task.

### DIFF
--- a/core/shared/src/main/scala/fs2/pipe.scala
+++ b/core/shared/src/main/scala/fs2/pipe.scala
@@ -374,6 +374,25 @@ object pipe {
   def scan1[F[_],I](f: (I, I) => I): Pipe[F,I,I] =
     _ pull { _.receive1 { (o, h) => _scan0(o)(f)(h) }}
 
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/pipe.scala
```diff
@@ -374,6 +374,25 @@ object pipe {
   def scan1[F[_],I](f: (I, I) => I): Pipe[F,I,I] =
     _ pull { _.receive1 { (o, h) => _scan0(o)(f)(h) }}
 
+  private def _scanF0[F[_], O, I](z: O)(f: (O, I) => F[O]): Handle[F, I] => Pull[F, O, Handle[F, I]] =
+    h => h.await1Option.flatMap {
+      case Some((i, h)) => Stream.eval(f(z, i)).open.flatMap { _.await1.flatMap {
```
mpilquist: Can we do `Pull.eval(f(z, i)).flatMap { o => Pull.output(...)`

zainab-ali: Thanks, I've made the change

PR #835, opened 2017-03-25
Adding nonEmpty pipe to unwrap option values

yilinwei:


core/shared/src/main/scala/fs2/pipe.scala
```diff
@@ -336,6 +336,10 @@ object pipe {
       }
     }
 
+  /** Filters any 'None'. */
+  def nonEmpty[F[_], I]: Pipe[F, Option[I], I] =
```
mpilquist: Name wise, what do you think of `unNone` for symmetry with `unNoneTerminate`?

yilinwei:
@mpilquist Done.

core/shared/src/main/scala/fs2/pipe.scala
```diff
@@ -458,6 +458,9 @@ object pipe {
   def unchunk[F[_],I]: Pipe[F,I,I] =
     _ repeatPull { _.receive1 { case (i, h) => Pull.output1(i) as h }}
 
+  /** Filters any 'None'. */
+  def unNone[F[_], I]: Pipe[F, Option[I], I] = _.filter(_.isDefined).map(_.get)
```
mpilquist: Sorry, 1 more comment :)

This would be more efficient as `_.collect { case Some(i) => i }`

yilinwei:
@mpilquist Changed.

PR #836, opened 2017-03-27
Topic/leak join

### LINKED ISSUE #834: concurrent.join leaks memory
concurrent.join seems to leak memory, as following program will consume memory that retains in oldGen after manual GC. 

```scala 

object StreamsTestApp extends App {
  import Resources._

  def innerStream(idx:Int):Stream[Task, Unit] = {
    val allocate = "*"*10000
    Stream.eval(Task.delay(println(s"XXXG START $idx" + allocate.take(1))))
  }

  concurrent.join(Int.MaxValue)(
    time.sleep_[Task](10.seconds) ++
    Stream.range(0, 7000).map { i =>
      innerStream(i)
    } ++ Stream.eval_(Task.delay(println("XXXY DONE"))) ++
      time.sleep_[Task](1.hour)
  ).run.unsafeAttemptRun()


//  (
//    time.sleep_[Task](10.seconds) ++
//    Stream.range(0, 7000).flatMap { i =>
//      innerStream(i)
//    } ++ Stream.eval_(Task.delay(println("XXXY DONE"))) ++
//      time.sleep_[Task](1.hour)
//  ).run.unsafeAttemptRun()

}


```

After analysing through YourKit, the JVM seems to hold exactly 7000 references to stream in heap, together with their definition. 

The commented (sequential) variant is just working fine w/o leaking the memory. 

Any ideas of the source ?

pchlupacek:
@mpilquist @pchiusano @djspiewak this fixes #834. I'd like to make 0.9.5 with this fix today, if possible. Thanks

### DIFF
--- a/core/jvm/src/test/scala/fs2/ConcurrentSpec.scala
+++ b/core/jvm/src/test/scala/fs2/ConcurrentSpec.scala
@@ -27,7 +27,7 @@ class ConcurrentSpec extends Fs2Spec {
     }
 
     "join (1)" in forAll { (s1: PureStream[Int]) =>
-      runLog { concurrent.join(1)(s1.get.covary[Task].map(Stream.emit)) } shouldBe runLog { s1.get }
+      runLog { concurrent.join(1)(s1.get.covary[Task].map(Stream.emit)) }.toSet shouldBe runLog { s1.get }.toSet
     }
 
     "join (2)" in forAll { (s1: PureStream[Int], n: SmallPositive) =>
--- a/core/shared/src/main/scala/fs2/concurrent.scala
+++ b/core/shared/src/main/scala/fs2/concurrent.scala
@@ -39,59 +39,56 @@ object concurrent {
     * @param outer      Stream of streams to join.
     */
   def join[F[_],O](maxOpen: Int)(outer: Stream[F,Stream[F,O]])(implicit F: Async[F]): Stream[F,O] = {
-    assert(maxOpen > 0,"maxOpen must be > 0, was: " + maxOpen)
+    assert(maxOpen > 0, "maxOpen must be > 0, was: " + maxOpen)
 
-    def throttle[A](checkIfKilled: F[Boolean]): Pipe[F,Stream[F,A],Unit] = {
+    Stream.eval(async.signalOf(false)) flatMap { killSignal =>
+    Stream.eval(async.semaphore(maxOpen)) flatMap { available =>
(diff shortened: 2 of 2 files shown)


core/jvm/src/test/scala/fs2/ConcurrentSpec.scala
```diff
@@ -27,7 +27,7 @@ class ConcurrentSpec extends Fs2Spec {
     }
 
     "join (1)" in forAll { (s1: PureStream[Int]) =>
-      runLog { concurrent.join(1)(s1.get.covary[Task].map(Stream.emit)) } shouldBe runLog { s1.get }
+      runLog { concurrent.join(1)(s1.get.covary[Task].map(Stream.emit)) }.toSet shouldBe runLog { s1.get }.toSet
```
mpilquist: Why this change?

mpilquist:
Looks good to me though this is very much like the original version. There were subtle timing bugs in the original version. The current implementation tracked the inner streams running via resources (`Pull.acquireCancellable`) to prevent those leaks -- this was probably overkill. 

Given the importance of `concurrent.join`, I'm a little uncomfortable releasing 0.9.5 with a rewritten version without putting some more mileage on it. @pchlupacek -- can you run the join tests with 10x or 100x number of test cases?

pchlupacek:
@mpilquist yes. However if you find a way how to crack current implementation, feel free. I had trouble to find a way to do it, but perhaps as you suggest is maybe one-liner.

core/jvm/src/test/scala/fs2/ConcurrentSpec.scala
pchlupacek: sorry this is just overlook from my side. I did some experiments while implementing. Because maxopen is 1, the toSet is not necessary.

pchlupacek:
@mpilquist just runn full fs2.test scope 20 times, so far no issue. Will continue up to 100 times and lets see if that proves fine too.

pchlupacek:
Ok I have run 100 iteration of the test and all them passed

pchlupacek:
@mpilquist if you afraid about implication of this change how about publishing 0.9.4.1 for the while and have it under some stress in our production code? Then after week or so we may give it a clear go to 0.9.5 ?

mpilquist:
Any opinions on which PR we merge?

mpilquist:
@pchlupacek OK, let's merge this and then I'll release 0.9.5-RC1. If that looks good after a few days in your production stuff, we can release 0.9.5.

PR #837, opened 2017-03-27
Fixed memory leak in concurrent.join 

### LINKED ISSUE #834: concurrent.join leaks memory

mpilquist:
In the old implementation, the `doneQueue` was only processed when `open == maxOpen`. The new implementation drains the `doneQueue` eagerly, ensuring timely finalization, while also draining it after reaching the end of the outer stream.

Fixes #834. Alternate to #836.

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent.scala
+++ b/core/shared/src/main/scala/fs2/concurrent.scala
@@ -66,17 +66,56 @@ object concurrent {
         }
       }
 
-      def go(doneQueue: async.mutable.Queue[F,Pull[F,Nothing,Unit]])(open: Int): (Handle[F,Stream[F,A]], Handle[F,Pull[F,Nothing,Unit]]) => Pull[F,Nothing,Unit] = (h, d) => {
-        if (open < maxOpen)
-          h.receive1Option {
-            case Some((inner, h)) => runInnerStream(inner, doneQueue).flatMap { gate => go(doneQueue)(open + 1)(h, d) }
-            case None => Pull.done
+      def go(
+        doneQueue: async.mutable.Queue[F,Pull[F,Nothing,Unit]],
+        open: Int,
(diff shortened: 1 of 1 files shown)


mpilquist:
The old implementation was complicated and this fix makes it significantly worse. Not sure if we should merge this or go with @pchlupacek's new implementation of `join`, which is probably faster.

pchlupacek:
@mpilquist I am not quite sure that the amount of changes in this variant justifies the old code. I am not sure about the fact you indicating about performance benefit in my rewrite you indicating, but overall I think my rewritten version is easier to reason about.

PR #841, opened 2017-04-04
Removed Option from signature of await1Async - option was made redundant by NonEmptyChunk change

mpilquist:


pchlupacek:
+1. Just for curiosity, don't we have same situation elsewhere?

mpilquist:
Possibly -- I tried to catch all such things before we release 0.9 final. I definitely missed this one though.

pchlupacek:
ah ok

PR #842, opened 2017-04-04
Removed Strategy in favor of ExecutionContext

mpilquist:
The `Strategy` trait is basically the same as `scala.concurrent.ExecutionContext` -- the latter only adds a single method, `reportFailure`. Getting rid of `Strategy` has a number of benefits:
1) Better integration with Scala ecosystem
2) Better default implementations
3) Less concepts unique to FS2

Keeping it has no real advantages.


rossabaker:
I've recently been going back and forth on whether `Strategy` or `ExecutionContext` should be the primary user-facing concept in http4s, and wanted to discuss this unification.  I'm glad to see you got there first. :+1:

mpilquist:
By the way, I should point out that @alexandru's observations in https://github.com/typelevel/general/issues/66 and other places is what got me thinking about this.

alexandru:
This is good for interoperability 👍

pchiusano:
Don't care too much, but have some thoughts -

My only concern is that I think it is better if people are forced to think about what thread pool their FS2 code is running on. Being forced to construct a separate `Strategy` object achieves this... whereas I feel like a lot of code might just have some implicit `ExecutionContext` in scope for one reason or another, possibly for a COMPLETELY DIFFERENT PURPOSE. Like maybe they are using Scala futures in the same module... maybe they even have some blocking operations being done in that pool... you can see how using the same `ExecutionContext` type might be a bit too implicit and make it too easy to confuse which pool different code is running on.

I could also see the argument that this doesn't matter, just have everything running on 1 pool. In practice, that is often not the right choice.

OTOH, it's very possible that you have multiple uses of FS2 in your code and you want them to each have different thread pools... right now the only way to manage that is by altering the implicits that are in scope, which is just as error prone.

I think key question is whether we think it is likely that users will have FS2 code in a module that also uses an ExecutionContext for say working with `Futures` (or some other user of `ExecutionContext`), and we want to use different ExecutionContexts for FS2 usage vs the usage of `Future`. If we think that is a somewhat common case, then keeping `Strategy` separate makes sense. If we think that's a super rare case, then probably okay to get rid of `Strategy`.

Feels like the most explicit solution would be to get rid of implicit `Async` instance derived from implicit `Strategy` (or `ExecutionContext`). Users instead create a `Async` themselves by calling `Task.asyncFromExecutionContext` or something, and use that in their code (they can make the instance implicit at whatever scope they want). There's then no chance of magic-ing up the wrong `Async` tied to some random `ExecutionContext` that's in scope.

mpilquist:
@pchiusano I understand that rationale but feel it is too ad-hoc. If you extend that reasoning to the set of libraries used in a "typical" application, most or every library that deals with concurrency would have its own strategy-like interface. As an application developer, I want my libraries to interoperate when appropriate.

> Feels like the most explicit solution would be to get rid of implicit Async instance derived from implicit Strategy (or ExecutionContext).

I don't think this would solve anything. Rather, it would be like SIP-18 language imports -- some magic incantation that folks are told to copy/paste to make things work.

Overall, I think there's nothing *unique* about FS2 that warrants it having a newtype of execution context.

pchiusano:
> I don't think this would solve anything. Rather, it would be like SIP-18 language imports -- some magic incantation that folks are told to copy/paste to make things work.

Well, with explicit `Async` instances, you need to construct one, which forces you to make a decision about which `ExecutionContext` you want to give it. It won't just pull one in automatically from implicit scope. You have to make a decision. If you are going to argue that most people will make this decision by copying and pasting some code they don't understand, well, you could say that about any aspect of programming. The point is that the API forces you to be explicit, there's no "default" decision that will be made if the user takes no action.

I think the other arguments you made are stronger (some interop friction that's needless in many cases, and nothing really unique about FS2's usage of thread pools compared to other stuff).

There is always some tension about how you carve a system up into types... when types are more fine-grained, you get some additional safety and explicitness, at a cost of additional plumbing code (needing to convert between types that are "the same" at various boundaries). 

I don't have much of an opinion, you make the call. Am okay with merging... or not. Just raised these points because they weren't mentioned in your PR comments and wanted to make sure we thought about them! :)

alexandru:
Hi guys,

In my opinion execution contexts being passed implicitly around is fairly acceptable, because the [BlockingContext](http://www.scala-lang.org/api/current/scala/concurrent/BlockContext.html) while not very efficient, at least makes blocking I/O somewhat safe (with some caveats), but also because micro-managing thread-pools in complex applications becomes unproductive.

For instance in the application that I've been working on we have:

1. the application's global `ExecutionContext`, meant for CPU-bound tasks and backed by a `ForkJoinPool`
2. a special `ExecutionContext` meant for blocking I/O and backed by an unlimited Java `CachedExecutorService` ... basically blocking I/O can suffer from starvation if you have limits in your thread-pool; I actually ended up with deadlocks in my code because blocking stuff was keeping the thread-pool occupied
3. Netty's own thread-pool, because Netty always does that, keeping a couple of threads always busy and in a loop, probably reading from some sockets or whatever
4. [Slick's](slick.lightbend.com) `AsyncExecutor`, creating its own threads, but that I then reconfigured to use the official context for blocking I/O (point 2 above)
5. Akka's dispatcher, since we have been using actors for some of the communication; which I couldn't reconfigure to use the application's `global`, because Akka takes that configuration from `application.conf` and I didn't have the energy to workaround that
6. We also have had [ReactiveMongo](http://reactivemongo.org/)'s dispatcher, because it insisted on creating its own `ActorSystem` 😱 something which hurt my feelings and I pulled it out

Basically as an application ages, it acquires more and more thread-pools, which is pretty bad. Personally I have to actively fight this, I always take care of my thread-pools, but most people don't.

There's nothing wrong with reusing thread-pools. It's actually recommended. So what if FS2 Streams end up reusing the global EC meant for Futures? That's actually good, because blocking I/O stuff always needs special care, with operations interspersed with CPU-bound tasks, which should still be executed on the EC meant for CPU-bound stuff.

mpilquist:
Okay after thinking about this some more, I've decided to merge.

PR #847, opened 2017-04-17
Fixed stack safety of Task#attempt

mpilquist:


mpilquist:
BTW, this is a bug in 1.0 branch only (not 0.9), introduced when I combined `Task` and `Future` in to a single type.

mpilquist:
With this change, the following test fails:

```scala
Task.delay(throw Err).flatMap(x => Task.delay(x)).attempt.unsafeRun shouldBe Left(Err)
```

Hold off on merge until I find a stack safe way to fix.

mpilquist:
I'm going to close this and restore the 0.9 version of Task (based on Future). @pchiusano has an idea for an alternative representation of Task which takes advantage of the performance work he's doing.

PR #849, opened 2017-04-21
Added dependency on cats-core and cats-effect

mpilquist:
I've left `fs2.Task` in place for now and I've restored the 0.9 version. We'll see how the performance work goes before deciding what to do with `Task`. I'm still strongly in favor of removing it, but one step at a time.

Notable changes:
 - `Async` renamed to `Concurrent`
 - `Concurrent.mk` can provide a valid `Concurrent[F]` instance for any `F` given an `Effect[F]` and an `ExecutionContext`
 - Added `cats.Eq[A]` where appropriate
 - Ported tests to use `cats.effect.IO`

I didn't add cats type class instances everywhere given the work Paul's doing on performance. I'll make another pass later and add type class instances for Stream, Pull, etc.


pchlupacek:
@mpilquist we taking full cats-core? I was hoping only for cats-effects to depends on. I am really not sure if we shall depend on full cats.

pchlupacek:
I also don't think we have to port test to anything else than fs2.Task if fs2.Task is still included. We shall avoid having `F` in codebases that is virtually untested.

core/jvm/src/test/scala/fs2/TaskSpec.scala
```diff
@@ -13,26 +11,6 @@ class TaskSpec extends Fs2Spec{
       acc.toList shouldBe List(1, 2, 3, 4, 5)
     }
 
-    "Ref" - {
-
-      /**
-        * Tests whether set after access causes said access's
-        * set to default to no-op as the value in ref has changed
-        */
-      "Interleaving set and access " in {
```
mpilquist: This test was moved to the new ConcurrentSpec

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -695,12 +710,21 @@ object Stream {
 
   implicit def covaryPurePipe2[F[_],I,I2,O](p: Pipe2[Pure,I,I2,O]): Pipe2[F,I,I2,O] = pipe2.covary[F,I,I2,O](p)
 
-  implicit def streamCatchableInstance[F[_]]: Catchable[({ type λ[a] = Stream[F, a] })#λ] =
-    new Catchable[({ type λ[a] = Stream[F, a] })#λ] {
+  implicit def streamMonadErrorInstance[F[_]]: MonadError[Stream[F,?], Throwable] =
+    new MonadError[Stream[F,?], Throwable] {
```
mpilquist: No, the compiler plugin is not necessary for users -- it is compile time only. Users are free to use type lambda or type aliases.

mpilquist:
Re: using `IO` in the tests, I did that when I was certain that we'd be removing `Task`. Now that we aren't certain of that (though I feel pretty strongly still), I agree that we need more test coverage with it. The way I'd like to handle this is by writing an issue and addressing it pre-1.0 *if* we end up keeping `Task` around. Then I'd like to write some tests such that they are polymorphic in effect type and use both `IO` and `Task`. Not sure how well that will work but I think it's worth a shot.

build.sbt
```diff
@@ -168,7 +171,12 @@ lazy val root = project.in(file(".")).
 lazy val core = crossProject.in(file("core")).
   settings(commonSettings: _*).
   settings(
-    name := "fs2-core"
+    name := "fs2-core",
+    resolvers += Resolver.bintrayRepo("djspiewak", "maven"), // Temporary until cats-effect is published on Maven Central
+    libraryDependencies ++= Seq(
+      "org.typelevel" %%% "cats-core" % "0.9.0",
+      "cats-effect" %%% "cats-effect" % "0.1-650a34e"
```
djspiewak: cats-effect is now [published](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.typelevel%22%20AND%20a%3Acats-effect*) on Maven Central.

```sbt
"org.typelevel" %%% "cats-effect" % "0.1-d1b5231"
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -695,12 +707,21 @@ object Stream {
 
   implicit def covaryPurePipe2[F[_],I,I2,O](p: Pipe2[Pure,I,I2,O]): Pipe2[F,I,I2,O] = pipe2.covary[F,I,I2,O](p)
 
-  implicit def streamCatchableInstance[F[_]]: Catchable[({ type λ[a] = Stream[F, a] })#λ] =
-    new Catchable[({ type λ[a] = Stream[F, a] })#λ] {
+  implicit def streamMonadErrorInstance[F[_]]: MonadError[Stream[F,?], Throwable] =
```
djspiewak: Random thought: shouldn't `Stream[F, ?]` implement `Concurrent` for all `Concurrent[F]`?  By the same token, it could implement `Effect` for all `Effect[F]`, etc up the line.

core/shared/src/main/scala/fs2/util/Concurrent.scala
```diff
@@ -0,0 +1,299 @@
...
+    * continues to execute in the background though its result will be sent
+    * nowhere.
+   */
+  def race[A, B](fa: F[A], fb: F[B]): F[Either[A, B]]
+
+  /**
+   * Like `unsafeRunSync` but execution is shifted to the execution context used by this `Concurrent` instance.
+   * This method returns immediately after submitting execution to the execution context.
+   */
+  def unsafeRunAsync[A](fa: F[A])(f: Either[Throwable, A] => IO[Unit]): Unit
```
djspiewak: Is this just here for convenience?  If so, then you could give it a concrete implementation, since it can be implemented in terms of `runAsync`.

core/shared/src/main/scala/fs2/util/Concurrent.scala
```diff
@@ -0,0 +1,299 @@
...
+      def handleErrorWith[A](fa: F[A])(f: Throwable => F[A]): F[A] = F.handleErrorWith(fa)(f)
+      def raiseError[A](e: Throwable): F[A] = F.raiseError(e)
+      def async[A](k: (Either[Throwable, A] => Unit) => Unit): F[A] = F.async(k)
+      def runAsync[A](fa: F[A])(cb: Either[Throwable, A] => IO[Unit]): IO[Unit] = F.runAsync(fa)(cb)
+      def flatMap[A, B](fa: F[A])(f: A => F[B]): F[B] = F.flatMap(fa)(f)
+      def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B] = F.tailRecM(a)(f)
+      def liftIO[A](ioa: IO[A]): F[A] = F.liftIO(ioa)
+      def suspend[A](thunk: => F[A]): F[A] = F.suspend(thunk)
+      def unsafeRunAsync[A](fa: F[A])(f: Either[Throwable, A] => IO[Unit]): Unit =
+        F.runAsync(F.shift(fa)(ec))(f).unsafeRunSync
```
djspiewak: …like so.  :-)

core/shared/src/main/scala/fs2/util/Concurrent.scala
```diff
@@ -0,0 +1,299 @@
...
+              }
+              waiting = LinkedMap.empty
+            }
+            result = r
+
+          case Msg.TrySet(id, r, cb) =>
+            if (id == nonce) {
+              nonce += 1L; val id2 = nonce
+              waiting.values.foreach { cb =>
+                ec.executeThunk { cb((r: Either[Throwable, A]).map((_,id2))) }
```
djspiewak: I think I had this question previously…  But is this executing the callback multiple times?  `waiting` is a `LinkedMap`.

core/shared/src/main/scala/fs2/util/Concurrent.scala
```diff
@@ -0,0 +1,299 @@
...
+              // or the actor directly, and the winner destroys any
+              // references behind it!
+              if (won.compareAndSet(false, true)) {
+                val actor = ref.get
+                ref.set(null)
+                actor ! Msg.Set(res)
+              }
+            }
+            unsafeRunAsync(f1)(res => IO(win(res)))
+            unsafeRunAsync(f2)(res => IO(win(res)))
```
djspiewak: If either `f1` or `f2` is fully synchronous, then this function will not run them in parallel!  I think you either need to inject a `shift`, or make it clear in the documentation that the effects *may* be run simultaneously.

core/shared/src/main/scala/fs2/util/Concurrent.scala
```diff
@@ -0,0 +1,299 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  implicit def ioInstance(implicit ec: ExecutionContext): Concurrent[IO] = mk(IO.ioEffect)(ec)
```
djspiewak: Subjective question here: is it worth just straight-up making `mk` implicit?

core/shared/src/main/scala/fs2/util/Free.scala
```diff
@@ -11,7 +12,7 @@ sealed trait Free[+F[_],+A] {
   final def fold[F2[_],G[_],A2>:A](f: Fold[F2,G,A2])(implicit S: Sub1[F,F2], T: RealSupertype[A,A2]): G[A2] =
     this.step._fold(f)
 
-  final def translate[G[_]](u: F ~> G): Free[G,A] = {
+  final def translate[G[_]](u: UF1[F, G]): Free[G,A] = {
```
djspiewak: Future TODO: is `UF1` different from `FunctionK`?

core/shared/src/main/scala/fs2/util/UF1.scala
```diff
@@ -1,13 +1,13 @@
 package fs2.util
 
-/** A `forall a . f a -> g a`. Aliased as `F ~> G`. */
+/** A `forall a . f a -> g a`. */
 trait UF1[-F[_],+G[_]] { self =>
```
djspiewak: Oh, it's variant.  Got it!

docs/guide.md
```diff
@@ -108,17 +108,14 @@ val eff = Stream.eval(Task.delay { println("TASK BEING RUN!!"); 1 + 1 })
...
 
-The `eval` function works for any effect type, not just `Task`. FS2 does not care what effect type you use for your streams. You may use the included [`Task` type][Task] for effects or bring your own, just by implementing a few interfaces for your effect type ([`Catchable`][Catchable], [`Suspendable`][Suspendable], [`Effect`][Effect], and optionally [`Async`][Async] if you wish to use various concurrent operations discussed later). Here's the signature of `eval`:
+The `eval` function works for any effect type, not just `Task`. FS2 does not care what effect type you use for your streams. You may use the included [`Task` type][Task] for effects or bring your own, just by implementing a few interfaces for your effect type (`cats.effect.MonadError[?, Throwable]`, `cats.effect.Sync`, `cats.effect.Async`, `cats.effect.Effect`, and [`fs2.util.Concurrent`][Concurrent] if you wish to use various concurrent operations discussed later). Here's the signature of `eval`:
```
djspiewak: Might be worth mentioning `Concurrent.mk` here.

docs/guide.md
```diff
@@ -176,7 +173,7 @@ res15: Int = 2
 
 Here we finally see the tasks being executed. As is shown with `rc`, rerunning a task executes the entire computation again; nothing is cached for you automatically.
 
-_Note:_ The various `run*` functions aren't specialized to `Task` and work for any `F[_]` with an implicit `Catchable[F]`---FS2 needs to know how to catch errors that occur during evaluation of `F` effects.
+_Note:_ The various `run*` functions aren't specialized to `Task` and work for any `F[_]` with an implicit `MonadError[F, Throwable]`---FS2 needs to know how to catch errors that occur during evaluation of `F` effects.
```
djspiewak: Is it worth defining a type alias `type Catchable[F[_]] = MonadError[F, Throwable]` in `cats.effect`?

docs/guide.md
```diff
@@ -557,14 +554,14 @@ s.runLog.unsafeRun()
...
 
 ```scala
-import fs2.util.Async
-// import fs2.util.Async
+import cats.effect.Sync
+// import cats.effect.Sync
 
-val T = implicitly[Async[Task]]
-// T: fs2.util.Async[fs2.Task] = Async[Task]
+val T = implicitly[Sync[Task]]
```
djspiewak:
```scala
val T = Sync[Task]
```

docs/src/guide.md
```diff
@@ -80,17 +80,14 @@ val eff = Stream.eval(Task.delay { println("TASK BEING RUN!!"); 1 + 1 })
...
 
-The `eval` function works for any effect type, not just `Task`. FS2 does not care what effect type you use for your streams. You may use the included [`Task` type][Task] for effects or bring your own, just by implementing a few interfaces for your effect type ([`Catchable`][Catchable], [`Suspendable`][Suspendable], [`Effect`][Effect], and optionally [`Async`][Async] if you wish to use various concurrent operations discussed later). Here's the signature of `eval`:
+The `eval` function works for any effect type, not just `Task`. FS2 does not care what effect type you use for your streams. You may use the included [`Task` type][Task] for effects or bring your own, just by implementing a few interfaces for your effect type (`cats.effect.MonadError[?, Throwable]`, `cats.effect.Sync`, `cats.effect.Async`, `cats.effect.Effect`, and [`fs2.util.Concurrent`][Concurrent] if you wish to use various concurrent operations discussed later). Here's the signature of `eval`:
```
djspiewak: As above.

docs/src/guide.md
```diff
@@ -430,12 +427,12 @@ val s = Stream.eval_(Task.delay { destroyUniverse() }) ++ Stream("...moving on")
...
 
-The way you bring synchronous effects into your effect type may differ. [`Async.delay`](../core/shared/src/main/scala/fs2/util/Async.scala) can be used for this generally, without committing to a particular effect:
+The way you bring synchronous effects into your effect type may differ. `Sync.delay` can be used for this generally, without committing to a particular effect:
 
 ```tut:book
-import fs2.util.Async
+import cats.effect.Sync
 
-val T = implicitly[Async[Task]]
+val T = implicitly[Sync[Task]]
```
djspiewak: As above.

docs/src/guide.md
```diff
@@ -483,9 +480,10 @@ val c = new Connection {
   }
 }
 
-// recall T: Async[Task]
+// Effect extends both Sync and Async
+val T = implicitly[cats.effect.Effect[Task]]
```
djspiewak:
```scala
val T = cats.effect.Effect[Task]
```

core/shared/src/main/scala/fs2/util/Concurrent.scala
mpilquist: Both `f1` and `f2` are shifted to the EC captured by the `Concurrent` instance inside the implementation of `unsafeRunAsync`. Seems fully asynchronous to me but maybe I'm missing something subtle?

core/shared/src/main/scala/fs2/util/Concurrent.scala
mpilquist: Not that I see -- this is only executed once, on a successful set/tryset. The `waiting` map is then reset.

core/shared/src/main/scala/fs2/util/Concurrent.scala
mpilquist: Yeah, I'm not in love with the name -- it shifts the computation to the captured EC and then runs it. It's more than just convenience b/c we don't pass ECs throughout the Concurrency `carrying` functions.

An alternative design is making `Concurrency` a utility instead of a type class. Then we'd have to thread `Async[F]` and `ExecutionContext` through things.

core/shared/src/main/scala/fs2/util/Concurrent.scala
mpilquist: Done

docs/guide.md
mpilquist: Removed reference to `Concurrent` here.

docs/guide.md
mpilquist: IMO no, as I generally feel that type aliases add to the cognitive load needed to understand a library. They create a new name for an existing concept, requiring folks to know both names when puzzling through error messages. I'm fine with the type alias existing if others want it though.

pchlupacek:
@mpilquist was just wondering if with this change we couldn't tackle issue with Ref#set as well. I would like perhaps to emphasize that current `set` is forked, and returns earlier before actual set is performed. 
Thoughts?

mpilquist:
@pchlupacek Absolutely -- are you thinking we just rename `set` to `setAsync` and `setPure` to `setAsyncPure`?

pchlupacek:
@mpilquist yes, but not sure on naming tho. It is not really asynchronous operation. Also I am worid how many errors could be brought in scope by just changing so drastically `set` operation. 

So one option would be to drop `set` in favor of `put` (set that awaits confirmation) and  perhaps `shiftPut` that first shifts to another thread and then performs put (i.e current `set`) 

I am not that happy with `put` but having not better idea for now :-)

mpilquist:
@pchlupacek Hm, the existing `Ref#{set,setPure}` are truly asynchronous in this PR, just like they were in 0.9:

```scala
def set(t: F[A]): F[Unit] =
  F.liftIO(F.runAsync(F.shift(t)(ec)) { r => IO(actor ! Msg.Set(r)) })
```

Name-wise, I can handle renaming `set` and `setPure` to whatever names we collectively prefer. I'm also fine with adding `put` or something that does the set and waits for confirmation. I don't know if we should get rid of the async `set` though, as a lot of stuff is implemented in terms of it, and I don't want to subtly break that stuff. How about we discuss more on https://github.com/functional-streams-for-scala/fs2/issues/782 ?

pchlupacek:
@mpilquist ah this was not about getting rid of set (async). It is very usefull. I was not sure just by declaring the behaviour Asynchronous. If you are cool with it, then lets move on. 

So lets set on these `setAsync` for current `set` and `setAsyncPure` for current `setPure`. Then I would introduce their variants `setSync` and `setSyncPure` that awaits when the set is confirmed on the reference.

core/shared/src/main/scala/fs2/util/Concurrent.scala
djspiewak: As a sidebar, if it's just here for convenience, I would rather it be moved to a companion object or similar.  This is a bit of a subjective design decision, but I kind of like trying to keep the `unsafe` functions out of the typeclasses.

core/shared/src/main/scala/fs2/util/Concurrent.scala
djspiewak: Ok, I'm assuming I just wasn't following the logic correctly.  I was going off the types and some loose control flow analysis, and it seemed… odd.

core/shared/src/main/scala/fs2/util/Concurrent.scala
djspiewak: Didn't realize that `unsafeRunAsync` called `shift`.  That does indeed make this truly parallel.

djspiewak:
@mpilquist The question about whether or not `Concurrent.mk` should be implicit was actually just a question, not a request.  :-)  Do you think it's better to be explicit about which effect types support concurrency?  Or just make them *all* support it implicitly?  I can see arguments either way.

mpilquist:
@djspiewak Yeah, I've been back and forth on this one. I'm currently thinking that `Concurrent[F]` isn't a type class but rather a utility for an `Async[F]` given an `ExecutionContext`. Interested in thoughts on that. /cc @pchlupacek

pchlupacek:
@mpilquist @djspiewak Concurrent[F] to me is utility. In fact, we shall be always able to express it with cats.effects.* + EC. I would even consider to somehow hide it from API to not confuse users. That said we may perhaps going forward to remove all functionality to functions that just take cats.effects.* + EC and remove it from fs2 alltogether. 

However I would like to hold on that change until future of Task // IO in fs2 would be settled.

core/shared/src/main/scala/fs2/util/Concurrent.scala
mpilquist: @djspiewak Note that we can't move it to a companion b/c the execution context is captured in the `Concurrent` *instance*. I'll leave this here for now and then open a new PR after this one is merged that makes `Concurrent` a utility instead of a full type class.

mpilquist:
Alright, I'm merging this one and I'll open a few new issues to address some of the follow-up work, like deciding what to do with `Task`, making `Concurrent` a utility, and so on.

PR #850, opened 2017-04-24
clarifies nowrap argument for inflate / decompression

wookietreiber:
Reading the scaladoc, it is a bit irritating reading the word *compression* in the *inflate* aka *decompression* function. I suppose, this was introduced by copy-pasting.

---

> This is CONSISTENCY!
>
> -- Leonidas


wookietreiber:
Should probably be backported if there is another release in the 0.9 series.

PR #856, opened 2017-04-26
Changed Concurrent to be a utility instead of a type class

mpilquist:
`Concurrent` wasn't really a type class anymore -- it was just functions that used an `Effect[F]` and an `ExecutionContext`. This PR formalizes that by changing `Concurrent` to be a pure utility.

Merge after #854.


pchlupacek:
seems ok to me. Only thing I am worried here is that we have `concurrent` and  `Concurrent` now. How about having just `concurrent` as object ?

core/shared/src/main/scala/fs2/time/time.scala
```diff
@@ -46,7 +48,7 @@ package object time {
           val cancel = scheduler.scheduleAtFixedRate(d) {
             if (running.compareAndSet(false, true)) {
               val d = FiniteDuration(System.nanoTime, NANOSECONDS) - t0
-              F.unsafeRunAsync(signal.set(d))(_ => IO(running.set(false)))
+              Concurrent.unsafeRunAsync(signal.set(d))(_ => IO(running.set(false)))
```
pchlupacek: hm, we are referencing here IO directly ?

mpilquist:
> seems ok to me. Only thing I am worried here is that we have concurrent and Concurrent now. How about having just concurrent as object ?

Agreed. I'd like to move `fs2.concurrent.join` to the `Stream` companion, so usage would be `Stream.join(4)(streams)`. I'll update this PR if no objections.

core/shared/src/main/scala/fs2/time/time.scala
mpilquist: Yeah, sort of -- the `runAsync` method on `cats.effect.Effect` returns an `IO[Unit]` that kicks off the evaluation of `F`. https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Effect.scala#L35

mpilquist:
@pchlupacek OK that change is done. Do you think we should move `fs2.util.Concurrent` to `fs2` package? I'm leaning towards yes. Also, do you think we should name it like a mini-package (lowercase) or stick with standard scala naming conventions for objects?

pchlupacek:
@mpilquist yes to moving it to `concurrent`. And I would also leave join there for backward compatibility perhaps? 

I think with cats dependency, I would more look on util like `internal` perhaps? Or what utils we really want to expose to users now?

mpilquist:
@pchlupacek OK done -- `fs2.concurrent` it is, and I restored `concurrent.join` as a deprecated alias for `Stream.join`.

I hope to further whittle down the `util` package given some more time. Some stuff we won't be able to do without like `Sub1`, `Lub1`, `UF1` due to our use of variance.

pchlupacek:
this looks very nice @mpilquist lets merge once check gives green :-)

Issue #832, opened 2017-03-18
Debounce


yilinwei:
It would be useful to have some way to debounce a stream.

I think it ought to be a variant of a `Signal` rather than on the `Stream` itself.

mpilquist:
@yilinwei Could you elaborate a little on expected behavior?

yilinwei:
Yes.

fs2 currently has several tools when working across stream boundaries, `Signal` and `Queue`. However, it's a relatively common use case that you want to throttle the amount the updates to a reasonable rate. Adding extra facilities utility functions for that would be quite useful.

The marble diagram for [debounce](http://reactivex.io/documentation/operators/debounce.html) shows best.

Ideally, it would a a `Signal` which you would set and when calling `.discrete` debounced the updates.

zainab-ali:
The marble diagram seems to describe the following:

Let's say you have a `Stream[F, A]` and a debounce period `dt`.
If `a1` arrives at `t1` and no events arrive over `t1 + dt`, emit `a1`
If `a1` arrives at `t1` and `a2` arrives at `t2`, such that `t2 - t1 < dt`, ignore `a1`.

You could have a custom pipe which pulls from `Stream[F, A]`.  
 - On receiving an `a`, create an `F[Unit]` which emits after `dt`.
 - `awaitAsync` on the next handle and race the resulting future with `F[Unit]`.
 - If `F[Unit]` finishes first, emit `a` and repeat
 - If the next handle finishes first, create another `F[Unit]` and race again

I'm not sure if this is a good approach - it would be worth getting feedback on it.

yilinwei:
I'll give that approach a go next week and see how it's performance is.

yilinwei:
@zainab-ali  `race` is only on the `Task` and relies on `setRace` of the `Ref` of the task.

@mpilquist could it be moved to an arbitrary `Async[F]`? It doesn't seem a good idea to require people to implement it.

mpilquist:
Can we implement in terms of `merge` and `time.sleep` instead?

yilinwei:
I think so,

something like,

`time.sleep[Task](1 second).map(Left(_)).merge(Stream.eval(fa.map(Right(_)))).head`

should be technically the same as a race I think.

yilinwei:
Addressedin #861

PR #859, opened 2017-04-27
add debounce pipe

yilinwei:
#832

### DIFF
--- a/core/jvm/src/test/scala/fs2/time/TimeSpec.scala
+++ b/core/jvm/src/test/scala/fs2/time/TimeSpec.scala
@@ -63,5 +63,15 @@ class TimeSpec extends AsyncFs2Spec {
         case Vector(d) => assert(d >= delay)
       }
     }
+
+    "debounce" in {
+      val delay = 100 milliseconds
+      val s1 = Stream(1, 2, 3) ++ time.sleep[IO](delay * 2) ++ Stream() ++ Stream(4, 5) ++ time.sleep[IO](delay / 2) ++ Stream(6)
+      val t = s1.debounce(delay) runLog
+
+      t.unsafeToFuture() map { r =>
+        assert(r == Vector(3, 6))
+      }
+    }
   }
 }
--- a/core/shared/src/main/scala/fs2/pipe.scala
+++ b/core/shared/src/main/scala/fs2/pipe.scala
@@ -1,6 +1,7 @@
 package fs2
 
 import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.FiniteDuration
 
 import cats.{ Eq, Functor }
 import cats.effect.Effect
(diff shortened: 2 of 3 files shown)


core/shared/src/main/scala/fs2/pipe.scala
```diff
@@ -135,6 +136,28 @@ object pipe {
...
+      time.sleep[F](d).open.flatMap { h2 =>
+        h2.awaitAsync.flatMap { l =>
+          h1.awaitAsync.flatMap { r =>
+            (l race r).pull.flatMap {
+              case Left(_) => Pull.output1(i) >> r.pull.flatMap(identity).flatMap {
+                case (hd, tl) => go(hd.last, tl)
+              }
+              case Right(r) => r.optional.flatMap {
+                case Some((hd, tl)) => go(hd.last, tl)
+                case None => l.pull.flatMap(identity).flatMap { _ => Pull.output1(i) >> Pull.done }
```
yilinwei: I'm not entirely certain whether this should be emitted straight away or after the `time.sleep`.

yilinwei:
Seems to be a transient failure in a different [test](https://travis-ci.org/functional-streams-for-scala/fs2/jobs/226469397) on the Java 9 build.

core/shared/src/main/scala/fs2/pipe.scala
```diff
@@ -135,6 +136,28 @@ object pipe {
     }
   }
 
+  /** Debounce the stream with a minimum delay of `d` between each element */
+  def debounce[F[_], I](d: FiniteDuration)(implicit F: Effect[F], scheduler: Scheduler, ec: ExecutionContext): Pipe[F, I, I] = {
```
mpilquist: What do you think about implementing this with `in => pipe2.mergeHaltR(time.sleep[F](d), in).scan(...)`?

core/shared/src/main/scala/fs2/pipe.scala
```diff
@@ -135,6 +136,28 @@ object pipe {
     }
   }
 
+  /** Debounce the stream with a minimum delay of `d` between each element */
```
mpilquist: Nitpick: add period

core/shared/src/main/scala/fs2/pipe.scala
mpilquist: Add an alias in `Stream` too?

core/jvm/src/test/scala/fs2/time/TimeSpec.scala
```diff
@@ -63,5 +63,15 @@ class TimeSpec extends AsyncFs2Spec {
         case Vector(d) => assert(d >= delay)
       }
     }
+
+    "debounce" in {
+      val delay = 100 milliseconds
+      val s1 = Stream(1, 1, 1) ++ time.sleep[IO](delay * 2) ++ Stream(2, 2) ++ time.sleep[IO](delay / 2) ++ Stream(3)
```
mpilquist: To test that the last element is emitted, change to `Stream(1, 2, 3) ++ time.sleep[IO](delay * 2) ++ Stream(4, 5) ++ ... ++ Stream(6)` and assert result is `Vector(3, 6)`?

core/shared/src/main/scala/fs2/pipe.scala
yilinwei: Sure.

yilinwei: I'm very sorry, but I'm not too sure how this would work; the difficulty in the debounce seems to me to be in the fact that your `time.sleep` has to from the point of each new `A` which is emitted by `in` _and_ the ability to discard an element if the next one is emitted before the period.

Could you elaborate a little on what you mean?

core/shared/src/main/scala/fs2/pipe.scala
yilinwei: Sure.

core/shared/src/main/scala/fs2/pipe.scala
mpilquist: Ignore this, it didn't work out the way I was thinking. :+1: for this implementation.

yilinwei:
Mucked up the git history. Closed in favour of #861

PR #865, opened 2017-05-06
Make sure empty streams get deflated properly

mrdziuban:
The current behavior of `fs2.compress.deflate` is to skip deflation on empty streams. I would argue that this behavior is incorrect given the following example:

```scala
import java.io.ByteArrayOutputStream
import java.util.zip.{Deflater, DeflaterOutputStream}

val byteArrayStream = new ByteArrayOutputStream()
val deflater = new Deflater(Deflater.DEFAULT_COMPRESSION, true)
val deflaterStream = new DeflaterOutputStream(byteArrayStream, deflater)
deflaterStream.write(Array[Byte]())
deflaterStream.close()
byteArrayStream.toByteArray()
// => Array(3, 0)
```

These changes fix the behavior by making sure that `_deflate_finish` is called regardless of whether the stream was empty or not.


mpilquist:
:+1:

Issue #868, opened 2017-05-13
concurrent.join fails to propagate error of outer Stream 


pchlupacek:
This fails on 0.9.5
```scala
"join" - {

      "outer-failed" in {
        an[Err.type] should be thrownBy {
          (concurrent.join[Task, Unit](Int.MaxValue)(
            Stream(
              time.sleep_[Task](1 minute)
            ) ++ Stream.fail(new Throwable("Boom"))
          ) .run.unsafeRun()
        }
      }

    }
```

PR #869, opened 2017-05-13
Fixes Outer Stream in concurrent.join to fail join

pchlupacek:
@mpilquist I think this is not that critical fix, but perhaps we shall make 0.9.6? Not sure if anything else is on table for 0.9 branch

Shall I create another PR for 1.0.0 series?

### DIFF
--- a/core/jvm/src/test/scala/fs2/ConcurrentSpec.scala
+++ b/core/jvm/src/test/scala/fs2/ConcurrentSpec.scala
@@ -1,5 +1,7 @@
 package fs2
 
+import scala.concurrent.duration._
+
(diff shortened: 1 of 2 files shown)


pchlupacek:
fixes #868

mpilquist:
👍 We can release 0.9.6 (I plan for 0.9 to have fixes published for a long time due to some of the API changes in 1.0). Before doing so, I'd like to get #865 addressed too, but I've been very busy on 1.0 work.

mpilquist:
@pchlupacek Don't worry about 1.0 -- I just merged this there and I'm about to merge it to topic/fast.

Issue #876, opened 2017-05-23
Race condition failing outer stream in concurrent.join


fiadliel:
It seems like there is a race condition in `concurrent.join`:

if `runOuter` fails, then `decrementRunning` is called before enqueuing `Some(Left(failure))`. This will race (and maybe lose) against `doneMonitor` which will enqueue `None`.

As a result, a failure of the outer stream may not be detected.

@mpilquist WDYT?

pchlupacek:
@filadel you on 0.9.6 right?

fiadliel:
Yes, it's 0.9.6.

The current code is:
```scala
def runOuter: F[Unit] = {
  (outer.interruptWhen(killSignal) flatMap runInner onFinalize decrementRunning run).attempt flatMap {
    case Right(_) => F.pure(())
    case Left(err) => outputQ.enqueue1(Some(Left(err)))
  }
}
```

I think the issue is that `decrementRunning` is called before the enqueue. My first thought was something like this, but I don't think it's necessarily safe (e.g. `imap` on `Queue` could throw an exception):

```scala
def runOuter: F[Unit] = {
  (outer.interruptWhen(killSignal) flatMap runInner run).attempt flatMap {
    case Right(_) => F.pure(())
    case Left(err) => outputQ.enqueue1(Some(Left(err)))
  } flatMap { _ => decrementRunning }
}
```

pchlupacek:
@fiadliel you are right. I think I know how to fix this. Thanks for diagnosing it.

mpilquist:
@fiadliel Would you mind trying 0.9.0-SNAPSHOT now that #880 has been merged?

kalak451:
This: https://gist.github.com/kalak451/58701a62dc360984de719e8642c934ec  used to fail ~80% of the time in 0.9.6,  with a local build of 0.9.7-SNAPSHOT I can no longer get it to fail.

mpilquist:
@kalak451 Awesome, thanks very much for testing!

PR #880, opened 2017-05-31
Fixes the race condition when outer stream in concurrent.join failed

pchlupacek:


pchlupacek:
fixes #876

mpilquist:
Any reproduction case we can include in unit tests?

pchlupacek:
sadly didn't come with test case for this. The issue is that queue is internal, so the race may/may not occur. Not sure if creating test that does something 1000 time and may / may not fail is good here

PR #881, opened 2017-06-04
Recycle input buffer only in unsafeReadInputStream{,Async}

rossabaker:
The existing `readInputStream` recycles the input buffer.  This is akin to the behavior of scalaz-stream's `unsafeChunkR`, except it is not advertised as unsafe.

This passes the input buffer into `readInputStreamGeneric` in `F`.  `readInputStream` and `readInputStreamAsync` create a new array on each invocation, while `unsafeReadInputStream` and `unsafeReadInputStreamAsync` behave as the previous versions.

### DIFF
--- a/io/src/main/scala/fs2/io/io.scala
+++ b/io/src/main/scala/fs2/io/io.scala
@@ -17,8 +17,8 @@ package object io {
    *
    * Blocks the current thread.
    */
-  def readInputStream[F[_]: Sync](fis: F[InputStream], chunkSize: Int, closeAfterUse: Boolean = true): Stream[F, Byte] =
-    readInputStreamGeneric(fis, chunkSize, readBytesFromInputStream[F], closeAfterUse)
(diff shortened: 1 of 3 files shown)


io/src/main/scala/fs2/io/io.scala
```diff
@@ -31,7 +31,40 @@ package object io {
...
-    readInputStreamGeneric(fis, chunkSize, readAsync, closeAfterUse)
+    readInputStreamGeneric(fis, F.delay(new Array[Byte](chunkSize)), readAsync, closeAfterUse)
+  }
+
+  /**
+   * Reads all bytes from the specified `InputStream` with a buffer size of `chunkSize`.
+   * Set `closeAfterUse` to false if the `InputStream` should not be closed after use.
+   *
+   * Recycles an underlying input buffer for performance. It is safe to call
+   * this as long as whatever consumes this `Stream` stores the `Chunk` returned
```
mpilquist: Should this read "as long as whatever consumes this `Stream` *doesn't* store`

rossabaker: Eek! Yes.

Issue #891, opened 2017-07-04
Fix leak in Pipe.Stepper and Pipe2.Stepper


mpilquist:
The `Pipe.Stepper` and `Pipe2.Stepper` types allow stepping of pure pipes. We added these types in 0.9 as a replacement for the 0.8 pattern where a `Process1` was stepped via `feed1(a).unemit`. Scodec-protocols used that pattern extensively (e.g., https://github.com/scodec/scodec-protocols/blob/9888ebd0c5276269289174b6a7c92e853aa415bf/src/main/scala/scodec/protocols/process1ext.scala#L44-L60).

Unfortunately, the 0.9 stepper functionality is *much* slower than the old `feed1(a).unemit` pattern, due to how opaque pipes are in 0.9.

Nonetheless, I've ported the 0.9 stepper functionality to 0.10 on the topic/steppers branch (https://github.com/functional-streams-for-scala/fs2/blob/4d2f4bee0123d8e64cc6470130335dbba52aa315/core/shared/src/main/scala/fs2/Pipe.scala#L14-L46). It works but has a memory leak. To see the leak, run `coreJVM/test:run` and select the `StepperSanityTest2` test. Taking a heap dump soon after will show something like this:

<img width="885" alt="eclipse_memory_analyzer" src="https://user-images.githubusercontent.com/118078/27833160-888d5e3e-609f-11e7-8623-b6d312715275.png">

Both Paul and I have spent a bunch of time on the memory leak and we haven't been able to find the root cause. It seems like each tail stream that comes out of `uncons` accumulates a new `Bind` node which extends the `f` function by 1 lambda.

In general, I don't think there are good use cases for steppers, besides directly porting old `Process1` code. When working correctly, it's still an order of magnitude slower than the old `feed1(a).unemit` pattern from 0.8. Hence, I recommend any use of stepper be replaced with a different abstraction. For example, in scodec-protocols, I replaced `Pipe` stepping with a less opaque abstraction which is "compiled to" a `Pipe` after composition is complete: https://github.com/scodec/scodec-protocols/blob/49145b45e453f418f47cc4e669edc6bbd369b6fe/src/main/scala/scodec/protocols/Transform.scala

As a result, we're removing steppers from 0.10. I'll leave this issue open in case anyone wants to debug the memory leak.

mpilquist:
Fixed in #895

PR #895, opened 2017-07-12
Fixes memory leak in recursive streams

mpilquist:
This PR fixes a memory leak in `Stream#flatMap` when defining a recursive stream. Consider:

```scala
def loop: Stream[IO,Unit] = Stream(()).covary[IO].flatMap(_ => loop)
loop.run.unsafeRunSync
```

Before this PR, running this program would fill the heap within seconds. This was due to this code in `Stream#flatMap`:

```scala
hd.map(f).foldRightLazy(Stream.fromFreeC(tl).flatMap(f))(_ ++ _).get
```

Each element is mapped to a `Stream[F,O2]` and the resulting `Segment[Stream[F,O2],Unit]` is then lazily folded in to a `Stream[F,O2]` via the `++` operator.

Recursive streams of the form of `loop` emit a single element at a time, but the fold turns that in to `f(hd(0)) ++ Stream.empty.flatMap(f)`. This is equivalent to `f(hd(0))` but pushes the concatenation of `Stream.empty.flatMap(f)` in to a `Bind` node in the resulting `FreeC`. Hence, each iteration through the recursive `flatMap` lengthens the overall stream by an additional `Bind`.

This PR special cases `Stream#flatMap` for singleton streams in order to support this pattern. I think it's okay to special case singleton streams b/c the non-singleton case is expected to run in non-constant memory.

Note that this also fixes `Pipe.Stepper` and `Pipe2.Stepper` -- the `prompts` stream used in each had a recursive call inside `flatMap`.

### DIFF
--- a/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
+++ b/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
@@ -84,3 +84,8 @@ object ConstantEvalSanityTest extends App {
     }
   }) }.run.unsafeRunSync
 }
+
+object RecursiveFlatMapTest extends App {
+  def loop: Stream[IO,Unit] = Stream(()).covary[IO].flatMap(_ => loop)
+  loop.run.unsafeRunSync
+}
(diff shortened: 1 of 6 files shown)


mpilquist:
Fixes #891

pchiusano:
Did you happen to try bringing back viewL cache? Curious if the leaks we saw with that go away also.

pchiusano:
Lgtm - I think key is that non singleton case is expected to use more memory, so this is less of a hack than I was thinking initially. :)

mpilquist:
@pchiusano Even with this PR, the viewL cache still leaks (e.g., ConstantEvalSanityTest).

I know what you mean re: the hack. It took me a while to convince myself that special-casing singleton streams would be okay.

PR #896, opened 2017-07-12
Added toArray to Segment

mpilquist:


core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -835,6 +835,27 @@ abstract class Segment[+O,+R] { self =>
   }
 
   /**
+   * Converts this segment to an array, discarding the result.
+   *
```
pchlupacek: What exactly `discarding the result` means here ?

mpilquist: A `Segment[O,R]` outputs values of type `O` and when it finishes, outputs a single result of type `R`. Often `R` is `Unit`. When converting a segment to a standard collection type, we collect all the output `O` values and throw away the `R` value.

pchlupacek: Ah sorry. I am getting this. It was just unclear to me from the toArray signature what is lost. Thanks @mpilquist .

PR #899, opened 2017-07-14
Scheduler cleanup

mpilquist:
This PR is organized by individual commits in case there are objections to individual changes.

d3d6547 adds the ability to shutdown schedulers on the JVM. The `apply` method on `Scheduler` object returns a `Stream` that brackets the scheduler creation and shutdown. Alternatively, `Scheduler.allocate` can be used to manually control lifecycle of the scheduler (e.g., in unit tests).

e298149 defines the common Scheduler class in the shared directory.

67fa1ba gets rid of the `fs2.time` object and moves the methods to `fs2.Scheduler` class.


core/jvm/src/main/scala/fs2/internal/ThreadFactories.scala
```diff
@@ -0,0 +1,32 @@
+package fs2.internal
+
+import java.lang.Thread.UncaughtExceptionHandler
+import java.util.concurrent.{ Executors, ThreadFactory }
+import java.util.concurrent.atomic.AtomicInteger
+
+import scala.concurrent.ExecutionContext
+
+private[fs2] object ThreadFactories {
+  /** A `ThreadFactory` which creates daemon threads, using the given name. */
```
pchlupacek: do you think it is necessarry to make this private/internal. Was usefull in libraries so I couldn't repeat same boilerplate

mpilquist:
How about we leave it in the internal package but make it public? We can do the same for NonFatal.

pchlupacek:
@mpilquist completely ok with me.

core/shared/src/main/scala/fs2/Scheduler.scala
```diff
@@ -86,16 +97,27 @@ object time {
...
+      scheduleOnce(d) {
         ec.execute(() => cb(Right(())))
       }
       ()
     })
   }
 
   /** Identical to `sleep(d).drain`. */
-  def sleep_[F[_]](d: FiniteDuration)(implicit F: Effect[F], scheduler: Scheduler, ec: ExecutionContext): Stream[F, Nothing] =
+  def sleep_[F[_]](d: FiniteDuration)(implicit F: Effect[F], ec: ExecutionContext): Stream[F, Nothing] =
```
SystemFw: Sorry to meddle, but since you happen to be modifying this, you might want to replace `Effect` with `Async` here.

mpilquist:
@pchlupacek Done
@SystemFw Thanks for the reminder - fixed now.

PR #900, opened 2017-07-16
Add `retry` combinators 

SystemFw:
Once the Scheduler rework is merged, I will do the (very small) changes required to adapt this branch to the new Scheduler mechanism, but apart from that, this is ready to receive feedback I think.

The `delays` test has consistently succeeded on my machine, but let me know if you think there are risks of flakiness.

### DIFF
--- a/core/shared/src/main/scala/fs2/Scheduler.scala
+++ b/core/shared/src/main/scala/fs2/Scheduler.scala
@@ -114,6 +114,54 @@ abstract class Scheduler {
     sleep(d).drain
 
   /**
+   * Retries `fa` on failure, returning a single element
+   * stream with the result of `fa` as soon as it succeeds.
+   *
+   * @param delay Duration of delay before the first retry
+   *
+   * @param nextDelay Applied to the previous delay to compute the
+   *                  next, e.g. to implement exponential backoff
+   *
+   * @param maxRetries Number of attempts before failing with the
+   *                   latest error, if `fa` never succeeds
+   *
+   * @param retriable Function to determine whether a failure is
+   *                  retriable or not, defaults to retry every
+   *                  `NonFatal`. A failed stream is immediately
+   *                  returned when a non-retriable failure is
+   *                  encountered
+   */
+  def retry[F[_], A](fa: F[A],
+                     delay: FiniteDuration,
+                     nextDelay: FiniteDuration => FiniteDuration,
+                     maxRetries: Int,
+                     retriable: Throwable => Boolean = internal.NonFatal.apply)(
+    implicit F: Async[F], ec: ExecutionContext): Stream[F, A] = {
+     val delays = Stream.unfold(delay)(d => Some(d -> nextDelay(d))).covary[F]
+
+    attempts(fa, delays)
+      .take(maxRetries)
+      .takeThrough(_.fold(err => retriable(err), _ => false))
+      .last
+      .map(_.getOrElse(sys.error("[fs2] impossible: empty stream in retry")))
+      .rethrow
+  }
+
+  /**
+   * Retries `fa` on failure, returning a stream of attempts that can
+   * be manipulated with standard stream operations such as `take`,
+   * `collectFirst` and `interruptWhen`.
+   *
+   * Note: The resulting stream does *not* automatically halt at the
+   * first successful attempt. Also see `retry`.
+   */
+  def attempts[F[_], A](fa: F[A], delays: Stream[F, FiniteDuration])(
+    implicit F: Async[F], ec: ExecutionContext): Stream[F, Either[Throwable, A]] =
+    Stream.attemptEval(fa) ++ delays.flatMap { delay =>
+      sleep_(delay) ++ Stream.attemptEval(fa)
+    }
+
+  /**
    * Debounce the stream with a minimum period of `d` between each element.
    *
    * @example {{{
--- a/core/shared/src/main/scala/fs2/Scheduler.scala
+++ b/core/shared/src/main/scala/fs2/Scheduler.scala
@@ -114,8 +114,9 @@ abstract class Scheduler {
     sleep(d).drain
 
   /**
-   * Retries `fa` on failure, returning a single element
-   * stream with the result of `fa` as soon as it succeeds.
+   * Retries `s` on failure, returning a singleton stream with the
+   * result of `s` as soon as it succeeds. `s` must be a singleton
+   * stream.
    *
    * @param delay Duration of delay before the first retry
    *
@@ -131,7 +132,7 @@ abstract class Scheduler {
    *                  returned when a non-retriable failure is
    *                  encountered
    */
-  def retry[F[_], A](fa: F[A],
+  def retry[F[_], A](s: Stream[F, A],
                      delay: FiniteDuration,
                      nextDelay: FiniteDuration => FiniteDuration,
                      maxRetries: Int,
@@ -139,7 +140,7 @@ abstract class Scheduler {
     implicit F: Async[F], ec: ExecutionContext): Stream[F, A] = {
      val delays = Stream.unfold(delay)(d => Some(d -> nextDelay(d))).covary[F]
 
-    attempts(fa, delays)
+    attempts(s, delays)
       .take(maxRetries)
       .takeThrough(_.fold(err => retriable(err), _ => false))
       .last
@@ -148,18 +149,17 @@ abstract class Scheduler {
   }
 
   /**
-   * Retries `fa` on failure, returning a stream of attempts that can
+   * Retries `s` on failure, returning a stream of attempts that can
    * be manipulated with standard stream operations such as `take`,
-   * `collectFirst` and `interruptWhen`.
+   * `collectFirst` and `interruptWhen`. `s` must be a singleton
+   * stream.
    *
    * Note: The resulting stream does *not* automatically halt at the
    * first successful attempt. Also see `retry`.
    */
-  def attempts[F[_], A](fa: F[A], delays: Stream[F, FiniteDuration])(
+  def attempts[F[_], A](s: Stream[F, A], delays: Stream[F, FiniteDuration])(
     implicit F: Async[F], ec: ExecutionContext): Stream[F, Either[Throwable, A]] =
-    Stream.attemptEval(fa) ++ delays.flatMap { delay =>
-      sleep_(delay) ++ Stream.attemptEval(fa)
-    }
+    s.attempt ++ delays.flatMap(delay => sleep_(delay) ++ s.attempt)
 
   /**
    * Debounce the stream with a minimum period of `d` between each element.
--- a/core/shared/src/test/scala/fs2/RetrySpec.scala
+++ b/core/shared/src/test/scala/fs2/RetrySpec.scala
@@ -1,0 +1,93 @@
+package fs2
+
+import scala.concurrent.duration._
+import cats.effect.IO
+
+class RetrySpec extends AsyncFs2Spec {
+  case class RetryErr(msg: String = "") extends RuntimeException(msg)
+
+  "retry" - {
+    "immediate success" in {
+      var attempts = 0
+      def job: IO[String] = IO {
+        attempts += 1
+        "success"
+      }
+
+      runLogF(mkScheduler.flatMap(_.retry(job, 1.seconds, x => x, 100))).map { r =>
+        attempts shouldBe 1
+        r shouldBe Vector("success")
+      }
+    }
+
+    "eventual success" in {
+      var failures, successes = 0
+      def job: IO[String] = IO {
+        if (failures == 5) { successes += 1; "success" }
+        else { failures += 1; throw RetryErr() }
+      }
+
+      runLogF(mkScheduler.flatMap(_.retry(job, 100.millis, x => x, 100))).map { r =>
+        failures shouldBe 5
+        successes shouldBe 1
+        r shouldBe Vector("success")
+      }
+    }
+
+    "maxRetries" in {
+      var failures = 0
+      def job: IO[String] = IO {
+        failures += 1
+        throw RetryErr(failures.toString)
+      }
+
+      runLogF(mkScheduler.flatMap(_.retry(job, 100.millis, x => x, 5))).failed.map {
+        case RetryErr(msg) =>
+          failures shouldBe 5
+          msg shouldBe "5"
+      }
+    }
+
+    "fatal" in {
+      var failures, successes = 0
+      def job: IO[String] = IO {
+        if (failures == 5) { failures += 1; throw RetryErr("fatal") }
+        else if (failures > 5) { successes += 1; "success" }
+        else { failures += 1; throw RetryErr()}
+      }
+
+      val f: Throwable => Boolean = _.getMessage != "fatal"
+
+      runLogF(mkScheduler.flatMap(_.retry(job, 100.millis, x => x, 100, f)))
+        .failed
+        .map {
+          case RetryErr(msg) =>
+            failures shouldBe 6
+            successes shouldBe 0
+            msg shouldBe "fatal"
+        }
+    }
+
+    "delays" in {
+      val delays = scala.collection.mutable.ListBuffer.empty[Long]
+      val unit = 200
+      val maxTries = 5
+      def getDelays =
+        delays.sliding(2).map(s => (s.tail.head - s.head) / unit ).toList
+
+      def job: IO[Unit] = {
+        val start = System.currentTimeMillis()
+        IO {
+          delays += System.currentTimeMillis() - start
+          throw new Exception
+        }
+      }
+
+      runLogF(mkScheduler.flatMap(_.retry(job, unit.millis, _ + unit.millis, maxTries)))
+        .failed
+        .map { r =>
+          getDelays shouldBe List.range(1, maxTries)
+        }
+    }
+  }
+}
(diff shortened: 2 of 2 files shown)


core/shared/src/main/scala/fs2/retry.scala
```diff
@@ -0,0 +1,57 @@
...
+                     maxRetries: Int,
+                     retriable: Throwable => Boolean = NonFatal.apply)(
+    implicit F: Async[F], scheduler: Scheduler, ec: ExecutionContext): Stream[F, A] = {
+     val delays = Stream.unfold(delay)(d => Some(d -> nextDelay(d))).covary[F]
+
+    retries(fa, delays)
+      .take(maxRetries)
+      .takeThrough(_.fold(err => retriable(err), _ => false))
+      .last
+      .map(_.get) // `retries` is guaranteed to have at least one element
```
wedens: I'd rather do something like `.getOrElse(sys.error("[fs2] impossible: empty stream in `retry`"))`

core/shared/src/main/scala/fs2/retry.scala
```diff
@@ -0,0 +1,57 @@
+package fs2
+
+import cats.effect.Async
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.FiniteDuration
+import scala.util.control.NonFatal
```
mpilquist: Change to `fs2.internal.NonFatal`?

core/shared/src/main/scala/fs2/retry.scala
```diff
@@ -0,0 +1,57 @@
...
+      .map(_.get) // `retries` is guaranteed to have at least one element
+      .rethrow
+  }
+
+  /**
+   * Retries `fa` on failure, returning a stream of attempts that can
+   * be manipulated with standard stream operations such as `take`,
+   * `collectFirst` and `interruptWhen`.
+   *
+   * Note: The resulting stream does *not* automatically halts at the
```
mpilquist: s/halts/halt/

core/shared/src/main/scala/fs2/retry.scala
```diff
@@ -0,0 +1,57 @@
...
+    retries(fa, delays)
+      .take(maxRetries)
+      .takeThrough(_.fold(err => retriable(err), _ => false))
+      .last
+      .map(_.get) // `retries` is guaranteed to have at least one element
+      .rethrow
+  }
+
+  /**
+   * Retries `fa` on failure, returning a stream of attempts that can
```
mpilquist: WDYT about calling this method `attempts` instead of `retries`? Something that makes it clear that the result isn't consulted.

core/shared/src/main/scala/fs2/retry.scala
```diff
@@ -0,0 +1,57 @@
+package fs2
+
+import cats.effect.Async
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.FiniteDuration
+import scala.util.control.NonFatal
+
+object retry {
```
mpilquist: WDYT about moving this method to `Scheduler` class?

core/shared/src/main/scala/fs2/retry.scala
```diff
@@ -0,0 +1,57 @@
...
+   *                   latest error, if `fa` never succeeds
+   *
+   * @param retriable Function to determine whether a failure is
+   *                  retriable or not, defaults to retry every
+   *                  `NonFatal`. A failed stream is immediately
+   *                  returned when a non-retriable failure is
+   *                  encountered
+   */
+  def retry[F[_], A](fa: F[A],
+                     delay: FiniteDuration,
```
pchlupacek: I am not so sure by this retry combinator. It has a lot of parameters and seems to me quite heavy. 
But the main objection I have is  the fact that `f` here cannot have safe resource allocation we should not rather instead of `F` use Stream[F, A]. You can always do Stream.eval() if its only `F` you need to evaluate. 

May I suggest alternative signature here, that will be on Stream[F, A]: 

```scala 

/** runs this stream until `f` returns None. Otherwise waits given timeout before retrying. Int has number of attempts since last failure. Emits failure on left when fails.  **/
def retryWhen(f: (Throwable, Int) => F[Option[FiniteDuration]] ) : Stream[F, Either[Throwable, A]] = ??? 

def retryAlways(pause: FiniteDuration): Stream[F, Either[Throwable, A]] = 
  retryWhen { _ => F.pure(Some(period)) }

```

core/shared/src/main/scala/fs2/retry.scala
SystemFw: Of course :)

core/shared/src/main/scala/fs2/retry.scala
SystemFw: Yes, I wasn't really sure about the names, `attempts` sounds better

core/shared/src/main/scala/fs2/retry.scala
SystemFw: I would agree, `retry.retry` is a bit ugly

core/shared/src/main/scala/fs2/retry.scala
SystemFw: Thanks for the feedback :) 
A few comments:
- `F` vs `Stream`: the counterargument is that (I think) you would need to guarantee that a stream only returns one result, which didn't seem super clean.
- `retryWhen`: I actually started with a simpler recursive method using `flatMap` (as opposed to combinators over an infinite stream), along the lines of:
```scala
def retryWithDelay[F[_], A](
    fa: F[A],
    delay: FiniteDuration,
    nextDelay: FiniteDuration => FiniteDuration,
    maxRetries: Int
  )(implicit sch: Scheduler, F: Async[F]): Stream[F, A] =
    Stream.attemptEval(fa).flatMap {
      case Left(ex) if maxRetries <= 0 => Stream.fail(ex)
      case Left(ex)  => fs2.time.sleep_(delay) ++ retryWithDelay(fa, nextDelay(delay), nextDelay, maxRetries.map(_ - 1))
      case Right(a) => Stream.emit(a)
    }
```
 and you seem to suggest a similar approach (correct me if I'm wrong). The reason I moved away from it is that standard `Stream` combinators didn't seem to work (e.g. `take`), which I think is counterintuitive.
- Parameters: They roughly mirror the existing `retry` in Scala.Task (which several people have asked for a replacement of). Definitely open to a discussion over those though :)

wedens: Currently I use the following implementation exactly because I need safe resource allocation and standard combinators (especially `interruptWhen`). Also I need to do things like "log every 3rd failure".
```scala
    /**
      * Retries `fa` if it fails.
      * Returns a single (the first) value produced by `fa` on a successful
      * attempt or fails with the last error if `nextDelay` results in `None`.
      *
      * @param initialDelay delay after the first attempt
      * @param nextDelay function used to compute consequent delays
      *                  `None` terminates attempts
      * @param onFailedAttempt can be used to handle (but not recover) attempt failure
      * @return the first value produced by `fa`
      */
    def retryWithDelay[F[_]: Async, A](
      fa: Stream[F, A],
      initialDelay: FiniteDuration,
      nextDelay: (Int, FiniteDuration) => Option[FiniteDuration],
      onFailedAttempt: (Int, Throwable) => Stream[F, Nothing]
    )(implicit sch: Scheduler): Stream[F, A] =
      fa.attempt.repeat.mapAccumulate((1, initialDelay)) {
        case (t@(attempts, delay), Left(ex)) =>
          nextDelay(attempts + 1, delay) match {
            case Some(ndelay) =>
              val s = onFailedAttempt(attempts, ex) ++ fs2.time.sleep_[F](delay)
              ((attempts + 1, ndelay), s)
            case None =>
              (t, Stream.fail(ex))
          }
        case (t, Right(a)) => (t, Stream.emit(a))
      }.flatMap(_._2).head
```

SystemFw: Let's set  aside the `Stream` vs `F` debate for a second: it's a very small implementation change), and I do hear yours and @pchlupacek arguments about resource safety, so definitely open to change that bit.

As for the API, my idea was to provide a very simple, but powerful and compositional method  for people to build upon (that would be `retries/attempts`): I 100% agree that ability to work with standard combinators is paramount, and imo compositionality is the best feature of fs2 so I wouldn't want to lose that.

As for `retry`, I also wanted to provide people with an almost drop-in replacement to scalaz.Task.retry, hence the specific API. Would it be possible to build your `retryWithDelay` on top of attempts? I haven't tried, but my gut feeling is that it would.

wedens: API is already different from `z.c.Task.retry`. It produces a `Stream[F, A]`, not `F[A]` ;)

I think it's possible to implement many things on top of `attempts` as it is fairly low level.

SystemFw:
@mpilquist Addressed initial code review feedback

core/shared/src/main/scala/fs2/Scheduler.scala
```diff
@@ -114,6 +114,54 @@ abstract class Scheduler {
...
+   * @param maxRetries Number of attempts before failing with the
+   *                   latest error, if `fa` never succeeds
+   *
+   * @param retriable Function to determine whether a failure is
+   *                  retriable or not, defaults to retry every
+   *                  `NonFatal`. A failed stream is immediately
+   *                  returned when a non-retriable failure is
+   *                  encountered
+   */
+  def retry[F[_], A](fa: F[A],
```
mpilquist: Based on where the conversation in PR left off, I think we should change `fa` to `Stream[F,A]` here and then maybe add an overload for `F[A]`.

SystemFw: @mpilquist Should I add a comment saying that it expects a singleton Stream?

mpilquist: Yeah that sounds good.

SystemFw:
The requested changes have been done

pchlupacek:
@SystemFw @mpilquist I now this has merged already. I was sort of hoping it will hold until my review will be closed (still open though). Anyhow I still have problem with the signature, which is not composable and too complex IMHO.  

I really don't like that  retryable is not wrapped in `F`, at least. and I really do not thing that `nextdelay`, `delay` and `maxreteries` shall be separate parameters again, without `F` being involved here to allow some more complex computations. 

I would kindly ask you guys to reconsider this one. And perhaps revert it to more composable signature.

pchlupacek:
Also a small remark. I don't think so `s` must be an singleton. Really I would like to keep this signature similar to Attempt, allowing to return multiple values. If anyone wants to use with singleton stream fine, but that restriction is not necessary.

SystemFw:
@pchlupacek Do you consider `attempts` composable enough? My reasoning was to provide a composable primitive for complex computations, and an easy one for the most common scenario.
Otherwise you would have a `(Throwable, Int, FiniteDuration) => F[Option[FiniteDuration]]`, which is imo clunky to write, and would repeat the logic in `nextDelay` and `maxRetries` most of the time.

mpilquist:
Sorry for eager merge - let's find something we can all agree on. We can have both a lower level API and higher level API.

SystemFw:
Just for the record, I'm willing to do all the necessary changes until we find something we can all agree on :)

Re: not having singleton streams, what's the semantics you imagine @pchlupacek ? Having some elements of a partially succeeded stream (in a Right), then `Left(Exception)`, then delay, then elements of the second Stream? I believe you can do this with `attempts` right now, but I'm not sure how do you then manipulate the resulting stream in a useful way, so I'm genuinely curious what sort of semantics you envision

pchlupacek:
@SystemFw Yes, I agree the semantics shall be very same like we have for `attempt` now, with the added delay that retry adds. 

In fact I figured out we have Stream.attemptEval as well, so perhaps we shall make two variants of retry, one that is on the stream and one that is on the F[_]. 

That having said I am thinking of something like this:

```scala

// Retries current stream, until f, returns Some(). I believe you can express all situations like in proposed API by supplied (potentially) effectfull function. Note that F[Option[FiniteDuration]] may be backed by some external signal to allow you consult various comple states before deciding if retry shall / shall not be the case. 
def retry(f:(Throwable, Int) => F[Option[FiniteDuration]] ): Stream[F, Attempt[A]]

// Stream.retryEval
def retryEval[F[_]](f: F[A])(f: (Throwable, Int) => F[Option[FiniteDuration]]) :  Stream[F, Attempt[A]]

// then if you think that (Throwable, Int) => F[Option[FiniteDuration]] is complex to write, we may ///introduce simple combinators on some object (i.e. retry)

// retry infinitelly with supplied delay 
def infinite[F[_]](delay: FiniteDuration))(implicit F: Applicative[F]):  (Throwable, Int) => F[Option[FiniteDuration]] = 
 (_, _) => F.pure(Some(delay))

// incremental delay, not sure if it is usefull tho.... as delay may be potentially infinite
def incremental[F[_] : Applicative](
 initial: FiniteDuration
 , increments: FiniteDuration
 , retryIf: Throwable => F[Boolean]
): (Throwable, Int) => F[Option[FiniteDuration]] = { (rsn, attempts) =>
 retryIf(rsn) map { shallRetry =>
   if (shallRetry) Some(initial + (attempts*increments))
   else None
 }
} 

// that could give us syntax like 

val s: Stream[F, A] = ???
import retry._
s.retry infinite(1 second)
s.retry incremental(1.second, 100.milllis, rsn => F.pure(rsn.nonFatal))



```

pchlupacek:
As for the manipulation of the resulting stream, I think there is nothing really special here. You know that when Left is coming that you will receive again elements sooner or later, so you may use for example mapAccumulate to keep track of the state if you really need to.

pchlupacek:
Note that suggested syntax can be perhaps generalized even further takind delay off the signature completely, and perfrom delay only in scope of `F` which is perhaps usefull. 

i.e. 

```scala 

def retry(f: (Throwable, Int) => F[Boolean]): Stream[F, AttemptA] =???

then 

def infinite(delay: FiniteDuration)(implicit S: Scheduler, F: Async[F]): (Throwable, Int) => F[Boolean] = 
   ??? // perform delayed execution with scheduler and then complete asynchronous F to yield true

```

wedens:
The most general signature will be `def retry(f: (Throwable, Int, Option[FiniteDuration]) => Stream[F, Option[FiniteDuration]]): Stream[F, ?]` 😄

pchlupacek:
@wedens you don't need finiteduration in the scope of retry. Also there is no need to wrap that in stream, as the `f` will always produce only once and will live only short time (resources need to be released)

pchlupacek:
@wedens having said that, perhaps you can find something like pipe2, that will work ... I.e. creating a pipe that will consule from the left, and when left fails then it will consult right what to do next.

pchlupacek:
i.e. 
``` scala

def retry(f: Throwable => F[Boolean]): Pipe2[F, A, FiniteDuration] = ???

//then you can perhaps use it as follows
stream through retry(nonFatal) awakeEvery(1s)  


```

wedens:
I'll just list some features I'd like to be possible to express:
- decide whether `Thowable` is retriable
- calculate next delay based on previous delay and/or attempt number
- infinite retires
- max number of retries
- do something on failed attempt (e.g. log every third attempt) and continue retries
- be able to use `interruptWhen` to stop retries

pchlupacek:
@wedens I think you can express them all with all suggested signatures above. Is more about experimenting how easily different syntaxes may be used...

wedens:
Having said that, perhaps fs2 should have some basic implementation to allow most common use cases. And if you need something more convoluted, you can always roll your own version tailored to specific case.

For example, I use:
```scala
    def retryWithDelay[F[_]: Async, A](
      fa: Stream[F, A],
      initialDelay: FiniteDuration,
      nextDelay: (Int, FiniteDuration, Throwable) => Option[FiniteDuration],
      onFailedAttempt: (Int, Throwable) => F[Unit]
    )(implicit sch: Scheduler): Stream[F, A] =
      fa.attempt.repeat.mapAccumulate((1, initialDelay)) {
        case (t@(attempts, delay), Left(ex)) =>
          nextDelay(attempts + 1, delay, ex) match {
            case Some(ndelay) =>
              val s = Stream.eval_(onFailedAttempt(attempts, ex)) ++ fs2.time.sleep_[F](delay)
              ((attempts + 1, ndelay), s)
            case None =>
              (t, Stream.fail(ex))
          }
        case (t, Right(a)) => (t, Stream.emit(a))
      }.flatMap(_._2).head
```

But I don't think it's something you'd like to have in a library ;)

pchlupacek:
@wedens @SystemFw More I am thinking of it we should have implementation that DOES NOT have baked-in delay. I think hence there is F[] already in signature, we could do much more flexible stuff, like retry on signal not just after somehow computed delay. Then we could write some helpers producing (Throwable, Int) => F[Boolean] that will do whatever retry and timeout logic you need.

SystemFw:
Ok, I'll try to respond to everything.
First and foremost, the main criticism from @pchlupacek is that `retry` is not generic and composable enough: that's by design. `retry` is the special, common case. It's `attempts` which is the generic and composable one, and it hasn't been mentioned once in any comment.

> I agree the semantics should be the same we have for attempt now, with the added delay that retry adds.

This is what `attempts` already does. You can then manipulate the stream however you please.

>  More I am thinking of it we should have implementation that DOES NOT have baked-in delay.

I don't see how this could work. It's not just _when_ to delay, but also _how long_, which requires `sleep`, which requires a `FiniteDuration`. I agree that in the general case you don't want constant `FiniteDuration`, but `attempts` takes a `Stream[F, FiniteDuration]`, so you can decide when to emit the next delay by manipulating this stream however you please. How do you delay a single `F`, in fs2? I think that would require Stream.

I also have questions on your proposed implementation, but I'll ask those in your new PR

SystemFw:
@wedens I believe all the features you have listed can be implemented in terms of `attempts` right now.

PR #903, opened 2017-07-21
Fixed a bug in Resources#newSince

### LINKED ISSUE #902: [0.9.7] take() flips order of finalizers
```scala
var o: Vector[Int] = Vector.empty
val str = (Stream.emit(()) ++ Stream.fail(new RuntimeException("r")))
  .onFinalize(Task.delay(o = o :+ 1))
  .onFinalize(Task.delay(o = o :+ 2))
    
str.run.unsafeAttemptRun()
o shouldEqual Vector(1, 2) //Passes

o = Vector.empty
str.take(1).run.unsafeAttemptRun()
o shouldEqual Vector(1, 2) //Fails
```
Got bitten by this when outer stream allocates a thread pool. Inner stream can't release its resources because the thread pool it uses is already closed.

In 0.10.0-M4 it works as expected.

avakhrenev:
Now it actually returns newest tokens first.

Fixes #902

### DIFF
--- a/core/jvm/src/test/scala/fs2/ResourceSafetySpec.scala
+++ b/core/jvm/src/test/scala/fs2/ResourceSafetySpec.scala
@@ -184,6 +184,37 @@ class ResourceSafetySpec extends Fs2Spec with EventuallySupport {
       s.unsafeRun
       s.unsafeRun
     }
+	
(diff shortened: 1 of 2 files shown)


mpilquist:
Awesome!

PR #909, opened 2017-07-29
Update dependencies

aeons:
Including `cats-effect-0.4` which brings `cats-1.0-MF`, `sbt-0.13.16` and `scala-2.12.3`.


mpilquist:
Thanks!

PR #910, opened 2017-07-31
Fixes #907

### LINKED ISSUE #907: Signal regression from 0.9 -> 0.10: only first item emitted on `repeat` source.
The following chunk of code only results in a single item emitted on 0.10.0-M4 (also tested on M2).

```
import fs2._
import cats.effect.IO

def log[A](prefix : String) : Pipe[IO,A,A] = { in =>
  in.evalMap(i => IO { println(prefix + ">>" + i.toString) ; i} )
}

case class Price(key : Char, price : Double)
object Price {
  def random : IO[Price] = IO {
    val key = 'A' + (scala.math.random() * 10.0).toInt toChar
    val price = scala.math.random()
    Price(key, price)
  }
}

val in : Stream[IO,Price] = (Stream eval Price.random repeat) through log("in")

val sig : Stream[IO, async.immutable.Signal[IO,Option[Price]]] =
  async.immutable.Signal.holdOption(in)

val out : Stream[IO,Option[Price]] =
  sig flatMap (_.discrete) take 20 through log("out")

out.runLog.unsafeRunTimed(1 second)
```

I would expect this to emit the 20 items. It only outputs:
```
in>>Price(H,0.30250069403636715)
out>>None
out>>Some(Price(H,0.30250069403636715))
```

On 0.9.7, with the following code:

```
import fs2._

implicit def strat = Strategy.fromCachedDaemonPool()
def log[A](prefix : String) : Pipe[Task,A,A] =
  _.evalMap(i => Task.delay{ println(prefix + ">>" + i.toString) ; i } )

case class Price(key : Char, price : Double)
object Price {
  def random : Task[Price] = Task {
    val key = 'A' + (scala.math.random() * 10.0).toInt toChar
    val price = scala.math.random()
    Price(key, price)
  }
}

val in : Stream[Task,Price] = (Stream eval Price.random repeat) through log("in")

val sig : Stream[Task, async.immutable.Signal[Task,Option[Price]]] =
  async.immutable.Signal.holdOption(in)

val out : Stream[Task,Option[Price]] =
  sig flatMap (_.discrete) take 20 through log("out")

out.runLog.unsafeRun()
```
This prints out multiple `in`s and the 20 `out`s (can't see the output - it's cut off, sample below):

```
in>>Price(A,0.32971324020042525)
in>>Price(D,0.29919574847430286)
in>>Price(H,0.9111423166252853)
in>>Price(J,0.5500450896250947)
in>>Price(F,0.22712762271018438)
in>>Price(H,0.44739601322878864)
in>>Price(B,0.7649178222004639)
in>>Price(I,0.08453299145449888)
in>>Price(B,0.9298880933223229)
in>>Price(C,0.8484940029373574)
in>>Price(I,0.9747919045403481)
in>>Price(H,0.14794885880903685)
in>>Price(H,0.27389323535519183)
in>>Price(C,0.23817026370678984)
in>>Price(D,0.2903977744468541)
in>>Price(A,0.6209111201757896)
in>>Price(J,0.7897043931941421)
in>>Price(A,0.7167284670123723)
in>>Price(D,0.3363488372388195)
in>>Price(H,0.5082710285336824)
in>>Price(G,0.8791137829005642)
in>>Price(D,0.63527340408363)
out>>Some(Price(H,0.5082710285336824))
in>>Price(G,0.27663035256663004)
in>>Price(I,0.9854773973480855)
in>>Price(F,0.9448985707882575)
in>>Price(A,0.17042652049262041)
out>>Some(Price(F,0.9448985707882575))
in>>Price(C,0.4185048818658893)
out>>Some(Price(A,0.17042652049262041))
out>>Some(Price(C,0.4185048818658893))
in>>Price(I,0.4010571085499287)
out>>Some(Price(I,0.4010571085499287))
in>>Price(E,0.5089171525630363)
in>>Price(J,0.18523549641820947)
Output exceeds cutoff limit.
```

@SystemFw  reproduced this too.

mpilquist:
Changes hold and holdOption to be pipes that have proper merge and
termination behavior.

@SystemFw Let me know what you think

### DIFF
--- a/core/shared/src/main/scala/fs2/async/async.scala
+++ b/core/shared/src/main/scala/fs2/async/async.scala
@@ -49,7 +49,7 @@ package object async {
     mutable.Queue.circularBuffer[F,A](maxSize)
 
   /**
-   * Pipe that holds the last seen element of the source stream.
+   * Holds the last seen element of the source stream.
    *
    * Each element from the source stream is set to a signal.
    *
--- a/core/shared/src/main/scala/fs2/async/async.scala
+++ b/core/shared/src/main/scala/fs2/async/async.scala
@@ -51,28 +51,36 @@ package object async {
   /**
    * Pipe that holds the last seen element of the source stream.
    *
-   * Backed by a discrete or continuous signal, specified via `discrete` parameter.
+   * Each element from the source stream is set to a signal.
+   *
+   * The output stream is determined by applying `mk` with the signal. The default
+   * `mk` returns the discrete stream from the applied signal.
+   *
    * By default, the returned stream terminates when the source stream terminates, configured
    * via the `haltWhenSourceHalts` parameter.
    *
    * @param initial value emitted until first element from source stream is received
-   * @param discrete true if returned stream should be backed by a discrete signal or false if continuous
+   * @param mk function which generates the output stream from the source signal
    * @param haltWhenSourceHalts true if stream should be terminated when the source stream terminates
    */
-  def hold[F[_]:Effect,A](initial: A, discrete: Boolean = true, haltWhenSourceHalts: Boolean = true)(implicit ec: ExecutionContext): Pipe[F,A,A] =
-    source => Stream.eval(signalOf[F,A](initial)).flatMap { signal =>
-      val in = source.evalMap(signal.set).drain
-      val out = if (discrete) signal.discrete else signal.continuous
-      if (haltWhenSourceHalts) in mergeHaltBoth out
-      else in mergeHaltR out
-    }
+  def hold[F[_]:Effect,A](initial: A, source: Stream[F,A],
+    mk: immutable.Signal[F,A] => Stream[F,A] = (s: immutable.Signal[F,A]) => s.discrete,
+    haltWhenSourceHalts: Boolean = true)()(implicit ec: ExecutionContext): Stream[F,A] =
+      Stream.eval(signalOf[F,A](initial)).flatMap { signal =>
+        val in = source.evalMap(signal.set).drain
+        val out = mk(signal)
+        if (haltWhenSourceHalts) in mergeHaltBoth out
+        else in mergeHaltR out
+      }
 
   /**
    * Like `[[hold]]` but emits `None` until a value from the source has been set,
    * and then emits the last set value wrapped in `Some`.
    */
-  def holdOption[F[_]:Effect,A](discrete: Boolean = true)(implicit ec: ExecutionContext): Pipe[F,A,Option[A]] =
-    source => hold(Option.empty[A], discrete).apply(source.map(Some(_)))
+  def holdOption[F[_]:Effect,A](source: Stream[F,A],
+    mk: immutable.Signal[F,Option[A]] => Stream[F,Option[A]] = (s: immutable.Signal[F,Option[A]]) => s.discrete,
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/async/async.scala
```diff
@@ -49,20 +49,30 @@ package object async {
...
-   * be `initial`.
+   * Backed by a discrete or continuous signal, specified via `discrete` parameter.
+   * By default, the returned stream terminates when the source stream terminates, configured
+   * via the `haltWhenSourceHalts` parameter.
    *
-   * @param source   discrete stream publishing values to this signal
+   * @param initial value emitted until first element from source stream is received
+   * @param discrete true if returned stream should be backed by a discrete signal or false if continuous
+   * @param haltWhenSourceHalts true if stream should be terminated when the source stream terminates
    */
```
SystemFw: Even though, in light of the discussion on #907, I'm not sure the 0.9 behaviour can be obtained, I feel like this signature is less powerful: you lose the ability the asynchronously "sample" the Stream with `Signal.get`, and also need to know at the start whether you want a continuous or discrete Stream, so you can't share the Stream returned by `hold` with two consumers having different requirements.
Not quite sure how to improve it though.

mpilquist:
Instead of a Boolean, we could take a `Signal[F,A] => Stream[F,B]` and default that to `_.discrete`.

SystemFw:
@mpilquist I was coming around to something similar, since `hold` since to fall into the `bracket` (very loosely speaking) family of requiring a "body" to execute. I like it more than the Boolean, ideally we would get some more opinions on usability (even though as it stands, `hold` is unusable in 0.10, so anything is an improvement)

mpilquist:
@SystemFw OK take a look now. I changed the signatures to take a function instead of a boolean. To get reasonably convenient type inference, I had to ditch the `Pipe` and I couldn't define the `mk` function with an unconstrained output type. I'm comfortable with these compromises.

pchlupacek:
@mpilquist , @SystemFw I have few questions on this. 

First, we are losing by changing the signature feature of being able to do Stream[F,A] => F[A] (Signal's get) that I have been using a lot. What's replacement for this? 

Second, do you think it is really necessary to have so much baked in in the combinator? For example `termination boolean` can be expressed by Option (aka. unNoneTerminate), in both cases for hold and holdOption. Similarly, I think `mk` shall be rather used as flatMap in scope of resulting map. 

I maybe sort of missed discussing before why we doing this?

SystemFw:
@pchlupacek Just to be clear, my preference would be to keep the signature as it is, but working (I also use `Signal.get` in the same fashion you described).
However, I wasn't able to see how, but I find your comment on `merge` in #907 very interesting, maybe we should keep the discussion there to see if there are any more ideas worth exploring?

pchlupacek:
@SystemFw the suggesting to move this elsewhere is that this is more complex than just issue for the signal // hold. It also affects how errors are propagated (there is difference of behaviour now between merge and join, which I found very confusing). Essentially I would like to agree on laws for nondet stream merging and making sure they are consistent agains all nondet merge combinators (pipe2, join, ...) if possible.

SystemFw:
@pchlupacek Yes, I agree on that :) Could you please notify me on the issue when/if you open it? I'd like to hear that discussion

Issue #911, opened 2017-08-01
Laws for nondeterministic combinators


pchlupacek:
Currently we have few non deterministic operators in fs2
 - `pipe2`, i.e. merge, either ...
 - `join`

It seems that behaviour of these combinators subtle changed from 0.9 to 0.10 release that may cause some programs to compile under 0.10 but not work properly. Furthermore, there is difference now in 0.9 and 0.10 how `join(Stream(streamA1, streamA2))` works and how `streamA1 merge streamA2` will work (specifically for drained streams) 

I would like in this thread to settle on laws for these combinators. My ideas for the laws would be in areas: 

- how drained streams (ones that produce no values) are run
- if either of the streams fails, how that affects downstream 

@mpilquist @djspiewak @SystemFw @pchiusano @AdamChlupacek can you please share some thoughts here? 

This is continuation of discussion in #907

mpilquist:
@pchlupacek Could you describe the differences in error propagation between `merge` and `join`? I'm aware of the differences in drained stream handling.

pchlupacek:
@mpilquist lets say have these programs: 

```scala

val streamA: Stream[F, Int] = Stream.range(0, Int.MaxValue) onError { err => Stream.eval_(dumError(err)) }

val streamB: Stream[F, Nothing] = time.sleep_(1.second) ++ Stream.fail(new Throwable("BOOM"))


val program1 = (concurrent.join(Int.MaxValue)(Stream(streamA, streamB))) zip time.awakeEvery(20.seconds)

val program2 = (streamA merge streamB) zip time.awakeEvery(10.seconds)

```

then program1 will behave as follows: 

when StreamB fails (after 1s) then StreamA is killed, cleanups run and program will terminate after 20s with failure.  

however program2 will behave as follows: 

When StreamB fails (after 1s) then nothing will happen, however after 20s, when next element is requested, then cleanups for StreamA are run, and then program terminates with failure

Note the different time when cleanups on StreamA are called in both scenarios. 

This was observed on 0.9.6.

mpilquist:
@pchlupacek Ah yes, okay. And that behavior is the same in 0.9 and 0.10, right?

pchlupacek:
@mpilquist cannot confirm that 100% unfortunately.  I would assume however that yes, given 0.10 merge difference and fact that join has same code as 0.9, the behaviour shall be the same.

pchlupacek:
[...]

PR #914, opened 2017-08-04
Fixes #907

### LINKED ISSUE #907: Signal regression from 0.9 -> 0.10: only first item emitted on `repeat` source.

mpilquist:
Introduced `s.concurrently(t)` as a safe replacement for `s.mergeHaltR(t.drain)`. Removed `mergeHaltL` and `mergeHaltR`. Changed `hold` to be implemented with `concurrently`.

@pchlupacek This should cover everything from #911. /cc @SystemFw

### DIFF
--- a/core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
+++ b/core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
@@ -1,0 +1,37 @@
+package fs2
+
+import scala.concurrent.duration._
+import cats.effect.IO
+
+class ConcurrentlySpec extends Fs2Spec {
+
+  "concurrently" - {
+
+    "when background stream terminates, overall stream continues" in forAll { (s1: PureStream[Int], s2: PureStream[Int]) =>
+      runLog(Scheduler[IO](1).flatMap(scheduler => (scheduler.sleep_[IO](25.millis) ++ s1.get).concurrently(s2.get))) shouldBe s1.get.toVector
+    }
+
+    "when background stream fails, overall stream fails" in forAll { (s: PureStream[Int], f: Failure) =>
+      val prg = Scheduler[IO](1).flatMap(scheduler => (scheduler.sleep_[IO](25.millis) ++ s.get).concurrently(f.get))
+      val throws = f.get.drain.run.attempt.unsafeRunSync.isLeft
+      if (throws) an[Err.type] should be thrownBy runLog(prg)
+      else runLog(prg)
+    }
+
+    "when primary stream fails, overall stream fails and background stream is terminated" in forAll { (f: Failure) =>
+      var bgDone = false
+      val bg = Stream.repeatEval(IO(1)).onFinalize(IO(bgDone = true))
+      val prg = Scheduler[IO](1).flatMap(scheduler => (scheduler.sleep_[IO](25.millis) ++ f.get).concurrently(bg))
+      an[Err.type] should be thrownBy runLog(prg)
+      bgDone shouldBe true
+    }
+
+    "when primary stream termiantes, background stream is terminated" in forAll { (s: PureStream[Int]) =>
+      var bgDone = false
+      val bg = Stream.repeatEval(IO(1)).onFinalize(IO(bgDone = true))
+      val prg = Scheduler[IO](1).flatMap(scheduler => (scheduler.sleep_[IO](25.millis) ++ s.get).concurrently(bg))
+      runLog(prg)
+      bgDone shouldBe true
+    }
+  }
+}
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1772,7 +1772,7 @@ object Stream {
      * concurrent data structure and merging it with a consumer from the same
      * data structure. In such cases, use `consumer.concurrently(producer)`
      * instead of `consumer.mergeHaltR(producer.drain)` to ensure the producer
-     * continues to run in paralle with consumer processing.
+     * continues to run in parallel with consumer processing.
      *
      * @example {{{
      * scala> import scala.concurrent.duration._, scala.concurrent.ExecutionContext.Implicits.global, cats.effect.IO
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -941,7 +941,7 @@ final class Stream[+F[_],+O] private(private val free: FreeC[Algebra[Nothing,Not
    * res0: List[Int] = List(0, 1, 2, 3, 4)
    * }}}
    */
-  def takeWhile(p: O => Boolean): Stream[F,O] = this.pull.takeWhile(p).stream
+  def takeWhile(p: O => Boolean, takeFailure: Boolean = false): Stream[F,O] = this.pull.takeWhile(p, takeFailure).stream
 
   /**
    * Converts the input to a stream of 1-element chunks.
@@ -1436,7 +1436,8 @@ object Stream {
      * @example {{{
      * scala> import cats.effect.IO, scala.concurrent.ExecutionContext.Implicits.global
      * scala> val data: Stream[IO,Int] = Stream.range(1, 10).covary[IO]
-     * scala> Stream.eval(async.signalOf[IO,Int](0)).flatMap(s => Stream(s).concurrently(data.evalMap(s.set))).flatMap(_.discrete).takeWhile(_ < 9).runLast.unsafeRunSync
+     * scala> Stream.eval(async.signalOf[IO,Int](0)).flatMap(s => Stream(s).concurrently(data.evalMap(s.set))).flatMap(_.discrete).takeWhile(_ < 9, true).runLast.unsafeRunSync
+     * res0: Option[Int] = Some(9)
      * }}}
      */
     def concurrently[O2](that: Stream[F,O2])(implicit F: Effect[F], ec: ExecutionContext): Stream[F,O] = {
@@ -2608,7 +2609,7 @@ object Stream {
      * and returns the remaining `Stream`. If non-empty, the returned stream will have
      * a first element `i` for which `p(i)` is `false`.
      */
-    def takeWhile(p: O => Boolean): Pull[F,O,Option[Stream[F,O]]] = takeWhile_(p, false)
+    def takeWhile(p: O => Boolean, takeFailure: Boolean = false): Pull[F,O,Option[Stream[F,O]]] = takeWhile_(p, takeFailure)
 
     private def takeWhile_(p: O => Boolean, takeFailure: Boolean): Pull[F,O,Option[Stream[F,O]]] =
       uncons.flatMap {
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1417,6 +1417,48 @@ object Stream {
     def changes(implicit eq: Eq[O]): Stream[F,O] = self.filterWithPrevious(eq.neqv)
 
     /**
+     * Runs the supplied stream in the background as elements from this stream are pulled.
+     *
+     * The resulting stream terminates upon termination of this stream. The background stream will
+     * be interrupted at that point. Early termination of `that` does not terminate the resulting stream.
+     *
+     * Any errors that occur in either `this` or `that` stream result in the overall stream terminating
+     * with an error.
+     *
+     * Upon finalization, the resulting stream will interrupt the background stream and wait for it to be
+     * finalized.
+     *
+     * This method is similar to `this mergeHaltL that.drain` but ensures the `that.drain` stream continues
+     * to be evaluated regardless of how `this` is evaluated or how the resulting stream is processed.
+     * This method is also similar to `Stream(this,that).join(2)` but terminates `that` upon termination of
+     * `this`.
+     *
+     * @example {{{
+     * scala> import cats.effect.IO, scala.concurrent.ExecutionContext.Implicits.global
+     * scala> val data: Stream[IO,Int] = Stream.range(1, 10).covary[IO]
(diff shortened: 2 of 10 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1722,6 +1764,16 @@ object Stream {
...
      *
+     * Note: `this` and `that` are each pulled for a segment. Upon receiving
+     * a segment, it is emitted downstream. Depending on how that element is
+     * processed, the remainder of `this` and `that` may never be consulted
+     * again (e.g., `a.merge(b) >> Stream.constant(0)`). A common case where
+     * this can be problematic is draining a stream that publishes to a
+     * concurrent data structure and merging it with a consumer from the same
+     * data structure. In such cases, use `consumer.concurrently(producer)`
+     * instead of `consumer.mergeHaltR(producer.drain)` to ensure the producer
+     * continues to run in paralle with consumer processing.
```
pchlupacek: *parallel

core/shared/src/main/scala/fs2/async/async.scala
```diff
@@ -57,12 +57,14 @@ package object async {
    *
    * @param source   discrete stream publishing values to this signal
    */
-  def hold[F[_]:Effect,A](initial: A, source: Stream[F, A])(implicit ec: ExecutionContext): Stream[F, immutable.Signal[F,A]] =
-     immutable.Signal.hold(initial, source)
+  def hold[F[_],A](initial: A, source:Stream[F,A])(implicit F: Effect[F], ec: ExecutionContext): Stream[F,immutable.Signal[F,A]] =
+    Stream.eval(signalOf[F,A](initial)) flatMap { sig =>
+      Stream(sig).concurrently(source.flatMap(a => Stream.eval_(sig.set(a))))
```
pchlupacek: source evalMap (... ) ?

docs/guide.md
```diff
@@ -478,9 +478,9 @@ scala> Stream(1,2,3).merge(Stream.eval(IO { Thread.sleep(200); 4 })).runLog.unsa
 res38: Vector[Int] = Vector(1, 2, 3, 4)
 ```
 
-The `merge` function is defined in [`pipe2`](../core/shared/src/main/scala/fs2/pipe2.scala), along with other useful concurrency functions, like `interrupt` (halts if the left branch produces `false`), `either` (like `merge` but returns an `Either`), `mergeHaltBoth` (halts if either branch halts), and others.
+The `merge` function supports concurrency. FS2 has a number of other useful concurrency functions like `concurrently` (runs another stream concurrently and discards its output), `interrupt` (halts if the left branch produces `false`), `either` (like `merge` but returns an `Either`), `mergeHaltBoth` (halts if either branch halts), and others.
```
pchlupacek: Michael, I am just wondering, the interrupt now, wouldn't it suffer from the same issues like mergeDrainX? I mean currently, interrupt wouldn't be able to interrupt the other branch of merge, unless something processes the elements correct? Not sure if that couldn't be problematic too ?

mpilquist: Interrupt relies on merging both streams and inspecting the emitted interruption value in the merged output. As such, it's definitely not safe to use with a downstream that emits an infinite stream. I think this is OK though -- this has always been its behavior.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1417,6 +1417,48 @@ object Stream {
...
+     * The resulting stream terminates upon termination of this stream. The background stream will
+     * be interrupted at that point. Early termination of `that` does not terminate the resulting stream.
+     *
+     * Any errors that occur in either `this` or `that` stream result in the overall stream terminating
+     * with an error.
+     *
+     * Upon finalization, the resulting stream will interrupt the background stream and wait for it to be
+     * finalized.
+     *
+     * This method is similar to `this mergeHaltL that.drain` but ensures the `that.drain` stream continues
```
SystemFw: I like this explanation, but it refers to `mergeHaltL`, which is removed. Likely to be confusing to someone who's not familiar with the old API perhaps?

mpilquist: I left `mergeHalt{L,R,Both}` -- only removed `mergeDrain{L,R}`

SystemFw: Right, so there's a typo in the docs below I think

docs/migration-guide-0.10.md
```diff
@@ -126,6 +126,22 @@ Scheduler[IO](corePoolSize = 1).flatMap { scheduler =>
...
+  Stream.emit(s) mergeHaltR source.evalMap(s.set).drain
+}
+```
+
+The general idea behind this pattern is to create a producer and a consumer and then combine them in to a single stream by merging the consumer with the result of draining the producer. In 0.10, the two streams used in a merge are only consulted when downstream needs an element. As a result, it's possible to deadlock when using this pattern. This pattern relies on the fact that the drained producer will continue to execute in parallel with the consumer, *even when the downstream never asks for another element from the merged stream*.
+
+To address this use case, the `concurrently` method has been added. Instead of `consumer.mergeHaltR(producer.drain)`, use `consumer.concurrently(producer)`. See the ScalaDoc for `concurrently` for more details.
+
+Given that most usage of merging a drained stream with another stream should be replaced with `concurrently`, we've removed `mergeHaltL` and `mergeHaltR`.
+
```
SystemFw: Do you mean `mergeDrainL` and `mergeDrainR` here?

mpilquist: Yep :)

core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    "when primary stream termiantes, background stream is terminated" in forAll { (s: PureStream[Int]) =>
+      var bgDone = false
+      val bg = Stream.repeatEval(IO(1)).onFinalize(IO(bgDone = true))
+      val prg = Scheduler[IO](1).flatMap(scheduler => (scheduler.sleep_[IO](25.millis) ++ s.get).concurrently(bg))
+      runLog(prg)
+      bgDone shouldBe true
+    }
+  }
```
pchlupacek: @mpilquist can you please verify that program below terminates before 1 minute timeout ? Note the main program emits once then uses the discrete of signal in flatMap
```
Stream.eval(async.signalOf[IO, Int](-1)) flatMap { signal => 
  (Stream.emit(signal)).concurrently( 
     (Stream.range(0, 10) ++ Stream.fail(new Throwable("BOOM")) .evalMap(signal.set)
  )
} flatMap { signal => 
 signal.discrete  
}}

 

```

docs/guide.md
pchlupacek: I think in 0.9 this would have completely different behaviour than in 0.10

```scala

val s1: Stream[Int] = ???

val s2: Stream[Boolean]  = time.sleep_(100 millis) ++ Stream.emit(true)

(s1 interruptWhen s2) flatMap { i =>
 Stream.emit(i) ++ sleep_(1.minute)
}
```
In 0.9 this will interrupt s1 in 100 millis, however in 0.10 this will interrupt s1 in 1.minute

mpilquist: @pchlupacek Just tested this under 0.9.7 and confirmed that it interrupts after 1 minute, not after 100 milliseconds.

core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
mpilquist: @pchlupacek That program hangs. The background stream fails, which causes the primary stream to have its interruption signal set. But the primary stream never inspects the interruption signal b/c it is still focused on infinite stream returned from discrete. It's not clear to me why this pattern works in 0.9 given the way interrupt is implemented.

mpilquist: @pchlupacek Ah ha! It hangs in 0.9 as well for the same reason. E.g.

```scala
Stream.eval(async.signalOf[Task,Boolean](false)).flatMap(s => Stream.eval_(Task.start(s.set(true).schedule(1.second))) ++ Stream.emit(s).interruptWhen(s)).flatMap(s => s.discrete).runLog.unsafeRunAsync(println)
```

pchlupacek: @mpilquist in 0.9 it won't work. in 0.8 however, that will work. My concern here is that if we won't be able to somehow resolve this, then this `concurrently` pattern is inherently unsafe and will lead to unsound dead-locked programs. Also, current mergeDrainL, R, etc patterns are suffering from this, and as such migration of complex concurrents programs from 0.8 to 0.9 is really painful :-(. 

@djspiewak, @pchiusano  I would be really grateful on your input here, I think it is really important design decision we are going to make here, and hence your worked on many concurrent primitives, perhaps you may share your opinion on this ?

pchlupacek: @mpilquist yeh, you were faster than me :-)

mpilquist: @pchlupacek Note that it's not `concurrently` that's inherently unsafe but rather the fact that `interrupt` is implemented in terms of `merge`. If we could implement `interrupt` in terms of `concurrently` instead, it would work fine. I don't know how to do that though -- we'd need some way to communicate the interruption to the main interpreter loop.

mpilquist: Well that's kind of strange I guess - `concurrently` is implemented with `interruptWhen` and `interruptWhen` would be implemented in terms of `concurrently`. So ignore that comment.

pchlupacek: @mpilquist exactly. Is all about the laws here :-) essentially currently we are breaking the law that when right side (Stream[F, Unit]) fails, that resulting stream fails too :-)

pchlupacek: And that said, this is for any pipe2, which I am quite sure will suffer from this behaviour.

pchlupacek: :-) Catch-22

mpilquist:
Tell me more, I'm not seeing it...

pchlupacek:
I was merely just commenting on your idea with concurrently/interruptWhen which I considered `Catch-22` like

mpilquist:
After more discussion with @pchlupacek offline, we decided to merge this and open a new issue for the concerns around `interruptWhen`.

PR #918, opened 2017-08-09
It is possible - and indeed does happen intermittently when rapidly f…

sbuzzard:
…lat mapping segment - for inner to be referenced when null.  This fix was suggested by @mpilquist to plug this hole.

### DIFF
--- a/core/shared/src/main/scala/fs2/Segment.scala
+++ b/core/shared/src/main/scala/fs2/Segment.scala
@@ -333,11 +333,12 @@ abstract class Segment[+O,+R] { self =>
 
       outerStep.map { outer =>
         step {
+          val innerRem = if (inner eq null) Segment.empty[O2].asInstanceOf[Segment[O2,S]] else inner.remainder
           outerResult match {
             case Some(r) =>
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -333,11 +333,12 @@ abstract class Segment[+O,+R] { self =>
 
       outerStep.map { outer =>
         step {
+          val innerRem = if (inner eq null) Segment.empty[O2].asInstanceOf[Segment[O2,S]] else inner.remainder
```
mpilquist: Isolated to just this line, this cast is not safe. The `innerRem` val is used 3 places below -- the two `prepend(innerRem)` cases are safe though, as we will never look at the result of `innerRem`. The other use case is not safe though.

sbuzzard: I had a feeling since the result parameter is fixed to Unit for empty (since its just an alias to Chunk.empty).  So I guess Segment.empty won't work after all.

mpilquist:
Fixed in #922

PR #920, opened 2017-08-16
Added file/directory watching support to fs2.io.file based on NIO WatchService

mpilquist:


mpilquist:
@SystemFw Would you do a review of this? Thanks!

SystemFw:
Will do :)

SystemFw:
I'll get to this tomorrow

SystemFw:
The user facing API looks great. 
The implementation also looks good to me: the extra complexity from recursive watching and single file watching is justified imo. I definitely wouldn't want users to have to implement that.

Issue #923, opened 2017-08-20
Should Signal have a setSyncPure?


zainab-ali:
I have a case where a value needs to be set synchronously for an arbitrary effect type.  
In 0.10, this can be achieved by using the `setSyncPure` function on `Ref`:
```scala
F.runAsync(ref.setSyncPure(value))(_ => IO.unit).unsafeRunSync()
```

~~It's not possible to call this in 0.9 where the private `fs2.internal.Ref` is modeled very differently.~~ 

**EDIT**: fs2 exposes an `Async.Ref` which has an asynchronous set method.

This logic is much better described using a `Signal` with a synchronous set method.

Are there any thoughts on having a `setSync` and `setSyncPure` method on `Signal`, which defer to the methods on the underlying ref?  

I'd be happy to add this, if desired.

SystemFw:
Have you checked the right `Ref`? There's a `Ref` in `Task` which is roughly equivalent to the `Ref` in 0.10

SystemFw:
@zainab-ali do you still need this, or is `Async.Ref` enough?

zainab-ali:
You're right!  I was looking at the wrong `Ref`.  fs2 exposes an `Async.Ref` which is public.

However, this doesn't have a synchronous set method - I do still need this behavior.

mpilquist:
@zainab-ali Porting `setSyncPure` from 0.10's `fs2.async.Ref` to 0.9's `fs2.util.Async.Ref` should be straightforward. If you want to give it a shot, I'd be happy to merge a PR and answer any questions that arise.

zainab-ali:
That would be great!  I shall give it a go

zainab-ali:
As discussed in https://github.com/zainab-ali/fs2-reactive-streams/pull/31, this change is no longer necessary

PR #925, opened 2017-08-22
Add setSync to `Async.Ref`

zainab-ali:
Adds `setSync` and `setSyncPure` methods to `Async.Ref`, as discussed in #923 .

### DIFF
--- a/core/shared/src/main/scala/fs2/Task.scala
+++ b/core/shared/src/main/scala/fs2/Task.scala
@@ -280,7 +280,7 @@ object Task extends TaskPlatform with TaskInstances {
   private object Msg {
     final case class Read[A](cb: Callback[(A, Long)], id: MsgId) extends Msg[A]
     final case class Nevermind[A](id: MsgId, cb: Callback[Boolean]) extends Msg[A]
-    final case class Set[A](r: Attempt[A]) extends Msg[A]
+    final case class Set[A](r: Attempt[A], cb: () => Unit) extends Msg[A]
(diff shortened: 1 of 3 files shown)


zainab-ali:
Adding a method to the public `Async.Ref` trait breaks binary compatibility, hence the build failure.

mpilquist:
Yeah, we'll have to add that method to `Task.Ref` only.

core/shared/src/main/scala/fs2/Task.scala
```diff
@@ -342,6 +342,9 @@ object Task extends TaskPlatform with TaskInstances {
     def set(t: Task[A]): Task[Unit] =
       Task.delay { S { t.unsafeRunAsync { r => actor ! Msg.Set(r) } }}
 
+    def setSync(t: Task[A]): Task[Unit] =
+      Task.delay { t.unsafeRunAsync { r => actor ! Msg.Set(r) }}
```
mpilquist: Note that this is still asynchronous -- once the `Msg.Set(r)` message has veen sent to the actor, the resultant `Task[Unit]` is completed. We need to use `Task.async` and pass the callback along to the actor, letting the actor invoke the callback only after the `Set` has happened.

mpilquist:
@zainab-ali Looks good. The 3 remaining errors from MIMA are safe to ignore. Can you add the noted exclusions to build.sbt?

zainab-ali:
As discussed in https://github.com/zainab-ali/fs2-reactive-streams/pull/31, this change is no longer necessary.

PR #926, opened 2017-08-25
Avoids creating a new array in chunk if chunk is already a Chunk.Byte…

sbuzzard:
…s - which can be expensive depending on the size of the array and behavior of map strict in the chunk implementation, since the chunk.toArray is invoked on the selector thread.


sbuzzard:
closing pending re-work.

PR #927, opened 2017-08-26
Avoids converting from/to byte array from Chunk within UDP ASG single…

sbuzzard:
…ton socket selector thread by introducing a package private analog of the public Packet data type which holds a byte array.  The conversion from/to Chunk is done in the socket implementation prior to calling ASG.read or ASG write.  More a proposal or query for better solution at this point.  This does have the effect, at least in two applications, with eliminating the selector thread as a hot spot under load (one app, the client, creates 20 concurrent UDP connections and writes data at 38.8mbps to the 2nd app, which in turn manipulates the data and outputs it to third UDP destination).

### DIFF
--- a/io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
+++ b/io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
@@ -18,6 +18,7 @@ import java.util.concurrent.{ConcurrentLinkedQueue,CountDownLatch}
  *
  * Each `AsynchronousSocketGroup` is assigned a single daemon thread that performs all read/write operations.
  */
+
 sealed trait AsynchronousSocketGroup {
   private[udp] type Context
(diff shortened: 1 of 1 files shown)


io/src/main/scala/fs2/io/udp/AsynchronousSocketGroup.scala
```diff
@@ -193,6 +199,10 @@ object AsynchronousSocketGroup {
     }
 
     override def write(key: SelectionKey, packet: Packet, timeout: Option[FiniteDuration], cb: Option[Throwable] => Unit): Unit = {
+      val writerPacket = {
+        val b = packet.bytes.toBytes
+        new WriterPacket(packet.remote, ByteBuffer.wrap(if (b.size == b.values.size) b.values else b.values.slice(b.offset, b.size)))
```
mpilquist: Instead of using `slice`, let's use `Array.copy` instead

PR #932, opened 2017-09-17
Segment over Chunk for Stream Emits

ChristopherDavenport:
While [Algebra takes a Segment](https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/internal/Algebra.scala#L19) and I'm sure many of us have been told to use Segment for anything that does not require index based access. However [`Stream.emits`](https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/Stream.scala#L1189) uses `Chunk.seq`.

In order to help I build it out of Segment, to later realize [`Segment.seq`](https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/Segment.scala#L1235) uses [`Chunk.seq`](https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/Chunk.scala#L246) which uses [`Chunk.indexedSeq`](https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/Chunk.scala#L237) which is definitely designed for index based access. 

Is this intended and what should be the preferred invocation, both internally and for a user.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1186,7 +1186,7 @@ object Stream {
   def emits[O](os: Seq[O]): Stream[Pure,O] = {
     if (os.isEmpty) empty
     else if (os.size == 1) emit(os.head)
-    else fromFreeC(Algebra.output[Pure,O](Chunk.seq(os)))
+    else fromFreeC(Algebra.output[Pure,O](Segment.seq(os)))
   }
 
   private[fs2] val empty_ = fromFreeC[Nothing,Nothing](Algebra.pure[Nothing,Nothing,Unit](())): Stream[Nothing,Nothing]


mpilquist:
When writing pipes / sinks, prefer `Segment` over `Chunk` because you'll get operator fusion for free and overall much better performance. When constructing a `Segment` or `Chunk`, I generally use the most specific constructor. E.g., `Chunk.indexedSeq` for `Vector`, `Chunk.seq` for `List`, `Segment.unfold` for general cases. The common chunk constructors are aliased on `Segment` to aid in discoverability -- the aliases also return `Segment` instead of `Chunk`, "forgetting" the specialization, though this is typically used for inference improvements more than implementation hiding.

mpilquist:
Feel free to reopen if you disagree. Maybe we should update ScalaDoc and/or guide to talk about these types of differences.

PR #937, opened 2017-09-28
fix for npe in segment's flatmap

tscholak:
I detected this issue when I was running old code that I ported from 0.9 to 0.10. I discovered that there is the real possibility of a null pointer exception in the `flatMap` method of `Segment`, see the backtrace here: http://lpaste.net/358781

This problem is related to https://github.com/functional-streams-for-scala/fs2/pull/922 where a null pointer exception was discovered and fixed in the `flatMapAccumulate` method, also of `Segment`.

I tested the fix with my code, and the problem has disappeared.

Take this fix with a grain of salt. It might be incorrect. I can't say I understand the segment stepper code very well. I followed the types and used common sense to create this solution.

### DIFF
--- a/core/shared/src/main/scala/fs2/Segment.scala
+++ b/core/shared/src/main/scala/fs2/Segment.scala
@@ -284,12 +284,18 @@ abstract class Segment[+O,+R] { self =>
 
       outerStep.map { outer =>
         step {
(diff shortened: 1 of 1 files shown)


tscholak:
PS: no idea why gh picked up changes in `Stream.scala`.

mpilquist:
Thanks for finding this and fixing!

PR #939, opened 2017-09-30
Add a scope to a stream before append

### LINKED ISSUE #930: Sometimes resources are released too late
Version is `0.10.0-M6`. Here is a reproducer:

```scala
import cats.effect.IO
import cats.syntax.flatMap._
import fs2._
val scheduler   = Scheduler.allocate[IO](1).unsafeRunSync()
(
//Stream.emit("mess up resources") ++ //uncomment to se the bug
Stream.bracket(IO {
  println("->Acquire")
})(
    _ => scheduler._1.awakeEvery[IO](500.millis).map(_.toString)
  , _ => IO { println("<-Release") }
))
.interruptWhen(scheduler._1.sleep[IO](1.seconds).map(_ => true))
.repeat
.interruptWhen(scheduler._1.sleep[IO](2.seconds).map(_ => true))
.run
.unsafeRunSync()
```

prints
```
->Acquire
<-Release
->Acquire
<-Release
```

...but if one uncomment that innocent `Stream.emit`, then output would be
```
->Acquire
->Acquire
<-Release
<-Release
```
What is going on here?

avakhrenev:
Fixes #930

### DIFF
--- a/core/jvm/src/test/scala/fs2/ResourceSafetySpec.scala
+++ b/core/jvm/src/test/scala/fs2/ResourceSafetySpec.scala
@@ -205,6 +205,16 @@ class ResourceSafetySpec extends Fs2Spec with EventuallySupport {
       o shouldBe (0 until 10).toVector
     }
 
+    "finalizers are run before next appended stream" in {
+      val c = new AtomicLong(0)
+      val s1 = Stream.emit("a").covary[IO]
+      val s2 = Stream.bracket(IO { c.incrementAndGet() shouldBe 1L; () })(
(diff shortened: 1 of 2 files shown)


avakhrenev:
Don't quite understand the mechanic of scopes. How it is supposed to work when interpreter encounters `CloseScope` for already closed scope?

mpilquist:
That isn't supposed to be possible (hence the throw in the code). Not sure why this change introduces this issue though...

avakhrenev:
Ok, seems like scopes need another thought.

PR #942, opened 2017-10-04
Add doctest making Stream.syncInstance implicit

jdnavarro:
This doctest may help making the connection between using `cats` syntax for `Stream`s and making `syncInstance` implicit more obvious.


mpilquist:
Thanks!

PR #944, opened 2017-10-06
Make sure MessageDigest respects Chunk size and offset

SystemFw:
Fixes #943


adamretter:
Thanks for fixing this so quickly :-)

Just one comment - Would have been nice to see a test as well to prevent regressions. I did add a test to my PR for that purpose.

SystemFw:
Yes, you are absolutely right. Since I'm quite busy today I sent the fix as soon as possible :grin:

SystemFw:
I'm thinking of a test that wouldn't involve having to create a file, since this is unit testable in principle

aeons:
Cool. Any chance this would trigger another milestone?

PR #949, opened 2017-10-12
Changed UnconsAsync so that it does not capture Effect[F] instance, resulting in proper translate signature

mpilquist:
This is the more ambitious version of the fix for #948.

The `Algebra.UnconsAsync` case no longer captures the `Effect[F]` instance, which means we can have the standard definition of `Stream#translate`. However, running a stream now requires an `Effect[F]` instead of a `Sync[F]`. I provided an alternative for each run function that requires a `Sync[F]` but that fails if an `UnconsAsync` step is encountered.

Overall, the expressiveness of the library is not changed by this PR. Rather, we moved where `Effect` instances are required.

/cc @tpolecat

### DIFF
--- a/core/jvm/src/test/scala/fs2/StreamSpec.scala
+++ b/core/jvm/src/test/scala/fs2/StreamSpec.scala
@@ -100,17 +100,9 @@ class StreamSpec extends Fs2Spec with Inside {
         IndexedSeq.range(0, 100)
     }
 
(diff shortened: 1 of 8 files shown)


tpolecat:
Rock on. I like this better because it fixes my problem, but also because Stream becomes a more cromulent higher-kinded functor: `translate(F andThen G)` can now always be written `translate(F).translate(G)` which wasn't the case before. 👍

PR #950, opened 2017-10-15
Reintroduce repartition

soumyarupsarkar:
This PR reintroduces the `repartition` combinator. This was originally added to scalaz-stream by @fthomas in #93

### DIFF
--- a/core/jvm/src/test/scala/fs2/StreamSpec.scala
+++ b/core/jvm/src/test/scala/fs2/StreamSpec.scala
@@ -101,14 +101,14 @@ class StreamSpec extends Fs2Spec with Inside {
     }
 
     "repartition" in {
-      Stream("Lore", "m ip", "sum dolo", "r sit amet").repartition(_.split(" ").toIndexedSeq).toList ==
+      Stream("Lore", "m ip", "sum dolo", "r sit amet").repartition(s => Segment.array(s.split(" "))).toList ==
         List("Lorem", "ipsum", "dolor", "sit", "amet") &&
-      Stream("hel", "l", "o Wor", "ld").repartition(_.grouped(2).toVector).toList ==
+      Stream("hel", "l", "o Wor", "ld").repartition(s => Segment.indexedSeq(s.grouped(2).toVector)).toList ==
         List("he", "ll", "o ", "Wo", "rl", "d") &&
-      Stream(1, 2, 3, 4, 5).repartition(i => Vector(i, i)).toList ==
+      Stream(1, 2, 3, 4, 5).repartition(i => Segment.indexedSeq(Vector(i, i))).toList ==
         List(1, 3, 6, 10, 15, 15) &&
-      (Stream(): Stream[Nothing, String]).repartition(_ => Vector()).toList == List() &&
-      Stream("hello").repartition(_ => Vector()).toList == List()
+      (Stream(): Stream[Nothing, String]).repartition(_ => Segment.empty).toList == List() &&
+      Stream("hello").repartition(_ => Segment.empty).toList == List()
     }
 
     "translate" in forAll { (s: PureStream[Int]) =>
--- a/core/jvm/src/test/scala/fs2/StreamSpec.scala
+++ b/core/jvm/src/test/scala/fs2/StreamSpec.scala
@@ -101,14 +101,14 @@ class StreamSpec extends Fs2Spec with Inside {
     }
 
     "repartition" in {
-      Stream("Lore", "m ip", "sum dolo", "r sit amet").repartition(s => Segment.array(s.split(" "))).toList ==
-        List("Lorem", "ipsum", "dolor", "sit", "amet") &&
-      Stream("hel", "l", "o Wor", "ld").repartition(s => Segment.indexedSeq(s.grouped(2).toVector)).toList ==
-        List("he", "ll", "o ", "Wo", "rl", "d") &&
-      Stream(1, 2, 3, 4, 5).repartition(i => Segment.indexedSeq(Vector(i, i))).toList ==
-        List(1, 3, 6, 10, 15, 15) &&
-      (Stream(): Stream[Nothing, String]).repartition(_ => Segment.empty).toList == List() &&
-      Stream("hello").repartition(_ => Segment.empty).toList == List()
+      Stream("Lore", "m ip", "sum dolo", "r sit amet").repartition(s => Segment.array(s.split(" "))).toList shouldBe
+        List("Lorem", "ipsum", "dolor", "sit", "amet")
+      Stream("hel", "l", "o Wor", "ld").repartition(s => Segment.indexedSeq(s.grouped(2).toVector)).toList shouldBe
+        List("he", "ll", "o ", "Wo", "rl", "d")
+      Stream(1, 2, 3, 4, 5).repartition(i => Segment.indexedSeq(Vector(i, i))).toList shouldBe
+        List(1, 3, 6, 10, 15, 15)
+      (Stream(): Stream[Nothing, String]).repartition(_ => Segment.empty).toList shouldBe List()
+      Stream("hello").repartition(_ => Segment.empty).toList shouldBe List()
     }
 
     "translate" in forAll { (s: PureStream[Int]) =>
--- a/core/jvm/src/test/scala/fs2/StreamSpec.scala
+++ b/core/jvm/src/test/scala/fs2/StreamSpec.scala
@@ -105,10 +105,20 @@ class StreamSpec extends Fs2Spec with Inside {
         List("Lorem", "ipsum", "dolor", "sit", "amet")
       Stream("hel", "l", "o Wor", "ld").repartition(s => Segment.indexedSeq(s.grouped(2).toVector)).toList shouldBe
         List("he", "ll", "o ", "Wo", "rl", "d")
-      Stream(1, 2, 3, 4, 5).repartition(i => Segment.indexedSeq(Vector(i, i))).toList shouldBe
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1984,6 +1984,36 @@ object Stream {
...
+      * are emitted. The last element is then appended to the next input using the
+      * Semigroup `S`.
+      *
+      * @example {{{
+      * scala> import cats.implicits._
+      * scala> Stream("Hel", "l", "o Wor", "ld").repartition(_.split(" ").toIndexedSeq).toList
+      * res0: List[String] = List(Hello, World)
+      * }}}
+      */
+    def repartition(p: O => IndexedSeq[O])(implicit S: Semigroup[O]): Stream[F,O] = {
```
mpilquist: Could we change this signature to take a `p: O => Segment[O,Unit]`?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1984,6 +1984,36 @@ object Stream {
...
+      def go(carry: Option[O], s: Stream[F,O]): Pull[F,O,Unit] = {
+        s.pull.uncons1.flatMap {
```
mpilquist: It would be much more efficient to use `s.pull.uncons` here. It will complicate the body somewhat but likely worth the tradeoff.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1984,6 +1984,41 @@ object Stream {
...
+      * scala> import cats.implicits._
+      * scala> Stream("Hel", "l", "o Wor", "ld").repartition(s => Segment.array(s.split(" "))).toList
+      * res0: List[String] = List(Hello, World)
+      * }}}
+      */
+    def repartition(p: O => Segment[O, Unit])(implicit S: Semigroup[O]): Stream[F,O] = {
+      def go(carry: Option[O], s: Stream[F,O]): Pull[F,O,Unit] = {
+        s.pull.uncons.flatMap {
+          case Some((hds, tl)) =>
+            S.combineAllOption(hds.toVector) match {
```
mpilquist: Can we implement this without using `toVector` or any of the other strict operations on `Segment`? If `hds` is infinite, the `toVector` call will throw an OOM.

soumyarupsarkar: Done, thanks! Let me know if you had something different in mind.

core/jvm/src/test/scala/fs2/StreamSpec.scala
```diff
@@ -100,6 +100,17 @@ class StreamSpec extends Fs2Spec with Inside {
...
     }
 
+    "repartition" in {
+      Stream("Lore", "m ip", "sum dolo", "r sit amet").repartition(s => Segment.array(s.split(" "))).toList shouldBe
+        List("Lorem", "ipsum", "dolor", "sit", "amet")
+      Stream("hel", "l", "o Wor", "ld").repartition(s => Segment.indexedSeq(s.grouped(2).toVector)).toList shouldBe
+        List("he", "ll", "o ", "Wo", "rl", "d")
+      Stream(1, 2, 3, 4, 5).repartition(i => Segment.indexedSeq(Vector(i, i))).toList shouldBe
+        List(1, 3, 6, 10, 15, 15)
+      (Stream(): Stream[Nothing, String]).repartition(_ => Segment.empty).toList shouldBe List()
```
SystemFw: This can probably be `Stream.empty.covaryOutput[String]` or `Stream(): Stream[Pure, String]`

SystemFw:
Is this intended behaviour?

```scala
scala> import cats.implicits._

scala> def input = Stream("a", "b").repeat
input: fs2.Stream[fs2.Pure,String]

scala> def ones(s: String) = Segment vector s.grouped(1).toVector
ones: (s: String)fs2.Segment[String,Unit]

// OK

scala> input.take(2).repartition(foo).toVector
res8: Vector[String] = Vector(a, b)

scala> input.repartition(foo).take(2).toVector
res9: Vector[String] = Vector(a, b)

// Not what I'd expect
scala> input.take(10).repartition(foo).toVector
res10: Vector[String] = Vector(a, b)

scala> input.repartition(foo).take(10).toVector
res11: Vector[String] = Vector(a, b)

// works again without streaming
scala> Stream.emits(input.take(10).toVector).repartition(foo).toVector
res12: Vector[String] = Vector(a, b, a, b, a, b, a, b, a, b)
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1984,6 +1984,42 @@ object Stream {
...
+      * scala> import cats.implicits._
+      * scala> Stream("Hel", "l", "o Wor", "ld").repartition(s => Segment.array(s.split(" "))).toList
+      * res0: List[String] = List(Hello, World)
+      * }}}
+      */
+    def repartition(p: O => Segment[O, Unit])(implicit S: Semigroup[O]): Stream[F,O] = {
+      def go(carry: Option[O], s: Stream[F,O]): Pull[F,O,Unit] = {
+        s.pull.uncons.flatMap {
+          case Some((ohd, otl)) =>
+            ohd.uncons1 match {
```
mpilquist: Can you try using `Segment#scan` or `Segment#fold` here instead of doing an `uncons1` and pushing the remainder back in to `s`? We've hit a lot of performance issues due to patterns resembling an uncons followed by a push back.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1984,6 +1984,42 @@ object Stream {
...
+              case Right((hhd,htl)) =>
+                val tl = Stream.segment(htl) ++ otl
+                val next = carry.fold(hhd)(c => S.combine(c, hhd))
+                val parts = p(next).toVector
```
mpilquist: We should avoid using `toVector` here (e.g., b/c the segment could be infinite). One option is changing the signature of `repartition` to `def repartition(p: O => Chunk[O])(implicit S: Semigroup[O]): Stream[F,O]`. I'd be better to keep the current signature and use one of the lazy/fused combinators on `Segment` though. Let me know how it goes or if you get stuck -- I think we should be able to get it to work.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1984,6 +1984,42 @@ object Stream {
...
+                parts.size match {
+                  case 0 => go(None, tl)
+                  case 1 => go(Some(parts.head), tl)
+                  case _ => Pull.output(Chunk.indexedSeq(parts.init)) >> go(Some(parts.last), tl)
+                }
+              case Left(_) =>
+                carry.fold[Pull[F,O,Unit]](Pull.done)(c => Pull.output1(c))
```
mpilquist: If we reimplement this to avoid the `uncons1`, then you can ignore the following comment. However, I don't think we want to end the pull here -- we'd hit this case if an empty segment is emitted. I think the right thing to do here is to recurse, carrying over the carry in to the recursive call.

soumyarupsarkar: You're right, it should skip the segment in this case.

core/shared/src/main/scala/fs2/Stream.scala
soumyarupsarkar: Done, with a test case with an infinite segment. I had some trouble getting this to work with `uncons` instead of `uncons1` because `repartition` applies `p` to `carry ++ head`, and needs to emit all partitions before determining the next `carry`.

Do you have any suggestions on how to avoid `uncons1`? Thanks!

core/jvm/src/test/scala/fs2/StreamSpec.scala
soumyarupsarkar: Thanks! I think the surprising behavior you found was due to ending the pull too early. I've added some of those cases as test cases. Let me know if the behavior now makes sense.

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: General idea is to split the functionality of `repartition` up in to 2 layers -- an outer layer which does `s.pull.uncons.flatMap(...)` and an inner layer which statefully scans each segment. The segment scan state would be the current carry value. Each step of the scan would output an `Option[O]`.

Roughly, this would look something like:

```scala
def repartition(p: O => Segment[O,Unit])(implicit S: Semigroup[O]): Stream[F,O] = {
  def process(carry: Option[O], s: Segment[O,Unit]): Segment[O,Option[O]] = {
    s.scan(carry)((carry, o) => /* TODO compute an output value using carry and o */).collect { case Some(o) => o }
  }
  def go(carry: Option[O], s: Stream[F,O]): Pull[F,O,Unit] = {
    s.pull.uncons.flatMap {
      case None => carry.map(Pull.output1(_)).getOrElse(Pull.done)
      case Some((hd,tl)) =>
        Pull.segment(process(carry, hd)).flatMap { newCarry => go(newCarry, tl) }
    }
  }
  go(None, this)
}
```

soumyarupsarkar: Right now, what we do is effectively: given an input element and a carry, partition `carry |+| input`, and emit `partitions.init`; use `partitions.last` as the carry when the next input comes in.

In the body of the scan, how do we emit all finished partitions (`partitions.init`) while threading through the carry (`partitions.last`) for each element? Currently this is done by recursively returning pulls to output finished partitions until we get to the end, where we make a recursive call with a carry (the unfinished partition). I'm not sure how to do this with `Segment`.

PR #952, opened 2017-10-16
Make the constructor of Ref private. Fixes #951

### LINKED ISSUE #951: Public constructor of Ref allows breaking referential transparency
The expression `new Ref[IO, Int]` is not pure. Making the constructor private would force users to use the pure "equivalent": `Ref.uninitialized[IO, Int]`.

SystemFw:
Fixes #951

### DIFF
--- a/core/shared/src/main/scala/fs2/async/Ref.scala
+++ b/core/shared/src/main/scala/fs2/async/Ref.scala
@@ -15,7 +15,7 @@ import java.util.concurrent.atomic.{AtomicBoolean,AtomicReference}
 import Ref._
 
 /** An asynchronous, concurrent mutable reference. */
-final class Ref[F[_],A](implicit F: Effect[F], ec: ExecutionContext) { self =>
+final class Ref[F[_],A] private[fs2] (implicit F: Effect[F], ec: ExecutionContext) { self =>
 
   private var result: Either[Throwable,A] = null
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/async/Ref.scala
```diff
@@ -15,7 +15,7 @@ import java.util.concurrent.atomic.{AtomicBoolean,AtomicReference}
 import Ref._
 
 /** An asynchronous, concurrent mutable reference. */
-final class Ref[F[_],A](implicit F: Effect[F], ec: ExecutionContext) { self =>
+final class Ref[F[_],A] private[fs2] (implicit F: Effect[F], ec: ExecutionContext) { self =>
```
mpilquist: We could make this `private` instead and change async.scala to use `Ref.uninitialized`

SystemFw: That was my initial thought, then worried about the extra method call. Arguably over optimising though. (could also do `private[async]`).
Should I change it as you suggest?

SystemFw: scratch it, `private[async]` won't work since it's used in Scope as well

mpilquist: Ah right, merging then

Issue #892, opened 2017-07-04
Caching unrolled ViewL inside FreeC results in memory leak


mpilquist:
Unrolling a `FreeC` is accomplished by calling `viewL`. In theory, the unrolled structure can be cached inside the `FreeC` instance so that any subsequent usage of the same `FreeC` results in a lookup instead of unrolling.

We implemented this in early 0.10.0 milestones (e.g., 0.10.0-M2). However, we found that it results in a memory leak. As a result, we removed the caching in https://github.com/functional-streams-for-scala/fs2/pull/887.

Note: removing the caching had no detectable performance impacts on the test suite. I suspect that the use cases in which caching pays off are rare when working with `Stream`.

To see the leak, run `ConstantEvalSanityTest` from `coreJVM/test:run` with a version of the code prior to merge of 887 (e.g., f0c880c4fad6a4d53f263983242d20ede429b9e4).

When debugging this leak, we tried the following things to no avail:
 - Nulled out the reference to the original stream in `runFold_` to ensure nothing was closing over `stream`
 - Nulled out `viewL` cache on calls to `flatMap`, `map`, `onError`

I'm happy to add caching back if someone (who is not me) fixes the leak. Before doing so, I'd like to see some example stream computations which benefit from the FreeC caching though.

mpilquist:
If we ever add `viewL` caching again, we'll need to add the `Suspend` constructor back to `Algebra` -- reverting #953.

pchlupacek:
I am not sure if this is relevant with recent changes to viewL any thoughts @mpilquist?

mpilquist:
Agreed

PR #953, opened 2017-10-19
Removed the Suspend node from the Stream algebra

mpilquist:
We added the Suspend node back when we were trying to implement viewL
caching in FreeC. If viewL caches, then a Free.suspend(f) may be
evaluated and cached. However, suspend on Stream/Pull/Algebra relies on
always-evaluate semantics.

Now that we don't cache in FreeC, there's no reason for the Stream
algebra to have its own Suspend constructor.


SystemFw:
Just a curiosity since I wasn't around at the time, did we give up on FreeC caching because it couldn't be implemented nicely, or it's just not worth it?

mpilquist:
@SystemFw I gave up on it due to the memory leak in #892. I linked that issue to this one though, so if anyone ever fixes #892, we'll remember to restore a `Suspend` constructor.

SystemFw:
Makes sense, thank you

PR #954, opened 2017-10-21
Added Segment.unfoldChunk and changed Segment#run so that it requires an output type of Nothing

mpilquist:


core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -590,14 +590,14 @@ abstract class Segment[+O,+R] { self =>
 
   /**
    * Computes the result of this segment. May only be called when `O` is `Unit`, to prevent accidentally ignoring
```
SystemFw: the comment is out of date, it should read "when `O` is `Nothing`"

mpilquist: Thanks, fixed

PR #956, opened 2017-10-22
Fix Stream#runSync to use `Sync` instead of `Effect`

guersam:
The implementation of `Stream#runSync` does not match its documentation. Given the background which introduced this method, I guess we have to rewrite it.


SystemFw:
Thanks for spotting and fixing this!

PR #959, opened 2017-10-25
Fixed performance issue in Stream.flatMap exposed by streams like Stream.emits(...).flatMap(Stream(_))

### LINKED ISSUE #958: Very expensive flatMap() on streams backed by large chunks
`flatMap()` invocation on streams backed by large chunks (chunk size in the order of 1000) are extremely expensive. The following example illustrates the problem.

```scala
/* Using using scala 2.11.11 and fs2 version 0.10.0-M7 */

Stream.emits(0 until 20000).flatMap(Stream(_)).toVector  // Takes approx 25 seconds (!)
Stream.range(0, 20000).flatMap(Stream(_)).toVector  // Takes approx. 50 milliseconds
```

mpilquist:
Fixes #958.

Before this fix:

```
[info] Benchmark                                   Mode  Cnt           Score   Error  Units
[info] StreamBenchmark.emitsThenFlatMap_1          avgt             2527.329          ns/op
[info] StreamBenchmark.emitsThenFlatMap_10         avgt            34048.764          ns/op
[info] StreamBenchmark.emitsThenFlatMap_100        avgt          1298487.894          ns/op
[info] StreamBenchmark.emitsThenFlatMap_1000       avgt         79365209.923          ns/op
[info] StreamBenchmark.emitsThenFlatMap_10000      avgt       6324507135.000          ns/op
```

After this fix:

```
[info] Benchmark                                   Mode  Cnt        Score   Error  Units
[info] StreamBenchmark.emitsThenFlatMap_1          avgt          2311.262          ns/op
[info] StreamBenchmark.emitsThenFlatMap_10         avgt         10316.613          ns/op
[info] StreamBenchmark.emitsThenFlatMap_100        avgt         76033.740          ns/op
[info] StreamBenchmark.emitsThenFlatMap_1000       avgt        588334.872          ns/op
[info] StreamBenchmark.emitsThenFlatMap_10000      avgt       9377237.551          ns/op
```

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/StreamBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/StreamBenchmark.scala
@@ -2,7 +2,8 @@ package fs2
 package benchmark
 
 import cats.effect.IO
-import org.openjdk.jmh.annotations.{Benchmark, State, Scope}
+import org.openjdk.jmh.annotations.{Benchmark, BenchmarkMode, Mode, OutputTimeUnit, State, Scope}
+import java.util.concurrent.TimeUnit
 
 @State(Scope.Thread)
 class StreamBenchmark {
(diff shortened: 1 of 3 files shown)


mpilquist:
Main issue here is that `foldRightLazy` was implemented in terms of `unconsChunk` which would call `unconsChunks`, take the head chunk and push the remainder back. As a result, the `Catenable` backing the `Segment` ended up getting used as a stack, where elements are continuously pushed/popped. Fix is relatively simple -- `foldRightLazy` is now implemented in terms of `unconsChunks`.

Issue #961, opened 2017-10-27
>> vs *>


mpilquist:
`>>` is deprecated in Cats 1.0 in favor of `*>` (https://github.com/typelevel/cats/pull/1955). Should we follow suit on `Stream`, `Pull`, etc?

ChristopherDavenport:
I have already started switching to those semantics where available. So I would be all for consistent syntax.

mpilquist:
Fixed by #963

PR #963, opened 2017-11-01
Upgrade cats-effect

aeons:
This depends on a currently unpublished cats-effect 0.5, which in turn depends on cats 1.0-RC1.

As soon as cats-effect 0.5 is released, this should be good to go.


SystemFw:
Note: this also addresses #961

PR #964, opened 2017-11-01
Fixed memory leak in scope management.

### LINKED ISSUE #962: Big memory leak in 0.10.x
Hello.

I must say first that although I'm not a complete newbie in functional programming and FS2, I'm also not fully proficient at this and I hope there is no obvious mistake on my side.

I recently implemented several microservices with FS2 0.10.x that rely on Kafka topics to communicate, and when I started deploying them and let them run for some time, they all blew up in my face with a nice 'GC overhead exceeded' error.

After a lot of investigation, I managed to reduce the issue to this small piece of code.
```scala
import cats.effect.IO
import fs2.Stream

object Fs2Test extends App {
  Stream.eval(IO(Thread.sleep(1L)))
    .flatMap(_ => Stream.emits(Seq()))
    .map(x => x)
    .repeat
    .run
    .unsafeRunSync()
}

```

Here, the `Thread.sleep(1L)` is the equivalent of the polling call on the Kafka consumer. I put 1 ms here just to make the issue show up faster. The `flatMap(_ => Stream.emits(Seq()))` is where I flatten the bunch of consumer records to have them emitted one by one (and here I simulate the most common case where there is no new message), and the `map(x => x)` is where I JSON-decode the records.

_Important note_: the problem does not show up when the call to `map` is not present.

_Other importante note_: the problem does not exist with the following equivalent code in 0.9.7.
```scala
import fs2.{Strategy, Stream, Task}

object Fs2Test extends App {
  implicit val S: Strategy = Strategy.fromFixedDaemonPool(8)

  Stream.eval(Task.delay(Thread.sleep(1L)))
    .flatMap(_ => Stream.emits(Seq()))
    .map(x => x)
    .repeat
    .run
    .unsafeRun()
}
```

Taking two heap dumps in JVisualVM several seconds apart shows the following in version 0.10.0-M7.
Heap dump diff for 0.10.0-M7

In version 0.9.7, a heap dump look like this. As you can see, classes from fs2 are not even in the top 100 (as expected).
Heap dump for 0.9.7

All my attempts always displayed the same pattern: the top six classes you see in the first heap dump accumulate the same number of instances over time. Note also that the `LinkedHashMap` instances seem to be all empty.

Obviously all these instances reference each other, and analyzing the reference graph of, say, `fs2.Catenable$Singleton` shows the following.
Reference graph for fs2.Catenable$Singleton

Digging into the code, I tried to find the references to the constructor of `fs2.Catenable$Append`, and ended up in an `uncons` method somewhere I don't remember. From what I know, `uncons`-like methods are used a lot in the code base, and I'm pretty sure there are many other situations like this one that could lead to this issue. If this is true, it seems to me that in general, version 0.10.x in its current state is not suitable for long-lived streams.

As a side note, I noticed [this Stack Overflow question](https://stackoverflow.com/questions/46912745/expensive-flatmap-operation-on-streams-originating-from-stream-emits) that seems to mention this exact issue, although the answer only stresses on a performance problem.

If there is any obvious workaround for my case (and unless I'm wrong of course), I would be glad to read it from you guys, otherwise I'm doomed to downgrade my codebase to version 0.9.7, knowing that I'll have to upgrade again later.

Thanks in advance for your help, and congratulations for this amazing lib :)!

mpilquist:
When a scope is closed, it should be unregistered from its parent scope
so that it is eligible for GC. This PR implements this by tracking
spawned scopes via a token based LinkedHashMap and upon scope closure,
removing the entry in the parent's map.

Fixes #962.

### DIFF
--- a/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
+++ b/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
@@ -123,3 +123,10 @@ object StepperSanityTest2 extends App {
   }
   go(0)(Pipe.stepper(_.map(_ + 1)))
 }
+
+object EvalFlatMapMapTest extends App {
+  Stream.eval(IO(())).
+    flatMap(_ => Stream.emits(Seq())).
+    map(x => x).
+    repeat.run.unsafeRunSync()
+}
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -71,14 +71,14 @@ private[fs2] object Algebra {
 
   final class Token
 
-  final class Scope[F[_]] private (private val parent: Option[Scope[F]])(implicit F: Sync[F]) {
+  final class Scope[F[_]] private (private val id: Token, private val parent: Option[Scope[F]])(implicit F: Sync[F]) {
     private val monitor = this
     private var closing: Boolean = false
     private var closed: Boolean = false
     private var midAcquires: Int = 0
     private var midAcquiresDone: () => Unit = () => ()
     private val resources: LinkedHashMap[Token, F[Unit]] = new LinkedHashMap[Token, F[Unit]]()
-    private var spawns: Catenable[Scope[F]] = Catenable.empty
+    private val spawns: LinkedHashMap[Token, Scope[F]] = new LinkedHashMap[Token, Scope[F]]()
 
     def isClosed: Boolean = monitor.synchronized { closed }
 
@@ -109,8 +109,9 @@ private[fs2] object Algebra {
       val spawn = monitor.synchronized {
         if (closing || closed) None
         else {
-          val spawn = new Scope[F](Some(this))
-          spawns = spawns :+ spawn
+          val spawnId = new Token()
+          val spawn = new Scope[F](spawnId, Some(this))
+          spawns += (spawnId -> spawn)
           Some(spawn)
         }
       }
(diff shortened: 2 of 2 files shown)


core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
```diff
@@ -123,3 +123,10 @@ object StepperSanityTest2 extends App {
   }
   go(0)(Pipe.stepper(_.map(_ + 1)))
 }
+
+object EvalFlatMapMapTest extends App {
+  Stream.eval(IO(())).
```
aeons: `IO.unit` is easier to read imo :)

mpilquist:
@SystemFw I generally debug these types of issues by adding a test to `MemorySanityChecks` and running with `test:run` with `jvisualvm` attached and sampling memory usage. A test case which can reproduce the leak within a few seconds is important. Once I have that, I then bounce around a bit, looking at heap dumps, heap diffs (often just by using VisualVM's sampler delta feature), and changing the test case in various ways to see how memory profile is impacted. If I'm not making progress doing this, I'll load heap dumps in to Eclipse MAT and use various reports to explore the heap.

In this case, the error jumped out pretty quickly when looking at the memory sampler deltas in VisualVM. There were a few hundred thousand instances of `Scope`. The reason the bug was introduced by the addition of `map(x => x)` is that `map` is implemented with a pull, and when we convert a `Pull` to a `Stream`, we insert a scope (I might want to change this in the future...). So each time the stream repeated, it would do another `map` which would introduce a new scope. We also had a huge number of `Catenable` append nodes. Looking at `Scope`, I realized the `spawns` collection was stored as a `Catenable[Scope[F]]`, and then realized that as spawned scopes were being closed, they weren't being removed from their parent's `spawns` collection.

SystemFw:
@mpilquist Fantastic writeup, thank you

PR #965, opened 2017-11-01
Topic/crypto 0.9

pchlupacek:
@mpilquist @SystemFw @djspiewak @pchiusano TLS support for 0.9 series, 0.10 will be ported once that will gets in :-)


pchlupacek:
@mpilquist friendly reminder on this one. I will need to publish fs2 with crypto support, as some of the projects I am working on rely on this build. Could you please take a quick look on this so we can decide if it is go/no go ?

pchlupacek:
as agreed with @mpilquist we will be not adding this to core library now. so for those that are interested, TLS Support will live here: https://github.com/Spinoco/fs2-crypto

SystemFw:
@pchlupacek let's add `fs2-crypto` to the README though :)

pchlupacek:
yes, when the build will be finished, I'll update it

Issue #967, opened 2017-11-02
join and joinUnbounded release resources too early


mpilquist:
Consider:

```scala
Stream.bracket(IO(println("acquired")))(_ => Stream.emit(()), _ => IO(println("released"))).
  map(_ => Stream.eval(IO(Thread.sleep(500)) *> IO(println("using")))).
  joinUnbounded.run
```

Running this program results in:
```
acquired
released
using
```

Replacing `joinUnbounded` with `flatMap(identity)` results in the correct order.

Originally spotted by @tpolecat in `SocketSpec` where replacing `Stream.emit(s.flatMap(f))` with `s.map(f)` resulted in errors indicating the channel had been closed.

mpilquist:
@pchlupacek Seems like the `runOuter` method in `join` is missing an `onFinalizer` that blocks on the inners completing. WDYT?

pchlupacek:
@mpilquist take a look on that. I am just not 100% sure, but I recall fixing this in 0.9.x branch....

mpilquist:
I added a failing test case in #968. I spot checked 0.9 and it appears to have the same issue.

mpilquist:
Given `s: Stream[F,A]` and `f: A => Stream[F,B]`, resources allocated in scope of a single `A` must not be released until the inner `Stream[F,B]` completes. There are at least 2 cases to consider here: a resource allocated in process of generating a single `A` and resources allocated in generation of some/all `A`s in `s`. A client socket is an example of the first case whereas the bind socket is an example of the second case. The implementation should preserve the property that resources are freed eagerly (e.g., in the TCP example, releasing client sockets shouldn't happen at the end of the outer stream but rather as each inner stream completes).

I don't know if the current join implementation can be adapted to meet these requirements without some new APIs that let it manipulate scopes.

pchlupacek:
@mpilquist Thanks & I'll take a look on this. I have feeling that resource thing can be resolved by prohibiting the outer stream to terminate before inner streams will. 

I just need to think in context of interruption what that would that mean, and if we wouldn't end up wit some deadlock.

mpilquist:
@pchlupacek I have a prototype of a fix for this issue which involves adding a reference count to each resource and adding 2 new cases to the stream algebra -- one that exports all resources, incrementing their reference counts and another that imports all resources in to the current scope. The test case passes with this change (and with no other synchronization added to join).

mpilquist:
[...]

PR #968, opened 2017-11-02
Added test case showing that join/joinUnbounded release resources from outer stream too early

mpilquist:
See #967

### DIFF
--- a/core/jvm/src/test/scala/fs2/MergeJoinSpec.scala
+++ b/core/jvm/src/test/scala/fs2/MergeJoinSpec.scala
@@ -44,6 +44,15 @@ class MergeJoinSpec extends Fs2Spec {
       runLog { s1.get.flatMap(_.get) }.toSet
     }
 
+    "join - resources acquired in outer stream are released after inner streams complete" in {
+      val bracketed = Stream.bracket(IO(new java.util.concurrent.atomic.AtomicBoolean(true)))(Stream(_), b => IO(b.set(false)))
+      // Starts an inner stream which fails if the resource b is finalized
+      val s: Stream[IO,Stream[IO,Unit]] = bracketed.map { b =>
+        Stream.eval(IO(b.get)).flatMap(b => if (b) Stream(()) else Stream.fail(Err)).repeat.take(10000)
+      }
+      s.joinUnbounded.run.unsafeRunSync()
+    }
+
     "merge (left/right failure)" in forAll { (s1: PureStream[Int], f: Failure) =>
       an[Err.type] should be thrownBy {
         s1.get.merge(f.get).run.unsafeRunSync()
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1255,6 +1255,9 @@ object Stream {
     go(0)
   }
 
+  private[fs2] def exportResources[F[_]]: Stream[F,List[Algebra.Resource[F]]] =
+    Stream.fromFreeC(Algebra.exportResources[F,List[Algebra.Resource[F]]].flatMap(Algebra.output1(_)))
+
   /**
    * Creates a stream that, when run, fails with the supplied exception.
    *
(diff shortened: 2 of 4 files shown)


mpilquist:
I added a commit which fixes this test case. Since it adds 2 new primitives to the stream algebra, I'd really like to get @pchiusano to weigh in.

General idea behind the fix is the introduction of reference counting on resources along with 2 new primitives -- `ExportResources` and `ImportResources`. When a scope is closed, the reference count of each resource is decremented (via `maybeRelease`). If the count reaches 0, the finalizer is invoked. The reference count is initialized to 1.

Reference counts are incremented by using an `ExportResources` step, which returns *all* resources in the stream's scope and increments the reference count of each. We can then later use `ImportResources` on a different stream, which pulls those resources in to the scope of that stream.

pchlupacek:
@mpilquist I like the idea of reference counting on resources, perhaps that may simplify reasoning behind Scopes too. I'll take a look on the code.

pchlupacek:
@mpilquist obviously you were right, including the proposed fix. so unless @pchiusano has some objections to it, I suggest this will gets in as it is.

pchlupacek:
Hm, on the other thought. Isn't the `import`/`export` thing too much low-level? Essentially we need just manually signal when acquired scope shall be released manually to outlive stream closure. However I would like still to keep it within single stream, that owning the scope, instead of having user to manage it and resources manually. 

  So I was just thinking to have something like this perhaps : 
```scala
// on Stream 

/** 
 * Allows to manually close the resource scope of the stream. 
 * Guarantees that any resources are not frees until `f` finishes its evaluation. 
 * Note however, that this applies only if the stream terminated normally. In case the stream failed or was 
 * terminated, this will not be consulted, and resources are released immediately.      
 **/
def confirmRelease(f:  F[Unit]) : Stream[F, O]  = ???   


// Then we can perhaps use it simply on any stream like 
val outerStream: Stream[F, O] = 
   source.map { inner =>
    runInner(inner)
   } 
  .confirmRelease(awaitAllInnerStreamsToTerminate)
  .onFinalize(...)
  .onError(...)

```

Then I think that the implementation perhaps may be to add to Scope's constructor this `confirm`

```final class Scope[F[_]] private (private val id: Token, private val parent: Option[Scope[F]], private val confirm: Option[F[Unit]])(implicit F: Sync[F]) ```

and then when you do `closeAndReturnFinalizers` consult whether stream was failed or interrupted, and in that case just ignore `confirm` otherwise await confirm to complete `before` running finalizers. 

This is just rough idea. What do you think ?

mpilquist:
@pchlupacek I think that's basically what `onFinalize` does except it would register the finalizer on all scopes at or below the current scope. It's not clear to me if that will work or not though.

One thing I like about exporting resources is that we can extend it to new APIs that allow users to selectively export a resource, effectively taking over its lifecycle. No need for this now though.

pchiusano:
Hi sorry for delayed reply, I am really swamped at the moment. My feeling is: the join/joinUnbounded functions should be fixed, and maybe for now that is done via these new stream algebra primitives, but I'm more hesitant on exposing these primitives publicly unless we are super confident they are the right approach. I haven't spent nearly as much time looking at this as y'all though, so I may just not be understanding the problem well enough. @mpilquist and @pchlupacek you should decide. :)

mpilquist:
@pchlupacek For all `s: Stream[F,A]`, `f: A => Stream[F,B]`, finalization of resources in `s.map(f).join(n)` should be the same as `s.flatMap(f)`, regardless of failures in outer stream.

So if we went with the `confirmResources` approach, I think we'd have to invoke it no matter what (just like a finalizer) -- the difference between it and `onFinalize` is that it would register as a finalizer on the deepest right-most scope in the current scope tree whereas `onFinalize` simply registers on the current scope.

pchlupacek:
@mpilquist I agree with what you say. Also you are right about the fact that this shall be the seam behaviour whether stream is killed failed etc... 

Lets then merge this in, as it fixes the issues and offers great flexibility.

PR #970, opened 2017-11-06
Fixed 2 bugs in scope management

mpilquist:
Scope#releaseResource only searched the local resources map for a
resource, so if the resource was in a spawn's map or a parent map, the
release was skipped. This masked a bug in the new join import/export
support, which treated explicit resource releases as forced releases,
skipping reference counting.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -82,11 +82,9 @@ private[fs2] object Algebra {
 
   sealed trait Resource[F[_]] { self =>
     private[Algebra] def maybeRelease: F[Unit]
-    private[Algebra] def forceRelease: F[Unit]
     private[Algebra] def increment: F[Unit]
     private[Algebra] def translate[G[_]](u: F ~> G): Resource[G] = new Resource[G] {
       def maybeRelease = u(self.maybeRelease)
-      def forceRelease = u(self.forceRelease)
       def increment = u(self.increment)
     }
   }
@@ -100,13 +98,6 @@ private[fs2] object Algebra {
             case 0L => finalizer
             case _ => F.unit
           }
-        }
-        def forceRelease: F[Unit] = {
-          def loop: F[Unit] = F.delay(count.get).flatMap(cnt =>
-            if(cnt > 0) F.delay(count.compareAndSet(cnt, 0)).ifM(finalizer, loop)
-            else F.unit
-          )
-          loop
         }
         def increment: F[Unit] = F.delay {
           val cnt = count.incrementAndGet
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -239,11 +246,7 @@ private[fs2] object Algebra {
         val (spawns, resources) = s.monitor.synchronized(s.spawns.map(_._2).toList -> s.resources.map(_._2).toList)
         spawns.map(down).foldLeft(Catenable.empty: Catenable[Resource[F]])(_ ++ _) ++ Catenable.fromSeq(resources)
       }
-      def root(s: Scope[F]): Scope[F] = s.parent match {
-        case Some(p) => root(p)
-        case None => s
-      }
-      val all = down(root(this)).toList
+      val all = down(root).toList
```
pchlupacek: I am really worried by this. Essentially what worries me;
     - is that we are taking all resources from the stream at time when export resources was called
     - are we 100% sure there cannot be midAcquires at that time, 
     - or worse that some of that resources did not already begun its finalization due the monitor lock being placed only on #246 line?

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -147,8 +138,24 @@ private[fs2] object Algebra {
       if (midAcquires == 0) midAcquiresDone()
     }
 
-    def releaseResource(t: Token): Option[F[Unit]] = monitor.synchronized {
-      resources.remove(t).map(_.forceRelease)
+    /**
```
pchlupacek: the implementation is not stack safe. 
Also michael I think we have to clarify how the scopes and spawns work really.

pchlupacek:
Leaved few comments. I have still not a good feeling about synchronisation we doing there and whole concept of scopes.  Will think about it for a while. However if you feel good with this, lets merge this in, and then lets see if we can somehow improve that whole scope thing. I have something drafted already, but not that happy so far, so may need some time to think about it before submitting that...

mpilquist:
@pchlupacek Hm, I don't see any comments. Are you sure you submitted some?

Note that in 0.9, we managed scopes and resources using those internal refs. Paul explicitly decided that was too complicated during the 0.10 design work and replaced it all with a `ConcurrentMap`. But then that wasn't strong enough of a construct, so we ended up with `Scope`, which is really a rose tree of concurrent ordered maps.

mpilquist:
@pchlupacek Thanks, they just showed up

core/shared/src/main/scala/fs2/internal/Algebra.scala
mpilquist: Any recommendations for fixing stack safety? We could trampoline through F I suppose.

core/shared/src/main/scala/fs2/internal/Algebra.scala
mpilquist: We discussed some of this on Gitter. It's OK if there are mid acquires b/c if there are, that means there's an unconsAsync executing in the stream, which means the mid-acquired resources cannot be visible to the current step of execution.

If a resource has already begun finalization, when `increment` is called, it will no-op.

pchlupacek:
Michal let me sumarize how I got the scopes to work in 0.10 branch: 

OPENING: 
--------

There is one single place where scopes are closed and opened essentially: 
Algebra#58: 
``` scala
def scope[F[_],O,R](pull: FreeC[Algebra[F,O,?],R]): FreeC[Algebra[F,O,?],R] =
    openScope flatMap { newScope =>
      FreeC.Bind(pull, (e: Either[Throwable,R]) => e match {
        case Left(e) => closeScope(newScope) flatMap { _ => fail(e) }
        case Right(r) => closeScope(newScope) map { _ => r }
      })
    }
```

That essentially (for non-concurrent // merged streams) will create sequential linked list in current scope implementation with Scope.root at the `root`. So essentially I view them in this case as sort of `stack`, i.e. : 
Open S1 -> OPEN S2 -> OPEN S3 -> CLOSE S3 -> CLOSE S2 -> CLOSE S1. Because of the Algebra.scope implementation, I don't see any other way how they can be closed/opened except like that. This also assumes, that S2 will not be acquired before s1 will finish all resource midAcquires. 

However for concurrent streams, I see that this may get bit trickier. There may be concurrent branches being constructed where merge is happening , i.e. 
OPEN S1 -> [OPEN S2.1, OPEN S2.2] -> [CLOSE S2.2, CLOSE S2.1] -> CLOSE S1. Where order in which S2.1 and S2.2 are opened and closed is non deterministic, but they still will not be opened before S1 will be fully acquired (including mid acquires) and won't be closed after S1 will be closed.  

Can you perhaps just quickly look at that if I understood way how the scopes work now correctly? 

Thanks a lot.

mpilquist:
@pchlupacek Yes, that appears correct to me

pchlupacek:
Ok @mpilquist thanks for confirmation. I have in that case idea how to suggest the scope representation to be more readable and safer. lets get this in, and I will open another PR ok ? 
Last but not at least can you perhaps clarify why we need to traverse the resources all they way up and then all the way down through all the concurrent leafs? i.e. imagine I am the guy in S2.1, why I need to take care of any resources from S2.2?

mpilquist:
Sounds good.

In the test case I added for this, the scope tree was something like: S2.{1,2} with current scope S2 when exporting but with the acquired resource in S2.1 or S2.2. So somehow, we ended up with the current scope set to S2 despite S2.1 and S2.2 both being open still. I don't exactly follow why the scopes aren't acting as a perfect stack here.

pchlupacek:
:-) nice. I bet some of these may be due fact that scope's manipulation escapes `F` (i.e. when the heck the finishClose will actually get run having the monitor set there on 3 places? ). I have created simple SyncRef, that essentially only wraps atomicRef and provides synchronous ops on Ref, and with that I think we can make Scope to be much easier to reason about.

mpilquist:
It's all yours -- looking forward to seeing what you come up with. :)

PR #972, opened 2017-11-08
Make Sync instances for Stream and Pull implicit

mpilquist:
If you `import cats.implicits._`, the cats syntax for methods like `flatMap`, `*>`, and `handleErrorWith` overrides the methods from `Stream`. The cats versions of these methods don't support all the variance tricks that the FS2 equivalents support. Hence, users may need to use `covary` in more cases than they'd normally need to if they are importing cats syntax.

This exposed some inconsistencies with cats methods -- e.g., `onError` in cats means something different than `onError` in fs2, so I renamed `onError` to `handleErrorWith` to match the cats name. Similarly, I renamed `fail` to `raiseError` to match the cats name.

Note that the `Pull` API doesn't need the same variance tricks that `Stream` uses because (1) we don't create `Pull[Pure,O,R]` instances out of thin air (if we have one, it's because we are pulling on a pure stream) and (2) the pull constructors are all polymorphic in F.

### DIFF
--- a/core/jvm/src/main/scala/fs2/compress.scala
+++ b/core/jvm/src/main/scala/fs2/compress.scala
@@ -96,7 +96,7 @@ object compress {
     }
   }
   private def _inflate_finish[F[_]](inflater: Inflater): Pull[F, Nothing, Unit] = {
-    if (!inflater.finished) Pull.fail(new DataFormatException("Insufficient data"))
+    if (!inflater.finished) Pull.raiseError(new DataFormatException("Insufficient data"))
     else { inflater.end(); Pull.done }
   }
 }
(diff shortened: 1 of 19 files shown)


SystemFw:
I think we should consider having those instances in an instances import rather then the companion object, so that users that are happy with the status quo can still `import cats.implicits._` without being affected, at least until we figure out how painful the extra `covary`s will be.

I often have cats code (which requires cats syntax), and fs2 code (on which `covary` is annoying) in the same file. I do however approve of having the instances as implicits _somewhere_ to avoid replicating cats functionality (for me, `void` has been the most common deficiency when working with Stream)

tpolecat:
I would also be fine with an extra import.

ChristopherDavenport:
Here is all modifications required for http4s - Only main currently, will likely expand
https://github.com/http4s/http4s/pull/1534

SystemFw:
@mpilquist They do change my opinion to a large extent. I guess we will only see the real pain when more people start using it (especially those that aren't experienced enough to know where to put a `covary`), but maybe we can assume it won't be painful (and therefore put them in the companion), and perhaps change it later if it's actually a problem?

I agree with removing `fail`, given how much is needed to switch to 0.10 anyway

SystemFw:
Ah, thanks again @ChristopherDavenport for doing this

mpilquist:
Alright, I'm going to go ahead and merge this then. We can see how it goes after the next milestone build and if it ends up too painful, we can drop the implicit keyword from the instances and introduce an implicits object.

Issue #916, opened 2017-08-07
Add primitive support for interruption


mpilquist:
In the PR review of #914, @pchlupacek pointed out that `interruptWhen` in 0.9 and 0.10 cannot interrupt all stream types. The issue is caused by the fact that `interruptWhen` is implemented in terms of `merge`, and downstream from the application of `interruptWhen`, a caller can `flatMap` and turn a single element in to an infinite stream, causing the merger to never be re-evaluated.

```scala
// 0.9.7
Stream.eval(async.signalOf[Task,Boolean](false)).flatMap(s => Stream.eval_(Task.start(s.set(true).schedule(1.second))) ++ Stream.emit(s).interruptWhen(s)).flatMap(s => s.discrete).runLog.unsafeRunAsync(println)
```

Instead of implementing `interruptWhen` in terms of `merge`, we could add primitive interruption support to `runFold`. There are probably a few different ways to do this.

@pchiusano What do you think? Any advice on implementing this?

pchlupacek:
Just as remark note, this not only causes program to not interrupt but also swallows the exceptions in the stream graph causing the whole program to enter into unstable state.

mpilquist:
@pchlupacek I took a shot at implementing the start of this but ran in to some pretty big problems.

I added a `interrupted: AtomicBoolean` parameter to `runFold_` and checked on each invocation of `go`. I also added a new primitive to the algebra: `final case class GetInterrupt[F[_],O]() extends Algebra[F,O,F[Unit]]`

As an initial take on `interruptWhen`, I tried:
```scala
    def interruptWhen(haltWhenTrue: Stream[F,Boolean])(implicit F: Effect[F], ec: ExecutionContext): Stream[F,O] =
      fromFreeC(Algebra.interrupt[F,O].flatMap { interrupt =>
        self.concurrently(haltWhenTrue.takeWhile(!_, true).filter(identity).evalMap(_ => interrupt)).get
      })
```

This definition is problematic though because it uses `concurrently` and the implementation of `concurrently` uses `interruptWhen`. So then I tried a simple test:

```scala
    def interruptWhen(haltWhenTrue: Stream[F,Boolean])(implicit F: Effect[F], ec: ExecutionContext): Stream[F,O] =
      fromFreeC(Algebra.interrupt[F,O].flatMap { interrupt =>
        // TODO handle errors in monitor, terminate monitor at end of self
[...]

PR #974, opened 2017-11-10
Topic/scope

pchlupacek:
@mpilquist, @pchiusano, @SystemFw  
This is an attempt to move Scope to `F`. 

The reasoning behind Scope move to `F` is to prevent any monitors (synchronisations) in fs2 codebase and to improve correctness and reasoning about scopes. 

I also added documentation to Scope for people to get in internals easier. Perhaps one day we shall add this to Algebra too. 

Worth of note : 

- Introduction of SyncRef (internal only) that is essentially `F` wrapper around AtomicRef, providing the simplified API cosistent with `Ref`. It should provide less overhead than actor based `Ref` and just needs `Sync[F]` instance which is available in runFold. Not sure if that shall be public, but perhaps it is useful. 

- instead [Import/Export]Resources we have now possibility to acquire scope as resource to another stream. Doing so, shall prevent that scope to be released until stream that acquired that scope as resource will terminate and thus release its acquisition 

- The mechanism of scope-acquisition is based on midAcquires concept already present, where each acquisition is incrementing the mid-acquire (now renamed to refCount) value. 

- The scope can be now `visible` while evaluating the stream. I am not quite sure if that is useful, but I leaved this now public, perhaps we can make it fs2 private. One of the ideas I was playing here with was to use this to implement interruption of the downstream as discussed here #916 

Still have couple tests failing, so I will work on this during next few days, hopefully resolving all of them. 

I don't expect this to have significant impact on the benchmarks and performance. 

Any initial input is welcome.


core/shared/src/main/scala/fs2/internal/Scope.scala
```diff
@@ -0,0 +1,295 @@
...
+    *
+    * Linking is implemented as incrementing the `refCount` in `this` scope by one. The `remote` side will
+    * then register supplied callback as their normal resource provided returned callback as release of that resource.
+    *
+    * When this yields to None, that indicates that linking was not succesfull, due `this` scope being closed.
+    *
+    * Note that acquisition of the scope is synchronous, so it is safe to await its evaluation.
+    *
+    */
+  def acquire: F[Option[F[Unit]]] = {
```
mpilquist: Can we name this something else? I think the name is too close to the resource management functions above (e.g. beginAcquire)

core/shared/src/main/scala/fs2/internal/Scope.scala
```diff
@@ -0,0 +1,295 @@
...
+    */
+  final private[Scope] case class ScopeState[F[_]](
+    closed: Option[Boolean]
+    , refCount: Int
+    , refReleased: Option[F[Unit]]
+    , resources: ListMap[Token, Resource[F]]
+    , children: ListMap[Token, Scope[F]]
+  ) {
+
+    def open_? : Boolean = closed.isEmpty
```
mpilquist: Style thing: we don't use the `_?` syntax anywhere else in the project so I'd prefer to call these `isOpen`, `isClosed`, `isClosing`, `isClosed`

core/shared/src/main/scala/fs2/internal/Scope.scala
```diff
@@ -0,0 +1,295 @@
...
+    * @param resources          All acquired resources (that means synchronously, or the ones acquired asynchronously) are
+    *                           registered here. Note that the resources are released in reverse order as they were acquired.
+    *
+    * @param children           Children of this scope. Children may appear during the parallel pulls where one scope may
+    *                           split to multiple asynchronously acauired scopes and resources. Still, likewise for resources they are released in reverse order.
+    *
+    *
+    * @tparam F
+    */
+  final private[Scope] case class ScopeState[F[_]](
```
mpilquist: Since this is defined in `Scope` companion, can we name it `State`? Also, no need for the `[Scope]` modifier on private -- `Scope` can see private members of its companion.

core/shared/src/main/scala/fs2/internal/SyncRef.scala
```diff
@@ -0,0 +1,65 @@
...
+    F.delay { go() }
+  }
+
+
+}
+
+
+object SyncRef {
+
+  def apply[F[_] : Sync, A](a: A): SyncRef[F, A] =
```
mpilquist: Should return `F[SyncRef[F,A]]`

core/shared/src/main/scala/fs2/internal/SyncRef.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import scala.annotation.tailrec
+
+/**
+  * Reference to `A` that supports concurrent, but only synchronous operations.
+  *
+  * Think of it as an java AtomicReference wrapped in `F`
+  *
+  */
+final class SyncRef[F[_], A] (private val ar: AtomicReference[A])(implicit F: Sync[F]) {
```
mpilquist: I've wanted this type a number of times and I've seen people re-invent it. Perhaps we should consider putting a more polished version of it in `fs2.async`.

core/shared/src/main/scala/fs2/internal/Scope.scala
```diff
@@ -0,0 +1,295 @@
...
+import cats.implicits._
+import fs2.async.Ref.Change
+import fs2.internal.Scope.ScopeState
+
+import scala.collection.immutable.ListMap
+import scala.concurrent.ExecutionContext
+
+final class Scope[F[_]] private (private val id: Token, private val parent: Option[Scope[F]])(implicit F: Sync[F]) { self =>
+
+  private val state: SyncRef[F, ScopeState[F]] = SyncRef(ScopeState.initial.asInstanceOf[ScopeState[F]])
```
mpilquist: Change `initial` to be polymorphic effect and remove this cast

pchlupacek: hm, I wanted to have this inital as val, so it won't get recreated with every new Scope acquisition. and I wanted SyncRef to be not +F? I may put the initial to be def, and put case somehow there if you think is better style ? Or would you be so kind and suggest something instead?

core/shared/src/main/scala/fs2/internal/Scope.scala
pchlupacek: how about acquireScope?

core/shared/src/main/scala/fs2/internal/Scope.scala
pchlupacek: yes, sure.

core/shared/src/main/scala/fs2/internal/Scope.scala
pchlupacek: ok

core/shared/src/main/scala/fs2/internal/SyncRef.scala
pchlupacek: yeah, is a bit of awkward to put in async :-) beacuse it is sync, but I may just add it there and see if that works ?

core/shared/src/main/scala/fs2/internal/SyncRef.scala
pchlupacek: yeah sure, if we do it public

pchlupacek:
@mpilquist thanks so much. Will keep you updated.

SystemFw:
Not had any time to go over this, but I have a question: does this open the possibility to have a resource safe `F` (e.g.  `bracket` returning `F` rather than `Stream`). Not proposing we should pursue it at this time, just curious

pchlupacek:
@SystemFw just missed your question. I think sort of by definition of IO this is not possible. This change is sort just escaping from imperative code and monitors to `F`.

pchlupacek:
@mpilquist just pushed changes. I sort of reverted to your previous model with ExportResources, however ImportResources is not necessary. Couple of notes: 
- Resource was enhanced to support all the possible release mechanics, which allow in trun
- Greatly simplify the Scope's logic

Also, I think as result of this changes runFold may not need Effect/EC and may be now defined only with Sync. 

However there is still caveat (which doesn't seem to be related to my changes) and that is when code in eval_() fails, in that case somehow the exception is not getting propagates as it should be. This seems to deadlock interpreter, and I am still investigating it. The last point where I see exception is here : 
https://github.com/functional-streams-for-scala/fs2/blob/2a1a254a9241cc7e321f27cbc78f9ba426b5ff33/core/shared/src/main/scala/fs2/internal/Algebra.scala#L214
but not after that. 
I will be investigating that further. But I think this may go in as it is

Also I would like to extract SyncRef by another PR than this one ?

mpilquist:
@pchlupacek Yeah let's do the `SyncRef` PR first as that will be easy to get merged. Let me know how the bug hunting goes.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1299,6 +1291,10 @@ object Stream {
   def iterateEval[F[_],A](start: A)(f: A => F[A]): Stream[F,A] =
     emit(start) ++ eval(f(start)).flatMap(iterateEval(_)(f))
 
+  /** Allows to get current scope during evaluation of the stream **/
+  def getScope[F[_]]: Stream[F, Scope[F]] =
```
mpilquist: Since we're making `Scope` part of the public api here, we'll need a `trait Scope[F[_]]` in the `fs2` package. We can keep the implementation of that trait in `fs2.internal` if that makes sense.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -78,198 +72,10 @@ private[fs2] object Algebra {
...
-        }
-      }
-    }
-
-    @annotation.tailrec def root: Scope[F] = parent match {
-      case Some(p) => p.root
-      case None => this
-    }
+  final class Token {
+    override def toString: String = s"T[${hashCode.toHexString}]"
```
mpilquist: Did you mean to leave this?

core/shared/src/main/scala/fs2/internal/Resource.scala
```diff
@@ -0,0 +1,165 @@
...
+/**
+  * Represent resource acquired during stream interpretation
+  *
+  * Resource is acquired by Algebra.Acquire and then released either by Algebra.Release or Algebra.CloseScope.
+  *
+  * The acquisition of the Resource has three steps:
+  *
+  * 1. Resource is created and registered with the scope (Algebra.Acquire)
+  * 2. Resource is acquired
+  * 3. Resource `acquired` is invoked to conform acquisition of the resource.
```
mpilquist: s/conform/confirm/

core/shared/src/main/scala/fs2/internal/Resource.scala
```diff
@@ -0,0 +1,165 @@
...
+  *
+  * The resource may be released by any of the following methods:
+  *
+  * (1) The Algebra.Resource is interpreted. In this case resource finalization will be invoked if
+  *     the resource was successfully acquired (the `acquired` was evaluated and resource was not leased other scope)
+  * (2) The owning Scope was closed by `Algebra.CloseScope`. This essentially evaluates `release` of
+  *     the Resource and acts like (1).
+  * (3) The `acquired` was evaluated after scope was `released` by either (1) or (2). In this case
+  *     resource will invoke the finalization immediately if the resource is not leased.
+  * (4) The `cancelLEase` is invoked for previously leased resource. This will invoke resource
```
mpilquist: s/cancelLEase/cancelLease/

core/shared/src/main/scala/fs2/internal/Resource.scala
```diff
@@ -0,0 +1,165 @@
...
+  *     finalization if resource was already acquired and released and there are no more outstanding leases.
+  *
+  * Resources may be leased to other scopes. in that case, each scope must lease resource with `lease` and
```
mpilquist: s/in/In/

core/shared/src/main/scala/fs2/internal/Resource.scala
```diff
@@ -0,0 +1,165 @@
...
+  * when scope is closed (or when the resource lease is no longer required) release the lease with `cancelLease`.
+  *
+  * Note thate every method which as part of its invocation may potentially call resource finalizer
```
mpilquist: s/thate/that/

core/shared/src/main/scala/fs2/internal/Resource.scala
```diff
@@ -0,0 +1,165 @@
...
+  * is returning `F[Either[Throwable, Unit]]` instead just simple F[Unit] to make sure the errors when releasing resource
+  * are properly handled.
+  *
+  *
+  */
+private[internal] sealed trait Resource[F[_]] {
+
+  /** if of the resource **/
```
mpilquist: s/if/id/ -- also, end the summary doc with a period to be consistent with other doc in project

core/shared/src/main/scala/fs2/internal/Resource.scala
```diff
@@ -0,0 +1,165 @@
...
+  def lease: F[Boolean]
+
+  /**
+    * Each leased resource must be canceled when no longer needed by scope that leased it.
+    *
+    * When the lease is returned, this may result in resource's finalizer to be invoked and
+    * thus this may result in finalizer to fail.
+    *
+    */
+  def cancelLease: F[Either[Throwable, Unit]]
```
mpilquist: This might not be a good idea but we could encode the lease state machine directly in the types by changing `lease` to be `def lease: F[Option[Lease[F]]]` where:
```scala
trait Lease[F[_]] {
  def cancel: F[Either[Throwable,Unit]]
}
```

core/shared/src/main/scala/fs2/internal/Scope.scala
```diff
@@ -0,0 +1,293 @@
...
+    F.map(state.modify2 { _.unregisterChild(id) }){ _ => () }
+
+
+  /** returns all resources of this scope **/
+  private def resources: F[Catenable[Resource[F]]] = {
+    F.map(state.get) { _.resources }
+  }
+
+  /**
+    * Traverses supplued catenable with `f` that may produce failure, and collects theses failures.
```
mpilquist: s/supplued/supplied/

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: ok makes sense will do

core/shared/src/main/scala/fs2/internal/Resource.scala
pchlupacek: perhaps you right just lease: F[Lease[F]]. Not sure if we really need that option here.

core/shared/src/main/scala/fs2/CompositeFailure.scala
```diff
@@ -0,0 +1,10 @@
+package fs2
+
+/**
+  * Certain implementation in fs2 may cause multiple failures from the different part of the code to be thrown and collected.
+  * In these scenarios, COmposite failure assures that user won't lose any intermediate failures.
+  */
+class CompositeFailure(
+  first: Throwable
```
mpilquist: Perhaps use `cats.data.NonEmptyList` here?

core/shared/src/main/scala/fs2/CompositeFailure.scala
```diff
@@ -0,0 +1,10 @@
...
+  , others: List[Throwable]
+) extends Throwable(s"Resources failed to finalize: [${first.getClass.getName}: ${first.getMessage}] + ${others.size} others", first)
```
mpilquist: Recommend reorganizing this message to the "+ N others" part of the string shows up in the prefix and is hence unambiguous as to whether it is part of the first.getMessage. Something like: `if (exceptions.tail.headOption.isEmpty) exceptions.head.getMessage else s"Multiple exceptions were thrown (${exceptions.size}), first: ${exceptions.head.getMessage}"`

core/shared/src/main/scala/fs2/StreamScope.scala
```diff
@@ -0,0 +1,49 @@
+package fs2
+
+
+/**
+  * Scope represents a controlled block of execution of the stream to track resources acquired and released during
+  * the interpretation of the Stream.
+  *
+  * Scope's methods are used to perform low-level actions on stream interpretation, such as leasing the resources.
+  */
+trait StreamScope[F[_]] {
```
mpilquist: I'd prefer to call this `Scope` (and the one in `fs2.internal` could be called `ScopeImpl` or `DefaultScope` or something). Keep the uglier name for the internals and the nicer name for the API.

core/shared/src/main/scala/fs2/StreamScope.scala
```diff
@@ -0,0 +1,49 @@
...
+  def lease: F[Option[Lease[F]]]
+
+
+}
+
+
+/**
+  * Wraps leased resources from the scope of the other Stream.
+  */
+trait Lease[F[_]] {
```
mpilquist: Thoughts on this being in `Scope` companion, to keep top-level package more concise?

```scala
object Scope {
  trait Lease[F[_]] { ... }
}
```

core/shared/src/main/scala/fs2/CompositeFailure.scala
pchlupacek: ok

core/shared/src/main/scala/fs2/CompositeFailure.scala
pchlupacek: ok

core/shared/src/main/scala/fs2/StreamScope.scala
pchlupacek: Ok so how about Scope as public and ScopeRunFold in implementation?

core/shared/src/main/scala/fs2/StreamScope.scala
pchlupacek: Makes sense

PR #978, opened 2017-11-15
Iterator Handling in IO Package

### LINKED ISSUE #971: Iterator Integration
`iterator` and `iteratorR` were provided in scalaz-stream and I have need to write the interaction with iterators and wonder whether there was a reason this was removed. I'd be pleased to write the PR if you are interested and you can tell me where you would like its home to be.

[scalaz-stream code](https://github.com/scalaz/scalaz-stream/blob/series/0.8/src/main/scala/scalaz/stream/io.scala#L154-L185)

ChristopherDavenport:
Fixes #971 

If you are interested in this being implemented. It used to be present in scalaz-stream, and is a useful piece of functionality. If there is a problem in this implementation I would like to know so I can avoid it.

### DIFF
--- a/io/src/main/scala/fs2/io/io.scala
+++ b/io/src/main/scala/fs2/io/io.scala
@@ -130,7 +130,10 @@ package object io {
     */
   def readIterator[F[_], A](iterator: Iterator[A])(implicit F: Effect[F]): Stream[F, A] = {
     def getNext(i: Iterator[A]): F[Option[(A, Iterator[A])]] =
-      F.delay(i.hasNext).flatMap(b => if (b) F.delay(i.next()).map(a => (a, i).some) else F.pure(None))
+      F.delay(i.hasNext).flatMap(hasNext =>
+        if (hasNext) F.delay(i.next()).map(a => (a, i).some)
+        else F.pure(None)
+      )
 
     Stream.unfoldEval(iterator)(getNext)
   }
--- a/io/src/main/scala/fs2/io/io.scala
+++ b/io/src/main/scala/fs2/io/io.scala
@@ -124,4 +124,27 @@ package object io {
     */
   def toInputStream[F[_]](implicit F: Effect[F], ec: ExecutionContext): Pipe[F,Byte,InputStream] =
     JavaInputOutputStream.toInputStream
+
+  /**
+    * Reads all the values of the iterator into a Stream.
+    */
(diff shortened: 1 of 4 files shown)


io/src/main/scala/fs2/io/io.scala
```diff
@@ -124,4 +124,30 @@ package object io {
     */
   def toInputStream[F[_]](implicit F: Effect[F], ec: ExecutionContext): Pipe[F,Byte,InputStream] =
     JavaInputOutputStream.toInputStream
+
+  /**
+    * Reads all the values of the iterator into a Stream.
+    */
+  def readIterator[F[_], A](iterator: Iterator[A])(implicit F: Effect[F]): Stream[F, A] = {
```
mpilquist: I think we can move this method to `Stream` companion as `Stream.fromIterator` -- the effect constraint can be loosened to `Sync`.

io/src/main/scala/fs2/io/io.scala
```diff
@@ -124,4 +124,30 @@ package object io {
...
+        else F.pure(None)
+      )
+
+    Stream.unfoldEval(iterator)(getNext)
+  }
+
+  /**
+    * Brackets Resource Acquisition for an Iterator
+    */
+  def readBracketedIterator[F[_]: Effect, R, O](acquire: F[R])
```
mpilquist: I'm not convinced we need this method. If we have `Stream.fromIterator`, I think putting that together with `Stream.bracket` is simple enough.

io/src/main/scala/fs2/io/io.scala
SystemFw: You can make this a bit more concise with:
```scala
  def readIterator2[F[_], A](iterator: Iterator[A])(implicit F: Sync[F]): Stream[F, A] =
    Stream.unfoldEval(iterator)(i => F.delay(i.hasNext.guard[Option].as(i.next -> i)))
```

it's a style choice though, so up to you

io/src/main/scala/fs2/io/io.scala
SystemFw: I'm also a bit unconvinced we need this, especially because the user still has to do some massaging to get an `R => F[Iterator[O]]` (the wrapping in `F`), at which point one could use `bracket` directly

PR #979, opened 2017-11-18
Performance improvements in Chunk#toArray

### LINKED ISSUE #977: Streaming of very large documents is (relatively) slow
I originally raised this against http4s, but on further investigation narrowed it down to fs2.  The original issue is https://github.com/http4s/http4s/issues/1555#issuecomment-344463931.

Note the issue I'm presenting here is a bit contrived as I've narrowed it down to the smallest possible piece of code which highlights the issue.  My actual application is streaming results from the DB.

I've got the following code to copy a large (2GB) file.
```
 def copyFile(): Unit = {
    val file = new File(s"largefile.txt")

    val process = fs2.io.file.readAll[IO](file.toPath, DefaultBufferSize)
      .through(fs2.io.file.writeAll(Paths.get("copy.txt")))
      .run

    process.unsafeRunSync()
    println(s"Took ${end - start}ms")
  }
```

This code takes 45 seconds to run (macbook pro) and is rather CPU intensive for an IO activity.

If it helps at all, jvisualvm sampler snapshot looks like:
image

As a comparison, the equivalent commons-io example completes in 5 seconds.
```
import org.apache.commons.io.IOUtils;

import java.io.File;
import java.io.FileInputStream;
import java.io.FileOutputStream;

public class Test {
    public static void main(String[] args) throws Exception {
        File file = new File("/Users/luke.stephenson/projects/listings/listing-publisher-api/results4.txt");
        File out = new File("/Users/luke.stephenson/projects/listings/listing-publisher-api/results3.txt");

        long start = System.currentTimeMillis();
        IOUtils.copyLarge(new FileInputStream(file), new FileOutputStream(out));
        long end = System.currentTimeMillis();
        System.out.println("Took " + (end - start) + "ms");
    }
}
```

Curious to hear thoughts on this. Is it a bad idea to be working with large response streams with fs2?

Thanks!

mpilquist:
Fixes #977.

Includes a series of optimizations:
 - `Chunk#toArray` now simply uses a while loop instead of using the segment machinery to access each element (not sure why it ever did what it was doing...)
 - `toArray` is overridden on each array backed chunk subtype using `.slice` operation
 - `toByteBuffer` was added to `Chunk.Bytes` allowing a zero-copy view of the underlying array

@fiadliel could you take a look at this?

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -2,6 +2,7 @@ package fs2
 
 import cats.Eval
 import scala.reflect.ClassTag
+import java.nio.ByteBuffer
 
 /**
  * Segment with a known size and that allows index-based random access of elements.
@@ -65,7 +66,7 @@ abstract class Chunk[+O] extends Segment[O,Unit] { self =>
   override def toArray[O2 >: O: ClassTag]: Array[O2] = {
     val arr = new Array[O2](size)
     var i = 0
-    this.map { b => arr(i) = b; i += 1 }.drain.run
+    while (i < size) { arr(i) = apply(i); i += 1 }
     arr
   }
 
@@ -284,6 +285,7 @@ object Chunk {
     protected def splitAtChunk_(n: Int): (Chunk[O], Chunk[O]) =
       Boxed(values, offset, n) -> Boxed(values, offset + n, length - n)
     protected def mapStrict[O2](f: O => O2): Chunk[O2] = seq(values.map(f))
+    override def toArray[O2 >: O: ClassTag]: Array[O2] = values.slice(offset, offset + length).asInstanceOf[Array[O2]]
   }
   object Boxed { def apply[O](values: Array[O]): Boxed[O] = Boxed(values, 0, values.length) }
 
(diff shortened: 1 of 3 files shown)


fiadliel:
The specialized `toArray` was the same as what I had, and other changes look great! Throughput for the *sync* NIO calls is now amongst the fastest options; *async* is slower, but I think reasonably acceptable:

```
[info] Benchmark                                              Mode  Cnt  Score   Error  Units
[info] LargeStreamBenchmark.commonsIo                         avgt   10  3.467 ± 0.276   s/op
[info] LargeStreamBenchmark.fs2IoStream                       avgt   10  6.050 ± 0.268   s/op
[info] LargeStreamBenchmark.fs2NioAsyncStream                 avgt   10  7.764 ± 0.237   s/op
[info] LargeStreamBenchmark.fs2NioStream                      avgt   10  2.969 ± 0.102   s/op
[info] LargeStreamBenchmark.monixObservable                   avgt   10  4.307 ± 0.161   s/op
[info] LargeStreamBenchmark.monixObservableWithAsyncBoundary  avgt   10  2.833 ± 0.153   s/op
```

mpilquist:
Awesome, thanks for benchmarking!

PR #986, opened 2017-11-23
Added fs2.async.SyncRef

mpilquist:
...a lightweight alternative to Ref backed by an AtomicReference.

### DIFF
--- a/core/shared/src/main/scala/fs2/async/Ref.scala
+++ b/core/shared/src/main/scala/fs2/async/Ref.scala
@@ -3,8 +3,6 @@ package fs2.async
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 
-
-import cats.Eq
 import cats.implicits.{ catsSyntaxEither => _, _ }
 import cats.effect.{ Effect, IO }
 
@@ -15,7 +13,7 @@ import java.util.concurrent.atomic.{AtomicBoolean,AtomicReference}
 import Ref._
 
 /** An asynchronous, concurrent mutable reference. */
-final class Ref[F[_],A] private[fs2] (implicit F: Effect[F], ec: ExecutionContext) { self =>
+final class Ref[F[_],A] private[fs2] (implicit F: Effect[F], ec: ExecutionContext) extends RefOps[F,A] { self =>
 
   private var result: Either[Throwable,A] = null
   // any waiting calls to `access` before first `set`
@@ -23,7 +21,7 @@ final class Ref[F[_],A] private[fs2] (implicit F: Effect[F], ec: ExecutionContex
   // id which increases with each `set` or successful `modify`
   private var nonce: Long = 0
 
-  private val actor: Actor[Msg[A]] = Actor.actor[Msg[A]] {
+  private val actor: Actor[Msg[A]] = Actor[Msg[A]] {
     case Msg.Read(cb, idf) =>
       if (result eq null) {
         waiting = waiting.updated(idf, cb)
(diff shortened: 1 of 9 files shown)


core/shared/src/main/scala/fs2/async/async.scala
```diff
@@ -94,7 +96,7 @@ package object async {
    * bound. The inner `F[A]` will block until the result is available.
    */
   def start[F[_], A](f: F[A])(implicit F: Effect[F], ec: ExecutionContext): F[F[A]] =
-    ref[F, A].flatMap { ref => ref.setAsync(F.shift(ec) *> f).as(ref.get) }
```
mpilquist: This shift was redundant with the one in `setAsync`

core/shared/src/main/scala/fs2/async/RefOps.scala
```diff
@@ -0,0 +1,49 @@
+package fs2.async
+
+/**
+ * Operations on concurrent, mutable reference types.
+ */
+trait RefOps[F[_], A] {
```
mpilquist: Note there's no `setAsync` operation, as that's not possible to implement on `SyncRef`

core/shared/src/main/scala/fs2/async/Ref.scala
```diff
@@ -204,7 +186,7 @@ object Ref {
 
   /** Creates an asynchronous, concurrent mutable reference, initialized to `a`. */
   def initialized[F[_]: Effect, A](a: A)(implicit ec: ExecutionContext): F[Ref[F,A]] =
-    uninitialized[F, A].flatMap(r => r.setAsyncPure(a).as(r))
```
mpilquist: This was a bug or performance degradation depending on perspective

SystemFw:
It mostly looks good to me. I have one question regarding the semantics though: as it stands, `Ref` and `SyncRef` have different semantics wrt to errors. I.e. on a Ref, `set(F.fail(ex))` will succeed, and the subsequent `get` will fail, whereas `set(F.fail(ex))` will fail on a SyncRef.
In other words, a SyncRef is a mutable reference for As, whereas a Ref is a mutable reference for `Throwable or A` (which puzzled me at first).
Is this intended/desirable?

SystemFw:
I'd go with option 1.

(I guess my preference would be to have Ref to also not have that behaviour, but that's a deeper change that's best left out of this PR)

mpilquist:
Hm, what about leaving `setSync` on `RefOps` but changing its behavior such that a failure in `fa` fails result instead of setting a failure in to the ref? Then the only way to cause `get` to throw would be a failure in `fa` passed to `setAsync`, which is understandable since the task is forked before `fa` is run.

mpilquist:
OK, fixed in https://github.com/functional-streams-for-scala/fs2/pull/986/commits/f05db8818eb85bace11f6c9554034140c95df618

SystemFw:
Well, I thought the whole point of having `setSync` as a separate operation is that it behaves differently than flatMap then set.

(Otoh, I have tried not too hard once to remove the Throwable from Ref altogether)

mpilquist:
Alright, that's convincing. Just pushed a fix.

SystemFw:
:+1: for merging this

If you want to have a look at how Ref would look like with no throwable (in general, not for the sake of this PR), I have a (very probably buggy) version [here](https://github.com/SystemFw/fs2/blob/exp/failed/ref-no-throwable/core/shared/src/main/scala/fs2/async/Ref.scala)

I tacked this while trying to add `modifyF` (which I think can be done with a new message in Actor, or modifying the TrySet message)

Issue #888, opened 2017-06-28
RFE: Ref/Signal modifyF


djspiewak:
```scala
def modifyF[B](f: A => F[(A, B)]): F[(Change[A], B)]
```

This would be useful on both `Ref` and on `mutable.Signal`.  It's similar in principle to `modify2`, but allows an effect context to be threaded through.  This makes it possible to evaluate `StateT` through a `Signal` (with retry), which is sometimes very nice.

Also, this is somewhat trivial to implement.  Simply `flatMap` the results of the `f` inside `tryModify2`.

mpilquist:
No objections here. PR?

pchlupacek:
@djspiewak  I assume this shall acquire lock, eval `F` and release lock in case `F` evaluates or fails right? That would be welcome. In fact I would sugest to have these two : 
```scala 
def modifyF[B](f: A => F[A]]): F[Change[A]] = ???
def modify2F[B](f: A => F[(A, B)]): F[(Change[A], B)]
``` 
to have them consistent with current `modify` naming

wedens:
In haskell effectful modify for `MVar` is implemented like this (with removed irrelevant to scala things):
```haskell
modifyMVar :: MVar a -> (a -> IO (a,b)) -> IO b
modifyMVar m io = do
    a <- takeMVar m
    (a', b) <- io a `onException` putMVar m a
    putMVar m a'
    return b
```
The missing thing in `Ref` is something like `takeMVar` that works like `get` but also unsets `Ref`. 

`scalaz.concurrent.MVar` does it similarly to haskell `MVar`.

SystemFw:
So, I've spent some time deep diving into the internals of fs2 async, and I'm convinced that the fact a `Ref` can't be unset is a feature. 
The fact that most compound operations on `MVar` involve `takeMVar` is a weak point, since it can cause other threads to needlessly block: in fact, `readMVar` is no longer implemented as "take then put" for exactly this reason, and it now appears to be a native operation.
Conversely, it's easy to implement something like `takeMVar` in terms of `Ref`: just have a `Ref[F, (Option[A], Vector[Ref[F,A]])]`, where the `Vector` holds uninitialised `Ref`s for the readers which are queueing on `None` (similarly to what it's done in `Queue`).

Now, using `Ref`s `modify` semantics to implement `modifyF` does raise some interesting questions about the semantics of `modifyF`, especially with regards to how many times the effect is evaluated.
[...]

PR #989, opened 2017-11-23
Removes need for UnconsAsync in Algebra

pchlupacek:
+ Removes need of F:Effect / Sync in translate
needs #974 to get in first


pchlupacek:
@mpilquist this seems now done. Please review, if possible. Also before merging we have to squash it :-) 
Thank you.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2496,8 +2496,23 @@ object Stream {
      * For example, `merge` is implemented by calling `unconsAsync` on each stream, racing the
      * resultant `AsyncPull`s, emitting winner of the race, and then repeating.
      */
-    def unconsAsync(implicit ec: ExecutionContext, F: Effect[F]): Pull[F,Nothing,AsyncPull[F,Option[(Segment[O,Unit], Stream[F,O])]]] =
-      Pull.fromFreeC(Algebra.unconsAsync(self.get, ec)).map(_.map(_.map { case (hd, tl) => (hd, Stream.fromFreeC(tl)) }))
+    def unconsAsync(implicit ec: ExecutionContext, F: Effect[F]): Pull[F,Nothing, AsyncPull[F,Option[(Segment[O,Unit], Stream[F,O])]]] = {
+      type UO = Option[(Segment[O,Unit], FreeC[Algebra[F,O,?],Unit])]
+
+      Pull.fromFreeC {
+        val ref = new async.Ref[F, Either[Throwable, Option[(Segment[O,Unit], Stream[F,O])]]]
```
SystemFw: I would think this needs to be suspended in `F`, maybe it's ok since it's encapsulated?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2496,8 +2496,23 @@ object Stream {
...
+        Algebra.getScope[F, Nothing] flatMap { scope =>
+          val runStep =
+            Algebra.runFoldScope(
+              scope
+              , Algebra.uncons(self.get).flatMap(Algebra.output1(_))
+              , None : UO
+            ){ (_, uo) => uo.asInstanceOf[UO] } map { _ map { case (hd, tl) => (hd, fromFreeC(tl)) }}
+
+          Algebra.eval(ref.setAsync(F.attempt(runStep))) map { _ => AsyncPull.readAttemptRef(ref) }
```
SystemFw: As per the current implementation of `Ref`, `attempt(runStep)` is redundant, and `readAttemptRef` can take a `Ref[F, A]` rather than a `Ref[Throwable, A]` (with some implementation changes). However, I much prefer to be explicit just like you were: I think we should remove the internal throwable from Ref, and use a `Ref[F, Either[Throwable, A]` explicitly when we want to carry an exception within it.

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: yes, the reason is to optimize as much as we can, hence this is evaluated at every async pull.

SystemFw: :+1:

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: I think `setAsync` has weird semantics. Why does it fork the `F[A]` evaluation and why can't the caller do that instead? If the caller did the fork, then we'd be able to implement `setAsync` in a way that doesn't fail.

mpilquist:
This is awesome by the way!

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: I was sort of puzzled by this as well. I guess we have to trust that there were enough discussions on the Cats.Effects for this. Didn't follow all that recently, so I really don't have idea why that changed from the way how we had implemented it in 0.9.

pchlupacek:
thanks @mpilquist

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: @SystemFw lets revisit Ref implementation later. I think perhaps we can do a better with Ref. Also perhaps we can work with folks in CatsEffects to bring something that will be stronger.

mpilquist: Unrelated to cats-effect stuff. You can see the old 0.9 stuff worked the same way: https://github.com/functional-streams-for-scala/fs2/blob/108f94a0a4a4e21bcc99a900db28e1ca54439570/core/shared/src/main/scala/fs2/Task.scala#L342-L343

pchlupacek: Ah @mpilquist I see what do you mean. Perhaps we shall give it a try ?

pchlupacek: hm, cannot we have all the set's in fact synchronous? I am curios if we need in fact actor for defining the Ref?

mpilquist: Sure, I'm all for it assuming it doesn't break too much stuff. @SystemFw do you want to give this a shot?

SystemFw: I've been thinking a lot about `Ref`, and I think it can be improved (or at least it's worth a try), in a separate PR (I also want to tackle #888 at some point, with a new Actor message).

But yeah, it's a separate discussion. Btw, I second the awesomeness of removing an essential (I thought) algebra primitive

SystemFw:
I've been thinking a lot about Ref, and I think it can be improved (or at least it's worth a try), in a separate PR (I also want to tackle #888 at some point, with a new Actor message).

But yeah, it's a separate discussion. Btw, I second the awesomeness of removing an essential (I thought) algebra primitive

pchlupacek:
@SystemFw I think, (perhaps I am wrong), that we could implement Ref W/O actor, and perhaps only we can use the Actor`s queue to signal the listeners. I think generally that would mean a reasonable performance improvement if we would be able to do so ....

SystemFw:
I've haven't thought about removing Actor yet. My lines of thought were:
- do we need `setAsync`?
- do we need a `Throwable Or A` as the state of a Ref, or is just A enough?
- should we support a `modifyF` (I think we should)

pchlupacek:
@SystemFw I think if we implement it w/o Actor, we can perhaps get [Throable,A] => A only
also we may need then only `Sync` instance to build it and Async/Effect may be needed only for defining the `get` and cancellableGet. I am not 100% sure if it can be implemented, but I think so.

SystemFw:
I think whether or not we use `Throwable` depends on what async behaviour we want (I'll expand in a second). I also think it can be implemented without Actor, but not sure it's wise to do so: after all, using Actor we can abstract away most of the lockfree algorithm, which would otherwise have to be written by hand (and they are error prone).

As for the Throwable, `setSync` had the weird behaviour of succeeding on a failed `fa`, with the subsequent `get` failing (fixed now). `setAsync` has the same behaviour, but it makes more sense. Now we could decide to not have `setAsync`, since afterall is a leftover of the pre cats-effect days, where `shift` wasn't available, and could now be `shift *> setSync`. I like that.

I do however, have a counter point. With `setSync`, we can return the `fa` error to the caller, so the new behaviour is fine. With any form of `setAsync` (either primitive or in user land with shift), we'd have to discard the error. If we don't have a Throwable in the Ref, it means that if `fa` in `setAsync(fa)` fails, the `Ref` is not set and the error discarded, and `get` will deadlock.

Thoughts?

pchlupacek:
@SystemFw I still doubt the setting of ref should be able to fail. The reason for all the Throwable stuff comes imho form the way how the Ref is implemented. 

I may be wrong, but the performnace of Actor and non-actor based reference may be a winner, since we use ref a lot. 

Also, surprisingly the complexity of the Actor-free implementation may be not that bad. But lets start somewhere and we will see :-)

mpilquist:
We could change `setAsync(fa)` to `fa.flatMap(setAsyncPure)` -- the evaluation up to `a` would happen synchronously but the update to internal state would happen asynchronously. If a caller wants the original behavior, they could do `unsafeRunAsync(ref.set(fa))(res => /* handle error here*/ )`.

SystemFw:
Yeah generally speaking I think the old behaviour would simply require `Ref[F, Either[Throwable, A]` explicitly, so removing `setAsync` would make sense.

Another thing though, what about `modifyF`? The only way (I could think) of implementing it would be to send a closure (or IO) to the Actor, so that the effect is only executed when a concurrent race is won. I _think_ this can still be done without having a Throwable in the Ref (might require an AtomicReference in the body of `modifyF`), but I thought I'd mention it since allowing the `Ref` to effectfully modify its contents  is a significant conceptual change.

SystemFw:
btw, should we open an issue for this rather than continue the discussion in this PR?

mpilquist:
@SystemFw Yeah sounds good

SystemFw:
@mpilquist will do

PR #992, opened 2017-11-23
Fixed memory leak in readAllFromFileHandle due to use of for/yield

### LINKED ISSUE #987: Memory leak when reading files with a small buffer size
This is running against `"co.fs2" %% "fs2-io" % "0.10.0-SNAPSHOT"` containing the fix to #977 

I noticed a memory leak when using http4s which I raised https://github.com/http4s/http4s/issues/1561.  This may be related, or may not.  I was trying to track down a minimal test to reproduce and ended up here.

I ended up using the same code as what I was using for the file copy performance test before, but changed the read buffer size to 1.

https://github.com/lukestephenson/stream-benchmarks/blob/master/src/main/scala/benchmarks/CompressTest.scala
```
package benchmarks

import java.io.File
import java.nio.file.Paths

import cats.effect.IO
/**
 * results4.txt is a 4GB file.
 * run this with -Xmx56m
 **/
object CompressTest {
  val DefaultBufferSize = 1
  val src = "/Users/luke.stephenson/projects/listings/results4.txt"
  val target = "/Users/luke.stephenson/projects/listings/results3.txt"

    val file = new File(src)

    def process = fs2.io.file.readAll[IO](file.toPath, DefaultBufferSize)
      .through(fs2.io.file.writeAll(Paths.get(target)))
      .run

    process.unsafeRunSync()

    println(s"took ${end - start}ms")
  }
}
```
When running with a large buffer size, the code runs fine.  However when running with a small buffer  size it runs out of memory.  I'd expect it to be slower, but not impact memory usage.

The heap dump shows it holding onto a lot of `FreeC` instances.

mpilquist:
Fixes #987.

### DIFF
--- a/io/src/main/scala/fs2/io/file/FileHandle.scala
+++ b/io/src/main/scala/fs2/io/file/FileHandle.scala
@@ -111,9 +111,10 @@ private[file] object FileHandle {
         asyncCompletionHandler[F, Lock](f => chan.lock(position, size, shared, null, f))
 
       override def read(numBytes: Int, offset: Long): F[Option[Chunk[Byte]]] = {
-        val buf = ByteBuffer.allocate(numBytes)
-        asyncCompletionHandler[F, Integer](f => chan.read(buf, offset, null, f)).map { len =>
-          if (len < 0) None else if (len == 0) Some(Chunk.empty) else Some(Chunk.bytes(buf.array, 0, len))
+        F.delay(ByteBuffer.allocate(numBytes)).flatMap { buf =>
+          asyncCompletionHandler[F, Integer](f => chan.read(buf, offset, null, f)).map { len =>
+            if (len < 0) None else if (len == 0) Some(Chunk.empty) else Some(Chunk.bytes(buf.array, 0, len))
+          }
         }
       }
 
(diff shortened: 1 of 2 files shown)


io/src/main/scala/fs2/io/file/FileHandle.scala
```diff
@@ -111,9 +111,10 @@ private[file] object FileHandle {
         asyncCompletionHandler[F, Lock](f => chan.lock(position, size, shared, null, f))
 
       override def read(numBytes: Int, offset: Long): F[Option[Chunk[Byte]]] = {
-        val buf = ByteBuffer.allocate(numBytes)
```
mpilquist: Wrapped these buffer allocations in `F.delay` so the returned action is safe to use more than once. Unrelated to memory leak.

SystemFw:
wow nice catch

io/src/main/scala/fs2/io/file/pulls.scala
```diff
@@ -19,11 +19,11 @@ object pulls {
   def readAllFromFileHandle[F[_]](chunkSize: Int)(h: FileHandle[F]): Pull[F, Byte, Unit] =
     _readAllFromFileHandle0(chunkSize, 0)(h)
 
-  private def _readAllFromFileHandle0[F[_]](chunkSize: Int, offset: Long)(h: FileHandle[F]): Pull[F, Byte, Unit] = for {
-    res <- Pull.eval(h.read(chunkSize, offset))
-    next <- res.fold[Pull[F, Byte, Unit]](Pull.done)(o => Pull.output(o) *> _readAllFromFileHandle0(chunkSize, offset + o.size)(h))
-  } yield next
```
mpilquist: This `yield next` was the cause of the leak -- each time through this pull, we'd push a `map(identity)` in to the `FreeC`, which would accumulate on the return. In essence, the `yield` results in the recursive call being out of tail position.

lukestephenson: Thanks for the explanation.  It took me a while to realise what was going on here and I would have really struggled to find this.

PR #994, opened 2017-11-25
Add link to Actor docs

SystemFw:
This would have saved me  a  _ton_ of time when I was first learning about Actor, so I think it's worth linking.

I do have a question though: the article explains how the state of the queue is correctly shared in the presence of concurrency, but what about the state that the handler mutates?

In other words, let's say two calls to `act` happen on 2 different threads, and the handler is mutating a variable `myVar` , how will the update to `myVar` from one call be propagated to the other thread? (also assume that any call to `!` is happening on another thread still).  `act` doesn't write to any `volatile` (or Atomic*) field after running `handler`, so I don't see where the memory barrier is.

There's a comment saying:
"Memory consistency guarantee: when each message is processed by the `handler`, any memory that it mutates is guaranteed to be visible by the `handler` when it processes the next message, even if the execution context runs the invocations of `handler` on separate threads. This is achieved because the `Actor` reads a volatile memory location before entering its event loop, and writes to the same location before suspending."

but that doesn't seem to reflect what the code is doing (I think it refers to an older version of actor with an AtomicInteger called `suspended`). What am I missing?

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Actor.scala
+++ b/core/shared/src/main/scala/fs2/internal/Actor.scala
@@ -28,6 +28,8 @@ import java.util.concurrent.atomic.AtomicReference
  * Implementation based on non-intrusive MPSC node-based queue, described by Dmitriy Vyukov:
  * [[http://www.1024cores.net/home/lock-free-algorithms/queues/non-intrusive-mpsc-node-based-queue]]
  *
+ * A detailed analysis of the behaviour of Actor is available at [[https://gist.github.com/djspiewak/671deab9d7ea027cdc42]]
+ * 
  * @param handler  The message handler
  * @param onError  Exception handler, called if the message handler throws any `Throwable`.
  * @param ec       Execution context


mpilquist:
@SystemFw Indeed, the comment is out of date. The change which removed `suspended` is here: https://github.com/scalaz/scalaz/commit/dce8a804cfb79e5904cace2e6655dbc933c1d756#diff-7cf066c680ea8b7d1d83dfa1f1d86f74 with PR discussion here: https://github.com/scalaz/scalaz/pull/624

The only place I'm not sure of w.r.t. memory barriers is here: https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/internal/Actor.scala#L56 Though perhaps the barrier is provided by the submission to the execution context?

SystemFw:
@mpilquist We went to the same thought process, but:

A read from a volatile shouldn't mean that a memory barrier is introduced afaik, it's only when one threads writes to a volatile var, and another reads from the same var (again, afaik).

I also thought about the submission to the ec, but then EC is an trait with no implementation, so it seems risky to assume that this holds.

I had seen that PR, but it doesn't seem to discuss it, and at this point curiosity is killing me tbh, do you know who might know? Daniel perhaps?

mpilquist:
@plokhotnyuk or @djspiewak 

I'm more convinced now that this implementation is relying on the EC to providing the barrier for state mutated by the handler function and `head.get/compareAndSet` for barrier for `!` vs handler.

SystemFw:
I do agree that EC submission is the most likely point for the state memory barrier. `!` vs handler is clear. I think this is never a problem in practice given ECs are backed by java thread pools (which I assume insert the barrier), but strictly speaking I can break it by providing a custom EC (which is just a trait) that doesn't.

Anyway let's see if we get a definitive response and I'll update the outdated comment :)

SystemFw:
@mpilquist Found it

> Memory consistency effects: Actions in a thread prior to the submission of a Runnable or Callable task to an ExecutorService happen-before any actions taken by that task, which in turn happen-before the result is retrieved via Future.get()

It's for ExecutorServices. This is actually unsafe both in scalaz (a Strategy need not be an ExecutorService), and here (ECs need not be ExecutorServices either), but I don't suppose it matters in practice. Anyway mystery solved.

durban:
I'd argue that the same guarantee should be provided by any (useful) implementation of `ExecutionContext` even if it is not spelled out in the API docs. (If not, that would allow things like a `Runnable` not seeing its own fields. That would be ... broken.)

SystemFw:
@durban https://twitter.com/mpilquist/status/934450193403142146

All is good it seems :)

SystemFw:
I'll make it more clear in the scaladoc soon

plokhotnyuk:
You are right. Not any instance of Strategy is suitable for Actor.
It would be better to have separated ActorStrategy type.
Here is implementation of most efficient one for current Actor interfaces: https://github.com/plokhotnyuk/actors/blob/master/src/test/scala/com/github/plokhotnyuk/actors/ScalazActorSpec.scala#L11

plokhotnyuk:
In the following fork of Scalaz I have tried to solve this and other problems of Scalaz actors and strategies for them https://github.com/plokhotnyuk/scalaz/commits/series/7.2.x

PR #996, opened 2017-11-25
Queue#peek1 and timedPeek1

### LINKED ISSUE #983: Queue#peek
I think a `Queue#peek` method would be useful with the following signature:
```scala
def peek: F[A]
```
Its semantics would be like `dequeue1`, but without removing the element. It could have a `timedPeek` variant too.

If it's determined, that this indeed would be useful, I'm willing to work on a PR.

durban:
`Queue#peek1` and `timedPeek1` (fixes #983).

I'm unsure about the semantics for `synchronous` (and consequently for `synchronousNoneTerminated`). If I understand correctly, that queue is (conceptually) always empty. That would suggest, that `peek1` should never finish. However, it is in fact possible (sometimes) to dequeue an element from such a queue. Thus, peek could also work. This seems more useful, so this is what this PR currently implements.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -16,3 +16,4 @@ Also see the [GitHub contributor stats](https://github.com/functional-streams-fo
 - Rúnar Ó. Bjarnason ([@runarorama](https://github.com/runarorama)): Various processes and combinators, circular buffers, compression. I'm an ideas man, Michael.
 - Jed Wesley-Smith ([@jedws](https://github.com/jedws)): really very minor tweaks, cleanups and pestering, hardly worth the mention
 - Michael Pilquist ([@mpilquist](https://github.com/mpilquist)): 0.9 redesign work, maintenance
+- Daniel Urban ([@durban](https://github.com/durban)): queue peek implementation
--- a/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
+++ b/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
@@ -159,6 +159,7 @@ object Queue {
       * Internal state of the queue
       * @param queue    Queue, expressed as vector for fast cons/uncons from head/tail
       * @param deq      A list of waiting dequeuers, added to when queue is empty
+      * @param peek     The waiting peekers (if any), created when queue is empty
       */
     final case class State(
       queue: Vector[A],
@@ -202,10 +203,7 @@ object Queue {
 
         def dequeue1: F[A] = cancellableDequeue1.flatMap { _._1 }
 
-        def peek1: F[A] = peek1Impl.flatMap {
-          case Left(ref) => ref.get
-          case Right(a) => F.pure(a)
-        }
+        def peek1: F[A] = peek1Impl.flatMap(_.fold(_.get, F.pure))
 
         def timedDequeue1(timeout: FiniteDuration, scheduler: Scheduler): F[Option[A]] =
           timedDequeueBatch1(1, timeout, scheduler).map(_.map(_.strict.head))
@@ -218,16 +216,13 @@ object Queue {
         }
 
         private def peek1Impl: F[Either[Ref[F, A], A]] = ref[F, A].flatMap { ref =>
-          qref.modify2 { state =>
-            if (state.queue.isEmpty) {
-              state.peek match {
-                case None => (state.copy(peek = Some(ref)), Left(ref))
-                case Some(r) => (state, Left(r))
-              }
-            } else {
-              (state, Right(state.queue.head))
-            }
-          }.map(_._2)
+          qref.modify { state =>
+            if (state.queue.isEmpty && state.peek.isEmpty) state.copy(peek = Some(ref))
+            else state
+          }.map { change =>
+            if (change.previous.queue.isEmpty) Left(change.now.peek.get)
+            else Right(change.now.queue.head)
+          }
         }
 
         def cancellableDequeue1: F[(F[A],F[Unit])] =
--- a/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
+++ b/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
@@ -221,7 +221,7 @@ object Queue {
             else state
           }.map { change =>
             if (change.previous.queue.isEmpty) Left(change.now.peek.get)
-            else Right(change.now.queue.head)
+            else Right(change.previous.queue.head)
           }
         }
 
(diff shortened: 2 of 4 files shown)


SystemFw:
> I'm unsure about the semantics for synchronous (and consequently for synchronousNoneTerminated). If I understand correctly, that queue is (conceptually) always empty. 

Not really, if anything one might see it's a queue with always one element. A more apt description is that `synchronous` is a queue with _at most_ one element: i.e. a producer can't enqueue until the consumer has dequeued. `peek` makes perfect sense there so I assume the semantics you have now are correct (going to read the code in a minute)

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -144,10 +160,14 @@ object Queue {
       * @param queue    Queue, expressed as vector for fast cons/uncons from head/tail
       * @param deq      A list of waiting dequeuers, added to when queue is empty
       */
-    final case class State(queue: Vector[A], deq: Vector[Ref[F,Chunk[A]]])
+    final case class State(
+      queue: Vector[A],
+      deq: Vector[Ref[F,Chunk[A]]],
+      peek: Option[Ref[F, A]]
```
SystemFw: Adding a `@param` for this as well would be nice, for consistency

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -160,20 +180,56 @@ object Queue {
...
+              state.peek match {
+                case None => (state.copy(peek = Some(ref)), Left(ref))
+                case Some(r) => (state, Left(r))
+              }
+            } else {
+              (state, Right(state.queue.head))
+            }
+          }.map(_._2)
+        }
+
```
SystemFw: Generally speaking I like the approach here, not sure whether we should go the full route of having a `cancellablePeek1` like for `dequeue` or not

SystemFw:
Generally I like the approach, left a few comments. From an implementation point of view `peek1` on `synchronous` queue is ok. It's just unlikely to be very useful

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
durban: Ok.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -160,20 +180,56 @@ object Queue {
...
-            else // queue was empty, we had waiting dequeuers
-              c.previous.deq.head.setAsyncPure(Chunk.singleton(a)).as(true)
+            val dq = if (c.previous.deq.isEmpty) {
+              // we enqueued a value to the queue
+              signalSize(c.previous, c.now)
+            } else {
+              // queue was empty, we had waiting dequeuers
+              c.previous.deq.head.setAsyncPure(Chunk.singleton(a))
+            }
+            val pk = if (c.previous.peek.isEmpty) {
```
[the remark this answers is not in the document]
durban: I've used `isEmpty`/`get` to conform to the already existing code there (which uses `isEmpty`/`head`).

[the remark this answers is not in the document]
SystemFw: Fair enough :)

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -160,20 +180,56 @@ object Queue {
...
+              // notify peekers
+              c.previous.peek.get.setAsyncPure(a)
+            }
+            (dq *> pk).as(true)
           }
 
         def dequeue1: F[A] = cancellableDequeue1.flatMap { _._1 }
 
+        def peek1: F[A] = peek1Impl.flatMap {
+          case Left(ref) => ref.get
```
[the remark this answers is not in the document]
durban: Do you mean like `.flatMap(_.fold(_.get, F.pure))`? Pattern matching is [supposedly faster](https://github.com/typelevel/cats/issues/1951); but if `fold` is preferred in this codebase, I can change it.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
durban: I'll think about it.

durban:
I've pushed a commit addressing (hopefully) all comments, except the one about cancellable peek. I'll have to think about that one. (Actually I wish there was a general way of cancelling async stuff, not implementing separately one variant for every method...)

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -160,20 +181,50 @@ object Queue {
...
+          }
+        }
+
+        private def peek1Impl: F[Either[Ref[F, A], A]] = ref[F, A].flatMap { ref =>
+          qref.modify { state =>
+            if (state.queue.isEmpty && state.peek.isEmpty) state.copy(peek = Some(ref))
+            else state
+          }.map { change =>
+            if (change.previous.queue.isEmpty) Left(change.now.peek.get)
+            else Right(change.now.queue.head)
```
SystemFw: One super minor nitpick (sorry). `change.previous.queue.head` for symmetry with the `isEmpty` (since the queue hasn't changed between previous and now)

SystemFw:
Apart from one extra little comment, it's ready. Thank you very much for this :)
I'm thinking of merging as is, and reassess the cancellable thing later. I'd also prefer composable cancellation, but it's a problem we should address in cats-effect (we are talking about it).

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
durban: Fair enough.

CONTRIBUTORS.md
```diff
@@ -16,3 +16,4 @@ Also see the [GitHub contributor stats](https://github.com/functional-streams-fo
 - Rúnar Ó. Bjarnason ([@runarorama](https://github.com/runarorama)): Various processes and combinators, circular buffers, compression. I'm an ideas man, Michael.
 - Jed Wesley-Smith ([@jedws](https://github.com/jedws)): really very minor tweaks, cleanups and pestering, hardly worth the mention
 - Michael Pilquist ([@mpilquist](https://github.com/mpilquist)): 0.9 redesign work, maintenance
+- Daniel Urban ([@durban](https://github.com/durban)): queue peek implementation
```
mpilquist: Add to build.sbt too so it shows up in the generated POM.

durban: Ok, thanks.

mpilquist:
Looks great, thanks

CONTRIBUTORS.md
SystemFw: lol I need to remember to do that as well at some point

PR #999, opened 2017-11-27
Fixed memory leaks in compress.inflate/delate

### LINKED ISSUE #998: Memory leak in fs2.compress.deflate
There is a memory leak in fs2 when using `fs2.compress.deflate` with a large Stream.  Similar to #987, it results in a long chain of `FreeC` instances.

The code to reproduce is:
```
package benchmarks

import java.io.File
import java.util.zip.Deflater

import cats.effect.IO
import fs2.compress.deflate

object CompressTest {
  val DefaultBufferSize = 1
  val src = "/Users/luke.stephenson/projects/listings/results4.txt"
  val target = "/Users/luke.stephenson/projects/listings/results3.txt"

  val bufferSize: Int = 32
  val level: Int = Deflater.DEFAULT_COMPRESSION

    val file = new File(src)

    def process = fs2.io.file.readAll[IO](file.toPath, DefaultBufferSize)
        .through(deflate(
          level = level,
          nowrap = true,
          bufferSize = bufferSize
        ))
      .run

    process.unsafeRunSync()

    println(s"took ${end - start}ms")
  }

}
```

Be sure to run this against the latest version of fs2 containing PR #992. With the source being a very large file, the heap dump shows:
image

This is probably the root cause of a memory leak in the http4s GZip implementation. https://github.com/http4s/http4s/issues/1561

mpilquist:
I believe this fixes #998 but I haven't tried it. If it's not too much trouble, would you mind giving it a shot @lukestephenson and letting us know if it fixes the leak you saw?

### DIFF
--- a/core/jvm/src/main/scala/fs2/compress.scala
+++ b/core/jvm/src/main/scala/fs2/compress.scala
@@ -28,18 +28,21 @@ object compress {
       _deflate_stream(deflater, buffer)(in)
     }.stream
   }
-  private def _deflate_step[F[_]](deflater: Deflater, buffer: Array[Byte]): Option[(Chunk[Byte], Stream[F, Byte])] => Pull[F, Byte, Option[Stream[F, Byte]]] = {
-    case None => Pull.pure(None)
-    case Some((hd,tl)) =>
-      deflater.setInput(hd.toArray)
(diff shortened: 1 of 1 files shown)


mpilquist:
Thanks!

PR #1001, opened 2017-11-27
Renamed *> to >> b/c *> takes RHS as strict param whereas >> takes RHS as by-name, enabling recursion

mpilquist:


mpilquist:
Note cats 1.0 final will have a matching definition of `>>` (by-name, non-deprecated).

Issue #993, opened 2017-11-24
Improve Ref


SystemFw:
This can be fleshed out, but for now let's collate the ideas about things worth improving in Ref.

- Is Actor necessary or can we use something more lightweight? What's the actual overhead of Actor ? 

- The behaviour of `setSync(fa: F[A])` and `setAsync(fa: F[A]`) has been weird for a while. A failure in `fa` will make them succeed, and the subsequent `get` will fail. `setSync` could be easily fixed, and probably replaced by `fa.flatMap(setSyncPure(a))` altogether. Can we do the same with `setAsync`? Is having `setAsyncPure` enough (so that the `fa` will run synchronously, and the `setAsyncPure` asynchronously)

- Related to the above, if  actions in `set*` can't fail, do we need  `Either[Throwable, A]` as the state of the Ref? Shouldn't a reference to A be enough? Things like `start` could be recovered via an explicit `Ref[F, Either[Throwable, A]]`

- As a counterpoint to the above two points, what about `modifyF` (#888). To have the expected semantics (effect only run once, `Change` is faithful), the only way I could envision is sending a closure (or an IO) to the Actor, which is only run when it wins the contention race. Unclear at the moment if that invalidates the _actions that set the Ref shouldn't be able to fail_ hypothesis, and therefore cement the need for having an `Either[Throwable, A]` as the state of `Ref`.

- As a broader point, users now have `runAsync` and `shift` in their API, which they didn't before, which might influence how (or whether) we rethink Ref

If this sounds non-organic and confused, it's because it still is (in my head), please challenge/correct.

Relevant discussions:
#989 
#986
#888
#974 (initial push, not much in terms of discussion of Ref)

@mpilquist @pchlupacek

pchlupacek:
@SystemFw Just a small clarification #989 did not introduce the Ref, was already there :-) Also note that every nondet code in fs2 is using Ref, i.e. merge, join, essentially any Async operation at every single step.

SystemFw:
True, thanks. I'll just remove that point.

pchlupacek:
@SystemFw just maybe for some thoughts re performance: 
given this program https://gist.github.com/pchlupacek/d80a947ac4fdd6f72649a5ae57cced59
this are the results of sync/async comparison

```
Ops: 10000000
Execution of : SYNC: setSyncPure took 0.739 s
Execution of : SYNC: setSyncPure took 0.618 s
[...]

PR #1002, opened 2017-11-27
Remove Throwable from Ref

SystemFw:
As per the title, it partially addresses #993 .
 Note that the behaviour of all code that depends on Ref within fs2 has _not_ been changed, e.g. I haven't experimented with changing on what EC anything runs, or switching async to sync or anything like that.
This is a sheer refactor, that shouldn't affect the behaviour of the rest of fs2 at all. The semantics of Ref have slightly changed though (hopefully for the best), and the implementation should be a bit clearer without Throwable. 
The commits are self-contained if you prefer to look at them them individually.
Let me know what you think.

### DIFF
--- a/core/shared/src/main/scala/fs2/async/Ref.scala
+++ b/core/shared/src/main/scala/fs2/async/Ref.scala
@@ -141,6 +141,9 @@ final class Ref[F[_],A] private[fs2] (implicit F: Effect[F], ec: ExecutionContex
    * `set` to this `ref`. The loser continues running but its reference
    * to this ref is severed, allowing this ref to be garbage collected
    * if it is no longer referenced by anyone other than the loser.
+   *
+   * If the winner fails, the returned `F` fails as well, and this `ref`
+   * is not set.
    */
   def race(f1: F[A], f2: F[A]): F[Unit] = F.delay {
     val ref = new AtomicReference(actor)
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/fs2/async/Ref.scala
```diff
@@ -150,9 +153,13 @@ final class Ref[F[_],A] private[fs2] (implicit F: Effect[F], ec: ExecutionContex
       // or the actor directly, and the winner destroys any
       // references behind it!
       if (won.compareAndSet(false, true)) {
-        val actor = ref.get
-        ref.set(null)
-        actor ! Msg.Set(res, () => ())
+        res match {
+          case Left(e) => throw e
```
mpilquist: Should the `ref.set(null)` still happen in this case?

SystemFw: Good catch! I'll have to think about it

SystemFw: Probably yes, or the loser still has a reference, and won't be garbage collected when the winner fails

mpilquist:
LGTM. Not sure about the one TODO listed and the one comment I left. Otherwise looks ready to merge to me.

SystemFw:
Yeah, I wanted to check the general approach was OK. I think your comment is spot on, and I'll have to think a tad more about that TODO item, but I'll probably be able to finish today

SystemFw:
Wrt the Boolean in the callback to the Nevermind message to the actor, it's used to signal that the `id` of the listener you want to cancel doesn't correspond to any listener. However, the only place that sends a Nevermind message is `cancellableGet`, which however is sure that the `id` exists (since it's the one of the listener it just created), and therefore ignores the Boolean. I'm inclined to remove it.

PR #1006, opened 2017-11-28
New concurrency primitives

### LINKED ISSUE #993: Improve Ref

SystemFw:
~~This is a first sketch of a Ref that doesn't use Actor. I'm only running a very very simple benchmark, which shows performance to be identical to the Actor version. I think it can be  a good starting point, so I reckon having a PR will facilitate discussion.~~

This PR introduces a new concurrency scheme based on `Ref` + `Promise` . 
It offers a more orthogonal api which separates concurrent state and synchronisation, a simpler implementation (no Actor), and greatly improved performance (on a ballpark benchmark).
 Fixes #993


core/shared/src/main/scala/fs2/async/Ref.scala
```diff
@@ -7,57 +7,84 @@ import cats.implicits.{ catsSyntaxEither => _, _ }
...
-        waiting = waiting.updated(idf, cb)
-      } else {
-        val r = result.value
-        val id = nonce
-        ec.execute { () => cb(r -> id) }
-      }
+  private val state = new AtomicReference[Ref.State[A]]
+  state.set(new Ref.State[A](null, LinkedMap.empty, 0))
+
+  //TODO in the old implementation Read and Nevermind submit the continuation
```
mpilquist: Likely an oversight / bug.

core/shared/src/main/scala/fs2/async/Ref.scala
```diff
@@ -129,11 +156,17 @@ final class Ref[F[_],A] private[fs2] (implicit F: Effect[F], ec: ExecutionContex
...
   override def setAsyncPure(a: A): F[Unit] =
-    F.delay { actor ! Msg.Set(a, () => ()) }
+    async.fork {
+      F.async[Unit] { cb => set(a, () => cb(Right(()))) }
+    }
 
+  // TODO not sure why the shift here
   override def setSyncPure(a: A): F[Unit] =
-    F.async[Unit](cb => actor ! Msg.Set(a, () => cb(Right(())))) *> F.shift
+    F.async[Unit](cb => set(a, () => cb(Right(())))) *> F.shift
```
mpilquist: I think I just added this shift when doing the `SyncRef` PR. Without it a test was hanging -- I think because the subsequence actions flatmapped on to the returned `F[Unit]` were getting run on the actor thread (based on where the callback was being invoked). Perhaps this isn't necessary if we are consistent in only ever invoking callbacks from within an EC.

mpilquist:
Overall I really like the idea of getting rid of `Actor`.

I like the idea of getting rid of forking. Let's do all that stuff at the call site.

Some miscellaneous thoughts on refs:
- Many use cases are satisfied by `SyncRef`
- We need a ref-like type that starts unset for various algorithms
- Arguably, we could benefit from a ref-like type that supports explicitly clearing its value, returning to unset state
- Adding unset support to the actor-less `Ref` in this PR wouldn't be difficult

This leaves me wondering if we should rename this actor-less ref `MVar` and rename `SyncRef` to `Ref`.

pchlupacek:
just an initial comment. I really think we can do most of the complexity away by defining UnsetRef in terms of SyncRef. I have a feeling that 1:1 implementation with Actor variant won't work nicely here.

pchlupacek:
Ah @mpilquist excellent idea. with Ref/MVar I really like it. Lets do it!

pchlupacek:
Also I think we can remove `F` from Ref and MVar now. Do you agree?

mpilquist:
> Also I think we can remove F from Ref and MVar now. Do you agree?

How so? Do you mean moving `F` from type param at trait level down to individual methods?

SystemFw:
One thing I believe this makes possible is moving the EC and the cats-effect constraint at call site.
In particular it can be weakened in several cases (to Async at least)

pchlupacek:
@mpilquist exactly. You can then use MVar/Ref as boundary between two concurrent programs in `F`, one will be in IO fro example the other one may be scala's Future. :-)

mpilquist:
Yep, agreed then assuming we can make it work :)

pchlupacek:
I think the only places where we need that `async` functionality is actually MVar. For Ref it is necessary only with asyncXXX methods, and these can be in fact defined with sync (is not just actually async.fork(syncXXX).

SystemFw:
Sounds like this is worth pursuing, so I'll keep going at it :)

mpilquist:
Yeah agreed -- `Ref` can definitely be implemented without an `F` and `ExecutionContext` captured at trait level.

I'm not as sure about `MVar` since we have the `async` constructors and need to thread-shift callbacks (either at callback invocation site or via `IO.shift` on the returned actions). Interested to see though!

@SystemFw Looking forward to what you come up with. :)

SystemFw:
Yeah I'm not even sure that an unsettable MVar carries its weight tbh. In haskell that makes it quite deadlock prone. But definitely we can improve Ref and move EC down for a start.

pchlupacek:
@SystemFw I would start with mVar to just have read, and perhaps not implement take at first. I agree that `take` is error prone and difficult to read.

SystemFw:
`MVar` without `take` is `Ref` up to renaming (because without `take` the blocking behaviour on `set` that MVar has will deadlock).
I can implement a proper MVar if you prefer (and in fact when I originally looked at Ref it was for this reason), but then I realised, when looking at the concurrent data structures, that Ref works just fine and is simpler (in most cases)

pchlupacek:
@SystemFw up to you :-)

SystemFw:
So a couple of the things we might want to do, e.g. remove `fork`ed methods and leave those at call site, or move `F` , constraints and ECs to individual methods are blocked by the fact that we now have to conform to SyncOps.

On another note, one thing we might do is have Ref track its state (Set or Unset). When Unset the implementation will stay as it is now (or close), but when Unset it could delegate to a SyncRef. That way it would be "slow" on the blocking get and first set/modify, and faster afterwards. I'd have to think what use cases will actually benefit from this though, and if it can actually be made to work.

Thoughts on either paragraph?

mpilquist:
Do you mean `RefOps`? Feel free to delete that.

I thought about swapping out for a `SyncRef` after first set but you'd still incur some synchronization on the dispatch (even if only an additional atomic reference). I'm not necessarily against it though.

SystemFw:
Yeah you'd have to dispatch. I only want to add it (since it's going to complicate things) if we can show it's convenient from a performance perspective.

pchlupacek:
@SystemFw just a side idea. I really liked @mpilquist idea to put any unset / async behaviour behind mVar, and leave Ref essentially just synchronous with forking. I wouldn't complicate Ref with any of the `callbacks` etc. Ref shall be very fast, hence it forms primitive for all async code. 
On other hand mVar could be optimised only for initial Take/Put operation, as this is how we use it most of the time (i.e. signal of something being ready).

SystemFw:
@pchlupacek I'm not sure whether you mean having three types  (SyncRef, Ref, and MVar), or just renaming `SyncRef` to `Ref` and `Ref` to `MVar`.
I'm only using the old names (so I'm not calling Ref MVar for now) to avoid generating confusion in the discussion.

pchlupacek:
@SystemFw SyncRef => Ref, MVar as completely new type. Sorry to be not clear.

SystemFw:
Yeah I'm not against that. I'll see how far I go along modifying the current Ref, then tackle that.
The thing I'm really not sure about is having MVar be able to be emptied or not.

mpilquist:
Yeah I'm not stuck on ability to unset an `MVar` (or `take`).

SystemFw:
Just realised we are going in the direction of having IORef and MVar a la haskell

mpilquist:
Yep - not unintentional on my part. :)

pchlupacek:
yeah, thats what it is exactly :-)

SystemFw:
Right, seem we have a way forward then. I'll be working on it, let's see how it looks like in the end

SystemFw:
One thing I'd like to know from both of you is how often do you use an unitialised Ref, and then you _also_ modify it or set it more than once.
Asking because as far as I could see usage of Ref tend to fall into two main categories:
- Refs that are modified multiple times are created with `initialized`
- Refs that are created with `uninitialized` are only `set` once. They seems to act more like a pure Promise (with the enhanced ability that you can `get` from them multiple times) than MVars

I don't actually think we should remove `modify` functionality from the MVar, but perhaps it isn't worth optimising for that if that path is not often taken.

By the way, we can improve our current concurrent data structures by utilising SyncRef rather than Ref in key places (and I've heard complaints against their performance). Should I do it as part of this PR, or separately?

mpilquist:
Based on a very quick look, I found my usage is about 15:1 in favor of initialized ref vs uninitialized ref. In my cases of using an uninitialized ref, I only set the value once.

I don't have strong opinions on `MVar` vs a single-set `Promise` type.

I'm fine if the PR includes updating our concurrent data structures to use the new `Ref` type.

ChristopherDavenport:
My usage pattern for refs is similar to @mpilquist

pchlupacek:
@SystemFw the `SyncRef` or `Ref` is certainly much more used than the `Unset` type of reference. That's why we have to optimise all what we can. The MVar type reference usually forms a synchronisation block, but with Semaphore in fs2, there are very few and narrow usages for it in user code, and I never get across situation, where I need multiple reads from `unset` reference. However I see a very great deal using MVar for thread synchronisation, where exactly ability take/put multiple times may simplify a lot. Think of the fs2 synchronousQueue for example, in fact you just wrap mVar with our Queue interface and essentially that is it. 

I think any changes to async._ shall be done after this is merged. We may found surprising simplifications there...

SystemFw:
The design I'm exploring now is (Sync)Ref + Promise, and I think it's working out nicely: it's orthogonal, covers our use cases, and allows me to implement one on top of the other. You are right that a full MVar is more powerful, but it comes at a greater chance of deadlock ime, so I'd like to not use it for this. I can/will implement a full haskell style MVar as a separate concern if you prefer, but first I'd like to assess how having just (Sync)Ref + Promise works out

pchlupacek:
@SystemFw completely ok with me.

SystemFw:
Preliminary results show that Promise is waay faster than the Actor based Ref (and therefore, of this version as well), roughly 3x slower than SyncRef.

This is nice if you consider that Promise is implemented _on top_ of SyncRef, purely functionally. It does not have modify ((but it would be easy to add)  since I don't think it's needed, and  I'd like to keep (Sync)Ref and Promise orthogonal.

It also brings other things to the table which we will be able to discuss soon (today or tomorrow), e.g. moving `F: Effect` and the EC to the methods, and only requiring a Sync when creating a Promise (all ops require Effect though, like the old Ref)

pchlupacek:
@SystemFw thats excellent news! When you will have this PR for review LMK.

SystemFw:
@mpilquist @pchlupacek 

I have completed a first full sketch of the new design. There is still some work to do but don't be taken aback by the number of TODO comments: most of them are just me explaining the various design choices.

The gist: most of fs2 is now based on (Sync)Ref, a thin wrapper over AtomicReference, very performant and without any synchronisation mechanism/async capability. These are given by Promise, a super simple primitive, built on top of (Sync)Ref (only 3x slower than (Sync)Ref, and therefore way faster than the old Ref), which can be get and set with semantics similar to the ones of the old Ref.

There is still some consolidation work to do, but it's good enough to get some feedback I think

SystemFw:
I think I might have done some messy rebasing since the diff is showing some changes made by Michael :(

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * Runs `f1` and `f2` simultaneously, but only the winner gets to
+   * `set` to this `ref`. The loser continues running but its reference
+   * to this ref is severed, allowing this ref to be garbage collected
+   * if it is no longer referenced by anyone other than the loser.
+   *
+   * If the winner fails, the returned `F` fails as well, and this `ref`
+   * is not set.
+   */
+  def race(f1: F[A], f2: F[A])(implicit F: Effect[F], ec: ExecutionContext): F[Unit] = F.delay {
```
mpilquist: Might want to make this `private[fs2]` or rename it to `unsafeRace` or something. Or change to `raceWith(f1: F[A], f2: F[A])(finished: Either[A,A] => F[Unit])(implicit ..): F[Unit]` to allow cooperative cancelation of the losing task.

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -0,0 +1,167 @@
...
+    F.delay(new MsgId).flatMap(getOrWait)
+
+  // TODO I prefer to not have `setAsync`, and leave forking at call site,
+  // since we can't exploit SyncRef lazy set directly
+  // I've replaced any setAsyncPure in fs2 with fork(setSync), which would
+  // probably reveal that there's a bunch of unnecessary forks that can be avoided
+  //
+  // NOTE: this differs in behaviour from the old Ref in that by the time readers are notified
+  // the new value is already guaranteed to be in place.
+  def setSync(a: A)(implicit F: Effect[F], ec: ExecutionContext): F[Unit] = {
```
mpilquist: I agree with not providing `setAsync`. I think we can rename this to `set` now too.

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -0,0 +1,167 @@
...
+  // in Effect and use in a looser constraint, although that probably means loosing the
+  // flexibility of choosing a different EC per op
+  def empty[F[_], A](implicit F: Sync[F]): F[Promise[F, A]] =
+    F.delay(unsafeCreate[F, A])
+
+  // TODO inline?
+  private[fs2] def unsafeCreate[F[_]: Sync, A]: Promise[F, A] =
+    new Promise[F, A](new Ref(new AtomicReference(Promise.State.Unset(LinkedMap.empty))))
+
+  private final class MsgId
```
mpilquist: We could replace `MsgId` with `fs2.internal.Token`

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -0,0 +1,167 @@
...
+
+import fs2.Scheduler
+import fs2.internal.LinkedMap
+import java.util.concurrent.atomic.{AtomicBoolean,AtomicReference}
+
+import Promise._
+
+// TODO shall we move Effect and EC implicits back up here?
+// TODO scaladoc
+final class Promise[F[_], A] private [fs2] (ref: Ref[F, State[A]]) {
```
mpilquist: Do `Ref` and `Promise` need to be parameterized by `F` or can that be on each method instead?

core/shared/src/main/scala/fs2/async/async.scala
```diff
@@ -114,8 +111,8 @@ package object async {
     * nowhere.
    */
   def race[F[_]: Effect, A, B](fa: F[A], fb: F[B])(implicit ec: ExecutionContext): F[Either[A, B]] =
```
mpilquist: Perhaps we should move this `race` to `Promise` companion (we can leave the forwarder here though). We should define as `def raceWith[F[_]: Effect, A, B](fa: F[A], fb: F[B])(finished: Either[(A,F[B]),(F[A],B)] => F[Unit]): F[Either[A,B]]` and then define `race` in terms of `raceWith`.

mpilquist: Or actually:

```scala
def raceWith[F[_]: Effect, A, B, C](fa: F[A], fb: F[B])(finished: Either[(A,F[B]), (F[A],B)] => F[C]): F[C]
```

core/shared/src/main/scala/fs2/async/mutable/Semaphore.scala
```diff
@@ -85,20 +85,20 @@ object Semaphore {
...
     // semaphore is either empty, and there are number of outstanding acquires (Left)
     // or it is non-empty, and there are n permits available (Right)
-    type S = Either[Vector[(Long,async.Ref[F,Unit])], Long]
-    async.refOf[F,S](Right(n)).map { ref => new Semaphore[F] {
-      private def open(gate: async.Ref[F,Unit]): F[Unit] =
-        gate.setAsyncPure(())
+    type S = Either[Vector[(Long,async.Promise[F, Unit])], Long]
+    async.refOf[F,S](Right(n)).map { state => new Semaphore[F] {
+      private def open(gate: async.Promise[F, Unit]): F[Unit] =
+        async.fork(gate.setSync(()))
```
mpilquist: Doubt this will need to be forked

mpilquist:
@SystemFw This is looking really great. Note that most usage of `setAsync` and `setAsyncPure` in the old code was there only because those were the only operations supported in 0.9 -- back in 0.9, they were called `set`/`setPure`. We added `setSync`/`setSyncPure` in the early days of 0.10 but never went back and retrofitted the old code (especially given that the old version of `setSync` was significantly slower than `setAsync`.

SystemFw:
Right, before we continue this discussion and I address the review comments, a meta question wrt git history
Should I:
- Bite the bullet and keep working on this branch as it is, merging at the end.
- Rewrite the history now, and then keep working. It might lose the review comments but github should preserve them as "outdated"

I am inclined to do the latter as this will end up being a fairly large PR, and I'd like the final review to be as easy/uncluttered as possible. Thoughts?

mpilquist:
No preference here really -- I'd probably just rewrite history.

SystemFw:
Right, addressing review:

----------------------

 Re: position of `F`. We can either go all the way back to have `F[_]: Effect, Ec` in the constructor, or (if possible) move type parameter, constraint and EC down to the methods. The middle ground taken here is a bit weird so the two options above are better. As for which one of the two we should pick, I'm torn.

- The pro of moving everything to the methods site allows you to pick an EC at will, as well as perhaps using `Ref/Promise` at the boundary of two effect types, although I'm not sure how common the need for that is.
- The con is that more methods might need an `Effect` constraint, rather than just `Sync/Async`. Not too sure here tbh. Having Effect on the constructor means creating Promises needs an effect, and most ops need to create a Promise anyway, so perhaps this concern is not actually valid and we should indeed move everything to the methods?

-----------

Re: `race`. Good ideas there. I think it should be removed from the Promise class to the companion object, and be changed to `raceWith`.

-------------
Re: `fork` on `setSync`. I indeed suspect pretty much all the `fork(setSync)` can be changed to just `setSync` (especially because in most cases we already `fork` the `fa` that we `.flatMap(_.setSync)` on.
I had renamed it to `set` at first, but then changed my mind. The type alone doesn't tell you if the `set` is sync or `async`, so `setSync` makes it easier to know at a glance without having to check the scaladoc.

----------------

Any thoughts on Promise `set` failing on the second set, vs semantics implemented now?

SystemFw:
It's also worth mentioning that one of the tests for Topic is currently `pending` since it intermittently fail, although I am pretty sure the test has a race condition, rather than the code (since I exposed the same issue with the very first version of non-Actor Ref, which is completely different from `Promise`)

mpilquist:
> Re: position of `F`. 

How about we leave the constraints at the class level for now and get this PR to merge-able state. Then we can investigate a subsequent change that moves the constraints down to methods. That way, the impacts on existing code will be clearer.

> so setSync makes it easier to know at a glance without having to check the scaladoc

:+1:

> Any thoughts on Promise set failing on the second set, vs semantics implemented now?

:+1: for allowing double set without failing. I'm not sure whether subsequent sets should be ignored or take effect. I'm leaning towards former with a recommendation to use `MVar` if you want real double set ability.

SystemFw:
> How about we leave the constraints at the class level for now and get this PR to merge-able state. Then we can investigate a subsequent change that moves the constraints down to methods

:+1:

> for allowing double set without failing. I'm not sure whether subsequent sets should be ignored or take effect. I'm leaning towards former with a recommendation to use MVar if you want real double set ability.

I'm also generally leaning towards some version of set semantics that make Promise a single-set primitive. Single-set Promise and Ref allow you to implement any concurrent structure, including MVar (although perhaps we should provide an MVar built on Ref only for perf, since an MVar built on top on Ref + Promise is basically `Queue.bounded(1)` in pretty much every respect).
How about we add another  `setAndCheck` (open to a name change) to `Promise` that returns an `F[Boolean]` for cases when you want to know?  I don't want to make the general `set` do that to avoid and extra `.void` call in the 99% cases in which you know you only `set` once.
The only relevant code here is `AsyncPull.race` , perhaps you can have a look there to see if my assumptions make sense. The `setAndCheck` idea might not  be worth it, but it comes at a low cost so it's worth thinking about it

SystemFw:
Ah, a separate point about `MVar`.

I think that the approach taken in `Promise` (and in the old `Ref`) of not allowing unsetting makes the implementation of concurrent structures quite a bit harder to get wrong, as you have fewer points in which a deadlock can happen.

The cost is that direct usage is less powerful, but ime is pretty rare that you can work with a single MVar, most cases need structures built on top of multiple MVars, at which point the argument above applies

pchlupacek:
@mpilquist @SystemFw  would it be somehow possible to minimize this PR so I can review. I really am a bit lost what actually got changed here. 

As a result I wouldn't not change the `F` position in this PR nor introduce any new primitives. Lets do another PR to address these so we can better reason about implications.

SystemFw:
@pchlupacek 
Most of noise comes from my rebasing error. Once that's cleaned up it should be a lot clearer.
I agree with not changing the `F` position in this PR, I'll revert that and leave it to another PR.

Introducing the new primitive is however necessary, but it shouldn't be too hard to appreciate the differences once the extra noise is removed I think.

pchlupacek:
@SystemFw so you let me know when this is sort of ready for review?

SystemFw:
@pchlupacek Sure, sorry for pinging you too early, but Michael comments are enough for me to go and clean this up to a mergeable state (it might take me a couple of days to do all the legwork though)

pchlupacek:
@SystemFw ah sure no pressure :-)

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -0,0 +1,178 @@
...
+  // NOTE: this differs in behaviour from the old Ref in that by the time readers are notified
+  // the new value is already guaranteed to be in place.
+  def setSync(a: A): F[Unit] = {
+    def notifyReaders(r: State.Unset[A]): Unit =
+      r.waiting.values.foreach { cb =>
+        ec.execute { () => cb(a) }
+      }
+
+    ref.modify2 {
+      case s @ State.Set(_) => s -> F.unit
```
durban: Could this be `F.raiseError` instead? This way a second set is silently ignored. That way it would at least complain.

SystemFw: @mpilquist  said he prefers not failing on double set. I have added a `trySetSync` for cases in which you care (atm, it's just AsyncPull.race, which I will refactor a bit in a separate PR).
I don't have a strong opinion on this, I can see arguments for both.

SystemFw:
Does anyone know why the old `RefSpec` is in `shared` but all the other concurrency tests are in `jvm`?
I will have the new `RefSpec` and `PromiseSpec` in `jvm` as well, and deleted the old `RefSpec`, I think

core/shared/src/main/scala/fs2/async/Promise.scala
durban: In my opinion, silently ignoring subsequent sets is the worst of both worlds. (The 2 worlds being: (1) working mutiple sets and (2) first set succeeds and subsequent sets fail.) If this behavior remains, I'd recommend renaming it to `complete` or even `tryComplete`.

SystemFw: I'd be inclined to agree (but I do want set-once semantics). Let's see what Michael thinks.

SystemFw:
Re: why single set semantics for Promise.

Having multiple `set`s only makes sense if you can avoid overwriting a previous value that you know the reader has already consumed, which will imply blocking on a Set Promise, which in turn implies the it needs to be able to be Unset, which is basically MVar.

The key realisation however, is that this doesn't buy anything really. To build the vast majority of concurrent structures and combinators, you still need multiple MVars, i.e. MVar is a building block at the same level of abstraction of Promise, except with more complicated semantics, and a greater chance of deadlock due to multiple points in which blocking can happen. Promise achieves the same in a simpler, easier-to-get-right way.

OldRef allowed multiple `set` only to implement `modify`, which in turn was only needed because OldRef doubled down as both a synchronisation primitive and a concurrent mutable reference. Now that these two concerns have been split, single set semantics for Promise make the most sense imho.

Now the remaining question is what to do if a user tries to `set` a Promise twice, the possibilities are:
- the `F` returned by `set` is successful, but has no effect
- same as above, but a Boolean is returned to signal a successful `set` or a `noOp`
- the returned `F` fails

Note that scenarios where `set` has the potential of legitimately being called twice are pretty much "race-like" only, in which case a mechanism to signal the loser that it has lost is useful (so either a Bool or a failed F). The question is whether we want to make this the default behaviour or not, since it imposes a cost on the more common case.

I don't have a strong opinion on this latter point, but at the moment I'm edging towards having a single method for all cases, which fails when called on a Promise that's already Set

core/shared/src/main/scala/fs2/async/Promise.scala
mpilquist: I agree with set-once semantics. I don't have strong opinions on how subsequent sets are signaled. I didn't want them to fail b/c I thought it would be error prone in certain usage patterns but I suppose folks can always call .attempt and explicitly ignore the error if it's appropriate for their use case.

core/shared/src/main/scala/fs2/async/async.scala
SystemFw: I'm leaning towards leaving `race` as is in this PR and tackle it in a separate one, since it's mostly self-contained

core/shared/src/main/scala/fs2/async/Promise.scala
SystemFw: I will change that `F.unit` to `F.raiseError`

core/shared/src/main/scala/fs2/AsyncPull.scala
```diff
@@ -35,17 +35,17 @@ sealed abstract class AsyncPull[F[_],A] { self =>
...
+      promise <- async.promise[F,Either[Throwable, Either[A,B]]]
       t0 <- self.cancellableGet.run
       (a, cancelA) = t0
       t1 <- b.cancellableGet.run
       (b, cancelB) = t1
       fa = a.run.map(Left(_): Either[A, B])
       fb = b.run.map(Right(_): Either[A, B])
-      _ <-  async.fork(fa.attempt.flatMap(ref.setAsyncPure))
-      _ <-  async.fork(fb.attempt.flatMap(ref.setAsyncPure))
+      _ <-  async.fork(fa.attempt.flatMap(x => promise.setSync(x)))
```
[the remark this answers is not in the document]
SystemFw: Yes. I left it there because I already have code to make this function simpler, exploiting the fact that `set` will fail on second set (and that needs the `x`)

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -0,0 +1,178 @@
...
+import Promise._
+
+// TODO scaladoc
+final class Promise[F[_], A] private[fs2] (ref: Ref[F, State[A]])(implicit F: Effect[F], ec: ExecutionContext) {
+
+  // TODO add implementation comment on why new Token is side effectful in this case
+  // and the fact that suspend is roughly 2x faster than delay flatMap
+  def get: F[A] = F.suspend {
+    val id = new Token
+    getOrWait(id)
```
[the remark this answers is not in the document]
SystemFw: `new Token` is a side effect in this case, because `Token` are compared with reference equality, so I think that `suspend` it's needed

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  // TODO I prefer to not have `setAsync`, and leave forking at call site,
+  // since we can't exploit SyncRef lazy set directly
+  // I've replaced any setAsyncPure in fs2 with fork(setSync), which would
+  // probably reveal that there's a bunch of unnecessary forks that can be avoided
+  //
+  // NOTE: this differs in behaviour from the old Ref in that by the time readers are notified
+  // the new value is already guaranteed to be in place.
+  def setSync(a: A): F[Unit] = {
```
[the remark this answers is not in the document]
SystemFw: I had renamed it to `set`, then changed my mind. `set: F[Unit]` does not have enough information to tell you if the operation is synchronous or asynchronous, especially because it's now up to the user to `fork` to get async behaviour. `setSync` is more informative in this respect, and doesn't force you to go read the scaladoc. Not too hung up on this though

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -0,0 +1,178 @@
...
+    def notifyReaders(r: State.Unset[A]): Unit =
```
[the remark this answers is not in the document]
SystemFw: I'm thinking `trySet` can go altogether if we `raiseError` on second `set` (which I will do, I think), since one can use `handleError` or `attempt`

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -0,0 +1,178 @@
...
+
+    ref.modify2 {
+      case s @ State.Set(_) => s -> false.pure[F]
+      case u @ State.Unset(_) => State.Set(a) -> F.delay(notifyReaders(u)).as(true)
+    }.flatMap(_._2)
+  }
+
+  /** Like [[get]] but returns an `F[Unit]` that can be used to cancel the subscription. */
+  def cancellableGet: F[(F[A], F[Unit])] = F.delay {
+    val id = new Token
```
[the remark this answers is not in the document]
SystemFw: see reply above. `new Token` is a side effect, due to the use of reference equality for `Token`

core/shared/src/main/scala/fs2/async/Ref.scala
```diff
@@ -67,125 +34,70 @@ final class Ref[F[_],A] private[fs2] (implicit F: Effect[F], ec: ExecutionContex
...
-      F.map(getStamped(mid)) { case (a, id) =>
-        val set = (a: A) =>
-          F.async[Boolean] { cb => actor ! Msg.TrySet(id, a, x => cb(Right(x))) }
-        (a, set)
-      }
-    }
-
-  override def get: F[A] = F.flatMap(F.delay(new MsgId)) { mid => F.map(getStamped(mid))(_._1) }
+  def access: F[(A, A => F[Boolean])] = F.delay {
+    def snapshot = ar.get
```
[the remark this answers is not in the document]
SystemFw: mm it works regardless since `snapshot` is evaluated when I put it into the tuple, but `val` is clearer in any case, so I'll change it to that. Good catch!

core/shared/src/test/scala/fs2/async/RefSpec.scala
```diff
@@ -1,65 +1,65 @@
-package fs2
```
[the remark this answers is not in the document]
SystemFw: Yes, I already have that code, I'll push shortly

core/shared/src/main/scala/fs2/async/mutable/Signal.scala
```diff
@@ -47,8 +47,8 @@ abstract class Signal[F[_], A] extends immutable.Signal[F, A] { self =>
       def get: F[B] = self.get.map(f)
       def set(b: B): F[Unit] = self.set(g(b))
       def refresh: F[Unit] = self.refresh
-      def modify(bb: B => B): F[Change[B]] = modify2( b => (bb(b),()) ).map(_._1)
-      def modify2[B2](bb: B => (B,B2)):F[(Change[B], B2)] =
+      def modify(bb: B => B): F[Ref.Change[B]] = modify2( b => (bb(b),()) ).map(_._1)
```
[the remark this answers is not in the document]
SystemFw: I have a slight preference (in general) for keeping the top-level small but I can definitely move Change back to where it was if you prefer

SystemFw:
Replied to review comments, except

> in fact I would propose that return type to be explicitly suggesting that as F[Either[Throwable, Unit]] instead just F[Unit] 

I agree with raising error on second set, but not with an explicit Either I don't think. The vast majority of code already calls `set` once (the only exception is `race`s), and I don't want to add a `void` everywhere. Double setting here feels more suited to a failed F, and you can `attempt` in those rare cases you need to double `set` imho

core/shared/src/main/scala/fs2/async/Ref.scala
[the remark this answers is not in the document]
SystemFw: ah no, you're right!
`get` in `access` and `snapshot` is different. Great catch then!

pchlupacek:
@SystemFw re the Either[Throwable,Unit] signature. I did that for scope too, and in fact was only way to discover few subtle bugs I was trying to catch. I really think we shall do this there, as finding the double sets will be really hard.

SystemFw:
> as finding the double sets will be really hard.

I don't understand why this is though. With a double set your task will fail with an exception telling you "you have tried to set this to 1, but it was already set to 2", and you won't be forced to do `.flatMap(F.fromEither)` everywhere else. Are there any other cases apart from AsyncPull `race` where a double `set` is done?

In any case I'll do the rest of the work and we can then return on this, it's just an `attempt` away :)

core/shared/src/main/scala/fs2/async/mutable/Signal.scala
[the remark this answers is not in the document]
mpilquist: If we end up adding `MVar`, we'll need a top-level `Change` I think. If not, I'd rather keep it contained in `Ref` companion to keep top level concepts smaller.

SystemFw:
@mpilquist @pchlupacek  I think this PR is now in a mergeable state, bar the final minor decisions over the exact naming and type signatures of some operations.

Please note that I've left a few things to separate PRs:
- I haven't touched `race`. I agree with all the proposed changes, but it's a self-contained piece of work so I'd rather have it in another PR. I'm already working on it
- Benchmarks. It would be nice to benchmark this more extensively. Running the initial benchmark by @pchlupacek shows that `Promise` is just 2x slower than `Ref`, and they are both an order of magnitude faster than the old `Ref.
- Pending tests. This change has uncovered three race conditions in existing tests. Two should already be fixed in #1008 , `Topic: synchronous publish` needs fixing.
- Moving `F` to the methods. A previous commit explored the possibility of moving `F`, `Effect` and the EC down to the individual methods. This is probably the more open ended task.

Let me know what you think

PR #1007, opened 2017-11-28
[WIP] Actor-less Ref

durban:
This is an alternative to #1006. (Sorry @SystemFw, I didn't know you were also working on this, and already had tests almost passing when I've seen your PR.)

The main difference is that #1006 is (as far as I can tell) somewhat modeled after the actor-based code. This has the same API, but a from-scratch implementation.

It still has a ton of things I'm unsure about (some of them marked as FIXME). It also needs a lot of tests (as it is a whole new implementation, with unknown new bugs), and benchmarks (preferably also with contention). (I'm happy to work on all of these, if there is interest.) Tests currently pass with this, although I'm somewhat suprised by that :-)

A quick initial benchmark (with https://gist.github.com/pchlupacek/d80a947ac4fdd6f72649a5ae57cced59) seems somewhat better than the actor-based one (but of course still worse than `SyncRef`):
```
Ops: 10000000
Execution of : SYNC: setSyncPure took 0.85 s
Execution of : SYNC: setSyncPure took 0.839 s
Execution of : ASYNC: setSyncPure took 0.875 s
Execution of : SYNC: setAsyncPure took 1.184 s
Execution of : ASYNC: setAsyncPure took 1.961 s
Execution of : SYNC: get took 1.839 s
Execution of : ASYNC: get took 5.947 s
Execution of : SYNC: modify took 2.054 s
Execution of : ASYNC: modify took 6.77 s
```

### DIFF
--- a/core/jvm/src/test/scala/fs2/async/QueueSpec.scala
+++ b/core/jvm/src/test/scala/fs2/async/QueueSpec.scala
@@ -147,9 +147,9 @@ class QueueSpec extends Fs2Spec {
           f <- async.start(q.peek1)
           g <- async.start(q.peek1)
           _ <- q.enqueue1(42)
-          b <- q.offer1(43)
           x <- f
           y <- g
+          b <- q.offer1(43)
           z <- q.peek1
         } yield List(b, x, y, z)
       )).flatten shouldBe Vector(true, 42, 42, 43)
@@ -174,13 +174,11 @@ class QueueSpec extends Fs2Spec {
             q <- async.synchronousQueue[IO, Int]
             none1 <- q.timedPeek1(100.millis, scheduler)
             _ <- async.start(q.enqueue1(42))
-            f <- async.start(q.timedPeek1(1000.millis, scheduler))
             x <- q.dequeue1
-            y <- f
             none2 <- q.timedPeek1(100.millis, scheduler)
-          } yield List(none1, x, y, none2)
+          } yield List(none1, x, none2)
         )
-      }).flatten shouldBe Vector(None, 42, Some(42), None)
+      }).flatten shouldBe Vector(None, 42, None)
     }
     "peek1 synchronous None-terminated queue" in {
       runLog(Stream.eval(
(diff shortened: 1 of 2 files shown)


pchlupacek:
@durban thanks for this work. Actually I think your performance results are awesome, they clearly show significant gains already, whereas some operations were like 20x slower with actor, not they seem to be like having 2x-3x penalty, which is really great.

Lets see how @SystemFw work will progress, I really like this to happen before next milestone release.

pchlupacek:
@durban also I think would be excellent to make a contention tests for Ref, can you perhaps code some against current version of the ref, so when @SystemFw work will be done we can just use them?

SystemFw:
I think the problem with Actor is that it was implementing a retry scheme based on recursive methods and a custom made identity (the `nonce`).
This is really evident if you look at the code that uses the same scheme and the AtomicRef, and you see it's doing the compareAndSet on two layers, natively with AtomicReference, and with its own TrySet scheme.

So obviously an MVar using AtomicReferences directly would be way faster (after all, it's how you typically implement MVar). 

However, I wonder  how the old Ref would perform it we used Actor only, for example `modify`  would be just a message on Actor that modifies the state directly, rather than using the "user-land" retry scheme on top of it. I suspect it will be equally fast, since changing the Actor parts for AtomicReference didn't bring much difference in perf (which proves the overhead of Actor is actually quite low).

pchlupacek:
@SystemFw I agree, that semantic of the implementation of the Actor based Ref may be the cause instead of the Actor itself. However there are still a lot of allocations with Actor, that specifically coming from Msg[] algebra and then Actor's Node itself. Thats all what we will now get rid of. Also any code with Actor must be asynchronous by Nature (A => Unit)  whereas now, it does not. 

I also like about SyncRef, so they are much easier to reason about, compare to ones implemented with Actor.

SystemFw:
@durban Would you mind submitting your fixes to the tests of  `Queue.peek1` as a separate PR if you have time?

durban:
@SystemFw See #1008.

SystemFw:
Wow that was fast, thanks :)

I'll try a few runs then merge today or tomorrow. Thanks again.

mpilquist:
@durban I'm going to close this one now that we have a different actor-less ref. Take a look at that one and feel free to open a new PR if there are optimizations we can make.

PR #1013, opened 2017-12-05
Refactor `race`

SystemFw:


core/shared/src/main/scala/fs2/AsyncPull.scala
```diff
@@ -35,22 +35,20 @@ sealed abstract class AsyncPull[F[_],A] { self =>
...
-      (FreeC.Eval(promise.get.flatMap {
-        case Left(e) => F.raiseError[Either[A, B]](e)
-        case Right(Left(a)) => cancelB.run.as(Left(a): Either[A, B])
-        case Right(Right(b)) => cancelA.run.as(Right(b): Either[A, B])
-      }), FreeC.Eval(cancelA.run *> cancelB.run))
-    })
+      fa = a.run.map(Left.apply).attempt
+      fb = b.run.map(Right.apply).attempt
+      _ <-  async.fork {
+        fa.flatMap(x => promise.complete(x).handleErrorWith(_ => cancelA.run))
```
mpilquist: I think `cancelA` and `cancelB` are swapped on lines 46/47?

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -140,9 +110,9 @@ object Promise {
   def empty[F[_], A](implicit F: Effect[F], ec: ExecutionContext): F[Promise[F, A]] =
     F.delay(unsafeCreate[F, A])
 
-  /** Raised when trying to set a [[Promise]] that's already been set once */
-  final class AlreadySetException extends Throwable(
-    s"Trying to set an fs2 Promise that's already been set"
+  /** Raised when trying to complete a [[Promise]] that's already been completed */
+  final class AlreadyCompletedException extends Throwable(
+    s"Trying to complete an fs2 Promise that's already been completed"
```
mpilquist: Don't need the `s` interpolator

core/shared/src/main/scala/fs2/AsyncPull.scala
SystemFw: They are, and that's by design.
Since now completing on a promise gives an error if the promise is already set, we can now now if we arrived second to the race, and cancel _ourselves_. Before, we had to check the result and cancel the _other_ runner.  wdyt?

core/shared/src/main/scala/fs2/async/Promise.scala
SystemFw: True! I'll remove

core/shared/src/main/scala/fs2/AsyncPull.scala
mpilquist: Hm I see. Wouldn't that delay cancelation until after the `complete` has occurred, which is after the action has finished?

SystemFw: Or I guess I can `flatMap` on `complete` and cancel the other `F`

mpilquist: Yeah let's do that instead of reverting

SystemFw: :+1:

core/shared/src/main/scala/fs2/async/async.scala
```diff
@@ -104,21 +105,37 @@ package object async {
...
   /**
     * Returns an effect that, when run, races evaluation of `fa` and `fb`,
     * and returns the result of whichever completes first. The losing effect
     * continues to execute in the background though its result will be sent
     * nowhere.
    */
-  def race[F[_]: Effect, A, B](fa: F[A], fb: F[B])(implicit ec: ExecutionContext): F[Either[A, B]] =
-    promise[F, Either[A,B]].flatMap { p =>
-      p.race(fa.map(Left.apply), fb.map(Right.apply)) *> p.get
+private[fs2] def race[F[_], A, B](fa: F[A], fb: F[B])(implicit F: Effect[F], ec: ExecutionContext): F[Either[A, B]] =
```
lukestephenson: Late feedback, does the `race` method need to be private?  It's used from http4s.

https://github.com/http4s/http4s/blob/0d1ebe8cfbe439703dd189dd0af663d8d9bbed16/server/src/main/scala/org/http4s/server/middleware/Timeout.scala#L26

lukestephenson: On second thoughts, I think it's correct for this to be private given the `unsafe` calls it contains

SystemFw: No that isn't the reason. Those `unsafeRunAsync` are wrapped in `F`, so the method is still perfectly pure. The reason is private is because it can't guarantee resource safety for the losing `F`. We will consider making it public again I think if  `http4s` needs it, I'll think about it

SystemFw: https://github.com/http4s/http4s/issues/1583

PR #1016, opened 2017-12-09
Add cats instances for segment

jmcardon:
Includes:

* `Traverse` and `Monad` implicitly for any `Segment[A, Unit]`
* A function to create a `Monad[Segment[A, ?]]`

### DIFF
--- a/core/shared/src/main/scala/fs2/Segment.scala
+++ b/core/shared/src/main/scala/fs2/Segment.scala
@@ -2,7 +2,7 @@ package fs2
 
 import java.util.{ LinkedList => JLinkedList }
 import cats._
-import cats.implicits._
+import cats.implicits.{ catsSyntaxEither => _, _ }
 
 import Segment._
 
@@ -751,7 +751,7 @@ abstract class Segment[+O,+R] { self =>
     }
 }
 
-object Segment extends SegmentInstances {
+object Segment {
 
   /** Creates a segment with the specified values. */
   def apply[O](os: O*): Segment[O,Unit] = seq(os)
@@ -1431,9 +1431,11 @@ object Segment extends SegmentInstances {
     val deferred: JLinkedList[() => Unit] = new JLinkedList[() => Unit]
     def defer(t: => Unit): Unit = deferred.addLast(() => t)
   }
-}
 
-trait SegmentInstances {
+  def segmentSemigroupT[A, B]: Semigroup[Segment[A, B]] = new Semigroup[Segment[A, B]]{
+    def combine(x: Segment[A, B], y: Segment[A, B]): Segment[A, B] = x ++ y
+  }
+
   implicit def segmentMonoidInstance[A]: Monoid[Segment[A, Unit]] = new Monoid[Segment[A, Unit]]{
     def empty: Segment[A, Unit] = Segment.empty[A]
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -1431,3 +1432,50 @@ object Segment {
...
+      def tailRecM[A, B](a: A)(f: (A) => Segment[Either[A, B], Unit]): Segment[B, Unit] =
+        f(a).flatMap[B, Unit]{
+          case Left(l) => tailRecM(l)(f)
+          case Right(r) => Segment(r)
+        }.mapResult(_ => ())
+
+      def pure[A](x: A): Segment[A, Unit] = Segment(x)
+    }
+
+  def segmentRMonad[T]: Monad[Segment[T, ?]] = new Monad[Segment[T, ?]]{
```
jmcardon: I'm a bit iffy on this name but I Wasn't sure what to call it...

`rightAppliedSegmentMonad`???

core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -1431,3 +1432,50 @@ object Segment {
     def defer(t: => Unit): Unit = deferred.addLast(() => t)
   }
 }
+
+trait SegmentInstances {
```
mpilquist: Let's put these directly in the companion instead of in a trait.

core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -1431,3 +1432,50 @@ object Segment {
...
+  implicit def segmentMonoidInstance[A]: Monoid[Segment[A, Unit]] = new Monoid[Segment[A, Unit]]{
```
mpilquist: We could define a `Semigroup[Segment[A,R]]` too, which might necessitate instance prioritization (though perhaps not).

core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -1431,3 +1432,50 @@ object Segment {
...
+    def combine(x: Segment[A, Unit], y: Segment[A, Unit]): Segment[A, Unit] = x ++ y
+  }
+
+  implicit val defaultSegmentMonadInstance: Traverse[Segment[?, Unit]] with Monad[Segment[?, Unit]]  =
+    new Traverse[Segment[?, Unit]] with Monad[Segment[?, Unit]] {
+      def traverse[G[_], A, B](fa: Segment[A, Unit])(f: (A) => G[B])(implicit evidence$1: Applicative[G]): G[Segment[B, Unit]] =
+        Traverse[List].traverse(fa.force.toList)(f).map(Segment.seq)
+
+      def foldLeft[A, B](fa: Segment[A, Unit], b: B)(f: (B, A) => B): B =
+        fa.force.toList.foldLeft(b)(f)
```
mpilquist: Would be more efficient to implement this as `fa.fold(b)(f).force.run` I think

core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -1431,3 +1432,50 @@ object Segment {
...
+
+      def foldRight[A, B](fa: Segment[A, Unit], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+        Foldable[List].foldRight(fa.force.toList, lb)(f)
+
+      def flatMap[A, B](fa: Segment[A, Unit])(f: (A) => Segment[B, Unit]): Segment[B, Unit] =
+        fa.flatMap(f).mapResult(_ => ())
```
mpilquist: Can use `.voidResult` as short cut for `.mapResult(_ => ())`

core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -1431,3 +1432,50 @@ object Segment {
...
+        f(a).flatMap[B, Unit]{
+          case Left(l) => tailRecM(l)(f)
+          case Right(r) => Segment(r)
+        }.mapResult(_ => ())
+
+      def pure[A](x: A): Segment[A, Unit] = Segment(x)
+    }
+
+  def segmentRMonad[T]: Monad[Segment[T, ?]] = new Monad[Segment[T, ?]]{
+    def flatMap[A, B](fa: Segment[T, A])(f: (A) => Segment[T, B]): Segment[T, B] =
```
mpilquist: Extra parens on `(A)` here

core/shared/src/main/scala/fs2/Segment.scala
mpilquist: How about `def resultMonad`

mpilquist:
2.11 build failed due to differences `Either` syntax being available from both cats and fs2. Change `import cats.implicits._` to `import cats.implicits.{ catsSyntaxEither => _, _ }` to fix.

core/shared/src/main/scala/fs2/Segment.scala
jmcardon: Could the `Semigroup` be non-implicit?

The implicit monoid allows for `runFoldMonoid` easily on the default case of segment.

core/shared/src/main/scala/fs2/Segment.scala
jmcardon: Sorry 'twas my editor

aeons:
What's the reason for a `Semigroup` when we also have a `Monoid`?

PR #1019, opened 2017-12-15
Topic/interrupt scope

### LINKED ISSUE #916: Add primitive support for interruption

pchlupacek:
Resolves #916 

This still needs a cleanup and perhaps more cases. 

Still the `drained` streams won't get interrupted. 

@mpilquist @pchiusano @SystemFw  pls feedback. 

Thanks.


core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -114,32 +130,45 @@ private[fs2] object Algebra {
     , g: (B, O) => B
     , v: FreeC.ViewL[Algebra[F,O,?], Option[(Segment[O,Unit], FreeC[Algebra[F,O,?],Unit])]]
   )(implicit F: Sync[F]): F[B] = {
-
+  //  println(s"RFL(${scope.id}): ${v.get}")
```
mpilquist: Remove

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1721,15 +1718,49 @@ object Stream {
      * because `s1.interruptWhen(s2)` is never pulled for another element after the first element has been
      * emitted. To fix, consider `s.flatMap(_ => infiniteStream).interruptWhen(s2)`.
      */
-    def interruptWhen(haltWhenTrue: Stream[F,Boolean])(implicit F: Effect[F], ec: ExecutionContext): Stream[F,O] =
-      haltWhenTrue.noneTerminate.either(self.noneTerminate).
-        takeWhile(_.fold(halt => halt.map(!_).getOrElse(false), o => o.isDefined)).
-        collect { case Right(Some(i)) => i }
+    def interruptWhen(haltWhenTrue: Stream[F,Boolean])(implicit F: Effect[F], ec: ExecutionContext): Stream[F,O] = {
```
mpilquist: Do we need a similar refactoring for `pauseWhen`?

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -175,7 +202,8 @@ private[fs2] object Algebra {
             }
 
           case o: Algebra.OpenScope[F,_] =>
-            F.flatMap(scope.open) { innerScope =>
+            F.flatMap(scope.open(o.interruptible)) { innerScope =>
+              //println(s"RFL(${scope.id}): --- OPEN SCOPE -- ${innerScope.id} interruptible: ${innerScope.interruptible.nonEmpty}")
```
mpilquist: Remove

core/shared/src/main/scala/fs2/internal/RunFoldScope.scala
```diff
@@ -88,11 +97,12 @@ private[internal] final class RunFoldScope[F[_]] private (val id: Token, private
    * If this scope is currently closed, then the child scope is opened on the first
    * open ancestor of this scope.
    */
-  def open: F[RunFoldScope[F]] = {
+  def open(interruptible: Option[(Effect[F], ExecutionContext, Promise[F, Throwable])]): F[RunFoldScope[F]] = {
+   // println(s"OPENING new scope at $id")
```
mpilquist: Remove

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -52,6 +55,19 @@ private[fs2] object Algebra {
       })
     }
 
+  /**
+    * Like `scope` but allows this scope to be interrupted.
+    * Note that this may fail with `Interrupted` when interruption occured
+    */
+  private[fs2] def interruptScope[F[_], O, R](pull: FreeC[Algebra[F,O,?],R])(implicit effect: Effect[F], ec: ExecutionContext): FreeC[Algebra[F,O,?],R] = {
```
mpilquist: Can we define both this and `scope` in terms of a new method that has the bulk of this body?

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: I would like to go over all nondet combinators when that is done and clean them up. I think we may need to touch mergeXX as well. But prefferably I would like to do this as separate pr ?

core/shared/src/main/scala/fs2/internal/Algebra.scala
pchlupacek: ok

pchlupacek:
@mpilquist pushed changes, I think this shall be good to go. 
Now I will review mergeXXX pause, either etc to make sure all of them works as expected 
That will come as another PRs

pchlupacek:
@mpilquist of the final build failed, so I need to find out reason why again ...

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -205,4 +238,24 @@ private[fs2] object Algebra {
...
+  def asHandler[F[_], O](free: FreeC[Algebra[F,O,?],Unit], e: Throwable): FreeC[Algebra[F,O,?],Unit] = {
+    free.viewL.get match {
+      case FreeC.Pure(_) => FreeC.Fail(e)
+      case FreeC.Fail(e2) => FreeC.Fail(e)
+      case bound: FreeC.Bind[Algebra[F,O,?], _, Unit] =>
+        val f = bound.f.asInstanceOf[
+          Either[Throwable,Any] => FreeC[Algebra[F,O,?], Unit]]
+        val fx = bound.fx.asInstanceOf[FreeC.Eval[Algebra[F,O,?],_]].fr
+        fx match {
+          case _: Algebra.CloseScope[F,_] => FreeC.Bind(FreeC.Eval(fx), { (_: Either[Throwable, Any]) => f(Left(e)) })
```
mpilquist: What if there's another `CloseScope` in the result of `f(fx)`?

pchlupacek: Not sure if we need deep search this. I am quite certain we always get the closest to root scope that may be needed to close when flatMap fails. I was not able to come with stream structure where that would not be enough.

core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -14,6 +14,11 @@ private[fs2] sealed abstract class FreeC[F[_], +R] {
       case Left(e) => FreeC.Fail(e)
     })
 
+  def flatMap2[R2](f: Either[Throwable, R] => FreeC[F, R2]): FreeC[F, R2] =
```
mpilquist: Rename to `transformWith` (matching naming convention used in other types)

pchlupacek: kk

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -205,4 +238,24 @@ private[fs2] object Algebra {
     }
     fr.translate[Algebra[G,O,?]](algFtoG)
   }
+
+  /**
+    * Specific version of `Free.asHandler` that will output any CloseScope that may have
+    * been opened before the `rsn` was yielded
+    */
+  def asHandler[F[_], O](free: FreeC[Algebra[F,O,?],Unit], e: Throwable): FreeC[Algebra[F,O,?],Unit] = {
```
mpilquist: This seems too adhoc to me, like we're working around a flaw somewhere else in the design.

core/shared/src/main/scala/fs2/internal/Algebra.scala
mpilquist: Why do we only care about the closest `CloseScope`?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1663,11 +1660,35 @@ object Stream {
...
-              hd.map(f).foldRightLazy(Stream.fromFreeC(tl).flatMap(f))(_ ++ _).get
-            case Some(o) => f(o).get
+              // tail mail contain algebra that is critical for proper sequenced scope closure
+              // as such here in the event of the failure we will always drain the algebra
+              // for any `CloseScope` so we will close correct resources and set correct parent scope.
+              // note that resources are guaranteed to be closed by the way how Scope.close is defined, however
+              // in that case, in certain situations, the flatMap followed by `handleErrorWith` may
+              // without this change postpone closure of resources until the handler in handleErrorWith is interpreted.
+
+              def fby( s1: Either[O, Stream[F, O]], s2: => Stream[F,O2]): Stream[F, O2] = {
```
mpilquist: I'm not convinced we should ever be searching the tail for more work -- seems like this is patching for something else that's wrong.

core/shared/src/main/scala/fs2/internal/Algebra.scala
pchlupacek: Well this is all realted to https://github.com/functional-streams-for-scala/fs2/issues/1022. Any shortcuting of propert CloseScope causes the (A) resource leakage (B) interuption to be not possible. May be there is more princcipled approach, but that may // likely will require to touch algebra of either FreeC or Algebra itself

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: yes, what wrong is that we skip all cleanups on error. Thats it. And in that case we rely on the cleanups to be run by `Algebra.runFold`.

pchlupacek: If we would signal interrupt by exception, that's likely only way. We could perhaps decide it not to signal by exception but somehow build it in the FreeC itself, but I do not have a good solution for that either.

core/shared/src/main/scala/fs2/internal/Algebra.scala
mpilquist: Re: #1022, `handleErrorWith` can be defined like I suggested if we want it to introduce a scope and hence, release resources acquired during that stream. Critically, there's no resource leak there at all, just a mismatched expectation as to scoping.

mpilquist: As a thought experiment, what would happen if `++` on `Stream` introduced a new scope? Would we be able to keep the old definition of `flatMap` then?

pchlupacek: It is not unfortunatelly helping (the issue with handleErrorWith). I will bring you example of resource leak thats because of the mismatched scoping later. See the issue is that with interrupt we need to interrupt flatMaps that are defined in program AFTER the scope was defined, however stays within the `scope` of execution. i.e.  
```scala
Stream(1).flatMap { i => xxx}.interruptWhen(...).scope // that will work
Stream(1).interruptWhen(...).scope.flatMap { i => xxx} // thats the problem
```

pchlupacek: Re ++'s scope. That was one of the earliest attempts. If we introduce scope with every ++, performance sucks.

mpilquist: Yes I understand performance will be unacceptable, but my point is that all this adhoc machinery is really just trying to make up for lack of scopes on each append in a use case specific way. I'm really not convinced we should be doing this.

pchlupacek: Not sure I have good answer to that. If we won't we couldn't have interruption done properly (not just ++, but also error propagation) and that is cutting fs2 as option to define more complex programs. If we introduce scope to ++ we will sacrifice perfromance in place where we don't want to.

mpilquist: I'd like to get some thoughts from @pchiusano -- maybe he has some ideas.

pchlupacek: Ok, sure.

pchlupacek:
@mpilquist this is next iteration. 
I have introduced here the `onInterrupt` to clearly define the demarcation point of the interruption. 
Removed specific flatMap implementation. 
Although all tests pass, I am not still 100% convinced that this is right approach. For example : 

```scala
val s1 = Stream(1,0,2)
val s2 = Stream(3,4,5)
 (s1.interruptWhen(x) onInterrupt s2).flatMap { x => 
   Stream.eval(IO { x }) 
 }

```
will still fail if, the flatMap is evaluated during interruption 

What do you think?

pchlupacek:
@mpilquist hm, surprisingly the above example won't fail, was my bad. So this seems to be pretty working well. Would be nice if we could bring some type of stream that will actually fail...

mpilquist:
@pchlupacek This is looking really good, thanks for all the word you've done on this!

Can we move `Interrupted` to the top-level package since it's observable by clients now?

pchlupacek:
yes and thanks for support on this one

pchlupacek:
@mpilquist unfortunatelly this is still not good enough
```scala 
(Stream(1,0,2).covary[IO].interruptWhen(interrupt.get.attempt).onInterrupt(Stream(3,4,5)))
    .flatMap { i => Stream.eval(promise.get.map { _ => i }) }
      .collect { case x => x }

```
will fail with `Interrupted` cause. If we remove the `collect` things works as expected. 

I may need to look further how the pulls are altering the scope hierarchy, as this seems to be related to that.

mpilquist:
The collect example should have the same scope structure as:

```scala
(Stream(1,0,2).covary[IO].interruptWhen(interrupt.get.attempt).onInterrupt(Stream(3,4,5)))
    .flatMap { i => Stream.eval(promise.get.map { _ => i }) }
      .scope
```

pchlupacek:
@mpilquist , @SystemFw, pushed another iteration. 
Good news : 
 - `onInterrupt` is no longer necessary 
- #1022 is resolved (and all other interruption issues too)
- performance is fine, almost no impact (see bellow)
- added many more tests for interruption (see specifically, no 15 and 16) 

Bad news : 
- The `def s: Stream[Pure,O] = Stream(o).flatMap { _ => s }` cannot be interrupted, due to specific  implementation in flatMap. See tests here and proposed solution here that unfortunately SoE. 
- The infinite streams (i.e. `def s: Stream[Pure,O] = Stream(o) ++ s `  ) are interruptible, but requires a limit of `asHandlers` to be processed before no more stream elements are consulted. Limit is now scope based and may be configurable. I was not able to figure out something smarter. 
- The solution of #1022 introduced noticeable overhead on `handleErrorWith`. That I think is acceptable, but worth of mentioning 

Moving forward: 

I really need to resolve that first issue with SoE an single element flatMap. This is show stopper, and I was not able to figure solution to that 

I am sort of out of options how to improve this any further. So I'll appreciate any ideas guys. My guts are that to make this more cleaner, we will need to heavily alter algebra, and to be honest I even don't know how :-( atm. 

The whole complexity comes from the `Algebra.uncons` that essentially causes to put cleanups to `tail` position of the returned stream. This causes all that mess that need to be solved with this PR. I was also thinking to bring uncons in Algebra, but hence it is sort of `uncons` and `F ~> F2` I am not really sure if that would be possible at all . 

Guys, i think we have to give this a priority, hence first user examples are showing that may be related to interruption (#1030) for example

pchlupacek:
@SystemFw @mpilquist This is next iteration. All seems to apss now except one `takeThrough` doctests, which I am still investigating. Seems somehow the `run` interpreter of segment is not correct (appreciate if you take a look on it. 

I have no ideas how to make interruption in different way, that hopefully will be easy to understand.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -728,7 +728,13 @@ final class Stream[+F[_],+O] private(private val free: FreeC[Algebra[Nothing,Not
    * res0: List[Int] = List(1, 2, 3, 1, 2, 3, 1, 2)
    * }}}
    */
-  def repeat: Stream[F,O] = this ++ repeat
+  def repeat: Stream[F,O] =
```
mpilquist: How does this behave differently than `this ++ repeat`? I see that the handling of `Interrupted` is different but I'm not sure of the implications of `pure(())` vs `interruptEventually(s2.get, interrupted)`.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1977,15 +2034,24 @@ object Stream {
      * res0: List[Int] = List(1, 2, 3, 0)
      * }}}
      */
-    def handleErrorWith[O2>:O](h: Throwable => Stream[F,O2]): Stream[F,O2] =
-      Stream.fromFreeC(self.get[F,O2] handleErrorWith { e => h(e).get })
+    def handleErrorWith[O2>:O](h: Throwable => Stream[F,O2]): Stream[F,O2] = {
+      fromFreeC(Algebra.openScope[F, O2](None) flatMap { scope =>
```
mpilquist: This can be rewritten more simply as `Stream.fromFreeC(self.scope.get[F,O2] handleErrorWith { e => h(e).get })`

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -50,12 +54,23 @@ import cats.effect.Sync
...
+ * @param parent          If empty indicates root scope. If non-emtpy, indicates parent of this scope.
+ * @param interruptible   If defined, allows this scope to interrupt any of its operation. Interruption
+  *                       is performed using the supplied context.
+  *                       Normally the interruption awaits next step in Algebra to be evaluated, with exception
+  *                       of Eval, that when interruption is enabled on scope will be wrapped in race,
+  *                       that eventually allows interruption while eval is evaluating.
+  *
  */
-private[internal] final class CompileScope[F[_]] private (val id: Token, private val parent: Option[CompileScope[F]])(implicit F: Sync[F]) extends Scope[F] { self =>
+private[fs2] final class RunFoldScope[F[_]] private (
```
mpilquist: Looks like the rename from `RunFoldScope` to `CompileScope` was overridden in a merge conflict resolution.

mpilquist:
FWIW, I routinely get this failure when building this branch locally:
```
[info] - asynchronous resource allocation (3) *** FAILED *** (20 seconds, 12 milliseconds)
[info]   TestFailedException was thrown during property evaluation.
[info]     Message: -1 was not equal to 0
[info]     Location: (ResourceSafetySpec.scala:144)
```

mpilquist:
@pchlupacek After merging in the fix for `Segment#splitAt`, the build is passing again.

pchlupacek:
@mpilquist thanks a lot. I am locally experimenting still with interruption. I am not happy with the way how it is signalled (specifically the arbitrary int parameter). However still nothing I would like to get in. 
I will give it a short for this week, then if still no success I suggest (If you would agree) I will cleanup this PR and we can merge it in ?

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: ok

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: Idea here is that repeat, will always produce same chain of stream to be attached, so this does not make sense to seek the `CloseScope` in the code. Essentially shall be safe to exit the stream chain right away.

core/shared/src/main/scala/fs2/Segment.scala
```diff
@@ -1034,6 +1034,21 @@ object Segment {
       result.get
     }
 
+    /**
+      * Like `run` but allows to run `f` for each `O`.
+      * as they are processed when running this segment.
+      * Allows to perfrom efficient accumulation of `O` while running the stream.
+      */
+    final def runForEach(f: O => Unit): R = {
```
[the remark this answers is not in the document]
pchlupacek: yes. will do.

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: To clarify the interruptEventually will at best end up with `asHandler` instance of `this` and hence we are already at the end of `this` there will not be any relevant `CloseScope` attached in next of `this` evaluation.

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: after second thougth we need to close the scope also in case of normal termination. But we can write something like Algebra.scope(self).handleErrorWith(...)

pchlupacek:
@mpilquist @SystemFw after playing with this little bit more, I run out of ideas on more proper way how to implement interruption than it is currently done. 
My guess is this shall work ok in all cases I have been thinking of, but I really couldn't find a way how to counter `uncons` behaviour any better, and `uncons` gives that much flexibility, so we don't want to get back from it I bet. 

So If you agree, I would like you to do final review of this, and then we could merge. 

If you are unhappy with current interrupt implementation, I bet the only way around it is to say that only arbitrary Throwable is propagated correctly (there we don't have to look up for the correct point to recover from) and leave interrupt in old implementation with all the drawbacks it may bring.

mpilquist:
@pchlupacek A final review sounds good. I'll do that this morning and then we can merge after. Then hopefully get a milestone release out today.

core/jvm/src/test/scala/fs2/Pipe2Spec.scala
```diff
@@ -150,17 +151,70 @@ class Pipe2Spec extends Fs2Spec {
...
 
     "interrupt (2)" in forAll { (s1: PureStream[Int]) =>
+      val s = async.mutable.Semaphore[IO](0).unsafeRunSync()
+      val interrupt = Stream.emit(true) ++ Stream.eval_(s.increment)
+      // tests that termination is successful even if stream being interrupted is hung
+      runLog { s1.get.covary[IO].evalMap(_ => s.decrement).interruptWhen(interrupt) } shouldBe Vector()
+    }
+
+    "interrupt (3)" in {
+      // tests the interruption of the constant stream
```
mpilquist: Super minor style nitpick -- these comments can be either put in to the test name -- e.g. `intterupt (3) - interruption of constant stream` or wrapped in an `info("interruption of constant stream")`

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -101,12 +127,34 @@ private[internal] final class CompileScope[F[_]] private (val id: Token, private
...
+          case Some(parent) => parent.open(interruptible)
           case None => F.raiseError(throw new IllegalStateException("cannot re-open root scope"))
         }
     }
   }
 
+  def acquireResource[R](fr: F[R], release: R => F[Unit]): F[Either[Throwable, (R, Token)]] = {
+    val resource = Resource.create
+    F.flatMap(register(resource)) { mayAcquire =>
+      if (!mayAcquire) F.raiseError(Interrupted(id, 0)) // todo: not sure if we shall signal by interrupted anymore
```
mpilquist: TODO?

core/shared/src/main/scala/fs2/Interrupted.scala
```diff
@@ -0,0 +1,18 @@
+package fs2
+
+import fs2.internal.Token
+
+/**
+  * Signals interruption of the evaluation. Contains id of last scope that shall be interrupted and
+  * any children of that scope.
```
SystemFw: perhaps we can reword a bit?
"Contains id of last scope that shall be interrupted along with its children"

Assuming that the meaning is "this will interrupt the scope with id scopeId and also interrupt all the children of said scope"

core/shared/src/main/scala/fs2/Interrupted.scala
```diff
@@ -0,0 +1,18 @@
...
+  * @param scopeId  Id of the scope that shall be the last interrupted by this message
+  * @param loop     In case of infinite recursion this prevents interpreter to search for cleanups indefinitely.
+  *                 In each recursive iteration, this will increment by 1 up to limit defined in current scope,
+  *                 After which this will Interrupt stream w/o searching further for any cleanups.
+  */
+final case class Interrupted(scopeId: Token, loop: Int) extends Throwable {
```
mpilquist: Hm, the internal `Token` type leaks to public API here. Can we make `scopeId` and `loop` private vals?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1978,14 +2035,15 @@ object Stream {
...
-      Stream.fromFreeC(self.get[F,O2] handleErrorWith { e => h(e).get })
+      fromFreeC(Algebra.scope(self.get[F,O2]).handleErrorWith { e =>  h(e).get[F, O2] } )
 
     /**
      * Run the supplied effectful action at the end of this stream, regardless of how the stream terminates.
      */
     def onFinalize(f: F[Unit])(implicit F: Applicative[F]): Stream[F,O] =
       Stream.bracket(F.pure(()))(_ => self, _ => f)
 
+
```
mpilquist: Extra whitespace line

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2095,8 +2153,8 @@ object Stream {
     /** Deprecated alias for `compile.toVector`. */
     @deprecated("Use compile.toVector instead", "0.10.0")
     def runLog(implicit F: Sync[F]): F[Vector[O]] = compile.toVector
-
     /** Deprecated alias for `compile.last`. */
+
```
mpilquist: Extra whitespace line

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -88,11 +103,22 @@ private[internal] final class CompileScope[F[_]] private (val id: Token, private
...
-        val scope = new CompileScope[F](new Token(), Some(self))
+        val newScopeId = new Token
+        def iCtx = interruptible.map { case (effect, ec) =>
+          InterruptContext(
+            effect = effect
+            , ec = ec
+            , promise = Promise.unsafeCreate[F, Throwable](effect, ec)
+            , ref = Ref.unsafeCreate[F, (Option[Throwable], Boolean)]((None, false))
+            , interruptScopeId = newScopeId
+            , maxInterruptDepth = 256 // todo: want we somehow configure this ?
```
mpilquist: Can we write a separate issue for configuring this value along with maxSteps and chunkSize in Algebra?

core/shared/src/main/scala/fs2/internal/CompileScope.scala
mpilquist: Oh I see -- we used `Interrupted` to signal this condition before but now `Interrupted` has meaning that's specific to `interruptWhen`. How about a new internal exception type here, `AcquireAfterScopeClosed` or something?

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -178,6 +226,11 @@ private[internal] final class CompileScope[F[_]] private (val id: Token, private
     go(self, Catenable.empty)
   }
 
+  /** yields to true, if this scope has ancestor with given scope Id **/
+  def hasAncestor(scopeId: Token): F[Boolean] = {
+    F.map(ancestors) { c => Catenable.instance.exists(c)(_.id == scopeId) }
```
mpilquist: Note that if you import `cats.implicits._`, you can write this line as `ancestors.map(_.exists(_.id == scopeId))`

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -199,11 +252,91 @@ private[internal] final class CompileScope[F[_]] private (val id: Token, private
       }
     }
   }
+
+  // See docs on [[Scope#interrupt]]
+  def interrupt(cause: Either[Throwable, Unit]): F[Unit] = {
+    interruptible match {
+      case None => F.raiseError(new Throwable("Scope#interrupt called for Scope that cannot be interrupted"))
```
mpilquist: s/Throwable/IllegalStateException/

core/shared/src/main/scala/fs2/internal/CompileScope.scala
SystemFw: There's quite a lot of `F.map` and `F.flatMap`, but imho it hurts readability a bit due to all the parenthesis, so unless there's a specific reason not to, definitely 👍 to use cats syntax

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -199,11 +252,91 @@ private[internal] final class CompileScope[F[_]] private (val id: Token, private
...
+      case Some(iCtx) =>
+        F.flatMap(iCtx.ref.get) { case (_, signalled) =>
+          if (signalled) F.attempt(f)
+          else {
+            // we assume there will be no parallel evals/interpretation in the scope.
+            // the concurrent stream evaluation (like merge, join) shall have independent interruption
+            // for every asynchronous child scope of the stream, so this assumption shall be completely safe.
+            F.flatMap(iCtx.promise.cancellableGet) { case (get, cancel) =>
+              F.flatMap(fs2.async.race(get, F.attempt(f))(iCtx.effect, iCtx.ec)) {
+                case Right(result) => F.map(cancel)(_ => result)
```
mpilquist: With cats syntax this becomes `cancel.as(result)`

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -199,11 +252,91 @@ private[internal] final class CompileScope[F[_]] private (val id: Token, private
...
+            }
+          }
+        }
+    }
+  }
+
+
+  /**
+    * When the stream is evaluated, there may be `Eval` that needs to be cancelled early, when asynchronous interruption
+    * is taking the place.
```
SystemFw: `staking the place/taking place`

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -199,11 +252,91 @@ private[internal] final class CompileScope[F[_]] private (val id: Token, private
...
+    * This allows to augment eval so whenever this scope is interupted it will return on left the reason of interruption.
```
SystemFw: `s/interupted/interrupted`

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -1,24 +1,29 @@
...
+import cats.effect.{Effect, Sync}
 import cats.implicits._
 import fs2._
 
+import scala.concurrent.ExecutionContext
+
 private[fs2] sealed trait Algebra[F[_],O,R]
 
+// Shared algebra between fold/uncons
+private[fs2] sealed trait AlgebraShared[F[_], O, R] extends Algebra[F, O, R]
```
mpilquist: I'd like a better name here - also, let's move this trait inside Algebra companion object.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -64,126 +85,140 @@ private[fs2] object Algebra {
...
+    compileFoldLoop[F,O,B](scope, init, g, stream)
 
-  private def compileLoop[F[_],O,B](
+  private def compileUncons[F[_],X,O](
+    scope: CompileScope[F]
+    , s: FreeC[Algebra[F,O,?],Unit]
+    , chunkSize: Int
+    , maxSteps: Long
+  )(implicit F: Sync[F]): F[(CompileScope[F], Option[(Segment[O,Unit], FreeC[Algebra[F,O,?],Unit])])] = {
+    F.delay(s.viewL.get) flatMap {
```
mpilquist: Do we need to wrap `s.viewL.get` in a delay constructor?

core/shared/src/main/scala/fs2/internal/CompileScope.scala
SystemFw: are we ok with the side effects here (creating promise, ref, token)?
It's guarded by `flatMap`, but it's threading the line (however, it might be hard to write it otherwise)

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -64,126 +85,140 @@ private[fs2] object Algebra {
...
+      case done: FreeC.Pure[Algebra[F,O,?], Unit] => F.pure((scope, None))
+      case failed: FreeC.Fail[Algebra[F,O,?], Unit] => F.raiseError(failed.error)
+      case bound: FreeC.Bind[Algebra[F,O,?], _, Unit] =>
```
mpilquist: This lost a cleanup change that replaced the existential with a type variable, which then reduced need for casting below. I can make this change again after we merge though.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -64,126 +85,140 @@ private[fs2] object Algebra {
...
+        val fx = bound.fx.asInstanceOf[FreeC.Eval[Algebra[F,O,?],_]].fr
+        F.flatMap(scope.shallInterrupt) {
+          case None =>
+            fx match {
+              case output: Algebra.Output[F, O] =>
+                try {
+                  compileFoldLoop(scope, output.values.fold(acc)(g).force.run._2, g, f(Right(())))
+                }
+                catch {
+                  case err: Throwable => compileFoldLoop(scope, acc, g, f(Left(err)))
```
mpilquist: Replace `case err: Throwable` with a `case NonFatal(err)` throughout here.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -64,126 +85,140 @@ private[fs2] object Algebra {
...
-          F.suspend {
-            try compileLoop[F,O,B](scope, hd.fold(acc)(g).force.run._2, g, uncons(tl).viewL)
-            catch { case NonFatal(e) => compileLoop[F,O,B](scope, acc, g, uncons(tl.asHandler(e)).viewL) }
-          }
-      }
-
-      case failed: FreeC.Fail[Algebra[F,O,?], _] => F.raiseError(failed.error)
+              case run: Algebra.Run[F, O, r] =>
+                try {
+                  val (r, b) = run.values.fold(acc)(g).force.run
```
mpilquist: Seems like we need to use `chunkSize` and `maxSteps` here to limit this call to `run`?

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -193,6 +228,7 @@ private[fs2] object Algebra {
         case o: Output[F,O2] => Output[G,O2](o.values)
         case Run(values) => Run[G,O2,X](values)
         case Eval(value) => Eval[G,O2,X](u(value))
+        case un:Uncons[F,x,O2] => Uncons[G,x,O2](FreeC.suspend(un.s.translate(algFtoG)), un.chunkSize, un.maxSteps).asInstanceOf[Algebra[G,O2,X]]
```
mpilquist: Why is suspend needed here?

core/shared/src/main/scala/fs2/internal/Algebra.scala
mpilquist: How about `Algebra.Effectful`? Because the only constructors which aren't subtypes of it are `Output`, `Run`, and `Uncons` which by themselves are all non-effectful (stepping through an uncons may involve evaluating some nested effectful nodes of course)

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: Could you add some comments that describe why `repeat` isn't simply `this ++ repeat`? That way I don't try to simplify it 6 months from now when I forget about this. ;p

core/shared/src/main/scala/fs2/internal/CompileScope.scala
pchlupacek: was trying to avoid cats.syntax in whole CompileScope. Syntax is zero cost right?

mpilquist: Generally no -- some cats syntax is zero cost but anything provided via simulacrum is not. I'm fine with not using it in `CompileScope` to avoid allocations.

core/shared/src/main/scala/fs2/internal/Algebra.scala
pchlupacek: Yes. I am still playing with Scope to have specific implementation for Uncons/Fold to resolve the interrupt. But yes lets name it effectfull

core/shared/src/main/scala/fs2/internal/Algebra.scala
pchlupacek: Yes, there are situation that will SoE without it. I think one of them was `transform` stack safety

core/shared/src/main/scala/fs2/internal/Algebra.scala
pchlupacek: no prob I put the existential there

core/shared/src/main/scala/fs2/internal/Algebra.scala
pchlupacek: ok

core/shared/src/main/scala/fs2/internal/Algebra.scala
pchlupacek: run shall always terminate after the fixes should it? Except when you define one that never terminates and as pure. If that would be the case, then, perhaps I will revert to code that stepped through uncons, but I don't think so it is necessary.

core/shared/src/main/scala/fs2/internal/Algebra.scala
pchlupacek: I had problems with SoE and transform. Perhaps with delaying viewL, this could be removed too.

core/shared/src/main/scala/fs2/Interrupted.scala
pchlupacek: ok

core/shared/src/main/scala/fs2/internal/Algebra.scala
mpilquist: `Stream.segment(Segment.constant(0)).evalMap(Sink.showLinesStdOut).compile.drain` <-- I'd expect this to print stuff forever but as defined now, I don't think it will print anything.

core/shared/src/main/scala/fs2/internal/CompileScope.scala
pchlupacek: kk makes sense

core/jvm/src/test/scala/fs2/Pipe2Spec.scala
pchlupacek: not sure about info sytax, xoun you just write quick example pls?

core/shared/src/main/scala/fs2/Interrupted.scala
mpilquist: Okay yeah, that's fine with me.

core/jvm/src/test/scala/fs2/Pipe2Spec.scala
mpilquist:
```scala
"interrupt (3)" in {
  info("tests interruption of the constant stream")
  ...
}
```

It just prints the info string to the test output in test reports. In this case, it's probably better in the test name itself.

core/shared/src/main/scala/fs2/internal/Algebra.scala
mpilquist: The above example works fine with this branch (because `to` uses `evalMap` which uses `flatMap` which calls `uncons`. The result is the full infinite segment, which is then lazily folded an element at a time.

I think you can ignore the need for `chunkSize` and `maxSteps` here until/if I come up with an example that fails.

core/jvm/src/test/scala/fs2/Pipe2Spec.scala
pchlupacek: I have recently experimented with that, however info() causes to print out the text at every `Gen` case, so perhaps 100x or so for every test. 
I am not sure about moving test cases to the names, not sure what that will do with tools that using names of tests to filter execution of only one single test case.

mpilquist: Ouch. Moving text description to test name works fine -- e.g., `testOnly *MergeJoinSpec -- -z interrupt` will run all the tests that include the word `interrupt` in their name

core/shared/src/main/scala/fs2/internal/CompileScope.scala
pchlupacek: @SystemFw they are guarded by modify2, I think we are on the safe side here. The scope allocation is pretty costly now. It is also quite often, i.e. every `repeatPull` (that is used by `map` for example) closes and opens scope. So I would like to leave it as it is.

core/shared/src/main/scala/fs2/internal/CompileScope.scala
pchlupacek: Are we sure there is no extra allocation coming from the cats syntax implicit?

core/shared/src/main/scala/fs2/internal/Algebra.scala
pchlupacek: I think this shall work always, hence the uncons is implemented with `splitAt`, honoring the chunkSize and maxSteps, however the runFold is implemented in `fold`. 

Also on the ChunkSize, wouldn't that mean here that chunk is always like 256 items? Just thinking loudly what that does with Stream[F, Byte]....

core/shared/src/main/scala/fs2/internal/CompileScope.scala
mpilquist: No, there's an allocation involved. I'm fine with the policy of `fs2.internal` not using cats syntax.

mpilquist:
Looks good to me. Ready for merge?

pchlupacek:
Just pushed the changes. 
I am still not giving up on resolving Interrupt in other way. However I believe we shall be in better shape that w/o this in 0.10, so I don't want to delay it. 
I am hoping to come with solution that will resolve interruption solely in compilation, but still not sure of all the aspects how to implement this.
Would be excellent to do M release now. I will work then immediately on improving the concurrent primitives (i.e. pause, merge etc) that may benefit from the new interruption mechanics.

mpilquist:
Sounds like a good plan to me. I'll merge and then release M11.

PR #1025, opened 2017-12-19
Upgrade to cats-effect-0.6

aeons:
And transitively cats-1.0.0-RC2.


rossabaker:
Reversal of prior plans: it's called 0.6, not 0.6-RC1.

PR #1028, opened 2017-12-20
Deprecated run methods and...

### LINKED ISSUE #1017: Change Stream.run to Stream.compile
Hi,

I don't know if there is a strong theoretical reason behind this choice, but every time someone explains `Stream.run` it is said that it "compiles" the `Stream`, nothing is really "ran" and it doesn't actually perform any of the effects.
Wouldn't it be better if it were already `Stream.compile`? It's, anyway, a nice analogy with a program: it is compiled and then run. Like:

`stream.compile.unsafeRunSync()`

Instead of:

```
stream.run.unsafeRunSync()
        ↓         ↳ this runs
this doesn't run
```

mpilquist:
introduced `s.compile.{drain,toVector,toList,last,fold,foldMonoid,foldSemigroup}`. Fixes #1017. See discussion in #1017 for motivation. Also, see the section added to the 0.10 migration guide:

----

In 0.9, a stream was compiled in to an effectful value via one of the `run` methods -- e.g., `run`, `runLog`, `runLast`. In 0.10, all methods which compile a stream in to an effectful value are under the `compile` prefix:

|0.9|0.10|
|---|---|
|s.run|s.compile.drain|
|s.runLog|s.compile.toVector|
| |s.compile.toList|
|s.runLast|s.compile.last|
|s.runFold|s.compile.fold|
| |s.compile.foldSemigroup|
| |s.compile.foldMonoid|

### DIFF
--- a/core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
+++ b/core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
@@ -13,7 +13,7 @@ class ConcurrentlySpec extends Fs2Spec {
 
     "when background stream fails, overall stream fails" in forAll { (s: PureStream[Int], f: Failure) =>
       val prg = Scheduler[IO](1).flatMap(scheduler => (scheduler.sleep_[IO](25.millis) ++ s.get).concurrently(f.get))
-      val throws = f.get.drain.run.attempt.unsafeRunSync.isLeft
+      val throws = f.get.drain.compile.toVector.attempt.unsafeRunSync.isLeft
       if (throws) an[Err.type] should be thrownBy runLog(prg)
       else runLog(prg)
     }
--- a/core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
+++ b/core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
@@ -13,7 +13,7 @@ class ConcurrentlySpec extends Fs2Spec {
 
     "when background stream fails, overall stream fails" in forAll { (s: PureStream[Int], f: Failure) =>
       val prg = Scheduler[IO](1).flatMap(scheduler => (scheduler.sleep_[IO](25.millis) ++ s.get).concurrently(f.get))
-      val throws = f.get.drain.compile.toVector.attempt.unsafeRunSync.isLeft
+      val throws = f.get.drain.compile.drain.attempt.unsafeRunSync.isLeft
       if (throws) an[Err.type] should be thrownBy runLog(prg)
       else runLog(prg)
     }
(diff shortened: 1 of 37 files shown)


docs/guide.md
```diff
@@ -722,6 +722,9 @@ Stream(1).covary[IO].
           onFinalize(IO { println("finalized!") }).
           take(1).
           runLog.unsafeRunSync()
```
andrelfpinto: Use `compile.toVector`.

core/shared/src/main/scala/fs2/Pipe.scala
```diff
@@ -22,7 +22,7 @@ object Pipe {
 
     // Steps `s` without overhead of resource tracking
     def stepf(s: Stream[Read,O]): Read[UO] = {
-      Algebra.runFold(
+      Algebra.compile(
```
andrelfpinto: Why isn't this `compile.fold` (the same for Pipe2)?

core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
```diff
@@ -13,7 +13,7 @@ class ConcurrentlySpec extends Fs2Spec {
 
     "when background stream fails, overall stream fails" in forAll { (s: PureStream[Int], f: Failure) =>
       val prg = Scheduler[IO](1).flatMap(scheduler => (scheduler.sleep_[IO](25.millis) ++ s.get).concurrently(f.get))
-      val throws = f.get.drain.run.attempt.unsafeRunSync.isLeft
+      val throws = f.get.drain.compile.toVector.attempt.unsafeRunSync.isLeft
```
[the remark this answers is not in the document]
mpilquist: Fixed

core/shared/src/main/scala/fs2/Pipe.scala
mpilquist: This was a hold over from an earlier version -- with the current code, this can be reimplemented in terms of `pull.uncons` and `streamNoScope`. Fixed.

docs/guide.md
mpilquist: Fixed

docs/ReadmeExample.md
```diff
@@ -94,11 +94,14 @@ scala> val written: Stream[IO, Unit] = encodedBytes.through(io.file.writeAll(Pat
 written: fs2.Stream[cats.effect.IO,Unit] = Stream(..)
 ```
 
-There are a number of ways of interpreting the stream. In this case, we call `run`, which returns a val value of the effect type, `IO`. The output of the stream is ignored - we run it solely for its effect.
+There are a number of ways of interpreting the stream. In this case, we call `compile.drain`, which returns a val value of the effect type, `IO`. The output of the stream is ignored - we compile it solely for its effect.
 
 ```scala
 scala> val task: IO[Unit] = written.run
```
andrelfpinto: Use `compile.drain`.

mpilquist: Thanks, fixed

SystemFw:
Nice cleanup on the core interpreter 👍

PR #1034, opened 2017-12-27
Fix map on chunk specializations, test laws

rossabaker:
Fixes map on the array specializations of chunk.

Introduces cats-laws to test for lawful instances.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -4,7 +4,7 @@ import scala.collection.immutable.VectorBuilder
 import scala.reflect.ClassTag
 import java.nio.{ByteBuffer => JByteBuffer}
 
-import cats.{Applicative, Eval, Foldable, Monad, Traverse}
+import cats.{Applicative, Eq, Eval, Foldable, Monad, Traverse}
 import cats.instances.all._
 
 /**
@@ -69,7 +69,12 @@ abstract class Chunk[+O] {
   def last: Option[O] = if (isEmpty) None else Some(apply(size - 1))
 
   /** Creates a new chunk by applying `f` to each element in this chunk. */
-  def map[O2](f: O => O2): Chunk[O2]
+  def map[O2](f: O => O2): Chunk[O2] = {
+    val b = new VectorBuilder[O2]
+    b.sizeHint(size)
+    for (i <- 0 until size) b += f(apply(i))
+    Chunk.indexedSeq(b.result)
+  }
 
   /** Splits this chunk in to two chunks at the specified index. */
   def splitAt(n: Int): (Chunk[O], Chunk[O]) = {
@@ -343,7 +348,6 @@ object Chunk {
     def apply(i: Int) = values(offset + i)
     protected def splitAtChunk_(n: Int): (Chunk[O], Chunk[O]) =
       Boxed(values, offset, n) -> Boxed(values, offset + n, length - n)
-    override def map[O2](f: O => O2): Chunk[O2] = seq(values.map(f))
     override def toArray[O2 >: O: ClassTag]: Array[O2] = values.slice(offset, offset + length).asInstanceOf[Array[O2]]
   }
   object Boxed { def apply[O](values: Array[O]): Boxed[O] = Boxed(values, 0, values.length) }
@@ -361,7 +365,6 @@ object Chunk {
     def at(i: Int) = values(offset + i)
     protected def splitAtChunk_(n: Int): (Chunk[Boolean], Chunk[Boolean]) =
       Booleans(values, offset, n) -> Booleans(values, offset + n, length - n)
-    override def map[O2](f: Boolean => O2): Chunk[O2] = seq(values.map(f))
     override def toArray[O2 >: Boolean: ClassTag]: Array[O2] = values.slice(offset, offset + length).asInstanceOf[Array[O2]]
   }
   object Booleans { def apply(values: Array[Boolean]): Booleans = Booleans(values, 0, values.length) }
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -456,6 +463,8 @@ object Chunk {
   }
   object Doubles { def apply(values: Array[Double]): Doubles = Doubles(values, 0, values.length) }
 
+  implicit def fs2EqForChunk[A: Eq: ClassTag]: Eq[Chunk[A]] = Eq.by(_.toArray.toVector)
```
rossabaker: This was necessary only for testing, and perhaps should accordingly move into TestUtil.  The question is whether it's wrong.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -465,6 +474,7 @@ object Chunk {
     def traverse[F[_], A, B](fa: Chunk[A])(f: A => F[B])(implicit G: Applicative[F]): F[Chunk[B]] =
       G.map(Traverse[Vector].traverse(fa.toVector)(f))(Chunk.vector)
     def pure[A](a: A): Chunk[A] = Chunk.singleton(a)
+    override def map[A, B](fa: Chunk[A])(f: A => B): Chunk[B] = fa.map(f)
```
rossabaker: The default map was lawful.  This map was not.

core/shared/src/test/scala/fs2/ChunkSpec.scala
```diff
@@ -1,128 +1,21 @@
...
+    for {
+      n <- Gen.choose(0, 100)
+      values <- Gen.containerOfN[Array, Byte](n, Arbitrary.arbByte.arbitrary)
+      offset <- Gen.choose(0, n)
+      sz <- Gen.choose(0, n - offset)
+    } yield Chunk.bytes(values, offset, sz)
+  }
+
+  checkAll("Eq[Chunk[Byte]]", EqTests[Chunk[Byte]].eqv)
+  checkAll("Monad[Chunk.Bytes]", MonadTests[Chunk].monad[Byte, Byte, Byte])
```
rossabaker: This is a bit awkward: `Chunk.Bytes` is not the right kind to be a `Monad`, but we need to test all the specializations, as that's where the bugs are.  If we proceed with this, we'll have Chunk.Ints, Chunk.Longs, etc.

Maybe we should pass the `arbByteChunk` instance in explicitly?

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -12,7 +13,8 @@ import cats.instances.all._
  * `Chunk`s can be created from a variety of collection types using methods on the `Chunk` companion
  * (e.g., `Chunk.vector`, `Chunk.seq`, `Chunk.array`). Additionally, the `Chunk` companion
  * defines a subtype of `Chunk` for each primitive type, using an unboxed primitive array.
- * To work with unboxed arrays, use methods like `toBytes` to convert a `Chunk[Byte]` to a `Chunk.Bytes`
+ * To work with unboxed arrays, use methods like `toBytes
+` to convert a `Chunk[Byte]` to a `Chunk.Bytes`
```
mpilquist: errant CRLF?

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: I think this should probably go in `TestUtil` -- or if not, then at least drop the `toArray` so we can drop the `ClassTag` constraint.

core/shared/src/test/scala/fs2/ChunkSpec.scala
mpilquist: Ah yeah, no complaints here either way

mpilquist:
Looks great!

core/shared/src/main/scala/fs2/Chunk.scala
rossabaker: Fixed

core/shared/src/main/scala/fs2/Chunk.scala
rossabaker: Since a Chunk is strict and finite, I don't see a lot of harm.  I'd overlooked `toVector`, which removes the `ClassTag` constraint.

If I were to argue for its banishment to `TestUtil`, I'd say it's because it's inconsistent with (more generous than) universal equality.

core/shared/src/test/scala/fs2/ChunkSpec.scala
```diff
@@ -19,101 +24,110 @@ class ChunkSpec extends Fs2Spec {
...
+      "isempty" in propIsEmpty[A, Chunk[A]]
+      "toarray" in propToArray[A, Chunk[A]]
+
+      if (implicitly[ClassTag[A]] == ClassTag.Byte)
+        "tobytebuffer.byte" in propToByteBuffer[Chunk[Byte]]
+
+      checkAll(s"Eq[Chunk[$of]]", EqTests[Chunk[A]].eqv)
+      checkAll(s"Monad[Chunk]", MonadTests[Chunk].monad[A, A, A])
+
+      if (testTraverse)
```
rossabaker: I'm not proud of this or the conditional above.  Maybe `testChunk` is too fancy.  If we were to repeat these tests for each specialization, we might want to formalize `ChunkLaws` in the discipline style.

core/shared/src/main/scala/fs2/Chunk.scala
rossabaker: I guess we could do better by comparing sizes and then short circuit on the first unequal A.  We don't need to materialize a vector.

core/shared/src/test/scala/fs2/TestUtil.scala
```diff
@@ -38,6 +39,9 @@ trait TestUtil extends TestUtilPlatform {
     )
   )
 
+  implicit def cogenChunk[A: Cogen: ClassTag]: Cogen[Chunk[A]] =
```
mpilquist: Can we use a `Cogen[List[A]]` to avoid the need for the `ClassTag` here?

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -343,7 +348,6 @@ object Chunk {
     def apply(i: Int) = values(offset + i)
     protected def splitAtChunk_(n: Int): (Chunk[O], Chunk[O]) =
       Boxed(values, offset, n) -> Boxed(values, offset + n, length - n)
-    override def map[O2](f: O => O2): Chunk[O2] = seq(values.map(f))
```
mpilquist: It surprises me that we don't have specialized implementations of map for array backed chunks. Maybe that's a good optimization for a separate PR.

PR #1036, opened 2017-12-27
Create Ref.fromF

Jacoby6000:
Without this, I find myself having to do things like `fa.flatMap(Ref(_))`.  That could be because I'm missing something, but I can't find any indication of that being the case.

### DIFF
--- a/core/shared/src/main/scala/fs2/async/Ref.scala
+++ b/core/shared/src/main/scala/fs2/async/Ref.scala
@@ -128,6 +128,9 @@ object Ref {
   def apply[F[_], A](a: A)(implicit F: Sync[F]): F[Ref[F, A]] =
     F.delay(new Ref[F, A](new AtomicReference[A](a)))
 
+  def fromF[F[_], A](fa: F[A])(implicit F: Sync[F]): F[Ref[F, A]] =
+    fa.map(a => new Ref[F, A](new AtomicReference(a)))
+
   /**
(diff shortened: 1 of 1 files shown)


SystemFw:
I'm a  mild 👎 to adding ops that are indeed easily recreated with `fa.flatMap(Ref(_))`.
The old Ref used to have them and imho they only add confusion.

Btw, note that `fa.map(x = new Ref[F, A](new AtomicReference(a))` is side effectful, one should indeed prefer `fa.flatMap(Ref(_)`, i..e `fa.flatMap(x => F.delay(new Ref[F, A](new AtomicReference(x))`

Jacoby6000:
Ah I didn't think about the implications of the side effect.  I had assumed that since it was happening inside of an F anyway, that it would be okay.

ChristopherDavenport:
Perhaps `Ref.eval(fa) => fa.flatMap(a => Ref(a))` which would allow there to be sound semantics with existing fs2 structures. I don't think its necessary, but could make code a little cleaner.

mpilquist:
I'm also mildly against this addition because I think the API footprint increase doesn't outweigh simplifying `fa.flatMap(Ref(_))`. I think it would also obscure the `apply` method for folks unfamiliar with the `Ref` API -- e.g., I could see folks writing things like `Ref.eval(IO.pure(x))`.

mpilquist:
Going to close this for now but feel free to reopen if you want to discuss more.

Issue #1009, opened 2017-12-02
Tests need some love


SystemFw:
First, a little puzzle:
```scala
class Sane extends AsyncFs2Spec {
  "nope!" in { IO(1).unsafeToFuture.map(_ => false shouldBe true) }
}
```
```scala
class InSane extends Fs2Spec {
  "I guess so..." in { IO(1).unsafeToFuture.map(_ => false shouldBe true) }
}
```
The first test correctly fails, but the second succeeds! This means that, at a minimum, the current RefSpec is completely broken (I'm fixing it as part of the Ref PR). I haven't investigated but there might well be other tests that should be failing and aren't. ( 😱 )

Then, I'd say we need to bring some order to the way assertions are run: some are run with `unsafeRunSync`, others with `unsafeToFuture.map` (up to the bug above), plus there are some helpers for Streams like `runLog` and `runLogF`. Ideally a single way of running them (or some guidelines on how to pick) would be great, or even better we can create a custom matcher that does what we want (I've done it for Task in specs2, don't know about scalatest).

Third, I'm a bit unsure about the criterion to put tests in jvm rather than shared, thoughts?

Fourth, if/when these points are addressed, a cleanup of the test suite in this direction needs to be done

mpilquist:
Good catch on `RefSpec`. I agree tests need to be organized / cleaned up / etc.

Ideally, tests should be in shared -- however, tests which rely on `unsafeRunSync` don't work on Scala.js if the underlying `IO` has an async boundary. In such cases, we have to use `unsafeToFuture` with assertions in the `map` call on the `Future`. However scalacheck doesn't support async props (https://github.com/rickynils/scalacheck/issues/214) so we need to call `unsafeRunSync` in any property based tests, which means either defining them in `jvm` or defining them in `shared` only when they don't have async steps.

SystemFw:
Ah, the explanation makes sense actually :)
It's good to have it written here because it's certainly non-obvious

SystemFw:
Is there any particular reason why `EventuallySupport` can't be in `shared`?

mpilquist:
Not that I recall

SystemFw:
Just found out the reason I think. `eventually` doesn't really work with `Future` (I think), so it can only be used in `unsafeRunSync` style tests, which have to be in `jvm`

SystemFw:
Ok, I'm posting rough guidelines about what can be done here:

[...]

PR #1037, opened 2017-12-27
Fix a purity breakage in `unconsAsync`, resulting in wild behaviour

SystemFw:
Anything involving `unconsAsync` (e.g.` merge`, `interruptWith`) was affected, and the side-effecting behaviour resulted in weird things happening. As an example:
```scala
object Bug {
  import fs2._
  import cats._, implicits._
  import cats.effect._

  def stepAsync(s: Stream[IO, Unit]): Stream[IO, Unit] =
    s.pull.unconsAsync.flatMap { ap =>
      ap.pull.flatMap {
        case None => Pull.done
        case Some((hd, tl)) => Pull.output(hd) >> tl.pull.echo
      }
    }.stream

  def stream = Scheduler[IO](2).flatMap { s =>
    val a = s.sleep_[IO](1.seconds) ++ Stream.eval(IO(println("tick")))

    //side effects
    val broken = stepAsync(a)
    def works = stepAsync(a)

    broken ++ broken ++ broken ++ works ++ works ++ works ++ broken ++ broken ++ broken
  }.run.unsafeRunSync
}
```
all the `broken` except the first would complete immediately.

I would have liked to add a test for this, but struggling for ideas to make it nice

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2440,12 +2440,12 @@ object Stream {
 
           Algebra.eval {
             Promise.empty[F, Either[Throwable, Res]] flatMap { p =>
-            async.fork(runStep.attempt.flatMap(p.complete(_))) as AsyncPull.readAttemptPromise(p)
+              async.fork(runStep.attempt.flatMap(p.complete(_))) as AsyncPull.readAttemptPromise(p)
+            }
           }
         }
       }
     }
-  }
 
     /**
      * Like [[uncons]], but returns a segment of no more than `n` elements.
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -7,6 +7,7 @@ import scala.concurrent.duration._
 import cats.{Applicative, Eq, Functor, Monoid, Semigroup, ~>}
 import cats.effect.{Effect, IO, Sync}
 import cats.implicits.{catsSyntaxEither => _, _}
+import fs2.async.Promise
 import fs2.async.mutable.Queue
 import fs2.internal.{Algebra, FreeC, Token}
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2426,20 +2425,22 @@ object Stream {
...
-              scope
-              , Algebra.uncons(self.get).flatMap(Algebra.output1(_))
-              , None : UO
-            ){ (_, uo) => uo.asInstanceOf[UO] } map { _ map { case (hd, tl) => (hd, fromFreeC(tl)) }}
-
-          Algebra.eval(async.fork(F.flatMap(F.attempt(runStep))(x => async.fork(p.complete(x))))) map { _ => AsyncPull.readAttemptPromise(p) }
+      type Res = Option[(Segment[O,Unit], Stream[F,O])]
+
+      Pull.eval(Promise.empty[F, Either[Throwable, Res]]) flatMap { p =>
+        Pull.fromFreeC {
```
avakhrenev: You could inline this `Pull`:
```scala
Pull.fromFreeC { Algebra.eval[F, Nothing, Promise[F, Res]](async.Promise.empty).flatMap { p =>
```

avakhrenev:
About tests: there only 2 tests for `Scheduler`. Maybe you could add a couple of tests for other `Scheduler` methods, which *also* demonstrate this issue?

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: yeah that's a good idea, I'll do that

SystemFw: In fact, the whole operation seems quite similar to `start`, so maybe it's worth thinking for a while how to make it a bit nicer

SystemFw:
Note that e.g. `retry` has its own set of tests, and now that I look at it, there's no reason `SchedulerSpec` should be in jvm and not in shared, and be perhaps unified with Retry Spec perhaps (more items for #1009 )

It's probably a good idea to have more tests for that, but to test the behaviour exhibited here I'd have to involve both a scheduler and merge, plus something like repeat... it feels a bit weird testing for referential transparency :)

pchlupacek:
lgtm. I don't think so we need tests for that a such it was more like escaping the ref transparency bounds.

SystemFw:
let me reformat real quick

PR #1040, opened 2017-12-30
Adjusted signature of fold and scan on Segment so the original result…

mpilquist:
… type is not discarded


pchlupacek:
@mpilquist thank you so much.

Issue #1046, opened 2018-01-03
Sbt `test` runs off Metaspace


pchlupacek:
After running couple of tests in sbt (~ 5), then CallSiteInitialization and MetaspaceOOM are thrown. My suspicion is that StreamApp tests seems to hold JVM Exit Hooks, that cause classes to be held in memory between `test` invocations.

guersam:
Just a workaround, but `fork in Test` for jvm projects works for repeated test invocations (scala-js doesn't run in a forked jvm.)

pchlupacek:
I think is ok to close this. Don't see any harm to fork the tests always and I think is not worth of doing something dramatic here. Thanks @guersam for help there. Closed by #1047

PR #1047, opened 2018-01-03
Workaround for #1046; use forked JVM in tests

guersam:
Not a fundamental fix, but handy in repeated tests.


guersam:
I'm able to run 10 consequent tests after this patch, instead of metaspace OOM in the third invocation before.

PR #1048, opened 2018-01-03
Auto-format source on each compilation

mpilquist:
This PR adds scalafmt to the build and configures it to run on each compilation. It's using the default scalafmt configuration plus 3 non-controversial rewrite rules (see .scalafmt.conf).

I know folks have mixed feelings on code formatters. I'm very much in favor of enabling this though, as keeping the code base in a consistent style has been a challenge given that each of the core contributors has a significantly different syntax style.

If you'd like to propose some changes from the scalafmt default configuration, please chime in -- the configuration options are documented here: http://scalameta.org/scalafmt/#Configuration I'd really like to avoid bike-shedding the configuration though, and the closer to the standard configuration, the better support we get from scalafmt. No configuration is going to make everyone happy so let's not get wrapped up in a perfect configuration.

One config param we may want to consider changing is the default line length -- scalafmt defaults this to 80 and this PR shows the results of reformatting everything to 80 columns. If anyone strongly objects to 80, we could increase slightly (100?).

Take a look at the individual commits to see the interesting changes. 99.9% of the diff is from 
6248d7d, which is the result of running default style rules against current code base.

If possible, I'd like to get a quick merge on this PR since it will conflict with any new PRs that are opened. Looking for approvals from @pchlupacek and @SystemFw.


mpilquist:
I updated the configuration using a slight variant of the settings used by http4s. Notable changes are increasing line width to 100 chars and enabling removal of redundant braces. I didn't enable `AvoidInfix` because it makes the ScalaTest matchers unreadable. I didn't turn off vertical alignment b/c I'd like to see how bad diffs are with it turned on.

pchlupacek:
I have mixed feeling about this, but hey if you like it I am on the boat.

PR #1050, opened 2018-01-04
Added special handling of Uncons and CloseScope in case of interrupt

pchlupacek:
@mpilquist there is the fix. 

I am still thinking of more proper way of interruption. I think there must be proper way to do it just in compile.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,7 +27,7 @@ lazy val commonSettings = Seq(
     "-language:postfixOps",
     "-Ypartial-unification"
   ) ++
-    (if (scalaBinaryVersion.value startsWith "2.12")
+    (if (scalaBinaryVersion.value.startsWith("2.12"))
        List(
          "-Xlint",
          "-Xfatal-warnings",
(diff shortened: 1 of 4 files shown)


mpilquist:
Awesome, thanks. I suspect we'll still have failures in at least the async resource allocation tests, unrelated to interruption -- tests 3 and 6 have failed on Travis since we merged the Algebra refactor.

pchlupacek:
But would open another PR if I will find reasons for that and fix

mpilquist:
@pchlupacek SemaphoreSpec failed: https://travis-ci.org/functional-streams-for-scala/fs2/jobs/325132310#L885

pchlupacek:
@mpilquist was looking up the semaphore failure but somehow the link you pasted is just refering to `- dequeueAvailable (118 milliseconds)` that seems to be passing. By any chance do you have the correct link?

mpilquist:
@pchlupacek Not sure, I guess Travis deleted that build log. I'll keep looking for it to repeat in subsequent builds.

pchlupacek:
@mpilquist excellent, thanks a lot

PR #1053, opened 2018-01-05
Added longer delay to `eventually` in Async Resource specs

### LINKED ISSUE #1052: Async ResourceSafety Spec is failing spuriously on (3) and (6)
There is a good chance that AsyncResource allocation tests manifests failure. This seems to be reported earlier, but since M11 the failure rate seems to be much higher.

pchlupacek:
Fixes #1052. Apparently asynchronous tests was too eager to compare that finalisers were called. The issue then manifested when system was slower or under higher load. 

Additionally, thanks to new `CompileScope` definition merged earlier, the resulting stream is not guaranteed to have all resources released when it terminates. As such, the test that required the cleanup to be called once the stream terminated was failing spuriously.

### DIFF
--- a/core/jvm/src/test/scala/fs2/ResourceSafetySpec.scala
+++ b/core/jvm/src/test/scala/fs2/ResourceSafetySpec.scala
@@ -4,6 +4,8 @@ import java.util.concurrent.atomic.AtomicLong
 import cats.effect.IO
 import cats.implicits._
 import org.scalacheck._
+import org.scalatest.concurrent.PatienceConfiguration.Timeout
+import scala.concurrent.duration._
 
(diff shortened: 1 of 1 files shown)


pchlupacek:
Seems the (6) of async resource still fails. I will investigate that further. Cannot reproduce it so far locally.

pchlupacek:
(6) seems to occasionally get in state where it would not terminate at all. so thats completely different issue.

pchlupacek:
(6) seems to be related to interruption. So I'll look into that in the different P/R that will fix it ok ?

mpilquist:
Sure

Issue #1056, opened 2018-01-06
Remove//Improve UnconsAsync


pchlupacek:
UnconsAsync is currently used by `merge`, `prefetch` and `pause`. Implementation of UnconsAsync is currently costly and every step requires non-negligible amount of work and allocations. 
There is chance, that UnconsAsync can be removed and similar functionality implemented with already presented higher-level `async._` combinators.

mpilquist:
Interested to see what you've got here. If `UnconsAsync` were removed, that would likely impact API rather drastically (e.g., would `AsyncPull` be gone then too?). Or are you thinking we'd still have `AsyncPull` and `ToPull#unconsAsync` but we wouldn't implement that API with an `UnconsAsync` constructor?

SystemFw:
I have similar thoughts. I'd definitely prefer `unconsAsync` be a normal combinator rather than a constructor, but removing its functionality altogether seems like a net decrease in expressiveness, maybe it's worth it though?
I do have to say I'm finding myself using `merge` less and less in favour of `join` for example, maybe someone can offer a counterargument in favour of "unconsAsync" style asynchrony/concurrency?

pchlupacek:
Let me play with some ideas. I think I may have something sketched in few days. Not sure if I will make it till tomorrow, but will try. Current implementation of UnconsAsync sort of `hurts` the interruption, and I think we don't need that combinator. Perhaps I am wrong, but lets see. I just want to have at lest merge to work as fast as join (or nearly as) and have same interruption guarantees as join/concurrently.

pchlupacek:
closed by #1066

PR #1059, opened 2018-01-06
Feature/interrupt with algebra

pchlupacek:
@mpilquist @SystemFw there is the interrupt implemented with Algebra. 

The idea is pretty simple, I had it on mind couple of weeks, but always failed to implement up to until now :-). Essentially the idea is that scope that is first to be interrupted is caching it`s next step in case of interruption and using that step when the interruption occurs. 

Also as part of this - 
 - Interrupted exception is gone
- Scope now takes type `O`
- `++`, `repeat`, `flatMap` reverted to M10 implementations

The implementation is bit noisy (if you can find a way how to make it cleaner I will really appreciate). Going forward I am thinking of introducing new type `Compile[F, O]` that will Contain current Scope[F, O] to cleanup algebra and also allow different compilation implementations to be plugged in (i.e. debugging compiler, javascript-optimized compiler etc..)


pchlupacek:
FWIW I think the travis failure is unrelated

SystemFw:
I think some instability in the `retry` spec is unavoidable (is testing repetitions in time), let's see if the failure is indeed a random one

pchlupacek:
@SystemFw I think the issue is that delays is mutable unguarded ListBuffer. I think that spec needs to be made stronger, as the delays are accesses potentially from multiple threads.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2793,7 +2771,7 @@ object Stream {
           val runStep =
             Algebra
               .compileScope(
-                scope,
+                scope.asInstanceOf[fs2.internal.CompileScope[F, UO]], // todo: resolve cast
```
mpilquist: Can the TODO be resolved by specifying UO as type param to getScope (instead of Nothing)?

pchlupacek: Unfortunatelly the result of `unconsAsync` is `Pull[F, Nothing, AsyncPull[F, Option[(Segment[O, Unit], Stream[F, O])]]]` so I would need then typecase UO back to Nothing, at least that's what compiler is asking me to do. 
I have plan to address in #1056.

pchlupacek:
@mpilquist @SystemFw pls let me know when you will be done reviewing this. I have ready `merge` pull request that uses new interruption, and can just make PR once this is merged.

SystemFw:
I'm in airports all day today, so it's more likely going to be tomorrow.
However, feel free to go ahead with only Michael's approval, I'm still
catching up on the core interpreter stuff

PR #1061, opened 2018-01-07
Feature/merge interrupt

pchlupacek:
merge is now implemented w/o unconsAsync : 
- there is no `race` combinator required, except the one resulting from `interrupt` in streams
- the implementation now behaves equally to `s1.concurrently(s2)`, however `concurrently` has little less ovrhead. 

There is spurious failure of `Pipe.handle errors from observing sink` test. I am not sure if the `sink` behaviour expected in test is correct. I feel very weird, that when sink is about to fail for each `O` it will just to be stopped to be observed for next `O`. I am perhaps missing something here @mpilquist any thoughts?

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2113,15 +2113,22 @@ object Stream {
       * eventually terminate with `raiseError(e)`, possibly after emitting some
       * elements of `s` first.
       *
-      * Note: `this` and `that` are each pulled for a segment. Upon receiving
-      * a segment, it is emitted downstream. Depending on how that element is
-      * processed, the remainder of `this` and `that` may never be consulted
-      * again (e.g., `a.merge(b) >> Stream.constant(0)`). A common case where
-      * this can be problematic is draining a stream that publishes to a
-      * concurrent data structure and merging it with a consumer from the same
-      * data structure. In such cases, use `consumer.concurrently(producer)`
-      * instead of `consumer.mergeHaltR(producer.drain)` to ensure the producer
-      * continues to run in parallel with consumer processing.
+      * The implementation always tries to pull one chunk from each side
+      * before waiting for it to be consumed by resulting stream.
+      * As such, there may be up to two chunks (one from each stream)
+      * waiting to be processed while the resulting stream
+      * is processing elements.
+      *
+      * Also note that if either side produces empty chunk,
+      * the processing on that side continues,
+      * w/o downstream requiring to consume result.
+      *
+      * If either side does not emit anything (i.e. as result of drain) that side
+      * will continue to run even when the resulting stream did not ask for more data.
+      *
+      * Note that even when `s1.merge(s2.drain) == s1.concurrently(s2)`, the `concurrently` alternative is
+      * more efficient.
+      *
       *
       * @example {{{
       * scala> import scala.concurrent.duration._, scala.concurrent.ExecutionContext.Implicits.global, cats.effect.IO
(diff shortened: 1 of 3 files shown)


mpilquist:
That test looks over-specified to me.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2134,43 +2141,59 @@ object Stream {
...
+                                            ec: ExecutionContext): Stream[F, O2] =
+      Stream.eval(async.semaphore(0)).flatMap { doneSem =>
+        Stream.eval(async.promise[F, Unit]).flatMap { interruptL =>
+          Stream.eval(async.promise[F, Unit]).flatMap { interruptR =>
+            Stream.eval(async.promise[F, Throwable]).flatMap { interruptY =>
+              Stream
+                .eval(async.unboundedQueue[F, Option[(F[Unit], Chunk[O2])]])
+                .flatMap { outQ => // note that the queue actually contains up to 2 max elements thanks to semaphores guarding each side.
+                  def runUpstream(s: Stream[F, O2], interrupt: Promise[F, Unit]): F[Unit] =
+                    async.semaphore(1).flatMap { quard =>
```
mpilquist: s/quard/guard/

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2134,43 +2141,59 @@ object Stream {
...
+                      s.chunks
```
mpilquist: Why operate in chunks instead of segments?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2134,43 +2141,59 @@ object Stream {
...
+                        .interruptWhen(interrupt.get.attempt)
+                        .evalMap { chunk =>
+                          if (chunk.isEmpty) F.pure(())
```
mpilquist: Why elide empty chunks?

pchlupacek: If we don't do it, stream may in fact hang with exception unable to propagate it to downstream. I had new merge (that I have added) tests failing w/o this, as some of the `Failed` Stream examples seems to emit empty chunk and then failure. I don't think so this does any harm, because if you care about empty chunks, you could always do s1.chunk.merge(s2.chunk) and they won't be elided in that case. For the Stream[F,A] I think use doesn't care about empty chunks?

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: a) Segement cannot be tested for emptiness 
b) I want the cost of processing of the `Segment` to be done concurrently between s1 and s2 instead on resulting stream, essentially utilising more concurrency.

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: Hm, this makes me thing there's something subtly wrong. We don't elide empty chunks and we use segments in `join` -- why would `merge` be different?

pchlupacek:
Re the failing test, I am more concerned how it shall behave to me. IMHO the observe shall not somehow automagically swallow failures in sink, as it does today. I think this may be failure (async/sync consistency) in diamond that manifests itself now. 
Do you think that we really need diamond? I never used it myself, and i think it is in fact quite very low level and dangerous. What do you think about getting rid of it and just implement `observe` correctly ? 
I just want to sort of agree on this before digging deeper to diamond.

mpilquist:
I'm fine with removing `diamond` though I don't see how `observe` is swallowing failures?

pchlupacek:
@mpilquist perhaps I don't understand the contract, but how if observer is a => Stream.raiseError, can terminate with something else than left and exactly once? I mean it fails once and then just emits inputs?

mpilquist:
It will terminate with a `Left` but due to the concurrency, the full segment may be emitted

pchlupacek:
Ah ok, I see now. I also see now why sync/async behaves differently. So I will think how to fix that and update this PR. Thanks.

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: ok, I'll investigate this again and see exact cause of this.

pchlupacek: Ok I remmebered the scenario why we have to emit empty streams. Essentially this comes from the following spec :
```scala
(s1: PureStream[Int], f: Failure) =>
        an[Err.type] should be thrownBy {
          s1.get
            .merge(f.get)
            .flatMap { _ =>
              Stream.eval(IO.async[Unit](_ => ()))
            }
            .compile
            .drain
            .unsafeRunSync()
        }

``` 
if `f` stream is structure of `Stream.emits() ++ Stream.fail` or similar, the stream will not fail unless that empty chunk get processed. That may not be a big deal for that example code, but Empty chunks may not be that explicit as you see there, they in fact may be a result of stream like ` Stream.emit(42).map(_ => throw Err))` where the exception appeared, was thrown but until the empty chunk preceding the failure processing(`Algebra.raiseError())is processed, stream won't fail. 

I think the `join` may perhaps need similar treatment, hence I didn't really wrote tests of its behaviour on blocking `flatMap`.

mpilquist: I'm not convinced. I think both `merge` and `join` should operate on segment structure and emit empty segments. As a property, this looks something like `s1.either(s2).segments.compile.toVector.separate == (s1.segments.compile.toVector, s2.segments.compile.toVector)`

pchlupacek: @mpilquist generally no problem with that, however we have then to release some properties, as the f: Failure currently prohibits certain tests to terminate. I mean I can always try to do `f.get.drain` instead of only `f.get` and that shall likely to work. Any thoughts?

mpilquist: Could you do `f.get.chunks.filter(_.nonEmpty)` in the impacted tests?

pchlupacek: sure, yes, no problem with that although, still, b/c if you look on Failure examples do you think user will have an idea there are empty chunks emitted before failure ? I mean how he should know in case of deadlocks that this `filter` has to be added for the combinator to work as expected? Again just reassuring myself that your suggested path is correct.

mpilquist: There's nothing particularly special about empty chunks leading to deadlocks due to the shape of the stream after merge. You could construct an example that deadlocks on singleton chunks or any other shape.

pchlupacek: ok well, lets start with this and well see how that goes. In fact, the effects shall be run by concurrently anyhow where that should not be an issue at all.

pchlupacek: @mpilquist seems that only `sink` tests are affected when I change this. So, I gonna revert it back w/o emptiness guard (I still leave this as chunk for (b) reason mentioned above). ok?

mpilquist: I think it should be at segment level -- you can always resegment/rechunk to get more concurrency out of it if workload calls for it.

mpilquist:
@pchlupacek I fixed the 2 merge tests that were failing (they just needed a `chunk`/`unchunk` normalization to force any error generating segments to happen prior to merge). This isn't a problem with `merge` but rather a requirement of the test case.

Looks like there's still an intermittent failure in `PipeSpec - observe/observeAsync - no hangs on failures`.

pchlupacek:
@mpilquist I have been investigating failures and there seems to be something odd with concurrently or at least how the sink is implemented. I have no solution neither cause identified yet.

mpilquist:
Okay, I'll take a look at that test and see if I can spot anything. I'll open add the failing case to #1063 too.

PR #1062, opened 2018-01-08
Fixed #957

### LINKED ISSUE #957: Socket.readN yields more data than requested
`Socket.readN(n)` does not in fact read "exactly N bytes" but rather *at least* `n` bytes because the backing buffer is never shrunk; it is only embiggened as necessary. So `readN` will read as many bytes as in the biggest chunk requested thus far on that `Socket`. A quick fix is to change `<` to `!=` [here](https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/io/src/main/scala/fs2/io/tcp/Socket.scala#L205). but that may have an unacceptable performance impact so shrinking may need to be special-cased; or the `readN0` method may need to be modified pull only the requested number of bytes.

mpilquist:


io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -236,28 +236,33 @@ protected[tcp] object Socket {
...
               bufferRef
                 .modify { _ =>
                   ByteBuffer.allocate(sz)
                 }
-                .map { _.now } else {
+                .map { _.now }
+            } else {
               buff.clear()
+              buff.limit(sz)
               F.pure(buff)
```
SystemFw: Is there any reason not to have 
```scala
F.delay {
  buff.clear()
  buff.limit(sz)
  buff
}
``` 
instead

io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -236,28 +236,33 @@ protected[tcp] object Socket {
...
+            if (buff.capacity() < sz) {
               bufferRef
                 .modify { _ =>
                   ByteBuffer.allocate(sz)
                 }
-                .map { _.now } else {
+                .map { _.now }
+            } else {
               buff.clear()
+              buff.limit(sz)
```
SystemFw: Is there any reason not to have

F.delay {
  buff.clear()
  buff.limit(sz)
  buff
}
instead

mpilquist: No thanks, will fix

io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -236,28 +236,34 @@ protected[tcp] object Socket {
         // buffer is also reset to be ready to be written into.
         def getBufferOf(sz: Int): F[ByteBuffer] =
           bufferRef.get.flatMap { buff =>
-            if (buff.capacity() < sz)
+            if (buff.capacity() < sz) {
               bufferRef
                 .modify { _ =>
```
SystemFw: is this any different from `setSync(ByteBuffer.allocatte(sz)`?

mpilquist: We'd need `setAndGet` or otherwise have to write `F.delay(ByteBuffer.allocate).flatMap(b => ref.setSync(b).as(b))`

SystemFw: sorry, edited too late. `flatTap` should make that more bearable, but again, not a big deal at all

mpilquist: Yay for `flatTap` -- changed!

PR #1066, opened 2018-01-09
Removed unconsAsync and AsyncPull

mpilquist:


SystemFw:
Generally speaking, I'm liking the look of this. It shrinks the api, and it always felt a bit strange to me that there were two "styles" of concurrency/asynchrony in the library.
Do we have any idea if there's a perf impact for this?
On one hand is allocating data structure, but on the other no more recursive things allocating AsyncPulls?

mpilquist:
Right, I'm pretty sure this approach is significantly faster for the reason you pointed out.

mpilquist:
Just saw your edit. I'm not sure on `merge` either. I've run in to some edge cases where the current implementation behaves a little bit differently than `join(2)`, which is definitely wrong. I'm leaning towards temporarily changing `merge` to be an alias for `join(2)` and then restoring the custom implementation once we are certain it behaves exactly the same in all cases.

SystemFw:
yeah, I should get rid of the habit of editing stuff...

> I'm not sure on merge either

Do you mean you are also not sure it should exists if not as a syntactically convenient way of joining two streams?

mpilquist:
Oh, well I guess I hadn't thought of removing it completely (especially given the `mergeHaltX` variants), but rather just defining it as an alias for `join(2)`.

SystemFw:
yeah, an alias would be good (that's what I meant by syntactically convenient).
I guess my question is:
- Am I missing something as to why `merge`'s implementation should be any different than `join(2)`?

A similar point is to what extent `concurrently` is now needed, now that interruption is fixed.
Again, not proposing we actually eliminate it, just trying to get a consistent picture

mpilquist:
`concurrently` is like `merge` in that it really doesn't need to be more than a simple alias. The only reason we'd keep the custom implementations is for performance reasons. Notably, `concurrently` is implemented with only 3 promises whereas `merge` uses 3 promises, a semaphore, and a queue, and `join` uses 2 signals, a semaphore, a queue, and a ref per inner stream.

SystemFw:
@mpilquist ok that matches the idea I was forming. I guess at the very least we can update the respective scaladoc with equivalence laws, and a note about perf

mpilquist:
OK I'm going to merge this and then look at the merge/join(2)/concurrently consistency issues for a bit. Then maybe release M12 depending on results.

SystemFw:
👍 
There's still a `todo` in the code as well

mpilquist:
Yeah, the one about using concurrently instead of join, right? That's caused by the same issue @pchlupacek has been looking in to and I'm going to investigate some now.

PR #1067, opened 2018-01-10
Bugfix/hanging nested observe

pchlupacek:
@mpilquist that fixes subtle issue with termination of the concurrently in case of failure. 

I also noticed, that merge, may suffer from similar issue. 

Also before RC1 I think we need to look on join, as currently my guess is that the join won't interrupt flatMap over blocking stream.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1654,10 +1654,14 @@ object Stream {
                 .attempt
                 .map { _.left.toOption }
                 .flatMap { r =>
-                  doneR.complete(r) >>
+                  doneR.complete(r) >> // to prevent deadlock, done must be signalled before `interruptL`
                     r.fold(F.pure(()))(interruptL.complete)
                 }
 
+            // There is slight chance that interruption in case of failure will arrive later than
+            // `self` terminates.
+            // To prevent such interruption to be `swallowed` we append stream, that results in
+            // evaluation of the result.
             Stream.eval(async.fork(runR)) >>
               self
                 .interruptWhen(interruptL.get.map(Left(_): Either[Throwable, Unit]))
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1644,7 +1644,7 @@ object Stream {
     def concurrently[O2](that: Stream[F, O2])(implicit F: Effect[F],
                                               ec: ExecutionContext): Stream[F, O] =
       Stream.eval(async.promise[F, Unit]).flatMap { interruptR =>
-        Stream.eval(async.promise[F, Unit]).flatMap { doneR =>
+        Stream.eval(async.promise[F, Option[Throwable]]).flatMap { doneR =>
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1644,23 +1644,32 @@ object Stream {
...
+                    r.fold(F.pure(()))(interruptL.complete)
                 }
 
+            // There is slight chance that interruption in case of failure will arrive later than
+            // `self` terminates.
+            // To prevent such interruption to be `swallowed` we append stream, that results in
+            // evaluation of the result.
             Stream.eval(async.fork(runR)) >>
               self
                 .interruptWhen(interruptL.get.map(Left(_): Either[Throwable, Unit]))
```
SystemFw: `.get.map(Either.left)` to remove that type annotation?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1644,23 +1644,32 @@ object Stream {
...
                 .interruptWhen(interruptR.get.attempt)
                 .compile
                 .drain
                 .attempt
-                .flatMap {
-                  case Right(_)  => doneR.complete(())
-                  case Left(err) => interruptL.complete(err) *> doneR.complete(())
+                .map { _.left.toOption }
+                .flatMap { r =>
+                  doneR.complete(r) >> // to prevent deadlock, done must be signalled before `interruptL`
```
SystemFw: I'm having some difficulty understanding why this is the case. Let's say you have the opposite:
`r.fold(F.unit)(interruptL.complete) >> doneR.complete(r)`

could you describe a scenario when this causes a deadlock?

pchlupacek: It only manifests itself when failing with exception. Think the situation when the `R` fails, but not yet signals its termination (doneR is after interruptL), then, with that failure, the `L` side is interrupted and in its onFinalize code, it awaits the `R` to complete (doneR.get) (in new implementation this is in append). That is indeed deadlock. 
Also note that failure may happen immediately `after` the `L` completes, and w/o this approach failure failed to be propagated to `L` stream in such case. 

Not sure if it is clear with explanation, but given the fact it took me like 3 days to find whats going on, I am quite sure it is not very `transparent` issue.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1644,23 +1644,32 @@ object Stream {
...
+                    r.fold(F.pure(()))(interruptL.complete)
```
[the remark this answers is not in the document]
pchlupacek: cool done

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: ok thanks

mpilquist:
Looks like the change to `Either.left` won't work

pchlupacek:
yes, seems 2.11 related ....

SystemFw:
ah, sorry about that. I had tested it on the repl and it worked with no types, but didn't think of trying with 2.11

PR #1072, opened 2018-01-21
Fixed issue with interrupting inner scope from interruptable parent.

AdamChlupacek:
Fixes the scenarios in the two new tests

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/CompileScope.scala
+++ b/core/shared/src/main/scala/fs2/internal/CompileScope.scala
@@ -113,37 +113,108 @@ private[fs2] final class CompileScope[F[_], O] private (
       interruptible: Option[
         (Effect[F], ExecutionContext, FreeC[Algebra[F, O, ?], Unit])
       ]
-  ): F[CompileScope[F, O]] =
-    F.flatMap(state.modify2 { s =>
-      if (!s.open) (s, None)
-      else {
-        val newScopeId = new Token
-        val iCtx = interruptible.map {
-          case (effect, ec, whenInterrupted) =>
-            InterruptContext[F, O](
-              effect = effect,
-              ec = ec,
-              promise = Promise.unsafeCreate[F, Option[Throwable]](effect, ec),
-              ref = Ref.unsafeCreate[F, Option[Option[Throwable]]](None),
-              interruptRoot = newScopeId,
-              whenInterrupted = whenInterrupted
-            )
-        }
-        val scope = new CompileScope[F, O](newScopeId, Some(self), iCtx.orElse(self.interruptible))
+  ): F[CompileScope[F, O]] = {
 
-        (s.copy(children = s.children :+ scope), Some(scope))
+    /**
+      * Creates a context for a new scope.
+      *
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -113,37 +113,108 @@ private[fs2] final class CompileScope[F[_], O] private (
...
+      *  The new scope is interruptible but this scope is not interruptible -
+      *     This is a new interrupt root that can be only interrupted from bottom. Create new interrupt context.
+      *
+      *  The new scope is interruptible as well as this scope is interruptible -
+      *     This is a new interrupt root that can be interrupted from bottom or as a result of
+      *     interrupting this scope. But it should not propagate its own interruption to this scope.
+      *
+      */
+    def createScopeContext: F[(Option[InterruptContext[F, O]], Token)] =
+      interruptible match {
```
pchlupacek: I see lot of noise/redundancies in this code. Cannot we simplify it ? i.e. effect, ec, promise, ref, seems to be same or fold over option, and real different is when parent is interruptible and new scope shall be root of another interruptible context.

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -113,37 +113,108 @@ private[fs2] final class CompileScope[F[_], O] private (
...
-      case (_, None)    =>
-        // This scope is already closed so try to promote the open to an ancestor; this can fail
-        // if the root scope has already been closed, in which case, we can safely throw
-        self.parent match {
-          case Some(parent) => parent.open(interruptible)
-          case None =>
-            F.raiseError(throw new IllegalStateException("cannot re-open root scope"))
+
+    F.flatMap(createScopeContext) {
+      case (context, newScopeId) =>
```
pchlupacek: can we just day this this is `iCtx` instead just `context`

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -346,30 +420,32 @@ private[fs2] final class CompileScope[F[_], O] private (
...
-    * the next stream step to take, when the evaluation was interrupted together with next scope.
-    * or the failure if the scope was interrupted by failure.
+    * However when the evaluation is normally interrupted the this evaluates on `Right` - `Left` where we signal
+    * what is the next scope from which we should calculate the next step to take.
     *
+    * Or if the evaluation is interrupted by a failure this evaluates on `Left` where the exception
+    * that caused the interruption is returned so that it can be handled.
     */
-  private[internal] def interruptibleEval[A](f: F[A]): F[Either[Throwable, Option[A]]] =
+  private[internal] def interruptibleEval[A](f: F[A]): F[Either[Throwable, Either[Token, A]]] =
```
pchlupacek: Perhaps we have to change this to F[Either[Either[Throwable, Token]], A] to prevent one allocation in case of normal evaluation.

pchlupacek:
thanks

PR #1073, opened 2018-01-24
Bugfix/translate2

pchlupacek:
This fixes translate bug in M11/RC1 plus
- fixes nested scope interruption (fix by @AdamChlupacek )
- simplifies interruption implementation in Algebra

@SystemFw @mpilquist thanks for comments

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -135,7 +135,8 @@ private[fs2] object Algebra {
   )(implicit F: Sync[F]): F[B] = {
 
     // Uncons is interrupted, fallback on `compileFoldLoop` has to be invoked
-    case class Interrupted[X]() extends UR[X]
+    // The token is a scope from which we should recover.
+    case class Interrupted[X](token: Token) extends UR[X]
     // uncons is done
     case class Done[X](result: Option[(Segment[X, Unit], FreeC[Algebra[F, X, ?], Unit])])
         extends UR[X]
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -341,26 +408,137 @@ private[fs2] object Algebra {
...
+            Algebra.segment(run.values).transformWith { r =>
+              translate0(fK, f(r), effect)
+            }
+
+          case u: Algebra.Uncons[G, x, O] =>
+            uncons[x](u.s, u.chunkSize, u.maxSteps) match {
+              case Done(r) => translate0(fK, f(Right(r)), effect)
+              case Interrupted(_) =>
+                translate0(fK,
+                           f(Left(new Throwable("Translated scope cannot be interrupted"))),
```
AdamChlupacek: Seems you forgot to properly implement interruption of translated scope here.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -63,13 +101,30 @@ private[fs2] object Algebra {
...
+        s.transformWith {
+          case Right(_) => closeScope(scope.id)
+          case Left(err) =>
+            closeScope(scope.id).transformWith {
+              case Right(_)   => raiseError(err)
+              case Left(err0) => raiseError(CompositeFailure(err, err0, Nil))
+            }
+        }
+    }
+  // FreeC.Eval[Algebra[F, O, ?], Unit](OpenScope(s, interruptible))
```
AdamChlupacek: Commented out, remove?

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -111,31 +172,39 @@ private[fs2] object Algebra {
...
-    type UnconsF[X] =
-      Function2[
-        CompileScope[F, O],
-        Either[Throwable, Option[(Segment[X, Unit], FreeC[Algebra[F, X, ?], Unit])]],
-        F[(CompileScope[F, O], FreeC[Algebra[F, O, ?], Unit])]
-      ]
+
+    // Uncons is interrupted, fallback on `compileFoldLoop` has to be invoked
+    // The token is a scope from which we should recover.
+    case class Interrupted[X](token: Token) extends UR[X]
```
mpilquist: Can we move these classes to directly under `Algebra` to ensure they don't accidentally close over any state from this method?

mpilquist: Oh I guess the close over `F`

pchlupacek: Well, there is unfortunatelly disrepance (see Continue) between ones used in compile and translate. In fact closing them inside F/G allowed me to discover another typecast bug. 

Having said that, once we will battle-test this implementation, we could mover them outside compile//translate

mpilquist:
Looks great. Let's merge once you address Adam's comments and then I'll release RC2.

pchlupacek:
ok shall be later today. so we shall be good to go fro RC2 tonite.  Also could you pls test this against scodec before releasing?

mpilquist:
Yep already did and commented on the issue - all tests pass

SystemFw:
Great stuff!

Should be ask for e.g. doobie to give this a try? Or release RC2 directly?
(although I doubt something like scodec won't fail if the implementation was faulty)

ChristopherDavenport:
Testing doobie locally against this branch my only test errors are in postres and examples, which I believe are because I don't have a posgres instance up to test against.
```
[error] (postgres / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] (example / Test / test) sbt.TestsFailedException: Tests unsuccessful
```
I don't know what tests were necessary to check, but all everything not requiring that appears to be passing on the series/0.5.x branch.

mpilquist:
@ChristopherDavenport Thanks very much for testing!

@pchlupacek Ready for merge?

ChristopherDavenport:
I have now gotten the DB operation and 100% of doobie tests now passed.

pchlupacek:
@ChristopherDavenport cool and thanks. @mpilquist yes, we could make a build.

Issue #1041, opened 2017-12-31
Create micro-site


keirlawson:
I'm new to FS2 and whilst it seems to be a really handy library I've found the documentation a little sparse which has made using/learning FS2 a bit intimidating.  Being unfamiliar with the library at this point I'm not in a good place to start helping with said documentation, however I figured what I could do is set up the various bits-and-bobs for an FS2 micro-site, as mentioned in #748 that hopefully others more knowledgeable that I could then help to populate.  Let me know if this would be helpful and if so, any particular requirements etc.

SystemFw:
:+1: :)

mpilquist:
This would be great! Looking forward to seeing what you come up with.

PR #1074, opened 2018-01-27
Microsite

keirlawson:
This PR is my first stab at setting up a FS2 microsite per #1041 

There are some limitations to this approach however after surveying various different
project's microsites I think this is probably the best possible approach.

* Links in `README.md` to other parts of the repo or site now need to be absolute,
as this file is now served as part of the site which has a different structure and
file extensions (.md converted .html).

* Having a single `tut` source directory for both site content and repo-only markdown
was not possible, as the microsite plugin insists on overriding the tut output
directory to its jekyll area in `target/`

* Currently uses the default sbt-microsite logo as there appears to be no FS2 one

* Currently not integrated into the build system, though publishing by hand is straight
forward (once the gh-pages branch is created, anyone with a private key for the
repo can simply run `sbt microsite/publishMicrosite`).  If Travis integration is
desirable it could be handled in a subsequent PR.

* I wasn't able to use the latest microsite plugin as it brings in a newer version of tut, and seemingly the fs2 docs do not compile with tut > 0.5.5

Appreciate any feedback :)


keirlawson:
Its not quite the same code as in this PR, but for a rough idea of what the site will look like I published this: https://keirlawson.github.io/fs2/

To run the site locally, install Jekyll then run `sbt microsite/makeMicrositre`, then `jekyll serve` in `site/target/site`

mpilquist:
This looks great! I upgraded to SBT 1.1 before releasing 0.10 final and that included an upgrade to the latest tut plugin. I suspect we'll be able to upgrade to latest microsites plugin now. Want to give that a shot?

keirlawson:
Ok, updated to the latest microsites plugin.

keirlawson:
Build failure appears unrelated to this PR?

SystemFw:
Yeah it should be unrelated

PR #1076, opened 2018-01-29
Changed the definition of prefetchN to use concurrently instead of join

mpilquist:


pchlupacek:
👍 btw. interruptWhen(s:Stream[F,Boolean]) could be expressed by `concurrently` too.

PR #1079, opened 2018-02-02
fs2.async.once

durban:
This PR adds `fs2.async.once`. We already have `start` for eager memoization; `once` has the same signature, but it memoizes lazily.

### DIFF
--- a/core/shared/src/main/scala/fs2/async/async.scala
+++ b/core/shared/src/main/scala/fs2/async/async.scala
@@ -109,6 +109,33 @@ package object async {
     }
 
   /**
+    * Lazily memoize `f`. For every time the returned `F[F[A]]` is
+    * bound, the effect `f` will be performed at most once (when the
+    * inner `F[A]` is bound the first time).
+    *
+    * @see `start` for eager memoization.
+    */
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/async/async.scala
```diff
@@ -108,6 +108,33 @@ package object async {
...
+        p <- promise[F, Either[Throwable, A]]
+        ch <- ref.modify {
+          case None =>
+            Some(p)
+          case s @ Some(_) =>
+            s
+        }
+        ma <- ch.previous match {
+          case None =>
+            f.attempt.flatMap(ma => p.complete(ma) *> F.pure(ma))
```
SystemFw: this can be `f.attempt.flatTap(p.complete)`.

I'm also coming around to preferring `modify2` now, like:
```scala
ref.modify2 {
   case None => Some(p) -> f.attempt.flatTap(p.complete)
   case s @ Some(other) => s -> other.get
}.flatMap(_._2)
```

But it's style, so totally up to you. 
I'm not entirely sure this belongs in fs2, but personally I like it :)

PR #1080, opened 2018-02-04
Fix documentation example

pvillega:
As per https://gitter.im/functional-streams-for-scala/fs2?at=5a778d38e217167e2c40cc1a the old code doesn't work with 0.10.0


mpilquist:
Thanks -- can you update `docs/src/guide.md` and then re-run `sbt docs/tut`?

pvillega:
I see the file moved, I'll close this PR and send another

PR #1084, opened 2018-02-14
fix of zip 

pchlupacek:
@mpilquist @SystemFw @AdamChlupacek this is a rough idea on fixing the issue with Zip. Essentially I have introduced the Tee algebra, which I think gives us much more flexibility in merging streams than we have currently. 
I believe this approach enhances currentfucntionality and is easier to reason about than somehow swapping manually scopes when necessary. 
If you can just take a quick look and give me feedback what do you thinkg. 

Also @mpilquist I found there is some tests failing now, and that seems to be related to usage of `val (chunks, r) = run.values.force.unconsAll` in compileLoop which as not used before, and seems it is not honoring the drop/tail combinators on segments. Can you perhaps take a look on that ?


SystemFw:
I'm not clear if `Tee` is supposed to be a user facing concept, or an internal detail.
My first thought would be that I prefer it as an internal detail (just to keep the concepts needed to learn the library as few as possible), in which case it should be in `internal`, right?

mpilquist:
@pchlupacek I'm not sure about `val (chunks, r) = run.values.force.unconsAll`.

Can you rewrite `zipWith_` in terms of `Tee` so we can see what it would look like? I'd rather come up with a solution that still lets folks write n-way pulls (n > 1) using the existing API instead of a new Tee API.

SystemFw:
>  I'd rather come up with a solution that still lets folks write n-way pulls (n > 1) using the existing API instead of a new Tee API.

That would be my preference as well, if possible. (I appreciate it's a hard problem, though)

pchlupacek:
@mpilquist, @SystemFw I don't think so that we could come-up with something ultimately correct yet simple and user friendly to allow n-way uncons. Only solution I am thinking of will require scope swapping before and after uncons that will require manual user procedure, and in fact this is why I came up with `Tee`.  

Essentially w/o Tee users will have to implement their `Tee`'s w/o algebra of course, but still it will not be error prone to scope switching and resource management. 

I will complete this PR to show you full implementation. Hoe then my intents will be clear. My idea with Tee was to have it as `advanced` user API, that we won't push forward, instead, users will use zipWith et all. If you will need anything more advanced, you will use `Tee` to get resource safety and scope switching for free.

pchlupacek:
Having said that all above, I hear your concerns and will try to comeup with something closer we have now.

pchlupacek:
@mpilquist @SystemFw I did couple of experiments and still cannot find a reliable way to merge two streams w/o some algebra. The main reason is that we have only one scope in compileFoldLoop, and essentially any solution with current api seems not to be able to do job. to fix this we need a Stream tobe inerpreted like this : set scope of left stream, uncons it, set scope to one of right stream, uncons it and repeat. That will severely impact uncons performance for any other streams than ones that are zipped. Imagine that this `Set/SwapScope` will have to be performed on every uncons in flatMap, pull etc. 

Note that there is no simple solution to this, hence just creation of independent scopes is not enough as resources need to be attached to correct stream scope, and they may  not interleave at all. 

Another approach, is to change that compile signature a bit, which is essentially `unsafeUncons`. Like uncons, where you must call cleanups whenever stream terminates. Because that escapes Algebra, there is need for other Zip/Tee Algebra, that allows users to work with that w/o forgetting to run cleanups at various point in time. 

If you have another idea, please share...
Thanks.

pchlupacek:
@mpilquist @SystemFw I just pushed update. I have removed Tee, and implemented solution with minimalistic changes to current API, in fact just addition to current API. 

I think regarding performance this shall have minimal to zero impact, and shall resolve the issue. 

However as mentioned earlier there is still one test failing, and that seems to be related because of the Segment bug. Can you please take a look on that ?

pchlupacek:
edit: I see there are still some zip cases that do not pass through. I am investigating these

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2448,29 +2448,28 @@ object Stream {
...
+        Pull.segment(leg1.head.zipWith(leg2.head)(f)).flatMap {
+          case Left(((), extra2)) =>
+            leg1.uncons.flatMap {
+              case None      => k2(Left((extra2, leg2.stream)))
+              case Some(tl1) => go(tl1, leg2.setHead(extra2))
+            }
+          case Right(((), extra1)) =>
+            leg2.uncons.flatMap {
+              case None      => k1(Left((extra1, leg1.stream)))
+              case Some(tl2) => go(leg1.setHead(extra1), leg2)
```
AdamChlupacek: The issues with zips you are seeing may be because you accidentally used `leg2` here instead of `tl2`

pchlupacek: ok nice catch

pchlupacek:
@mpilquist @SystemFw I think the only remaining is the Segment issue demonstrated in 
```Stream.range(0,1000).takeThrough(_ != 5).toList``` doctest. Couldy ou be so kind please and take a look on it?

**Comment by SystemFw** [comment]
I'll try looking at the Segment problem, but won't have time until the weekend (tomorrow, most likely)

**Comment by mpilquist** [comment]
@pchlupacek Fixed via https://github.com/functional-streams-for-scala/fs2/pull/1087.

File: core/shared/src/main/scala/fs2/StreamLeg.scala [hunk 2]
```diff
@@ -0,0 +1,36 @@
+package fs2
+
+import fs2.internal.{Algebra, CompileScope, FreeC}
+
+final class StreamLeg[F[_], O](
```
mpilquist: Thoughts on defining this inside `Stream` companion? `Stream.Leg[F, O]`

mpilquist: Also, can we add some ScalaDoc here?

SystemFw: :+1: to `Stream.Leg`

SystemFw:
First of all, nice work :)

Second: any thoughts on whether or not users are going to need to use `Leg` directly or not?
I guess where I'm coming from is: is it necessary to expose Leg in the public api, or should it be considered an implementation detail of `zipWith`?

AdamChlupacek:
The problem here isnt only with `zipWith` it concerns any kind of synchronous interleaving of streams. As such if user would like to create any other combinator where they would interleave multiple streams they would run into issue such as we had with `zipWith`. As such I think we should have `Leg` as a public API.

SystemFw:
@AdamChlupacek fair enough :)

core/shared/src/main/scala/fs2/StreamLeg.scala
pchlupacek: Good idea, will maek the chane + add the doc

pchlupacek:
@SystemFw @mpilquist pushed changes: 
- Moved Leg to companion and renamed it to StepLeg
- unconsLeg -> stepLeg to clearly indicate it is much different from `uncons`

If you ok with this shall we merge it & release?

SystemFw:
Great work. LGTM

pchlupacek:
Thanks. Lets see if Travis will be green. You think we can just release it as 0.10.2? I mean we added new api, but I think that change is backward binary compatible isn't it ?

pchlupacek:
ok travis is green.

SystemFw:
Someone else will have to release though. I think 0.10.2 is fine.

PR #1094, opened 2018-02-20
Adding Concurrency Primitives examples section

gvolpe:
Adding examples of the concurrency primitives present in the `fs2.async` package object, taken from the examples I've been working on [here](https://github.com/gvolpe/advanced-http4s/tree/master/src/main/scala/com/github/gvolpe/fs2).

Not sure if the layout structure seems okay to you but I can always change it, this is just the first attempt to have something. ~~I didn't use `tut` for the examples because I do not know how. It seems to work only with trivial examples but not when you use different classes and objects.~~

~~Also, I'm not sure that the relatives links added (exactly two) are working. When I generate the site locally the relative links don't get parsed correctly but neither does the Scala code.~~

Submitting this PR as @SystemFw pointed out it would be nice to have it in the microsite.

### DIFF
--- a/site/src/concurrency-primitives.md
+++ b/site/src/concurrency-primitives.md
@@ -20,7 +20,18 @@ These data structures could be very handy in a few cases scenarios. See below ex
 
 ### Concurrent Counter
 
-This is probably one of the most common uses of the primitive `Ref[F, A]`.
+This is probably one of the most common uses of the primitive `fs2.async.Ref[F, A]`.
+
+The workers will concurrently run and modify the value of the Ref so this is one possible outcome showing "#worker >> currentCount":
+
+```
+#1 >> 0
+#3 >> 0
+#2 >> 0
+#1 >> 2
+#2 >> 3
+#3 >> 3
+```
 
 ```scala
 import cats.effect.{Effect, IO}
@@ -35,15 +46,6 @@ object CounterApp extends Counter[IO]
 /**
   * Concurrent counter that demonstrates the use of [[fs2.async.Ref]].
   *
-  * The workers will concurrently run and modify the value of the Ref so this is one
-  * possible outcome showing "#worker >> currentCount":
-  *
-  * #1 >> 0
-  * #3 >> 0
-  * #2 >> 0
-  * #1 >> 2
-  * #2 >> 3
-  * #3 >> 3
   * */
 class Counter[F[_] : Effect] extends StreamApp[F] {
 
@@ -77,7 +79,13 @@ class Worker[F[_]](number: Int, ref: Ref[F, Int])
 
 ### Only Once
 
-Whenever you are in a scenario when many processes can modify the same value but you only care about the first one in doing so and stop processing, then this is the use case of a `Promise[F, A]`.
+Whenever you are in a scenario when many processes can modify the same value but you only care about the first one in doing so and stop processing, then this is the use case of a `fs2.async.Promise[F, A]`.
+
+Two processes will try to complete the promise at the same time but only one will succeed, completing the promise exactly once. The loser one will raise an error when trying to complete a promise already completed, that's why we call `attempt` on the evaluation.
+
+Notice that the loser process will remain running in the background and the program will end on completion of all of the inner streams.
+
+So it's a "race" in the sense that both processes will try to complete the promise at the same time but conceptually is different from "race". So for example, if you schedule one of the processes to run in 10 seconds from now, then the entire program will finish after 10 seconds and you can know for sure that the process completing the promise is going to be the first one.
 
 ```scala
 import cats.effect.{Effect, IO}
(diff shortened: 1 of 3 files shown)


gvolpe:
**UPDATE**: Relative links are working and all the code examples are now compiling with tut. I had a problem in my local environment (outdated jekyll) that generated weird results. 

Could you guys please take a look? @mpilquist @SystemFw

site/src/concurrency-primitives.md
```diff
@@ -0,0 +1,292 @@
...
+In the [`fs2.async` package object](https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/async/async.scala) you'll find a bunch of useful concurrency primitives:
+
+- `Ref[F, A]`
+- `Promise[F, A]`
+- `Queue[F, A]`
+- `Topic[F, A]`
+- `Signal[F, A]`
+- `Semaphore[F]`
+
+These data structures could be very handy in a few cases scenarios. See below examples of each of them originally taken from [gvolpe examples](https://github.com/gvolpe/advanced-http4s/tree/master/src/main/scala/com/github/gvolpe/fs2):
```
SystemFw: both saying cases and scenarios is redundant :)

site/src/guide.md
```diff
@@ -394,7 +394,7 @@ It flattens the nested stream, letting up to `maxOpen` inner streams run at a ti
 
 The `Effect` bound on `F` along with the `ExecutionContext` implicit parameter is required anywhere concurrency is used in the library. As mentioned earlier, users can bring their own effect types provided they also supply an `Effect` instance and have an `ExecutionContext` in implicit scope.
 
-In addition, there are a number of other concurrency primitives---asynchronous queues, signals, and semaphores. See the [`async` package object](https://github.com/functional-streams-for-scala/fs2/blob/series/0.10/core/shared/src/main/scala/fs2/async/async.scala) for more details. We'll make use of some of these in the next section when discussing how to talk to the external world.
+In addition, there are a number of other concurrency primitives---asynchronous queues, signals, and semaphores. See the [Concurrency Primitives section](concurrency-primitives) for more examples. We'll make use of some of these in the next section when discussing how to talk to the external world.
```
SystemFw: I'm thinking we should link to both the guide and the `async` file

site/src/concurrency-primitives.md
gvolpe: Maybe it should be "case scenario" instead of "cases scenarios" but as far as I know it's normal to use them in conjunction.

site/src/guide.md
gvolpe: The async package object is linked at the beginning of the concurrency primitives page, shouldn't be necessary, but I'm not against it.

SystemFw: > The async package object is linked at the beginning of the concurrency primitives page
cool, never mind then :)

site/src/concurrency-primitives.md
ChristopherDavenport: Definitely agree this a little weird as written.

site/src/concurrency-primitives.md
```diff
@@ -0,0 +1,292 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class Worker[F[_]](number: Int, ref: Ref[F, Int])(implicit F: Effect[F]) {
+
+  private val sink: Sink[F, Int] = _.evalMap(n => F.delay(println(s"#$number >> $n")))
+
+  def start: Stream[F, Unit] =
+    for {
+      _ <- Stream.eval(ref.get) to sink
```
ChristopherDavenport: personally prefer `Stream.eval(ref.get).to(sink)`

site/src/concurrency-primitives.md
```diff
@@ -0,0 +1,292 @@
...
+      _ <- Stream.eval(ref.modify(_ + 1))
+      _ <- Stream.eval(ref.get) to sink
```
ChristopherDavenport: Same as above

site/src/concurrency-primitives.md
```diff
@@ -0,0 +1,292 @@
...
+import fs2.{Scheduler, Stream, StreamApp, async}
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+
+class Buffering[F[_]](q1: Queue[F, Int], q2: Queue[F, Int])(implicit F: Effect[F]) {
+
+  def start: Stream[F, Unit] =
+    Stream(
+      Stream.range(0, 1000).covary[F] to q1.enqueue,
```
ChristopherDavenport: Again here

gvolpe:
I made the suggested changes @SystemFw @ChristopherDavenport

mpilquist:
Awesome, thanks!

Issue #969, opened 2017-11-06
Integration with cats Parallel


SystemFw:
cats has recently added a Parallel typeclass which relates types such as `Either` and `Validation`, and lets you do `(fa, fb).parMap(foo)`, which will use the "related" Applicative, while keeping consistent `<*>` and `flatMap`, and without requiring any explicit tagging by the user.
This gives us an opportunity to provide an Applicative Api in addition to `start` for parallel `F` (I like `start` so I want to keep it regardless). Should we:

- Do nothing
- Add a Parallel instance for any `F[_]: Effect`. This will be an orphan so perhaps unadvisable, but it will mean we can remove `parSequence` and `parTraverse` (already defined for every `Parallel`)
- Add a ParIO newtype, and add a `Parallel[IO, ParIO]` instance

More info on Parallel: https://github.com/typelevel/cats/pull/1837

mpilquist:
👍 

I like the idea of a parallel instance for any effect. Is there a way we could do this as a newtype so as to avoid the orphan?

SystemFw:
I don't think we can do it for any effect without orphans without
introducing explicit tagging, which would make it moot. I haven't really
thought about it that much yet though

mpilquist:
Why can't we do something like:

```scala
case class Par[F[_],A](value: F[A])
object Par {
  implicit def parallelInstanceForEffect[F[_]](implicit M: Effect[M]): Parallel[F, Par[F,?]] = new Parallel[F,Par[F,?]] {
    def applicative: Applicative[Par[F,?]] = Par.applicativeInstance
    def monad: Monad[F] = F
  }
  def applicativeInstance[F[_]](implicit F: Effect[F]): Applicative[F] = ...
}
```

SystemFw:
I'm actually thinking we will have the same problem regardless of whether we do it for `IO` or for `F` (and therefore we should prefer a solution like the one you propose). 
Namely that we would need an import or a type annotation to make the syntax work:
`(ioA, ioB).parMap((x,y) => ...)`
`parMap` would need an instance for `Parallel[F, Par[F, ?]]`, which will only be in scope if we put it in the companion object of F (and let the "functional dependency" trigger), or if we specify `parMap[IO, ParIO]((x,y) => )` or if we bring it into scope with an import.
However I haven't written any code to try, so I might well be wrong.

SystemFw:
[...]

PR #1095, opened 2018-02-25
Upgrade to cats-effect 0.9

guersam:
Now `IO` has a `Parallel` instance.

Related issue: #969


rossabaker:
I came here to see what the strategy would be on cats-effect-0.9.  I built http4s on fs2-0.10.2 and cats-effect-0.9, and all tests pass.  There were a bunch of binary compatibility exceptions, but I think they were entirely in private classes.

mpilquist:
I haven't looked at the differences yet between 0.8 and 0.9. If the public API is binary compatible, that makes this much easier.

Issue #1099, opened 2018-03-07
Zipped streams of Free blow up with CCE


tpolecat:
Running a zipped `Stream[Free` translated to `IO` blows up with a `ClassCastException` as was seen some time ago. This is with fs2 0.10.2

```scala
import cats._
import cats.effect.IO
import cats.free.Free
import cats.implicits._
import fs2.Stream

object FreeGoBoom {

  val trans: Free[Id, ?] ~> IO =
    Free.foldMap(λ[Id ~> IO](_.pure[IO]))

  val empty: Stream[Free[Id, ?], Nothing] =
    Stream.empty.covary[Free[Id, ?]]

  def main(args: Array[String]): Unit =
    (empty zip empty)
      .translate(trans)
      .compile
      .drain
      .unsafeRunSync

}
```

The exception looks like

```
[error] (run-main-3f) java.lang.ClassCastException: cats.effect.IO$Bind cannot be cast to cats.free.Free
[error] java.lang.ClassCastException: cats.effect.IO$Bind cannot be cast to cats.free.Free
[error] 	at FreeGoBoom$$anon$1.apply(SmartGcalImporter.scala:13)
[error] 	at fs2.internal.Algebra$AlgEffectSyntax$.translate$extension(Algebra.scala:54)
[error] 	at fs2.internal.Algebra$.translate0(Algebra.scala:564)
...
```

I'm sorry.

AdamChlupacek:
The issue here is that in zip we are running the two stream from within the stepLeg with an `F (Sync[F])` that is taken from the scope. As such suddenly the zip of the two `Free` have an `F` that is an `IO` thus this happens. 

One solution for this is that the zip will be taking the `Sync` for its current `F` thus getting the proper values.

AdamChlupacek:
This has an obvious additional requirement for `Sync[F]` when zipping/merging deterministically two stream. Not sure if this is acceptable for us to require in case of pure streams and such.

mpilquist:
@AdamChlupacek I really don't want to require a `Sync[F]` constraint on `zip` and derivatives. I think we should be able to zip pure streams.

AdamChlupacek:
@mpilquist I agree, thats why I did not really made a PR for it. But as of now I am not entirely sure how to do this proper, we will probably need a new case class in Algebra. I may get something like that together today. It may not look too bad.

pchlupacek:
[...]

PR #1101, opened 2018-03-08
Fix for #1099

AdamChlupacek:
Moved the running of the leg inside of algebra.


core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -62,6 +65,7 @@ private[fs2] object Algebra {
             }
           ).asInstanceOf[AlgEffect[G, O, R]]
 
+        case r: RunLeg[F, x, O]   => r.asInstanceOf[AlgEffect[G, O, R]]
```
pchlupacek: I don't think so this is sound. The inner stream has to be translated instead of cast.

AdamChlupacek:
@mpilquist @SystemFw This is now ready for review.

Issue #1100, opened 2018-03-08
Support cats-effect 0.10 and 1.0


mpilquist:
There are some major new features coming to cats-effect shortly: https://github.com/typelevel/cats-effect/milestones

Notably:
 - A new [Timer](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Timer.scala) type (0.10)
 - Cancelation support in `IO` and in the type class hierarchy (in the `Concurrent` type class) (0.10)
 - Bracket support in `IO` and in the type class hierarchy (1.0)

I've started the [topic/cats-effect-0.10](https://github.com/functional-streams-for-scala/fs2/tree/topic/cats-effect-0.10) branch to track changes in support of the 0.10 preview builds. That branch includes:
 - removal of `fs2.Scheduler` in favor of `cats.effect.Timer`
 - changing async data structures to take a `Concurrent[F]` constraint instead of an `Effect[F]` constraint
 - changing async data structures to utilize cancelation via `Concurrent` instead of ad-hoc solutions like `cancellableGet`, `timedGet`, etc.

I'm not sure if there's anything we'd like the change in FS2 once `bracket` lands in cats-effect 1.0.

I'm currently thinking we should release a binary compatible version of FS2 0.10 against cats-effect 0.10 with none of the above mentioned changes -- simply rebuild FS2 against cats-effect 0.10. Then we should release FS2 0.11 against cats-effect 1.0 when it is ready in May. Alternatively, we could release FS2 0.11 against cats-effect 0.10 with the above mentioned changes and then FS2 0.12 (or 1.0!) against cats-effect 1.0. Thoughts?

pchlupacek:
+1 to 0.11 / 0.10 and 1.0 with 1.0 
also @mpilquist  I would like to get resolution on https://github.com/typelevel/cats-effect/issues/104 before 0.11 release, as this may be necessary for any _Interruptible_ merge in fs2, so basically any concurrent primitive.

SystemFw:
> I'm currently thinking we should release a binary compatible version of FS2 0.10 against cats-effect 0.10 with none of the above mentioned changes -- simply rebuild FS2 against cats-effect 0.10. 

I like this one better, mostly because every time we break things the whole ecosystem has to do so as well. Since the changes in cats effect _might_ end up being significant, I'd err on the side of caution and avoid breaking things twice.

Not too hung up on it though.

pchlupacek:
[...]

PR #1103, opened 2018-03-10
Initial support for cats-effect 0.10 Timer and Concurrent

mpilquist:
Initial work to support cats-effect 0.10 as discussed in #1100. `Promise`, `Semaphore`, and `Queue` all support native cancelation now without resorting to ad-hoc methods like `timedGet`. `Topic` doesn't (yet) but it has never supported cancelation (it should). 

I started a 1.0 migration guide here: https://github.com/functional-streams-for-scala/fs2/blob/topic/cats-effect-0.10/docs/migration-guide-1.0.md

There are some other places we might be able take advantage of cancelation, like in `Pull.acquireCancellable` though maybe not, I'm not sure.


mpilquist:
(I accidentally targeted series/0.10 initially -- just retargeted to series/1.0)

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1291,6 +1319,14 @@ object Stream {
   def constant[O](o: O, segmentSize: Int = 256): Stream[Pure, O] =
     segment(Segment.constant(o).take(segmentSize).voidResult).repeat
 
+  /**
+    * Returns a stream that when run, sleeps for duration `d` and then pulls from `s`.
+    *
+    * Alias for `sleep_[F](d) ++ s`.
+    */
+  def delay[F[_], O](s: Stream[F, O], d: FiniteDuration)(implicit F: Timer[F]): Stream[F, O] =
```
SystemFw: I didn't really like this name in `Scheduler`, but in `Stream` is worse. I mean, it's perfectly clear what it does, and I can't think of an alternative, but it overlaps with `Sync.delay`....

core/shared/src/main/scala/fs2/async/Promise.scala
```diff
@@ -29,48 +25,37 @@ import Promise._
   *
   * Finally, the blocking mentioned above is semantic only, no actual threads are blocked by the implementation.
   */
-final class Promise[F[_], A] private[fs2] (ref: Ref[F, State[A]])(implicit F: Effect[F],
-                                                                  ec: ExecutionContext) {
+final class Promise[F[_], A] private[fs2] (ref: AtomicReference[State[A]])(
+    implicit F: Concurrent[F]) {
```
SystemFw: Since the ExecutionContext is not needed here anymore, can this change ripple up and we remove it pretty much everywhere? (not saying it should be done in this PR, but I want to know if it's possible).

I've also almost entirely convinced myself that removing it here is good, and that the callbacks don't have to be executed on the Promise ec, but I'd like to hear a few words motivating this change to have full peace of mind, if possible :)

core/shared/src/main/scala/fs2/async/async.scala
```diff
@@ -80,33 +78,32 @@ package object async {
...
 
   /**
     * Begins asynchronous evaluation of `f` when the returned `F[F[A]]` is
     * bound. The inner `F[A]` will block until the result is available.
     */
-  def start[F[_], A](f: F[A])(implicit F: Effect[F], ec: ExecutionContext): F[F[A]] =
-    promise[F, Either[Throwable, A]].flatMap { p =>
-      fork(f.attempt.flatMap(p.complete)).as(p.get.flatMap(F.fromEither))
-    }
+  def start[F[_], A](f: F[A])(implicit F: Concurrent[F], ec: ExecutionContext): F[F[A]] =
```
SystemFw: This will be eventually removed since we have it in cats effect, right?
Or do we want to keep a different version that `shifts`? and if so, why?

core/shared/src/main/scala/fs2/async/async.scala
```diff
@@ -134,44 +131,14 @@ package object async {
     * Begins asynchronous evaluation of `f` when the returned `F[Unit]` is
     * bound. Like `start` but is more efficient.
     */
-  def fork[F[_], A](f: F[A])(implicit F: Effect[F], ec: ExecutionContext): F[Unit] =
-    F.liftIO(F.runAsync(F.shift *> f) { _ =>
-      IO.unit
-    })
+  def fork[F[_], A](f: F[A])(implicit F: Concurrent[F], ec: ExecutionContext): F[Unit] =
+    F.start(Async.shift(ec) *> f).void
```
SystemFw: Any reason for the `shift`? No opinions, but curious if it was needed to make things work.
Ideally we could get rid of `EC`s in the api

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: Yeah agreed -- I think we could remove this method entirely.

core/shared/src/main/scala/fs2/async/Promise.scala
mpilquist: I tried heading down that path but wanted to get your opinion. Specifically, `Signal` still requires an `ec` in order to fork the completion of the promise (and hence, fork the callback invocations that are blocked on the promise's `get`).

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: Actually yeah, it could just be removed. The alternative with `sleep` is still super concise.

core/shared/src/main/scala/fs2/async/async.scala
mpilquist: Only to keep the old behavior -- unlike `fs2.async.start`, `Concurrent[F].start` does not introduce a thread shift.

core/shared/src/main/scala/fs2/async/Promise.scala
SystemFw: Right, in that case I'd say this deserves its own exploration/issue/PR.
Just wanted to make sure I wasn't missing something obvious that would make it a clear decision one way or another.

core/shared/src/main/scala/fs2/async/async.scala
mpilquist: It's needed for the implementation of `parallelTraverse` due to the thread shifting here. If we just used `Concurrent[F].start` in `parallelTraverse`, we won't get as much concurrency as folks expect. Perhaps we should deprecate `parallelTraverse` in favor of the new `Parallel` support in cats-effect though?

SystemFw: I think we should. I like that fs2 has its own concurrency scheme, but I'm the way I look at `parallelTraverse` here is just parallelism, in which case using `Parallel` is probably enough.
When you need more complicated concurrency, you should go with `join`.

So I'm slightly leaning toward removing both `start` and `parallel*` in favour of the ones in cats-effect. Not a strong opinion though.

core/shared/src/main/scala/fs2/async/async.scala
SystemFw: Ok, then I'd say it goes into the broader exploration of `ec`s and when and where we want to shift.

SystemFw:
I don't think any of the comments should hold merging into master btw, they can be addressed later

core/shared/src/main/scala/fs2/Stream.scala
aeons: An alternative name could be `delayFor`.

mpilquist: Oh and this should move to a `Stream` instance method too. `s.sleepFor(5.seconds)`?

SystemFw: Right, I think it could be a method on the Stream class, with that name.
`Stream(....).delayFor(3.seconds)`, or `delayBy`

mpilquist: :+1: for `delayFor` or `delayBy` as instance method on `Stream`

Issue #1107, opened 2018-03-14
StackOverflowException in core interpreter with stream shapes involving nested bind/uncons


mpilquist:
Reproduction case:

```scala
package example

import fs2._
import cats.effect.IO
import scala.concurrent.ExecutionContext.Implicits.global

object ObserveSoe extends App {
  val stream = Stream
    .range(0, 100000)
    .unchunk                    // Create lots of segments
    .covary[IO]
    .prefetch                   // Push segments through a queue
    .flatMap(_ => Stream.empty) // Generate a bind node that emits no elements
    .drain                      // Add a combinator that pulls a segment
  stream.compile.drain.unsafeRunSync
}
```

mpilquist:
Confirmed with some trace logging that the SOE is here: https://github.com/functional-streams-for-scala/fs2/blob/ed15bbc569ff81b01b276ec51ce1f00cfc6691aa/core/shared/src/main/scala/fs2/internal/Algebra.scala#L267-L270

SystemFw:
Thanks for minimising this, it was driving me insane :)

pchlupacek:
ok I did a bit of investigation there, and this is what I found : 

It seems to related only to queue, and particularly dequeue, which is defined as : 
```scala 
Stream
      .bracket(cancellableDequeue1)(d => Stream.eval(d._1), d => d._2)
      .repeat
```
now the issue seems to be related to bracket which is defined as 
```scala 
fromFreeC(Algebra.acquire[F, O, R](r, release).flatMap {
      case (r, token) =>
        use(r).onComplete { fromFreeC(Algebra.release(token)) }.get
    })

```
and particularly the `onComplete` which is defined as: 

```scala
(self.handleErrorWith(e => s2 ++ Stream.raiseError(e))) ++ s2
```
note tha since 0.10 the hadleError is defined as (note the Scope): 

```scala
 fromFreeC(Algebra.scope(self.get[F, O2]).handleErrorWith { e =>
        h(e).get[F, O2]
      })
```
as such I think it also uncovers the problem with queue, which when it takes one element only, 
it will in fact always open/close scope as result.

pchlupacek:
Just to clear things up, as it may not be obvious, every `bracket` now opens and closes scope, making the Acquire/Release sort of redundant from the scope now.

mpilquist:
[...]

PR #1108, opened 2018-03-19
Temporary fix for #1107 - removes scope from handleErrorWith, dramati…

mpilquist:
…cally reducing risk of SOE and memory leak caused by root bug in #1107


mpilquist:
Not merging this since the `observe` tests intermittently fail due to a change in error handling as a result of the scope removal.

PR #1110, opened 2018-03-20
Simplification of interpreter, fixing the SOE

AdamChlupacek:
@mpilquist @SystemFw 

Pavel have came up with simplified interpreter, that did multiple things for us: 

+ Fixed the SOE and memory leak for the #1107 .
+ Merged `StepLeg` and `Uncons` into one operation, which simplified translate for these cases.
+ Remove of `SetScope` from algebra.

I implemented largely simplified translate. 

The main idea of the new interpreter is that there is no more specialised interpretation of the `Algrebra` for when in normal stream, uncons stream or stepLeg stream. They all have been merged. The interpreter no longer needs to go to he lowest level when opening scope, but does it where it requires. The only exception for this is when creating an interruptible scope, which needs to be done on the `compileLoop` level, this is however optimised with passing the next step the interpreter should take. These things together will prevent the SOE.

### DIFF
--- a/core/jvm/src/main/scala/fs2/StreamApp.scala
+++ b/core/jvm/src/main/scala/fs2/StreamApp.scala
@@ -75,7 +75,6 @@ abstract class StreamApp[F[_]](implicit F: Effect[F]) {
             halted.set(true) *>
             exitCodePromise.complete(ExitCode.Error)
         case Right(exitCode) =>
-          println("XXXA returning code: " + exitCode)
           halted.set(true) *>
             exitCodePromise.complete(exitCode.getOrElse(ExitCode.Success))
       } *> exitCodePromise.get
--- a/core/jvm/src/main/scala/fs2/StreamApp.scala
+++ b/core/jvm/src/main/scala/fs2/StreamApp.scala
@@ -75,6 +75,7 @@ abstract class StreamApp[F[_]](implicit F: Effect[F]) {
             halted.set(true) *>
             exitCodePromise.complete(ExitCode.Error)
         case Right(exitCode) =>
+          println("XXXA returning code: " + exitCode)
           halted.set(true) *>
             exitCodePromise.complete(exitCode.getOrElse(ExitCode.Success))
       } *> exitCodePromise.get
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -8,7 +8,7 @@ import cats.{Applicative, Eq, Functor, Monoid, Semigroup, ~>}
 import cats.effect.{Effect, IO, Sync}
 import cats.implicits.{catsSyntaxEither => _, _}
 import fs2.async.Promise
-import fs2.internal.{Algebra, CompileScope, FreeC, Token}
+import fs2.internal.{Algebra, FreeC, Token}
 
 /**
   * A stream producing output of type `O` and which may evaluate `F`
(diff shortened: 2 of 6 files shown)


core/jvm/src/main/scala/fs2/StreamApp.scala
```diff
@@ -75,6 +75,7 @@ abstract class StreamApp[F[_]](implicit F: Effect[F]) {
             halted.set(true) *>
             exitCodePromise.complete(ExitCode.Error)
         case Right(exitCode) =>
+          println("XXXA returning code: " + exitCode)
```
mpilquist: Forgot to remove

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3177,13 +3177,15 @@ object Stream {
       * Note that resulting stream won't contain the `head` of this leg.
       */
     def stream: Stream[F, O] =
-      Stream.fromFreeC(Algebra.setScope(scope.id).flatMap { _ =>
-        next
-      })
+      Pull
```
mpilquist: Could you describe this change a little? I don't understand the need for looping here and it seems like we should still be consulting `next` at some point.

AdamChlupacek: The `next` is consulted by using the `leg.stepLeg` on each loop. This way we can step through to the end of the leg without the need for special algebra that will set the scope in the interpreter. There should not be much of overhead, if any, doing it this way instead of via the `SetScope` algebra.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -99,9 +92,33 @@ private[fs2] object Algebra {
...
+      .Eval[Algebra[F, X, ?], Option[(Segment[O, Unit], Token, FreeC[Algebra[F, O, ?], Unit])]](
+        Algebra.Step[F, O, X](stream, scopeId, chunkSize, maxSteps))
+
   def stepLeg[F[_], O](
       leg: Stream.StepLeg[F, O]): FreeC[Algebra[F, Nothing, ?], Option[Stream.StepLeg[F, O]]] =
-    FreeC.Eval[Algebra[F, Nothing, ?], Option[Stream.StepLeg[F, O]]](RunLeg(leg))
+    step[F, O, Nothing](
+      leg.next,
+      Some(leg.scopeId),
+      1024,
```
mpilquist: At some point, I'd like to surface these values in to the API or otherwise make them configurable but definitely don't need to do that now.

mpilquist: Although since they appear both here and in `compile`, maybe we should refactor this a bit so they are hard coded once for now?

mpilquist:
Looks great. The Travis failure is due to Mima. Could you add this to the mima config?

```
ProblemFilters.exclude[Problem]("fs2.internal.*")
```

core/shared/src/main/scala/fs2/internal/Algebra.scala
AdamChlupacek: We can just hard code these values now to places where they are required? That way it will be the easiest to refactor once we decide how to expose these values.

mpilquist: Yeah that sounds good

AdamChlupacek:
Can add, there is as well an issue with the `scopeId` in the `StepLeg`, I assume there has to be another rule to ignore that?

mpilquist:
Yeah a separate rule for ignoring the `StepLeg` change is fine (the change was to a `private[fs2]` val so no worries about compatibility)

Issue #1058, opened 2018-01-06
Optimize interruption of Eval


pchlupacek:
Currently when eval is interpreted in interruptible stream, every eval races with interruption signal to assure that it will be interrupted in case of the interruption signalled. 

However for synchronous `F`, this can be optimized, so the `F` that is known to not contain asynchronous code, may not be raced with interruption signal and rather checked before // after if the stream is interrupted. That will make overhead of interruption almost negligible. 

This requires https://github.com/typelevel/cats-effect/issues/104 to be resolved.

mpilquist:
I think we can defer this one to post 0.10 -- it wouldn't involved an API change so we can easily pick it up in 0.10.1 or something.

pchlupacek:
sure

mpilquist:
If cats-effect gets the requisite support in the next few days, then I'm fine with including this. I just don't want to hold 0.10 for it.

pchlupacek:
makes sense

mpilquist:
@pchlupacek Once #1149 is merged, this can be implemented via `Effect#runSyncStep`. Do you want to take a shot at implementing this PR?

pchlupacek:
yes. I will do this.

pchlupacek:
@mpilquist I just tried to implement it, but I am quite surprised that `runSyncStep` lives on Effect, which will require ConcurrentEffect on interrupt instead of Concurrent. I am sort of puzzled why this requires Effect?

What do you think is the course action we have to set ?

mpilquist:
That makes sense - Effect and ConcurrentEffect represent the ability to run an F. Not sure what course to take here.

pchlupacek:
Would the ConcurrentEffect be a problem as implicit to most Concurrent operators?

mpilquist:
Ehhh, I prefer only having a Concurrent bound. I’d like to hear from @systemfw and @alexandru

pchlupacek:
hm, sort of feel bad for @alexandru here. That feature is what we have requested to optimize eval, and now, we won't be able to use it :-(

alexandru:
The argument against `Effect` and `ConcurrentEffect` is that these restrictions are not very generic. At that point you're essentially working with `IO` or `Task` and this matters even for the monad transformers.

I don't mind having a `runSyncStep` in `Effect`. It will prove useful, as I've been using it with `Task` for some optimizations.

But yes, I would rather keep the `Concurrent` restriction instead of this optimization.

[...]

PR #1112, opened 2018-03-20
Topic/queue performance

### LINKED ISSUE #1111: Optimize Queue#dequeue
Current dequeue implementation uses `bracket` around each dequeue operation. That given current bracket implementation involves 5 Algebra evaluations: OpenScope, AcquireResource, Eval, ReleaseResource and CloseScope to be executed for each dequeued elements. 

Rather we have to open and close scope only at start / end of dequeu stream and reduce the operation only to Eval() for each element.s

mpilquist:
Fixes #1111

### DIFF
--- a/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
+++ b/core/shared/src/main/scala/fs2/async/mutable/Queue.scala
@@ -8,7 +8,7 @@ import cats.Functor
 import cats.effect.Concurrent
 import cats.implicits._
 
-import fs2.internal.Canceled
+import fs2.internal.{Canceled, Token}
 
 /**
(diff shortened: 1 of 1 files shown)


mpilquist:
In some ad-hoc performance testing done with `Stream.range(0, 1000000).unchunk.covary[IO].prefetchN(..)`, these changes don't make an improvement as the bottleneck is unrelated to raw interpreter speed.

pchlupacek:
@mpilquist its quite interesting. Also I think we have to do #1058 as each eval is quite complex computation now due to interruption.

pchlupacek:
Also perhaps we have to backport this to 0.10.x?

PR #1113, opened 2018-03-21
Improved performance of Queue

mpilquist:
@pchlupacek Here's the backport I did yesterday. I didn't open an PR for it since I wanted to make sure you were okay with the changes to 1.0 first.


mpilquist:
Not sure what to do about the mima issues -- they are real API breakages but we don't really expect folks to implement their own subtypes of `Queue`. On the other hand, there was some discussion in Gitter yesterday of doing that very thing.

pchlupacek:
Ok, I see your point. Hm perhaps the we should only make it as 1.0 feature? I mean if you anyhow measured that performance is not that bad?

mpilquist:
I'm fine with 1.0 only. Closing for now.

PR #1114, opened 2018-03-21
Removed new scope in bracket + optimised eval_

AdamChlupacek:
Since we started to look a bit to performance, I noticed these two things:

As of now we created new scope for each bracket, which is a minor issue in terms of performance, but since we already have the Aquire/Release algebra, why not to use it properly right? 

Second, thing that bugged me was that `eval_` was implemented via `eval(fa).drain` which adds an unnecessary `Output` and `Step` algebra. 

Running a stream that did 10000 `eval_` seen about 25% increase in performance.


mpilquist:
Awesome!

PR #1124, opened 2018-04-18
Upgrade to cats-effect 1.0.0-RC

mpilquist:
This isn't ready for merge yet. I'll add definitions for `bracketCase` for `Stream` and `Pull` along with laws checking.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -1,5 +1,6 @@
 package fs2
 
+import cats.~>
 import cats.effect.{ExitCase, Sync}
 import fs2.internal.{Algebra, FreeC, Token}
 
@@ -67,6 +68,9 @@ final class Pull[+F[_], +O, +R] private (private val free: FreeC[Algebra[Nothing
 
   /** Applies the resource of this pull to `f` and returns the result in a new `Pull`. */
   def map[R2](f: R => R2): Pull[F, O, R2] = Pull.fromFreeC(get.map(f))
+
+  /** Applies the outputs of this pull to `f` and returns the result in a new `Pull`. */
+  def mapOutput[O2](f: O => O2): Pull[F, O2, R] = Pull.mapOutput(this)(f)
 
   /** Run `p2` after `this`, regardless of errors during `this`, then reraise any errors encountered during `this`. */
   def onComplete[F2[x] >: F[x], O2 >: O, R2 >: R](p2: => Pull[F2, O2, R2]): Pull[F2, O2, R2] =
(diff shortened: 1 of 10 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -729,10 +729,9 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
     * }}}
     */
   def map[O2](f: O => O2): Stream[F, O2] =
-    this.repeatPull(_.uncons.flatMap {
-      case None           => Pull.pure(None);
-      case Some((hd, tl)) => Pull.output(hd.map(f)).as(Some(tl))
-    })
+    mapSegments(_.map(f))
+  // TODO replace the above with the following once ResourceSafetySpec passes
```
mpilquist: @pchlupacek @AdamChlupacek Could you take a look at this? If I replace the definition of `map` with the commented out version, the `finalizers are run in LIFO order - scope closure` test fails due to the innermost resource not being released first but rather last. That is, the ordering of finalizers ends up 1, 2, 3, ..., 8, 9, 0 instead of the correct 0, 1, 2, 3, ... 8, 9. I added a bunch of logging and noticed the following difference: with the old definition of `map`, once the exception is thrown, the scope is closed, resulting in all 10 resources being released. With the new definition, once the exception is thrown, the 1st resource is released, then the 2nd, then the 3rd, and so on up to and including the 9th -- then the scope is closed, resulting in the outstanding resource 0 getting finalized. Hence, it appears we are somehow skipping some error handling?

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -68,6 +69,9 @@ final class Pull[+F[_], +O, +R] private (private val free: FreeC[Algebra[Nothing
   /** Applies the resource of this pull to `f` and returns the result in a new `Pull`. */
   def map[R2](f: R => R2): Pull[F, O, R2] = Pull.fromFreeC(get.map(f))
 
+  /** Applies the outputs of this pull to `f` and returns the result in a new `Pull`. */
+  def mapOutput[O2](f: O => O2): Pull[F, O2, R] = Pull.mapOutput(this)(f)
```
mpilquist: If we can't figure out why `ResourceSafetySpec` fails when `Stream#map` is implemented in terms of `mapOutput`, then we'll at least need to make this function private or move it to the test class that uses it.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -262,15 +262,15 @@ private[fs2] object Algebra {
           fx match {
             case output: Algebra.Output[F, X] =>
               interruptGuard(
-                F.pure(Out(output.values, scope, f(Right(()))))
+                F.pure(Out(output.values, scope, FreeC.Pure(()).transformWith(f)))
```
mpilquist: @pchlupacek @AdamChlupacek Can you double check this change? I think this is necessary to preserve any error handling provided by `f`. If we just call `f(Right(()))`, we end up discarding the root level error handler -- but the exception handler in `compileScope` calls `tail.asHandler` in the event that folding the segment results in an exception.

mpilquist:
Besides the caveats on `Pull#mapOutput` which needs investigation, this is ready for merge. I removed the `Sync` instance for `Stream`.

mpilquist:
@pchlupacek @AdamChlupacek Friendly ping on the open questions above. :)

core/shared/src/main/scala/fs2/internal/Algebra.scala
AdamChlupacek: Good catch, we should never call `f(x)` directly, we should always use transform with it. As to the `tail.asHandler`, that seems to be there only for the case of error coming from `g`.

core/shared/src/main/scala/fs2/Stream.scala
AdamChlupacek: Ok, So the issue here is that `Segment` is actually lazy. As such we do not know about the error until we force the segment. 

This is happening in `compileScope`. Where once we encounter error in the segment we actually use the `tail` as the error handling. Thus we skip the error handling for the `emit` of `0` and go  straight to the error handling of `1`. 

The difference between this and how it works with `mapSegments`, is the existence of `uncons` which is now defined using `Step`, as the `Step` has the tail of the `stream` in it, it is not consulted and the scope is directly closed thus we get the correct order of resources released. 

A possible fix is that we will add the error handler of the of the `Segment` to the `Out` and then inside of the `compileScope` we use this error handler instead of `tail.asHandler`.

mpilquist:
@AdamChlupacek Okay with merging this as-is then and then following up with a separate PR?

AdamChlupacek:
@mpilquist Yeah, sounds good to me.

PR #1125, opened 2018-04-20
fix behavior in access

jmcardon:
Fixes the behavior of `access` to provide "only once" modification.

I spoke with @SystemFw a while ago about this (I think a month ago) but I forgot about it.

### DIFF
--- a/core/shared/src/main/scala/fs2/async/Ref.scala
+++ b/core/shared/src/main/scala/fs2/async/Ref.scala
@@ -1,13 +1,12 @@
 package fs2.async
 
-import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
 
 import cats.{Eq, Show}
 import cats.effect.Sync
 import cats.implicits._
 
 import scala.annotation.tailrec
-
 import Ref._
 
 /**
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/async/Ref.scala
```diff
@@ -65,7 +64,15 @@ final class Ref[F[_], A] private[fs2] (private val ar: AtomicReference[A])(impli
     */
   def access: F[(A, A => F[Boolean])] = F.delay {
     val snapshot = ar.get
-    def setter = (a: A) => F.delay(ar.compareAndSet(snapshot, a))
+    val hasBeenCalled = new AtomicBoolean(false)
+    def setter =
+      (a: A) =>
+        F.delay {
+          if (hasBeenCalled.compareAndSet(false, true))
```
durban: I think you can use `getAndSet` here (it's supposedly faster).

jmcardon: Why would that be the case? Compare and swap is a single atomic instruction on some architectures, 
while `getAndSet` is essentially an optimized loop of `compareAndSwap` until it succeeds.

[This](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/sun/misc/Unsafe.java#1104) is the pure java implementation of CAS. However, it delegates to an intrinsic on openjdk/oracle jvm.

durban: Yes, both should be a single instruction (at least on openjdk and intel). Apparently, performance should be the same (https://stackoverflow.com/questions/5339769/relative-performance-of-swap-vs-compare-and-swap-locks-on-x86), I misremembered something. So yeah, in this case it doesn't seem to matter, sorry for the noise.

Issue #1128, opened 2018-04-24
Error handling from forcing a Segment


AdamChlupacek:
To carry on from the talk in the PR #1124 . 

The fact that we try to recover from errors within a `Segment.map` is doable for  single streams. But once we introduce any kind of concurrency (applies for deterministic merging as well) it all falls apart. The issue as I outlined before is that when the segment gets to the end of the stream we do not know from where it originated, as such in case of an exception while forcing a segment, we have no idea what errorHandler to call.

Lets assume this code: 

```scala
val s1: Stream[F, Int] = ???
val s2: Stream[F, Int] = ???
val s3: Stream[F, Int] = ???

Stream(
 s1.map(_ => throw new Throwable("Boom")).handleErrorWith(_ => Stream.empty), 
 s2.handleErrorWith(_ => Stream.emit(1)),
 s3
).joinUnbounded
```

Now we force the segments at the end of the stream. Where we already have no notion of which segment came from where. Which of error handler should be called to handle this issue? 

I can imagine we can figure out some tagging style that would allow us to distinguish between different elements for combinators defined inside of fs2, but once we allow users passing `Segment` around, we cannot track from where the segment came from.

As such I see we have two options: 

1) Once we encounter an error from forcing a segment, we close all scopes and terminate. 

2) We use the error handler that is closest to the point where we can track the `Segment`. As to demonstrate on a example: 

```scala
Stream(???)
.joinUnbounded
.handleErrorWith(err => Stream.emit(1)) // <- This handler would be called
```

The number 2 could lead to really hard to debug behaviour, as in some cases would the error handlers from the inner stream take effect and sometimes the error handler from outside of join would take effect.

AdamChlupacek:
I gave this a bit more thought and I think that number one is the only valid solution. It is the only one that will behave the same in all situations and thus the `Stream` will remain compossible and easy to reason about.

if we were to go with number 2 the behaviour of which handler would be call will depend on the whole structure of the `Stream`. Ie if you have multiple `.join`s where after each one you have the `handlerErrorWith` we would call only the last `handlerErrorWith` which would be completely different behaviour to when the Streams would be joined only once and then run separately.

Now another issue is that sometimes we force the `Segment` a bit earlier than the end of the program. ie `flatMap`, `unconsChunk` and such. This once again will make a mess out of the order of error handlers. Currently resources are correctly cleared in case of any kind of `uncons` so that part is fine.

[...]

PR #1132, opened 2018-04-30
Fixes the order of resource release if pure function throws.

AdamChlupacek:
A first PR that tackles #1128. 
This one ensures that if we encounter an error from forcing an segment we will close resources in the correct order. And puts back the new definition for `map`.

We still need to go through algebra and see if we are capturing errors from all places where they can be raised. Namely most of the `f(Right(r))` and `f(Left(err))` places. Most of them are suspended in `flatMap` of `F` but I think we should still try to capture them if we can.


AdamChlupacek:
The reason why travis is not passing is since our generation of failed stream we are generating a stream which has an exception thrown from within `Segment`s transformation. I am thinking we should remove such case and have special cased test for such behaviour. As it will behave differently now.

mpilquist:
Sounds good

mpilquist:
@AdamChlupacek Any luck?

AdamChlupacek:
@mpilquist sorry, I was really busy for past week :/ will try to at least get this PR into mergeable state as soon as possible.

AdamChlupacek:
@mpilquist I really have no time for at least 2 more weeks :/ For this fix to work it is mostly only needed to change the line 200 in Algebra. That will fix the order of the resources closing. The rest of the changes was just an attempt to catch exceptions earlier. If you have time I would be glad if you could bring this through the finish line.

mpilquist:
@AdamChlupacek Do we need this still? Since this PR, we've changed the implementation of `map` to the implementation in this PR and we did not change line 200 (now line 184) in Algebra and tests still pass.

AdamChlupacek:
@mpilquist Interesting, I will get the latest 1.0 see if I can create a test case that would require this fix, otherwise we can close this for sure if all tests are passing.

AdamChlupacek:
@mpilquist with getting rid of segment, this will not be required at all. I am now closing this.

PR #1133, opened 2018-04-30
add mapAsync and mapAsyncUnordered

derekjw:
Should help aid people migrating from akka-streams to fs2, these methods should provide the same behaviour as the akka-streams versions. If more tests are needed around handling different failure conditions I can work on adding those as well.


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1787,6 +1787,59 @@ object Stream {
       }.stream
     }
 
+    /**
+      * Like [[Stream#evalMap]], but will evaluate effects in parallel, emitting the results
+      * downstream in the same order as the input stream. The number of concurrent effects
+      * is limited by the `parallelism` parameter.
+      *
+      * See [[Stream#mapAsyncUnordered]] there is no requirement to retain the order of
```
SystemFw: missing an `if` here

derekjw:
Yeah, I only ran the tests in intellij. I'll get it fixed up once I get home.

derekjw:
I think I have it figured out, should be all good now

SystemFw:
I'll merge once the builds complete :)

Thanks!

derekjw:
Thanks!

PR #1136, opened 2018-05-04
(WIP): separate queue implementations

jmcardon:
Following the discussion [here](https://gitter.im/functional-streams-for-scala/fs2?at=5aecc32b6f9af87e0443254d):

Queue can be separated into `Queue` and `SignallingQueue` , the latter being essentially that's on `0.10`, and the former giving us a mechanism to create interop with other libraries (i.e conversant disruptor, JCTools) with `BlockingQueue`-like synchronous or asynchronous queue implementations for less-overhead than what's currently the status quo, which requires all queues to perform signalling work when it's not needed.

Stuff this PR can tackle:
- [x] Separate Queue into two impls.
- [x] Synchronous blocking queues made pure
- [ ] Asynchronous blocking queues based off of pure fs2.

Opened for discussion.

### DIFF
--- a/core/jvm/src/test/scala/fs2/async/QueueSpec.scala
+++ b/core/jvm/src/test/scala/fs2/async/QueueSpec.scala
@@ -80,7 +80,7 @@ class QueueSpec extends Fs2Spec {
       runLog(
         Stream
           .eval(async.unboundedQueue[IO, Int])
-          .flatMap(_.size.discrete)
+          .flatMap(_.sizeSignal.discrete)
           .take(1)) shouldBe Vector(0)
     }
     "peek1" in {
@@ -161,7 +161,7 @@ class QueueSpec extends Fs2Spec {
       runLog(
         Stream.eval(
           for {
-            q <- async.mutable.Queue.synchronousNoneTerminated[IO, Int]
+            q <- async.mutable.SignallingQueue.synchronousNoneTerminated[IO, Int]
             f <- async.shiftStart(q.peek1)
             g <- async.shiftStart(q.peek1)
             _ <- q.enqueue1(None)
--- a/core/shared/src/main/scala/fs2/Pipe.scala
+++ b/core/shared/src/main/scala/fs2/Pipe.scala
@@ -3,7 +3,7 @@ package fs2
 import scala.concurrent.ExecutionContext
 import cats.effect.Concurrent
 import cats.implicits._
-import fs2.async.mutable.Queue
+import fs2.async.mutable.SignallingQueue
 import fs2.internal.{FreeC, NonFatal}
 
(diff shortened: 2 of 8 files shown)


pchlupacek:
I noticed that you have changed to have thread blocking operations on the synchronous queues and I am 👎 on that. I don't think so we shall have any F.delay { blocking_thread } code in library. 

I would like though to explore ways how to make queueing algorithm configurable in queue, given we will leave all queue behaviour as it is now.

jmcardon:
Fair, I can remove that.

jmcardon:
That said @pchlupacek , a lot of the code that goes on in queue current queue implementation isn't too different from "fork and spinlock", such as [this](https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/core/shared/src/main/scala/fs2/async/Promise.scala#L50) and [this](https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/core/shared/src/main/scala/fs2/async/mutable/Queue.scala#L196). It shouldnt be much different if I `Async.shift(ec) >> F.delay(block_thread)` from what I'm looking @ in the other implementations. 
I think we could explore different waiting strategies similar to the ones provided by [lmax disruptor](https://github.com/LMAX-Exchange/disruptor).

pchlupacek:
@jmcardon any thread blocking code in fs2 is breaking one of the main principle and that is that we can run in fixed number of threads. If the third party lib requires blocking the thread it shall have its own EC and F.shit shall `shift` us between boundaries of `fs2` EC and that library `EC`.  This mean that shall not be part of core fs2. 

Another problem with using blocking_thread approach in fs2-core, is that I am not sure if the `blocking_thread` code will compile on scala.js platform, where all core classes are available and supported, including the Queue and all derived primitives from it. 

I think spinlock strategy shall incur much less cost compare thread synchronisation, even under decent thread contention.

jmcardon:
@changlinli I am totally ok with not having basic blocking queues.

I think the main benefit in this PR is also for implementers downstream to have a weaker `Queue[F, A]` contract really. Maybe not necessarily part of fs2 directly.

jmcardon:
Also I apologize for the delay in this PR. Been finishing up some http4s Prs and then getting back to this one in a bit.

SystemFw:
Btw as discussed [here](https://gitter.im/typelevel/cats-effect?at=5af8a5f15a1d895fae387ff2) the waiting on Queue does **not** work like a fork and spin at all. 
The spinning is only there to modify the mutable state, the waiting is done ultimately via `async`, which asynchronously suspends the computation until another `F` awakens it.

mpilquist:
@jmcardon Are you still planning on working on this PR?

PR #1137, opened 2018-05-13
WIP: Migrate to Ref and Deferred

### LINKED ISSUE #1135: Handle Ref/Promise migration
We are in the process of moving Ref and Promise to cats-effect, which means we'd need to update fs2 accordingly. I'll be doing this work since I've got some spare time this week.

SystemFw:
Fixes #1135 

This is obviously pervasive, and perhaps surprisingly, not as mechanical as I thought given the changes to Ref


core/shared/src/main/scala/fs2/async/async.scala
```diff
@@ -18,8 +19,8 @@ package object async {
     mutable.Signal(initialValue)
 
   /** Creates a `[[mutable.Semaphore]]`, initialized to the given count. */
-  def semaphore[F[_]: Concurrent](initialCount: Long): F[mutable.Semaphore[F]] =
-    mutable.Semaphore(initialCount)
+  def semaphore[F[_]: Concurrent](initialCount: Long): F[Semaphore[F]] =
```
mpilquist: Think we still need these constructors here?

SystemFw: I wasn't too sure, I kinda liked having them as they made things easier to migrate, and in general I like the fact they you can save some imports, e.g. call `refOf` without having to import `Ref`.
Not too hung up on it one way or another.

I was also thinking of making a small PR to cats-effect to rename `modifyAndReturn` to `modify` and `modify` to `modify_`, as you can see `modifyAndReturn` is needed pretty much everywhere, and `modify_` is consistent with `traverse_` and co (throw away the result). WDYT? (very minor, cosmetic issue)

mpilquist: I lean towards removing the aliases as I don't like introducing multiple ways to do the same thing. I think if folks are using `Ref` and `Deferred` and `Semaphore`, an extra import is okay.

No strong opinion on the rename to `modify_`.

mpilquist: BTW, if you want to do the rename, let's get that in before the M2 release, which I think is tomorrow.

SystemFw: Cool, I'll remove the aliases then :)

And I'll ask about the rename to see what people think, and if positive, I'll open a PR for that

SystemFw:
I'll wait for the RC2 to get the latest changes in

mpilquist:
:+1: How about adopting `Resource` as a replacement for singleton streams (when it makes sense)?

SystemFw:
Re: `Resource`, I have been entertaining the same idea. I'd probably do it as a separate PR though, because I have a hunch it will be more difficult than expected, mostly because once you get a Stream, you can go back to a Resource, so it might be quite pervasive.

Definitely :+1: to trying though :)

PR #1148, opened 2018-05-23
Add Sink.either combinator

Daenyth:
We added this at work and it's been useful in our codebase, plus I've
seen the exact need come up in gitter before as well.


core/shared/src/main/scala/fs2/Sink.scala
```diff
@@ -28,4 +30,19 @@ object Sink {
     * using the `Show` instance for the input type.
     */
   def showLinesStdOut[F[_]: Sync, I: Show]: Sink[F, I] = showLines(Console.out)
+
+  /**
+    * Sink that routes each element to one of two syncs based on Either's Left/Right
```
mpilquist: s/syncs/sinks/

Maybe reword the end of this a bit? Or just go with:

```scala
* Sink that routes each element to one of two sinks -- `Left(x)` values get routed to the `left` sink and `Right(x)` values get routed to the `right` sink.
```

Issue #1104, opened 2018-03-10
Evaluate need for ExecutionContext and thread shifting throughout API


mpilquist:
Now that we have `cats.effect.Concurrent`, our dependency on `ExecutionContext` can be reduced quite a bit. `Promise` no longer depends on an `ExecutionContext` at all. `Signal` depends on an `ec` only for completing an internal promise asynchronously.

We should analyze our current uses of `ExecutionContext` and see if there are any that should be removed.

Also, we should consider whether we really want `async.shiftStart` and `async.unsafeRunAsync` or whether these should be deprecated / moved to cats-effect.

SystemFw:
I'll be taking a closer look at this soon. However my view has shifted a bit and I think that something like `shiftStart` (renamed to `fork`) would be quite useful. The fact that `shifting` and `starting` are separated is imho a bit too confusing for beginners. In any case, I'll hopefully be able be more precise soon.

notxcain:
As discussed on gitter with @SystemFw, it's probably possible to replace `ExecutionContext` with `Timer[F]`

Daenyth:
Can this now be closed since #1149 is merged?

PR #1149, opened 2018-05-23
Removed requirement for ExecutionContext from API

mpilquist:
…now that Concurrent type class guarantees shifting in start and race.

The `fs2-io` package still requires an EC in places where it binds callbacks to Java NIO. We could instead replace the `EC` requirement with a `Timer[F]` since we really just want to shift off the NIO pool in such cases.


SystemFw:
I think this fixes #1104 :)

mpilquist:
I removed EC from fs2-io too, replacing it with `Timer[F]`. I also added `blocking(...)` calls around the `SyncVar` waits in the Java input/output stream integration, hopefully helping prevent deadlock. It would be really nice to get rid of that `SyncVar` and make the input/output stream integration safer.

SystemFw:
Just a question, with `Timer`, the "shift and shift back" pattern is harder than with `ec`. 
Are we changing the recommended strategy to `blocking` when dealing with blocking code? If not, `ec` in the api might actually be better....

mpilquist:
Hm, AFAIK we don't use that pattern in fs2-io. The only use of `ExecutionContext` was calling `unsafeRunAsync` from java callbacks.

SystemFw:
Right, I didn't mean that we use it directly, but rather that our users might want to for the subset of  `fs2.io` that blocks, and it's easier to do so with `Ec` than with `Timer`.

SystemFw:
For non-blocking code :+1: :+1: to replace `ec` with `Timer` ofc. For blocking code that's a lingering question I think.

pchlupacek:
How about making the list of these `blocking` implementations (if we have any). Then I would like to see if we really need to have them, or if there is a way to make them async. If not, then I would like to require to pass some EC/Timer/... to clearly force user to know that this is going to drain threads, eventually.

pchlupacek:
just as comment, Is just about the naming, but I really do not like `Timer` implicit where there is nothing to do with time.

SystemFw:
@pchlupacek 
I think `Timer` should be renamed to `RTS`, but that's an issue for cats-effect > 1.0

I'd also like to list all blocking methods. We could even put them in a separate folder or something, if you can't remove them/improve them.

pchlupacek:
Usually in our code base we always create specific execution context for each blocking domain (for example Apache Tika) then we require implicit EC: ExecutionContext @@ ApacheTika. That way you can reasonably control the resource usage. 

I think for input stream we may really require EC/RTS/Timer to clearly requrie user to specify where blocking methods will be run and then just return shifted Async `F` from our API.

mpilquist:
Okay I pushed an update that restores the use of `ExecutionContext` for the functions that do blocking reads/writes with an `InputStream`/`OutputStream` and I made the EC argument explicit. I also ensured we do such operations inside a call to `blocking`, which will allow some execution contexts to adapt thread count dynamically if needed.

I think this PR is ready for merge unless there are more comments.

mpilquist:
@ChristopherDavenport Are you okay with the changes to io package w.r.t. http4s?

SystemFw:
I'll merge as soon as Chris responds, this is great :)

SystemFw:
Can we now say that this statement holds?

> If an fs2 function takes an `EC`, it's blocking. There's no blocking in `core` .

mpilquist:
> There's no blocking in core .

Yes this is basically true. Only exception I can think of is `Sink.{lines, showLines, showLinesStdOut}` which technically blocks writing to the `PrintStream`.

> If an fs2 function takes an EC, it's blocking.

Yep this is true too, though I might reword to "it's doing something that blocks" to clarify that the returned value is not necessarily the thing doing the blocking directly.

Another caveat is that things like `fs2.io.files.readAll(..)` will block when reading from the file system. For each of these blocking file apis, we have an async version which shifts the blocking to a dedicated EC, but I don't think we'd want to remove them entirely.

mpilquist:
Oh one caveat about `Timer[F]` vs `RTS[F]` for some future cats-effect version -- I want the ability to shift & sleep without having the ability to call `unsafeRunAsync`. Hence I think we'd need to separate those functions regardless of names.

SystemFw:
100% agreed on your first comment.

As for the RTS thing, I envisioned something like a pure RTS (basically Timer, and _perhaps_ something to deal with blocking), which is then extended by the impure RTS, or something along those lines. 
But yeah absolutely they need to be separate.

io/src/main/scala/fs2/io/io.scala
```diff
@@ -28,19 +28,21 @@ package object io {
...
+  def readInputStreamAsync[F[_]](
+      fis: F[InputStream],
+      chunkSize: Int,
+      blockingExecutionContext: ExecutionContext,
+      closeAfterUse: Boolean = true)(implicit F: Async[F], timer: Timer[F]): Stream[F, Byte] =
+    readInputStreamGeneric(
+      fis,
+      F.delay(new Array[Byte](chunkSize)),
+      (is, buf) =>
+        Async.shift(blockingExecutionContext) *> readBytesFromInputStream(is, buf) <* timer.shift,
```
mpilquist: @SystemFw Should this shift/shift-back pattern be done via `bracket`?

SystemFw: I think it should actually, :+1:

SystemFw:
There was a weird build failure, which I don't think is related.

Am I ok to merge this?

mpilquist:
Yep

PR #1152, opened 2018-05-30
Added support for cats.effect.Resource

mpilquist:
Removed all singleton streams that existed solely for the purpose of bracketing with resources.

### DIFF
--- a/io/src/main/scala/fs2/io/tcp/Socket.scala
+++ b/io/src/main/scala/fs2/io/tcp/Socket.scala
@@ -140,9 +140,9 @@ protected[tcp] object Socket {
       implicit AG: AsynchronousChannelGroup,
       F: ConcurrentEffect[F],
       timer: Timer[F]
-  ): Stream[F, Either[InetSocketAddress, Stream[F, Socket[F]]]] = {
+  ): Stream[F, Either[InetSocketAddress, Resource[F, Socket[F]]]] = {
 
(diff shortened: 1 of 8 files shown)


io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -147,67 +140,63 @@ protected[tcp] object Socket {
...
-        ch.setOption[java.lang.Boolean](StandardSocketOptions.SO_REUSEADDR, reuseAddress)
-        ch.setOption[Integer](StandardSocketOptions.SO_RCVBUF, receiveBufferSize)
-        ch.bind(address)
-        ch
-      }
-
-      def cleanup(sch: AsynchronousServerSocketChannel): F[Unit] = F.delay {
-        if (sch.isOpen) sch.close()
-      }
+  ): Stream[F, Either[InetSocketAddress, Stream[F, Socket[F]]]] = {
```
SystemFw: I'm not sure whether this should be `Stream[F, Either[InetSocketAddress, Resource[F, Socket[F]]]]` or not

mpilquist: Yeah good point, it should

SystemFw:
I think we might need some more scaladoc around this, warning that:

`Resource -> F.bracket -> evalMap` is not the same as `Resource -> Stream.bracket -> flatMap` so to speak, and that the second is what you want

Issue #1154, opened 2018-06-01
Converting from Pull to Stream should introduce a scope


mpilquist:
Consider this example:

```scala
val s = Stream(1, 2, 3).onFinalize(IO(println("finalized")))
val prg = s.take(1).repeat.take(10).append(Stream.eval(IO(Thread.sleep(5000))))
```

When `prg` is run, the finalizers are all delayed until the overall stream terminates. The root cause of this is `take(1)`, which invokes uncons once and then discards the tail stream. Because the tail stream is discarded, the finalizer for the resource associated with the `onFinalize` call is never explicitly released.

Manually inserting a scope after `take(1)` fixes this issue. Specifically:

```
val prg2 = s.take(1).scope.repeat.take(10).append(Stream.eval(IO(Thread.sleep(5000))))
```

In 0.9, *all* pull-to-scope conversions inserted a scope. We removed this in 0.10 b/c we thought it was unnecessary but in hindsight, I believe this was a mistake. I suspect this is the root of many of the issues exposed in #1151.

SystemFw:
I can't remember if there were any other reasons for not having a scope after every Pull-to-Stream apart from thinking it wasn't necessary

mpilquist:
I added `streamNoScope` back in October: https://github.com/functional-streams-for-scala/fs2/commit/a3019678b086b5554ede9f331515c54b8ee47285

This bugged me at the time, and I really liked the idea of zero cost from going between `Stream` and `Pull` -- just two different monads for the same underlying data structure.

I then removed it in January: https://github.com/functional-streams-for-scala/fs2/commit/47e41dbbea554fa391ff51d48dfd926e8c0c3822

mpilquist:
Fixed by #1155

PR #1155, opened 2018-06-01
Fixes #1151 #1154

### LINKED ISSUE #1151: Memory leak with 'flatMap', 'zip', 'fixedDelay'...
After spending a day at work doing various experiments to fix a memory leak in the long-lived stream that makes up one of my applications, I think streams shaped around some combinations of `flatMap`, `zip`, `joinUnbounded` and `Scheduler.fixedDelay` create an accumulation of FS2 objects.

I have observed the problem in both versions 0.10.2 and 0.10.4.

My stream's shape is pretty complex, and I can disclose some of it if really needed, but I have a small example that shows the problem. It probably deserves some explanation to map it with my real case.

The core logic is that I have an application that handles so-called processes, each of which consists in picking files from a dedicated directory, reading each of them and sending their contents as a message in a Kafka topic. To support a dynamic set of processes, at each 'heartbeat' (every second for example), I read a set of process definitions that is fed from an external administration web service, and for each of them, create a stream that consists in doing all that. Those streams are then joined. Those sub-streams are not long-lived; they end as soon as one file is processed, so that the joined stream ends as soon as possible, and another heartbeat can take place, with possibly a different set of processes, or with different configurations. In particular, this set can be empty.

```
import cats.effect.IO
import fs2.{Scheduler, Stream}

object SandboxApp extends App {
  // The inner stream embodies the set of processes, always empty in this example
  private val sources: Stream[IO, Stream[IO, Int]] = Stream(Stream.empty.covaryAll).repeat

  Scheduler[IO](8)
    .flatMap { sched =>
      sched.fixedDelay[IO](10.milliseconds) // At each heartbeat...
        .zip(sources) // ... obtain the set of 'processes'
        .flatMap { case (_, s) =>
          s.map(Stream.constant(_).covary) // ... create a stream that does something with each (irrelevant here)...
           .joinUnbounded // ... and join all processes
        }
    }
    .compile
    .drain
    .unsafeRunSync()
}
```

In the example, the set of processes is always empty, and the infinite stream created with `Stream.constant` is never executed, but the memory leak is already present at this point.

I initially thought `joinUnbounded` was the only culprit because when replacing `s.map(Stream.constant(_).covary).joinUnbounded` with `s.flatMap(Stream.constant(_).covary)`, the problem is gone in this example. However, in my real stream it is still there, though more subtle. So maybe there are several issues after all.

Here are the memory usage charts for my real application: https://imgur.com/a/h0brZbS, as well as a link to the complete snapshots that you can open in JVisualVM to check the heap dumps I took and compare them. The link is only valid until June 5th: https://www.transfernow.net/314kp1x23v2q. Before each heap dump, I forced a GC.

All of my tests are done with `-Xms64m -Xmx64m`, and the data I provide here is for the tests done with version 0.10.4. Note also that in the real cases, the heartbeat frequency is a hundred times lower than in the example.

I have tried to change a bit my stream's shape hoping to find a workaround, but to no avail. Sometimes I made it even worse, though I haven't kept track of all my attempts. This is probably subjective, but sometimes I found the problem to be a bit less serious in version 0.10.2.

This is a serious issue for me and I hope you guys will be able to at least find a workaround.

Aside from that, great library ;)!

mpilquist:
I confirmed this fixes the memory leak described in #1151.


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -872,11 +872,12 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
   /**
-    * Tracks any resources acquired during this stream and releases them when the stream completes.
+    * Scopes are typically inserted automatically, at the boundary of a pull (i.e., when a pull
+    * is converted to a stream). This method allows a scope to be explicitly demarcated so that
+    * resources can be freed earlier than when using automatically inserted scopes.
     *
-    * Scopes are sometimes inserted automatically, (e.g., as a result of calling `handleErrorWith`).
-    * This method allows a scope to be explicitly demarcated, so that resources can be freed earlier
-    * than when using automatically inserted scopes.
+    * One use case is scoping the left hand side of an append: `(s1.scope ++ s2)`, which ensures
```
SystemFw: A bit confused if this is the case now or not.
If I don't insert `scope` explicitly, are resources in `s1 ++ s2` released at the end of `s1` or not?

mpilquist: Yeah, the comment is pretty weak. If `s1` reaches a natural termination point, then it's resources are released before moving on to `s2`. However, if it doesn't run to completion, e.g. because of a pull that discarded the tail and used `streamNoScope`, then the explicit resource release call is lost and the resource lifecycle ends up promoted to the lifecycle of the active scope.

I don't know how to document this but suggestions are welcome. :)

SystemFw: > However, if it doesn't run to completion, e.g. because of a pull that discarded the tail and used streamNoScope, then the explicit resource release call is lost

That's only true if `streamNoScope` is used though, right? (and it's true currently, which is the source of the leaks?)
Or can it also happen to a Stream that didn't run to completion but didn't use `streamNoScope`?

SystemFw: I'm being picky because:
- That's one of the few comments addressing `Scopes`, and it's confusing :) If we can't make it clearer (and I agree that it's hard to document), we might be better off removing it
- Given the findings in #1151 , perhaps `streamNoScope` should come with a "only use it if you know what you're doing" warning

mpilquist: Agreed on both counts. Will push an update now.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -872,11 +872,12 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    * resources acquired during `s1` are released onces the end of `s1` has been passed.
     */
   def scope: Stream[F, O] = Stream.fromFreeC(Algebra.scope(get))
```
SystemFw: Tangential question: what would be a use case for `scope`, now they're back to being inserted automatically on `Pull-to-Stream`?

mpilquist: I think my commit just now addresses this question in the updated ScalaDoc.

SystemFw: Right, that's roughly what I thought: basically `streamNoScope` and `scope` let you do manual resource management instead of relying on fs2's own mechanism.
I'm ok with this as you never know what people might come up with, and the comment on `streamNoScope` is a sufficient warning :)

PR #1157, opened 2018-06-03
(Proposal/WIP) Adding `groupedTimed`/`groupedWithin`

changlinli:
This is a WIP proposal based on some gitter comments/requests for a method that is similar to akka-streams' `groupedWithin`.

Note that this is WIP, both in concept and implementation (and in particular does not yet pass the last test I added).


SystemFw:
Haven't looked at this code yet, but I have a wip PR as well, we can
compare and contrast

ahjohannessen:
Would be nice if output was NonEmptyList

PR #1158, opened 2018-06-03
Add `groupWithin`

SystemFw:
This is a prototype for `groupWithin`, same as #1157 

Still missing tests and docs, but worth submitting since it uses a different approach.

Looking for feedback from @Daenyth and @changlinli

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4,7 +4,6 @@ import cats._
 import cats.data.NonEmptyList
 import cats.effect._
 import cats.effect.concurrent._
-import cats.effect.implicits._
 import cats.implicits.{catsSyntaxEither => _, _}
 import fs2.internal.FreeC.Result
 import fs2.internal._
@@ -962,13 +961,16 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     *
     * Note: a time window starts each time downstream pulls.
     */
-  def groupWithin[F2[x] >: F[x]: Concurrent](n: Int, d: FiniteDuration)(
-      implicit timer: Timer[F2]): Stream[F2, Chunk[O]] =
-    Stream.eval(async.boundedQueue[F2, Option[Either[Int, O]]](n)).flatMap { q =>
-      def startTimeout(tick: Int): Stream[F2, Nothing] =
+  def groupWithin[F2[x] >: F[x]](n: Int, d: FiniteDuration)(
+      implicit timer: Timer[F2],
+      F: Concurrent[F2]): Stream[F2, Chunk[O]] =
+    Stream.eval(async.boundedQueue[F2, Option[Either[Token, O]]](n)).flatMap { q =>
+      def startTimeout: Stream[F2, Token] =
         Stream.eval {
-          (timer.sleep(d) *> q.enqueue1(tick.asLeft.some)).start
-        }.drain
+          F.delay(new Token).flatTap { t =>
+            F.start { timer.sleep(d) *> q.enqueue1(t.asLeft.some) }
+          }
+        }
 
       def producer = this.map(_.asRight.some).to(q.enqueue).onFinalize(q.enqueue1(None))
 
@@ -976,27 +978,29 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
         if (c.nonEmpty) Stream.emit(Chunk.concat(c.toList))
         else Stream.empty
 
-      def go(acc: Catenable[Chunk[O]], elems: Int, currentTimeout: Int): Stream[F2, Chunk[O]] =
+      def go(acc: Catenable[Chunk[O]], elems: Int, currentTimeout: Token): Stream[F2, Chunk[O]] =
         Stream.eval(q.dequeue1).flatMap {
           case None => emitNonEmpty(acc)
           case Some(e) =>
             e match {
               case Left(t) if t == currentTimeout =>
-                emitNonEmpty(acc) ++ startTimeout(currentTimeout + 1) ++ go(Catenable.empty,
-                                                                            0,
-                                                                            currentTimeout + 1)
+                emitNonEmpty(acc) ++ startTimeout.flatMap { newTimeout =>
+                  go(Catenable.empty, 0, newTimeout)
+                }
+
               case Left(t) => go(acc, elems, currentTimeout)
               case Right(a) if elems >= n =>
-                emitNonEmpty(acc) ++ startTimeout(currentTimeout + 1) ++ go(
-                  Catenable.singleton(Chunk.singleton(a)),
-                  1,
-                  currentTimeout + 1)
+                emitNonEmpty(acc) ++ startTimeout.flatMap { newTimeout =>
+                  go(Catenable.singleton(Chunk.singleton(a)), 1, newTimeout)
+                }
               case Right(a) if elems < n =>
                 go(acc.snoc(Chunk.singleton(a)), elems + 1, currentTimeout)
             }
         }
 
-      startTimeout(0) ++ go(Catenable.empty, 0, 0).concurrently(producer)
+      startTimeout.flatMap { t =>
+        go(Catenable.empty, 0, t).concurrently(producer)
+      }
     }
 
   /**
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -964,52 +964,54 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
   def groupWithin[F2[x] >: F[x]](n: Int, d: FiniteDuration)(
       implicit timer: Timer[F2],
       F: Concurrent[F2]): Stream[F2, Chunk[O]] =
-    Stream.eval(async.boundedQueue[F2, Option[Either[Token, Chunk[O]]]](n)).flatMap { q =>
-      def startTimeout: Stream[F2, Token] =
-        Stream.eval {
-          F.delay(new Token).flatTap { t =>
-            F.start { timer.sleep(d) *> q.enqueue1(t.asLeft.some) }
-          }
-        }
-
-      def producer = this.chunks.map(_.asRight.some).to(q.enqueue).onFinalize(q.enqueue1(None))
-      def emitNonEmpty(c: Catenable[Chunk[O]]): Stream[F2, Chunk[O]] =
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1899,6 +1899,54 @@ object Stream {
...
+
+      Stream.eval(async.unboundedQueue[F, Option[Either[Tick, O]]]).flatMap { q =>
+        def startTimeout(i: Tick): Stream[F, Nothing] =
+          Stream.eval {
+            async.fork {
+              timer.sleep(d) *> q.enqueue1(i.asLeft.some)
+            }
+          }.drain
+
+        def producer = self.map(Right(_)).noneTerminate.to(q.enqueue)
```
Daenyth: Where is `self` coming from?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1899,6 +1899,54 @@ object Stream {
...
+          else Stream.empty
+
+        def go(acc: Catenable[Segment[O, Unit]],
+               elems: Int,
+               currentTimeout: Tick): Stream[F, Segment[O, Unit]] =
+          Stream.eval(q.dequeue1).flatMap {
+            case None => emitNonEmpty(acc)
+            case Some(e) =>
+              e match {
+                case Left(t) if t > currentTimeout => throw new Exception
```
Daenyth: Do you want to throw instead of `Stream.raiseError`?

What does this case mean semantically? How is this branch hit?

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: `InvariantOps`, it's `this` plus the covariance machinery

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: that's a wip line. I want an `error` eventually because that branch can't be hit I don't think. Sloppy writing to get this in tonight :)

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1899,6 +1899,54 @@ object Stream {
     def foldMonoid(implicit O: Monoid[O]): Stream[F, O] =
       self.fold(O.empty)(O.combine)
 
+    def groupWithin(n: Int, d: FiniteDuration)(
+        implicit F: Effect[F],
+        timer: Timer[F],
+        ec: ExecutionContext): Stream[F, Segment[O, Unit]] = {
+      type Tick = Int
+
+      Stream.eval(async.unboundedQueue[F, Option[Either[Tick, O]]]).flatMap { q =>
```
changlinli: If I understand this correctly, this `unboundedQueue`, coupled with your `concurrently` producer later on I think leads to unbounded space usage. You might be able to recover some of your bounded behavior by using something like `synchronousNoneTerminated` or `bounded`, but I'm not sure. I think you run into problems with interrupting a pull in those cases. In particular if you have a really fast upstream stream, I'm not sure whether it's possible for `startTimeout`s `enqueue1` attempt to consistently get beat out.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1899,6 +1899,54 @@ object Stream {
...
+                  emitNonEmpty(acc) ++ startTimeout(currentTimeout + 1) ++ go(
+                    Catenable.singleton(Segment.singleton(a)),
+                    0,
+                    currentTimeout + 1)
+                case Right(a) if elems < n =>
+                  go(acc.snoc(Segment.singleton(a)), elems + 1, currentTimeout)
+              }
+          }
+
+        startTimeout(0) ++ go(Catenable.empty, 0, 0).concurrently(producer)
```
changlinli: `concurrently` feels like it's the wrong combinator here. My intuition for how this method should work is that ideally you don't pull until your timeout starts, whereas `concurrently` will always have you pulling in the background.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1899,6 +1899,54 @@ object Stream {
     def foldMonoid(implicit O: Monoid[O]): Stream[F, O] =
       self.fold(O.empty)(O.combine)
 
+    def groupWithin(n: Int, d: FiniteDuration)(
+        implicit F: Effect[F],
+        timer: Timer[F],
+        ec: ExecutionContext): Stream[F, Segment[O, Unit]] = {
```
changlinli: In general given that `Segment`s can be potentially infinite while the `Segment`s generated by this can only be finite (which is very valuable for downstream consumers to know), it seems like you'd probably either want to restrict the type to a `Chunk` or potentially a non-`Segment` collection (e.g. `List`).

SystemFw:
@changlinli So, generally speaking I think having most logic in `dequeue`, rather than triggering dequeues on certain conditions (most login in `enqueue`, loosely speaking) is key to be able to do this without requiring too much concurrent synchronisation logic.

However I do hear your concern about a slow consumer, I'd like to understand the problem a bit better (example perhaps?). I also wonder if a `bounded` queue might solve it without adding too much complication, and I'm looking at `merge` vs `concurrently` as well. Thanks for the feedback :)

core/shared/src/main/scala/fs2/Stream.scala
Daenyth: I disagree here; I think it's fine to expose segment with a guarantee in the docs that these segments will be non-empty and finite.

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: `noneTerminate` here is not enough, we need `onFinalize(enqueue(None))`. 
If `producer` is interrupted, `concurrently` won't interrupt the consumer (since producer is on the right side), and `None` won't be enqueued. In fact we might need to introduce interruption, which is probably another sign that `concurrently` is not the right choice here, as per @changlinli 's comment below

core/shared/src/main/scala/fs2/Stream.scala
changlinli: Eh... I don't think `Segment` buys you much at all here. Given most use cases I think of for this method, you end up working with the entire batch at once, i.e. you'll have some function of `Segment[O, Unit] => X` which almost certainly means you're going to call `.force.toList` or something similar on it immediately. I can only see `Segment` being useful if you `Pull.output` some a transformed version of it back into a new `Stream`, but if you're going to do that (e.g. with `map` or the like, you could probably have just done that on the parent `Stream` directly. If you just started out with a `List` (or `Vector`), you'll probably have better perf and save a conversion since far far more APIs are going to take a `List` than a `Segment`.

SystemFw:
Getting close to finishing this. I've used @changlinli tests (which caught a bug :) )
The akka stream suite is quite thorough : https://github.com/akka/akka/blob/master/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGroupedWithinSpec.scala#L24-L175

But all their tests rely on the push-based reactive spec, so a little more work is required to port what's possible to port

changlinli:
Sorry I've been completely away from OSS stuff for the past week. I'd be happy to try to port some of the akka stream tests here to help out.

changlinli:
It still feels somewhat weird that potentially you end up buffering up to `2n` elements for a parameter `n` (or equivalently that an element might have to wait up to twice the specified duration from being consumed to being re-emitted) because of the `n`-sized `boundedQueue`.

SystemFw:
> I'd be happy to try to port some of the akka stream tests here to help out.

Feel free to make a PR to this PR (you need to fork my branch and open a PR there)

> you end up buffering up to 2n elements for a parameter n

Why `2n`?

mpilquist:
@SystemFw Do you still have a WIP of this combinator on a branch somewhere?

SystemFw:
I do, sorry for the delay.

I think we're still a bit stuck between my initial implementation and the @changlinli's one.

You can have a look here

https://github.com/SystemFw/fs2/pull/1

SystemFw:
Ok did some work to get this done:
- I'm now targeting 1.0
- The combinator works at the Chunk level rather than individual element, making the `Option[Either` wrapping not too onerous
- The key realisation to make this work is that you can never get a timeout mid-chunk, or in other words you can never receive a single element faster than you can receive the chunk that element is in, since everything is expressed in terms of `unconsChunk` (including `uncons1`). If this assumption is not correct, we need to go back to working with single elements I'm afraid
- Use of `Token` instead of `Int` to track expired timeouts, to avoid overflows

All that being said, please review the logic very carefully, because this has proven tricky to write.
Also, we do have some tests but I'm fairly sure they wouldn't catch many possible bugs here. It is however very hard to write an exhaustive manual test for this, ideas welcome.

SystemFw:
Closed and reopened to trigger travis

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -966,6 +966,66 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+  def groupWithin[F2[x] >: F[x]](n: Int, d: FiniteDuration)(
+      implicit timer: Timer[F2],
+      F: Concurrent[F2]): Stream[F2, Chunk[O]] =
+    Stream
+      .eval(async.mutable.Queue.synchronousNoneTerminated[F2, Either[Token, Chunk[O]]])
+      .flatMap { q =>
+        def startTimeout: Stream[F2, Token] =
+          Stream.eval {
+            F.delay(new Token).flatTap { t =>
+              F.start { timer.sleep(d) *> q.enqueue1(t.asLeft.some) }
```
mpilquist: Might not be necessary, especially for small `d`, but we might want to cancel this fiber on stream cancelation.

SystemFw: good point, I can probably do this now?

```scala
Stream.eval(F.delay(new Token)).flatTap(t => Stream.supervise(timer.sleep(d) *> q.enqueue1(...)).void
```

mpilquist: Yep, that looks like it would do it.

mpilquist: I feel better about having `supervise` now 😀

SystemFw: :)

I was actually missing something that `start`ed at the stream level (even though I wasn't coming from the pov of finalising `cancel`), so that's good to have

mpilquist:
Looks good to me -- ready for merge after the fiber cancellation fix.

SystemFw:
Ok, I'm going to merge this if there are no objections

mpilquist:
:+1:

PR #1159, opened 2018-06-08
Optimized Chunk.seq/Chunk.apply to avoid creation of Vectors

mpilquist:
Noticed by @jkachmar.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -338,7 +338,41 @@ object Chunk {
       }
 
   /** Creates a chunk backed by a `Seq`. */
-  def seq[O](s: Seq[O]): Chunk[O] = vector(s.toVector)
+  def seq[O](s: Seq[O]): Chunk[O] = s match {
+    case a: collection.mutable.WrappedArray[O] => array(a.array)
+    case v: Vector[O]                          => vector(v)
+    case ix: IndexedSeq[O]                     => indexedSeq(ix)
+    case _                                     => buffer(collection.mutable.Buffer(s: _*))
+  }
+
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -338,7 +338,41 @@ object Chunk {
       }
 
   /** Creates a chunk backed by a `Seq`. */
-  def seq[O](s: Seq[O]): Chunk[O] = vector(s.toVector)
+  def seq[O](s: Seq[O]): Chunk[O] = s match {
+    case a: collection.mutable.WrappedArray[O] => array(a.array)
+    case v: Vector[O]                          => vector(v)
+    case ix: IndexedSeq[O]                     => indexedSeq(ix)
+    case _                                     => buffer(collection.mutable.Buffer(s: _*))
```
mpilquist: Note we can't convert the seq to an array here b/c we don't have a `ClassTag[O]`.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -338,7 +338,41 @@ object Chunk {
       }
 
   /** Creates a chunk backed by a `Seq`. */
-  def seq[O](s: Seq[O]): Chunk[O] = vector(s.toVector)
+  def seq[O](s: Seq[O]): Chunk[O] = s match {
+    case a: collection.mutable.WrappedArray[O] => array(a.array)
```
mpilquist: This is the type you get when passing varargs, so things like `Chunk(1, 2, 3)` will be *significantly* faster with this change -- before, the 3-element array was promoted to a `Vector` which is slow and wasteful.

Issue #1160, opened 2018-06-14
Stream#map destroys chunking (sometimes?)


djspiewak:
Using `map`:

```scala
    val sliceStream =
      s
        .mapChunks({ chunk => println(s"before RValue> ${chunk.size}"); chunk.toSegment })
        .map(data => RValue.fromJValueRaw(JValue.fromData(data))
        .mapChunks({ chunk => println(s"after RValue> ${chunk.size}"); chunk.toSegment })
```

Using `mapChunks`:

```scala
    val sliceStream =
      s
        .mapChunks({ chunk => println(s"before RValue> ${chunk.size}"); chunk.toSegment })
        .mapChunks(chunk => chunk.map(data => RValue.fromJValueRaw(JValue.fromData(data))).toSegment)
        .mapChunks({ chunk => println(s"after RValue> ${chunk.size}"); chunk.toSegment })
```

There's some incidental complexity in this example because I haven't actually minimized it, but you get the idea. Here are some println excerpts: 

```
before RValue> 93
after RValue> 1
after RValue> 1
after RValue> 1
after RValue> 1
after RValue> 1
after RValue> 1
after RValue> 1
after RValue> 1
after RValue> 1
after RValue> 1
after RValue> 1
```

And with `mapChunks`:

```
before RValue> 93
after RValue> 93
in yield> 94
before RValue> 94
after RValue> 94
^Cin yield> 93
before RValue> 93
after RValue> 93
in yield> 93
before RValue> 93
after RValue> 93
in yield> 93
before RValue> 93
```

Yeah, so… keep your hands off my chunking? ;-) I'm not sure I understand the point of a `Stream#map` which *doesn't* preserve chunk boundaries. This manifested as a 3x performance regression between fs2 0.10.0 and 1.0.0-M.

For reference, the surrounding commit state: https://github.com/wemrysi/quasar/commit/50ab260d37

mpilquist:
I don't think chunking is getting destroyed here, but rather that `Stream.chunk(c).map(f)` is equivalent to `Stream.segment(Segment.chunk(c).map(f))`. When you ask for the chunks that make up `Segment.chunk(c).map(f)`, the individual elements are emitted as singleton chunks. I don't think this is the root cause of the performance issue you are seeing, as in general, the fused segment will perform better than creating a bunch of intermediate chunks (though not always!).

[...]

PR #1161, opened 2018-06-15
Optimized unconsChunk

mpilquist:
See #1160 for details, but basically, operations like `Stream#map` result in dechunking if an `unconsChunk` is encountered. To fix, this PR changes `Segment#unconsChunk` such that when stepping, all of the emitted chunks from a single step get copied to a new chunk instead of taking the first and pushing the rest in to the tail.

In the test added to `StreamPerformanceSpec`, this change results in an improvement from about 2500ms to 15ms.

To implement an efficient `Chunk.concat`, I lifted the implementation from fs2 0.9 and added support for the primitive chunk types. One note: calling `concat` with `ByteVectorChunk`s results in `instanceOf[Byte]` checks on each element. I have two ideas on how to fix this:
1) Introduce a trait like: `trait KnownClass[A] { self: Chunk[A] => val classTag: ClassTag[A] }` and then use an `instanceOf` check on this type in `concat`
2) Move `ByteVectorChunk` in to fs2-core, including the dependency on `scodec-bits`. Note scodec hasn't broken binary compatibility since Feb 2016.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -1,6 +1,5 @@
 package fs2
 
-import scala.collection.immutable.VectorBuilder
 import scala.reflect.ClassTag
 import java.nio.{ByteBuffer => JByteBuffer}
 
@@ -70,10 +69,10 @@ abstract class Chunk[+O] extends Serializable {
 
   /** Creates a new chunk by applying `f` to each element in this chunk. */
   def map[O2](f: O => O2): Chunk[O2] = {
-    val b = new VectorBuilder[O2]
+    val b = collection.mutable.Buffer.newBuilder[O2]
     b.sizeHint(size)
     for (i <- 0 until size) b += f(apply(i))
-    Chunk.indexedSeq(b.result)
+    Chunk.buffer(b.result)
   }
 
   /** Splits this chunk in to two chunks at the specified index. */
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -268,6 +268,15 @@ abstract class Chunk[+O] extends Serializable { self =>
       buf.result
     }
 
+  /**
+    * Returns true if this chunk is known to have elements of type `B`.
+    * This is determined by checking if the chunk type mixes in `Chunk.KnownElementType`.
+    */
+  def knownElementType[B](implicit classTag: ClassTag[B]): Boolean = this match {
+    case ket: Chunk.KnownElementType[_] if ket.elementClassTag eq classTag => true
+    case _                                                                 => false
+  }
+
   override def hashCode: Int = toVector.hashCode
 
   override def equals(a: Any): Boolean = a match {
@@ -282,6 +291,11 @@ abstract class Chunk[+O] extends Serializable { self =>
 }
 
 object Chunk {
+
+  /** Optional mix-in that provides the class tag of the element type in a chunk. */
+  trait KnownElementType[A] { self: Chunk[A] =>
+    def elementClassTag: ClassTag[A]
+  }
 
   private val empty_ : Chunk[Nothing] = new Chunk[Nothing] {
     def size = 0
@@ -468,8 +482,10 @@ object Chunk {
     Booleans(values, offset, length)
 
   final case class Booleans(values: Array[Boolean], offset: Int, length: Int)
-      extends Chunk[Boolean] {
+      extends Chunk[Boolean]
+      with KnownElementType[Boolean] {
     checkBounds(values, offset, length)
+    def elementClassTag = ClassTag.Boolean
     def size = length
     def apply(i: Int) = values(offset + i)
     def at(i: Int) = values(offset + i)
@@ -507,8 +523,11 @@ object Chunk {
   def bytes(values: Array[Byte], offset: Int, length: Int): Chunk[Byte] =
     Bytes(values, offset, length)
 
-  final case class Bytes(values: Array[Byte], offset: Int, length: Int) extends Chunk[Byte] {
+  final case class Bytes(values: Array[Byte], offset: Int, length: Int)
+      extends Chunk[Byte]
+      with KnownElementType[Byte] {
     checkBounds(values, offset, length)
+    def elementClassTag = ClassTag.Byte
     def size = length
     def apply(i: Int) = values(offset + i)
     def at(i: Int) = values(offset + i)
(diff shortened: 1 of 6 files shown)


mpilquist:
I introduced `Chunk.KnownElementType` to solve the `ByteVectorChunk` problem. Should be ready for merge now.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -248,6 +268,15 @@ abstract class Chunk[+O] extends Serializable {
...
     }
 
+  /**
+    * Returns true if this chunk is known to have elements of type `B`.
+    * This is determined by checking if the chunk type mixes in `Chunk.KnownElementType`.
+    */
+  def knownElementType[B](implicit classTag: ClassTag[B]): Boolean = this match {
+    case ket: Chunk.KnownElementType[_] if ket.elementClassTag eq classTag => true
+    case _                                                                 => false
+  }
```
djspiewak: I'm really not sure this is actually enough, since wouldn't `ByteVectorChunk` have the same element type as `ByteChunk` and `ByteBufferChunk`?

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -436,12 +482,20 @@ object Chunk {
...
+    def elementClassTag = ClassTag.Boolean
     def size = length
     def apply(i: Int) = values(offset + i)
     def at(i: Int) = values(offset + i)
 
+    def copyToArray[O2 >: Boolean](xs: Array[O2], start: Int): Unit =
+      if (xs.isInstanceOf[Array[Boolean]])
+        System.arraycopy(values, offset, xs, start, length)
+      else
+        values.iterator.slice(offset, offset + length).copyToArray(xs, start)
```
djspiewak: Actually you can just cast and use `arraycopy`. `Array[Boolean] <: Array[AnyRef]` in the JVM's typesystem.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -703,6 +828,144 @@ object Chunk {
       Doubles(values, 0, values.length)
   }
 
+  /** Concatenates the specified sequence of chunks in to a single chunk, avoiding boxing. */
+  def concat[A](chunks: Seq[Chunk[A]]): Chunk[A] =
+    if (chunks.isEmpty) {
+      Chunk.empty
+    } else if (chunks.forall(c => c.knownElementType[Boolean] || c.forall(_.isInstanceOf[Boolean]))) {
```
djspiewak: Huh… I vaguely remember this code. Did I actually write this predicate?

```scala
c.forall(_.isInstanceOf[Boolean])
```

It would be radically more efficient to have a `allTypeOf[Boolean]` check, since that can be answered without actually iterating over the chunk. This is especially important for direct `ByteBuffer` chunks.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -703,6 +828,144 @@ object Chunk {
...
+      concatBooleans(chunks.asInstanceOf[Seq[Chunk[Boolean]]]).asInstanceOf[Chunk[A]]
+    } else if (chunks.forall(c => c.knownElementType[Byte] || c.forall(_.isInstanceOf[Byte]))) {
+      concatBytes(chunks.asInstanceOf[Seq[Chunk[Byte]]]).asInstanceOf[Chunk[A]]
```
djspiewak: This doesn't solve the `ByteVectorChunk` problem. Or `ByteBuffer`. Both of them will be paged out to `Array[Byte]` and then run through `concatBytes`.

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: Hm:

```scala
scala> val xs = new Array[Any](10)
xs: Array[Any] = Array(null, null, null, null, null, null, null, null, null, null)

scala> System.arraycopy(xs.asInstanceOf[Array[Byte]], 0, Array[Byte](1, 2, 3), 0, 3)
java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [B
  ... 29 elided
```

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: `c.knownElementType[Boolean]` will return true if `c` is an instance of `Booleans`. Otherwise, we resort to checking instance of each value. How would `allTypeOf` be implemented? If it's specialized in each of the primitive chunk types, then `knownElementType` accomplishes the same thing, right?

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: Both `ByteVectorChunk` and `ByteBuffer` implement `KnownElementType[Byte]` and both have reasonably efficient `copyToArray` implementations (`ByteBuffer` more so than `ByteVector`). What am I missing?

core/shared/src/main/scala/fs2/Chunk.scala
djspiewak: I… do not have a brain. :-D Not sure what I was thinking on this one.

core/shared/src/main/scala/fs2/Chunk.scala
djspiewak: Ah it's an `||`. Brain similarly not working on this one. You're very correct.

core/shared/src/main/scala/fs2/Chunk.scala
djspiewak: I guess there really isn't an efficient way to do `ByteBuffer` append intrinsically. One of the reasons why `ByteVector` is nicer. ;-)

With `ByteVector` though, you don't *want* to copy out to an array: you want to use the intrinsic append. This is actually really really important, because if you have a `ByteVector` which is wrapped around a direct `ByteBuffer`, you can append it without ever paging those bytes into heap memory. If you use `copyToArray`, you're sort of screwed.

mpilquist: Right, agreed. It would be much nicer to use stuff like `ByteVector.++` if most chunks are `ByteVector` backed, though that's not *much* different than using `Catenable`. I think we had polymorphic dispatch based on type of first chunk at some point, but maybe that never made it passed the discussion phase. That's definitely not what ended up in 0.9 final though.

Whatever we do, it has to account for things like `Chunk.empty` and `Chunk.singleton` being intermixed with the high density chunk types. Also should probably handle `Chunk.buffer` as well since that can show up when dealing with small literal chunks (e.g. `Chunk(1, 2, 3)`).

djspiewak: > Whatever we do, it has to account for things like Chunk.empty and Chunk.singleton being intermixed with the high density chunk types. Also should probably handle Chunk.buffer as well since that can show up when dealing with small literal chunks (e.g. Chunk(1, 2, 3)).

I'll give it some thought. I think this is doable and not too difficult to PR.

PR #1163, opened 2018-06-15
Fix groupAdjacentBy handling empty chunks

### LINKED ISSUE #1162: Stream#groupAdjacentBy can error calling Chunk.empty.apply
I'm not sure of the reproduction case yet - as soon as I can I'll attach one.

Based on reading the code, the way to hit this code path is if `unconsChunk` returns `Some((Chunk.empty, rest))`

Daenyth:
Add a new test generator that gives us a filtered stream, which can emit
empty chunks.

The groupAdjacentBy test goes red with only the generator change; the
code change brings it back to green

Closes #1162 

-----

I don't know the right way to fix the `mapSegments` test failure yet.

I think the test is nonsensical; The test says

```scala
runLog(s.get.mapSegments(identity).segments) shouldBe runLog(s.get.segments)
```

but the `Segment` docstring says

>   The `equals` and `hashCode` methods are not defined for `Segment`.

So this means that `shouldBe` is not safe to use with `Segment` values

What if we change the test to say this?

```scala
runLog(s.get.mapSegments(identity).segments.map(_.force.toVector) shouldBe runLog(s.get.segments.map(_.force.toVector))
```

### DIFF
--- a/core/jvm/src/test/scala/fs2/PipeSpec.scala
+++ b/core/jvm/src/test/scala/fs2/PipeSpec.scala
@@ -275,7 +275,7 @@ class PipeSpec extends Fs2Spec {
     }
 
     "mapSegments" in forAll { (s: PureStream[Int]) =>
-      runLog(s.get.mapSegments(identity).segments) shouldBe runLog(s.get.segments)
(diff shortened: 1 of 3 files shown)


Daenyth:
@mpilquist @SystemFw any idea on the segments test?

mpilquist:
Looks good to me. I agree with the proposed change to the `mapSegments` test.

PR #1164, opened 2018-06-18
Proof of concept of how to implement Segment#unconsChunk to avoid Cat…

mpilquist:
Not for merge -- for review/awareness only.

@djspiewak This is basically what I was thinking -- in effect, `Segment#unconsChunk` gets its own interpreter which means we have custom callbacks for `emit` and `emits` that can build the resulting chunk directly.

In this proof of concept, I used `collection.mutable.Buffer` but this is terrible -- boxing and inefficient copying. To make this mergeable, we'd at least need to specialize the logic for primitives and use an array instead of a buffer (and hence add size hinting).

### DIFF
--- a/core/shared/src/main/scala/fs2/Catenable.scala
+++ b/core/shared/src/main/scala/fs2/Catenable.scala
@@ -179,6 +179,7 @@ object Catenable {
   /** Creates a catenable from the specified sequence. */
   def fromSeq[A](s: Seq[A]): Catenable[A] =
     if (s.isEmpty) empty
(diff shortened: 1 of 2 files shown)


djspiewak:
Browser tab has been added to my queue! 🎉

mpilquist:
@djspiewak I just pushed an additional commit which replaces `Catenable` with `List` in `unconsChunks`, which *should* result in better performance all around.

mpilquist:
No longer needed now that chunk is gone

PR #1165, opened 2018-06-20
Another stab at groupedWithin

changlinli:
I basically think @SystemFw's approach in #1158 is the way to go. This take tries to add several new features on his approach:

+ Better handling of segments, not breaking them up when not necessary
+ More accurate buffering: we only ever buffer up to the specified n
  elements instead of potentially 2n elements
+ No integer overflow to deal with by using arbitrary java Objects for
  comparison instead of Ints


SystemFw:
Closing this to consolidate all work on #1158 
Sorry for the noise

PR #1167, opened 2018-06-22
Put Linebacker in place as the blocking mechanism for io project

ChristopherDavenport:
Order of parameters may not be to taste. 

This is off published milestone since we're off cats-effect but will happily go to 1.0 with extended support as necessary with cats-effect and fs2.

Left the blocking in place in the odd locations, but we can add it in as well inside where ConcurrentEffect is currently being utilized.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -253,6 +253,7 @@ lazy val io = project
   .settings(mimaSettings)
   .settings(
     name := "fs2-io",
+    libraryDependencies ++= Seq("io.chrisdavenport" %% "linebacker" % "0.2.0-M1"),
     OsgiKeys.exportPackage := Seq("fs2.io.*"),
     OsgiKeys.privatePackage := Seq(),
     OsgiKeys.importPackage := {
--- a/io/src/main/scala/fs2/io/JavaInputOutputStream.scala
+++ b/io/src/main/scala/fs2/io/JavaInputOutputStream.scala
@@ -5,17 +5,21 @@ import scala.concurrent.{SyncVar, blocking}
 
 import java.io.{IOException, InputStream, OutputStream}
 
-import cats.effect.{ConcurrentEffect, ExitCase, IO, Sync, Timer}
+import cats.effect.{Async, ConcurrentEffect, ExitCase, IO, Sync, Timer}
 import cats.implicits.{catsSyntaxEither => _, _}
 
 import fs2.Chunk.Bytes
 import fs2.async.mutable
 import fs2.internal.Canceled
 
+import _root_.io.chrisdavenport.linebacker.Linebacker
+
 private[io] object JavaInputOutputStream {
   def readBytesFromInputStream[F[_]](is: InputStream, buf: Array[Byte])(
-      implicit F: Sync[F]): F[Option[Chunk[Byte]]] =
-    F.delay(blocking(is.read(buf))).map { numBytes =>
+      implicit F: Async[F],
+      linebacker: Linebacker[F],
+      timer: Timer[F]): F[Option[Chunk[Byte]]] =
+    Linebacker[F].blockTimer(Sync[F].delay(is.read(buf))).map { numBytes =>
       if (numBytes < 0) None
       else if (numBytes == 0) Some(Chunk.empty)
       else if (numBytes < buf.size) Some(Chunk.bytes(buf.slice(0, numBytes)))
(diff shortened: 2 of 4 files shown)


mpilquist:
Looks good to me. I'd love to see linebacker folded in to cats-effect at some point.

What do you mean here?
> Left the blocking in place in the odd locations, but we can add it in as well inside where ConcurrentEffect is currently being utilized.

ChristopherDavenport:
There are some places you are calling `blocking` where you are using concurrenteffect without a blocking pool. That this does not directly address but leaves in place. If we remove those it requires a second pass. As it's a more intensive change.

ChristopherDavenport:
The remaining scala `blocking` are mainly where you are attempting to communicate blocking in a pure rather then effectful position - https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/io/src/main/scala/fs2/io/JavaInputOutputStream.scala#L167

I'd be very happy to see this moved into cats-effect as well. My understanding is that won't be the case at least for a little while.

Anything else I can help clean up or help with this?

mpilquist:
@pchlupacek Okay to merge?

pchlupacek:
@ChristopherDavenport @mpilquist 
Some points to consider: 
- is licence of Linebacker compatible with fs2 ?
- I saw linebacker is not that much of LoCs. Would it be somehow feasible to get in in fs2 before it somehow appears in cats? 

I think it is cool to solve it conceptually, but at same time I would love to leave dependencies at minimum. 

Any thoughts ?

ChristopherDavenport:
- It is MIT license as well, so I don't believe there is any license incompatibility
- Linebacker aims to be the solid foundation for functional thread management. I believe it can be in fs2, but due to that ownership of concerns I question whether fs2 is the right home for it. Its also minimal as you said, and has been very stable thus far.

The `0.1.0` is seeing use in many companies production systems currently and the changes to bracket and timer make the entire `0.2.x` version more reliable(considerring instead going straight to `1.0.0` with cats-effect). I expect that several downstreams will also integrate with it as this functionality is something we are currently missing in the community.

If I can assist or better respond to your concerns I'm happy to help.

pchlupacek:
@ChristopherDavenport thanks for reply and explanation. Could you give me some time to take a look on this? I want just to avoid situation where we will find subtle blocking issue, and to fix that we will need to rebuild the graph with your dependency. 
Any suggestions on how that could be avoided?

wedens:
`Linebacker` is not a typeclass, why is it implicit like it's a typeclass (I have similar concerns about `Timer` but it's another story)?

pchlupacek:
Just a very rough idea. If we cannot get somehow cats and linabcker fuse, I am using this construct which seems to be pretty easy understand on what it does and as far as I can read does what we need from linebacker: 

```scala
private[effect] class EvalOnPartialSyntax[F[_]](val workEc: ExecutionContext) extends AnyVal {
    /** evaluates `f` on provided context and shifts back execution on the context supplied by `EC` **/
    def apply[A](f: F[A])(implicit F: Async[F], EC: ExecutionContext): F[A] =
      Async.shift(workEc) >>
      F.rethrow {
        f.attempt.flatMap { r =>
          Async.shift(EC) as r
        }
      }
  }

  /** evaluates `f` on provided context and shifts back execution on the context supplied by `EC` **/
  def evalOn[F[_]](workEc: ExecutionContext) : EvalOnPartialSyntax[F] = new EvalOnPartialSyntax[F](workEc)

/// usage

evalOn(BlockingEC)(F.delay { /* any blocking code */})
 
```

pchlupacek:
that said we can just add this to async. 

Also on another topic @mpilquist what do you think about migrating all the [F[_]...] constructs in fs2 to the partial eval syntax approach?

mpilquist:
Partial eval?

pchlupacek:
Sorry for confusion  ^^ EvalOnPartialSyntax approach where we have [F[_],A] to split F[_] and A types in construct

mpilquist:
Oh, I think that ends up being a lot of boilerplate which would complicate things like finding methods in ScalaDoc. I’d rather push to get the dotty syntax added to 2.13 or 2.14 (eg, eval[F = IO](prg)).

pchlupacek:
ah cool. Didn't know that this is coming.

mpilquist:
Yep, works now in Dotty:

```scala
scala> def eval[F[_], A](fa: F[A]): Int = 0
def eval[F <: [_$2] => Any, A](fa: F[A]): Int

scala> eval[F = List](List(1))
val res1: Int = 0

scala> eval[A = Int](List(1))
val res2: Int = 0
```

I don't know how likely it is that this syntax makes Scala 2 but I can wish. :)

BTW, I tried some syntax for partially apply `F` in the constructor methods on the `Stream` companion but it didn't work out too well. Methods which infer correctly normally (like `eval`) needed to be left alone whereas methods that had `F` unconstrained needed to move to the template trait. However, calling such methods look different then. E.g. you have to do `Stream[IO].emits` instead of `Stream.emits[IO, Int](...)`.

pchlupacek:
ok thanks @mpilquist.

PR #1168, opened 2018-06-25
Add other Chunk types backed by different native buffer types

yilinwei:
Not particularly fond of this PR since the solution isn't particularly elegant and putting it out for comments.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -1,8 +1,17 @@
 package fs2
 
 import scala.reflect.ClassTag
-import java.nio.{ByteBuffer => JByteBuffer}
 import scodec.bits.ByteVector
+import java.nio.{
+  Buffer => JBuffer,
+  ByteBuffer => JByteBuffer,
+  ShortBuffer => JShortBuffer,
+  LongBuffer => JLongBuffer,
+  CharBuffer => JCharBuffer,
+  IntBuffer => JIntBuffer,
+  FloatBuffer => JFloatBuffer,
+  DoubleBuffer => JDoubleBuffer
+}
 
 import cats.{Applicative, Eq, Eval, Monad, Traverse}
 import cats.implicits._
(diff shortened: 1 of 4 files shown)


pchlupacek:
Just a remark could we restrict the buffers to be read-only buffers? I am really worried to have chunk backed by mutable java buffers. 

I know we do that for arrays, but these imho are not that often subject of mutations as Buffers are. 

Any thoughts on that ?

mpilquist:
We do the same for ByteBuffer and Chunk.buffer. There’s only so much we can do here — big warnings in ScalaDoc or some scary naming convention would both be fine.

pchlupacek:
I like the `view` approach from ByteVector, perhaps we can have naming convention liek that? I.e. default names will make defensive copy and view will be really `view` on BB instance ?

mpilquist:
That sounds good to me

mpilquist:
I likely won't have time to review in depth this week but I'm in favor of this approach in general. Naming wise, perhaps we should go with `NativeBuffer` and then things like `NativeIntBuffer` or something, to differentiate from the chunk that's backed by a `collection.mutable.Buffer`. Also curious about the `@specialized` use and why it is only specialized for 4 of the types.

yilinwei:
@mpilquist 

I started to specialise the types but I want to double check the byte code and how it plays with the superclass (since `Chunk`) isn't specialised. I should have some time tomorrow or the day after to take a look at that.

yilinwei:
As expected, the specialization needs to be on the superclass (unless, you explicitly use the subclasses). Is there any reason why the specialization can't be on `Chunk`?

yilinwei:
OK, looking at the java byte code I think the specialisation works as expected now.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -20,7 +30,7 @@ import cats.implicits._
   * intermediate chunks being created (1 per call to `map`). In contrast, a chunk can be lifted to a segment
   * (via `toSegment`) to get arbitrary operator fusion.
   */
-abstract class Chunk[+O] extends Serializable {
+trait Chunk[@specialized(Byte, Short, Long, Char, Int, Float, Double) +O] extends Serializable {
```
yilinwei: Unfortunately, scala didn't allow me to keep this as an abstract class.

mpilquist: Changing to a trait means it will bbe harder to preserve backwards compatibility on 2.11. Not a deal breaker if the specialization pays off though.

yilinwei: Hmm.

It's hard to say whether there's much of a benefit. Methods such as `apply` work better, but like `Stream`, `map` and other generic methods don't see any other benefit.

I'm leaning more of removing the specialization since `Stream` doesn't have it and I can't imagine most users are using `apply` very much and in the cases that matter using `toArray` or adding helper methods to get the underlying `Buffer` would serve the user better.

wdyt?

djspiewak: My two-cents: `Chunk#apply` probably shouldn't exist at all if we can avoid it, and that's really the only place where the specialization would matter. In all other cases, users are forced to respecialize *anyway* in order to avoid boxing, so it shouldn't matter.

yilinwei:
I've removed the specialisation because I don't think it adds enough.

Finally regarding the name, I don't think that a prefix of `Native` is a good idea - it's somewhat ambiguous because it could mean native as in living outside the JVM heap as well.

mpilquist:
Okay looks good to me. Can we add tests to `ChunkSpec`?

yilinwei:
> Okay looks good to me. Can we add tests to `ChunkSpec`?

@mpilquist

Done.

mpilquist:
@yilinwei Thanks for the patience in getting this PR over the finish line. One more comment -- do you need this in 0.10 or should we retarget to 1.0? I'd rather keep this only in 1.0 if possible.

yilinwei:
No problem.

1.0 is fine for me; I've got these classes locally anyway.

mpilquist:
Okay sounds good - I changed the target branch but it looks like we'll need to resolve some conflicts.

yilinwei:
Hmm. It seems as though there's some problem with the js build - is this expected?

mpilquist:
@yilinwei I looked at these errors on Friday and noticed the class cast exceptions all occur on the JVM. Note these tests pass on series/1.x. Let me know if you need help in tracking down the cause.

yilinwei:
@mpilquist You're right - I didn't check the tail end of the logs.

I'm having real trouble replicating the error on my side unfortunately.

Do you mind checking out my branch and running `testOnly fs2.ChunkSpec` to see if you get the error please?

mpilquist:
@yilinwei Hm, I can't reproduce the failure locally either...

Let's look at just the `CompressSpec` failure - it fails on the following line, failing due to `hd.toArray` returning an `Array[Int]` instead of `Array[Byte]` contrary to what the types claim:

```scala
deflater.setInput(hd.toArray)
```

The test case that fails asks for an implicit `PureStream[Byte]`. Could there be something about the modified `arbChunk` that causes this? I don't see anything but I'm not familiar with `Buildable`.

The Travis build uses the Oracle JDK 8 on Linux. Locally, I'm using Oracle JDK 8 on a Mac.

yilinwei:
@mpilquist I don't think so - we don't bring in the implicits from the ChunkSpec at all and I get the same test failures when commenting out all the ChunkSpec code.

When actually looking at the `Stream` we can actually see that we construct a `Chunk.booleans` rather than a `Chunk.bytes`

I'm going to have to take a look at this in more detail at the weekend because I have no idea what's going on.

mpilquist:
I think the bug might be this line:
```scala
def elementClassTag = implicitly
```

Try changing to:
```scala
def elementClassTag: ClassTag[C] = implicitly[ClassTag[C]]
```

mpilquist:
That seems to have helped - now only ChunkSpec.concat is failing

yilinwei:
Hmm, are you talking about the inflate/deflate? I still seem to get the same error locally.

mpilquist:
@yilinwei Hm okay, inflate/deflate is failing on the JVM build while IntBuffer.concat, ShortBuffer.contact, and some others are failing on the JS build. I can work on this some more if you could first resolve the conflicts against series/1.0.

yilinwei:
@mpilquist How did you work out that it was a scalajs compiler bug might I ask?

mpilquist:
I don’t know for sure that it’s a compiler bug but it certainly appears so. I added some printlns in copyToArray and confirmed the then branch was executing, not the else branch. I also confirmed we could cast to the array type without an exception. So then I tried adding an override directly in the various subclasses that don’t have corresponding primitives on Scala.js and noticed the same exact code worked (adjusting for the type param difference).

Issue #1169, opened 2018-07-04
Presence of cats.implicits._ can cause type mismatches


paulcarey:
This may not be a bug, but it is a rough edge and is likely to trip up newcomers who have imported `cats._, cats.implicits._` and expect them to play nicely with fs2.

```
scala> import cats.effect._
import cats.effect._

scala> import fs2._
import fs2._

scala> val s = Stream(1).flatMap(i => Stream.eval(IO { println(i) }))
s: fs2.Stream[cats.effect.IO,Unit] = Stream(..)

scala> import cats.implicits._
import cats.implicits._

scala> val s = Stream(1).flatMap(i => Stream.eval(IO { println(i) }))
<console>:20: error: type mismatch;
 found   : cats.effect.IO[Unit]
 required: fs2.Pure[?]
       val s = Stream(1).flatMap(i => Stream.eval(IO { println(i) }))
                                                     ^
```

mpilquist:
Yeah, this is an unfortunate consequence of the trickery we play with `Pure` to make inference work nicely. Specifically, `Stream(1)` gives you a `Stream[Pure, Int]` -- when you have `cats.implicits._` imported, you get `flatMap` from `FlatMapSyntax`, which means you end up needing a function `[B] A => Stream[Pure, B]`.

Here's an example from the test suite: https://github.com/functional-streams-for-scala/fs2/blob/800ae03dbf34799ba98e898595e3c8dd1e43954a/core/shared/src/test/scala/fs2/CompilationTest.scala#L59-L64

You can work around it in a few ways -- either adding manual `covary` calls such that the cats provided `flatMap` ends up with a compatible signature, or importing `fs2.Stream.{PureOps, InvariantOps}` (one or both depending on what syntax you are using) *after* you've imported `cats.implicits._`.  E.g.:

```
import cats.implicits._
import fs2.Stream.{PureOps, InvariantOps}
val s = Stream(1).flatMap(i => Stream.eval(IO { println(i) }))
```

This would be a good addition to the faq.md file (PR appreciated!)

smarter:
> Scala 3 might offer some relief via polymorphic function values + fixed inference bugs related to Nothing, the latter of which would enable us to replace Pure with Nothing and keep variance.

[...]

PR #1173, opened 2018-07-09
Remove the InvariantOps/PureOps encoding in favor of covariance

mpilquist:
In response to #1169 and the subsequent blog post, @smarter recommended a trick for convincing Scala 2 to infer unconstrained type params (https://gitter.im/typelevel/general?at=5b3e73a63572e970c17951f7).

We tried this when working on 0.9, but for whatever reason, decided it didn't work -- perhaps we didn't try hard enough or perhaps there were other compiler bugs at the time that prevented us from using this trick.

In this PR, I was able to get rid of almost all use of `InvariantOps`, though not completely. ~~We still need `InvariantOps` for `through` and `to` or otherwise we run in to issues with inference of function parameters. To get a feel for the issue, try moving those definitions directly to `Stream`.~~

~~Also, the current version requires a manual covary before joining a pure stream. E.g., `Stream(s, s).joinUnbounded` does not compile whereas `Stream(s, s).covary[IO].joinUnbounded` does. We could likely fix this by moving `join` and `joinUnbounded` to both `InvariantOps` and `PureOps`.~~

~~Another issue is that scalac crashes on expressions like `Stream.empty.merge(s)` but `Stream.empty[IO].merge(s)` works fine.~~

~~Another issue is that `Stream.empty.covary[IO]` no longer compiles, though `Stream.empty[IO]` or `Stream.empty[Pure].covary[IO]` both compile. You can see a few instances of this change in the diff.~~

Looking for feedback for now, not a merge.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -157,8 +157,8 @@ object Pull {
         .handleErrorWith(t => Algebra.pure[F, Nothing, Either[Throwable, R]](Left(t))))
 
   /** The completed `Pull`. Reads and outputs nothing. */
-  val done: Pull[Nothing, Nothing, Unit] =
-    fromFreeC[Nothing, Nothing, Unit](Algebra.pure[Nothing, Nothing, Unit](()))
+  val done: Pull[Pure, Nothing, Unit] =
+    fromFreeC[Pure, Nothing, Unit](Algebra.pure[Pure, Nothing, Unit](()))
 
   /** Evaluates the supplied effectful value and returns the result as the resource of the returned pull. */
   def eval[F[_], R](fr: F[R]): Pull[F, Nothing, R] =
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -20,8 +20,6 @@ import fs2.internal.{Algebra, FreeC, Token}
   *
   * `raiseError` is caught by `handleErrorWith`:
   *   - `handleErrorWith(raiseError(e))(f) == f(e)`
-  *
-  * @hideImplicitConversion covaryPure
   */
 final class Pull[+F[_], +O, +R] private (private val free: FreeC[Algebra[Nothing, Nothing, ?], R])
     extends AnyVal {
@@ -59,29 +57,23 @@ final class Pull[+F[_], +O, +R] private (private val free: FreeC[Algebra[Nothing
 
   /** Applies the resource of this pull to `f` and returns the result. */
   def flatMap[F2[x] >: F[x], O2 >: O, R2](f: R => Pull[F2, O2, R2]): Pull[F2, O2, R2] =
-    Pull.fromFreeC(get[F2, O2, R].flatMap { r =>
-      f(r).get
-    })
+    Pull.fromFreeC(get[F2, O2, R].flatMap(r => f(r).get))
 
   /** Alias for `flatMap(_ => p2)`. */
   def >>[F2[x] >: F[x], O2 >: O, R2](p2: => Pull[F2, O2, R2]): Pull[F2, O2, R2] =
-    this.flatMap { _ =>
-      p2
-    }
+    flatMap(_ => p2)
 
   /** Lifts this pull to the specified effect type. */
   def covary[F2[x] >: F[x]]: Pull[F2, O, R] = this.asInstanceOf[Pull[F2, O, R]]
 
   /** Lifts this pull to the specified effect type, output type, and resource type. */
-  def covaryAll[F2[x] >: F[x], O2 >: O, R2 >: R]: Pull[F2, O2, R2] =
-    this.asInstanceOf[Pull[F2, O2, R2]]
+  def covaryAll[F2[x] >: F[x], O2 >: O, R2 >: R]: Pull[F2, O2, R2] = this
 
   /** Lifts this pull to the specified output type. */
-  def covaryOutput[O2 >: O]: Pull[F, O2, R] = this.asInstanceOf[Pull[F, O2, R]]
+  def covaryOutput[O2 >: O]: Pull[F, O2, R] = this
 
   /** Lifts this pull to the specified resource type. */
-  def covaryResource[R2 >: R]: Pull[F, O, R2] =
-    this.asInstanceOf[Pull[F, O, R2]]
+  def covaryResource[R2 >: R]: Pull[F, O, R2] = this
 
   /** Applies the resource of this pull to `f` and returns the result in a new `Pull`. */
   def map[R2](f: R => R2): Pull[F, O, R2] = Pull.fromFreeC(get.map(f))
@@ -96,9 +88,7 @@ final class Pull[+F[_], +O, +R] private (private val free: FreeC[Algebra[Nothing
   /** If `this` terminates with `Pull.raiseError(e)`, invoke `h(e)`. */
   def handleErrorWith[F2[x] >: F[x], O2 >: O, R2 >: R](
       h: Throwable => Pull[F2, O2, R2]): Pull[F2, O2, R2] =
-    Pull.fromFreeC(get[F2, O2, R2].handleErrorWith { e =>
-      h(e).get
-    })
+    Pull.fromFreeC(get[F2, O2, R2].handleErrorWith(e => h(e).get))
 
   /** Tracks any resources acquired during this pull and releases them when the pull completes. */
   def scope: Pull[F, O, Unit] = Pull.fromFreeC(Algebra.scope(get[F, O, R].map(_ => ())))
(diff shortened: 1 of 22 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -293,6 +398,29 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+  /**
+    * Lifts this stream to the specified effect type.
+    *
+    * @example {{{
+    * scala> import cats.effect.IO
+    * scala> Stream(1, 2, 3).covary[IO]
+    * res0: Stream[IO,Int] = Stream(..)
+    * }}}
+    */
+  def covary[F2[x] >: F[x]]: Stream[F2, O] = this.asInstanceOf[Stream[F2, O]]
```
smarter: This asInstanceOf isn't necessary here.

smarter:
> Another issue is that Stream.empty.covary[IO] no longer compiles, though Stream.empty[IO] or Stream.empty[Pure].covary[IO] both compile.

Weird. As a workaround, you can get it to work by defining `covary` in both `InvariantOps` and `PureOps`: https://github.com/smarter/fs2/commits/topic/variance%2Bfix-covary

mpilquist:
@smarter Thanks, I added that commit

smarter:
> Another issue is that scalac crashes on expressions like Stream.empty.merge(s) but Stream.empty[IO].merge(s) works fine.

That's definitely a bug that needs to be reported. Meanwhile, going back to the non-polymorphic version of `Stream.empty` avoids the crash: https://github.com/smarter/fs2/commits/topic/variance%2Bfix-Stream-empty (For convenience, you could still define `def apply[F[_]]: Stream[F, O] = covary` in `PureOps` to be able to write `Stream.empty[IO]` instead of `Stream.empty.covary[IO]`).

smarter:
> We still need InvariantOps for through and to or otherwise we run in to issues with inference of function parameters.

`Stream#through` can be made to work by generalizing its type signature, though that may not be what you want your API to look like:
```scala
def through[F2[X] >: F[X], O2](f: Stream[F, O] => Stream[F2, O2]): Stream[F2, O2] = f(this)
// instead of
def through[F2[X] >: F[X], O2](f: Stream[F2, O] => Stream[F2, O2]): Stream[F2, O2] = f(this)
```

smarter:
There's a few uses of covary that are no longer needed with this PR: https://github.com/smarter/fs2/commits/topic/variance%2Bless-covary

smarter:
There's a lot of interesting type inference patterns going on here, it'd be great if we could come up with an "idealized" version of the fs2 API that we can make sure compiles with Dotty without workarounds (by improving how we do type inference if necessary) and add to our test suite.

(Also please avoid using `@smarter` in commit messages, otherwise github will send me an email everytime the PR is rebased!)

mpilquist:
The `StreamFactory` approach works well in some cases but makes inference worse in others. For example, if `eval` was moved up from `Stream` to `StreamFactory`, this expression fails to compile: `Stream.eval(IO(1))`. I suppose you meant only moving constructors that had unconstrained `F` params?

smarter:
>  I suppose you meant only moving constructors that had unconstrained F params?

Didn't think about it but yes that makes more sense.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -193,46 +183,42 @@ object Pull {
...
 
-  private def release[F[_]](token: Token): Pull[F, Nothing, Unit] =
+  private def release[F[x] >: Pure[x]](token: Token): Pull[F, Nothing, Unit] =
     fromFreeC[F, Nothing, Unit](Algebra.release(token))
 
-  /** Implicitly covaries a pull. */
-  implicit def covaryPure[F[_], O, R, O2 >: O, R2 >: R](p: Pull[Pure, O, R]): Pull[F, O2, R2] =
-    p.asInstanceOf[Pull[F, O, R]]
-
   /** `Sync` instance for `Stream`. */
```
SystemFw: `s/Stream/Pull`

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -678,6 +1074,159 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    *
+    * Finalizers on the returned stream are run after the outer stream has finished
+    * and all open inner streams have finished.
+    *
+    * @param maxOpen    Maximum number of open inner streams at any time. Must be > 0.
+    */
+  def join[F2[_], O2](maxOpen: Int)(implicit ev: O <:< Stream[F2, O2],
+                                    ev2: F[_] <:< F2[_],
+                                    F2: Concurrent[F2]): Stream[F2, O2] = {
+    assert(maxOpen > 0, "maxOpen must be > 0, was: " + maxOpen)
```
SystemFw: Not entirely clear about how this is working. is `ev2` taking `F[_]` and `F2[_]` existentially? What makes this op different from the rest, the bound on `O`?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1705,1257 +2684,116 @@ object Stream {
...
-                .attempt
-                .map { _.left.toOption }
-                .flatMap { r =>
-                  // to prevent deadlock, done must be signalled before `interruptL`
-                  // in case the interruptL is signaled before the `L` stream may be in
-                  // its `append` code, that requires `get` to complete, which won't ever complete,
-                  // b/c it will be evaluated after `interruptL`
-                  doneR.complete(r) >>
-                    r.fold(F.unit)(interruptL.complete)
+    def observe(sink: Sink[F, O])(implicit F: Concurrent[F]): Stream[F, O] =
```
SystemFw: What's about `observe` that needs to be in InvariantOps? the use of `Sink`?

Also at the top of the file (can't comment there for some reason) it says:

> Most of the api is defined in InvariantOps

which can probably be removed

core/shared/src/test/scala/fs2/CompilationTest.scala
```diff
@@ -56,10 +61,17 @@ object ThisModuleShouldCompile {
...
     import cats.implicits._
-    Stream(1,2,3).covary[IO].flatMap(i => Stream.eval(IO.pure(i)))
-    (Stream(1,2,3).covary[IO].flatMap(i => Stream.eval(IO(i)))): Stream[IO,Int]
+    1 |+| 1 // Mask unused warning from cats.implicits._ import
+    Stream(1,2,3).flatMap(i => Stream.eval(IO.pure(i)))
+    (Stream(1,2,3).flatMap(i => Stream.eval(IO(i)))): Stream[IO,Int]
   }
+
+  // Join a pure stream of effectful streams without type annotations
+  Stream(s, s).joinUnbounded
```
SystemFw: This is great :)

core/shared/src/test/scala/fs2/CompilationTest.scala
```diff
@@ -56,10 +61,17 @@ object ThisModuleShouldCompile {
...
 
-  // With cats implicits enabled, some of the above fail to compile due to the cats syntax being invariant:
+  // Ensure that Stream#flatMap is favored over cats's flatMap
   {
     import cats.implicits._
-    Stream(1,2,3).covary[IO].flatMap(i => Stream.eval(IO.pure(i)))
-    (Stream(1,2,3).covary[IO].flatMap(i => Stream.eval(IO(i)))): Stream[IO,Int]
+    1 |+| 1 // Mask unused warning from cats.implicits._ import
+    Stream(1,2,3).flatMap(i => Stream.eval(IO.pure(i)))
+    (Stream(1,2,3).flatMap(i => Stream.eval(IO(i)))): Stream[IO,Int]
```
SystemFw: 👍 

I think we used to have another problem because of this conflict, but can't remember what :(

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: A couple of things here -- `F[_] <:< F2[_]` isn't existential but rather evidence that `F` is a subtype of `F2`. This is the typelevel equivalent of the `F2[x] >: F[x]` bound on the type parameter.

In this case though, if we put that constraint on `F2` directly, then scalac fails to prove that `O` is `Stream[F2, O2]`, due to limitation in inferencing / type checking AFAICT. So instead, we leave `F2` unbound, let scalac derive `F2` via `O <:< Stream[F2, O2]` constraint, and then check if `F2` is a super type of `F`.

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: I tried inlining this similar to `through` and `to` -- basically this:
```scala
def observe[F2[x] >: F[x]: Concurrent](sink: Stream[F, O] => Stream[F2, Unit]): Stream[F2, O] = ...
```

Unfortunately, this doesn't work because in the implementation of `observeAsync`, we do more than just call `sink(this)` -- rather, we create a `Queue[F2, ...]` and then call `sink(q.dequeue)`.

core/shared/src/test/scala/fs2/CompilationTest.scala
mpilquist: Yeah, we'd pick up `FlatMapOps#flatMap` and it would complain b/c it wanted a `[B] Int => Stream[Pure, B]` but got `Int => Stream[IO, Int]`

mpilquist:
Pushed review updates

Issue #1181, opened 2018-07-19
Make ByteVectorChunk a case class


lJoublanc:
The read/write methods in `io.tcp` and friends return `F[Option[Chunk[Byte]]]`. This minor change allows pattern matching against that to extract the bytes.

I will submit a pull req in a sec.

mpilquist:
Fixed by #1881

PR #1182, opened 2018-07-19
Changed ByteVectorChunk to case class as in #1181

lJoublanc:
Allowing extractor to be used with io.tcp.Socket.read/write.


mpilquist:
Thanks!

PR #1183, opened 2018-07-21
Upgraded to latest cats-effect hash build

mpilquist:


SystemFw:
The failure is `upstream.isClosed` again, so I'll merge

PR #1186, opened 2018-07-24
Changed core interpreter and API to be chunk based instead of segment based

### LINKED ISSUE #1175: Remove Segment?
I'm not sure `Segment` is really helping performance in real-world use cases. This was noticed recently in https://github.com/functional-streams-for-scala/fs2/issues/1160, where even with some optimizations in 1.0.0-SNAPSHOT, a real world use case is about 2x slower under 1.0 than under 0.9.

Last time I did some benchmarking of 0.9 vs 0.10, I found that while 0.10 was significantly faster than 0.9 in most use cases, writing algorithms in terms of chunks instead of segments was generally faster (https://speakerdeck.com/mpilquist/fs2-internals?slide=24). Segment did win with certain chunk patterns though. Those results were with 0.10.0-M8. Running the same tests against 1.0.0-SNAPSHOT should favor the chunk backed algorithms even more.

Algorithms are often much harder to implement in terms of `Segment` than in terms of `Chunk`. Users often don't know which to implement and we typically offer the advice that `Segment` based algorithms will be faster. However the `take` benchmark shows that this isn't always true. Given the complexity of working with `Segment`, I feel like the performance gains need to be great in order to outweigh the complexity.

Like I said [here](https://github.com/functional-streams-for-scala/fs2/issues/1160#issuecomment-397812785), we shouldn't make any decisions without experiments. I'd appreciate feedback though. One option could be supporting both `Segment` and `Chunk` internally instead of lifting every chunk in to a segment? Or keep `Segment` but deprecate it?

mpilquist:
Fixes #1175

See #1175 for performance test results before and after this change.

Overall, chunk based streams perform better than segment based streams. It's possible to construct specialized benchmarks where segment based streams perform better but such benchmarks aren't transferrable to real stream programs. Given that we get better performance and the API ends up *significantly* simpler to work with, these issues outweigh the API breakage concerns, which I do take very seriously. Also note that we no longer have weirdness around when evaluation occurs and when exceptions are thrown due to the lazy nature of `Segment`.

I left `Segment` in place, and kept the `Stream.segment` and `Pull.segment` constructors, which now lazily unfold the segment in to chunks. This is nice in that users can opt-in to full fusion for workloads that require it, while not paying the cost of `Segment` for the more common cases.

Note: I'll work on a migration guide soon, which will include directions on dealing with the conversion. We should probably also consider providing a scalafix to help with the upgrade. We won't be able to rewrite pulls that used `Segment` to pulls that use `Chunk` but we will be able to handle method renames like `outputChunk` to `output`.

Credit to @djspiewak for getting me to re-benchmark FS2 with and without `Segment` and challenging the assumptions that `Segment` based streams are faster due to fusion.

### DIFF
--- a/core/jvm/src/main/scala/fs2/compress.scala
+++ b/core/jvm/src/main/scala/fs2/compress.scala
@@ -31,19 +31,19 @@ object compress {
 
   private def _deflate_stream[F[_]](deflater: Deflater,
                                     buffer: Array[Byte]): Stream[F, Byte] => Pull[F, Byte, Unit] =
-    _.pull.unconsChunk.flatMap {
+    _.pull.uncons.flatMap {
       case Some((hd, tl)) =>
         deflater.setInput(hd.toArray)
         val result =
           _deflate_collect(deflater, buffer, ArrayBuffer.empty, false).toArray
-        Pull.outputChunk(Chunk.bytes(result)) >> _deflate_stream(deflater, buffer)(tl)
+        Pull.output(Chunk.bytes(result)) >> _deflate_stream(deflater, buffer)(tl)
       case None =>
         deflater.setInput(Array.empty)
         deflater.finish()
         val result =
           _deflate_collect(deflater, buffer, ArrayBuffer.empty, true).toArray
         deflater.end()
-        Pull.outputChunk(Chunk.bytes(result))
+        Pull.output(Chunk.bytes(result))
     }
 
   @tailrec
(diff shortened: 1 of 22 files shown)


pchlupacek:
excellent @mpilquist lgtm

mpilquist:
Okay merging this one. If anyone wants to look at optimized chunk operations in chunk subtypes, feel free, especially for the primitive array backed entries.

dsebban:
Great! From a user point of view, the segment concept was indeed making the API confusing.

PR #1187, opened 2018-07-25
Merge fix

mraulim:
- finish upstream finalizers first
  - correctly propagate exceptions


AdamChlupacek:
Similar patch has to be applied to `concurrently` as well.

pchlupacek:
yes, given the fact that some issues still prevail in 0.10 series (i.e the fold issue).

PR #1189, opened 2018-07-26
Added ByteVectorChunk directly to core project and removed scodec module

mpilquist:


mpilquist:
Here's what I was thinking:
- We removed the scodec-bits dependency when we did 0.9, banishing the chunk instance to a submodule. At the time, fs2 had no other dependencies.
- By banishing the interop to a submodule, we aren't able to use it in FS2 itself and folks tend not to use it in general.
- FS2 now has both the cats and cats-effect dependency, each of which have broken BC many times since last scodec-bits dependency (Feb 2016).
- scodec-bits is used as a super-stable dependency in other projects, like fastparse
- In the rare event that scodec-bits did break BC, we'd only maintain one version anyway - we're not cross building nor pulling it out in to its own repo

Given all this, I think keeping it in its own submodule vastly reduces its usage without any of the flexibility we get by building a compat project. I'm hoping to wrap up fs2 1.0 over the next month and this was on my personal list of annoyances. Happy to debate though. :)

djspiewak:
> By banishing the interop to a submodule, we aren't able to use it in FS2 itself and folks tend not to use it in general.

This is probably the strongest point, IMO. The entire fs2-io package would benefit from this. It might also be possible to take advantage of it in `ByteBufferChunk`, since producing a `ByteVectorChunk` from most operations is likely more efficient than recreating a full `ByteBufferChunk`.

PR #1197, opened 2018-07-31
Bugfix/interrupt fold

### LINKED ISSUE #1179: `innterruptWhen` resets `fold` to zero
After `interruptWhen` triggers, `fold` emits it's zero value, even when it was accumulating values before interruption:
```scala
import cats.effect.IO
import cats.instances.int._
import fs2.{ Sink, Stream, StreamApp }

object InterruptFold extends StreamApp[IO] {
  override def stream(args: List[String], requestShutdown: IO[Unit]): fs2.Stream[IO, StreamApp.ExitCode] =
    Stream
      .unfold(0)(i => Some((i, i + 1)))
      .flatMap(Stream.emit(_).delayBy[IO](1.second))
      .interruptWhen(Stream.emit(true).delayBy[IO](15.seconds))
      .fold(0) { (t, i) =>
        print(s"$i "); t + i
      }
      .to(Sink.showLinesStdOut)
      .drain ++ Stream.emit(StreamApp.ExitCode.Success)
}
```
will output: `0 1 2 3 4 5 6 7 8 9 10 11 12 13 0`.

If we surround emitting stream with another one and join afterwards:
```scala
    Stream(
      Stream
        .unfold(0)(i => Some((i, i + 1)))
        .flatMap(Stream.emit(_).delayBy[IO](1.second))
        .interruptWhen(Stream.emit(true).delayBy[IO](15.seconds))
    ).join(1)
      .fold(0) { (t, i) =>
        print(s"$i "); t + i
      }
```
results will be of 'normal' fold: `0 1 2 3 4 5 6 7 8 9 10 11 12 13 91`.

Is it intended? And if so, is there a way to signal 'graceful' interruption with `Stream` or `Signal`?

Edit: scala 2.12.6, cats 1.1.0, cats-effect 1.0.0-RC2, fs2 1.0.0-M1

pchlupacek:
@mpilquist @SystemFw 
This fixes #1179 
Some notable changes: 

- FreeC now have special case for stream evaluation being interrupted
- FreeC now in `bind` takes `FreeC.Result` as argument, instead of Either. Result Type are pure cases of FreeC, so this also saves a bit on allocation
- Interruption is now signalled via CompileScope, however no longer scope keeps the `case` to run when interrupted
- Interpreter now does not have to `jump-off` in Uncons cases to open scope. Not only this makes it cleaner, but also it shall increase performance in certain scope-heavy interrupted streams. 
- Stream#flatMap has been simplified, which not only made code cleaner, but also leads to up to 70% speed increase in certain tests (See StreamPerformanceSpec)
- I am not sure 100% about Pipe.Stepper and Pipe2.Stepper implementation. There are no tests whatsoever for these. I think we shall either write some, or perhaps remove Stepper.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -811,8 +811,10 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
                 f(hd(idx)).get.transformWith {
                   case Result.Pure(_)   => go(idx + 1)
                   case Result.Fail(err) => Algebra.raiseError(err)
-                  case Result.Interrupted(scopeId, err) =>
+                  case Result.Interrupted(scopeId: Token, err) =>
                     Stream.fromFreeC(Algebra.interruptBoundary(tl, scopeId, err)).flatMap(f).get
+                  case Result.Interrupted(invalid, err) =>
+                    sys.error(s"Invalid interruption context: $invalid (flatMap)")
                 }
               }
 
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -797,15 +797,14 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
   def flatMap[F2[x] >: F[x], O2](f: O => Stream[F2, O2]): Stream[F2, O2] =
     Stream.fromFreeC[F2, O2](Algebra.uncons(get[F2, O]).flatMap {
       case Some((hd, tl)) =>
-        // nb: If tl is Pure, there's no need to propagate flatMap through the tail. Hence, we
-        // check if hd has only a single element, and if so, process it directly instead of folding.
-        // This allows recursive infinite streams of the form `def s: Stream[Pure,O] = Stream(o).flatMap { _ => s }`
-        val only: Option[O] = tl match {
-          case FreeC.Result.Pure(_) => if (hd.size == 1) hd.head else None
-          case _                    => None
-        }
-        only match {
-          case None =>
+        tl match {
+          case FreeC.Result.Pure(_) if hd.size == 1  =>
+            // nb: If tl is Pure, there's no need to propagate flatMap through the tail. Hence, we
+            // check if hd has only a single element, and if so, process it directly instead of folding.
+            // This allows recursive infinite streams of the form `def s: Stream[Pure,O] = Stream(o).flatMap { _ => s }`
+            f(hd(0)).get
+
+          case _ =>
             def go(rem: Chunk[O]): FreeC[Algebra[F2, O2, ?], Unit] =
               rem.head match {
                 case Some(o) =>
@@ -819,11 +818,8 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
               }
 
             go(hd)
-
-          case Some(o) =>
-            f(o).get
-
         }
+ 
       case None => Stream.empty.get
     })
 
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -798,28 +798,27 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     Stream.fromFreeC[F2, O2](Algebra.uncons(get[F2, O]).flatMap {
       case Some((hd, tl)) =>
         tl match {
-          case FreeC.Result.Pure(_) if hd.size == 1  =>
+          case FreeC.Result.Pure(_) if hd.size == 1 =>
             // nb: If tl is Pure, there's no need to propagate flatMap through the tail. Hence, we
             // check if hd has only a single element, and if so, process it directly instead of folding.
             // This allows recursive infinite streams of the form `def s: Stream[Pure,O] = Stream(o).flatMap { _ => s }`
             f(hd(0)).get
 
           case _ =>
-            def go(rem: Chunk[O]): FreeC[Algebra[F2, O2, ?], Unit] =
-              rem.head match {
-                case Some(o) =>
-                  f(o).get.transformWith {
-                    case Result.Pure(_)   => go(rem.drop(1))
-                    case Result.Fail(err) => Algebra.raiseError(err)
-                    case Result.Interrupted(scopeId, err) =>
-                      Stream.fromFreeC(Algebra.interruptBoundary(tl, scopeId, err)).flatMap(f).get
-                  }
-                case None => Stream.fromFreeC(tl).flatMap(f).get
+            def go(idx: Int): FreeC[Algebra[F2, O2, ?], Unit] =
+              if (idx == hd.size) Stream.fromFreeC(tl).flatMap(f).get
+              else {
+                f(hd(idx)).get.transformWith {
+                  case Result.Pure(_)   => go(idx + 1)
+                  case Result.Fail(err) => Algebra.raiseError(err)
+                  case Result.Interrupted(scopeId, err) =>
+                    Stream.fromFreeC(Algebra.interruptBoundary(tl, scopeId, err)).flatMap(f).get
+                }
               }
 
-            go(hd)
+            go(0)
         }
- 
+
       case None => Stream.empty.get
     })
 
@@ -1062,14 +1061,6 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     */
   def interruptWhen[F2[x] >: F[x]](haltOnSignal: F2[Either[Throwable, Unit]])(
       implicit F2: Concurrent[F2]): Stream[F2, O] =
-//    Stream.fromFreeC(
-//      Algebra.interruptScope(
-//        Algebra.getScope[F2, O, O].flatMap { scope =>
-//          Algebra
-//            .eval(F2.start(haltOnSignal.map { x => println("INTERR"); x }.flatMap(scope.interrupt)))
-//            .flatMap(_ => get[F2, O]) // todo bracket this and issue cancel on fiber
-//        }
-//      ))
     Stream
       .getScope[F2]
       .flatMap { scope =>
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -76,7 +76,10 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
 
   /** Appends `s2` to the end of this stream. Alias for `s1 ++ s2`. */
   def append[F2[x] >: F[x], O2 >: O](s2: => Stream[F2, O2]): Stream[F2, O2] =
-    Stream.fromFreeC(get[F2, O2].flatMap(_ => s2.get))
+    Stream.fromFreeC(get[F2, O2].transformWith {
+      case Result.Pure(_) => s2.get
+      case other          => other.asFreeC[Algebra[F2, O2, ?]]
+    })
 
   /**
     * Alias for `_.map(_ => o2)`.
@@ -798,14 +801,24 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
         // check if hd has only a single element, and if so, process it directly instead of folding.
         // This allows recursive infinite streams of the form `def s: Stream[Pure,O] = Stream(o).flatMap { _ => s }`
         val only: Option[O] = tl match {
-          case FreeC.Result.Pure(_) => hd.head
+          case FreeC.Result.Pure(_) => if (hd.size == 1) hd.head else None
           case _                    => None
         }
         only match {
           case None =>
-            hd.map(f)
-              .foldRightLazy(Stream.fromFreeC(tl).flatMap(f))(_ ++ _)
-              .get
+            def go(rem: Chunk[O]): FreeC[Algebra[F2, O2, ?], Unit] =
+              rem.head match {
+                case Some(o) =>
+                  f(o).get.transformWith {
+                    case Result.Pure(_)   => go(rem.drop(1))
+                    case Result.Fail(err) => Algebra.raiseError(err)
+                    case Result.Interrupted(scopeId, err) =>
+                      Stream.fromFreeC(Algebra.interruptBoundary(tl, scopeId, err)).flatMap(f).get
+                  }
+                case None => Stream.fromFreeC(tl).flatMap(f).get
+              }
+
+            go(hd)
 
           case Some(o) =>
             f(o).get
@@ -1053,6 +1066,14 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     */
   def interruptWhen[F2[x] >: F[x]](haltOnSignal: F2[Either[Throwable, Unit]])(
       implicit F2: Concurrent[F2]): Stream[F2, O] =
+//    Stream.fromFreeC(
+//      Algebra.interruptScope(
+//        Algebra.getScope[F2, O, O].flatMap { scope =>
+//          Algebra
+//            .eval(F2.start(haltOnSignal.map { x => println("INTERR"); x }.flatMap(scope.interrupt)))
+//            .flatMap(_ => get[F2, O]) // todo bracket this and issue cancel on fiber
+//        }
+//      ))
     Stream
       .getScope[F2]
       .flatMap { scope =>
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -5,7 +5,9 @@ import cats.data.NonEmptyList
 import cats.effect._
 import cats.effect.concurrent.{Deferred, Ref, Semaphore}
 import cats.implicits.{catsSyntaxEither => _, _}
+import fs2.internal.FreeC.Result
 import fs2.internal.{Algebra, Canceled, FreeC, Token}
+
 import scala.collection.generic.CanBuildFrom
 import scala.concurrent.duration._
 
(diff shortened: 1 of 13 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -792,23 +797,29 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+            // check if hd has only a single element, and if so, process it directly instead of folding.
+            // This allows recursive infinite streams of the form `def s: Stream[Pure,O] = Stream(o).flatMap { _ => s }`
+            f(hd(0)).get
+
+          case _ =>
+            def go(rem: Chunk[O]): FreeC[Algebra[F2, O2, ?], Unit] =
+              rem.head match {
+                case Some(o) =>
+                  f(o).get.transformWith {
+                    case Result.Pure(_)   => go(rem.drop(1))
```
mpilquist: I'd avoid using `rem.drop(1)` here as it will be expensive for large chunks. Instead, pass the index as a parameter to go and lookup the value via apply in each iteration.

pchlupacek: btw wouldn't apply(idx) be to slow for large Seq/List instances?

mpilquist: No, b/c there are no such instances -- `Chunk` guarantees efficient indexed based access. The `Chunk.seq` constructor copies the seq in to a mutable buffer for that reason.

pchlupacek: cool

core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -12,138 +12,204 @@ import scala.util.control.NonFatal
...
+  def eval[F[_], A](f: F[A]): FreeC[F, A] = Eval(f)
+
+  def raiseError[F[_], A](rsn: Throwable): FreeC[F, A] = Result.Fail(rsn)
+
+  def interrupted[F[_], A](scopeId: Token, failure: Option[Throwable]): FreeC[F, A] =
+    Result.Interrupted(scopeId, failure)
+
+  sealed trait Result[+R] { self =>
+
+    def asFreeC[F[_]]: FreeC[F, R] = self.asInstanceOf[FreeC[F, R]]
```
mpilquist: Can we avoid this case by declaring `self` as `self: FreeC[F, R]`?

core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -12,138 +12,204 @@ import scala.util.control.NonFatal
...
+
+    def asExitCase: ExitCase[Throwable] = self match {
+      case Result.Pure(_)           => ExitCase.Completed
+      case Result.Fail(err)         => ExitCase.Error(err)
+      case Result.Interrupted(_, _) => ExitCase.Canceled
+    }
+
+    def covary[R2 >: R]: Result[R2] = self.asInstanceOf[Result[R2]]
```
mpilquist: Don't need the cast here -- just define as `def covary[R2 >: R]: Result[R2] = self`

core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -12,138 +12,204 @@ import scala.util.control.NonFatal
...
+    }
+
+    // finalizerErrors contains accumulated errors during finalization of interruption.
+    // they will be presented to freeC after interupted scope will be closed
+    final case class Interrupted[F[_], R](scope: Token, finalizerErrors: Option[Throwable])
+        extends FreeC[F, R]
+        with Result[R] {
+      override def translate[G[_]](f: F ~> G): FreeC[G, R] =
+        this.asInstanceOf[FreeC[G, R]]
+      override def toString: String = s"FreeC.Interrupted($scope, ${})"
```
mpilquist: Is `${}` a typo?

core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -12,138 +12,204 @@ import scala.util.control.NonFatal
...
+    }
+
+    implicit class ResultSyntax[R](val self: Result[R]) extends AnyVal {
+
+      def map[R2](f: R => R2): Result[R2] = self match {
```
mpilquist: I don't think this will ever be called b/c `map` is defined directly on `FreeC`. Hence, scalac will never look for this extension method.

core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -12,138 +12,204 @@ import scala.util.control.NonFatal
...
+
+    final case class Fail[F[_], R](error: Throwable) extends FreeC[F, R] with Result[R] {
+      override def translate[G[_]](f: F ~> G): FreeC[G, R] =
+        this.asInstanceOf[FreeC[G, R]]
+      override def toString: String = s"FreeC.Fail($error)"
+    }
+
+    // finalizerErrors contains accumulated errors during finalization of interruption.
+    // they will be presented to freeC after interupted scope will be closed
+    final case class Interrupted[F[_], R](scope: Token, finalizerErrors: Option[Throwable])
```
mpilquist: The terminology here scares me a little bit -- `scope` and `finalizer` are stream concerns, not `FreeC` concerns. Could you add some docs either here or at the class level that describes this interruption model a little bit?

core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -12,138 +12,204 @@ import scala.util.control.NonFatal
...
-  private val pureContinuation_ = (e: Either[Throwable, Any]) =>
-    e match {
-      case Right(r) => Pure[Any, Any](r)
-      case Left(e)  => Fail[Any, Any](e)
-  }
 
-  def pureContinuation[F[_], R]: Either[Throwable, R] => FreeC[F, R] =
-    pureContinuation_.asInstanceOf[Either[Throwable, R] => FreeC[F, R]]
+  def pureContinuation[F[_], R]: Result[R] => FreeC[F, R] =
+    _.asFreeC[F]
```
mpilquist: Don't want to create one instance here and share, like we did before?

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: No `F` in Result type :-(

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: yeah

core/shared/src/main/scala/fs2/internal/FreeC.scala
mpilquist: Ah right of course

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: actually the problem is that `map` in FreeC is now implemented with this `map` and they both have different signatures (return types). This is picked over one on FreeC when FreeC is result. We could perhaps move `map` implementation to each FreeC case instead this magic?

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: ok yeah. It is a bit confusing, unfortunatelly we had not find a better way. I'll think of something.

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: do you think this won't get inlined? I can ofc get instance back.

core/shared/src/main/scala/fs2/internal/FreeC.scala
mpilquist: Ah, so `Result[A]` is *not* a subtype of `FreeC[F, A]`, which is why scalac doesn't see the `map` method on `FreeC`. This feels pretty delicate to me -- I think it would be clearer to override map on each instance but will defer to your opinion here.

core/shared/src/main/scala/fs2/internal/FreeC.scala
mpilquist: I gave up predicting what gets inlined a long time ago. Seems like it would be a Function1 allocation but who knows.

pchlupacek: ok :-) will then reintroduce that val

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: ok

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: Ok I think I know why I have decided to not have val there anymore: 

former code 
```
private val pureContinuation_ = (e: Either[Throwable, Any]) =>
    e match {
      case Right(r) => Pure[Any, Any](r)
      case Left(e)  => Fail[Any, Any](e)
  }

  def pureContinuation[F[_], R]: Either[Throwable, R] => FreeC[F, R] =
    pureContinuation_.asInstanceOf[Either[Throwable, R] => FreeC[F, R]]

```
As you see there is anyhow typeCat essentially what _.asFreeC does :-) 
One of the advantages we can pass along Result to bind

pchlupacek:
@mpilquist pushed changes. Could you please review comment on Interrupted case, if that is sort of explaining how the interruption works?

core/shared/src/main/scala/fs2/internal/FreeC.scala
mpilquist: I prefer the current definition of `map` -- I don't like adding the extra type param (`F`) and having result type be more general than `Result[B]`.

core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -12,138 +12,223 @@ import scala.util.control.NonFatal
...
+     * Essentially interpreter uses this to find any parents of this scope that has to be interrupted, and guards the
+     * interruption so it won't propagate to scope that shall not be anymore interrupted.
+     *
+     * Parameter `finalizerErrors` allows to collect errors that may accumulate during interruption propagation.
+     * When interruption is executed, all finalizers are executed, until interruption resumes. During this time,
+     * finalizers may accumulate errors. These errors are then stored here, and are raised when interruption resumes.
+     * Reason for this delayed reporting is to allow proper recovery with `handleErrorWith` handlers.
+     *
+     */
+    final case class Interrupted[F[_], R](scope: Token, finalizerErrors: Option[Throwable])
```
mpilquist: Maybe this is silly, but I would feel better about `Interrupted` class if the names of `scope` and `finalizerErrors` were more general. For example:

```scala
final case class Interrupted[F[_], R](token: Token, deferredError: Option[Throwable])
```

Or maybe even making the token parameterized:

```scala
final case class Interrupted[F[_], T, R](tag: T, deferredError: Option[Throwable])
```

mpilquist: Thanks for the docs by the way -- I'd prefer if we found a place for those docs in Algebra or Stream though. Conceptually, I'd like to keep `FreeC` as a standalone thing with no references to Streams.

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: the trouble is that for example Pure : 

Pure[A] extends FreeC[F, A] which requires to have map[B](f: A => B): FreeC[F, B]

but also 

Pure[A] extends Result[A] which requires to have map[B](f: A=>B) : Result[B] 

and compiler doesn't obviously like that. 

Not sure if I know the solution or understand what do you mean ....

mpilquist: Ah yes, okay I understand. We could fix by making `FreeC` covariant in both type params and making `Result[A] <: FreeC[Nothing, A]` but that's a lot of annoying machinery.

How about we rename `map` on `Result` to something else? `mapR` or something?

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: ok I like idea with type parameters. Not sure how that will play with current pattern match, in worst case I can have unapply variant on Algebra
 

```scala
final case class Interrupted[F[_], A, R](interruptContext: A)
```
and then I may move the explanation/doc to Algebra interpreter?

mpilquist: That sounds good

core/shared/src/main/scala/fs2/internal/FreeC.scala
pchlupacek: yeah, we could. Or you know, I could simply add monad instance for result and call F.map and thats it ?

mpilquist:
@SystemFw I'll wait on your review before merging. I think with next commit from @pchlupacek, `FreeC` will be completely devoid of any stream logic and the `Interrupted` constructor will simply allow early termination of a `FreeC` computation.

pchlupacek:
@mpilquist @SystemFw just pushed changes

mpilquist:
@SystemFw I'm going to merge now but please review anyway. We can address any comments in a subsequent PR. Pavel would like an M3 with this PR merged and I'll be away from a computer for a few days so I'm going to do the build now.

SystemFw:
No problem. I found myself without an internet connection yesterday 😞 
I'll review straight away.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -193,54 +221,53 @@ private[fs2] object Algebra {
...
     sealed trait R[X]
 
     def go[X](
         scope: CompileScope[F, O],
         stream: FreeC[Algebra[F, X, ?], Unit]
     ): F[R[X]] = {
-      F.flatMap(F.delay(stream.viewL.get)) {
-        case _: FreeC.Pure[Algebra[F, X, ?], Unit] =>
+      F.flatMap(F.map(F.delay(stream.viewL.get)) { a =>
+        //println(s"XXXA $scope : $a"); a
```
SystemFw: leftover

SystemFw: Actually the whole `map` can be eliminated I think?

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -282,115 +304,173 @@ private[fs2] object Algebra {
...
+      case _: FreeC.Result.Pure[Algebra[F, O, ?], Unit] =>
+        FreeC.interrupted(interruptedScope, interruptedError)
+      case failed: FreeC.Result.Fail[Algebra[F, O, ?], Unit] =>
+        Algebra.raiseError(
+          CompositeFailure
+            .fromList(interruptedError.toList :+ failed.error)
+            .getOrElse(failed.error))
+      case interrupted: FreeC.Result.Interrupted[Algebra[F, O, ?], _, Unit] =>
+        // impossible
+        FreeC.interrupted(interrupted.context, interrupted.deferredError)
```
SystemFw: Does this mean this case is unreachable? 
If so, wouldn't it be better to `sys.error` like in the other cases?

pchlupacek: Yes, but here we are able to recover ok. I mean there is no harm when compiler wll reach this state although I believe it is not possible to enter it. 

Hence therefore, we just do not fail here.

SystemFw: I know that it's harmless, I just find it a bit confusing (seeing code forces me to think about that branch, even though it's impossible to get there). Anyway I leave this up to you :)

pchlupacek:
@SystemFw I went through the comments on CompileScope, but did not find anything what striked me as completely wrong. Can you please guide me ?

SystemFw:
Ah, I didn't mean completely wrong, just a bit out of date, sorry.
I'm referring to the scaladoc here, which still mention EC and describe types that have been changed (they still talk about nested Options). From here down:

https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/core/shared/src/main/scala/fs2/internal/CompileScope.scala#L473

pchlupacek:
Actually, I think all the comments still holds. Scope is used to signal interruption and still we need to hold status for interruption of pure evaluations, so all of this damn machinery is still unfortunately relevant. 

We just removed whenInterrupted case, that is no longer necessary, and thats where FreeC comes in play.

SystemFw:
I'm only referring to this bit

```scala/**
    * A context of interruption status. This is shared from the parent that was created as interruptible to all
    * its children. It assures consistent view of the interruption through the stack
    * @param concurrent   Concurrent, used to create interruption at Eval.
    * @param ec       Execution context used to create promise and ref, and interruption at Eval.
    * @param promise  Promise signalling once the interruption to the scopes. Only completed once.
    *                 If signalled with None, normal interruption is signalled. If signaled with Some(err) failure is signalled.
    * @param ref      When None, scope is not interrupted,
    *                 when Some(None) scope was interrupted, and shall continue with `whenInterrupted`
    *                 when Some(Some(err)) scope has to be terminated with supplied failure.
    * @param interruptRoot Id of the scope that is root of this interruption andis guaranteed to be a parent of this scope.
    *                      Once interrupted, this scope must be closed and `whenInterrupted` must be consulted to provide next step of evaluation.
    * @param cancelParent  Cancels listening on parent's interrupt.
    */
  final private[internal] case class InterruptContext[F[_], O](
      concurrent: Concurrent[F],
      deferred: Deferred[F, Either[Throwable, Token]],
      ref: Ref[F, Option[Either[Throwable, Token]]],
      interruptRoot: Token,
      cancelParent: F[Unit]
```

pchlupacek:
all of this really holds :-)

pchlupacek:
In fact the interruption _SIGNALLING_ is still the same. What is new is the way how we compute `next` state after interruption or better to say `recover` from the interruption. Thats where FreeC helps.

SystemFw:
Right, but Execution context is no longer there, `Ref` is not `Option[Option[Throwable]]` anymore, and same for `Deferred`. I think you are talking about general strategy whether I'm talking about very boring little details :joy:

Anyway I can do a PR to clean that up later

pchlupacek:
aha sorry

PR #1201, opened 2018-07-31
Remove `equals` introduced by case class

eikek:
This part in `Segment` (line 1393) compares a segment against the empty segment

```scala
        val step = self
          .stage(
            Depth(0),
            trampoline.defer,
            o => emits(Chunk.singleton(o)),
            os => emits(os),
            r => {
              if (result.isEmpty || result
                    .flatMap(_.toOption)
                    .fold(false)(_ == Segment.empty)) result = Some(Left(r));
              throw Done
            }
          )
          .value
```

There is a class `SingleChunk` implementing `Segment` that is defined as a case class. Thus it gets `equals` implicitely. In my case that lead to comparing to SingleChunk segments where one was not empty, calling `equals` on its chunks which dramatically affected performance. In my case the code that took >3s before runs in ~200ms with the `case` modifier removed.

### DIFF
--- a/core/shared/src/main/scala/fs2/Segment.scala
+++ b/core/shared/src/main/scala/fs2/Segment.scala
@@ -1383,7 +1383,7 @@ object Segment {
             r => {
               if (result.isEmpty || result
                     .flatMap(_.toOption)
-                    .fold(false)(_ == Segment.empty)) result = Some(Left(r));
+                    .fold(false)(_ eq Segment.empty)) result = Some(Left(r));
               throw Done
             }
           )


eikek:
Ah…, I didn't think about binary compat! Good that mima does :-) I can change the `==` to `eq` then? This doesn't affect the 1.0 right? (I can't find Segment there…)

mpilquist:
Yep, changing to `eq` should be fine. `Segment` still exists on series/1.0 though it's not used in the main stream interpreter anymore. It would be good to get the fix in to both branches though.

PR #1204, opened 2018-08-02
Introduced FreeC.ViewL type

pchlupacek:
@mpilquist, @SystemFw this is a slight modification to ViewL, that is now trait with own case class for Eval Type. I think this cleans pattern match a bit (no Eval case anymore). 

Also this removes need for viewL suspension in main interpreter, savings on 2 allocation through every step of the interpreter (Suspend + Bind of F).

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -4,7 +4,7 @@ import cats.~>
 import cats.effect.{Concurrent, Sync}
 import cats.implicits._
 import fs2._
-import fs2.internal.FreeC.Result
+import fs2.internal.FreeC.{Result, ViewL}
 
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -229,10 +229,7 @@ private[fs2] object Algebra {
         scope: CompileScope[F, O],
         stream: FreeC[Algebra[F, X, ?], Unit]
     ): F[R[X]] = {
-      F.flatMap(F.map(F.delay(stream.viewL.get)) { a =>
-        //println(s"XXXA $scope : $a"); a
-        a
-      }) {
+      stream.viewL match {
```
SystemFw: ah, I left a comment about this in the other PR but you fixed it here :)

PR #1213, opened 2018-08-10
Fixes memory leak of Interrupt 

pchlupacek:
@mpilquist, @SystemFW fixes potential memory leak of interruptWhen and also improves concurrently tests a bit.

### DIFF
--- a/core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
+++ b/core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
@@ -60,6 +60,9 @@ class ConcurrentlySpec extends Fs2Spec with EventuallySupport {
 
     "run finalizers of background stream and properly handle exception" in forAll {
       s: PureStream[Int] =>
+        val streamStarted = new java.util.concurrent.atomic.AtomicBoolean(false)
+
+        val Boom = new Err
         val prg = Stream
           .eval(Deferred[IO, Unit])
           .flatMap { halt =>
@@ -73,18 +76,21 @@ class ConcurrentlySpec extends Fs2Spec with EventuallySupport {
                 s.get
                   .covary[IO]
                   .concurrently(
-                    (Stream.eval_(IO.sleep(50.millis)) ++
+                    (Stream.eval_(IO(streamStarted.set(true))) ++
                       Stream
                         .eval_(halt.complete(())))
                       .onFinalize(
                         IO.sleep(100.millis) >>
-                          (if (b.get) IO.raiseError(new Err) else IO(()))
+                          (if (b.get) IO.raiseError(Boom) else IO(()))
                       ))
               }
               .interruptWhen(halt.get.attempt)
           }
 
-        an[Err] should be thrownBy runLog(prg)
+        val r = prg.compile.drain.attempt.unsafeRunSync()
+
+        if (streamStarted.get) r shouldBe Left(Boom)
+        else r shouldBe Right(())
 
     }
   }
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1066,7 +1066,9 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     Stream
       .getScope[F2]
       .flatMap { scope =>
-        Stream.eval(F2.start(haltOnSignal.flatMap(scope.interrupt))).flatMap(_ => this)
+        Stream.eval(F2.start(haltOnSignal.flatMap(scope.interrupt))).flatMap { fiber =>
+          this.onFinalize(fiber.cancel)
+        }
       }
       .interruptScope
 
--- a/core/jvm/src/test/scala/fs2/MergeJoinSpec.scala
+++ b/core/jvm/src/test/scala/fs2/MergeJoinSpec.scala
@@ -74,12 +74,13 @@ class MergeJoinSpec extends Fs2Spec {
             )
 
           val s: Stream[IO, Stream[IO, Int]] = bracketed.flatMap { b =>
-            val s1 = (Stream.eval_(IO.sleep(20.millis)) ++
-              Stream.eval_(halt.complete(())))
-              .onFinalize(
-                IO.sleep(100.millis) >>
-                  (if (b.get) IO.raiseError(new Err) else IO(()))
-              )
+            val s1 =
+              Stream
+                .eval_(halt.complete(()))
+                .onFinalize(
+                  IO.sleep(100.millis) >>
+                    (if (b.get) IO.raiseError(new Err) else IO(()))
+                )
 
             Stream(s1, s2.get.covary[IO]).covary[IO]
           }
@@ -142,9 +143,8 @@ class MergeJoinSpec extends Fs2Spec {
               s.get
                 .covary[IO]
                 .merge(
-                  (Stream.eval_(IO.sleep(20.millis)) ++
-                    Stream
-                      .eval(halt.complete(())))
+                  Stream
+                    .eval(halt.complete(()))
                     .onFinalize(
                       IO.sleep(100.millis) >>
                         (if (b.get) IO.raiseError(new Err) else IO(()))


pchlupacek:
@SystemFw I will be coming with more fixes here. It seems we still have issue with correct sequencing of finalizers, that we are trying to address. They are very corner case, but still may cause some program to enter invalid state and may be very hard to debug.

so please hold on with merge for now

SystemFw:
@pchlupacek  will do :)

mpilquist:
I really don't like the special casing of onFinalize -- `bracket(unit)(finalizer)` must have the same behavior as `onFinalize(finalizer)`

pchlupacek:
@mpilquist this is just now where we are trying to understand issue deeper
There is definitelly issue with sequencing finalizers due to asynchronous code allocation of Acquire, which i am not sure currently how to fix. But, I don't either think that it is correct to have finaizers to be executed in random order

pchlupacek:
there is still something with parJoin tests, but still, the issue seems to manifests itself like race in finaizer evaluation due orchestration of current Scope#acquireResource implementation which is correct for asynchronous resources, but for finalizers it introduces non-determinism.

pchlupacek:
Also, the fix of `cancellation` is unrelated to issues in finalisers, it just manifests them more often.

mpilquist:
There's no difference between a finalizer and a resource. It's not possible to be correct for resources but not for finalizers. I'm happy to wait until there's an overall resolution but I think any solution which specializes `s.onFinalize(f)` beyond `Stream.bracket(unit)(f).flatMap(_ => s)` is incorrect.

pchlupacek:
Yeah, I hear you. However I am not 100% yet convinced that this is correct. If we get current state of affairs then: 
a) we allow for asynchronous resource allocation (Yes)
b) we are correct guaranteeing sequencing of finalizers (Yes)
currently a) and b) seems to be mutually exclusive

pchlupacek:
Also to be clear I am quite convinced there is better solution than special case, this was just something to prove the root cause. So no worries....

pchlupacek:
@mpilquist consider this:
currently the resource allocation consist of three phases : 
- 1. registration to the scope
- 2. acquisition 
- 3. cleanup 

In normal synchronous `F` you would assume that (1, 2) are provided as one step and nothing can get `in between`. Also there is a `catch` that cleanup may be actually invoked in two phases, 
(a) when the stream finishes through normal cleanup procedure of the scope 
(b) as a result of acquisition, that finished _after_ scope performed its cleanup.

In fs2 implementation when we register the resource we include `F.unit` as cleanup first, and only after the resource is acquired we either register the cleanup or just run it immediately, when resource  shall be released due scope being closed. 

This technique is completely ok for resources, where you really cannot do too much else. But hence we add to finalizer this asynchronous behaviour (by defining it as Bracket) we sort of introducing nondeterminism to something which is completely deterministic. 

So more and more I am digging into the problem, I am more and more convinced that defining the `onFinalize` in form of bracket is incorrect, and it shall have its own algebra constructor. 

Having said that I am still struggling to fix this cancellation issue. I found couple of places in code that likely need improvement, but still did not find the root of the cause why the finalisers are acting so weird. It seems this is combination of multiple issues and I need to identify all of them.

pchlupacek:
@mpilquist just a quick update on this one. Waiting for next release of the cats (RC3) and then I am gonna revisit this one. I am still solid on grounds re finaliser, but lets see how that goes. I think the concurrence operators will need to be improved a bit with cats changes.

mpilquist:
@pchlupacek FYI, series/1.0 now has the cats-effect RC3 changes, using a preview build of RC3.

pchlupacek:
@mpilquist @SystemFw this seems to be finally done. 

Couple of notes : 

- concurrently, merge, interrupt, parJoin all have been fixed to properly handle early terminations, finalizer resource propagation etc. 
- .onFinalize is again defined as `bracket`. However I have a feeling that using it may lead to hard tracking errors (i.e. ones this PR is fixing), and in fact, we may consider to remove it allTogether in favor of bracket-only syntax. 
- couple of minor fixes in algebra, compile scope et all. 

If you ok with this, it could me squashed and merged to get in in M4 tomorrow.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -353,41 +350,31 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
-                // in case the interruptL is signaled before the `L` stream may be in
-                // its `append` code, that requires `get` to complete, which won't ever complete,
-                // b/c it will be evaluated after `interruptL`
-                doneR.complete(r) >>
-                  r.fold(F.unit)(_ => interruptL.complete(()))
+    Stream.eval {
+      Deferred[F2, Unit].flatMap { interrupt =>
+        Deferred[F2, Either[Throwable, Unit]].map { doneR =>
+          def runR: F2[Unit] =
+            that.interruptWhen(interrupt.get.attempt).compile.drain.attempt.flatMap { r =>
```
mpilquist: Instread of `.drain.attempt.flatMap`, should this be `.drain.guaranteeCase`? If the `F[_]` is canceled at effect level (not stream level), you still want to run the body of the `flatMap`, right?

core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
```diff
@@ -60,31 +60,56 @@ class ConcurrentlySpec extends Fs2Spec with EventuallySupport {
...
+                    Stream.eval_(halt.complete(())) // immediately interrupt the outer stream
+
+                val prg0 =
+                  bracketed
+                    .flatMap { b =>
+                      s.get.covary[IO].concurrently(runner)
+                    }
+                    .interruptWhen(halt.get.attempt)
+
+                prg0.compile.drain.attempt.flatMap { r =>
```
mpilquist: Same comment here -- in general, most `.attempt.flatMap` calls should be replaced by `.guaranteeCase` or similar method from `Bracket[F]`.

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: the `F` cannot be cancelled here. Note the user does not have access to `F` (due to `F.start(runR)`) construct. 

I think whole cancellation story doesn't really play well in fs2, hence the finalisers may not kick in if you cancel. I think we have to make clear distinction between interruption (what we do in fs2, that is also recoverable) and cancellation, that is on `F` and is not recoverable in stream scope. 

It sad, I was hoping we would be able to use cancellation instead of interruption in merge, but that seems to be not possible, or at least I don't know how.

core/jvm/src/test/scala/fs2/ConcurrentlySpec.scala
pchlupacek: Yeah we could. I sort of used this patter, as this won't give us anything b/c we don't have special case for `cancel`. So was like sort of resistant to introduce another technique if we could live with attempt.flatMap.

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: > I think whole cancellation story doesn't really play well in fs2, hence the finalisers may not kick in if you cancel. I think we have to make clear distinction between interruption (what we do in fs2, that is also recoverable) and cancellation, that is on F and is not recoverable in stream scope.

Just to be clear, cancelling a compiled stream still guarantees all resources are finalized correctly.

pchlupacek: Well, I am not sure of that. I think cancellation may or may not run the finalisers. 

What do you think proves that?

pchlupacek: Ok I take it back. In Algebra#compile there is bracket, so yes, indeed the root scope will get finalised always even when `F` is cancelled, and now, thanks to new `cancel` behaviour user may decide to wait/not on finalisers to complete. 

 I may take another shot after this PR to further optimize `merge` and `concurrently` with this.

PR #1215, opened 2018-08-12
Fix bug with .pull.take computing wrong remainder

SystemFw:
Small but nasty bug :)

We'll probably need an M4 soon, fs2-reactive-streams relies on this op and broke on M2-M3


mpilquist:
Ouch

PR #1218, opened 2018-08-14
Add switchMap

tom91136:
So this is my attempt at implementing a fully working `switchMap` which should be equivalent to the Monix [switchMap](https://monix.io/api/3.0/monix/reactive/Observable.html#switchMap[B](f:A=%3Emonix.reactive.Observable[B]):monix.reactive.Observable[B]) and also semantically similar to the [switch](http://reactivex.io/documentation/operators/switch.html) operator in ReactiveX. 
To summarise, a `switchMap` is like `flatMap`(the signatures are identical) but interrupts the inner stream when new elements arrive in the outer stream.
This is useful for cases where downstream is working on time sensitive data and requires the latest element. This is also potentially useful for cases where the inner stream is infinite.

I've chosen what I think is the most sensible behaviours for finalisers and failures. The following are copied from the Scaladoc:
 * Finializers of each inner stream are guaranteed to run before the next inner stream starts.
 * When the outer stream stops gracefully, the currently running inner stream will continue to run.
 * When either the inner or outer stream fails, the entire streams fails.

There was some mention of needing a switchMap in #1139 and in the gitter channel so hopefully this resolves those.


SystemFw:
> When the outer stream stops gracefully, the currently running inner stream will continue to run.

I'm not entirely sure about this. I also have the dual question:
What happens if the inner stream stops gracefully or it's interrupted?

tom91136:
Cheers! I sort of experimented with a version where the inner stream terminates when the outer one terminates. I dropped the experiment because in the case of `parJoin`, the inner streams will continue to run so I thought I would match that. TBH I'm not too sure which behaviour makes the most sense so please let me know what you think.
For the second question, if the inner stream terminates/interrupts then nothing really happens until the next element arrives in the outer stream(the outer stream holds the stream open during this time or else the stream terminates).  
Let me double check my tests re the second question just to be sure.

tom91136:
Yep, the behaviour I described for when an inner stream terminates is correct.

SystemFw:
Upon further thought, I think the current semantics wrt inner stream termination are more desirable, so forget my question. Still haven't had time to go through the implementation details fully though.

tom91136:
No worries, I'm already using this as a `Pipe` locally so no real rush. I'm still very new to the cats ecosystem so I'm sure this will need more commits.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1237,6 +1237,57 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+                    guard.withPermit(queue.enqueue1(Left(e).some))
+                case Right(x) =>
+                  Deferred[F2, Unit].flatMap { halt =>
+                    haltRef.modify {
+                      case None       => halt.some -> F2.unit
+                      case Some(last) => halt.some -> last.complete(())
+                    }.flatten *> pulse.enqueue1((x -> halt).some)
+                  }
+              }))
+    } yield o2
```
mpilquist: Minor optimization: the final `yield o2` will result in a `map(o2 => o2)` call, which will then result in chunk copying (each chunk will have `.map(o2 => o2)` called on it). To fix, we could manually expand the for-comprehension.

SystemFw: probably true, but I tend not to use `for` with streams anyway, because you end needing a lot of methods on stream (e.g. `evalMap`, `concurrently`, `to`, `take`) and so on, and with explicit `flatMap` is easier to mix them in that with `for`

pchlupacek:
@mpilquist I leaved two minor comments. I will need a bit more time to make full review. Just as side remark would be excellent to name Streams with def, to understand better whats going on like 

def runInner = ???
def runOuter = ??? 

Also will be excellent if the interruption logic will somehow be briefly explained.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1244,61 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+                    .flatMap(Stream.chunk(_))
+                    .concurrently(pulse.dequeue.unNoneTerminate
+                      .onFinalize(queue.enqueue1(None))
+                      .evalMap {
+                        case (o, haltInner) =>
+                          guard.withPermit(
+                            f(o).chunks
+                              .evalMap(x => queue.enqueue1(Right(x).some))
+                              .interruptWhen(haltInner.get.attempt)
+                              .interruptWhen(haltOuter.get.attempt)
```
SystemFw: I'm not sure `.interruptWhen(haltOuter.get.attempt` is necessary. 
As far as I can see it's only completed in `handleError` of this same stream, but if there's an error this stream will already be terminated

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1244,61 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    *
+    * When either the inner or outer stream fails, the entire streams fails.
+    *
+    */
+  def switchMap[F2[x] >: F[x], O2](f: O => Stream[F2, O2])(
+      implicit F2: Concurrent[F2]): Stream[F2, O2] =
+    Stream.eval(Ref.of[F2, Option[Deferred[F2, Unit]]](None)).flatMap { haltRef =>
+      Stream.eval(Deferred[F2, Unit]).flatMap { haltOuter =>
+        Stream.eval(async.unboundedQueue[F2, Option[Either[Throwable, Chunk[O2]]]]).flatMap {
+          queue =>
```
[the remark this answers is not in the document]
tom91136: Yeah, I was trying to match the style of the first 7 lines in `merge`, probably took it too far.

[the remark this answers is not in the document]
SystemFw: sorry, I don't mean these, these are fine. I mean the ones below, the actual streams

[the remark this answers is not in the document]
SystemFw:
```scala
 queue.dequeue.unNoneTerminate.rethrow
                     .flatMap(Stream.chunk(_))
                     .concurrently
```

```scala
pulse.dequeue.unNoneTerminate
                       .onFinalize(queue.enqueue1(None))
                       .evalMap {...}.merge
```
```scala
this.attempt
                         .interruptWhen(haltOuter.get.attempt)
                         .onFinalize(pulse.enqueue1(None))
                         .evalMap {
                           case Left(e) =>
                             haltRef.modify(x => None -> x.fold(F2.unit)(_.complete(()))).flatten *>
                               guard.withPermit(queue.enqueue1(Left(e).some))
                           case Right(x) =>
                             Deferred[F2, Unit].flatMap(halt =>
                               haltRef.modify {
                                 case None       => halt.some -> F2.unit
                                 case Some(last) => halt.some -> last.complete(())
                               }.flatten *> pulse.enqueue1((x -> halt).some))
                         }))
```

```scala
case (o, haltInner) =>
                           guard.withPermit(
                             f(o).chunks
                               .evalMap(x => queue.enqueue1(Right(x).some))
                               .interruptWhen(haltInner.get.attempt)
                               .interruptWhen(haltOuter.get.attempt)
                               .compile
                               .drain
                               .handleErrorWith(e =>
                                 haltOuter.complete(()) *> queue.enqueue1(Left(e).some)))
```

at least some of these should be named, ideally you will end up with something like
```scala
foo.concurrently(bar merge baz)
```

Rough guidelines ofc, use your judgement. At the moment it's a bit hard to follow the logic completely (although tbf this is a tricky combinator). Same goes for `queue` and `pulse`, the names don't tell much, is `queue` something like `outQ` (so the final output stream).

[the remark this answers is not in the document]
tom91136: Done. Yeah, my initial implementation on gitter it way too optimistic, concurrency is hard.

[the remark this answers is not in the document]
SystemFw: Speaking of which, what's the role of the Semaphore here?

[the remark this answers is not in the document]
tom91136: Basically, the idea is that it prevents the downstream from getting multiple errors as both the inner and outer can terminate the downstream. It seems redundant now because there's a rethrow at downstream and after some testing it seems to work fine without it.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1245,74 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    *
+    * When the outer stream stops gracefully, the currently running inner stream will continue to run.
+    *
+    * When an inner stream terminates/interrupts, nothing happens until the next element arrives
+    * in the outer stream(i.e the outer stream holds the stream open during this time or else the
+    * stream terminates)
+    *
+    * When either the inner or outer stream fails, the entire streams fails.
+    *
+    */
```
SystemFw: Ok the logic is really starting to appear now :)
That makes it a lot easier to review, and I hope you don't mind all the nitpicking, but this is fairly tricky so I'd like to get it in the best shape possible.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1245,74 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+  def switchMap[F2[x] >: F[x], O2](f: O => Stream[F2, O2])(
+      implicit F2: Concurrent[F2]): Stream[F2, O2] =
+    Stream.eval(Ref.of[F2, Option[Deferred[F2, Unit]]](None)).flatMap { haltRef =>
+      Stream.eval(Deferred[F2, Unit]).flatMap { haltOuter =>
+        // consumes the upstream by sending a packet(stream and a ref) to the next queue and
+        // also does error handling/inturruption/switching by terminating the currently running
+        // inner stream
+        def runUpstream(next: Queue[F2, Option[(O, Deferred[F2, Unit])]],
+                        downstream: Queue[F2, Option[Either[Throwable, Chunk[O2]]]]) =
```
SystemFw: This doesn't really need to be a function, it's ok to just have `runUpstream` close over `next` and `downstream` like you're doing with `haltRef` and `haltOuter`

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1245,74 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+          this.attempt
+            .interruptWhen(haltOuter.get.attempt)
+            .onFinalize(next.enqueue1(None))
+            .evalMap {
```
SystemFw: I feel like `attempt....evalMap { case Left(e)` is best put in an `handleError` with at the end, so it's easier to focus on the `evalMap` logic in the happy path (the way it is now it's not immediately clear that the `Left` case can only happen once and not per element/chunk)

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1245,74 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+                  }.flatten *> next.enqueue1((x -> halt).some))
+            }
+
+        // consumes the packets produced by runUpstream and drains the content to downstream
+        // upon error/interruption, the downstream will also be signalled for termination
+        def runInner(next: Queue[F2, Option[(O, Deferred[F2, Unit])]],
+                     downstream: Queue[F2, Option[Either[Throwable, Chunk[O2]]]]) =
+          next.dequeue.unNoneTerminate
+            .onFinalize(downstream.enqueue1(None))
+            .evalMap {
```
SystemFw: this can go back to being a `flatMap` without `compile.drain` I think

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1245,74 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+              case (o, haltInner) =>
+                f(o).chunks
+                  .evalMap(x => downstream.enqueue1(Right(x).some))
+                  .interruptWhen(haltInner.get.attempt)
+                  .compile
+                  .drain
+                  .handleErrorWith(e => haltOuter.complete(()) *> downstream.enqueue1(Left(e).some))
```
SystemFw: I'm wondering if this logic in `handleErrorWith` is necessary at all (not too sure, maybe it is).

If this stream fails, `runUpstream` will fail as well (it's in a `merge`), and `downstream` will fail as well (it's in `concurrently`). So is this related to the order of finalisers? I'm actually not super sure what's guaranteeing the order of finalisers here

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1245,74 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+            }
+
+        val outQueue = Stream.eval(async.unboundedQueue[F2, Option[Either[Throwable, Chunk[O2]]]])
+        val nextQueue = Stream.eval(async.unboundedQueue[F2, Option[(O, Deferred[F2, Unit])]])
+
```
SystemFw: These two can go back at the top, and always be in scope through a `flatMap`, they weren't really a problem for clarity, the lack of `runInner` and `runOuter` was, and you can save some boilerplate

core/shared/src/main/scala/fs2/Stream.scala
tom91136: Agree. The main reason I didn't use `handleErrorWith` was because it needs a `Throwable => Stream[F2, O2]` and it just so turns out that `downstream.enqueue1(Left(e).some)` is a `F2[_]` so I thought that fits better in an `evalMap`.

tom91136: Speaking of which, will there ever be a `handleErrorWith` that takes `Throwable => F2[O2]`?

core/shared/src/main/scala/fs2/Stream.scala
tom91136: Done.

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: it'll have to be a specific Stream method cause it can't fit into `MonadError`. Otoh error handling might not be needed at all here (see my comment below)

core/shared/src/main/scala/fs2/Stream.scala
tom91136: After replacing `evalMap` and lifting the error to a Stream I'm getting one test failure: `when inner stream fails, overall stream fails`. The reason was because no exception was thrown. Investigating now.

tom91136: Ah! I accidentally wrote `Stream( * : F2[Unit])`, should be `Stream.eval( * : F2[Unit])`. Damn, I had similar bugs like this when working with `merge` and `concurrently`.

tom91136: Ok, there was an actual bug in there where a `None` would get enqueued before an failure causing downstream to terminate too early. I guess that was why the semaphore was needed previously. I've rearranged that logic such that failures will always get enqueued first.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1244,66 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+                    f(o).chunks
+                      .evalMap(x => downstream.enqueue1(Right(x).some))
+                      .interruptWhen(haltInner.get.attempt)
+                      .handleErrorWith(e => Stream.eval(downstream.enqueue1(Left(e).some)))
+                }
+
+            // continuously dequeue the downstream while we process the upstream
+            downstream.dequeue.unNoneTerminate.rethrow
+              .flatMap(Stream.chunk(_))
+              .concurrently(runInner.merge(runUpstream))
```
tom91136: So something like `concurrently(runInner.concurrently(runUpstream))`? Or do we want a parJoin of some sort?

SystemFw: I'd try with the `concurrently(... concurrently`, no queues are allocated. `parJoin` allocates one
Do test it though 😄

SystemFw:
I think something along these lines will work (no `handleError`), although not tested it

```scala
  def switchMap[F[x] :> F2[x]: Concurrent, O2](f: O => Stream[F2, O2]): Stream[F2, O2] =
    Stream.force {
      for {
        downstream <- async.unboundedQueue[F2, Option[Chunk[O2]]]
        next <- async.unboundedQueue[F2, Option[(O, Deferred[F2, Unit])]]
        halt <- Ref.of[F2, Option[Deferred[F2, Unit]]](None)
      } yield {

        def runUpstream =
          this
            .onFinalize(next.enqueue1(None))
            .evalMap { nextElem =>
              for {
                haltNext <-  Deferred[F2, Unit]
                haltCurrent <- halt.modify(c => haltNext.some -> c)
                _ <- haltCurrent.traverse(_.complete(()))
                _ <- next enqueue1 (nextElem -> haltNext).some
              } yield ()
            }


        def runInner =
          next.dequeue.unNoneTerminate
            .onFinalize(downstream.enqueue1(None))
            .flatMap {
              case (o, haltInner) =>
               f(o).chunks
                 .evalMap(x => downstream enqueue1 x.some)
                 .interruptWhen(haltInner.get.attempt)
          }

        downstream.dequeue.unNoneTerminate
          .flatMap(Stream.chunk(_))
          .concurrently(runInner concurrently runUpstream)
      }
    }
```

tom91136:
There was one thing left unclear up till now: the order of finalisers for when either stream fails. The current behaviour is that the outer one runs before the inner one in both cases. Before I document this, do you think this makes sense?

mpilquist:
Inner should run first, because the inner stream might be using a resource acquired in the outer stream.

tom91136:
Cool, will do that. The test looks quite funny for this, something along the lines of:

SystemFw:
Lol ofc you can. Do test it though cause I haven't. Also we should probably get the better-monadic-for plugin

tom91136:
Ok, got most things resolved. I've left the for-comp. for later because I just about ran out of ideas with handing finaliser orders upon failure. Most solutions I can think of goes back to having complicated logic in error handlers. Might need some help on this.

SystemFw:
Yeah, it's not excluded that some complexity will get back once this aspect is handled, but the underlying logic is now very clear (imho) so I feel more confident reasoning about it. Gtg sleep for tonight, but I'll think about the finalisation logic. Thanks for all the work so far!

tom91136:
So I'm quite busy from now till 31/8 so I'll probably resume work after that. Hopefully coming back with a solution for finaliser orders as well.

tom91136:
Apologies, turns out writing parser combinators in Java sucked but I'm back working on this now. I'm getting some progress with scoping via ` Stream.getScope[F2]`, will experiment on it some more to see if it passes the tests. Is there a chance for this to be included in `1.0.0` after this is done?

mpilquist:
@tom91136 No worries. Definitely would like to get this in to 1.0.0 if possible.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1244,6 +1244,62 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    * in the outer stream(i.e the outer stream holds the stream open during this time or else the
+    * stream terminates)
+    *
+    * When either the inner or outer stream fails, the entire stream fails and the finalizer of the
+	* inner stream runs before the outer one.
+    *
+    */
+  def switchMap[F2[x] >: F[x], O2](f: O => Stream[F2, O2])(
+      implicit F2: Concurrent[F2]): Stream[F2, O2] =
+    Stream.eval(async.unboundedQueue[F2, Option[Chunk[O2]]]).flatMap { downstream =>
```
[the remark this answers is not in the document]
tom91136: Done

tom91136:
So I'm still stuck on how to implement the required order for finaliser on failure.  @SystemFw gave me a few options on gitter:
> 1 don't do anything, and see if #1213 fixes it
> 2 Rather than delaying the stream finaliser, delay the stream from finishing, i.e. use Deferreds as latches to enforce happens-before relationships
> 3 Last resort, manual scope manipulation like you were attempting. I do feel like this should be tried last though, if all else has failed

@AdamChlupacek also mentions #1213 and that there seems to be an issue with the ordering of finalisers.

With the current design, I couldn't come up with a solution using the suggested option 2 and option 3 still has the issue where `lease.cancel` inside a concurrent does the wrong thing.

So with all that, I decided to give this another go from scratch. This version is much simpler, passes all the tests, and seems to work correctly. The core idea is similar but this time I'm using `parJoin`.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1311,6 +1311,56 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+                .handleErrorWith { e =>
+                  Stream.eval(innerFailed.set(true)) >> Stream.raiseError[F2](e)
+                }
+                // upon failure, do not release the guard(prevents the next stream from running)
+                .onFinalize(innerFailed.get.flatMap { if (_) F2.unit else guard.release })
+
+          this
+          // although failure on inner propagates, it is still possible for parJoin to
+          // keep pulling before it fails so we interrupt it anyway
+            .interruptWhen(innerFailed)
```
SystemFw: I'm curious if you have seen any failing tests because of this. Even with that, it's still possible for `this` to keep pulling before being interrupted, and the mechanism for error propagation in `join` is similar to this anyway, so I'm wondering if this adds anything or not, especially given that the `Semaphore` should be guarding things regardless.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1311,6 +1311,56 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+  def switchMap[F2[x] >: F[x], O2](f: O => Stream[F2, O2])(
+      implicit F2: Concurrent[F2]): Stream[F2, O2] =
+    Stream.eval(Semaphore[F2](1)).flatMap { guard =>
+      Stream.eval(Ref.of[F2, Option[Deferred[F2, Unit]]](None)).flatMap { haltRef =>
+        Stream.eval(SignallingRef.apply(false)).flatMap { innerFailed =>
+          def runInner(o: O, halt: Deferred[F2, Unit]) =
+            Stream.eval(guard.acquire) >> // guard inner to prevent parallel inner streams
+              f(o)
+                .interruptWhen(halt.get.attempt)
+                .handleErrorWith { e =>
```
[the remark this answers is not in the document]
tom91136: Done

core/shared/src/main/scala/fs2/Stream.scala
tom91136: Investigating, currently leaving ` .interruptWhen(innerFailed)` out causes tests to fail.

SystemFw: It's quite possible that I'm wrong and you need that. I'm only worried that `interruptWhen` makes the failure _less likely_ rather than solving it, so I'd like to understand what's going on

tom91136: Ok, this seems more like a test issue now.  The ordering of finalisers are correct after manually inspecting the test output for several failed runs. The test was flakey because the `verdict` checking (which is also a stream) ran before the outer finaliser had a chance to record it. Working on a fix now.

SystemFw:
Test failures are due to a stupid whitespace error I made in the doctests for an unrelated PR

pchlupacek:
@SystemFw ping me just when this is ready for final review ok?

SystemFw:
Yeah, couple final comments and then I think it's going to be ready

core/shared/src/main/scala/fs2/Stream.scala
tom91136: Done

SystemFw: Cool. I think `innerFailed` can be a normal `Ref` now (more lightweight), and eventually it can go away entirely once `bracketCase` is in (#1269 )

SystemFw:
@pchlupacek @mpilquist I think this is ready for a final review , keeping [this](https://github.com/functional-streams-for-scala/fs2/pull/1218#discussion_r219983659) comment in mind

mpilquist:
Awesome, looks great. I added `onFinalizeCase` in #1269 btw.

SystemFw:
`switchMap` is a rather standard name for this combinator in several libraries, probably better to use a familiar name. The rationale is that `flatMap` is `map` then `flatten` (concatenate all inner streams), whereas `switchMap` is `map` then `switch` (inner streams interrupt each other)

SystemFw:
@tom91136 I think that after rebasing and using `onFinalizeCase` to replace that `SignallingRef`, this would be ready for merge. Thanks for your work, this is an impressive first contribution :)

SystemFw:
In this specific combinator you need to `release` on completion and interruption, but not on error, so if you do ` ++ Stream.eval(release)`, you won't release on interruption (so deadlock, in this case), and if you do `onFinalize(release)`, you will release _also_ when errors happen, which is wrong (and causes a race).
Hence the `Ref` to signal when things have failed, that you consult in `onFinalize`.
With this new addition, you can do
```scala
.onFinalizeCase {
  case Completed => release
   case Interrupted => release
  case Error => F.unit
}
```

pchlupacek:
Just to clarify `s.interrupt(byThat) ++ release`, will call `release` when `byThat` will signal interruption. Essentially `append` cancels interruption boundary. So I am still convinced ++ is enough there.

SystemFw:
ah, you're right actually! Got confused.

Even better then, just `++` is enough  :)

Btw I don't see your comments, is it possible that they're stuck in review stage?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1311,6 +1311,51 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+                // upon failure, do not release the guard(prevents the next stream from running)
+                .onFinalize(innerFailed.get.flatMap { if (_) F2.unit else guard.release })
+
+          this
+            .evalMap { o =>
+              Deferred[F2, Unit].flatMap { halt =>
+                haltRef.modify {
+                  case None       => halt.some -> F2.unit
+                  case Some(last) => halt.some -> last.complete(()) // interrupt the previous one
+                }.flatten *> F2.pure(runInner(o, halt))
```
SystemFw: `.flatten *> F2.pure(runInner(o, halt))` should be equivalent

`.flatten.as(runInner(o, halt))`

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1311,6 +1311,51 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+  def switchMap[F2[x] >: F[x], O2](f: O => Stream[F2, O2])(
+      implicit F2: Concurrent[F2]): Stream[F2, O2] =
+    Stream.eval(Semaphore[F2](1)).flatMap { guard =>
+      Stream.eval(Ref.of[F2, Option[Deferred[F2, Unit]]](None)).flatMap { haltRef =>
+        Stream.eval(SignallingRef.apply(false)).flatMap { innerFailed =>
+          def runInner(o: O, halt: Deferred[F2, Unit]) =
+            Stream.eval(guard.acquire) >> // guard inner to prevent parallel inner streams
+              f(o)
+                .interruptWhen(halt.get.attempt)
+                .onError { case e => Stream.eval(innerFailed.set(true)) }
```
SystemFw: like @pchlupacek said, you can get rid of the `SignallingRef` and just do `++ Stream.eval(guard.release)`

pchlupacek:
@SystemFw damn I am still confused with need of submitting review

core/shared/src/main/scala/fs2/Stream.scala
tom91136: If I understand correctly, we can just do `Stream.eval(guard.acquire) >> f(o).interruptWhen(halt.get.attempt) ++ Stream.eval(guard.release)` but how do I get the return type to match `Stream[F2, O2]`?

SystemFw: use `Stream.eval_`, which is basically `Stream.eval(..).drain`

SystemFw: `drain` gets from `Stream[F, A]` to `Stream[F, Nothing]`, which makes things compile since `Nothing` is a subtype of everything. On a logical level, `drain` evaluates the Stream but doesn't emit anything, hence why `Stream[F, Nothing]`

pchlupacek:
@SystemFw ok I finally managed to hit that submit button for review. It just makes me cry :-)

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1311,6 +1311,51 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    * in the outer stream(i.e the outer stream holds the stream open during this time or else the
+    * stream terminates)
+    *
+    * When either the inner or outer stream fails, the entire stream fails and the finalizer of the
+	* inner stream runs before the outer one.
+    *
+    */
+  def switchMap[F2[x] >: F[x], O2](f: O => Stream[F2, O2])(
+      implicit F2: Concurrent[F2]): Stream[F2, O2] =
+    Stream.eval(Semaphore[F2](1)).flatMap { guard =>
```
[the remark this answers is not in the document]
SystemFw: that's `Stream.force` btw (`Stream.eval(...).flatten)`)

core/shared/src/main/scala/fs2/Stream.scala
tom91136: Ah right, perfect, tests are still passing, I'm just gonna check everything again and rebase anyway

core/shared/src/main/scala/fs2/Stream.scala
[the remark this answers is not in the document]
pchlupacek: cool :-)

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1311,6 +1311,51 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+              f(o)
+                .interruptWhen(halt.get.attempt)
+                .onError { case e => Stream.eval(innerFailed.set(true)) }
+                // upon failure, do not release the guard(prevents the next stream from running)
+                .onFinalize(innerFailed.get.flatMap { if (_) F2.unit else guard.release })
+
+          this
+            .evalMap { o =>
+              Deferred[F2, Unit].flatMap { halt =>
+                haltRef.modify {
```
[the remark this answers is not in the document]
SystemFw: Not to be a contrarian, but I find the explicit `case` to be a bit clearer for this case, if a bit longer. Don't really mind either way though :)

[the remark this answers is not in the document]
tom91136: How about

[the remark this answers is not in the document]
pchlupacek: :-) yeah its just ok, the `case` is not that importantas such here, is more `getAndSet` vs `modify`.  We are not in hot spot here so the additional cost of modify (1xGET, 1xCAS vs 1 GAS) doesn't really play significant role.

core/shared/src/main/scala/fs2/Stream.scala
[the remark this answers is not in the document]
tom91136: Could you give me an example here, we have 3x `F2` things, how does `Stream.force` help?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1311,6 +1311,51 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    Stream.eval(Semaphore[F2](1)).flatMap { guard =>
+      Stream.eval(Ref.of[F2, Option[Deferred[F2, Unit]]](None)).flatMap { haltRef =>
+        Stream.eval(SignallingRef.apply(false)).flatMap { innerFailed =>
+          def runInner(o: O, halt: Deferred[F2, Unit]) =
+            Stream.eval(guard.acquire) >> // guard inner to prevent parallel inner streams
+              f(o)
+                .interruptWhen(halt.get.attempt)
+                .onError { case e => Stream.eval(innerFailed.set(true)) }
+                // upon failure, do not release the guard(prevents the next stream from running)
+                .onFinalize(innerFailed.get.flatMap { if (_) F2.unit else guard.release })
```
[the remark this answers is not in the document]
tom91136: Done, this combinator is becoming really short

core/shared/src/main/scala/fs2/Stream.scala
tom91136: Done, combined with the `getAndSet` suggested by @pchlupacek

core/shared/src/main/scala/fs2/Stream.scala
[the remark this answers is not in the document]
tom91136: Nevermind, I see how this works now

core/shared/src/main/scala/fs2/Stream.scala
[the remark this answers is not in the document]
SystemFw: heh, the git history is going to be pretty cool once you see the starting and finishing points :)

pchlupacek:
ok, lgtm :-)

tom91136:
Yay! A big thanks to everyone for reviewing this and @SystemFw for holding my hand throughout this feature. Cheers!

PR #1221, opened 2018-08-16
Upgraded to latest cats-effect hash release

mpilquist:
This includes support for `ContextShift` (aka, the cats-effect version of `Linebacker`).

Note: in a separate PR, after this is merged, I am planning on getting rid of the synchronous/blocking versions of various `io` streams. E.g., deleting `fs2.io.stdout` and renaming `fs2.io.stdoutAsync` to `stdout`. The synchronous ones are dangerous and if we fix that by adding a blockingEC + `ContextShift[F]`, we've made them redundant with the async ones.

Similarly, we might want to do something similar with the `fs2.io.file` package, though there I can see keeping the synchronous ones, changing them to take the blocking ec + `ContextShift[F]`, and leaving the async ones use java.nio async file handle stuff.

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/ConcurrentBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/ConcurrentBenchmark.scala
@@ -1,16 +1,19 @@
 package fs2
 package benchmark
 
-import scala.concurrent.ExecutionContext.Implicits.global
-import cats.effect.IO
+import cats.effect.{ContextShift, IO}
 import org.openjdk.jmh.annotations.{Benchmark, Scope, State}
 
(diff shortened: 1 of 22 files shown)


SystemFw:
There is a compilation error here
https://github.com/functional-streams-for-scala/fs2/blob/a341857f4f0b09675e635fa0d34fa07a347b851b/benchmark/src/main/scala/fs2/benchmark/ConcurrentBenchmark.scala

about not finding `Concurrent[F2]`. I suspect it has to do with there not being an implicit `Timer` anymore.

mpilquist:
Thanks, I routinely forget to compile the benchmarks project. Fix incoming.

PR #1222, opened 2018-08-16
add Stream.fromEither

bpholt:
These changes were backported from the series/1.0 PR https://github.com/functional-streams-for-scala/fs2/pull/1220

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -5,11 +5,11 @@ import cats.data.NonEmptyList
 import scala.collection.generic.CanBuildFrom
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
-import cats.{Applicative, Eq, Functor, Id, Monoid, Semigroup, ~>}
-import cats.effect.{Effect, IO, Sync}
+import cats._
+import cats.effect._
 import cats.implicits.{catsSyntaxEither => _, _}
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1393,6 +1393,19 @@ object Stream {
...
+  class PartiallyAppliedFromEither[F[_]] {
+    def apply[A](either: Either[Throwable, A])(
+        implicit ev: ApplicativeError[F, Throwable]): Stream[F, A] = {
+      val _ = ev
+      either.fold(Stream.raiseError[A], Stream.emit(_))
+    }
+  }
+
+  /**
+    * Lifts an Either[Throwable, A] to an effectful Stream
```
mpilquist: Can you add example usage here following the doctest format used in other methods? Also, add punctuation to the scaladoc.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1393,6 +1393,19 @@ object Stream {
     go(0)
   }
 
+  class PartiallyAppliedFromEither[F[_]] {
```
mpilquist: Add final?

core/shared/src/main/scala/fs2/Stream.scala
bpholt: Sure, will do.

PR #1239, opened 2018-08-25
Fixed #1236 - stream cancelation hang involving queues

### LINKED ISSUE #1236: With 1.0.0-M4 `.cancel` on stream with inner queue never returns
Not sure if this is issue with `fs2` or `cats-effect` itself, but following code:
```scala
object FiberCancelTest extends IOApp {
  def run(args: List[String]): IO[ExitCode] =
    async
      .synchronousQueue[IO, Int]
      .flatMap { q =>
        Stream(
          Stream
            .unfold(0)(i => (i + 1, i + 1).some)
            .flatMap { i =>
              Stream.sleep_(1.second) ++ Stream.emit(i)
            }
            .to(q.enqueue),
          q.dequeue.to(Sink.showLinesStdOut)
        ).parJoin(2).compile.drain.start.flatMap { f =>
          IO.sleep(5.seconds) *>
            IO(println("Canceling...")) *>
            f.cancel *>
            IO(println("Done."))
        }
      }
      .as(ExitCode.Success)
}
```
will perform as expected with `1.0.0-M3`: print four numbers, then `Canceling..`, then `Done`, then exit with code 0.

With `1.0.0-M4` app will print four numbers, then `Canceling`, then print fifth number and hang forever till killed with `SIGINT`.

If we remove `synchronousQueue` usage and will print number directly, everything will cancel just fine. But because queue constructor requires `Concurrent[F]` I think it should support cancellation.

mpilquist:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1056,9 +1056,7 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
     Stream
       .getScope[F2]
       .flatMap { scope =>
-        Stream.eval(F2.start(haltOnSignal.flatMap(scope.interrupt))).flatMap { fiber =>
-          this.onFinalize(fiber.cancel)
-        }
+        Stream.supervise(haltOnSignal.flatMap(scope.interrupt)) >> this
```
SystemFw: It seemed like the old code was still finalising the fiber. A bit unclear wrt difference now

mpilquist: Yeah, this change is gratuitous and unrelated to the fix. I could back it out if you think it's not needed. The real fix is the `F.never -> F.unit` change.

SystemFw: I don't mind it being there as a refactor. Just wanted to know if behaviour was changing or not.
I guess as it is, it's slightly surprising since it (seems to) warrant the introduction of a combinator, `supervise`, which isn't used anywhere else.

No real preference here tbh.

core/shared/src/main/scala/fs2/async/mutable/Queue.scala
```diff
@@ -199,7 +199,7 @@ object Queue {
                   F.guaranteeCase(d.get) {
                     case ExitCase.Completed => F.unit
                     case ExitCase.Error(t)  => cleanup *> F.raiseError(t)
-                    case ExitCase.Canceled  => cleanup *> F.never
+                    case ExitCase.Canceled  => cleanup *> F.unit
                   }
```
SystemFw: I think this is a remnant of the old `F` cancellation model where you'd make canceled computations non-terminating and use `onCancelRaiseError`. That would explain why it regressed in `M4`. The fix makes sense to me intuitively.

mpilquist: Yep, that's a good way to think about it. Makes sense now in hindsight.

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: On second look, this change isn't gratuitous though it doesn't help with the failing test case. The problem here is that a fiber is started and then it separately registered for cancelation. It could have been fixed by replacing eval + flatMap + onFinalize with bracket.

SystemFw: ah, true, if `start` succeeds but then things get interrupted the `cancel` finaliser never gets registered, 👍

Issue #1117, opened 2018-04-03
Improve queue


SystemFw:
Queue is quite an essential structure, and it works fairly to very well already, but there are a few ideas floating around on how to flesh it out/improve it.
I'm opening this issue to keep track of things to be explored before 1.0

- Adding a `tryDequeue: F[Option[A]]`
- Adding a `drainQueue: F[Option[Vector[A]]`
- Explore extracting the queue strategy from the queue implementation itself
- Evaluate solutions to the problem of expressing "this queue will not receive any elements anymore". This is currently encoded with `Option` and `dequeue.unNoneTerminate`, is this solution ok? Can it be improved?

paul-lysak:
Hi. `complete` and `fail` methods like in `akka.stream.scaladsl.SourceQueueWithComplete` would be quite useful, to my oppinion.

mpilquist:
@SystemFw Any thoughts on what's left here for 1.0?

SystemFw:
@pchlupacek I'm not too hung on the specific type (`Vector` or whatever else).
I guess what I was going for with `tryDequeue` and `drainQueue` was:
Try to give me an element, and immediately return is none is available `tryDequeue` (hard to express this with the current combinators) and, try to give me all the elements there, and immediately return if there's none (`drainQueue`). On further thought, maybe `drainQueue` is not safe, so perhaps we should only have `tryDequeue` and `tryDequeueBatch1`? But tbh, I'm not too hung up on either since I don't have a use case anymore, and I would probably build a custom structure if I need that behaviour.

So, unless either of you really wants that, we can proceed without I think. Thoughts?

PR #1240, opened 2018-08-27
Renamed fs2.async package to fs2.concurrent

mpilquist:
I got rid of the `immutable` and `mutable` packages -- only `Signal` used this split, which I replaced with `ReadableSignal` & `WritableSignal` & `Signal` which mixes in both traits.

Similarly, I introduced `Enqueue` and `Dequeue` as super-traits to `Queue`, allowing methods to be written which constrain usage to producing or consuming. I also removed the size tracking from `Queue` by default (we only used it in 1 place in the library) for performance reasons. To get size tracking back, I introduced `InspectableQueue` (not tied to this name).

I also removed the various deprecated functions from the old package object & got rid of `unsafeRunAsync`. In so doing, I noticed we were using `Effect` and `ContextShift` to implement `unsafeRunAsync` when it can be done better with just `ConcurrentEffect`.


mpilquist:
Note I didn't touch the stuff listed in #1117. If we still want to get that stuff in by 1.0, it would be good to start it soon.

SystemFw:
Got a couple of (minor) ideas but it's very late. Proper review tomorrow

pchlupacek:
@mpilquist Cool. I have in works QueueStrategy, as pure representation of queues. I will review this later on today.

pchlupacek:
As for signal naming is related I don't like too much Readable / Writeable. 

Perhaps only difference between Signal and Ref is the ability to have discrete / continuous streams. 

How about having Signal only with `discrete` and `continuous` whereas the `mutable` variant will be named `SignalRef` essentially describing  that this is Ref that can signal too. 

In fact, perhaps it could have the whole signal attached as `def signal: Signal[F, A]` instead of super trait.

SystemFw:
I think that for `Signal` it's enough to have `ReadableSignal` (not super keen about the name) and `Signal <: ReadableSignal`.
I'm not sure `WritableSignal` adds much since you can `.get = .modify(a => (a, a))` anyway.

pchlupacek:
@SystemFw just to clarify I suggest 
``` Signal == ReadableSignal ```  and ```SignalRef extends Ref with Signal ```

**Comment by SystemFw** [comment]
mm that's also an interesting idea. I'd like it a little better if methods that I consider implementations leak weren't added to `Ref`, like `compareAndSet` and `lazySet`

**Comment by pchlupacek** [comment]
Hm. `lazySet` is strange.   I think `compareAndSet` is also not perfect, as it assumes identity equality. Wonder why these been added? I think we would be able somehow to implement them, but they really ARE implementation leaks. I would fight for them to be removed be4 effects 1.0

**Comment by mpilquist** [comment]
If we made `SignallingRef <: Ref`, we'd pick up a bunch of new methods. This complicates the implementation though, as many of those methods aren't natural to implement with a single underlying `Ref[F, (A, Long, Map[Token, Deferred[F, (A, Long)])]` -- we'd probably need two refs: `Ref[F, A]` and `Ref[F, (Long, Map[Token, Deferred[F, (A, Long)])]`. Thoughts?

**Comment by pchlupacek** [comment]
@mpilquist Not sure on that. Which methods seems to be complicated?

**Comment by mpilquist** [comment]
@pchlupacek I was able to implement it with a single `Ref` but note some of the implementations are uglier than their non-signalling counterparts.

```scala
  private final class SignallingRefImpl[F[_], A](
      state: Ref[F, (A, Long, Map[Token, Deferred[F, (A, Long)]])])(implicit F: Concurrent[F])
      extends SignallingRef[F, A] {

    override def get: F[A] = state.get.map(_._1)

    override def continuous: Stream[F, A] =
      Stream.repeatEval(get)

    override def discrete: Stream[F, A] = {
      def go(id: Token, lastUpdate: Long): Stream[F, A] = {
        def getNext: F[(A, Long)] =
          Deferred[F, (A, Long)]
            .flatMap { deferred =>
              state.modify {
                case s @ (a, updates, listeners) =>
                  if (updates != lastUpdate) s -> (a -> updates).pure[F]
                  else (a, updates, listeners + (id -> deferred)) -> deferred.get
              }.flatten
            }

        Stream.eval(getNext).flatMap { case (a, l) => Stream.emit(a) ++ go(id, l) }
      }

      def cleanup(id: Token): F[Unit] =
        state.update(s => s.copy(_3 = s._3 - id))

      Stream.bracket(F.delay(new Token))(cleanup).flatMap { id =>
        Stream.eval(state.get).flatMap {
          case (a, l, _) => Stream.emit(a) ++ go(id, l)
        }
      }
    }

    override def set(a: A): F[Unit] = update(_ => a)

    override def lazySet(a: A): F[Unit] = F.start(set(a)).void

    override def getAndSet(a: A): F[A] = modify(old => (a, old))

    override def compareAndSet(expected: A, newValue: A): F[Boolean] =
      access.flatMap {
        case (old, setter) =>
          if (old.asInstanceOf[AnyRef] eq expected.asInstanceOf[AnyRef]) setter(newValue)
          else F.pure(false)
      }

    override def access: F[(A, A => F[Boolean])] = state.get.flatMap { snapshot =>
      F.delay {
        val hasBeenCalled = new java.util.concurrent.atomic.AtomicBoolean(false)
        val setter =
          (a: A) =>
            F.delay(hasBeenCalled.compareAndSet(false, true))
              .ifM(state
                     .compareAndSet(snapshot, (a, snapshot._2, snapshot._3)),
                   F.pure(false))
        (snapshot._1, setter)
      }
    }

    override def tryUpdate(f: A => A): F[Boolean] =
      F.map(tryModify(a => (f(a), ())))(_.isDefined)

    override def tryModify[B](f: A => (A, B)): F[Option[B]] =
      state
        .tryModify {
          case (a, updates, listeners) =>
            val (newA, result) = f(a)
            val newUpdates = updates + 1
            val newState = (newA, newUpdates, Map.empty[Token, Deferred[F, (A, Long)]])
            val action = listeners.toVector.traverse {
              case (_, deferred) =>
                F.start(deferred.complete(newA -> newUpdates))
            }
            newState -> (action *> result.pure[F])
        }
        .flatMap {
          case None     => F.pure(None)
          case Some(fb) => fb.map(Some(_))
        }

    override def update(f: A => A): F[Unit] =
      modify(a => (f(a), ()))

    override def modify[B](f: A => (A, B)): F[B] =
      state.modify {
        case (a, updates, listeners) =>
          val (newA, result) = f(a)
          val newUpdates = updates + 1
          val newState = (newA, newUpdates, Map.empty[Token, Deferred[F, (A, Long)]])
          val action = listeners.toVector.traverse {
            case (_, deferred) =>
              F.start(deferred.complete(newA -> newUpdates))
          }
          newState -> (action *> result.pure[F])
      }.flatten

    override def tryModifyState[B](state: State[A, B]): F[Option[B]] = {
      val f = state.runF.value
      tryModify(a => f(a).value)
    }

    override def modifyState[B](state: State[A, B]): F[B] = {
      val f = state.runF.value
      modify(a => f(a).value)
    }
  }
```

Also, `SignallingRef` has an `Invariant` instance, which means `compareAndSet` will in practice always fail due to reference equality (unless the `f`/`g` functions passed to `imap` return references):

```scala
          override def compareAndSet(expected: B, newValue: B): F[Boolean] =
            fa.compareAndSet(g(expected), g(newValue))
```

mpilquist:
I pushed the `ReadableSignal -> Signal, Signal -> SignallingRef` change.

mpilquist:
What do you think about renaming `InspectableQueue` to `SignallingQueue`?

pchlupacek:
No strong preference there. I like that `peeking` is separate functionality from Queue. in fact how about putting it to `def peek` or `def signal` ?  In my QueueStrategy branch I have removed `signal` variants in favor to discrete stream variants, as they are lazy and more generic. Whatever suits better ....

pchlupacek:
Also, perhaps when QueueStrategy will be ready, perhaps we can implement the Signal via separate QueueStrategy...

mpilquist:
Hm, I left `peek` as a method on `Dequeue` and hence `Queue`. I could see it moving to `InspectableQueue` though, in which case the current name is better than `SignallingQueue`. I'm fine with leaving things as they are in this PR and then re-evaluating queue after your strategy changes. If that sounds good to you and @SystemFw, we can merge this one. Though Fabio might have some comments yet to make.

pchlupacek:
@mpilquist cool whatever you preffer :)

pchlupacek:
yeah lets leave push button to Fabio :-)

SystemFw:
I'm going to merge :)

I very much like the general direction, with _maybe_ some changes to:
- naming (well, more ergonomics of the constructors, but no better ideas for now, we can deal with this in a following PR.
- The extraneous methods in `Ref`. There's a case to be made for them being removed in cats-effect imho.

Nice work!

pchlupacek:
@mpilquist just clarifying if you have been able to fix that `Signal.constant` ?

mpilquist:
@pchlupacek Hm what do you mean? AFAIK there are no outstanding issues with `Signal.constant`.

pchlupacek:
above my review SIgnal.scala#49? It must not complete (the discrete signal), so it cannot be defined as Stream(a)

mpilquist:
Hm, I don't see any review comments in Signal.scala. Are you saying it should be `Stream(a) ++ Stream.eval(F.never)`?

pchlupacek:
Ah ok I didnt close the review. Sorry.

pchlupacek:
@mpilquist yes. I think there was a discussing about that a while ago...

pchlupacek:
And no worries, we could do these later :-)

PR #1241, opened 2018-08-27
Addressed additional review comments from #1240

mpilquist:


SystemFw:
Sorry for the over-eager merge 😬

mpilquist:
@pchlupacek With the change to `discrete`, the applicative law tests hang.

pchlupacek:
@mpilquist not sure why is that, but the current definition of discrete is simply not correct, as the `Signal.of(1).discrete != Signal.constant(1).discrete`, and I think they shall be.

pchlupacek:
lgtm :-)

PR #1243, opened 2018-08-30
Delete fs2.concurrent.once

SystemFw:
Depends  on https://github.com/typelevel/cats-effect/pull/328


SystemFw:
`memoize` is in cats-effect now, so this can be merged

SystemFw:
ah, I actually didn't know what was that for! 
Will do that then, just a sec

SystemFw:
ugh, added this and then realised the old wording is different. Do you prefer we keep what was there before?

mpilquist:
No strong preference

PR #1246, opened 2018-09-03
Add readRange function reading a range of bytes from a file

### LINKED ISSUE #1227: Read a range of bytes from a file
Currently, we have:

```scala
  def readAll[F[_]: Sync: ContextShift](
      path: Path,
      chunkSize: Int,
      blockingExecutionContext: ExecutionContext): Stream[F, Byte] 
```

There should be a way to read a range of bytes.  Maybe this?

```scala
  def readRange[F[_]: Sync: ContextShift](
      path: Path,
      start: Long,
      end: Long
      chunkSize: Int,
      blockingExecutionContext: ExecutionContext): Stream[F, Byte] 
```

http4s has an [internal solution](https://github.com/http4s/http4s/blob/a2a7b8e40a1270919479d2ceb1641974ec7b4df3/core/src/main/scala/org/http4s/StaticFile.scala#L171-L198) which could be cleaned up and contributed.

asoltysik:
Fixes #1227

cc @rossabaker because I stole the code from him.

### DIFF
--- a/io/src/main/scala/fs2/io/file/file.scala
+++ b/io/src/main/scala/fs2/io/file/file.scala
@@ -25,6 +25,22 @@ package object file {
       .stream
 
   /**
+    * Reads a range of data synchronously from the file at the specified `java.nio.file.Path`.
+    * `start` is inclusive, `end` is exclusive, so when `start` is 0 and `end` is 2,
+    * two bytes are read.
+    */
+  def readRange[F[_]: Sync: ContextShift](path: Path,
+                                          blockingExecutionContext: ExecutionContext,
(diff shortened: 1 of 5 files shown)


djspiewak:
Wouldn’t this be much more efficiently achieved using a RandomAccessFile?

mpilquist:
Hm, `FileHandle#read` delegates to a `java.nio.channels.FileChannel`, which is at least competitive with `RandomAccessFile`. Some benchmarks show it performing better too: https://mechanical-sympathy.blogspot.com/2011/12/java-sequential-io-performance.html

PR #1247, opened 2018-09-04
Added fallible streams and made compilation more flexible

mpilquist:
This PR makes a few changes:
 - Adds the `Fallible` effect type, which is like `Pure` but allows exceptions to be raised
 - Replaces usage of `ApplicativeError` with `RaiseThrowable` instances, which better captures the fact that we only use it as a witness
 - Introduces the `Stream.Compiler[F[_], G[_]]` type class, which provides a compilation strategy for a `Stream[F, O]` to a `G[?]`
 - Adds compilation support for `F = Pure, Id, Fallible` allowing all methods on the old `ToEffect` to be used with each of these effect types
 - Renamed `ToEffect` to `CompileOps`
 - Replaced usage of `Nothing` in API with `INothing` to get better type inference -- this is the same trick we use with `Pure`
 - Removed `Pipe2.stepper` which was missed in a prior PR

The `Fallible` effect type has special syntax for `toList`, `toVector`, `toChunk`, and `to[Collection]`, just like `Pure` streams. E.g.:

```scala
scala> (Stream(1, 2, 3) ++ Stream.raiseError[Fallible](new Exception)).toList
res2: Either[Throwable,List[Int]] = Left(java.lang.Exception)
```

Some other examples of compilation:

```scala
scala> (Stream(1, 2, 3) ++ Stream.raiseError[Fallible](new Exception)).compile.foldMonoid
res7: scala.util.Either[Throwable,Int] = Left(java.lang.Exception)

scala> Stream(1, 2, 3).compile.foldMonoid
res8: cats.Id[Int] = 6

scala> (Stream(1, 2, 3) ++ Stream.eval[Id, Int](4)).compile.foldMonoid
res9: cats.Id[Int] = 10
```

As a side effect of using `RaiseThrowable`, we get a nicer error message when one cannot be found:

```scala
scala> Stream.raiseError(new Exception)
<console>:31: error: Cannot find an implicit value for RaiseThrowable[F]: an instance is available for any F which has an ApplicativeError[F, Throwable] instance or for F = Fallible. If getting this error for a non-specific F, try manually supplying the type parameter (e.g., Stream.raiseError[IO](t) instead of Stream.raiseError(t)). If getting this error when working with pure streams, use F = Fallible.
       Stream.raiseError(new Exception)
                        ^

scala> Stream.eval(IO(1)) ++ Stream.raiseError(new Exception)
<console>:31: error: Cannot find an implicit value for RaiseThrowable[F]: an instance is available for any F which has an ApplicativeError[F, Throwable] instance or for F = Fallible. If getting this error for a non-specific F, try manually supplying the type parameter (e.g., Stream.raiseError[IO](t) instead of Stream.raiseError(t)). If getting this error when working with pure streams, use F = Fallible.
       Stream.eval(IO(1)) ++ Stream.raiseError(new Exception)
                                              ^
```

Compared with prior to this PR:

```scala
@ Stream.raiseError(new Exception)
res1: Stream[scala.util.Try, Nothing] = Stream(..)

@ Stream.eval(IO(1)) ++ Stream.raiseError(new Exception)
res2: Stream[Object, Int] = Stream(..)
```


pchlupacek:
👍 LGTM

SystemFw:
will review today

mpilquist:
> I do think that if we make it public though, we need to specify it a bit more tightly: it hasn't much to say wrt how resource safety ought to work, for example.

How about we make it `sealed` for now at least?

> Meh on Fallible, I feel like I'm missing its advantages, do they offset having to learn (or even just wonder about) a new name and concept?

`Fallible` allows us to support pure streams that can fail while avoiding the unexpected exceptions. This came up while porting scodec-stream, which used the following:
```scala
object StreamEncoder {
  type Step[A] = Stream[Pure, A] => Pull[Pure, BitVector, Option[(Stream[Pure, A], StreamEncoder[A])]]
  ...
}
```

As a result of the `ApplicativeError` requirement, I had to change that to:
```scala
  type Failable[A] = Either[Throwable, A]
  type Step[A] = Stream[Pure, A] => Pull[Failable, BitVector, Option[(Stream[Pure, A], StreamEncoder[A])]]
```

I then evolved that idea to an uninhabited `Fallible` effect type to avoid the need for translating (e.g., b/c of possibility of `Stream.eval(Right(..))`.

This also lets us use things like `compress` with pure streams.

SystemFw:
> How about we make it sealed for now at least?

:+1:

> This came up while porting scodec-stream

If you have a concrete use case where this popped, I'll happily defer then :)

Ready for merge after `seal`ing `Compiler` imo

PR #1253, opened 2018-09-08
Added compile.lastOrError convenience function

mpilquist:
Per discussion in Gitter, this is a common function so we might as well provide it directly

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3364,6 +3364,25 @@ object Stream extends StreamLowPriority {
       foldChunks(Option.empty[O])((acc, c) => c.last.orElse(acc))
 
     /**
+      * Compiles this stream in to a value of the target effect type `F`,
+      * raising a `NoSuchElementException` if the stream emitted no values
+      * and returning the last value emitted otherwise.
+      *
+      * When this method has returned, the stream has not begun execution -- this method simply
+      * compiles the stream down to the target effect type.
+      *
+      * @example {{{
+      * scala> import cats.effect.IO
+      * scala> Stream.range(0,100).take(5).covary[IO].compile.lastOrError.unsafeRunSync
+      * res0: Int = 4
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
My only question if this would be better by 

```scala
def lastOrError(t: Throwable)(implicit G: MonadError[G, Throwable]): G[O] = ...
```

Otherwise a lot of times this will be directly followed by `adaptError` and it confused the `NoSuchElementException` and a business logic exception as you leave streaming. Not super opinionated here, just throwing out ideas.

SystemFw:
I'm not sure how to word it, but some comments about how this means that the stream has terminated/finished emitting would be good.

I mean things like:
```scala
Stream.hold(...).lastOrError.flatMap { .... signal never updated again...}
```

mpilquist:
@SystemFw Note this `lastOrError` is on `CompileOps` -- I think your comment assumes it's on `Stream` instead.

@ChristopherDavenport No strong opinion here. I kind of like the current signature more as it avoids the ceremony of `new NoSuchElementException` at each call site. Also note that we'd need to take the throwable as a by-name to avoid cost of instantiating it (and filling a stack trace) in the common case.

SystemFw:
> Note this lastOrError is on CompileOps -- I think your comment assumes it's on Stream instead.

:+1: I did assume it was on Stream, but I think the comment still holds some value, I can see people expecting that e.g. `hold` returns a `Stream[F, Signal]`, and since it's a singleton stream they can just `compile.lastOrError` and just keep working with it in `F`, which isn't the case. A similar thing would happen for finalisers. In many ways `.compile.drain` keeps you honest, but then again this helper _is_ useful, so not quite sure how to word a possible cautionary comment. Don't let this hold merging though

ChristopherDavenport:
Good call. I like this solution. I'm doing the whole ceremony now, so it's a reduction in boilerplate regardless.

mpilquist:
@SystemFw Yeah I see what you mean. If you come up with wording that works, let's update it in a new PR.

PR #1261, opened 2018-09-19
Update bracket signature in guide

Zelenya:
Either the guide is slightly outdated or I am looking in the wrong place


mpilquist:
Thanks!

PR #1262, opened 2018-09-20
Added timed actions for Queue

Twizty:
Hi everyone. This is my attempt to bring back timed actions for Queue from 0.10.

### DIFF
--- a/core/jvm/src/test/scala/fs2/concurrent/QueueSpec.scala
+++ b/core/jvm/src/test/scala/fs2/concurrent/QueueSpec.scala
@@ -1,8 +1,11 @@
 package fs2
 package concurrent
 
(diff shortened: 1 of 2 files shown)


mpilquist:
Thanks for the PR! We intentionally removed the timed actions from queues in favor of using [Concurrent.timeout](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Concurrent.scala#L388). The existing enqueue and dequeue APIs should all work correctly when used with `Concurrent.timeout`. Are you seeing something different?

Twizty:
I haven't actually seen this method in the `Concurrent`. Well, then this PR makes no sense.

PR #1269, opened 2018-09-24
Added Stream.bracketCase

mpilquist:
This needs some tests but I'm interested in feedback on the implementation. I basically just chased types.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -139,9 +139,9 @@ object Pull extends PullLowPriority {
     acquireCancellableCase(r)((r, _) => cleanup(r))
 
   /**
-   * Like [[acquireCancellable]] but provides an `ExitCase[Throwable]` to the `cleanup` action,
-   * indicating the cause for cleanup execution.
-   */
+    * Like [[acquireCancellable]] but provides an `ExitCase[Throwable]` to the `cleanup` action,
+    * indicating the cause for cleanup execution.
+    */
   def acquireCancellableCase[F[_]: RaiseThrowable, R](r: F[R])(
       cleanup: (R, ExitCase[Throwable]) => F[Unit]): Pull[F, INothing, Cancellable[F, R]] =
     Stream
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1455,6 +1455,13 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     Stream.bracket(F2.unit)(_ => f) >> this
 
   /**
+    * Like [[onFinalize]] but provides the reason for finalization as an `ExitCase[Throwable]`.
+    */
+  def onFinalizeCase[F2[x] >: F[x]](f: ExitCase[Throwable] => F2[Unit])(
+      implicit F2: Applicative[F2]): Stream[F2, O] =
+    Stream.bracketCase(F2.unit)((_, ec) => f(ec)) >> this
+
+  /**
     * Nondeterministically merges a stream of streams (`outer`) in to a single stream,
     * opening at most `maxOpen` streams at any point in time.
     *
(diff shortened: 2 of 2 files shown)


pchlupacek:
:-) This seems to enter shallow waters of resource management. Just to understand whats the benefit of adding this? (I am sure there is one, just need to quickly get on it)

mpilquist:
@pchlupacek Read from this comment to the end: https://github.com/functional-streams-for-scala/fs2/issues/1140#issuecomment-400077235

djspiewak:
This sort of thing would be profoundly useful in many cases. One example that we ran into in quasar is this general pattern:

1. Allocate some sort of bucket
2. Do work on completing that bucket
3. Bracket around **2** with a release which closes and publishes said bucket

In the even that the above is interrupted, step 3 happens, but what we *actually* want is to close and *clean up* the bucket, rather than publishing it. The way we're achieving this right now is with a `Signal` which we `set` as an action which is appended to **2** and then checked in **3**. This is quite clunky, and I think `bracketCase` would provide a more elegant answer.

SystemFw:
I'd like to have a look at this before merging.
Definite :tada: for having it though

pchlupacek:
@mpilquist @djspiewak , please note there is a significant difference between `interruption` and `cancellation`.  I am definitely in favor somehow to add this feature, I am just wondering between somehow clearly understand all the implications what this change will mean to interruption and how we can make sure the users won't somehow mismatch these two concepts that do completely different stuff :-) 

For example in case of what I understand from @djspiewak comment, this talks about `interruption` which is however completely unrelated to `cancellation`, and I didn't completely understand if the `interruption` of the stream causes somehow mimic the `cancellation` as you requested.

pchlupacek:
Just probably to clarify, what I am pointing out to this PR is that `interrupting` a stream will allow to signal `ExitCase.Cancel` and as such signal/ (or eventually recover) from this.

core/shared/src/main/scala/fs2/internal/Resource.scala
```diff
@@ -147,8 +147,10 @@ private[internal] object Resource {
                   else {
                     // scope is closed and this is last lease, assure finalizer is removed from the state and run
                     F.flatten(state.modify { s =>
-                      // previous finalizer shall be alwayy present at this point, this shall invoke it
-                      s.copy(finalizer = None) -> s.finalizer.getOrElse(F.pure(Right(())))
+                      // previous finalizer shall be always present at this point, this shall invoke it
+                      s.copy(finalizer = None) -> s.finalizer
+                        .map(_(ExitCase.Canceled))
```
AdamChlupacek: I think this should be `ExitCase.Completed` as releasing a lease seems like normal completion of the resource. It is just delayed.

mpilquist: Thanks, good catch

djspiewak:
> For example in case of what I understand from @djspiewak comment, this talks about interruption which is however completely unrelated to cancellation, and I didn't completely understand if the interruption of the stream causes somehow mimic the cancellation as you requested.

Ah, I could use some clarification on this point! What I'm talking about is stream interruption, for example with things like `interruptWhen`.

AdamChlupacek:
Good job. This all looks good. I am just afraid that people will mistaken `cancelation` and `interruption`. As for interrupted streams this will pass the `ExitCase.Completed` to the `bracketCase`. Now I think we could easily make it that in case of interruption we could pass `ExitCase.Canceled` to the `bracketCase`, but I am not entirely sure if we want to make them one case without the possibility of saying which is which.

AdamChlupacek:
@mpilquist oh, I wrote my comment too soon, seems like you already put the cases together.

mpilquist:
@AdamChlupacek @djspiewak `ExitCase.Canceled` should work now for `stream.interruptWhen`. I'll work on some tests & docs tonight.

SystemFw:
> Ah, I could use some clarification on this point! 

Same here actually, although it looks like things have been unified now regardless

mpilquist:
Okay this should be good for review now. Tests are messy but par for the course for `ResourceSafetySpec`. I'd really like to redo them at some point but I'd rather we do that later than doing it incrementally on each PR.

pchlupacek:
@mpilquist excellent. LGTM

PR #1271, opened 2018-09-25
Updated ScalaDoc for new fanout operations

mpilquist:


core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -106,22 +106,22 @@ private[fs2] final class CompileScope[F[_], O] private (
...
-      *
-      *  The new scope is interruptible but this scope is not interruptible -
-      *     This is a new interrupt root that can be only interrupted from within the new scope or its children scopes.
-      *
-      *  The new scope is interruptible as well as this scope is interruptible -
-      *     This is a new interrupt root that can be interrupted from within the new scope, its children scopes
-      *     or as a result of interrupting this scope. But it should not propagate its own interruption to this scope.
-      *
-      */
+    /*
```
Daenyth: I think you may want the second `*` so this is treated as scaladoc

mpilquist: I remove the second `*` to avoid a ScalaDoc warning about an unmoored doc comment -- ScalaDoc doesn't allow doc comments on local methods.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -118,45 +118,50 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
     attempt ++ delays.flatMap(delay => Stream.sleep_(delay) ++ attempt)
 
   /**
-    * Alias for `this.through(Broadcast(1))`
+    * Returns a stream of streams where each inner stream sees all elements of the
+    * source stream (after the inner stream has started evaluation).
+    * For example, `src.broadcast.take(2)` results in two
+    * inner streams, each which see every element of the source.
```
chuwy: "each which" is a typo I guess.

mpilquist: Thanks fixed.

PR #1272, opened 2018-09-25
Reimplemented mapAsync -> evalMapPar

### LINKED ISSUE #1226: Slow parallel processing
Hello guys, I'm afraid I've run into similar issue as was #977.
My scenario is actually very similar as there: streaming big data from sharded database, that means running multiple DB queries in parallel where each query returns some data (potentially GBs in production so we need to it have streamed).  
There is a huge difference between processing this data with `Observable` (Monix) and FS2 `Stream`, visible even in unit tests with some funny-small data. Based on the finding I've implemented a [repo](https://github.com/jendakol/streaming-benchmark) with benchmark and the scenario described more into deep.  
The actual code is quite short and naive but something has to be wrong there:
```scala
    Stream
      .range(1, items)
      .mapAsyncUnordered[IO, Array[Byte]](parallelism) { _ =>
        IO {
          DataGenerator.get(size)
        }
      }
      .map(_.length) // don't want to drain the stream so I "process" the data this way
      .reduce(_ + _)
      .compile
      .toList
      .runAndWait
```
The same implementation with `Observable` is about 30 times faster:
```scala
    Observable
      .range(1, items)
      .mapParallelUnordered[Array[Byte]](parallelism) { _ =>
        Task {
          DataGenerator.get(size)
        }
      }
      .map(_.length)
      .reduce(_ + _)
      .toListL
      .runAndWait
```

Results (shortened, see the repo):
```
[info] Benchmark                            (items)  (parallelism)  (size)  (threads)  Mode  Cnt    Score   Error  Units
[info] StreamingBenchmark.testFs2IO             500              5   10000         10  avgt    2   90,676          ms/op
[info] StreamingBenchmark.testMonix             500              5   10000         10  avgt    2    3,665          ms/op
```

What am I doing wrong? :-)

P.S. I am currently running full test with current setup (as is in the repo) but it will take few hours. I'll post it here later but intermediate results show no surprise.

pchlupacek:
+ mapAsyncUnordered -> evalmMapParUnordered

----
fixes #1226

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -775,6 +775,101 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
   }
 
   /**
+    * Like [[Stream#evalMap]], but will evaluate effects in parallel, emitting the results
+    * downstream in the same order as the input stream. The number of concurrent effects
+    * is limited by the `maxConcurrent` attribute and by size of chunk, whichever is lower.
+    *
+    * For example chunks of size 1 will always get evaluated synchronously, even when `maxConcurrent`
+    * will be set to Int.MaxValue.
+    *
+    * Similarly, if the chunk has size of 100, but `maxConcurrent` is set to 1, then effects will be evaluated
+    * sequentially.
+    *
+    *
+    * If any of concurrently running computations fails, the remainder of the computations iimmediately
+    * cancelled, and ocne all computations are cancelled, this will fail.
+    *
+    * @param maxConcurrent  Maximum number of concurrent effects.
+    * @param f              A function to transform `O` to `O2` evaluating effect `F`
+    */
+  def evalMapPar[F2[x] >: F[x]: Concurrent, O2](maxConcurrent: Int)(
+      f: O => F2[O2]): Stream[F2, O2] = {
+    def go(remains: Vector[O]): F2[Vector[O2]] = {
+      val (head, tail) = remains.splitAt(maxConcurrent)
+      if (head.isEmpty) Applicative[F2].pure(Vector.empty)
+      else {
+        head.traverse(o => Concurrent[F2].start(f(o))).flatMap { fibers =>
+          fibers.traverse(_.join).attempt.flatMap {
+            case Right(v) =>
+              go(tail).map(v ++ _)
+            case Left(err) =>
+              fibers.traverse_(_.cancel).attempt >> Sync[F2].raiseError(err)
+          }
+        }
+      }
+    }
+
+    this.chunks.flatMap { chunk =>
+      Stream.evalUnChunk(go(chunk.toVector).map(Chunk.vector))
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


mpilquist:
Any thoughts on an alias here? One thing that's nice about the current names is that they align with akka-stream and folks often come looking for them.

mpilquist:
Hm following conventions like `parTupled, parMapN, parJoin`, should these be `parEvalMap(Unordered)`?

SystemFw:
`parEvalMap` makes sense given convention, although tbh I'm not sure we should change names for the reason mentioned above. Especially because `evalMapParUnordered` is quite a mouthful...

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -774,6 +774,101 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    * downstream. The number of concurrent effects is limited by the `maxConcurrent` attribute
+    * and by size of chunk, whichever is lower.
+    *
+    * If any of concurrently running computations fails, the remainder of the computations iimmediately
+    * cancelled, and ocne all computations are cancelled, this will fail.
+    *
+    * See [[Stream#evalMapPar]] if retaining the original order of the stream is required.
+    *
+    */
+  def evalMapParUnordered[F2[x] >: F[x]: Concurrent, O2](maxConcurrent: Int)(
```
mpilquist: Seems like using this method on a stream with singleton chunks results in the same behavior as `evalMap`, which I find unintuitive. That is, I don't think chunk size should impact number of concurrent effects running.

mpilquist:
Let's push this one to post RC1.

pchlupacek:
No strong preference on naming. I thought the evalXXX is consistent with what we have now, if you really want to see the old names back just let me know.

jilen:
Maybe it could be `parEvalMap` just like `parJoin`

core/shared/src/main/scala/fs2/Stream.scala
pchlupacek: @SystemFw current implementation runs essentially whole lot of orchestration for processing the single element. This is not necessarily, hence for example you don't need any resource management, interruption etc. for `F[B]`

Just to be on same page. Are you talking about Unordered variant or both ?

mpilquist: When maxConcurrent == chunkSize, doesn't this implementation wait until each of the tasks finish before taking the next chunk? This is different behavior than the old version which attempted to make sure there were *always* maxConcurrent tasks running, regardless of chunk structure.

pchlupacek: @mpilquist Ah, I see what you mean. Sorry this took me that long to realize. Ok then, lets wait a while and I gonna improve this :-)

pchlupacek:
yes

mpilquist:
Closing this one since we haven't worked on it in a very long time. Feel free to open a new PR if you get a chance to work on this again.

Issue #1273, opened 2018-09-26
Restore `Queue#dequeueBatch1`?


rossabaker:
`dequeueBatch1` was [removed](https://github.com/functional-streams-for-scala/fs2/commit/4ddd75a2dc032b7604dc1205c86d7d6adc993859#diff-6ae893d277869fb64e113ba04fbd8b59L67) in #1258.

The closest replacement I've found is `Stream(size).through(queue.dequeueBatch).chunks.compile.lastOrError`.  Was there a strong reason to remove it?

mpilquist:
Fixed by #1274 -- take a look at the new `dequeueChunk1(maxSize)`

rossabaker:
I didn't write the test, but it looks like the intent is to [ensure that `Close` is the last message](https://github.com/http4s/http4s/blob/ec1e0600ece470815e01cb093c756830f4c9ca76/blaze-core/src/test/scala/org/http4s/blazecore/websocket/Http4sWSStageSpec.scala#L86) in a reasonable amount of time.  `pollBatchOutpubound` calls [this](https://github.com/http4s/http4s/blob/master/blaze-core/src/test/scala/org/http4s/blazecore/websocket/WSTestHead.scala#L66), which used to call `dequeueBatch1`.  If I understand the test, we could have polled for these messages indvidually, raced against a timeout, just as well.

It was moderately annoying for it to disappear, but if it cleans up the internals, I can live without it.

PR #1274, opened 2018-09-26
Added dequeueChunk1 to Queue

mpilquist:
Opening this for discussion only at this point. This is one potential solution to #1273.

If we're open to something like this, then I think we might want to change `DequeueChunk1` such that calling `dequeueChunk1` on a `NoneTerminated` queue returns `F[Option[Chunk[A]]]` instead of a `F[Chunk[Option[A]]]`. Perhaps via something like:

```scala
trait DequeueChunk1[F[_], G[_], A] {
  def dequeueChunk1(maxSize: Int): F[G[Chunk[A]]]
  def tryDequeueChunk1(maxSize: Int): F[Option[G[Chunk[A]]]]
}

trait Queue[F[_], A]
    extends Enqueue[F, A]
    with Dequeue1[F, A]
    with DequeueChunk1[F, Id, A]
    with Dequeue[F, A] { self =>
}

trait NoneTerminatedQueue[F[_], A]
    extends Enqueue[F, Option[A]]
    with Dequeue1[F, Option[A]]
    with DequeueChunk1[F, Option, A]
    with Dequeue[F, A] { self =>
}
```

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/Queue.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Queue.scala
@@ -46,6 +46,16 @@ trait Dequeue1[F[_], A] {
   def tryDequeue1: F[Option[A]]
 }
 
+/** Provides the ability to dequeue individual chunks from a `Queue`. */
+trait DequeueChunk1[F[_], A] {
+
+  /** Dequeues one `Chunk[A]` with no more than `maxSize` elements. Completes once one is ready. */
+  def dequeueChunk1(maxSize: Int): F[Chunk[A]]
+
+  /** Tries to dequeue a single chunk of no more than `max size` elements; yields to `None` if the element cannot be dequeued. */
+  def tryDequeueChunk1(maxSize: Int): F[Option[Chunk[A]]]
+}
+
 /** Provides the ability to dequeue chunks of elements from a `Queue` as streams. */
 trait Dequeue[F[_], A] {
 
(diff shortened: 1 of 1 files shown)


pchlupacek:
@mpilquist ah ok, that would work for sure 👍

mpilquist:
Cool -- are you okay with the addition suggestion I made about adding a `G` param to `DequeueChunk` to avoid the weird `Chunk[Option[A]]` we get back now?

core/shared/src/main/scala/fs2/concurrent/Queue.scala
```diff
@@ -46,6 +46,16 @@ trait Dequeue1[F[_], A] {
   def tryDequeue1: F[Option[A]]
 }
 
+/** Provides the ability to dequeue individual chunks from a `Queue`. */
+trait DequeueChunk1[F[_], A] {
+
+  /** Dequeues one `Chunk[A]` with no more than `maxSize` elements. Completes once one is ready. */
+  def dequeueChunk1(maxSize: Int): F[Chunk[A]]
+
+  /** Tries to dequeue a single chunk of no more than `max size` elements; yields to `None` if the element cannot be dequeued. */
```
Daenyth: > if the element cannot be dequeued

When does this happen?

mpilquist:
Cool, I pushed that update. Should be ready for review/merge now.

core/shared/src/main/scala/fs2/concurrent/Queue.scala
mpilquist: If there's no element available at the time of the `tryDequeueChunk1` call. Instead of semantically blocking for a chunk, it returns immediately with a None.

Daenyth: That wording would be more clear in the doc, for me at least. Thanks!

PR #1277, opened 2018-10-02
Bugfix/topic discrete subs

### LINKED ISSUE #1276: Topic.subscribers behaves as continuous stream rather than discrete after recent change
(Discussed in Gitter, issue created on request of @pchlupacek )

In a recent commit `Topic reimplemented with strategy` the signature for Topic.subscribers was changed from:
`  def subscribers: Signal[F, Int]`
to:
`  def subscribers: Stream[F, Int]`

However, the previous behaviour of:
 `topic.subscibers.discrete.map(_ => ...)  // pulls on change`
does not match the new behaviour of:
 `topic.subscribers.map(_ => ...)  // pulls continuously`

@SystemFw suggested the following testcase:
```
def p = Stream.eval(fs2.concurrent.Topic[IO, Int](0)).flatMap { topic => topic.subscribers.to(Sink.showLinesStdOut) concurrently Stream.fixedDelay(1.second).evalMap(_ => topic.subscribe(10).compile.drain.start.void) }.interruptWhen(timer.sleep(10.seconds).attempt)
```

pchlupacek:
fixes #1276

### DIFF
--- a/core/jvm/src/test/scala/fs2/concurrent/TopicSpec.scala
+++ b/core/jvm/src/test/scala/fs2/concurrent/TopicSpec.scala
@@ -1,8 +1,8 @@
 package fs2
 package concurrent
 
+import cats.implicits._
 import cats.effect.IO
-
 import fs2._
 import fs2.Stream._
 
@@ -80,6 +80,25 @@ class TopicSpec extends Fs2Spec {
           assert(diff.min == 0 || diff.min == 1)
           assert(diff.max == 0 || diff.max == 1)
       }
+    }
+
+    "discrete-size-signal-is-discrete" in {
+
+      def p =
+        Stream
+          .eval(Topic[IO, Int](0))
+          .flatMap { topic =>
+            def subscribers = topic.subscribers
+            def subscribe = Stream.fixedDelay[IO](200.millis).evalMap { _ =>
+              topic.subscribe(10).compile.drain.start.void
+            }
+            subscribers.concurrently(subscribe)
+          }
+          .interruptWhen(Stream.sleep[IO](2.seconds).as(true))
+
+      p.compile.toList
+        .unsafeRunSync()
+        .size shouldBe <=(11) // if the stream won't be discrete we will get much more size notifications
     }
   }
 }
--- a/core/shared/src/main/scala/fs2/concurrent/PubSub.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/PubSub.scala
@@ -612,7 +612,7 @@ private[fs2] object PubSub {
           def publish(i: I, state: State[S]): State[S] = {
             val qs1 = strategy.publish(i, state.qs)
             if (Eq[S].eqv(state.qs, qs1)) state.copy(qs = qs1)
-            else State(state.qs, Set.empty)
+            else State(qs1, Set.empty)
           }
 
           def get(selector: Either[Option[Token], Sel],
--- a/core/shared/src/main/scala/fs2/concurrent/PubSub.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/PubSub.scala
@@ -1,6 +1,6 @@
 package fs2.concurrent
 
-import cats.Applicative
+import cats.{Applicative, Eq}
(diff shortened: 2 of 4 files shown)


SystemFw:
> I took the liberty to use your code to demonstrate the issue in this PR and added it as test. Hope thats ok :-)

Lol, ofc it is

SystemFw:
I will add a test case for Queue as well before merge

mpilquist:
Is that assertion correct? Seems like it should be 0.

pchlupacek:
@mpilquist yeah, seems 0 is correct here. @SystemFw did you had anything else in mind ?

SystemFw:
I wrote that very late at night so maybe I'm just dead wrong, let me take another look

SystemFw:
The assertion is wrong (I always get `range` wrong :P) but the bug is there I think.
This is the unminimised version:
```scala
  def p =
    Stream
      .eval(fs2.concurrent.InspectableQueue.unbounded[IO, Int])
      .flatMap { q =>
        def changes =
          (Stream.range(1, 6).to(q.enqueue) ++ q.dequeue.take(5)).zip(Stream.fixedRate[IO](100.millis))
        q.size.concurrently(changes)
      }
      .interruptWhen(Stream.sleep[IO](5.seconds).as(true))
      .to(Sink.showLinesStdOut)
      .compile.drain.unsafeRunSync
```
```scala
scala> p
0
1
2
3
4
5
0
// wait
```
My assumption is that the size should go down to `4,3,2,1`. Is that unreasonable?

pchlupacek:
@SystemFw I think behaviour is correct. 

Note that `def dequeue = dequeueChunk(Int.MaxValue)`, So there are 5 publish and one dequeue (chunk) operation at the end.

SystemFw:
@pchlupacek Actually just realised this by doing:
`Stream.repeatEval(q.dequeue1)`

:)

I think the behaviour is correct but surprising. I'll delete the test, add a comment to InspectableQueue, and merge this, by the end of the day

pchlupacek:
@SystemFw hey, if you dequeue with ` q.dequeueBatch(1).take(5)` that will perform as you would expect :-)

SystemFw:
@mpilquist you beat me to it by a split second ;)

mpilquist:
Of course I did it wrong... fixing now

SystemFw:
Fixed. I had done the same mistake and had to correct it btw :P

nikelin:
I'm not sure, is it the right place to raise this issue but it seems related to this PR or some recent changes around topics.

Since 1.0.0-RC2 I started seeing some uncaught `StackOverflowException`. After debugging for some time, I realised that the problem is actually coming from the new PubSub implementation. In particular it happens on this like when topic has no active subscribers receiving its elements:

https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/core/shared/src/main/scala/fs2/concurrent/Topic.scala#L166

The problem can be reproduced with this code:
```
implicit val executionContext = ExecutionContext.global

    implicit val timer: Timer[IO] = IO.timer(executionContext)
    implicit val contextShift: ContextShift[IO] = IO.contextShift(executionContext)
    implicit val concurrent: ConcurrentEffect[IO] = IO.ioConcurrentEffect(contextShift)

    val topicA = fs2.concurrent.Topic[IO, String]("").unsafeRunSync()

    fs2.Stream.range(0, 10000000)
      .evalMap[IO, Unit](i =>
        topicA.publish1(i.toString)
      )
      .compile
      .drain
      .unsafeRunSync()
```

Seems to be related to the problem described here, but I'm not 100% sure: https://pavkin.ru/multiple-faces-of-scala-iterator-trap/

I'm not sure is this code semantically correct, but intuitively it looks that way. Let me know if it worth having a separate issue.

SystemFw:
@nikelin  maybe it was this?
https://github.com/functional-streams-for-scala/fs2/pull/1280

nikelin:
> #1280

Thanks! Looks like that. Are you planning to release it any time soon?

SystemFw:
1.0 is scheduled tomorrow :)

PR #1278, opened 2018-10-02
add jdk11 test

xuwei-k:


.travis.yml
```diff
@@ -7,6 +7,11 @@ scala:
 jdk:
   - oraclejdk8
 
+matrix:
+  include:
+  - jdk: oraclejdk11
```
SethTisue: nbd, but arguably better to default to `openjdk11` since `oraclejdk11` is now commercial. (the Oracle license does allow use during development and testing, so `oraclejdk11` isn't an outright wrong choice, but it still seems better to me to steer entirely clear of commercially licensed software)

PR #1280, opened 2018-10-03
Avoid stackoverflow in Topic

lucidd:
During some stress testing of my app i encountered a stackoverflow in [Topic](https://github.com/functional-streams-for-scala/fs2/blob/ff319f2310ef94d6c8781b5de46203f5dfdda47f/core/shared/src/main/scala/fs2/concurrent/Topic.scala#L166). The problem seems to be the use of `mapValues` in `publish` since it only wraps the `Map` in `MappedValues` ([known issue](https://github.com/scala/bug/issues/9499)). So if `accepts` is called after a lot of successive `publish` calls it blows up with a stackoverflow.

After applying this patch i did not encounter any stackoverflows anymore.

### DIFF
--- a/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
+++ b/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
@@ -2,10 +2,17 @@ package fs2
 
 import scala.concurrent.ExecutionContext
 import cats.effect.{ContextShift, IO, Timer}
-import fs2.concurrent.{Queue, SignallingRef}
+import fs2.concurrent.{Queue, SignallingRef, Topic}
 
 // Sanity tests - not run as part of unit tests, but these should run forever
 // at constant memory.
+
+object TopicContinuousPublishSanityTest extends App {
+  implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.Implicits.global)
+  Topic[IO, Int](-1)
+    .flatMap(topic => Stream.repeatEval(topic.publish1(1)).compile.drain)
+    .unsafeRunSync()
+}
 
 object ResourceTrackerSanityTest extends App {
   val big = Stream.constant(1).flatMap { n =>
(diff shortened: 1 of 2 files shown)


SystemFw:
@lucidd You don't happen to have a test case for this do you?

I will probably merge regardless, but it would be good to have one. We already have quite a lot of tests that check for memory consumption

SystemFw:
Thanks for fixing this btw :)

lucidd:
@SystemFw i don't have a minimal test case for it yet but i will try to write one. Thanks for the pointers.

SystemFw:
Thanks to you again  :)

> So if `accepts` is called after a lot of successive `publish` calls it blows up with a stackoverflow.

It seems like you've nailed down the specific problematic behaviour, so hopefully it shouldn't be too hard to isolate that with some code

lucidd:
@SystemFw I added a minimal test program that runs forever with the patch but stackoverflows without it. It was simpler then expected. Thanks for encouraging me to write the test case :+1:

core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
```diff
@@ -2,11 +2,17 @@ package fs2
...
 import cats.effect.{ContextShift, IO, Timer}
-import fs2.concurrent.{Queue, SignallingRef}
+import fs2.concurrent.{Queue, SignallingRef, Topic}
 
 // Sanity tests - not run as part of unit tests, but these should run forever
 // at constant memory.
 
+object TopicContinuousPublishSanityTest extends App {
+  implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.Implicits.global)
+  val topic = Topic[IO, Int](-1).unsafeRunSync()
```
SystemFw: Would you mind using `flatMap` instead of `unsafeRunSync` here :)

PR #1281, opened 2018-10-03
Improved performance of takeRight/dropRight

mpilquist:
It seems like `ChunkQueue` would be generally useful when writing pulls that attempt to preserve chunk boundaries.


pchlupacek:
hey that looks cool. We could perhaps add this to PubSubStrategy of Queue.

pchlupacek:
+1 to add it to Chunk.Queue. I meant that we can use it in implementation of the PubSub for queues that do accept chunks.

mpilquist:
No strong opinion on `Chunk.Queue` vs `ChunkQueue`. The `Chunk` companion has a bunch of subtypes which implement `Chunk` and I don't think this type should (though it would be possible). Assuming you are okay with `Chunk.Queue` not extending `Chunk`, I'll move it now.

SystemFw:
Yeah I didn't mean to extend it, just to move it away from the top level package given that it's fairly specialised (but not complicated enough to belong in `internal`)

SystemFw:
btw, does this need benchmark to justify its existence?

mpilquist:
Using this time method:

```scala
def time[A](f: => A): (FiniteDuration, A) = { val start = System.nanoTime; val r = f; ((System.nanoTime - start).nanoseconds, r) }
```

1.0.0-RC2:

```scala
@ val s = Stream.chunk(Chunk.ints(Array.fill(1024 * 1024 * 50)(0))).takeRight(5)
s: Stream[Nothing, Int] = Stream(..)

@ time(s.toList)
res27: (FiniteDuration, List[Int]) = (18280158622 nanoseconds, List(0, 0, 0, 0, 0))

@ time(s.toList)
res28: (FiniteDuration, List[Int]) = (22994350608 nanoseconds, List(0, 0, 0, 0, 0))

@ time(s.toList)
res29: (FiniteDuration, List[Int]) = (19344467809 nanoseconds, List(0, 0, 0, 0, 0))

@ time(s.toList)
res30: (FiniteDuration, List[Int]) = (18502128289 nanoseconds, List(0, 0, 0, 0, 0))

@ time(s.toList)
res31: (FiniteDuration, List[Int]) = (25060266522 nanoseconds, List(0, 0, 0, 0, 0))

@ time(s.toList)
res32: (FiniteDuration, List[Int]) = (18067113835 nanoseconds, List(0, 0, 0, 0, 0))
```

SNAPSHOT:

```scala
scala> val s = Stream.chunk(Chunk.ints(Array.fill(1024 * 1024 * 50)(0))).takeRight(5)
s: fs2.Stream[[x]fs2.Pure[x],Int] = Stream(..)

scala> time(s.toList)
res12: (scala.concurrent.duration.FiniteDuration, List[Int]) = (6132673774 nanoseconds,List(0, 0, 0, 0, 0))

scala> time(s.toList)
res13: (scala.concurrent.duration.FiniteDuration, List[Int]) = (9592186810 nanoseconds,List(0, 0, 0, 0, 0))

scala> time(s.toList)
res14: (scala.concurrent.duration.FiniteDuration, List[Int]) = (5708929299 nanoseconds,List(0, 0, 0, 0, 0))

scala> time(s.toList)
res15: (scala.concurrent.duration.FiniteDuration, List[Int]) = (5540729768 nanoseconds,List(0, 0, 0, 0, 0))

scala> time(s.toList)
res16: (scala.concurrent.duration.FiniteDuration, List[Int]) = (5567032563 nanoseconds,List(0, 0, 0, 0, 0))

scala> time(s.toList)
res17: (scala.concurrent.duration.FiniteDuration, List[Int]) = (5642898514 nanoseconds,List(0, 0, 0, 0, 0))

scala> time(s.toList)
res18: (scala.concurrent.duration.FiniteDuration, List[Int]) = (5637929957 nanoseconds,List(0, 0, 0, 0, 0))
```

Throw out the second sample of each as I accidentally ran them at the same time and you end up with these averages:
- 1.0.0-RC2 = ~19.85s
- 1.0.0-SNAPSHOT = ~5.7s

About a 4x speedup in this case.

SystemFw:
I wasn't actually questioning that :P
More asking if we should start adding benchmark code to this kind of PR, or if the cost/benefit ratio isn't high enough for that

mpilquist:
Ha, no worries. I'm not sure where to capture these types of benchmarks honestly. We have the JMH stuff but they don't get run and are hard to compare. There's the fs2-benchmarks project but no one has really taken an interest there.

SystemFw:
> There's the fs2-benchmarks project but no one has really taken an interest there.

Yeah fair enough. Ideally that's where those would belong I think, but maybe we're stretched too thin to deal with that as well right now. Just wanted to ask the question, go ahead and merge if Pavel has no further comments :)

pchlupacek:
Well actually, we could add them to benchmark, but the problem here you compare two versions OLD/NEW and that would mean to maintain old code :-)

PR #1289, opened 2018-10-11
Add isOpen to Socket Interface

ChristopherDavenport:
I was trying to create a ConnectionPool for a client. However these sockets sitting somewhere may be dead whenever we come back to them, presently I have no way to tell which of a pool of sockets would be still active to be reused. This would hopefully restore my ability in this regard.

### DIFF
--- a/io/src/main/scala/fs2/io/tcp/Socket.scala
+++ b/io/src/main/scala/fs2/io/tcp/Socket.scala
@@ -65,6 +65,8 @@ trait Socket[F[_]] {
   /** Indicates to peer, we are done writing. **/
   def endOfOutput: F[Unit]
 
+  def isOpen : F[Boolean]
+
   /** Closes the connection corresponding to this `Socket`. */
   def close: F[Unit]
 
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
Failure is the MiMa check.

mpilquist:
@ChristopherDavenport Do you still need this operation? We can suppress the mima check since folks don't implement this trait.

ChristopherDavenport:
I cannot satisfy the http4s pool without it. I have built out my own pooling mechanism in the meantime, but this is something I think will be useful for connection pooling in general.

djspiewak:
I can confirm that this is indeed pretty important. Obviously, ultimately determining whether a socket *is* open in the sense that data can be transmitted requires trying to send packets, but having a faster determination when one can be made is a significant optimization and one that helps quite a lot in any pool-like scenario.

mpilquist:
At NEScala, @ChristopherDavenport pointed out that fs2-crypto extends `Socket`  (https://github.com/Spinoco/fs2-crypto/blob/9acb97daed666587568f6732aa1c48e198595a48/src/main/scala/spinoco/fs2/crypto/io/tcp/TLSSocket.scala#L22).

If we just suppress the mima error, then anyone who uses fs2-crypto *and* calls `isOpen` on a `TLSSocket` will get a `NoSuchMethodError`. I think that's an okay tradeoff assuming @pchlupacek is okay with making a new release of fs2-crypto against an fs2 1.0.5.

AdamChlupacek:
@mpilquist Releasing new fs2-crypto with the next release is fine, I can do it then, this seems like quite important API addition, just to block it on this.

PR #1291, opened 2018-10-12
Add spawn

SystemFw:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2021,6 +2021,12 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
       .stream
   }
 
+  /**
+    * Starts this stream and cancels it as finalization of the returned stream.
```
Daenyth: I think it's worth noting that this is distinct from `evalMap(_.start).void` in the cancellation behavior.

PR #1294, opened 2018-10-17
Add metered

vpavkin:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2789,6 +2789,12 @@ object Stream extends StreamLowPriority {
   def iterateEval[F[_], A](start: A)(f: A => F[A]): Stream[F, A] =
     emit(start) ++ eval(f(start)).flatMap(iterateEval(_)(f))
 
+  /**
+    * Like [[iterate]], but emits elements with the fixed rate of `d`.
+    */
+  def iterateEvery[F[_], A](d: FiniteDuration)(start: A)(f: A => A)(implicit timer: Timer[F]): Stream[F, A] =
```
pchlupacek: Not sure about the name here. I think this is related to scheduling and perhaps we have somehow express this. Something like iterateAtRate or similar.

vpavkin: @pchlupacek I named it consistently with `awake` family (awakeEvery). But `iterateAtRate` also fits.
I'm fine with both. 

Actually, pronouncing `iterateAtRate` feels a bit awkward :)

pchlupacek: I am not expert on the naming these mouthful operators. Lets see what @SystemFw @mpilquist thinks about this.

pchlupacek: My whole idea is that `awake` sort of implies the scheduling, where `iterate` does not plus, we already use iterate for bunch of scheduling unrelated methods.

SystemFw: `iterateEvery` is fine as a name, it's consistent with `awakeEvery`. I'm slightly more doubtful that this op warrants inclusion (you can add an `*Every` to every op in the lib), but not really opposed to it either, whatever you guys prefer

mpilquist: How about a new method on `Stream`:
```scala
def metered[F2[x] >: F[x]: Timer](period: FiniteDuration): Stream[F, O] =
  fixedRate[F2](period).zipRight(this)
```

Not sure we need `iterateEvery` then as `iterate(start)(f).metered(d)` is pretty short and `metered` is very general.

SystemFw: :+1:

vpavkin: @mpilquist @SystemFw @pchlupacek  Fair enough, a I can do that. 
What concerns me in this case, is that user might expect this to always produce at the specified rate, while in reality general `this` can have longer gaps and the output rate will not be `metered`.

WDYT?

pchlupacek: Also, as side note we could make perhpas `metered` to be more effective than zipping, by just inserting it at every value emitted.... like `this.evalMap { a => Timer.sleep(d) as a }`

mpilquist: I think that's okay though should be noted in the ScalaDoc. Certainly open to other names as well if you come up with something clearer. In many of the systems I work on, metering refers to the steady output of buffered data & the behavior when the buffer is empty is often very domain specific.

vpavkin: in this sense smthing like `atMaxRateOf` could be more precise

vpavkin: > Also, as side note we could make perhpas `metered` to be more effective than zipping, by just inserting it at every value emitted.... like `this.evalMap { a => Timer.sleep(d) as a }`

Doesn't this change the semantics from `fixedRate` to `fixedDelay`?

SystemFw: >  this sense smthing like atMaxRateOf could be more precise

That's not the case, if the other stream is slower if will be slower.

> ike this.evalMap { a => Timer.sleep(d) as a }

These semantics are different, that's `fixedDelay`, not `fixedRate`

pchlupacek: yeah it does. :-)

vpavkin: @SystemFw if the other stream is slower it will be slower.

I might be wrong, but I think that's what `maxRate` implies - if `this` is too fast, the rate is limited, but if it's slow, the rate is slow as well.

`metered` is more concise though, I'll go with it.

vpavkin:
I've spent like 15 min on the scala doc, but still feel it can be worded better :)

PR #1299, opened 2018-10-24
Fix concurrency-primitives.md

kczulko:
Fix concurrency-primitives.md
- `SignallingRef` instead of `Signal`
- more relevant termination example with `SignallingRef`
- fix compilation issues (e.g `Stream.awakeEvery[F]`)


kczulko:
Hi,

According to [that](https://github.com/functional-streams-for-scala/fs2/issues/1298#issuecomment-432629597) reply from @SystemFw I've changed the example a little bit to make it more relevant when it comes to the usage of `SignallingRef`. Please let me know if that looks fine.
BTW: I'm still learning fs2 and this PR gave me a lot of fun. I'm amazed how convenient and expressive is this API. Bravo!

Regards,
Karol

mpilquist:
Thanks!

PR #1301, opened 2018-10-30
Fix lifo/fifo terminology in Queue

ghostdogpr:
The implementation of `FIFO` is actually `LIFO` and vice-versa. The default strategy is `LIFO` which actually acts as a `FIFO`, which is quite confusing.

This PR fixes the function names (`FIFO` to `LIFO` and vice versa) and sets the default to `FIFO`.

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/Queue.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Queue.scala
@@ -151,12 +151,12 @@ object Queue {
 
   /** Creates a queue with the specified size bound. */
   def bounded[F[_], A](maxSize: Int)(implicit F: Concurrent[F]): F[Queue[F, A]] =
-    forStrategy(Strategy.boundedLifo(maxSize))
+    forStrategy(Strategy.boundedFifo(maxSize))
 
   /** Creates a bounded queue terminated by enqueueing `None`. All elements before `None` are preserved. */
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/concurrent/Queue.scala
```diff
@@ -143,11 +143,11 @@ object Queue {
...
-    forStrategy(Strategy.lifo[A])
+    forStrategy(Strategy.fifo[A])
 
   /** Creates an unbounded queue that distributed always at max `fairSize` elements to any subscriber. */
   def fairUnbounded[F[_], A](fairSize: Int)(implicit F: Concurrent[F]): F[Queue[F, A]] =
-    forStrategy(Strategy.lifo[A].transformSelector[Int]((sz, _) => sz.min(fairSize)))
+    forStrategy(Strategy.fifo[A].transformSelector[Int]((sz, _) => sz.min(fairSize)))
 
   /** Creates a queue with the specified size bound. */
   def bounded[F[_], A](maxSize: Int)(implicit F: Concurrent[F]): F[Queue[F, A]] =
```
mpilquist: Looks like `bounded`, `boundedNoneTerminated`, and `fairBounded` need to be updated too.

ghostdogpr: You're right! That should explain the failed test.

Issue #1196, opened 2018-07-31
Create scalafix rules to implement as much of the 0.10 to 1.0 upgrade as possible


mpilquist:
For starters, it would be nice to handle the various term and type renames described in https://github.com/functional-streams-for-scala/fs2/pull/1195/files.

I spent a few hours on getting a few simple renames working here: https://github.com/functional-streams-for-scala/fs2/compare/series/1.0...mpilquist:topic/scalafix?expand=1

I also added support for `Stream.bracket` changes.

olafurpg:
Very cool! A few comments

* note that `ctx.replaceTree` replaces the full tree with new tokens. If you expect to run further rewrites for the subtrees inside `$aquire` and `$release` then this approach will produce unexpected results. Example case where this might be a problem (but I have no idea if that code makes sense): 
```scala
Stream.bracket(IO(1))(Stream(_), i =>  Stream.bracket(IO(1))(Stream(_), i => IO.unit))
```
* `case Term.Select(_, bracket(_)) =>` might be possible to simplify to `case bracket(_: Term.Name) =>`, I don't know if it works in v0.5.10 but it will work in v0.6 at least.
* heads ups, the method symbol format will change in the upcoming v0.6 release. Example, the method `_root_.fs2.Stream.InvariantOps#observe1(Lscala/Function1;Lcats/Functor;)Lfs2/internal/FreeC;.` becomes `fs2/Stream.InvariantOps#observe1().` instead. The new method symbols are both easier on the eye and also have the benefit of being possible to produce with only a parser (no compiler). The changelog will document further breaking changes, and there is also a migration script for scalafix.g8 generated project https://github.com/scalacenter/scalafix.g8/tree/master/scalafix

olafurpg:
I hope to have a stable Scalafix v0.6 release out this August, with a revamp of the documentation both for end-users and rule authors. I'm happy to help migrating existing rules to the new API so don't hesitate to ping me.

SystemFw:
Thanks Olaf :)

PR #1304, opened 2018-11-01
Scalafix from 0.10 to 1.0

amarrella:
Attempts a very first stab at #1196. I know there is still a lot to do though :) 
Feedback is more than welcome!
I'm willing to try to do the fixes for all the changes (in the limits of the possible) in [migration guide](https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/docs/migration-guide-1.0.md) if there is interest


scalafix/build.sbt
```diff
@@ -0,0 +1,58 @@
+lazy val V = _root_.scalafix.sbt.BuildInfo
+inThisBuild(
+  List(
+    organization := "com.alessandromarrella",
```
amarrella: this and below can/should be changed to fs2 stuff :)

scalafix/rules/src/main/scala/fix/StreamAppRules.scala
```diff
@@ -0,0 +1,51 @@
...
+      Patch.replaceTree(d, newDef.toString)
+    }.asPatch
+
+  private[this] def addCatsEffectImports(implicit c: SemanticContext): Patch =
+    Patch.addGlobalImport(Symbol("cats/effect/IO.")) +
+      Patch.addGlobalImport(Symbol("cats/effect/ExitCode.")) +
+        Patch.addGlobalImport(Symbol("cats/effect/IOApp."))
+
+  private[this] def addCatsSyntaxImport(implicit c: SemanticDocument): Patch =
+    Patch.addGlobalImport(importer"cats.syntax.functor._")
```
Daenyth: Is it possible to make this detect whether cats syntax is already present, and if so skip this?

Also it might be good to check for both `cats.implicits._` as well as functor syntax

amarrella: Yeah this is a refinement I want to do in a second time. For now I'm focusing on the migration itself. Though good point! I'll keep it in mind :)

amarrella: Fixed in https://github.com/functional-streams-for-scala/fs2/pull/1304/commits/ba16b95bc4df6aa87558453370f5f5ea203a6d17

mpilquist:
Awesome work!

amarrella:
Maybe this is a good moment to stop and see if the pr is going well. Wdyt? Can we merge/review what has been done so far? We have a few renames missing and import handling

amarrella:
Sure, no rush :) I won't be able to address any comment until next week anyway (traveling) . Thanks for your time!

amarrella:
I refactored the fix to be github based to be compatible with https://github.com/fthomas/scala-steward/issues/149

amarrella:
This is now testable in your project with the following
```
sbt ";scalafixEnable;scalafix github:amarrella/fs2/v1?sha=672ea4f9"
```
replace sha with the commit you want to test.

amarrella:
I reduced a little bit the impact of the scalafix on StreamApp to IOApp (should preserve formatting better now)

mpilquist:
Thanks for the patience on this one. I'll spend some time this Wednesday on it.

scalafix/rules/src/main/scala/fix/v1.scala
```diff
@@ -0,0 +1,511 @@
...
+      case t: TypeRef    => Some(t.symbol)
+      case t: SingleType => Some(t.symbol)
+      case t: ThisType   => Some(t.symbol)
+      case t: SuperType  => Some(t.symbol)
+      case _             => None
+    }
+
+  def getEffectType(symbol: Symbol)(implicit doc: SemanticDocument): String =
+    getType(symbol).toString
+      .takeWhile(_ != '[') // There might be a better way, but for our purposes it's enough
```
mpilquist: :)

mpilquist:
This looks great, awesome work!  A couple comments:
1) Can we add something to the migration guide instructing folks how to run the upgrade?
2) To run the rules on a sample project, I had to upgrade from SBT 0.13 and manually add the scalafic plugin to project/plugins.sbt. Might be worth mentioning this somewhere.
3) References to `Segment[O, Unit]` are being rewritten to `Chunk[O, Unit]` but should be rewritten to `Chunk[O]`
4) Is there any way to use non-qualified imports? E.g., `Chunk` instead of `fs2.Chunk`?

Here's the result of running the upgrade on scodec-protocols: https://github.com/scodec/scodec-protocols/commit/ae1882e838b6ff6823c47f4dc36ff0d37c365b83

amarrella:
Apologies for the delay, I was very busy! :( 

I think I tackled 3 & 4 now, not sure if you want to replace that with simple linting messages, or keep this as is now. 

I'll do 5 as soon as I can for the other imports.

For 1 & 2 I'll try to do that as well as soon as possible.

amarrella:
This is how it looks now with scodec-protocols https://github.com/scodec/scodec-protocols/compare/4e6933c108788eb872a39f5889903efa4e5da124...amarrella:example/fs2-upgrade-via-scalafix

I agree that the `Chunk` changes are complex and would probably break the client code anyway, maybe we should remove them from this PR and eventually add them later if we find a cleaner solution? 

Let me know :)

mpilquist:
@amarrella Thanks for all the work on this one!

Issue #1293, opened 2018-10-16
Queue (consistently) misses an element on certain usage


svalaskevicius:
When a dequeue stream is manipulated (merged with another stream, and split into two via pull), then dequeuing misses an element (always the first element on a subsequent pull). an example of the code showing this can be seen here: https://scastie.scala-lang.org/LBTK7fSnQIC0kLEkvk9cYA

Not only the element (4) is missing from the second printed list, but it is also not shown in the dequeue's evalTap logging - there is no "deq: 4" line.

This behaviour is also consistent - all repeated runs result in the same error.

`run1` in the example is added as a passing example - if the stream is split by amount, dequeuing works ok.

related conversation (in case there are more replies) on gitter starts here: https://gitter.im/functional-streams-for-scala/fs2?at=5bc642ed82893a2f3bf0cef9

fs2 version: 1.0.0
The same function (`splitByTimeout`) was working ok with an earlier fs2 release (0.10.x)

mpilquist:
@pchlupacek I think this is a bug in `dequeue` which is defined as:

```scala
def dequeueChunk(maxSize: Int): Stream[F, A] =
  Stream.evalUnChunk(pubSub.get(maxSize)).repeat
```

There's no bracketing occurring here, so when the stream is canceled, this isn't finalized. I think we need something like `Stream.bracket(pubSub.get(maxSize).start)(_.cancel).flatMap(Stream.chunk)`

pchlupacek:
@mpilquist yeah seems so will take a look on this.

pchlupacek:
@mpilquist its strange. the pubsub get actually performs cleanup in subscribers on every get. so there shall not be a need for that bracket imho.

pchlupacek:
ah i see the issue.  the get is not cancelled, as it is not treated as resource in stream context.  I think I will come with some generic fix for that  :)

mpilquist:
Right we either supervise an incur the cost of the fiber or we have an alternative to get that returns something like Resource[F, A]

mpilquist:
@pchlupacek Any luck? I'm fine with the simple fiber based solution and can PR that if you are busy.

pchlupacek:
@mpilquist could you give me a day or two? I have solution in head, essentially I want to add subscribe: Stream[F, A] to PubSub for these situations, is quite simple actually, its only that you bracket the stream instead every `get`.

mpilquist:
:+1: no problem at all

[...]

PR #1306, opened 2018-11-05
Introduces Resource on PubSub#get

pchlupacek:
I think this shall fix the #1293, but somehow test I brought up did not pass. 

I am quite not confident if test is correct. @mpilquist can you take a very quick on QueueSpec? I just took your minimalistic example and somehow this still fails.

### DIFF
--- a/core/jvm/src/test/scala/fs2/concurrent/QueueSpec.scala
+++ b/core/jvm/src/test/scala/fs2/concurrent/QueueSpec.scala
@@ -75,6 +75,21 @@ class QueueSpec extends Fs2Spec {
         }
       }
     }
+
+    "dequeue releases subscriber on interrupt" in {
+
+      Queue
+        .unbounded[IO, Int]
+        .flatMap { q =>
+          q.dequeue.interruptAfter(1.second).compile.drain *>
+            q.enqueue1(1) *>
+            q.enqueue1(2) *>
+            q.dequeue1
+        }
+        .unsafeRunSync shouldBe 1
+
+    }
+
     "size signal is initialized to zero" in {
       runLog(
         Stream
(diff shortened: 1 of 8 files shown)


mpilquist:
@pchlupacek Yeah, the test is correct. The dequeue is interrupted and after that completes, we enqueue 2 elements and dequeue 1 element. We must get back the first element we dequeued or otherwise, the original `dequeue` still has an outstanding effect reading from the queue.

I briefly skimmed this new implementation and I think the issue is that nothing interrupts the `Deferred#get` call upon cleanup of the resource returned from `get`. That is, the finalizer is correctly removing the token & associated state from the Queue state but not doing anything about outstanding calls to the returned `get_`.

pchlupacek:
@mpilquist thats where I am confused. If we finalize Resource and Subscriber is cleared there shall not be a way, how that subscriber may be completed again? 

The get_ performs 
1. registration to subscribers
2. deferred.get

So if we interrupt awaiting `get`
1. cancel the `deferred.get` // note `get` is not as resource, it shall be cancellable as such 
2. unregister subscriber

that means any subsequent publishes may not see subscriber again. 

Or you saying that Resource is not cleared in this queue dequeueChunk implementation? 

```scala

      def dequeueChunk(maxSize: Int): Stream[F, A] =
          Stream.resource(pubSub.get(maxSize)).flatMap { get =>
            Stream
              .repeatEval(get)
              .unNoneTerminate
              .flatMap(Stream.chunk)
          }

```

As I read the code, when `get` is cancelled the (1) from cleanup is called and then (2) is called as a matter of finaliser on resource being invoked. 

Or am I getting this completely wrong?

mpilquist:
See comments below:

```scala
 def dequeueChunk(maxSize: Int): Stream[F, A] =
          Stream.resource(pubSub.get(maxSize)).flatMap { get =>
            Stream.evalUnChunk(get).repeat // <-- the evaluation of `get` here isn't interrupted, it's just ignored -- the stream interpreter will evaluate `get` and when stream is interrupted, `get` will still finish in the background
          }
```

pchlupacek:
Hm that leaves the question if we shall not have 2 implementations, really, one for F[A] where the cleanup is called on cancel, and the other with Stream[F, A] where cleanup is invoked always, when the  stream finishes. As leaving it only on resource will require the cleanup to be called always when `F` completes, which is a bit overhead for normal Get? 
What do you think ?

mpilquist:
Yeah, I think 2 implementations could be a good solution

pchlupacek:
Ok I'll gonna fix that. So lets close this and I open another PR

PR #1307, opened 2018-11-05
Resource: small code improvements.

diesalbla:
This PR suggests some changes to the code of the `Resource`, which I hope may improve it a bit:  

- Since Leases are built without parameters, and they only used the immutable value `state: Ref[F]` (immutable reference to Ref object), we can just have a single Lease object for all those using it.
- We extract a function `State.isFinished` to indicate when the state  allows running the `finalizer`.
- We extract a simple variable for the `F.pure(Right(()))` expression.
- Makes the resource `state` into a private val.


mpilquist:
Thanks!

PR #1308, opened 2018-11-06
Bugfix/queue get

### LINKED ISSUE #1293: Queue (consistently) misses an element on certain usage

pchlupacek:
Fixes #1293. 

Needs to be squashed before merge.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -197,9 +197,9 @@ lazy val mimaSettings = Seq(
   mimaBinaryIssueFilters ++= Seq(
     ProblemFilters.exclude[Problem]("fs2.internal.*"),
     ProblemFilters.exclude[Problem]("fs2.Stream#StepLeg.this"),
-    ProblemFilters.exclude[Problem]("fs2.concurrent.Publish"),
-    ProblemFilters.exclude[Problem]("fs2.concurrent.Subscribe"),
-    ProblemFilters.exclude[Problem]("fs2.concurrent.PubSub")
+    ProblemFilters.exclude[Problem]("fs2.concurrent.Publish.*"),
+    ProblemFilters.exclude[Problem]("fs2.concurrent.Subscribe.*"),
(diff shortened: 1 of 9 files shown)


pchlupacek:
huh travis complains about bincompat? All I have touched is private, so this seems odd :-(

mpilquist:
We can add an exclusion for the Travis error.

Does this test pass too?

```scala
      Queue
         .unbounded[IO, Int]
         .flatMap { q =>
           q.dequeue.timeout(1.second) *>
             q.enqueue1(1) *>
             q.enqueue1(2) *>
             q.dequeue1
         }
         .unsafeRunSync shouldBe 1
```

pchlupacek:
@mpilquist tests passes. As regard the exclusions, where to add them ?

mpilquist:
https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/build.sbt#L197

pchlupacek:
Also going forward perhaps fro next release of fs2, I would like to explore, if we could implement `Stream.bracket` with similar signatures, essentially making `canceling` a stream instead of `interrupting` it. I think it could be possible and have bracket propagate the `Case` to finalizer, similarly like `Bracket` does in cats.

PR #1309, opened 2018-11-07
create Text message with different milliseconds

erlangxk:


guersam:
How about utilizing [`cats.effect.Clock#realTime`](https://typelevel.org/cats-effect/api/cats/effect/Clock.html#realTime(unit:concurrent.duration.TimeUnit):F[Long])?

mpilquist:
@erlangxk Sorry for the delay on this one! Could you update this PR to use `Clock` as suggested by @guersam?

erlangxk:
using clock now, but I commit via another github account, is it ok?

PR #1311, opened 2018-11-07
Fix parEvalMap getting stuck on exception

### LINKED ISSUE #1297: Stream hangs non-deterministically when using mapAsync
The following program:
```scala
import cats.effect.{ExitCode, IO, IOApp}
import cats.implicits._
import fs2.Stream

object Main extends IOApp {
  override def run(args: List[String]): IO[ExitCode] = {
    val stream =
      Stream
        .iterate(0)(_ + 1)
        .covary[IO]
        .mapAsync(25) { n =>
          if (n > 100) IO.raiseError(new RuntimeException)
          else IO(println(n))
        }

    stream.compile.drain.as(ExitCode.Success)
  }
}
```
when run somtimes hangs (never completes) indefinitely. It might take several runs/attempts for the hang to occur, but I'm always able to eventually trigger it locally.

vlovgr:
Fix #1297. Explanation is in https://github.com/functional-streams-for-scala/fs2/issues/1297#issuecomment-436666729.

### DIFF
--- a/core/jvm/src/test/scala/fs2/PipeSpec.scala
+++ b/core/jvm/src/test/scala/fs2/PipeSpec.scala
@@ -183,6 +183,12 @@ class PipeSpec extends Fs2Spec {
       runLog(r) shouldBe runLog(s.get).map(f)
     }
 
+    "mapAsync.exception" in forAll { s: PureStream[Int] =>
+      val f = (_: Int) => IO.raiseError[Int](new RuntimeException)
+      val r = s.get.covary[IO].mapAsync(1)(f).attempt
+      runLog(r).size == 1
+    }
+
     "mapAsyncUnordered" in forAll { s: PureStream[Int] =>
       val f = (_: Int) + 1
       val r = s.get.covary[IO].mapAsyncUnordered(16)(i => IO(f(i)))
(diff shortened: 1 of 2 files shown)


SystemFw:
it would be good to have a test for this scenario, given that it's tricky, but if it's even trickier to write one, never mind

vlovgr:
@SystemFw Looks like there might be more issues at play here.

While trying to get some tests together, I wrote this:

```scala
import cats.effect.IO

class MapAsyncSpec extends Fs2Spec {
  "mapAsync should terminate on exceptions" in {
    Stream
      .iterate(0)(_ + 1)
      .covary[IO]
      .mapAsync(1) { n =>
        if (n == 5) IO.raiseError(new RuntimeException)
        else IO.unit
      }
      .compile
      .drain
      .unsafeRunSync
  }
}
```

which hangs forever, even with the fix in this pull request. It looks like neither finalizers ever get invoked. If the `maxConcurrent` parameter for `mapAsync` / `parEvalMap` is increased enough, I'm no longer able to reproduce it. From my short investigation, it looks like this issue occurs when `queue.enqueue1(Some(value.get))` blocks while the error is raised, and for some reason, finalizers are not invoked, so we can't cancel the blocking. I'm not sure how stream errors are handled internally.

vlovgr:
@SystemFw Thanks for the super quick reply! :) It looks like we're getting stuck in the finalizer from `concurrently`. Do you know why it's getting executed before the one in `parEvalMap`? I thought finalizers should always be called in reverse order, but that does not seem to be the case here.

vlovgr:
@SystemFw I've added a test case for the finalizer behaviour, which is what previously prevented stream interruption (and which prevented it when I got the order of the finalizers wrong above). It does not test that the blocking `enqueue` is cancelled in `evalMap`, and I've not been able to write a test for it, but have manually managed to verify that it can happen, and that this fixes it.

PR #1312, opened 2018-11-08
groupWithin leak

### LINKED ISSUE #1303: Memory leak with `groupWithin`
Hi all, I have observed a memory leak problem in my long-lived stream application.
After doing various experiments, I located the `groupWithin`
My sample:
```scala
Stream.eval(IO.never).covary[IO]
.groupWithin(Int.MaxValue, 1 millis)
.compile.drain.unsafeRunSync()
```
YourKit profiling(the old gen keeps growing and OutOfMemery finally):
img_20181101_183624

I am using this function to perform a time windowing in my stream. Is there any workaround in case the groupWithin is leaking ?
Kindly correct me if anything I missed.

SystemFw:
Attempts to fix #1303

In particular, it definitely fixes the original leak, and it appears to not leak outstanding timeouts either, but I'm not 100% sure about how airtight the resource safety on that timeout being canceled is. 
It kinda relies on `onFinalise` not running until the `finaliser` in the `bracket` in `startTimeout` has completed. Ideas welcome

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4,6 +4,7 @@ import cats._
 import cats.data.{Chain, NonEmptyList}
 import cats.effect._
 import cats.effect.concurrent._
+import cats.effect.implicits._
 import cats.implicits.{catsSyntaxEither => _, _}
 import fs2.concurrent._
 import fs2.internal.FreeC.Result
@@ -1111,60 +1112,62 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
       implicit timer: Timer[F2],
       F: Concurrent[F2]): Stream[F2, Chunk[O]] =
     Stream
-      .eval(Queue.synchronousNoneTerminated[F2, Either[Token, Chunk[O]]])
-      .flatMap { q =>
-        import cats.effect.implicits._
-        def startTimeout: Stream[F2, (Token, CancelToken[F2])] =
-          Stream.eval(F.delay(new Token)).evalMap { t =>
-            val fiber = (timer.sleep(d) *> q.enqueue1(t.asLeft.some)).start
-            fiber.map(f => t -> f.cancel)
-          }
+      .eval {
+        Queue
+          .synchronousNoneTerminated[F2, Either[Token, Chunk[O]]]
+          .product(Ref[F2].of(F.unit))
+      }
+      .flatMap {
+        case (q, lastTimeout) =>
+          def startTimeout: Stream[F2, Token] =
+            Stream.eval(F.delay(new Token)).evalTap { t =>
+              val timeout = timer.sleep(d) *> q.enqueue1(t.asLeft.some)
 
-        def producer = this.chunks.map(_.asRight.some).to(q.enqueue).onFinalize(q.enqueue1(None))
-
-        def emitNonEmpty(c: Chain[Chunk[O]]): Stream[F2, Chunk[O]] =
-          if (c.nonEmpty) Stream.emit(Chunk.concat(c.toList))
-          else Stream.empty
-
-        def resize(c: Chunk[O], s: Stream[F2, Chunk[O]]): (Stream[F2, Chunk[O]], Chunk[O]) =
-          if (c.size < n) s -> c
-          else {
-            val (unit, rest) = c.splitAt(n)
-            resize(rest, s ++ Stream.emit(unit))
-          }
-
-        def go(acc: Chain[Chunk[O]],
-               elems: Int,
-               currentTimeout: Token,
-               cancelCurrentTimeout: CancelToken[F2]): Stream[F2, Chunk[O]] =
-          Stream.eval(q.dequeue1).flatMap {
-            case None => emitNonEmpty(acc)
-            case Some(e) =>
-              e match {
-                case Left(t) if t == currentTimeout =>
-                  emitNonEmpty(acc) ++ startTimeout.flatMap {
-                    case (newTimeout, newCancel) =>
-                      go(Chain.empty, 0, newTimeout, newCancel)
-                  }
-                case Left(t) if t != currentTimeout =>
-                  go(acc, elems, currentTimeout, cancelCurrentTimeout)
-                case Right(c) if elems + c.size >= n =>
-                  val totalChunk = Chunk.concat((acc :+ c).toList)
-                  val (toEmit, rest) = resize(totalChunk, Stream.empty)
-
-                  toEmit ++ Stream.eval_(cancelCurrentTimeout) ++ startTimeout.flatMap {
-                    case (newTimeout, newCancel) =>
-                      go(Chain.one(rest), rest.size, newTimeout, newCancel)
-                  }
-                case Right(c) if elems + c.size < n =>
-                  go(acc :+ c, elems + c.size, currentTimeout, cancelCurrentTimeout)
+              timeout.start.bracket(_ => F.unit) { fiber =>
+                lastTimeout.getAndSet(fiber.cancel).flatten // start the cancel? (now using flatten)
               }
-          }
+            }
 
-        startTimeout.flatMap {
-          case (t, c) =>
-            go(Chain.empty, 0, t, c).concurrently(producer)
-        }
+          def producer = this.chunks.map(_.asRight.some).to(q.enqueue).onFinalize(q.enqueue1(None))
+
+          def emitNonEmpty(c: Chain[Chunk[O]]): Stream[F2, Chunk[O]] =
+            if (c.nonEmpty) Stream.emit(Chunk.concat(c.toList))
+            else Stream.empty
+
+          def resize(c: Chunk[O], s: Stream[F2, Chunk[O]]): (Stream[F2, Chunk[O]], Chunk[O]) =
+            if (c.size < n) s -> c
+            else {
+              val (unit, rest) = c.splitAt(n)
+              resize(rest, s ++ Stream.emit(unit))
+            }
+
+          def go(acc: Chain[Chunk[O]], elems: Int, currentTimeout: Token): Stream[F2, Chunk[O]] =
+            Stream.eval(q.dequeue1).flatMap {
+              case None => emitNonEmpty(acc)
+              case Some(e) =>
+                e match {
+                  case Left(t) if t == currentTimeout =>
+                    emitNonEmpty(acc) ++ startTimeout.flatMap { newTimeout =>
+                      go(Chain.empty, 0, newTimeout)
+                    }
+                  case Left(t) if t != currentTimeout => go(acc, elems, currentTimeout)
+                  case Right(c) if elems + c.size >= n =>
+                    val totalChunk = Chunk.concat((acc :+ c).toList)
+                    val (toEmit, rest) = resize(totalChunk, Stream.empty)
+
+                    toEmit ++ startTimeout.flatMap { newTimeout =>
+                      go(Chain.one(rest), rest.size, newTimeout)
+                    }
+                  case Right(c) if elems + c.size < n =>
+                    go(acc :+ c, elems + c.size, currentTimeout)
+                }
+            }
+
+          startTimeout
+            .flatMap { t =>
+              go(Chain.empty, 0, t).concurrently(producer)
+            }
+            .onFinalize(lastTimeout.get.flatten)
       }
 
   /**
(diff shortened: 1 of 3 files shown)


SystemFw:
(I'd recommend to ignore whitespace changes)

SystemFw:
Ok we have a problem, at least with current stable versions of fs2 and cats-effect (there's a release idempotency bug in cats-effect, fixed in master)

```scala
  def w =
    Stream.eval {
      IO.sleep(3.seconds).bracket(_ => IO.unit)(_ => IO(println("IO done")))
    }
      .interruptAfter(1.second)
      .onFinalize(IO(println("stream done")))
      .compile.drain.unsafeRunSync
```

```
  scala> w
stream done

scala> w
IO done
stream done

scala> w
stream done
IO done

scala> w
IO done
stream done

scala> w
stream done
IO done

scala> w
IO done
stream done
IO done

scala> w
stream done

scala> w
IO done
stream done
IO done

scala> w
stream done
IO done
scala> w
stream done

scala> w
IO done
stream done

scala> w
stream done
IO done

scala> w
IO done
stream done

scala> w
stream done
IO done

scala> w
IO done
stream done
IO done

scala> w
stream done

scala> w
IO done
stream done
IO done

scala> w
stream done
IO done
```

There's all sorts of issues here:
- Sometimes the IO finaliser runs twice, that's probably the bug in cats effect
- Sometimes the IO finaliser doesn't run at all, which is scary. That could happen if the stream gets interrupted before the finaliser is installed, but the stream is being interrupted after 1 whole second, that should be plenty of time
- There's no ordering between the IO finaliser and the stream finaliser, which means in my code there will be a race condition between storing the current `.cancel` in the `Ref`, and the stream finaliser getting it from the Ref and cancelling it.

I think I might have to think of something else for this specific issue, but this is worth investigating regardless. Keen to hear your thoughts @mpilquist @pchlupacek

SystemFw:
Just to be clear, just removing `supervise` does improve the situation, the combinator is usable, but there will be a slow leak in timeouts, and/or not full resource safety on them

pchlupacek:
@SystemFw you sort of confirm my suspicion that `IO.bracket` and cancellation doesn't play well sometimes. 

I will look closely on groupWithin implementation and see if somehow issues can be break down and minimalised. To be groupWithin was always scary (:-) ) so I need some time for this.

Would be excellent that if we can minimalise the issues you mentioning with simple examples so we can pinpoint whether this is fs2 issue or really cats.effects.

SystemFw:
I think I have a better solution, which _does_ assume that IO bracket finalisers will be run, but does _not_ assume any ordering between IO finalisers and Stream finalisers, they can be concurrent. I will push it in a few minutes, just need to run the sanity tests to ensure there are no leaks again (and then we need to think about resource safety tests)

mpilquist:
Okay a couple of things that might help explain the behaviors you are seeing.

The first is that the acquire that's passed to `F.bracket` is made uncancelable, so the following expression cannot be canceled:

`IO.sleep(3.seconds).bracket(_ => IO.unit)(_ => IO(println("IO done")))`

The second is that we support interruption and finalization of streams with hung effect evaluations -- e.g. `Stream.eval(IO.never).interruptAfter(...)`. As a consequence of this, when the original stream is interrupted, the sleep continues to execute.

With the `w` program above, I *always* get the output of:
```
> w
(1 second delay)
stream done
(2 second delay)
IO done
```

SystemFw:
> you sort of confirm my suspicion that IO.bracket and cancellation doesn't play well sometimes.

Looking again at the example above, I actually think it does.
Canceling `eval` does not cancel the `IO` there all the time, so I think the IO finaliser will just run 3 seconds after. I think my solution for `groupWithin` should work I think, posting soon

SystemFw:
> With the w program above, I always get the output of:

Yeah just realised this :)
Therefore I think what I'm about to push will work, but please review very carefully

mpilquist:
Looks like the spawn and metered methods were accidentally deleted in the PR

SystemFw:
oh sorry, I didn't even see that, I'll fix it now

SystemFw:
Must have been some weird rebase errors because some comments were removed too, probably forked this branch from an old version of master

SystemFw:
Ok, history fixed

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1111,51 +1112,85 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+        case (q, currentTimeout) =>
+          def startTimeout: Stream[F2, Token] =
+            Stream.eval(F.delay(new Token)).evalTap { t =>
+              val timeout = timer.sleep(d) *> q.enqueue1(t.asLeft.some)
+
+              // Just using `supervise(timeout)` will ensure resource safety
+              // but finalisers accumulate and cause a leak, so we
+              // need to dispose of outstanding timeouts manually
+              timeout.start
+                .bracket(_ => F.unit) { fiber =>
```
mpilquist: BTW, another technique is to use `Pull.acquireCancellable` or `Stream.bracketWithToken`, which lets us invoke a finalizer early *and* unregister it from the resource map.

SystemFw: > which lets us invoke a finalizer early

ah! I actually didn't think of those combinators, but they sound promising. I might see how things look like with them.

mpilquist: `bracketWithToken` is private[fs2] and fairly low level but a `bracketCancellable` might be useful to add to the API. Take a look at the implementation of `Pull.acquireCancellable` for an example of how to use `bracketWithToken` -- of particular note is the `release` call, which turns in to an `Algebra.Release(t)`, which is what prevents the finalizer leak.

SystemFw: However, wouldn't that need a similar machinery to the code above, to  ensure that `release` is always called?

SystemFw: Mm maybe not, it's not like `lease` right?
it's still bracketed, and _in addition to that_ it allows early finalisation?
if that's the case, it would be nice, but I'd rather we added a higher-level version to the public api before using it here

mpilquist: Right. If `release(token)` is never called, the finalizer will still run at the normal place it would run with `bracket`. Note the similar in implementation: https://github.com/functional-streams-for-scala/fs2/blob/9a4074df5763341b82828c02d6eb9db5c5c260c9/core/shared/src/main/scala/fs2/Stream.scala#L2555-L2572

SystemFw: Yeah, makes sense. I will still need the `Ref` to store the previous `(Token, R)`, but I could then deal with running those in the happy path with a simple flatMap, given that in case of problems (basically interruption) it will still be run

SystemFw: Wishing I had known this before thinking so hard about synchronisation in the current code :P

Anyway, I'll add a `bracketCancelable` (or `earlyBracket` or whatever) in another PR, and then probably port this to use that instead, if the approach outlined in the comment right above makes sense to you

mpilquist: Here's an example of a public API we could add:

```scala
  def bracketCaseCancellable[F[x] >: Pure[x], R](acquire: F[R])(
      release: (R, ExitCase[Throwable]) => F[Unit]): Stream[F, (Stream[F, INothing], R)] =
    bracketWithToken(acquire)(release).map {
      case (token, r) => (Stream.fromFreeC(Algebra.release(token, None)), r)
    } 
```

SystemFw:
Waiting on  #1317

SystemFw:
Minimal example: same recursion pattern within a `flatMap`, early finalisation of the previous computation via `bracketWithToken`, still leaks

```scala
object MinST7 extends App {
  import cats.effect.concurrent._
  import fs2.internal._

  implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
  implicit val timer: Timer[IO] = IO.timer(ExecutionContext.global)

  val d = 1.millis

  Stream
    .eval(Ref[IO].of(Stream.eval(IO.unit)))
    .flatMap { ref =>
      val prog: IO[Unit] = timer.sleep(d)

      def startTimeout: Stream[IO, Unit] =
        Stream
          .bracketWithToken(prog) {
            case (unit, exitCase) =>
              IO(println("release"))
          }
          .flatMap {
            case (t, _) =>
              val release = Stream.fromFreeC(Algebra.release(t, None)) ++ Stream.emit(())

              Stream.eval(ref.getAndSet(release)).flatten
          }

      def read: Stream[IO, Unit] = startTimeout.flatMap(_ => read)

      read
    }
    .compile
    .drain
    .unsafeRunSync
}
```

mpilquist:
Hm, I suspect the `Algebra.release` call ends up being a noop b/c it doesn't find that token in the current scope - and it remains in the original scope.

SystemFw:
> Hm, I suspect the Algebra.release call ends up being a noop b/c it doesn't find that token in the current scope - and it remains in the original scope.

That was my guess as well initially, but the thing is that it does print "release".

Minimised that behaviour here https://github.com/functional-streams-for-scala/fs2/compare/series/1.0...SystemFw:groupWithin/leakBracketWithToken?expand=1#diff-124fdf506a5147d4b51ef4f94ae4ae56R211

SystemFw:
I don't think it would print "release" if the resource was not in the current scope anymore
```scala
 def releaseResource(id: Token, ec: ExitCase[Throwable]): F[Either[Throwable, Unit]] =
    F.flatMap(state.modify { _.unregisterResource(id) }) {
      case Some(resource) => resource.release(ec)
      case None           => F.pure(Right(())) // resource does not exist in scope any more.
    }
```

mpilquist:
@SystemFw Okay, I'm no longer convinced that we can/should support recursive brackets. Each bracket will install a finalizer in to the overall stream structure, regardless of whether we clean up the resource or not.

Here's a smaller reproducer:

```scala
object ExplicitRelease extends App {
  def go: Stream[IO, Unit] = Stream.bracketWithToken(IO.unit)((_, _) => IO.unit).flatMap {
    case (t, _) =>
      Stream.fromFreeC(internal.Algebra.release(t, None)) ++ go
  }
  go.compile.drain.unsafeRunSync()
}
```

This correctly removes the resource from the resources collection but runs out of memory anyway, because of the tail-accumulated release calls.

SystemFw:
Ok, I have a series of questions :) (sorry)

- How did you figure out the problem was with bracket in the first place? Profiling mostly shows `FreeC.Bind`, and I guess one could use the usual technique of playing with the code, minimising until the bug disappears, but I'm interested on whether you had a specific clue _from the memory analysis_ (trying to get better at this)

- What should we do for this specific combinator? I think the solution in this PR works (no leak, timeouts are resource safe), if a bit ad hoc

- More in general, should we support early finalisation in a way that makes the snippet you just posted work? It's a rare case, but iirc other streaming libraries had to deal with this problem as well (specifically `pipes`)

mpilquist:
> How did you figure out the problem was with bracket in the first place? Profiling mostly shows FreeC.Bind, and I guess one could use the usual technique of playing with the code, minimising until the bug disappears, but I'm interested on whether you had a specific clue from the memory analysis (trying to get better at this)

I spot checked the live instances count of various classes -- `FreeC.Bind`, closures, tokens, and `Algebra.Release` instances all roughly had the same count (there might have been a 2x factor in there somewhere but there was clearly a linear relationship). From there, I theorized about what could be keeping `Algebra.Release` instances live on the heap, in a linear relationship with `Bind` instances. I added printlns and confirmed the resource was truly getting removed from the scope (and noticed I didn't see `Resource` instances in the heap).

> What should we do for this specific combinator? I think the solution in this PR works (no leak, timeouts are resource safe), if a bit ad hoc

I think it's good as-is.

> More in general, should we support early finalisation in a way that makes the snippet you just posted work? It's a rare case, but iirc other streaming libraries had to deal with this problem as well (specifically pipes)

Perhaps? If we want to support it, we'd have to fundamentally change our approach to installing the `Release` call.

SystemFw:
Thanks for the explanation. I think I need to get a bit more used to thinking about the underlying low level structure. I think the idea of having a debug print interpreter for Stream is still valid :)

Do you think it's a good idea to add a few comments to that `startTimeout` implementation?  I have some, but maybe explaining the problem a bit more will help clarity. Also, did you have time to run the memory safety tests I've added? I think the implementation is solid now, but it's been tricky.

> If we want to support it, we'd have to fundamentally change our approach to installing the Release call.

Probably best to put it on hold for now :)

PR #1317, opened 2018-11-09
Add bracketCancellable

SystemFw:
Useful for cases like #1312 
Better have something like this in the public than creating an Algebra node in a high level operation like `groupWithin`

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2559,6 +2559,30 @@ object Stream extends StreamLowPriority {
         Stream.emit(r).covary[F].get[F, R].transformWith(bracketFinalizer(token))
     })
 
+  /**
+    * Like [[bracket]] but the result value consists of a cancellation
+    * Stream and the acquired resource. Running the cancellation Stream frees the resource.
+    * This allows the acquired resource to be released earlier than at the end of the
+    * containing Stream scope.
+    * Note that this operation is safe: if the cancellation Stream is not run manually,
+    * the resource is still guaranteed be release at the end of the containing Stream scope.
+    */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2559,6 +2559,30 @@ object Stream extends StreamLowPriority {
...
+  /**
+    * Like [[bracketCancellable]] but the release action is passed an `ExitCase[Throwable]`.
+    *
+    * `ExitCase.Canceled` is passed to the release action in the event of either stream interruption or
+    * overall compiled effect cancelation.
+    */
+  def bracketCaseCancellable[F[x] >: Pure[x], R](acquire: F[R])(
+      release: (R, ExitCase[Throwable]) => F[Unit]): Stream[F, (Stream[F, INothing], R)] =
+    bracketWithToken(acquire)(release).map {
+      case (token, r) => (Stream.fromFreeC(Algebra.release(token, None)), r)
```
SystemFw: I'm not sure about the `None` here

SystemFw: I've come to the conclusion that `None` is fine there, given that the canceler stream we are building here is not linked to the termination of the bracketed stream, and actually can't be because that defeats the point of an early finaliser. I therefore think this is ready for merge.

SystemFw:
I decided to change the finaliser from `Stream[F, INothing]` to `Stream[F, Unit]`, because the former proved error prone and counterintuitive, e.g in 

```scala
...
Stream.emit(1).flatTap(releaseStreamFromBracketCancelable)
```

Which would compile and return an empty stream.

On a separate note, it's not clear whether the initial motivation for making this combinator public (a leak in `groupWithin`) is actually fixable with this, so we also need to decide if having this in the public api is useful in general

PR #1322, opened 2018-11-17
Removed ConcurrentEffect from tcp/udp packages - NOT FOR MERGE

mpilquist:
This is a proof of concept PR which removes `ConcurrentEffect` from the `tcp` and `udp` packages. I'd like to also remove it from `JavaInputOutputStream` if possible. Assuming we're good with this approach, I'll figure out how to update this PR to preserve binary compatibility.

cc @tpolecat

### DIFF
--- a/io/src/main/scala/fs2/io/tcp/Socket.scala
+++ b/io/src/main/scala/fs2/io/tcp/Socket.scala
@@ -17,7 +17,7 @@ import java.nio.channels.{
 import java.util.concurrent.TimeUnit
 
 import cats.implicits._
-import cats.effect.{ConcurrentEffect, Resource}
+import cats.effect.{Async, Concurrent, ContextShift, Resource}
 import cats.effect.concurrent.{Ref, Semaphore}
 
 import fs2.Stream._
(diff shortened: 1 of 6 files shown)


io/src/main/scala/fs2/io/tcp/Socket.scala
```diff
@@ -116,18 +117,18 @@ protected[tcp] object Socket {
...
             def completed(result: Void, attachment: Void): Unit =
-              invokeCallback(cb(Right(ch)))
+              cb(Right(ch))
             def failed(rsn: Throwable, attachment: Void): Unit =
-              invokeCallback(cb(Left(rsn)))
+              cb(Left(rsn))
           }
         )
-      }
+      } <* cs.shift
```
SystemFw: I think we could do 
```scala
def yieldBack = F.start(F.unit).flatMap(_.join)
```
and only require a `Concurrent` constraint, which imho better represents what we're doing here.

SystemFw: Ah, and probably we want to yield back in a `bracket`? (although I appreciate this is a POC PR :) )

mpilquist: @SystemFw Fixed now, let me know what you think

SystemFw:
@mpilquist The only other remaining thing I'm not sure about, it's between `<* yieldBack` and `.guarantee(yieldBack)`

SystemFw:
I'm not entirely sure that `fa.start.bracket(_.join)(_.cancel)` guarantees that execution is shifted back to when we want it, given that `fa` is async itself. `fa.guarantee(yieldBack)` seems more straightforward in expressing "after the async task is done, go back to the runtime"

mpilquist:
Gotcha, fixed

mpilquist:
Alright this one is ready for final review

PR #1323, opened 2018-11-18
Fix thread pool leak in README

kubukoz:


README.md
```diff
@@ -24,19 +24,21 @@ Quick links:
...
 import cats.implicits._
 import fs2.{io, text}
 import java.nio.file.Paths
 import java.util.concurrent.Executors
 import scala.concurrent.ExecutionContext
 
 object Converter extends IOApp {
-  def run(args: List[String]): IO[ExitCode] = {
+  private val blockingExecutionContextResource =
+    Resource.make(IO(ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(2))))(ec => IO(ec.shutdown()))
```
SystemFw: _Maybe_ `Stream.bracket` here, since we're on topic?

SystemFw: Or perhaps keep the Resource there, but `Stream.resource` rather than `use` later on

kubukoz: Fine with me, but I prefer to use Resource if there's a single thing it'll emit, plus we'll have to carry it around the whole thing.

kubukoz: Just saw your second comment, that still leaves the second issue...

SystemFw: > Fine with me, but I prefer to use Resource if there's a single thing it'll emit,

Yeah having Resource there it's a good idea, since that's how we expose our apis as well now (e.g. `Socket`), but I'd rather showcase `Stream.resource` since we're at it, given that in many cases you want to use a Resource but stay in Stream

SystemFw:
```scala
object Converter extends IOApp {

  val blockingExecutionContext: Resource[IO, ExecutionContext] =
    Resource.make(IO(ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(2))))(ec => IO(ec.shutdown()))

  def converter: Stream[IO, Unit] = blockingExecutionContext.flatMap { blockingEC =>
    def fahrenheitToCelsius(f: Double): Double = (f - 32.0) * (5.0/9.0)

    io.file.readAll[IO](Paths.get("testdata/fahrenheit.txt"), blockingEC, 4096)
      .through(text.utf8Decode)
      .through(text.lines)
      .filter(s => !s.trim.isEmpty && !s.startsWith("//"))
      .map(line => fahrenheitToCelsius(line.toDouble).toString)
      .intersperse("\n")
      .through(text.utf8Encode)
      .through(io.file.writeAll(Paths.get("testdata/celsius.txt"), blockingEC))
  }


    converter.compile.drain.as(ExitCode.Success)
}
```

Something like that perhaps?

kubukoz: Done, I also changed the ReadmeExample tut doc - please take a look, especially at the ending part (in which I close the EC that's allocated somewhere in the middle of the page)

kubukoz: Should be good now, I'll take a look tomorrow and see if I can come up with some good description of what's going on with the EC.

docs/ReadmeExample.md
```diff
@@ -59,6 +61,9 @@ import java.util.concurrent.Executors
 import scala.concurrent.ExecutionContext
 
 implicit val cs: ContextShift[IO] = IO.contextShift(scala.concurrent.ExecutionContext.Implicits.global)
+
+//note: this should be shut down when it's no longer necessary - normally that's at the end of your app.
+//See the whole README example for proper resource management in terms of ExecutionContexts.
 val blockingExecutionContext = ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(2))
```
Daenyth: This is a little confusing to me in that the changes above name `blockingExecutionContext: Resource[F, EC]` but this is `blockingExecutionContext: EC`.

I'm concerned that a reader unfamiliar with `Resource` won't quite understand the implications and end up with opening a resource multiple times instead of opening it once and then sharing it.

Perhaps we might name the `Resource` versions with `make` or `get` or something like that in the name?

PR #1324, opened 2018-11-19
Expands the ScalaDocs for the Stream class.

diesalbla:
We extend the ScalaDocs description of the Stream class.

- We adds some descriptions for the notions of pull-based evaluation, and laziness. We show some special cases of Streams.
- We add a section describing the equivalence between some of the methods of the Stream class, and similar methods in the List class. This equivalences hold true for finite and pure streams.
- We add into the docs of some of methods a description of how the methods handles non-terminating streams.

Thanks to @SystemFw for clarifying some doubts and questions for this PR.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
 
 /**
   * A stream producing output of type `O` and which may evaluate `F`
-  * effects. If `F` is [[Pure]], the stream evaluates no effects.
+  * effects.
+  * Or: a lazy sequence of computations that may evaluate `F` effects `F` and yield values of type `O`.
+  *
+  * Streams are purely functional: a `Stream[F, O]` is a description for a computation,
+  * much like `cats.effect.IO` or `scalaz.Task` are just descriptions of computations.
+  * The methods in the `Stream` class only calculate new descriptions from others.
+  *
+  * Streams are lazy, pull-based data sources. To '''evaluate''' a stream is to extract from it
+  * as many elements as a _consumer_ of the stream may need for the result.
+  *   by performing as many `F`-effectful computations as necessary to get those elements.
+  *   Thus, the effects in `F` are not evaluated unless and until they are required (pulled) by a stream consumer.
+  *
+  * - '''Laziness''': streams are lazily evaluated sources for effectful computations.
+  *   The operations on a Stream only evaluate a finite prefix,
+  *   a finite part of a whole stream, as needed to supply the requests of the stream _consumers_.
+  *   Thus, there can be infinite streams, or Streams based on
+  *   an unlimited source of information (e.g. a Signal that may be sampled with as frequently as desired).
+  *
+  * - '''Source''': a stream may simply be a fixed list of pre-defined F-effectful computations.
+  *   However, it can also be backed by an external data source (e.g. a network connection).
+  *   It can also be backed by a effectful computation that can be re-evaluated,
+  *   such as a Stream to read the current system time.
+  *
+  * There are some special properties or cases of streams:
+  *  - A '''pure''' stream is a `Stream` in which the `F` is [[Pure]], which evaluates no effects.
+  *  - A stream is '''terminating''', or '''finite''', if pulling from it reaches the end after yielding
+  *    a finite number of elements.
+  *  - In particular, a stream is ''''empty''' if it terminates after yielding no value.
+  *  - A '''singleton''' stream is a stream that ends after yielding one single value.
+  *  - A '''never''' stream is a stream that never terminates and never yields any value.
+  *
+  * == List-Like API and operations ==
+  *
+  * A `Stream[F, O]` can be (naively) seen as a list of elements with `F`-effects.
+  * Following the analogy between lists and streams, the API of the `Stream` class
+  * defines several methods similar to those of the `List` class.
+
+  * We can convert every ''pure and finite''' stream into a `List[O]` with the `.toList` method.
+  * This conversion obeys some equivalences between `Stream` and `List` operations.
+  *
+  *  - For any pure stream `s`, `s.foo().toList == s.toList.foo()`, is true if `foo` is one of `map`, `as`, `delete`, `take`, `drop`, `takeWhile`, `dropWhile`, `zipWithIndex`, `filter`, or `collect`.
+  *  - Likewise, `s.foo().toList == List(s.toList.foo() )` holds if `foo` is one of:  `exists`, `forall`, `fold`.
+  *  - Other equalities of the form `leftSide.toList == rightSide` are shown in this table, where we use `SP[A]` as alias for `Stream[Pure, A]`
+  *
+  * | Method    | Left side         | Right side | Variables |
+  * | --------- | ----------------: | :----------------------------- | -------- |
+  * | head      | `s.head`          | `s.toList.headOption.toList` |
+  * | last      | `s.last`          | `s.toList.lastOption.toList` |
+  * | append    | `s1 ++ s2`        | `s1.toList ++ s2.toList`  | `s1, s2: SP[O]` |
+  * | takeRight | `s.takeRight(n)`  | `s.toList.reverse.take(n).reverse` | `n: Int` |
+  * | dropRight | `s.dropRight(n)`  | `s.toList.reverse.drop(n).reverse` | `n: Int` |
+  * | takeThrough | `s.takeThrough(n)`  | `s.toList.takeWhile(p) ++ s.toList.dropWhile(p).take(1)` | `p: O => Boolean` |
+  * | dropThrough | `s.dropThrough(n)`  | `s.toList.dropWhile(p).drop(1)` | `p: O => Boolean` |
+  * | flatMap   | `s.flatMap(f)`    | `s.toList.flatMap(f andThen toList)`  | `s: SP[O]`, and `f: O => SP[B]` |
+  * | flatten   | `ss.flatten`      | `s.toList.flatMap(_.toList)`  | `ss: SP[SP[O]]` |
+  * | zip       | `s.zip(t)`        | `s.toList zip t.toList`     | `s: SP[O]` and `t: SP[B]` |
+  * | zipWith   | `s.zipWith(t)(f)` | `(s.toList zip t.toList).map(f)`  | `s: SP[O]`, `t: SP[B]`, and `f: (O,B) => C` |
+  * | zipWithPrevious | `s.zipWithPrevious` | `(None :: s.toList.map(Some)) zip s.toList` |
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
 
 /**
   * A stream producing output of type `O` and which may evaluate `F`
-  * effects. If `F` is [[Pure]], the stream evaluates no effects.
+  * effects.
+  * Or: a lazy sequence of computations that may evaluate `F` effects `F` and yield values of type `O`.
```
mpilquist: Can we combine the first sentence with this one (the one that starts with Or) -- they aren't different enough to both exist IMO.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  *
+  * Streams are purely functional: a `Stream[F, O]` is a description for a computation,
```
mpilquist: s/description for/description of/

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  * much like `cats.effect.IO` or `scalaz.Task` are just descriptions of computations.
```
mpilquist: I'd replace the reference to `scalaz.Task` with Monix and ZIO perhaps. The old scalaz task was `scalaz.concurrent.Task` but that's ancient history now.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  * The methods in the `Stream` class only calculate new descriptions from others.
+  *
+  * Streams are lazy, pull-based data sources. To '''evaluate''' a stream is to extract from it
+  * as many elements as a _consumer_ of the stream may need for the result.
+  *   by performing as many `F`-effectful computations as necessary to get those elements.
+  *   Thus, the effects in `F` are not evaluated unless and until they are required (pulled) by a stream consumer.
+  *
+  * - '''Laziness''': streams are lazily evaluated sources for effectful computations.
```
mpilquist: Seems like there's a sentence missing before this bulleted list starts, perhaps introducing these properties.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  *   The operations on a Stream only evaluate a finite prefix,
+  *   a finite part of a whole stream, as needed to supply the requests of the stream _consumers_.
```
mpilquist: I think consumers should be consumer here, since evaluation is controlled by a single fold.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  * much like `cats.effect.IO` or `scalaz.Task` are just descriptions of computations.
+  * The methods in the `Stream` class only calculate new descriptions from others.
+  *
+  * Streams are lazy, pull-based data sources. To '''evaluate''' a stream is to extract from it
+  * as many elements as a _consumer_ of the stream may need for the result.
+  *   by performing as many `F`-effectful computations as necessary to get those elements.
+  *   Thus, the effects in `F` are not evaluated unless and until they are required (pulled) by a stream consumer.
+  *
+  * - '''Laziness''': streams are lazily evaluated sources for effectful computations.
+  *   The operations on a Stream only evaluate a finite prefix,
```
mpilquist: Style-wise, I recommen "stream" in prose and `Stream` when referencing the type directly. I would avoid "Stream".

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  *   a finite part of a whole stream, as needed to supply the requests of the stream _consumers_.
+  *   Thus, there can be infinite streams, or Streams based on
+  *   an unlimited source of information (e.g. a Signal that may be sampled with as frequently as desired).
+  *
+  * - '''Source''': a stream may simply be a fixed list of pre-defined F-effectful computations.
```
mpilquist: I want to be careful with the term Source as it doesn't appear in FS2 proper but it does appear in other libraries. E.g., once you introduce this term, then both `Stream` and `Pull` act as sources while the compilation actions (e.g. compile.drain) act as sinks.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  *   However, it can also be backed by an external data source (e.g. a network connection).
+  *   It can also be backed by a effectful computation that can be re-evaluated,
```
mpilquist: "It can also be backed by" shows up twice here. Reword one.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  *   such as a Stream to read the current system time.
+  *
+  * There are some special properties or cases of streams:
+  *  - A '''pure''' stream is a `Stream` in which the `F` is [[Pure]], which evaluates no effects.
+  *  - A stream is '''terminating''', or '''finite''', if pulling from it reaches the end after yielding
+  *    a finite number of elements.
+  *  - In particular, a stream is ''''empty''' if it terminates after yielding no value.
```
mpilquist: Remove "In particular" to match the style of this list?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  *  - A '''singleton''' stream is a stream that ends after yielding one single value.
+  *  - A '''never''' stream is a stream that never terminates and never yields any value.
```
mpilquist: Is `s.drain` a never stream using this classification system? I.e. s may evaluate arbitrary effects forever, never emitting a value.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -15,7 +15,67 @@ import scala.concurrent.duration._
...
+  * == List-Like API and operations ==
+  *
+  * A `Stream[F, O]` can be (naively) seen as a list of elements with `F`-effects.
+  * Following the analogy between lists and streams, the API of the `Stream` class
+  * defines several methods similar to those of the `List` class.
+
+  * We can convert every ''pure and finite''' stream into a `List[O]` with the `.toList` method.
+  * This conversion obeys some equivalences between `Stream` and `List` operations.
+  *
+  *  - For any pure stream `s`, `s.foo().toList == s.toList.foo()`, is true if `foo` is one of `map`, `as`, `delete`, `take`, `drop`, `takeWhile`, `dropWhile`, `zipWithIndex`, `filter`, or `collect`.
```
mpilquist: I'm not too keen on enumerating these methods here or in the table below. I think they will end up out of sync as new methods are added. Rather, I think this section can show some examples but shouldn't attempt to be exhaustive.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1000,6 +1085,9 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
   /**
     * Alias for `map(f).foldMonoid`.
     *
+    * @returns A singleton stream whose value is the result of the fold, if `this` is a finite stream.
```
mpilquist: Might be better to elide this one as the docs for foldMonoid have the full details.

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: Perhaps we could add one phrase on _Pure Semantics_ on each method's Scaladoc? The examples sort-of show that, but writing the equivalence would also be more informative.

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: :+1: I would advise against "sequence of computations" btw, it's subtle, but it can be misleading. There are some key differences between `Stream[F, A]` and `List[F[A]]`, for example the fact that you can always recover and continue from error given `List[F[A]]`, but you have to terminate the Stream with `Stream[F, A]`.

This is because subsequent computations in Stream might depend on previous ones monadically (from a data structure perspective, a Stream is closer to a tree of computations, given that it's a Free monad).

However, none of this should make into the scaladoc imho (you still want it to be concise and to the point), so I'd stick with "a computation emitting output of type `O` and evaluating `F` effects"

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: Yes.

I used _never_ after the `Stream.never` method, but I see that evaluation of effects is orthogonal to yielding values, so there can be a stream with unlimited valueless effects, or with effects with values. IIUC, however, a stream can never yield a value without at least performing one effect, so every yield value comes out of an effect.

The reason I want to differentiate between a _never_-yielding computation and an infinite one is that  lazy operations, like `take`, always yield values when applied to infinite non-failing stream; others, like `exists` may not yield a value out of an infinite stream or not; and others, like `fold`, yield nothing until they reach the stream's end.

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: Yes, I will indicate that the absence of explicit constraints is a tacit description of purity.

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: Done.

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: Perhaps "lazy monadic chain of computations that evaluate effects in `F` and may yield `O` values"?
Monadic, of course, on the `FreeC`, not on the `F` itself.

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: Done.

diesalbla: Rewritten.

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: Done. Replaced this point with `abstract`, to suggest that streams can be fetching data from a repeating computations, an input-output processes, a finite in-memory data structures, or "continuously" reading from a global mutable indistinctly.

PR #1325, opened 2018-11-20
Internal changes to the implementation of `translate`.

diesalbla:
We make some changes to the internal Stream implementation, in the `Algebra` file, to avoid using implicit classes.

- We remove the syntax extension implicit class `AlgScopeSyntax`, which was not being used anywhere.
- We remove the syntax extension `AlgEffectSyntax`, and extract its only method as an internal private function. We move this method to the bottom of the file, closer to the related _translate_ methods. We replace the only two uses needed for this function.
- Without `AlgScopeSyntax`, the sub-trait `AlgScope` is not used anywhere, so we can remove it and replace it with `AlgEffect`.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -28,44 +28,18 @@ private[fs2] object Algebra {
       extends AlgEffect[F, O, Unit]
 
   final case class OpenScope[F[_], O](interruptible: Option[Concurrent[F]])
-      extends AlgScope[F, O, Token]
+      extends AlgEffect[F, O, Token]
 
   // `InterruptedScope` contains id of the scope currently being interrupted
   // together with any errors accumulated during interruption process
   final case class CloseScope[F[_], O](scopeId: Token,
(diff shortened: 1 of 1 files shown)


djspiewak:
> to avoid using implicit classes

Why? If you look at the bytecode between those two implementations, I'm relatively certain there is no meaningful difference. `@inline` almost doesn't work at all, and when it does work, most of the time it would have been better if it hadn't, given that it actually works against HotSpot's own inlining in many cases. Furthermore, implicit value classes (such as this one) are elided by the compiler already, and the implicit enrichment turns into an `invokestatic`.

diesalbla:
> Why? 

Because of the style guideline, usually mentioned by M. Odersky, that given two code solutions to a same problem, all things being equal, the simpler one is better. Here, function and calls seems simpler than wrapping that through syntax extension, specially when they are confined to this file.

> `@inline` almost doesn't work at all. 

I have heard. I have dropped that.

mpilquist:
I'm good with replacing extension methods with regular methods for internals for code discoverability & ergonomic reasons. This PR removes 2 extension methods and an unneeded sub-trait, which is a nice simplification.

PR #1326, opened 2018-11-20
Remove the EitherSyntax class

diesalbla:
We remove the `EitherSyntax` implicit class. 

This implicit class was introduced to provide a Right-biased `flatMap` for the `Either` type when compiling with `Scala 2.11`. However, although `fs2` still supports Scala 2.11, the special right-biased `flatMap` was used nowhere in `fs2`. The only method being used was the `toOption`; and since that is was followed with `Option.get`, we can just use `.right.get` instead by way of the `Either.RightProjection` class.

### DIFF
--- a/core/jvm/src/test/scala/fs2/PipeSpec.scala
+++ b/core/jvm/src/test/scala/fs2/PipeSpec.scala
@@ -569,7 +569,7 @@ class PipeSpec extends Fs2Spec {
             .attempt
         }
         r1 should have size (1)
-        r1.head.swap.toOption.get shouldBe an[Err]
+        r1.head.swap.right.get shouldBe an[Err]
         val r2 = runLog {
           s.get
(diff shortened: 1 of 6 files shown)


core/jvm/src/test/scala/fs2/PipeSpec.scala
```diff
@@ -569,7 +569,7 @@ class PipeSpec extends Fs2Spec {
             .attempt
         }
         r1 should have size (1)
-        r1.head.swap.toOption.get shouldBe an[Err]
```
mpilquist: `.swap.right.get` can be `.left.get` here and in the other tests

mpilquist:
Looks good to me -- need to add mima exclusions for the removed stuff. Doing so is safe since that syntax was private to fs2.

PR #1329, opened 2018-11-20
allow creation of an empty Topic

Porcupine96:


pchlupacek:
I was always kind of reluctant on this. We had same issues on Queue I think, perhaps same sort of approach would work there. 

Specifically, I would like to avoid having same type (i.e. Topic) that behaves completely differently, so you can hardly tell in code if it is one that blocks or one that does not.  That will definitely in large programs lead to erroneous code and deadlocks. 

So my suggestion would be to do with Topic exactly what we did in case of Queue, introduce NoneTerminatedTopic as subtype. 

Note that in case of of `Queue` we did `NoneTerminatedQueue` that does not extends `Queue` but does extends `Dequeue1` etc, so it is much safer to know what we are passing around. 

Not sure if all that machinery for topic will be 1.0 compatible, if not I think addition to next major release will be ok.

pchlupacek:
After reading this carefully, this talks about unset Topic. Hence the Topic is sort of Queue + Ref together I am -1 on this, hence Ref cannot be unset as well,  Sorry for misunderstanding the topic here.

mpilquist:
@pchlupacek Counterpoint is that using `Topic[F, Option[A]]` along with `.collect { case Some(a) => a }` is equally prone to deadlocks. What's nice about this PR is that it encapsulates the `Option` pattern without the overhead while not changing the default semantics of `Topic.apply`. Does that make you reconsider?

pchlupacek:
I am sort of still inclining to leave topic as it is. The reason for it is that the value part of the topic is really just `Signal` with stronger guarantees that you do see all changes (instead only discrete changes). 

 I think it is really weird, that there would be different behaviour when you attempt to subscribe before anything was published and AFTER something was published. to topic. 

However I understand the need for being able to register all subscribers and then guarantee by publishing the first value that all subscribers will see first value and then all other following values. That is hard  achievable with todays Topic signature and guarantees.  

As such, I would rather propose to add some constructors to Broadcast, that has similar signature like this (hey not sure about the name though :-): 

``` scala 
def readyN[F, O](minSubscribers: Int): Resource[F, Broadcast[F, O]] = ???

trait Broadcast[F[_], O] extends Topic[F, O] { }
```

That will in fact allow us to synchronise between nondeterministically registering subscribers and on other hand publishing the first value. I believe this will give much better control than allow topic to have unset value. 

Any thoughts?

PR #1332, opened 2018-11-21
Removed ConcurrentEffect constraint from toInputStream

mpilquist:
WIP - going to try removing the Concurrent[IO] constraint next

### DIFF
--- a/io/src/main/scala/fs2/io/JavaInputOutputStream.scala
+++ b/io/src/main/scala/fs2/io/JavaInputOutputStream.scala
@@ -1,32 +1,37 @@
 package fs2
 package io
 
-import scala.concurrent.{SyncVar, blocking}
-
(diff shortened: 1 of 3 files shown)


mpilquist:
This PR uses the technique of implementing the core logic concretely in `IO` and lifting to `F` via `liftIO` as necessary. Unfortunately, this results in needing both a `Concurrent[F]` and `Concurrent[IO]` instance. The latter is used for creating a `Queue` and two `SignallingRef`s.

If `Concurrent[IO]` was globally available, this wouldn't be a problem but that instance is dependent on an `ExecutionContext`. One option is requiring an explicit `ExecutionContext` argument to `toInputStream`.

I'm going to close this since I'm now convinced `ConcurrentEffect[F]` is the best we can do with the current type class hierarchy.

PR #1334, opened 2018-11-24
Copied over (mostly) implementation from fs2-gzip

### LINKED ISSUE #1313: Support for gzip compress/decompress pipes?
scalaz-stream used to have gzip support way back in the day, but it never worked particularly well and it was tied to `Process`'s internals. @fthomas rightfully deleted it quite some time ago. I have resurrected it in a new and totally-not-hacky way! https://github.com/slamdata/fs2-gzip

Is there interest in a PR which moves this into -core? Or perhaps -io? I'm more than happy to do so. We needed it right away, which is the only reason I spun it up as a separate project.

---

For those who wish to code review the project I linked, here's a quick summary of the challenges involved and what exactly is being done to address them.

GZIP is basically an unwrapped inflate datastream wrapped by a CRC header/footer and some other magic bytes. These headers and footers aren't *entirely* trivial, but they are reasonably straightforward. Note that this does *not* mean that inflate with `nowrap = true` is the same as gzip! You cannot decompress the output of `gzip` with `deflate`, nor can you compress data for `gunzip` with `inflate`. They're still different formats.

The `java.util.zip` package contains an implementation of `Inflate` which is able to deal with variable-sized input chunks in an incremental way. By which I mean, you can pass a chunk of bytes to it, and it will process that chunk and give you as much output as it can, updating its state to accept the *next* chunk of bytes and returning control flow. This property (returning control flow and updating internal state) is critical to use in an async context.

By way of contrast, `GZIPInputStream` does *not* exhibit this property. The `InputStream#read` function is expected to block until input is available. You cannot simply return zero bytes to signal "return control flow back to whoever called you", even though that would have been an extremely sensible design for `InputStream` 20+ years ago. Naively, this would seem to imply that the only way to use `GZIPInputStream` within a `Pipe` would be to spin up a separate thread which blocks for `read` calls, context shifting chunks to and from that thread. This would be quite horrible.

Unfortunately, really no one has implemented a particularly good GZIP parser which is async-friendly. Spray comes the closest, and they do so by implementing GZIP header/footer parsing themselves and delegating to `Inflate` for the datastream encoding/decoding. This isn't a horrible strategy, but it makes me a bit nervous since a) it's tied to Spray in some unfortunate ways, and b) it's not used by anyone other than Spray.

`GZIPInputStream`/`GZIPOutputStream` are horrible, but they've been used by all the people since forever. It is desirable to use them *if possible*.

In this case, it *is* possible using a trick similar to what happens inside of Jawn's `AsyncParser`. Basically, we use a static no-trace exception as a signal to suspend an effective continuation, where the continuation closure is saved via mutable state. This is a little trickier than Jawn's approach though, since the mutable state in question is *inside* of `GZIPInputStream`.

The key property which allows this to work is the fact that the only places where `GZIPInputStream` takes control flow (via a method call) and performs reads on the underlying byte stream are strictly arranged into the following pattern:

1. Idempotent operations on mutable state
2. Read, read, read, read
3. Write new mutable state
4. Return

Critically, we do *not* have non-idempotent mutations interleaved with reads. All the non-idempotent writes are at the very end, after all reads are completed.

This allows us to throw an exception whenever a read is attempted with insufficient data available. This exception passes up through `GZIPInputStream` uncaught until it reaches our code on the *other* side of the `read` call, effectively unwinding the call stack and giving us control again. We then asynchronously suspend (i.e. `in.pipe.uncons`) until we have more bytes to supply. Since nothing has been non-idempotently mutated, we can safely just restart the same `read` call as soon as we get new bytes. And since traceless exceptions are insanely fast on HotSpot, we don't even lose any meaningful performance by doing so!

In order to support multiple sequential reads without cursor offsets, we have a single-cell `checkpoint`/`restore` mechanism on the underlying byte stream (which we bypass for performance as soon as we know we're beyond the header and into the `Inflate`-managed datastream). This deals with the fact that, technically, a *successful* `read` on the underlying byte stream represents a non-idempotent mutation (due to the fact that the cursor moves forward). But since that mutation is entirely within our control (rather than in the internals of `GZIPInputStream), it doesn't invalidate the technique.

Overall, I'm pretty happy with the implementation. I don't have any benchmarks written, but it should be relatively competitive with any approach which involves `Stream`, and it is entirely based on a well-tested and standard implementation of the core algorithm. I also have tests which show compatibility with the GNU `gzip`/`gunzip` utilities, just to be extra paranoid.

Also, the technique itself seems to be relatively general. We should be able to use this to make other `InputStream`s async-friendly, provided their internal state management adheres to the pattern I described above.

djspiewak:
Also requesting a review from @wemrysi on the new tests.

This is basically fs2-gzip, but copied over into fs2. I originally intended to put this in fs2-io, per the suggestions, but the problem is that `fs2.compress` is *already* in fs2-core, and it really makes sense for `gzip`/`gunzip` to be in that package. It is now a package object, which I think might screw up binary compatibility. Didn't actually run mima. I can revert that if necessary, though it would be a shame since we would have to put the supporting infrastructure somewhere awkward.

Either way. One thing I *didn't* copy over from fs2-gzip was the "adaptive" variants, which infer the buffer size based on the size of the first chunk of the input stream. I wanted to do this, but unfortunately it seems there are some previously-undiscovered problems with those functions that the property tests revealed (they can actually cause hangs). That concerns me just a bit, but everything is clean on the explicit variants, and mutally-prime stuff works, so… not sure?

Either way, here you go!

Fixes #1313


djspiewak:
Oh yeah, mima is very unhappy. I'll fix.

mpilquist:
Yep - not particularly elegant but safely ignorable given the other trickery in this PR ;)

wemrysi:
@djspiewak Tests look good, though I think we might want smaller buffer sizes to ensure multiple steps as that was what bit us previously. I'm not sure what size streams the `PureStream[A]` generator will result in.

wemrysi:
@djspiewak or possibly better yet, pick smallish buffer sizes and rechunk the generated streams to ensure multiple steps.

mpilquist:
@djspiewak Want any help pushing this one over the finish line?

djspiewak:
@mpilquist I would definitely appreciate it! It's on my TODO list and I will get to it, but I would absolutely not say "no" to someone else jumping in.

PR #1336, opened 2018-11-26
Fix #1335 - stack safety of map

### LINKED ISSUE #1335: Stack-overflow occurring on 1.0.x (Works in 0.10.x)
Here is a minimal rose-tree implementation using fs2:
```
import fs2.{Pure, Stream}

case class Tree[A](label: A, subForest: Stream[Pure,Tree[A]]) {
  /** Depth-first pre-order traversal */
  def flatten: Stream[Pure, A] =
    Stream(this.label) ++ this.subForest.flatMap(_.flatten)
}

object Tree {
  def unfoldForest[A, B](s: Stream[Pure,A])(f: A => (B, Stream[Pure,A])): Stream[Pure,Tree[B]] =
    s.map(unfoldTree(_)(f))

  def unfoldTree[A, B](seed: A)(f: A => (B, Stream[Pure, A])): Tree[B] =
    f(seed) match {
      case (a, bs) => Tree(a, unfoldForest(bs)(f))
    }
}
```

This behaves as expected in `1.0.x`:
```
//         1
//       /   \
//      2     2
//     /|\   /|\
//    3 3 3 3 3 3
//    : : : : : :

val tree1 = Tree.unfoldTree(1)(x => (x, Stream(x+1).repeat.take(x+1)))
```

But this stack-overflows on `1.0.x` but works in `0.10.x`:

```
//    1
//    |
//    2
//    |
//    3
//    :

val tree2 = Tree.unfoldTree(1)(x => (x, Stream(x+1)))
```

A work around in `1.0.x` is something like:
`val tree2 = Tree.unfoldTree(1)(x => (x, Stream(x+1) ++ Stream.empty))`

Adding `++ Stream.empty` seems a bit weird?

https://scalafiddle.io/sf/oi8aWb1/0

mpilquist:


mpilquist:
Before merging, see this comment: 
https://github.com/functional-streams-for-scala/fs2/issues/1335#issuecomment-441819203

mpilquist:
Closing this one b/c even if we think this behavior should change, it's unlikely we'd want to change only `map` and not, for example, `translate` as well.

mpilquist:
Update -- `FreeC.translate` was already suspending if the node was a `FreeC.Bind`. I updated this PR to suspend when the node is a `FreeC.Eval` as well, and this restores the lazy behavior of things like `Stream.{map, translate}`.

SystemFw:
Has this update swayed your opinion wrt tradeoff?

mpilquist:
Yeah, with this change we only add a FreeC.suspend call on translated Eval nodes, where we are already invoking the FunctionK. We aren’t “double suspending” like in the original PR m.

SystemFw:
Right, so if the perf hit is not too much I'd rather laziness to just work (so, merge this), even though `suspend` in that case is not too bad

mpilquist:
Okay merging based on your feedback and Pavel's. I ran the stream JMH benchmarks that use map and saw they haven't meaningfully changed with this PR.

PR #1346, opened 2018-11-29
Removed dependency on parallel collections

mpilquist:


SethTisue:
you also need to remove this from your build.sbt:

```
    libraryDependencies ++= {
      CrossVersion.partialVersion(scalaVersion.value) match {
        case Some((2, minor)) if minor >= 13 =>
          Seq("org.scala-lang.modules" %% "scala-parallel-collections" % "0.1.2" % "test")
        case _ =>
          Seq()
      }
    }
```

mpilquist:
Fixed in 2a8200840fb2ae01cdd20f97414eaae2cbefb737

PR #1349, opened 2018-12-01
Changes and Syntactic Sugaring in the Internals

diesalbla:
We make some changes in the internals of the `Stream` implementation, so as to try to decouple some of the several pieces of it, and help make the code more readable. 

- We add some comments to describe the `Algebra` GADT. 
- We make the `O` parameter of the `Algebra` GADT covariant. This allows us to define the sub-trait `AlgEffect` as Algebra operations with `O=INothing`, in other words, operations that are not concerned with output. As a consequence, for those operations we can remove the middle type parameter.

We also introduce several type aliases, with the intent of simplifying some of the nested type signatures inside it, and to try to make more obvious the patterns between algebra operations. In particular: 
- We use  `ExprFA` as alias for `FreeC[Algebra[F, O, ?], R]`. 
- We use `ProgFA` as alias for `FreeC[Algebra[F, O, ?], Unit]`, which is used to compute a `Stream[F, A]`. 
- We use `StepFA` to represent the output of a compilation step, which is a chunk of outputs, a middle result, and the continuation.
- We add a function `liftFA`, to _lift_ an instance of the `Algebra` GADT into an `FreeC`. 

Hopefully, these type aliases should make it more clear the relation between these classes. It could also guide how to decouple, or extend independently, each part. 

We also introduce some type aliases related to resource cleanup:
- We use `Finaliser[F]` as an alias for  `ExitCase[Throwable] => F[Unit]`, which is to say, an side-effectful operation to be run when a process is finalised, . 
- We use `Cleanup[F, R]` as an alias for `(R, ExitCase[Throwable]) => F[Unit]`, which is to say, a side-effectful function that can be run on any "resource" of type `R` when a process is finalised.


mpilquist:
I'm not generally fond of adding type aliases, especially in the internals. They add a bunch of new terms that need to be understood individually.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -8,55 +8,65 @@ import fs2.internal.FreeC.{Result, ViewL}
...
+                                    interruptedScope: Option[(Token, Option[Throwable])],
+                                    exitCase: ExitCase[Throwable])
+      extends AlgEffect[F, Unit]
 
-  final case class GetScope[F[_], O, X]() extends AlgEffect[F, O, CompileScope[F, X]]
+  final case class GetScope[F[_], X]() extends AlgEffect[F, CompileScope[F, X]]
 
-  sealed trait AlgEffect[F[_], O, R] extends Algebra[F, O, R]
+  @inline
+  private[this] final def liftFA[F[_], O, R](alg: Algebra[F, O, R]): ExprFA[F, O, R] =
```
mpilquist: s/liftFA/liftAlg/ ? No strong opinion but seems clearer to me.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -8,55 +8,65 @@ import fs2.internal.FreeC.{Result, ViewL}
...
-  final case class Step[F[_], X, O](
-      stream: FreeC[Algebra[F, X, ?], Unit],
-      scope: Option[Token]
-  ) extends Algebra[F, O, Option[(Chunk[X], Token, FreeC[Algebra[F, X, ?], Unit])]]
+  final case class Step[F[_], X, O](stream: ProgFA[F, X], scope: Option[Token])
+      extends Algebra[F, O, StepFA[F, X, Token]]
+
+  /* The `AlgEffect` trait is for operations on the `F` effect that create no `O` output.
+   * They are related to resources and scopes. */
+  sealed trait AlgEffect[F[_], R] extends Algebra[F, INothing, R]
```
mpilquist: :+1: to the `INothing` change here.

diesalbla:
> I'm not generally fond of adding type aliases, especially in the internals. They add a bunch of new terms that need to be understood individually.

Of course, type aliases do create an indirection to an extra definition that a code reader needs to keep in mind. Much like `implicit` values or parameters, they add some cognitive overhead that the programmer needs to keep. And, one when writing code, one should check if the improvement in clarity is worth that overhead, much like implicits. In my humble opinion, there are several reasons why I would defend that this is the case: 

First, each type alias is parameterised over a plain list with two or three parameters. The type expressions on the right-hand sides include rather complex nested type expressions. One clear case being the alias `Step[F, O, M]` for `Option[ (Chunk[O], M, FreeC[ Algebra[F, O, ?], Unit])]`, a type expression with three parameters, six literals, one type lambda, all nested into four levels. 
 
Next, type aliases can have a better meaning, an intuition that someone can make to describe how the library works. In this sense, I personally can not give any clear meaning in `FreeC[Algebra[F, O, ?], Unit]` at first, but I can connect a type `ProgFA[F, O]` with the intution of: 
> "a program, an evolving sequence of instructions, each one an element from the `Algebra[F, O, ?]` language of control instructions, that produces `O` as a side-output.

Finally, by introducing these type aliases, you minimise the interfaces between the parts of your program (the type signatures) to their essence, and clear out accidental implementation details. For instance, the essential parameters of a `ExprFA[F, O, A]` are in what effect `F` it's run, what side-output `O` it gives, and what information `A` does it pass to the rest of the _program_. The accidental details, such as the algebra being a GADT,or the program being as a `FreeC`, are hidden behind the type alias.

mpilquist:
I'm not convinced the aliases are worth it in this case, though I do appreciate the detailed perspective. Let's see what @pchlupacek and @SystemFw think though.

I could see an alias for `StepFA` -- I'm less convinced of the other too though, b/c they are sort of redundant with the `Stream` and `Pull` new types:

```scala
type ExprFA[F[_], O, R] = FreeC[Algebra[F, O, ?], R] // This is Pull as a type alias instead of anyval
type ProgFA[F[_], O] = ExprFA[F, O, Unit]. // This is Stream as a type alias instead of anneal
```

We purposefully don't use `Stream` and `Pull` in the main GADT b/c we want to avoid circular dependencies (though we don't currently succeed at that 100%).

diesalbla:
> They seem like a good idea but in my experience they only generate confusion

Well, It helps to think of type aliases as functions, except that they work on types instead of the value-level. Much as `cats.Functor` adds special alias methods, such as `def as(a: A)` for `map( _ => a )`,  or `void` for `map(_ => () )`, or the many special methods in `Foldable` that can be seen as a simple `foldMap`.

In any case, I think I will submit the `AlgEffect - INothing` stuff separately.

PR #1354, opened 2018-12-02
CompileScope: several changes. 

diesalbla:
Several changes in the `CompileScope` scala file: 
- We replace some uses of `flatMap` or `map` with some operators from the `Functor and `Applicative` type classes, such as `Functor.as` to return a constant, or `Applicative.{productL, productR, map2}` to combine independent operators, or apply a function to their result. 
- We extract the special `Lease` of the `CompileScope`, which is an aggregation of leases, into a separate class. 
- We replace the `unsafeFromInterruptible` with `unsafeFromConcurrent`, which we can use in other places.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/CompileScope.scala
+++ b/core/shared/src/main/scala/fs2/internal/CompileScope.scala
@@ -2,7 +2,7 @@ package fs2.internal
 
 import scala.annotation.tailrec
 
-import cats.{Traverse, TraverseFilter}
+import cats.{Applicative, Traverse, TraverseFilter}
 import cats.data.Chain
 import cats.effect.{Concurrent, ExitCase, Sync}
 import cats.effect.concurrent.{Deferred, Ref}
@@ -341,14 +341,8 @@ private[fs2] final class CompileScope[F[_]] private (
       else {
         val allScopes = (s.children :+ self) ++ ancestors
         F.flatMap(Traverse[Chain].flatTraverse(allScopes)(_.resources)) { allResources =>
-          F.map(TraverseFilter[Chain].traverseFilter(allResources) { r =>
-            r.lease
-          }) { allLeases =>
-            val lease = new Scope.Lease[F] {
-              def cancel: F[Either[Throwable, Unit]] =
-                traverseError[Scope.Lease[F]](allLeases, _.cancel)
-            }
-            Some(lease)
+          F.map(TraverseFilter[Chain].traverseFilter(allResources)(_.lease)) { allLeases =>
+            Some(new CompileScope.Lease[F](allLeases))
           }
         }
       }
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -516,27 +510,39 @@ private[internal] object CompileScope {
...
+        deferred = Deferred.unsafe[F, Either[Throwable, Token]](concurrent),
+        ref = Ref.unsafe[F, Option[Either[Throwable, Token]]](None),
+        interruptRoot = scopeId,
+        cancelParent = cancel
+      )
+
+  }
+
+  /* A Lease for the compile scope, which is just an aggregation of leases */
+  private class Lease[F[_]](leases: Chain[Scope.Lease[F]])(implicit F: Applicative[F])
```
diesalbla: @mpilquist Per our recent conversation: the idea of a _Lease_ made of many other leases, is it general enough to merit  adding it to the `fs2.Scope` companion object?

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -222,20 +224,24 @@ private[fs2] final class CompileScope[F[_]] private (
     * finalized after this scope is closed, but they will get finalized shortly after. See [[Resource]] for
     * more details.
     */
-  def close(ec: ExitCase[Throwable]): F[Either[Throwable, Unit]] =
+  def close(ec: ExitCase[Throwable]): F[Either[Throwable, Unit]] = {
+    val selfClean: F[Unit] = F.productR(
```
diesalbla: This was nested inside the flatMap continuation, but it happens to be independent of the variables bound in that flatMap.

core/shared/src/main/scala/fs2/internal/CompileScope.scala
```diff
@@ -121,40 +121,41 @@ private[fs2] final class CompileScope[F[_]] private (
...
-              case Some(parent) =>
-                F.flatMap(self.interruptible.map(_.cancelParent).getOrElse(F.unit)) { _ =>
-                  parent.open(interruptible)
-                }
 
-              case None =>
-                F.pure(Left(new IllegalStateException("cannot re-open root scope")))
-            }
+    val newScopeId = new Token
+    F.flatMap(createScopeContext(newScopeId)) { iCtx =>
```
diesalbla: Something I have noticed  is that this gets created, but immediately becomes garbage, if the state is not open. Would it make more sense to put it inside that branch?

core/shared/src/main/scala/fs2/internal/CompileScope.scala
mpilquist: Not public at least, since I don't think we'd want to expose the underlying chain to users.

pchlupacek:
Hi just a small note on these changes. I am quite scared on significant changes to core interpreter w/o any measurable benefits to interpreter. If the change is obvious then fine, but if these are cosmetics, I would say to keep them until next iteration of fs2, if possible. 

Although we have numerous tests to assure the core interpreter is sound, and correct, we don't have that perfect coverage on all scenarios of resources, scopes etc, to make 100% sure that none changes had no side-effecting impact. 

@diesalbla I really appreciate the excellent work you doing, but would be perfect to tie some of the changes with some improvements that are somehow user-facing and could positively impact the performance of the fs2 overall. 

Just a thought a expression on may feelings, I don't want to be negative here :-)

SystemFw:
I kinda feel the same as Pavel here, I hope I didn't come across negatively on the PRs, but that's where I'm coming from as well :)

PR #1362, opened 2018-12-06
Replace the use of ViewL.View with an "unfoldView"

### LINKED ISSUE #1360: Question / Spike - Internals: turn View into Tagged Type
In the `fs2.internals` package, there is a `ViewL[F[_], R]` trait, with a single case class `ViewL.View` and a single factory method `ViewL.apply`. The goal of this class is to represent expressions of the `FreeC[F, R]` type that have been _unrolled_ deep enough until it could no longer be unrolled. In particular, a View can be either:

- An instance of the sub-type `FreeC.Result`, with no continuation; or
- A `ViewL.View`, which corresponds to a `FreeC.Bind` where the left hand side is a `FreeC.Eval` (the lift operation), and the rest is a continuation on the result.

We could get the same functionality using a tagged type, which would wrap a `FreeC[F, R]`. Using a tagged type could remove the need for an extra `ViewL` ADT (which overlaps with `FreeC`), at no memory cost.

diesalbla:
In the `fs2.internals`, besides  `FreeC` there is an overlapping ADT, `ViewL`, used to represent `FreeC` expressions at which evaluation in `FreeC` can stop. This evaluation proceeds by
unrolling left-nested `Bind` and unrolled, composing Bind-continuations to the right, until it gets either a final result or  a `ViewL`.

It turns out, the `ViewL` is not stored anywhere, nor used as type parameters anywhere.
Instead, a `ViewL` is emitted by using a method `viewL` (on the `FreeC` class) and, in all usages, is immediately pattern-matched. This suggests that we can remove the `ViewL` type altogether. 

This PR replaces any use of `viewL` by a higuer-order function `viewFold`, which unrolls the `FreeC` and takes two continuations: one is applied if the unrolling stops at a `Result`, and another one if the unrolling ends up at a `Eval-Bind` (represented with the `View` class. 

Resolves #1360, although using a different solution to the one proposed.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -213,19 +213,20 @@ private[fs2] object Algebra {
         scope: CompileScope[F],
         stream: FreeC[Algebra[F, X, ?], Unit]
     ): F[R[X]] = {
-      stream.viewL match {
-        case _: FreeC.Result.Pure[Algebra[F, X, ?], Unit] =>
+
+      val goResult: FreeC.Result[Unit] => F[R[X]] = {
+        case Result.Pure(_) =>
           F.pure(Done(scope))
-
-        case failed: FreeC.Result.Fail[Algebra[F, X, ?], Unit] =>
-          F.raiseError(failed.error)
-
-        case interrupted: FreeC.Result.Interrupted[Algebra[F, X, ?], _, Unit] =>
-          interrupted.context match {
-            case scopeId: Token => F.pure(Interrupted(scopeId, interrupted.deferredError))
+        case Result.Fail(error) =>
+          F.raiseError(error)
+        case Result.Interrupted(context, deferredError) =>
+          context match {
+            case scopeId: Token => F.pure(Interrupted(scopeId, deferredError))
             case other          => sys.error(s"Unexpected interruption context: $other (compileLoop)")
           }
+      }
 
+      val goView: ViewL.View[Algebra[F, X, ?], _, Unit] => F[R[X]] = {
         case view: ViewL.View[Algebra[F, X, ?], y, Unit] =>
           def interruptGuard(scope: CompileScope[F])(next: => F[R[X]]): F[R[X]] =
             F.flatMap(scope.isInterrupted) {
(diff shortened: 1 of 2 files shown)


diesalbla:
> @diesalbla could you prove benefits of this change? Any tests // measurement proving that this has measurable end user benefits? [From [this comment](https://github.com/functional-streams-for-scala/fs2/issues/1360#issuecomment-445107466) ]

@pchlupacek I am afraid not. The changes are not motivated by any use case or memory leak. Still, I will look to run the benchmarks to see if there are any performance regressions. 

In any case, the changes are intended to improve a bit the design of the code. On that account, and  since you introduced the `ViewL` trait not long ago, could you point me to the discussion on those changes? I have no wish to push the changes in this PR if they are a step back from the direction you were following.

pchlupacek:
@diesalbla I am sorry, but I am not in favor of the PRs that only improve code style, hence in many ways they can introduce bugs and often are just reflection of personal preferences. 

Would be excellent if you can help with tagless implementation of the algebra as @mpilquist suggested, that will really have impact, hence it may reduce footprint of internal implementation resulting likely in better perfromance of the fs2 overall.

diesalbla:
> @diesalbla I am sorry, but I am not in favor of the PRs that only improve code style, hence in many ways they can introduce bugs and often are just reflection of personal preferences.

Given how many people are, informally, relying on `fs2`, your caution is more that justified. And, to be perfectly clear, if you already have a preference you can state them and I will respect them.

> Would be excellent if you can help with tagless implementation of the algebra as @mpilquist suggested, that will really have impact, hence it may reduce footprint of internal implementation resulting likely in better [performance] of the fs2 overall.

Well, I did not realised this at first, but this PR happens to be a small step in that direction. 

To go "tagless" (or rather _shallow_) means that, instead of creating a data structure (or AST) for some inner _language_ of instructions, and then unfolding that AST with an interpreter, you directly call the interpreter. Under this perspective, the `ViewL` type is like a "AST" to communicate two pieces of the internals: the one to unroll the `FreeC` until it reaches a (weak) normal form, and the one to interpret the root of that normal form. Under this perspective, this PR replaces creating or emitting a  `ViewL`, with a direct call to an interpreter for it. 

However, I am not so sure if it would reduce memory footprint, since creating interpreter objects may be just as expensive as creating the AST nodes. This PR may not be ready to merge just yet.

PR #1364, opened 2018-12-11
Express Pipe.join with broadcast

SystemFw:
No one must have used `Pipe.join` ever, because it looks very broken to me. Anyway, this PR reimplements it with `broadcast`.  Pavel already had a look on gitter

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -208,7 +208,9 @@ lazy val mimaSettings = Seq(
     ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.io.tcp.Socket.client"),
     ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.io.tcp.Socket.server"),
     ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.io.tcp.Socket.mkSocket"),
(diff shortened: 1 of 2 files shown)


mpilquist:
:+1:

This breaks binary compatibility due to `joinQueued` and `joinAsync` being removed but if they are that broken, maybe that's okay in this case?

SystemFw:
```scala
  val p1: Pipe[IO,Int,(Int, String)] = _.map(_+1).take(2).tupleRight("p1")
  val p2: Pipe[IO,Int,(Int, String)] = _.map(_*2).take(2).tupleRight("p2")
  val p3 = (s: Stream[IO,Int]) => s.through(Pipe.join(Stream(p1,p2)))

  def p =
    Stream.range(0, 20).through(p3).evalMap(put).interruptAfter(3.seconds)
```

this deadlocks and can't even be interrupted with `interruptAfter` (I don't even know how). I'd classify that as pretty broken :P

I don't know what to do in these cases, my feeling is that no one is using this, especially not in a library, but who knows.

mpilquist:
OK let's add Mima exclusions for these then.

SystemFw:
@mpilquist done

pchlupacek:
👍

PR #1365, opened 2018-12-12
Add recommendations for blockingExecutionContext

jferris:
* Document that the io implementations do, in fact, block
* Recommend an unbounded thread pool
* Reference Cats Effect docs on blocking threads


mpilquist:
Awesome, thanks!

PR #1368, opened 2018-12-15
Removed dead code path in Stream.bracket

mpilquist:
I noticed this when working on tagless branch -- the error handling case in `bracketFinalizer` could never be hit b/c it was only called immediately after a simple emit, which cannot fail with an exception.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ z_local.sbt
 .tags
 .metaserver
 test-output
+.metals
+.bloop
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -16,3 +16,5 @@ z_local.sbt
 .tags
 .metaserver
 test-output
+.metals
```
SystemFw: 👍 😉

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2743,33 +2743,9 @@ object Stream extends StreamLowPriority {
           .emit(r)
           .covary[F]
           .map(o => (token, o))
-          .get[F, (Token, R)]
-          .transformWith(bracketFinalizer(token))
+          .get[F, (Token, R)] >> Algebra.release(token)
```
SystemFw: What about the `Result.Interrupted` case? Why is it not relevant anymore? Or is it dead-code as well given that it's only `emit` and then `map`?

mpilquist: That case was an identity -- it indeed gets hit but doesn't alter the result in any way.

SystemFw: Ah, ofc, it's just propagating the same `Interrupted` node

PR #1369, opened 2018-12-15
Add repeatN combinator to Stream and a spec for repeat and repeatN

ybasket:
Lately, I found myself several times in the need of repeating a `Stream` a specific number of times and I don't see a good way of doing this with the existing combinators - one can do `repeat.take(n * s)` for a `Stream` of known size `s`, but that usually doesn't help as the length is unknown. Also added a scalacheck spec for testing `repeat` and `repeatN` on simple, pure streams.

Maybe not worth being added, but as there's a chance others need this as well, I wanted to give it a try ;)
Questions and/or comments welcome!

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2058,6 +2058,22 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     this ++ repeat
 
   /**
+    * Repeat this stream a given number of times.
+    *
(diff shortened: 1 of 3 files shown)


mpilquist:
Thanks!

SystemFw:
> repeating a Stream a specific number of times and I don't see a good way of doing this with the existing combinators - one can do repeat.take(n * s) for a Stream of known size s, but that usually doesn't help as the length is unknown

Just leaving this as a tip:
```scala
val in = Stream(1,2,3)
val repeatThreeTimes = Stream(in).repeat.take(3).flatten
```

PR #1370, opened 2018-12-17

diesalbla:
The methods `translate0` and `translateStep` were introduced in previous versions, to deal with different variants of the translations. `translate0` was introduced to remove translation from the algebra, and  `translateStep` was introduced to support the new compiler-interpreter.

Looking at them, the only differences between these methods were:
- That translate0 was not parameterised on the output type X.
- That, for the algebra operation of output, the `translate0` had a recursive call to `translate0`, whereas the `translateStep` method was performing a Cast to a result of the parametric type X.

Thus, we have unified these methods, by replacing the initial call to `translate0` as a call to `translateStep` with `X = O`.

We also wrap the `translateStep` inside a `translateAux` function, which is the main entry method of the recursive loop. Since neither the `fK` nor the `Concurrent[G]` change in the recursive calls of `translateStep`, we use a pattern useful in recursive functions, of not passing in the recursive calls the
parameters, either type or value, that never change.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -150,7 +150,7 @@ private[fs2] object Algebra {
       s: FreeC[Algebra[F, O, ?], Unit],
       u: F ~> G
   )(implicit G: TranslateInterrupt[G]): FreeC[Algebra[G, O, ?], Unit] =
-    translate0[F, G, O](u, s, G.concurrentInstance)
+    translateAux[F, G, O](u, s, G.concurrentInstance)
 
   def uncons[F[_], X, O](s: FreeC[Algebra[F, O, ?], Unit])
     : FreeC[Algebra[F, X, ?], Option[(Chunk[O], FreeC[Algebra[F, O, ?], Unit])]] =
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -410,115 +410,72 @@ private[fs2] object Algebra {
...
-              .transformWith { r =>
-                translate0(fK, view.next(r.asInstanceOf[Result[y]]), concurrent)
+                )
               }
 
-          case alg: Algebra.AlgEffect[F, r] =>
-            FreeC
-              .Eval[Algebra[G, O, ?], r](translateAlgEffect(alg, concurrent, fK))
-              .transformWith(r => translate0(fK, view.next(r), concurrent))
+            case alg: Algebra.AlgEffect[F, res] => {
```
diesalbla: After the changes in this PR, this is the only place in which the middle trait `AlgEffect` is used. If we replace this single match with the six matches below, we can remove that middle trait.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -410,115 +410,72 @@ private[fs2] object Algebra {
...
+    def translateStep[X](next: FreeC[Algebra[F, X, ?], Unit]): FreeC[Algebra[G, X, ?], Unit] =
+      next.viewL match {
+        case _: FreeC.Result.Pure[Algebra[F, X, ?], Unit] =>
+          FreeC.pure[Algebra[G, X, ?], Unit](())
 
-      case view: ViewL.View[Algebra[F, X, ?], y, Unit] =>
-        view.step match {
-          case output: Algebra.Output[F, X] =>
-            Algebra.output[G, X](output.values).transformWith {
-              case r @ Result.Pure(v) =>
```
diesalbla: This is the one point in which there is a difference between `translate0` and `translateStep`.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -410,115 +410,72 @@ private[fs2] object Algebra {
...
 
-      case interrupted: FreeC.Result.Interrupted[Algebra[F, O, ?], _, Unit] =>
-        FreeC.interrupted(interrupted.context, interrupted.deferredError)
+                case r @ Result.Fail(err) => translateStep(view.next(r))
 
-      case view: ViewL.View[Algebra[F, O, ?], y, Unit] =>
-        view.step match {
-          case output: Algebra.Output[F, O] =>
-            Algebra.output[G, O](output.values).transformWith { r =>
-              translate0(fK, view.next(r), concurrent)
```
diesalbla: This is the only point where `translate0` differs from `translateStep`. Not sure if this recursive call modifies the behaviour, with respect to the type case.

diesalbla:
Tests were failing in Travis. Will reopen when ready.

PR #1372, opened 2018-12-17

diesalbla:
The methods `translate0` and `translateStep` were introduced in previous versions, to deal with different variants of the translations: `translate0` was introduced to remove translation from the Algebra, and `translateStep` was introduced to support the new compiler-interpreter.

The differences between these methods were:
- That `translate0` was not parameterised on the output type `X`, and uses directly `O` instead.
- That, for the algebra operation of output, the `translate0` had a  recursive call to `translate0`, whereas the translateStep method was performing a Cast to a result of the parametric type X.
- That `translate0` first recurs with itself, until it resorts to `translateStep`, as in some layered approach.

Thus, we have unified these methods, by replacing the initial call to `translate0` as a call to `translateStep` with `X = O`. We add a boolean flag, `isMainLevel`, to indicate if it is the main level
(formerly `translate0`) or the lower levels (`translateStep`). We also wrap the `translateStep` inside a `translateAux` function, which is the main entry method of the recursive loop. Since neither the `fK` nor the `Concurrent[G]` change in the recursive calls of `translateStep`, we use a pattern useful in recursive functions, of not passing in the recursive calls the (type or value) parameters that never change.


core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -410,115 +410,77 @@ private[fs2] object Algebra {
...
-              .transformWith { r =>
-                translate0(fK, view.next(r.asInstanceOf[Result[y]]), concurrent)
+                )
               }
 
-          case alg: Algebra.AlgEffect[F, r] =>
-            FreeC
-              .Eval[Algebra[G, O, ?], r](translateAlgEffect(alg, concurrent, fK))
-              .transformWith(r => translate0(fK, view.next(r), concurrent))
+            case alg: Algebra.AlgEffect[F, res] => {
```
diesalbla: With the changes in this PR, this is the only use for the `AlgEffect` middle trait. If we unfold the 6 inner cases to the `view.step match{` level, we can remove `AlgEffect`.

core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -410,115 +410,77 @@ private[fs2] object Algebra {
...
 
-          case step: Algebra.Step[F, x, O] =>
-            FreeC
-              .Eval[Algebra[G, O, ?], Option[(Chunk[x], Token, FreeC[Algebra[G, x, ?], Unit])]](
-                Algebra.Step[G, x, O](
-                  stream = translateStep[F, G, x](fK, step.stream, concurrent),
+            case step: Algebra.Step[F, x, X] =>
+              evalCont[Option[(Chunk[x], Token, FreeC[Algebra[G, x, ?], Unit])]] {
+                Algebra.Step[G, x, X](
+                  stream = translateStep[x](step.stream, false),
```
diesalbla: This is the only recursive call to `translateStep` where it needs to change te `isMainLevel` to `false`.

PR #1374, opened 2018-12-20
Adding new logo to microsite

gvolpe:
Also re-organizing tut documentation files into the standard directory.

fs2

The logo doesn't seem to render very well but that's how it is :/


mpilquist:
Looks great. We don't need a build.sbt change corresponding to the tut source directory change?

gvolpe:
Apparently is not needed since the custom dir `/src` is still the top level of the standard `/src/main/tut/` so the sources get picked up anyway but to do things right I just removed that line from `build.sbt`. I didn't realized it was there to begin with :smile:

PR #1375, opened 2018-12-21
Readability - Rename variables of "scope" in the compileLoop function. 

diesalbla:
Rename variables `scope` in the `compileLoop` funciton. 

The code in the `compileLoop` function had several function parameters, or pattern bindings, all named `scope` and typed `CompileScope[F]`. This was a bit confusing, and at some points there could be up to three different variables in view. We renamed them, to avoid this confusion.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -209,7 +209,7 @@ private[fs2] object Algebra {
    */
 
   private[this] def compileLoop[F[_], O](
-      scope: CompileScope[F],
+      compileScope: CompileScope[F],
       stream: FreeC[Algebra[F, O, ?], Unit]
   )(implicit F: Sync[F]): F[Option[(Chunk[O], CompileScope[F], FreeC[Algebra[F, O, ?], Unit])]] = {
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -238,81 +238,83 @@ private[fs2] object Algebra {
...
             case u: Step[F, y, X] =>
               // if scope was specified in step, try to find it, otherwise use the current scope.
-              F.flatMap(u.scope.fold[F[Option[CompileScope[F]]]](F.pure(Some(scope))) { scopeId =>
-                scope.findStepScope(scopeId)
+              F.flatMap(u.scope.fold[F[Option[CompileScope[F]]]](F.pure(Some(goScope))) { scopeId =>
+                goScope.findStepScope(scopeId)
               }) {
                 case Some(stepScope) =>
                   F.flatMap(F.attempt(go[y](stepScope, u.stream))) {
-                    case Right(Done(scope)) =>
```
diesalbla: At these point, there were three shadowed variables called `scope`: the one in this line, the one in the recursive `go` function, and the one in the entry `compileLoop`.

mpilquist:
This is intentional -- we often purposefully shadow names to prevent the ability to pass the wrong value.

More generally, I want to stress that changes to the core interpreter should be motivated by performance or correctness reasons. We're generally pretty careful about merging changes to `Algebra`, `CompileScope`, etc. as they are very complex. It would be more fruitful to help on the tagless2 branch. :)

diesalbla:
That being the case, I can as well withdraw the PR.

PR #1382, opened 2018-12-31
Removed use of cats.data.Eval from stream compilation and added s.compile.string syntax

mpilquist:


SystemFw:
This appears to break bincompat. Any plans on how to proceed?

mpilquist:
Yeah I need to add a mima suppression - will update tomorrow.

PR #1384, opened 2019-01-01
Remove Type Parameters in FreeC and subclasses.

diesalbla:
Since `FreeC[F[_], +R]` is covariant in O, we modify its sub-classes `Fail` and `Interrupted`, so as to set `R` to `INothing`, and then we can remove the type parameter from the class. We apply the same change to some of the factory methods.


pchlupacek:
The implementaion seems correct. However I would like to keep the FreeC as it is, hence this does not bring any user-faced beenfit. 

I think as @mpilquist explained now the focus is on tagless implementation of core, and until that is done, or we will call it no-go, I think we shall cease modification of the current internals, that are not bringing immediate benefits.

mpilquist:
I'm good with this change if @pchlupacek doesn't object to a merge as it removes a cast or two and I feel there's no risk in merging.

I definitely would like us to spend more time on the tagless approach though. I'm not necessarily against improving the internals of 1.0 for clarity reasons but I fear that spending time on internals takes away time we could be spending on things like performance improvements, the cats-effect 2 hierarchy, better cancelation support, or the tagless architecture - especially when the changes are non-trivial to review.

PR #1385, opened 2019-01-01
Remove MonadInstance object

diesalbla:
The `monadInstance` value, in the `Result` object, which was not implicit, was only used at one place, and for the `map` method (through `flatMap`), explicitly. That being the case, we can drop that unused code and add a function `Result.map` function, directly on the companion object.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/FreeC.scala
+++ b/core/shared/src/main/scala/fs2/internal/FreeC.scala
@@ -1,6 +1,6 @@
 package fs2.internal
 
-import cats.{Monad, MonadError, ~>}
+import cats.{MonadError, ~>}
 import cats.effect.{ExitCase, Sync}
 import fs2.CompositeFailure
 import FreeC._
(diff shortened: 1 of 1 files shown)


pchlupacek:
Why we need to do that? I was under impression that it is nice that Result may form monad?

diesalbla:
@pchlupacek It is dead code, most of it, and you can cut the part that isn't as a plain function. 

It is helpful to know that the sealed type `Result[A]` type allows an instance of `Monad`, and `MonadError` too, but that can be just declared in a code comment. Personally, I do not find that instance too complicated: if you think of `Result[A]` like an `Either[Throwable, A]` with one more case, you have an instance like that of [`Either[Err,?]`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/instances/either.scala#L33-L61)

pchlupacek:
I do not think that removing the instance that we already have makes sense. The fact that we have it on companion object makes it easier for users. The fact that we do not use it in library significantly does not justifies that we have to remove it. 

Result is type that is introduced by fs2, so we shall be responsible for basic instances and their implementations instead of forcing users to implement them, however that easy may be.

diesalbla:
> [...]. The fact that we have it on companion object makes it easier for users. [...], so we shall be responsible for basic instances and their implementations instead of forcing users to implement them, however that easy may be.

I am afraid that usability would not matter for much, given that the `Result` sealed trait, with its subclasses, is defined within the `FreeC` companion object, which is `private` to `fs2`.

pchlupacek:
Yeah, thats true, that Result is now private. Still I am not convinced to pursue this direction, hence I do not see any difference from the current implementation, which is richer and could allow us to do more until hopefully we get rid of FreeC alltogether.

mpilquist:
Like #1384, I'm okay with merging this one since it's removing dead code only and hence very, very low risk.

PR #1386, opened 2019-01-02
Deprecated Sink in favor of Pipe

mpilquist:
`Sink[F, I]` was an alias for `Pipe[F, I, Unit]` - I don't think saving a few characters is worth the cost of a top level concept in the library. Having redundant types like this led to methods like `through` and `to`, where `to` is a slightly more specialized version of `through` (one less type parameter) and otherwise identical.

I'd like to reserve the word `Sink` for use in the next major version of FS2 -- perhaps to replace the "compilation" terminology we use now.

Also note that the `Pipe` and `Sink` aliases (necessarily) set the `F` type param to be invariant, which makes them more constrained than needed. Both `through` and the now deprecated `to` needed to be defined to *not* use the alias so that the input effect and the output effect could vary independently. E.g.:

```scala
// On Stream[F, O]:
def through[F2[x] >: F[x], O2](f: Stream[F, O] => Stream[F2, O2]): Stream[F2, O2]
```


SystemFw:
Controversial, but I'm :+1: personally.
The nature of `Sink` is pretty confusing as well given that you can `.to(thing).flatMap...`

djspiewak:
I'm 👍. I do use `Sink` quite a bit, but `Pipe` is kind of trivial to use instead.

pchlupacek:
+1, lgtm

PR #1388, opened 2019-01-04
Compile a Stream down to a Resource

SystemFw:
Motivation: be able to return a `Resource` instead of a singleton Stream, for an abstraction whose constructor uses Stream concurrency combinators.

What we can do today:
- Return a singleton stream. Example:
    ```scala
    def hold[F2[x] >: F[x], O2 >: O](initial: O2)(
      implicit F: Concurrent[F2]): Stream[F2, Signal[F2, O2]] =
    Stream.eval(SignallingRef[F2, O2](initial)).flatMap { sig =>
      Stream(sig).concurrently(evalMap(sig.set))
    }
  ```

- Fall back to use `start` + `Fiber` directly. Way more error prone and lower level.

Status: I thought using `Resource` to manage the top level scope is enough, but it's not.

Look at the `basic` test for the gist of the behaviour I want to enable.

### DIFF
--- a/core/jvm/src/test/scala/fs2/ResourceCompilationSpec.scala
+++ b/core/jvm/src/test/scala/fs2/ResourceCompilationSpec.scala
@@ -1,6 +1,78 @@
 package fs2
 
+import cats.implicits._
 import cats.effect.{IO, Resource}
+import cats.effect.concurrent.{Deferred, Ref}
+import scala.concurrent.duration._
+
+class ResourceCompilationSpec extends AsyncFs2Spec {
+
+  "compile.toResource - concurrently" in {
+    pending
+    val prog: Resource[IO, IO[Unit]] =
+      Stream
+        .eval(Deferred[IO, Unit].product(Deferred[IO, Unit]))
+        .flatMap {
+          case (startCondition, waitForStream) =>
+            val worker = Stream.eval(startCondition.get) ++ Stream.eval(waitForStream.complete(()))
+            val result = startCondition.complete(()) >> waitForStream.get
+
+            Stream.emit(result).concurrently(worker)
+        }
+        .compile
+        .toResource
+        .lastOrError
+
+    prog.use(x => x).timeout(5.seconds).unsafeToFuture
+  }
+
+  "compile.toResource - onFinalise" in {
+    pending
+    val expected = List(
+      "stream - start",
+      "stream - done",
+      "io - done",
+      "io - start",
+      "resource - start",
+      "resource - done"
+    )
+
+    Ref[IO]
+      .of(List.empty[String])
+      .flatMap { st =>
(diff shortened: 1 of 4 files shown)


SystemFw:
Adding this here from Gitter:

So, I think I fixed it (all new and existing tests pass), but I'd like to hear your thoughts on the approach.
So, the problem I was having is that Release nodes trigger finalisers via scope.releaseResource before the outer call to scope.close at the very end of the compile loop does (this would be a double release, except releaseResource also removes the resources from the Map).
So, I simply made releaseResource a no-op for root scope (the one with no parent). This guarantees prompt finalisation for the inner scopes (Releaseis still respected), but leaves the finalisation of the root scope down to the call to scope.close in the outer bracket, at which point replacing bracket with Resource.make achieves the desired behaviour.
Are there any flaws in this logic? All tests pass, but I'm not sure if/how heavily we're testing for prompt finalisation.

core/jvm/src/test/scala/fs2/ResourceCompilationSpec.scala
```diff
@@ -0,0 +1,95 @@
+package fs2
+
+import cats.implicits._
+import cats.effect.{IO, Resource}
+import cats.effect.concurrent.{Deferred, Ref}
+import scala.concurrent.duration._
+
+class ResourceCompilationSpec extends AsyncFs2Spec {
+  // NOTE: still doesn't work. There are Release nodes being inserted, whose finaliser gets called anyway.
```
mpilquist: This comment is out of date right? The disabling of release if parent is defined fixes this issue?

SystemFw: Yes, sorry, there is cleanup work I need to do, more comments etc.
I'm mainly interested to know if there is a logical fallacy in my basic idea, which somehow affects resource safety. If that's not the case, I'll do the necessary work to make this ready for merge :)

SystemFw: Especially because we need to fix `interruptibleEval` asap, so it would be nicer to somehow linearise the changes to Scope to avoid making a mess :)

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3819,6 +3864,97 @@ object Stream extends StreamLowPriority {
     def lastOrError(implicit G: MonadError[G, Throwable]): G[O] =
       last.flatMap(_.fold(G.raiseError(new NoSuchElementException): G[O])(G.pure))
 
+    /**
```
mpilquist: Awesome description!

SystemFw:
Ah, you both spotted the typo (realised yesterday and it was late) and fixed it, thanks :)

PR #1390, opened 2019-01-07
fix readme for correct library versions in example

visox:
Fixed as found in maven.
@SystemFw asked me if i could fix it.
edit: sorry wrong copy-paste name


SystemFw:
that was me actually ;)

PR #1403, opened 2019-01-27
scanMap and scanMonoid

### LINKED ISSUE #1184: Add `foldMap` alternative that uses `scan`, `scanMonoid`
Something along the lines of:

```scala
def scanMap[O2](f: (O2, O) => F[O2])(implicit O2: Monoid[O2], F: Functor[F]): fs2.Stream[F, O2] =
  stream.evalScan(O2.empty)((acc, el) => F.map(f(acc, el))(O2.combine(acc, _)))
```

Not sure about the naming of it, could be `evalScanMonoid` or something like that, but it looks similar to `foldMap` to me (apart from the `Functor[F]` requirement). Maybe it's possible to write it with `pull` and without the Functor bound?

Also, `scanMonoid` resembling `foldMonoid`:

```scala
def scanMonoid(implicit O: Monoid[O]): fs2.Stream[F, O]
```

tabdulradi:
Resolves #1184 

As per the conversation on the issue, this PR adds `scanMap` and `scanMonoid` with similar signature as `foldMap` and `foldMonoid`, but emit intermediate results.


mpilquist:
Thanks!

PR #1405, opened 2019-01-28
Fix groupWithin to not emit empty groups, resolves #1404

### LINKED ISSUE #1404: groupedWithin emits empty groups
Hi

I use fs2 version `1.0.2` and the `groupedWithin` function seems not  to behave as documented. 

```
/**
    * Divide this streams into groups of elements received within a time window,
    * or limited by the number of the elements, whichever happens first.
    * Empty groups, which can occur if no elements can be pulled from upstream
    * in a given time window, **will not be emitted.**
    *
    * Note: a time window starts each time downstream pulls.
    */
  def groupWithin[F2[x] >: F[x]](n: Int, d: FiniteDuration)(
``` 

The documentation says that empty groups will not be emitted. But when `numberOfElements % n == 0` then the last group is still empty.

Reproducer:

```scala
import fs2.Stream

object Fs2Test extends App {

  implicit val cs: ContextShift[IO] = IO.contextShift(concurrent.ExecutionContext.global)
  implicit val timer: Timer[IO]     = IO.timer(concurrent.ExecutionContext.global)

  val numberStrings = (1 to 30).map(_.toString).toList
  Stream
    .emits(numberStrings)
    .covary[IO]
    .groupWithin(30, 1.milli)
    .map(chunk => println(chunk.size))
    .compile
    .drain
    .unsafeRunSync()

  // Output is:
  // 30
  // 0
}
```

Is this a bug or is the documentation miss-leading?

Sebruck:


SystemFw:
Thank you!

Issue #1406, opened 2019-01-29
Topic suffers under heavy load, `Set` comparison CPU intensive


SystemFw:
There is a problem with Topic (and I suspect with Signal as well), probably due to how CPU intensive Set comparison is. One of my teams had problems were reasonably heavy traffic was causing thread starvation due to the CPU work needed for that comparison. I was able to fix by porting the old Topic (without PubSub), but we need to address this because it's a pretty serious regression right now. I'm trying to come up with a reproducible example, which isn't super easy.

SystemFw:
In the meantime, I can share the stack trace commonly seen in the problematic scenario
```
"ForkJoinPool-2-worker-6" #83 daemon prio=5 os_prio=0 tid=0x00007f56fb6d3000 nid=0x13bc runnable [0x00007f56db7fd000]
   java.lang.Thread.State: RUNNABLE
        at scala.runtime.ScalaRunTime$._hashCode(ScalaRunTime.scala:149)
        at scala.Tuple2.hashCode(Tuple2.scala:24)
        at scala.runtime.Statics.anyHash(Statics.java:122)
        at scala.collection.immutable.HashMap.elemHashCode(HashMap.scala:87)
        at scala.collection.immutable.HashMap.computeHash(HashMap.scala:96)
        at scala.collection.immutable.HashMap.contains(HashMap.scala:59)
        at scala.collection.MapLike$DefaultKeySet.contains(MapLike.scala:176)
        at scala.collection.GenSetLike.apply(GenSetLike.scala:48)
        at scala.collection.GenSetLike.apply$(GenSetLike.scala:48)
        at scala.collection.AbstractSet.apply(Set.scala:51)
        at scala.collection.immutable.MapLike$ImmutableDefaultKeySet.apply(MapLike.scala:112)
        at scala.collection.Iterator.forall(Iterator.scala:953)
        at scala.collection.Iterator.forall$(Iterator.scala:951)
        at scala.collection.AbstractIterator.forall(Iterator.scala:1429)
        at scala.collection.IterableLike.forall(IterableLike.scala:77)
        at scala.collection.IterableLike.forall$(IterableLike.scala:76)
        at scala.collection.AbstractIterable.forall(Iterable.scala:56)
        at scala.collection.GenSetLike.subsetOf(GenSetLike.scala:107)
        at scala.collection.GenSetLike.subsetOf$(GenSetLike.scala:107)
[...]

PR #1407, opened 2019-01-29
Implements fs2.concurrent.Topic with fs2.concurrent.Queue

vladimir-popov:
Implementation of the `fs2.concurrent.Topic` with  `fs2.concurrent.Queue` as subscriptions. Read more in this issue #1406 .

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -560,12 +560,12 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
   /**
     * Debounce the stream with a minimum period of `d` between each element.
     *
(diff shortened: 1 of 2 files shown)


pchlupacek:
Just a quick comment on this. If we wont manage to get perfromance correct, there may be also ather combinators affected. So instead of getting this as quick solution only for topic I would like to go through pubsub, and make sure that 
 - this cannot be solved with pubsub 
 - if this cannot be solved with pubsub make sure that no other combinators using the pubsub (i.e. broadcast) are not affected by this too 
 - fix all the other combinators, if this will be pubsub design flaw.

vladimir-popov:
Ok. I'm understand your point. Now I’ll close this PR, but we will continue use it as temporary workaround.

pchlupacek:
Thats completely fine @vladimir-popov

PR #1413, opened 2019-02-08
Fixed #1411 - updated Chunk to use scala.collection.Seq

### LINKED ISSUE #1411: Seq inconsistency between 2.12 and 2.13
Related to #1232 

The predef `Seq` in the scala language 2.13 appears to have changed form mutable to immutable. As a result this means that code using `Seq`s will behave differently across versions. For instance `Chunk.Seq`. 

I've created https://github.com/lJoublanc/fs2seqincompat to demonstrate this. `Chunk seq scala.collection.mutable.Seq(1,2,3)` won't compile under 2.13 but will under 2.12 (use e.g. ++2.13.0-M5 to switch between versions).

Not sure whether there are other parts of the API that take a seq which may be affected.

See also gitter conversation [February 8, 2019 11:04 AM](https://gitter.im/functional-streams-for-scala/fs2?at=5c5d6240604f233ab6d47d23).
@rossabaker @mpilquist

mpilquist:
Initial pass at fixing #1411. One issue that remains is that varargs on 2.13 passes an `ArraySeq` now, not a `WrappedArray`, so I'm unsure if the optimization in `Chunk.seq` is accomplishing anything. I think I looked in to this once before but will do so again.


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -534,16 +535,16 @@ object Chunk {
         override def map[O2](f: O => O2): Chunk[O2] = indexedSeq(s.map(f))
       }
 
-  /** Creates a chunk backed by a `Seq`. */
-  def seq[O](s: Seq[O]): Chunk[O] = s match {
+  /** Creates a chunk backed by a mutable `Seq`. */
```
lJoublanc: A collection.Seq an be either mutable _or_ immutable, right?

rossabaker: Correct. It is a parent of both `collection.immutable.Seq` and `collection.mutable.Seq`.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -534,16 +535,16 @@ object Chunk {
...
+  def seq[O](s: GSeq[O]): Chunk[O] = s match {
     case a: collection.mutable.WrappedArray[O] =>
       array(a.array.asInstanceOf[Array[O]])
     case v: Vector[O]      => vector(v)
     case ix: IndexedSeq[O] => indexedSeq(ix)
     case _ =>
       if (s.isEmpty) empty
       else if (s.tail.isEmpty) singleton(s.head)
-      else buffer(collection.mutable.Buffer(s: _*))
+      else buffer(collection.mutable.Buffer((s.toSeq): _*))
```
rossabaker: Would a match for `Buffer` to go through `buffer` without `s.toSeq` be worthwhile?  I doubt this gets passed `Buffer`s often, but it's what jawn-fs2 always sends.  I may end up special casing it there, even after this fix.

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: Fixed

mpilquist:
> This is not binary compatible on 2.13.0-M5, right? Is the strategy to wait until the next 2.13.0, or to say that's not subject to binary compatibility, being a standard library milestone? I'm content either way.

I was thinking of saying no binary compatibility on 2.13 milestones.

> 👍 to the spirit of the change. `collection.Seq` is not my favorite type, but we'll see it a lot until Scala 2.12 falls into disuse.

Same :)

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: Probably a good idea.

The `s.toSeq` here is weird -- I think that will no longer be necessary as of 2.13.0-RC1. I'll add a comment to double check that.

mpilquist:
Okay this one should be ready for re-review & merge.

build.sbt
```diff
@@ -249,6 +249,13 @@ lazy val core = crossProject(JVMPlatform, JSPlatform)
   .settings(
     name := "fs2-core",
     sourceDirectories in (Compile, scalafmt) += baseDirectory.value / "../shared/src/main/scala",
+    unmanagedSourceDirectories in Compile += {
+      val dir = CrossVersion.partialVersion(scalaVersion.value) match {
```
kubukoz: Doesn't sbt support this just by naming the source directory `scala-2.13-` (or something like that)? I'm looking at cats: https://github.com/typelevel/cats/tree/master/core/src/main

mpilquist: That appears to be done in build.sbt here: 
https://github.com/typelevel/cats/blob/master/build.sbt#L21

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -1,6 +1,7 @@
 package fs2
 
 import scala.collection.immutable.{Queue => SQueue}
+import scala.collection.{Seq => GSeq}
```
lJoublanc: I've noticed that this problem also extends to `IndexedSeq`. Is it necessary to give it the same treatment?

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -470,80 +474,109 @@ object Chunk {
...
+  def seq[O](s: GSeq[O]): Chunk[O] = iterable(s)
+
+  /** Creates a chunk from a `scala.collection.Iterable`. */
+  def iterable[O](i: collection.Iterable[O]): Chunk[O] = i match {
+    case ArrayBackedSeq(arr) =>
+      // arr is either a primitive array or a boxed array
+      // cast is safe b/c the array constructor will check for primitive vs boxed arrays
+      array(arr.asInstanceOf[Array[O]])
+    case v: Vector[O]                    => vector(v)
+    case ix: IndexedSeq[O]               => indexedSeq(ix)
```
lJoublanc: Suspect this is only picking up immutable `IndexedSeq`s, as explained above.

build.sbt
kubukoz: Never mind then :D

mpilquist:
> So playing around with `IndexedSeq` I've noticed that it has the same change from 2.12 -> 2.13 i.e. becomes immutable. I'm wondering whether you need to apply the same `import collection.IndexedSeq` to make this work with both types. Specifically, is there a way to check that inside `Chunk.seq` (now `Chunk.iterable`) mutable `IndexedSeq`s are matching correctly?

Thanks, fixed

PR #1419, opened 2019-02-23
Interruptible eval

SystemFw:
This makes `eval` nodes properly interruptible. For this PR I decided to go with the least intrusive approach possible, there is plenty of (relatively) low hanging fruit to make things better, simply because they were designed without taking F interruption into account.
As a mitigating factor though, a lot of code which would naively be leaky is actually in an `F.start` section plus custom Stream interruption,  and that makes it unreachable for F interruption, which is why I believe we are not seeing problematic behaviour across the board.
In any case, I'll be auditing and fixing for this when time permits, but the problematic behaviour should be fixed as is

### DIFF
--- a/core/jvm/src/test/scala/fs2/Pipe2Spec.scala
+++ b/core/jvm/src/test/scala/fs2/Pipe2Spec.scala
@@ -495,7 +495,6 @@ class Pipe2Spec extends Fs2Spec {
     }
 
     "interrupt (19)" in {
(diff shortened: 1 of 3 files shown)


SystemFw:
@kaishh This should unblock you

pchlupacek:
+1, seem good to me

kaishh:
@SystemFw Thanks!

PR #1424, opened 2019-03-02
Fix issue where switchMap could get stuck

rfkm:
This fixes issue where `switchMap` could easily get stuck due to deadlocked semaphore.

minimum repro:
```scala
import cats.effect.IOApp

object Demo extends IOApp {

  import cats.implicits._
  import cats.effect._
  import fs2.Stream

    Stream
      .range(1, 10)
      .covary[IO]
      .evalTap(i => IO(println(i)))
      .switchMap(Stream.emit)
      .compile
      .drain
      .as(ExitCode.Success)
}

// Output:
// 1
// 2
// 3
// 4
// ...no further output
```

### DIFF
--- a/core/jvm/src/test/scala/fs2/SwitchMapSpec.scala
+++ b/core/jvm/src/test/scala/fs2/SwitchMapSpec.scala
@@ -12,6 +12,15 @@ class SwitchMapSpec extends Fs2Spec with EventuallySupport {
 
   "switchMap" - {
 
+    "sanity check" in forAll { s: PureStream[Int] =>
+      runLog(
+        s.get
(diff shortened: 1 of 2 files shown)


rfkm:
Ah, `switchMap` doesn't guarantee the order? Is this intended behaviour?

rfkm:
I've closed this because I have no time to track this issue anymore. I hope someone follow up the issues:
- deadlock issue
- non-deterministic order

tom91136:
Thanks @rfkm, will look into those and the patch before the coming weekend. I use this combinator quite frequently so its good to have another pair of eyes on it.

PR #1425, opened 2019-03-05
add `toInputStreamResource` 

wedens:
I think it makes sense considering `toInputStream`  returned singleton stream.

Not sure what to do with bin compat though. Is there some branch for breaking changes?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -218,6 +218,7 @@ lazy val mimaSettings = Seq(
     ProblemFilters
       .exclude[DirectMissingMethodProblem]("fs2.io.package.invokeCallback"),
     ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.io.tcp.Socket.client"),
+    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.io.JavaInputOutputStream.toInputStream"),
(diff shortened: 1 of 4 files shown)


mpilquist:
This is a great idea. If we want to fix this in the 1.0 series, we could introduce a new name for the version that returns a resource -- `toInputStreamResource` perhaps. Then have the pipe version use the `toInputStreamResource` method in its implementation.

There's no series/1.1 yet -- we're working on a new interpreter on the topic/fs3 branch but it's a bit early to PR stuff there yet unless the changes are related to the interpreter.

SystemFw:
Data point: for `hold`, which is in a similar situation, I added `holdResource`.
Ideally we would just change it, but we can't break the api, as you point out

wedens:
> we could introduce a new name for the version that returns a resource -- toInputStreamResource perhaps. Then have the pipe version use the toInputStreamResource method in its implementation.

I've implemented this suggestion.

mpilquist:
@wedens This looks ready to merge to me once we add the mima exclusion for binary compatibility of `JavaInputOutputStream`.

wedens:
@mpilquist Done.

SystemFw:
Thank you!

PR #1427, opened 2019-03-10
Printing sinks should take ContextShift + EC

monadplus:
Add stdoutLines and stdinLines as commented in the issue https://github.com/functional-streams-for-scala/fs2/issues/1328

### DIFF
--- a/io/src/main/scala/fs2/io/io.scala
+++ b/io/src/main/scala/fs2/io/io.scala
@@ -2,8 +2,11 @@ package fs2
 
 import cats.effect.{ConcurrentEffect, ContextShift, Sync}
 
+import cats._
 import cats.implicits._
 import java.io.{InputStream, OutputStream}
+import java.nio.charset.Charset
+
 import scala.concurrent.{ExecutionContext, blocking}
 
 /**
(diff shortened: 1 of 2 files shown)


io/src/main/scala/fs2/io/io.scala
```diff
@@ -131,6 +135,25 @@ package object io {
...
+    * blocking so the execution context should be configured appropriately.
+    */
+  def stdoutLines[F[_], O](blockingExecutionContext: ExecutionContext, charset: Charset = utf8Charset)(
+      implicit F: Sync[F],
+      cs: ContextShift[F],
+      show: Show[O]): Pipe[F, O, Unit] =
+    _.map(_.show).through(text.encode(charset)).through(stdout(blockingExecutionContext))
+
+  /** Stream of `String` read asynchronously from standard input decoded in UTF-8. */
+  def stdinUtf8[F[_]](bufSize: Int, blockingExecutionContext: ExecutionContext)(
```
mpilquist: How about `stdinText` and taking a charset param defaulted to UTF8?

monadplus: Is this generalization possible ?
I though converting from bytes to string has to be done ad hoc.

mpilquist: Oh, right, we don't have a `decode(c: CharSet)` method at this point in time.

monadplus: So how do I proceed ?

PR #1432, opened 2019-03-21
Scalafix for replacing Sink with Pipe

tmciver:
One possible issue with this PR is that there are 2 input and output files associated with this scalafix rule.  Would it be preferred that they be combined into one?

### DIFF
--- a/scalafix/input/src/main/scala/fix/Fs2sinkremoval.scala
+++ b/scalafix/input/src/main/scala/fix/Fs2sinkremoval.scala
@@ -1,0 +1,16 @@
+/*
+rule = v1
+*/
+package fix
+
+import fs2._
+import fs2.{Pipe, Sink, Stream}
+import fs2.{Sink => X, Stream}
+
+object Fs2sinkremoval {
+
+  private def foo[F[_], A](s: Stream[F, A], sink: Sink[F, A]): Stream[F, Unit] =
+    s.to(sink)
+
+  def bar[F[_], A, B](): Pipe[F, A, B] = ???
+}
--- a/scalafix/input/src/main/scala/fix/Fs2sinkremoval.scala
+++ b/scalafix/input/src/main/scala/fix/Fs2sinkremoval.scala
@@ -13,4 +13,6 @@ object Fs2sinkremoval {
     s.to(sink)
 
   def bar[F[_], A, B](): Sink[F, A] = ???
+
+  def baz[F[_], A, B](): X[F, A] = ???
 }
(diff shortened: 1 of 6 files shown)


scalafix/input/src/main/scala/fix/Fs2sinkremoval.scala
```diff
@@ -0,0 +1,16 @@
+/*
+rule = v1
+*/
+package fix
+
+import fs2._
+import fs2.{Pipe, Sink, Stream}
+import fs2.{Sink => X, Stream}
```
mpilquist: Did you intend to add a `X[F, A]` example below?

scalafix/input/src/main/scala/fix/Fs2sinkremoval.scala
```diff
@@ -0,0 +1,16 @@
...
+
+object Fs2sinkremoval {
+
+  private def foo[F[_], A](s: Stream[F, A], sink: Sink[F, A]): Stream[F, Unit] =
+    s.to(sink)
+
+  def bar[F[_], A, B](): Pipe[F, A, B] = ???
```
mpilquist: Should this return `Sink[F, A]`?

tmciver: Yup. Fixed.

scalafix/input/src/main/scala/fix/Fs2sinkremoval.scala
tmciver: Not sure about this one.  @monadplus what was the intent of the `Sink` rename?

scalafix/rules/src/main/scala/fix/v1.scala
```diff
@@ -530,13 +530,20 @@ object SinkToPipeRules {
   ).asPatch
 
   def apply(t: Tree)(implicit doc: SemanticDocument): List[Patch] = {
+    //println("Tree.structureLabeled: " + doc.tree.structureLabeled)
+    //val sinkMatcher = SymbolMatcher.normalized("fs2.Sink")
+    val sinkMatcher = SymbolMatcher.exact("fs2/Sink.")
```
olafurpg: What happens with `fs2/Sink#`?

tmciver: It doesn't match but I'm not sure why.

olafurpg: What symbol is printed by `Tree.symbol` of the trees you want to match?

tmciver: BTW, what's the difference between `fs2/Sink.` and `fs2/Sink#`?  I don't think the docs are clear on that.

olafurpg: the symbol format is documented in the SemanticDB specification https://scalameta.org/docs/semanticdb/specification.html#symbol

`Sink.` is the object symbol and `Sink#` is the class or trait symbol

mpilquist:
Thanks for all the work on this one!

PR #1438, opened 2019-04-05
Add chunkMin Combinator

ChristopherDavenport:
This adds a chunkMin combinator which is like that pair of chunkLimit, but puts a combination into each chunk having atleast N elements, but does not stop their, continues with the values available to any particular Chunk if they are available and larger than the size.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -402,7 +402,7 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     *
     * @example {{{
     * scala> (Stream(1,2) ++ Stream(3,4) ++ Stream(5,6,7)).chunkMin(3).toList
-    * res0: List[Chunk[Int]] = List(Chunk(1,2,3,4), Chunk(5,6,7))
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -392,6 +392,54 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    * If `allowFewerTotal` is true,
+    * if the stream is smaller than N, should the elements be included
+    *
+    * @example {{{
+    * scala> (Stream(1,2) ++ Stream(3,4) ++ Stream(5,6,7)).chunkMin(3).toList
+    * res0: List[Chunk[Int]] = List(Chunk(1, 2, 3, 4), Chunk(5, 6, 7))
+    * }}}
+    */
+  def chunkMin(n: Int, allowFewerTotal: Boolean = true): Stream[F, Chunk[O]] = {
+    def go[A](accFull: Option[Chunk[A]],
```
mpilquist: Instead of `Option[Chunk[A]]`, can you use `Chunk.Queue[A]`? This lets us avoid concatenating each chunk with the accumulator as it is received -- instead, they are all stored in the queue and then can be concatenated/taken from/dropped from as necessary in an efficient way.

PR #1440, opened 2019-04-10
add FunctionK instances for Stream and Pull

bpholt:
I'm not sure if these are the right places for these functions, but it seemed like a good start. This PR was prompted by reading @gvolpe's TypeLevel blog post [“Tagless Final algebras and Streaming”](https://typelevel.org/blog/2018/05/09/tagless-final-streaming.html) and trying to implement an interpreter parameterized by `F[_], G[_]`. For example, given this algebra:

```scala
import cats._, cats.arrow.FunctionK, cats.effect.IO

trait FooAlg[F[_], G[_]] {
  def foo(): G[Int]
  def bar(): F[Int]
}
```

if I want to implement `foo()` in terms of `bar()` without fixing `G` as `Stream[F, ?]`, I think one must have an `F ~> G` available.

```scala
class FooImpl[F[_] : Applicative, G[_] : Monad](implicit FtoG: F ~> G) extends FooAlg[F, G] {
  def foo(): G[Int] = FtoG(bar()).flatMap(x => Applicative[G].pure(x)) // actually this strings together several consecutive functions in monadic fashion
  def bar(): F[Int] = Applicative[F].pure(42) // actually this does IO to call a remote service
}
```

```scala
scala> val foo = new FooImpl[IO, Stream[IO, ?]]
foo: FooImpl[cats.effect.IO,[β$0$]fs2.Stream[cats.effect.IO,β$0$]] = FooImpl@4ee2479d

scala> foo.foo()
res0: fs2.Stream[cats.effect.IO,Int] = Stream(..)

scala> foo.bar()
res1: cats.effect.IO[Int] = IO(42)

scala> foo.foo().compile.toList.unsafeRunSync
res2: List[Int] = List(42)
```

I'm mainly interested in the instance for `Stream`, but I thought I'd add the one for `Pull` too, for the sake of completion.

I didn't see any laws available for `FunctionK` in cats-laws, so I didn't add any tests beyond the doctests.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -1,6 +1,7 @@
 package fs2
 
 import cats._
+import cats.arrow.FunctionK
 import cats.effect._
 import fs2.internal._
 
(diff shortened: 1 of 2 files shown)


mpilquist:
Thanks! One thing I recently noticed is that `Stream#translate` doesn't take a `Concurrent[G]` anymore, which seems like an oversight to me. For example:

```scala

// Returns after 1 second

// Hangs
```

bpholt:
Thanks @mpilquist.

> The "right" fix with the 1.0 design is changing the signature of translate to take an Option[Concurrent[G]], but then this breaks mapK (and binary compatibility).
>
> Another option is documenting the limitations of translate and adding a translateConcurrent overload that takes an implicit Concurrent[G].

I agree this seems like a problem, but I don't have an opinion on how to address it. Sorry that's not more helpful.

PR #1442, opened 2019-04-11
Avoid calculate size of emits

ryoppy:
LinearSeq is caluculate every time of size. I refactored of emits.


mpilquist:
Thanks! I wasn't aware that `Nil` matched any seq (e.g. `Vector.empty`).

PR #1451, opened 2019-05-02
Add fs2-google-pubsub as a related project

CremboC:
Obviously up to you guys to decide whether this should be added or not. I guess as a word of warning, the API isn't very stable, but otherwise we have been using that library in production for around 6 months now.


mpilquist:
Always happy to add an FS2 based library! Thanks!

PR #1452, opened 2019-05-03
kind-projector has moved to typelevel org

SethTisue:
it would be helpful to have this merged for the Scala community
build. normally we're able to override this kind of thing in our
own config but it's not working in this case for some reason


mpilquist:
Thanks!

PR #1453, opened 2019-05-03
Update dependencies for 2.13.0-RC1

kubukoz:
Apparently, there are some new ambiguities in scalacheck under 2.11:

```
[error] /home/travis/build/functional-streams-for-scala/fs2/core/shared/src/test/scala/fs2/ChunkGen.scala:233:10: ambiguous reference to overloaded definition,
[error] both method apply in object Cogen of type (f: (org.scalacheck.rng.Seed, List[A]) => org.scalacheck.rng.Seed)org.scalacheck.Cogen[List[A]]
[error] and  method apply in object Cogen of type (implicit ev: org.scalacheck.Cogen[List[A]])org.scalacheck.Cogen[List[A]]
[error] match expected type ?
[error]     Cogen[List[A]].contramap(_.toList)
[error]          ^
[error] /home/travis/build/functional-streams-for-scala/fs2/core/shared/src/test/scala/fs2/FreeCSpec.scala:36:88: ambiguous reference to overloaded definition,
[error] both method apply in object Cogen of type (f: (org.scalacheck.rng.Seed, Unit) => org.scalacheck.rng.Seed)org.scalacheck.Cogen[Unit]
[error] and  method apply in object Cogen of type (implicit ev: org.scalacheck.Cogen[Unit])org.scalacheck.Cogen[Unit]
[error] match expected type ?
[error]       f <- arbFunction1[Result[A], FreeC[F, A]](Arbitrary(freeGen[F, A](fDepth)), Cogen[Unit].contramap(_ => ())).arbitrary
[error]                                                                                        ^
[error] /home/travis/build/functional-streams-for-scala/fs2/core/shared/src/test/scala/fs2/TestUtil.scala:100:73: ambiguous reference to overloaded definition,
[error] both method apply in object Cogen of type (f: (org.scalacheck.rng.Seed, List[A]) => org.scalacheck.rng.Seed)org.scalacheck.Cogen[List[A]]
[error] and  method apply in object Cogen of type (implicit ev: org.scalacheck.Cogen[List[A]])org.scalacheck.Cogen[List[A]]
[error] match expected type ?
[error]     implicit def pureStreamCoGen[A: Cogen]: Cogen[PureStream[A]] = Cogen[List[A]].contramap[PureStream[A]](_.get.toList)
[error]                                                                         ^
[error] three errors found
```

PR #1454, opened 2019-05-04
Remove scala-collection-compat dependency

travisbrown:
It seems like a shame to have a dependency on scala-collection-compat (which is currently broken and waiting for an incompatible 2.0 release) when all that fs2 needs from it is one alias and one extension method.

This change shouldn't affect compatibility for any Scala version and would unblock #1453.

r? @mpilquist

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -53,7 +53,6 @@ lazy val commonSettings = Seq(
     "org.typelevel" %%% "cats-laws" % "1.6.0" % "test",
     "org.typelevel" %%% "cats-effect" % "1.3.0",
     "org.typelevel" %%% "cats-effect-laws" % "1.3.0" % "test",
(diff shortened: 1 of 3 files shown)


mpilquist:
Love this idea! Thanks!

travisbrown:
The 2.13 build is failing because of #1452, but so is master, so ¯\_(ツ)_/¯. I've got a branch including the updates in #1453 and can verify that it works on 2.13 there.

travisbrown:
Uh, actually I was wrong about the 2.13 part—this change doesn't work on 2.13 alone. Will open a new PR now.

I'm a fan of the community build but it seems like a shame to break builds to make it happy. Maybe #1452 should have waited for the full 2.13.0-RC1 update?

mpilquist:
Yeah agreed on #1452 -- I should have checked CI status on that one too.

PR #1458, opened 2019-05-04
Added chunkMin (updated)

mpilquist:
This is an updated version of #1438. I simplified the implementation a bit and modified the logic to output an accumulated chunk as soon as the accumulator reaches the desired size. Before my commit, the accumulated chunk would only be output after a new chunk is received which has a size >= n. For a stream with small chunks, this could result in a memory leak.


mpilquist:
@ChristopherDavenport Can you take a look at 672b16c?

PR #1472, opened 2019-05-17
Adding Chromaprint.scala to projects using fs2

mgdigital:


mpilquist:
Looks like a really cool project!

PR #1475, opened 2019-05-22
Fix recursive aux functions differences

asakaev:
Sometimes it's "loop" and sometimes "go". If there is no semantics in "loop" I think is better to make them uniform.


diesalbla:
@asakaev I think that a few of these functions admit a `@tailrec` annotation. Could you add them where that is the case?

asakaev:
@diesalbla sure, done!

PR #1477, opened 2019-05-24
CompileScope.open: rearrange operations

diesalbla:
We reorganise the operations of the `open` method of the `CompileScope` class.

We change `createCompileScope` from being an `Option` of the `InterruptContext` to directly building the new scope. This hides the `newScopeId` in this F-value.

This operation saves the allocations of a `Some` and a Tuple2 objects, although now the `CompileScope` is now built regardless of path. Not a big loss, since this is likely the frequent case.

_Note to reviewers:_ most line changes are due to whitespace, indentation, and braces.

_Note:_ I have looked for a `modifyM` method, which could allow us to avoid allocations until the scope is known to be open. Reading in `cats-effect`, I just found [this issue](https://github.com/typelevel/cats-effect/issues/451), which mentions it tangentially.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/CompileScope.scala
+++ b/core/shared/src/main/scala/fs2/internal/CompileScope.scala
@@ -109,40 +109,39 @@ private[fs2] final class CompileScope[F[_]] private (
      *     or as a result of interrupting this scope. But it should not propagate its own interruption to this scope.
      *
      */
-    def createScopeContext: F[(Option[InterruptContext[F]], Token)] = {
+    val createCompileScope: F[CompileScope[F]] = {
       val newScopeId = new Token
       self.interruptible match {
         case None =>
-          F.pure(InterruptContext.unsafeFromInterruptible(interruptible, newScopeId) -> newScopeId)
+          val iCtx = InterruptContext.unsafeFromInterruptible(interruptible, newScopeId)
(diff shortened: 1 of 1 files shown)


pchlupacek:
I am ok with this change although I would like to explore modifyM approach more in detail before merging this.

diesalbla:
@pchlupacek I should have elaborated more. There is no method `modifyM` in the `Ref` class of `cats-effect`. The issue I linked was a proposal in `cats-effect`, to add a similar effect-full update  methods in `MVar`, not to `Ref`. In that issue, there is a reference to an implementation of [`RefM` in the `zio` library](https://github.com/scalaz/scalaz-zio/blob/master/core/shared/src/main/scala/scalaz/zio/RefM.scala#L37), library, but as you can see it is not a simple data structure. 

This makes me doubt if a `Ref` with an effect-full update method would be a simple thing, worth using here.

pchlupacek:
@diesalbla ok, I thought you meant it is already there, didn't check the link. Anyhow, I have no problem to get this in even-though I am not sure of measurable benefits this may bring. @mpilquist any thoughts?

mpilquist:
I'm okay with this too, though I'd like to be very conservative in changes to internals for allocation saving reasons unless we can benchmark the improvements.

PR #1478, opened 2019-05-24
Port unit tests from fs3 topic

geoffreytchew:
Existing series/1.0 test classes were deleted and replaced with those from the topic/fs3 branch. 
A few minor modifications were made: 

1. Disabled fatal warnings for test compilation because sbt-doctest generated tests generate warnings which lead to test failures.
2. Changed "repeatN 3" test in StreamSpec.scala to _pending_ due to frequent (albeit intermittent) failures
3. Re-enabled three tests that are currently marked pending in topic/fs3 because they consistently passed without error when run against the series/1.0 code. The re-enabled tests are: StreamSpec.scala: "resources acquired in outer stream are released after inner streams complete", "run finalizers of inner streams first", and "when inner stream fails, inner stream finalizer run before the primary one"
4. Removed calls to *void* which is not present in trait Pull in the series/1.0 branch


SystemFw:
Oh wow, that's a lot of work, thanks a lot!
Quick question, I've seen that some things were deleted (like `GroupWithinSpec` or `ResourceCompilationSpec`), where they moved in other files? (before I go review the whole thing)

mpilquist:
Yep, those are mostly in StreamSpec now

SystemFw:
quick heads up: there are several test failures , which at quick glance looks like timeouts (so possibly things hanging)

geoffreytchew:
I'll take a look at the test failures

core/shared/src/test/scala/fs2/Fs2Spec.scala
```diff
@@ -1,48 +1,114 @@
...
+    with GeneratorDrivenPropertyChecks
+    with Checkers
+    with MiscellaneousGenerators
+    with ChunkGenerators
+    with StreamGenerators
+    with EffectTestSupport
+    with TestPlatform {
 
-abstract class AsyncFs2Spec extends AsyncFreeSpec with Fs2SpecLike with AsyncTimeLimitedTests {
+  implicit val timeout: FiniteDuration = 10.seconds
```
mpilquist: Bump this back to 60 seconds -- this is likely the cause of most/all of the timeouts on Travis

SystemFw: btw, maybe this requires a different PR, but I think we should investigate using `TestContext` more instead of actual time

core/shared/src/test/scala/fs2/StreamSpec.scala
```diff
@@ -2125,6 +2125,7 @@ class StreamSpec extends Fs2Spec {
       }
 
       "resources acquired in outer stream are released after inner streams complete" in {
+        pending
```
mpilquist: Odd, these are passing on series/1.x AFAIK

geoffreytchew: I'm commenting them out for now, but may re-add them after I address the fact that flickersOnTravis is not behaving as expected (does not exclude test on Travis like it should)

geoffreytchew:
All are passing after my latest round of changes. In summary: Fixed the issue with flickersOnTravis not being applied (related to: https://stackoverflow.com/questions/27486540/jvm-options-not-passed-on-to-forked-process) and marked one additional test as pending that was failing both locally and on Travis.

mpilquist:
Awesome work!

PR #1479, opened 2019-05-26
WIP: Support Scala 2.13.0-RC2

rossabaker:


build.sbt
```diff
@@ -49,20 +49,13 @@ lazy val commonSettings = Seq(
...
-    "org.typelevel" %%% "cats-core" % "2.0.0-M1",
-    "org.typelevel" %%% "cats-laws" % "2.0.0-M1" % "test",
-    "org.typelevel" %%% "cats-effect" % "2.0.0-M1",
-    "org.typelevel" %%% "cats-effect-laws" % "2.0.0-M1" % "test",
-    "org.scalatest" %%% "scalatest" % "3.0.8-RC2" % "test"
+    "org.typelevel" %%% "cats-core" % "2.0.0-M2",
+    "org.typelevel" %%% "cats-laws" % "2.0.0-M2" % "test",
+    "org.typelevel" %%% "cats-effect" % "2.0.0-M2",
+    "org.typelevel" %%% "cats-effect-laws" % "2.0.0-M2" % "test",
+    "org.scalacheck" %%% "scalacheck" % "1.14.0" % "test",
```
rossabaker: If we're on cats-laws-2.x and cats-effect-laws-2.x, we should be on scalacheck-1.14.0.

build.sbt
```diff
@@ -73,7 +66,7 @@ lazy val commonSettings = Seq(
     implicit val contextShiftIO: ContextShift[IO] = IO.contextShift(global)
     implicit val timerIO: Timer[IO] = IO.timer(global)
   """,
-  doctestTestFramework := DoctestTestFramework.ScalaTest,
+  doctestTestFramework := DoctestTestFramework.ScalaCheck,
```
rossabaker: Working around https://github.com/tkawachi/sbt-doctest/issues/130

rossabaker:
Blocked on https://github.com/scalatest/scalatestplus-testng/issues/3.  We should consider porting those tests to reduce test dependency liability.

rossabaker:
Nope, easier said than done: it (appropriately) uses the reactive-streams-tck, which uses testng.

rossabaker:
Also blocked on https://github.com/scodec/scodec-bits/pull/118.

mpilquist:
I was hoping to get #1478 in first given the ScalaTest changes. WDYT?

rossabaker:
That looks reasonable, and this is blocked on testng, so :+1:*2.

mpilquist:
@rossabaker I updated this and fixed a few test issues introduced by the new test suite (unrelated to this PR).

rossabaker:
What is the version strategy going to be for this?  fs2 doesn't publish laws (I don't think?), so it doesn't have to be major.  But cats-effect-2.x will not be binary compatible in 2.11, and if the `fromFuture` ticket goes through, anywhere.  Does this mean fs2-2.0?

mpilquist:
Argh, good catch. Maybe 1.1?

mpilquist:
I'm going to merge this to series/1.0 at least and then figure out where to go from there.

PR #1480, opened 2019-05-27
Encapsulate CompileScope methods

diesalbla:
Narrow the CompileScope class. We remove the method "openAncestor", and hide as private `findSelfOrAncestor` and `findSelfOrChild`. We expose two methods, `descendsFrom` (a Boolean query), and the `closeRelative` method, which includes the logic to find a related scope, close it, and return any error that was risen in that Scope. We add documentation for these methods.

We rewrite the code of the `compileLoop-go` for the `CloseScope` node. We use of the `closeRelative` method. This removes the `closeAndGo`.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -301,49 +301,34 @@ private[fs2] object Algebra {
               }
 
             case close: CloseScope[F] =>
-              def closeAndGo(toClose: CompileScope[F], ec: ExitCase[Throwable]) =
-                F.flatMap(toClose.close(ec)) { r =>
-                  F.flatMap(toClose.openAncestor) { ancestor =>
-                    close.interruptedScope match {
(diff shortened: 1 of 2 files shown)


pchlupacek:
Could you please elaborate on benefits of this change?

diesalbla:
> Could you please elaborate on benefits of this change?

- It reduces the API of the `CompileScope` class, avoiding the need to expose the "family tree" of the compile scopes. 
- The changes also make the compilation case for the `CloseScope` operation more readable: by removing the inner method `closeAndGo` the flow in that case is top-down.
- It also adds or clarifies some comments.

pchlupacek:
I am not seeing enough benefits for this. My feeling is that this is just different approach to the implementation that is already there. As there are no performance/correctness improvements and scope's api is in fact private to fs2, I am -1 on this.

diesalbla:
This code has been changed in `series/1.1`, so this is no longer needed.

PR #1481, opened 2019-05-27
Performance trick: NoException.

diesalbla:
There are many places in FS2 Internals where an `Option[Throwable]`, or an `Either[Throwable, Unit]` is used to indicate presence or absence of an error. This has a memory cost: although the first one only has extra allocation when there is an error, the second one always allocates either a
Right or a Left.

It turns out there is a zero-extra-allocations way to represent the absence of exceptions: since `Throwable` is a non-sealed trait, we can represent any special case with another sub-type, namely an object, that extends from `Throwable`. In particular, we can represent that there are no Exceptions, by using a special sub-type of Exception.

### DIFF
--- a/core/shared/src/main/scala/fs2/CompositeFailure.scala
+++ b/core/shared/src/main/scala/fs2/CompositeFailure.scala
@@ -20,25 +20,26 @@ object CompositeFailure {
             rest: List[Throwable] = List.empty): CompositeFailure =
     new CompositeFailure(first, NonEmptyList(second, rest))
 
-  def fromList(errors: List[Throwable]): Option[Throwable] = errors match {
-    case Nil                     => None
-    case hd :: Nil               => Some(hd)
-    case first :: second :: rest => Some(apply(first, second, rest))
+  def fromList(errors: List[Throwable]): Throwable = errors match {
+    case Nil                     => NoExceptionIsGoodException
+    case hd :: Nil               => hd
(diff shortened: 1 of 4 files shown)


pchlupacek:
Not sure that pursuing this with erasing types from the result is good idea. I think performance benefits will have to be substantial, and I am quite conservative how `None` differs from `NoExceptionIsAGoodException` introduced. `Some(Throwable)` shall not be usually on main execution path of the program, so I don't really see any benefits here.

diesalbla:
@pchlupacek So, if the choice is between having a `Some(---)` instance or having the  `None` singleton, then there is only allocation costs when there is an error. The problem comes if you have an `Either[Throwable, Unit]`, in which case you may be adding extra cost. But there may be an easy solution for that... 

In general, I wonder if there could be a way to avoid using `Some`, `Left` or `Right` intermediaries, and use just special token-like values; even if it is less generic.

SystemFw:
I'm fairly negative on this: this sort of encoding cannot do `Option[Option[Throwable]]`, and in general would replace a set of useful transformations (`map` and so on) with ugly checks, in parts of the code which are already complex (internal interpreter, concurrent ops).
More in general, I don't think removing single allocations "randomly" pays for itself in the performance vs opaqueness scale, I'd like proof that the allocations being removed are actually bottlenecks in the hot path.

pchlupacek:
@SystemFW agree

PR #1482, opened 2019-05-27
Result: remove Unapply method.

diesalbla:
The `unapply` method in the Result companion object was completely
similar to a `isInstanceOf` method, without any value extraction.
Thus, we remove it, and we replace its only use with a combination of
a type-isInstanceof-pattern and an `asInstanceOf`.

_Note_ two of the line changes is whitespace by `scalafmt`.


pchlupacek:
I think current encoding is better -1 to this

diesalbla:
In light of a [recent comment](https://github.com/functional-streams-for-scala/fs2/pull/1483#issuecomment-496477480), it seems that this goes against the preferred code style. Withdrawing.

PR #1483, opened 2019-05-27
FreeC.flatMap: avoid creating objects

diesalbla:
A small performance tweak to FreeC.flatMap: in the `Interrupted` or `Failed` classes, instead of creating a new object, just return the receiver object as an instance of the new target type.

Also, unify the three `Result` cases of `FreeC.translate`.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/FreeC.scala
+++ b/core/shared/src/main/scala/fs2/internal/FreeC.scala
@@ -32,8 +32,8 @@ private[fs2] sealed abstract class FreeC[F[_], +R] {
           case Result.Pure(r) =>
             try f(r)
             catch { case NonFatal(e) => FreeC.Result.Fail(e) }
-          case Result.Interrupted(scope, err) => FreeC.Result.Interrupted(scope, err)
-          case Result.Fail(e)                 => FreeC.Result.Fail(e)
+          case r @ Result.Interrupted(scope, err) => r.asInstanceOf[FreeC[F, R2]]
+          case r @ Result.Fail(e)                 => r.asInstanceOf[FreeC[F, R2]]
       }
     )
 


pchlupacek:
The resulting casts are unsafe, resulting in potential bugs -1 to this

diesalbla:
@pchlupacek I am afraid I cannot see how these casts, from `Result`, could be unsafe. Could you elaborate?

raulraja:
@pchlupacek this is a common technique used across many FP libs to avoid extra allocation. For example in cats-effect https://github.com/typelevel/cats-effect/blob/cbd7a26cba0a6514912029a1c2c59ef813d28ca3/core/shared/src/main/scala/cats/effect/package.scala#L46

pchlupacek:
Any cast is essentially turning off the type checker. We use it very rarely, in situations where there is no other approach available. Since the original code does not require it, I am not seeing any benefit from introducing it to code. 

This is specifically case here, hence optimisation is on the fail-path, where there is no reason why performance will be any issue on interrupted/fail path in flatMap.

diesalbla:
Pavel is right in principle: any use of casting is a manual override, and inserts a direct assertion, into the facts inferred by the compiler. An override that, being human-written, could be wrong; and should therefore be used carefully and prudently. 

On the other hand, this is one of case in which we have some extra cost forced by the limitations of the type-checker. Ideally, one would like to express in the type system that a `Result` is a `FreeC[F, R]` _of every_ `F[_]` you may want. For instance, if the `F` was co- or contra-variant, you would just make `Result[R]` to extend from `FreeC[Nothing, R]`, or `FreeC[Any, R]`, but this leads to some problems in pattern-matches. 

For instance, I have tried writing something more simple, just saying that `r` is also a `FreeC[F, R2]`, 
```Scala
case r @ Result.Interrupted(scope, err) => (r: FreeC[F, R2])
case r @ Result.Fail(e)                 => (r: FreeC[F, R2])
```
But for this, the compiler gives me the following error: 
```Scala
[error] Note: Any >: F (and fs2.internal.FreeC.Result.Interrupted[Any,Any] <: fs2.internal.FreeC[Any,fs2.INothing]), but class FreeC is invariant in type F.
[error] You may wish to define F as -F instead. (SLS 4.5)
```

_Note:_ Above I refer to them as limitations of the type-checker, but I am not expert enough to be sure if any type-checker for the Scala type-system could do any different.

SystemFw:
Just for context: in the past we have actually worked hard to _remove_ casts, not so much in FreeC but in the core interpreter. We had to use them because of GADT limitations, but they made the code opaque and bug-prone, especially when you have to modify it 6 months later.

I'm not dead set against this PR because it's simple, however the problem is that you are turning off the type checker _for sure_, in exchange for a performance boost _in theory_.

Experience shows that you can't just reason about perf in a "mathematical" way, thinking that some approach  ought to be faster from first principles (see the `Segment` tale), you need to reason about it in a "scientific" way, i.e. measuring. 

So I guess all I'm saying is that if the only point of the PR is a perf boost, and it comes at a cost in safety or clarity, I'd like to see a benchmark showing the perf boost to be substantial enough to be worth it.

pchlupacek:
I think we shall make this a rule of thumb, that unless there is sufficient prove, that PR claims performance boost, it is not considered. That way owner of the PR has to make a bit more homework OTOH reviewers clearly see the benefit and immediately can reason about trade-off (if any).

mpilquist:
Thanks @diesalbla for the PR! Closing this one based on the comments from Pavel and Fabio. You can reopen if benchmarks show significant improvements.

PR #1486, opened 2019-05-31
Avoid executing function twice in `mapFilter`

SystemFw:
Originally reported by @Taig  

```scala
Stream.emit(1).mapFilter { x => println(x); Some(x) }
1
1
```

The reason is that `mapFilter` goes through `collect`, which takes a `PartialFunction` and says
```scala
val o = apply(i)
if (pf.isDefinedAt(o))
   b += pf(o)
```
generally `pf.isDefinedAt` does not require the body of `pf` to be evaluated when it's defined by pattern matching, but in this case it's coming through `Function.unlift`, and therefore the body executes twice.

In terms of correctness, this is only observable via side-effects, but it still feels weird and I don't think it was foreseen, hence this quick fix

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -1558,8 +1558,19 @@ object Chunk {
         Chunk.buffer(buf.result)
       }
       override def functor: Functor[Chunk] = this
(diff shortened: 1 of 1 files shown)


SystemFw:
I should stop writing PRs in early mornings. I'll fix the compile errors

PR #1490, opened 2019-06-03
Replace pool used in example blocking execution context

mtomko:
We probably don't want to use a fixed thread pool in a blocking context; update the examples accordingly. This is per a discussion with @djspiewak in the cats-effect gitter room this morning (couldn't figure out how to link directly to it).


SystemFw:
Thanks!

PR #1492, opened 2019-06-06
Propagate error from inner stream of parJoin before continue on with ++.

### LINKED ISSUE #1491: parJoin on error stream ++ Stream.never doesn't terminate
This stream doesn't terminate:
```scala
Stream.emit(
 Stream.raiseError[IO](new IllegalStateException("kaboom"))
).parJoinUnbounded ++ Stream.never[IO]
```

Full repro: https://scastie.scala-lang.org/5FqDk4E7T2i1NrMm81lVBg

AdamChlupacek:
Fixes #1491

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2040,13 +2040,15 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
                         .fold[F2[Unit]](F2.unit)(F2.raiseError)
                     }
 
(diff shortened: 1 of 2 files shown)


AdamChlupacek:
The test suit is being weird, it seems to be running all the tests at the same time. At least locally it fires off all the tests spikes cpu to max. Also the error with which it failed on travis seems to indicate something weird.

mpilquist:
Yeah, not sure what's going on with ScalaTest concurrency -- I've noticed the same recently. Bill Venners suggested mixing in OneInstancePerTest, which is on my list to try.

AdamChlupacek:
Alright, good to hear.

PR #1493, opened 2019-06-06
Restored default ScalaTest execution context in tests to avoid interm…

mpilquist:
…ittent ConcurrentModificationExceptions

### DIFF
--- a/core/shared/src/test/scala/fs2/EffectTestSupport.scala
+++ b/core/shared/src/test/scala/fs2/EffectTestSupport.scala
@@ -7,12 +7,12 @@ import scala.concurrent.{ExecutionContext, Future}
 import org.scalatest.{Assertion, Succeeded}
 
 trait EffectTestSupportLowPriority {
-  implicit def executionContext: ExecutionContext
+  protected def executionContext: ExecutionContext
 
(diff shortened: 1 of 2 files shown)


mpilquist:
This is a pretty ugly workaround but it works for now. I sent the following reproduction case to @bvenners:

```scala
 "org.scalatest" %% "scalatest" % "3.1.0-SNAP11" % "test",

package scratch

import scala.concurrent._

import org.scalatest.Succeeded
import org.scalatest.freespec.AsyncFreeSpec

class Repo extends AsyncFreeSpec {

  implicit override val executionContext: ExecutionContext =
    scala.concurrent.ExecutionContext.Implicits.global

  ('a' to 'z').map { x =>
    x.toString - {
      (1 to 10).map(i => i.toString in Future(Succeeded))
    }
  }
}

java.util.ConcurrentModificationException: Two threads have apparently attempted to run a suite at the same time. This has resulted in both threads attempting to concurrently change the current Informer. Suite class name: scratch.Repo
```

mpilquist:
@AdamChlupacek FYI I think this should help until ScalaTest finds the bug.

PR #1495, opened 2019-06-07
Added translateInterrupt - fixes #1443

### LINKED ISSUE #1443: Stream#translate forgets ability to interrupt effect evaluation
One thing I recently noticed is that `Stream#translate` doesn't take a `Concurrent[G]` anymore, which seems like an oversight to me. For example:

```scala

// Returns after 1 second

// Hangs
```

The interpreter uses the `Concurrent` instance to interrupt effect evaluation. When we translate with the identity `FunctionK`, the interpreter forgets about the old `Concurrent[IO]` it had.

Another option is documenting the limitations of `translate` and adding a `translateConcurrent` overload that takes an implicit `Concurrent[G]`.

mpilquist:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2533,10 +2533,23 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
+    * Note: the resulting stream is *not* interruptible in all cases. To get an interruptible
+    * stream, `translateInterruptible` instead, which requires a `Concurrent[G]` instance.
     */
   def translate[F2[x] >: F[x], G[_]](u: F2 ~> G): Stream[G, O] =
     Stream.fromFreeC[G, O](Algebra.translate[F2, G, O](get[F2, O], u))
 
+  /**
+    * Translates effect type from `F` to `G` using the supplied `FunctionK`.
+    *
+    * Note: the resulting stream is *not* interruptible in all cases. To get an interruptible
```
SystemFw: This should be deleted

mpilquist: Thanks fixed

PR #1497, opened 2019-06-08
Updated fs2-io to ensure blocking calls are not made via `F.delay`

### LINKED ISSUE #1431: Properly handle blocking IO in watch API
I've stumbled upon a fact that `fs2.io.file.watch` [uses blocking API internally](https://github.com/functional-streams-for-scala/fs2/blob/a34ed54d3b9c07b98e704862cd6d081edbcdbdca/io/src/main/scala/fs2/io/file/Watcher.scala#L280) - it's written in such a way that it is almost always is in `WatchService.poll(Long, TimeUnit)`. If someone runs `fs2.io.file.watch` on the CPU pool (like I did), it basically hogs a single thread for itself, which is especially noticable when running on a single CPU core.

It seems that the correct behavior here would be to use the non-blocking `WatchService.poll()` instead, and control the poll frequency using `cats.effect.Timer`. Otherwise, `Watcher` and related functions should use the standard `ExecutionContext`/`ContextShift` idiom.

mpilquist:
Updated file, file watching, tcp, and udp support. With the exception of tcp & udp, api changes are limited to adding `blockingExecutionContext: ExecutionContext` params.

The tcp and udp packages have more signification changes:
- Removed the deprecated constructors
- Removed the implicit dependencies on unsafe constructs -- `AsynchronousChannelGroup` and `AsynchronousSocketGroup`. Replaced these with safe `SocketGroup` objects that are constructed in a `Resource[F, SocketGroup]`. E.g. instead of `Socket.server` taking an implicit ASG, you now call `SocketGroup(...).use(sg => sg.server(...))`. The test classes (`tcp.SocketSpec` and `udp.SocketSpec`) have good examples of what would need to change at source level.

Fixes #1431


io/src/main/scala/fs2/io/io.scala
```diff
@@ -178,4 +178,16 @@ package object io {
...
       implicit F: ConcurrentEffect[F]): Resource[F, InputStream] =
     JavaInputOutputStream.toInputStream(source)
+
+  /**
+    * Like `Sync#delay` but evaluates the thunk on the supplied execution context
+    * and then shifts back to the default thread pool.
+    */
+  private[io] def blockingDelay[F[_], A](blockingExecutionContext: ExecutionContext)(
+      thunk: => A)(implicit F: Sync[F], cs: ContextShift[F]): F[A] =
+    cs.evalOn(blockingExecutionContext)(F.delay(blocking(thunk)))
```
SystemFw: Is the use of `blocking` needed here? I was under the impression that pretty much only `global` deals with it, and if the cats-effect change goes through, we'd probably always use a `cachedThreadPool` here anyway

mpilquist: Yeah, only used with contexts backed by FJPool. I'll remove

durban: The `blockingExecutionContext` seem to be passed in through a public API. Which means, that it may be a ForkJoinPool (even if it's not advisable). In which case `blocking` is useful. And in other cases it doesn't hurt. (Or does it?)

SystemFw: The idea is that it will be replaced by https://github.com/typelevel/cats-effect/pull/556

io/src/main/scala/fs2/io/tcp/SocketGroup.scala
```diff
@@ -0,0 +1,347 @@
...
+    Resource.make(socket)(_ =>
+      blockingDelay(blockingExecutionContext)(if (ch.isOpen) ch.close else ()).attempt.void)
+  }
+}
+
+object SocketGroup {
+
+  def apply[F[_]: Sync: ContextShift](
+      blockingExecutionContext: ExecutionContext,
+      nonBlockingThreadCount: Int = 4,
```
tpolecat: Why not 5? Should this be a function of available processors?

mpilquist: Fixed, thanks

io/src/main/scala/fs2/io/io.scala
mpilquist: I agree that the `blocking` call doesn't hurt though I don't want to encourage use of FJPool as that relies on all blocking thunks to be wrapped in `blocking`, which isn't true in the wider cats-effect ecosystem.

durban: Ok, I didn't realize the `ExecutionContext` will be replaced by this new `Blocker`.

PR #1500, opened 2019-06-10
Delete evalMap return restriction

AdrianRaFo:
This PR deletes the `evalMap` unnecessary restriction of receive a function returning `Unit` because the result type of that function will be actually ignored.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,3 +1,1 @@
--J-Xms2g
--J-Xmx4g
 -J-XX:MaxMetaspaceSize=512m
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,1 +1,3 @@
+-J-Xms2g
+-J-Xmx4g
 -J-XX:MaxMetaspaceSize=512m
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -972,7 +972,7 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     * Not as powerful as `observe` since not all pipes can be represented by `O => F[A]`, but much faster.
     * Alias for `evalMap(o => f(o).as(o))`.
     */
-  def evalTap[F2[x] >: F[x]: Functor, A](f: O => F2[A]): Stream[F2, O] =
+  def evalTap[F2[x] >: F[x]: Functor](f: O => F2[_]): Stream[F2, O] =
     evalMap(o => f(o).as(o))
 
   /**
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -976,13 +976,6 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     evalMap(o => f(o).as(o))
 
   /**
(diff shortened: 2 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -967,11 +968,11 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
...
 
   /**
-    * Like `observe` but observes with a function `O => F[Unit]` instead of a pipe.
-    * Not as powerful as `observe` since not all pipes can be represented by `O => F[Unit]`, but much faster.
+    * Like `observe` but observes with a function `O => F[A]` instead of a pipe.
+    * Not as powerful as `observe` since not all pipes can be represented by `O => F[A]`, but much faster.
     * Alias for `evalMap(o => f(o).as(o))`.
     */
-  def evalTap[F2[x] >: F[x]: Functor](f: O => F2[Unit]): Stream[F2, O] =
+  def evalTap[F2[x] >: F[x]: Functor, A](f: O => F2[A]): Stream[F2, O] =
```
djspiewak: You can avoid the extra type parameter here while still maintaining the same functionality by using `O => F[_]` rather than `O => F[A]`.

Daenyth: I'd prefer this as well, because it makes manual parameter passing more convenient, if people prefer it for readability etc.

AdrianRaFo:
@djspiewak addressed

core/shared/src/main/scala/fs2/Stream.scala
AdrianRaFo: it's without the `A` type parameter right now, but it's up to you to choose the way you want. Just tell me and I'll addressed

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -835,7 +836,7 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
     * scala> Stream.range(0,10).dropRight(5).toList
     * res0: List[Int] = List(0, 1, 2, 3, 4)
     * }}}
-    */
```
mpilquist: Extra `*` here

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -967,11 +968,11 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing,
   }
 
   /**
-    * Like `observe` but observes with a function `O => F[Unit]` instead of a pipe.
-    * Not as powerful as `observe` since not all pipes can be represented by `O => F[Unit]`, but much faster.
+    * Like `observe` but observes with a function `O => F[A]` instead of a pipe.
```
mpilquist: Change reference to `F[_]` here and below

.sbtopts
```diff
@@ -1,3 +1 @@
--J-Xms2g
```
mpilquist: Undo?

AdrianRaFo: Yes, sorry

AdrianRaFo:
Looks like I have some dependency resolution error, could someone retry the build?

mpilquist:
Thanks -- the build failure is related to pending 2.13 upgrade.

Issue #1502, opened 2019-06-10
[fs2.io.file] Functionality to `tail -f` a file, i. e. keep reading it while it's being written to


shn-amn:
While working on a log analysis app, I noticed that there are no methods in `fs2.io.file` that allow reading a file that at the same time as it is being written. What I needed was essentially to stream a file as in Linux's `tail -f`.

`file.readAll` as well as `file.readRange` finish the stream as soon as they reaches EOF.

I tweaked the methods `pulls._readAllFromFileHandle0` and `pulls.readAllFromFileHandle` to allow for a `file.keepReading` that does the job. I can make a pull request if the feature can be useful to others.

mpilquist:
That sounds really useful! PR welcome.

shn-amn:
I did that: PR #1506.

For testing, as I commented there, I used Scala's own `Future` and Java's `Thread.sleep` to write asynchronously to the file that is being read. Doesn't seem the best way to do it...

mpilquist:
Fixed by #1506

PR #1506, opened 2019-06-11
file.keepReading to keep reading a file as it's being appended. #1502

shn-amn:
Not so sure of my way of asynchronously writing to the temp file for test is alright...


io/src/main/scala/fs2/io/file/file.scala
```diff
@@ -39,6 +38,21 @@ package object file {
       .flatMap(c => pulls.readRangeFromFileHandle(chunkSize, start, end)(c.resource))
       .stream
 
+  /**
+    * Reads all data synchronously from the file at the specified `java.nio.file.Path`.
+    * Then waits for `delay` duration and reads whatever has been added to the file
+    * in the mean time. Repeats from delay.
+    */
+  def keepReading[F[_]: Sync: ContextShift: Timer](path: Path,
```
Daenyth: What happens if the file is deleted by some other process while we're reading?

shn-amn: As is, stream does not stop, it just emits nothing more. As if nothing new were being written to the file. It is the same behaviour as `file.readAll` and `file.readRange` if the file is deleted before reading is done.

What is the desired behaviour? For the stream to terminate, for example?

shn-amn: Actually, I just looked at `FileHandle`'s implementation, and what I said above is not true.

I will correct the code accordingly.

io/src/main/scala/fs2/io/file/pulls.scala
```diff
@@ -22,6 +23,22 @@ object pulls {
...
+
+  private def _keepReadingFromFileHandle0[F[_]](
+      chunkSize: Int,
+      offset: Long,
+      delay: FiniteDuration)(h: FileHandle[F])(implicit timer: Timer[F]): Pull[F, Byte, Unit] =
+    Pull.eval(h.read(chunkSize, offset)).flatMap {
+      case Some(bytes) =>
+        Pull.output(bytes) >> _keepReadingFromFileHandle0(chunkSize, offset + bytes.size, delay)(h)
+      case None =>
+        timer.sleep(delay)
```
mpilquist: Note this `timer.sleep(delay)` isn't being used -- change to `Pull.eval_(timer.sleep(delay)) >> _keepReadingFromFileHandle0(...)`

io/src/main/scala/fs2/io/file/pulls.scala
```diff
@@ -22,6 +23,22 @@ object pulls {
       h: FileHandle[F]): Pull[F, Byte, Unit] =
     _readRangeFromFileHandle0(chunkSize, start, end)(h)
 
+  def keepReadingFromFileHandle[F[_]: Timer](chunkSize: Int, delay: FiniteDuration)(
```
mpilquist: Name wise, how about `tailFileHandle`? Also, I think we should have a convenience function for this in `file` package object like `fs2.io.file.tail(...)`

io/src/test/scala/fs2/io/file/BaseFileSpec.scala
```diff
@@ -23,6 +24,9 @@ class BaseFileSpec extends Fs2Spec {
   protected def modify(file: Path): Stream[IO, Unit] =
     Stream.eval(IO(Files.write(file, Array[Byte](0, 1, 2, 3))).void)
 
+  protected def modifyLater(file: Path): Stream[IO, Unit] =
+    Stream.eval(IO(Future(List[Byte](0, 1, 2, 3).foreach(waitAndAppend(file)))).void)
```
mpilquist: Even in tests, we should avoid side effects. Instead of writing this to use `Future` and `foreach`, let's write it as a stream of time delayed chunks that writes to a file. Something like this:

```scala
val s = Stream.range(0, 50).map(i => Chunk.array(Array.fill[Byte](1024)(i.toByte))).covary[IO]
s.chunks.metered(25.millis).to(file.writeAll(...)
'''

File: io/src/main/scala/fs2/io/file/pulls.scala (hunk 2 above)
Comment (shn-amn) [reply io/src/main/scala/fs2/io/file/pulls.scala]: Thanks! Done.

File: io/src/test/scala/fs2/io/file/BaseFileSpec.scala (hunk 4 above)
Comment (shn-amn) [reply io/src/test/scala/fs2/io/file/BaseFileSpec.scala]: Thanks for the feedback. Tell me if the new version is alright. It changed a bit from the old version, the two streams need to run concurrently...

File: io/src/main/scala/fs2/io/file/pulls.scala (hunk 3 above)
Comment (shn-amn) [reply io/src/main/scala/fs2/io/file/pulls.scala]: Right. So, 
- Do you think `file.tail(...)` should ignore preexisting content of the file at the time it's run?
- Should there be a parameter saying how far _before_ the end of file the output should start, as in `tail -50f ...` (but probably in bytes rather than lines)?
- Is there any chance starting to read a text file at some arbitrary position would mess up UTF decoding?
(I didn't use the name `tail`, because, as is, the function reads the file from the beginning and only then tails it.)

File: io/src/test/scala/fs2/io/file/BaseFileSpec.scala [hunk 5]
```diff
@@ -25,7 +26,12 @@ class BaseFileSpec extends Fs2Spec {
     Stream.eval(IO(Files.write(file, Array[Byte](0, 1, 2, 3))).void)
 
   protected def modifyLater(file: Path): Stream[IO, Unit] =
-    Stream.eval(IO(Future(List[Byte](0, 1, 2, 3).foreach(waitAndAppend(file)))).void)
+    Stream
+      .range(0, 4)
+      .map(_.toByte)
+      .covary[IO]
+      .metered(250.millis)
+      .evalMap(b => IO(Files.write(file, Array(b), StandardOpenOption.APPEND)).void)
```
mpilquist: Instead of using `Files.write` here, can we use `fs2.io.file.writeAll`?

io/src/test/scala/fs2/io/file/FileSpec.scala
```diff
@@ -63,9 +63,14 @@ class FileSpec extends BaseFileSpec {
         .resource(blockingExecutionContext)
         .flatMap { bec =>
           tempFile
-            .flatTap(modifyLater)
-            .flatMap(path => file.keepReading[IO](path, bec, 4096, 250.millis))
+            .flatMap { path =>
+              file
+                .keepReading[IO](path, bec, 4096, 25.millis)
```
mpilquist: Do this instead: `file.keepReading(...).concurrently(modifyLater(path))` -- then you don't need the `Left`/`Right` mapping or the `collect`.

io/src/main/scala/fs2/io/file/pulls.scala
mpilquist: Great questions.

```scala
def tail[F[_]: Sync: ContextShift: Timer](path: Path, blocker: Blocker, chunkSize: Int, offset: Long = 0L, pollDelay: FiniteDuration = 1.second): Stream[F, Byte] = ???
```

The `offset` param is an offset from the *start* of the file, not the end.

Thoughts?

io/src/test/scala/fs2/io/file/FileSpec.scala
shn-amn: Great! I didn't know about `concurrently`...

By the way, is there any documentation, apart from in the source code, where all these methods are explained?

io/src/test/scala/fs2/io/file/BaseFileSpec.scala
shn-amn: Well, since `fs2.io.file.writeAll` is tested in the same file, I don't feel confortable using it here.

Otherwise, if ever `file.writeAll` breaks, even if `file.keepReading` is alright, neither test will pass...

What do you think?

mpilquist: I think that's okay and should be protected by unit tests for `writeAll`.

djspiewak:
Echoing a desire for `file.tail` rather than `file.keepReading`, though I do see merits to the latter name given that the former has a different meaning in conventional functional programming.

io/src/main/scala/fs2/io/file/pulls.scala
shn-amn: My concern was that since UTF-8 is variable width, starting at an arbitrary position, which could be in the middle of a rune, could maybe mess up UTF decoding. However, I looked up UTF-8 and since the prefixes of first and subsequent bytes in multi-byte encoded characters are different, there won't be any ambiguity for decoders. So, it's fine.

io/src/main/scala/fs2/io/file/file.scala
shn-amn: Okay, so... 

`FileHandle`s are based on `java.nio.channels.FileChannel`. `FileChannel`s seem to outlive their underlying files, in the sense that when the file is deleted, it's `FileChannel` can stay open and I couldn't find any way of asking the `FileChannel` wether the file is still there.

So, unless `FileHandle` is essentially modified, there seems to be no way of knowing that the file is deleted once the handle is created.

Any ideas for a solution?

io/src/test/scala/fs2/io/file/BaseFileSpec.scala
shn-amn: Okay... So I moved the test _after_ that of of `writeAll`.

io/src/main/scala/fs2/io/file/pulls.scala
shn-amn: Done.

io/src/main/scala/fs2/io/file/file.scala
```diff
@@ -36,6 +36,23 @@ package object file {
       .flatMap(c => pulls.readRangeFromFileHandle(chunkSize, start, end)(c.resource))
       .stream
 
+  /**
+    * Reads all data synchronously from the file at the specified `java.nio.file.Path`,
```
mpilquist: How about this description?

```
Returns an infinite stream of data from the file at the specified path. Starts reading from the specified offset and upon reaching the end of the file, polls every `pollDuration` for additional updates to the file.

Read operations are limited to emitting chunks of the specified chunk size but smaller chunks may occur.

If an error occurs while reading from the file, the overall stream fails.
```

mpilquist:
On second thought, I'll update the ScalaDoc on the one method after merging. Thanks for your work on this!

PR #1513, opened 2019-06-15
Algebra: replace the R ADT with a CompileCont trait

diesalbla:
The `compileLoop` method has a system of nested executions, which correspond to the Step algebra operations, in which an inner stream is executed with a different CompileScope.

This was implemented with an inner ADT, `R` with three cases: 1) the compilation on a certain Scope is _done_, 2) the compilation is stopped with some output values, and the rest of the stream; or 3) the compilation was interrupted. This ADT is only used as a way of messaging from one part of the method to another: objects created in one part are deconstructed in another part. This meant that several objects would be allocated without need.

This commit replaces the use of that ADT by a trait called `CompileCont` whose methods correspond to the three cases of that ADT. The `CompileCont` acts as a continuation: it indicates how the
compilation of the stream should proceed when and end-point is reached. There are two (sub)-types of `CompileCont`: 
- The first one, `LoopEnd`, is used for returning from the `compileLoop` method.
- The second one is used when dealing with a `Step` instruction.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -177,8 +177,14 @@ private[fs2] object Algebra {
         F.pure(init)
     }
 
+  private[this] trait CompileCont[F[_], X, Y] {
+    final type Out = Y
+    def done(scope: CompileScope[F]): F[Y]
+    def out(head: Chunk[X], scope: CompileScope[F], tail: FreeC[Algebra[F, X, ?], Unit]): F[Y]
+    def interrupted(scopeId: Token, err: Option[Throwable]): F[Y]
+  }
+
   /*
-   *
    * Interruption of the stream is tightly coupled between FreeC, Algebra and CompileScope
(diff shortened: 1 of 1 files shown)


diesalbla:
@mpilquist Since this is removing some of the classes in the compilation, could this also help in the  work of the "tagless" approach?

mpilquist:
@diesalbla Did you see the test failures on this one?

diesalbla:
@mpilquist I have been fixing the errors in the previous commits in this Pull Request. I have restored it back to a first version of it. The difference with the ones that was failing is that, in those, the `go` inner  method was directly exposed (privately) at the top scope of the `Algebra` object, and then the `compile` method itself was creating an instance of the `CompileCont` trait that would perform the actions in the `compile` method.  Unfortunately, this did not work ok with a series of tests involving a combination of  `bracket` and `handleError`. 

Currently, the only failing test only affects the `2.13` version of Scala, and it involves the message of `java.lang.NoClassDefFoundError: scala/reflect/internal/Trees`.

PR #1522, opened 2019-06-23
Avoid using `cats.implicits` syntax in internals.

diesalbla:
There is a stated preference to avoid using syntax extensions in the internal code. For consistency, this commit applies that to the `CompileScope` code.

Because the modified code avoids using the methods of `Either`, it saves some object allocations.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/CompileScope.scala
+++ b/core/shared/src/main/scala/fs2/internal/CompileScope.scala
@@ -1,12 +1,10 @@
 package fs2.internal
 
 import scala.annotation.tailrec
-
 import cats.{Traverse, TraverseFilter}
 import cats.data.Chain
 import cats.effect.{Concurrent, ExitCase, Sync}
 import cats.effect.concurrent.{Deferred, Ref}
-import cats.implicits._
 import fs2.{CompositeFailure, Pure, Scope}
 import fs2.internal.CompileScope.InterruptContext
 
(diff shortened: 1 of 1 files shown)


djspiewak:
Is this stated in fs2? Or in cats? The primary reason to avoid it in cats is just to cut circular dependencies. There's no particular technical reason to avoid syntax in fs2 or other derivative libraries.

I generally use the syntax implicits pretty aggressively in library construction, mostly because it's far more convenient. The only exception is when I'm in the hot path, where I generally want to avoid abstractions anyway.

diesalbla:
The code in fs2 internals avoids using `cats` syntax extensions, in favour of explicitly referring to the `Monad` or `Sync` instance. For example, [here](https://github.com/functional-streams-for-scala/fs2/blob/series/1.1/core/shared/src/main/scala/fs2/internal/CompileScope.scala#L213-L224), or [here](https://github.com/functional-streams-for-scala/fs2/blob/series/1.1/core/shared/src/main/scala/fs2/internal/Algebra.scala#L305-L306).

djspiewak:
> The code in fs2 internals avoids using cats syntax extensions, in favour of explicitly referring to the Monad or Sync instance.

I wonder if that's intentional though? Or just a convention that people happened to fall into? FS2 used to have its own `Monad` and such, and that hierarchy didn't have syntax extensions, so maybe that's where it comes from?

mpilquist:
I'd like to continue to use syntax for `Either` -- things like `map`, `leftMap`, etc. The core internals don't use type class syntax currently and I think that was partly a historical artifact from when we had our own type classes. However, that turned in to a guideline to avoid syntax in the core interpreter for performance reasons. With that said, I don't believe we'd see a measurable performance impact by wholesale switching to syntax style -- the FS3 branch is written that way and it cleans a lot of stuff up.

diesalbla:
Taking Michael Pilquist comment into account, I am closing this

djspiewak:
@mpilquist Having spent some time recently with C2's assembly output, I can say with some confidence that the syntax imports almost certainly do not have a measurable performance impact on fs2. They would have a measurable performance impact on certain possible codepaths in other projects, so this isn't to take away from the fact that they are a bit of an indirection, but fs2 specifically has vastly higher overhead in all cases where syntax would be used.

PR #1523, opened 2019-06-24
possibly typo fixed

ikr0m:
I'm not sure if those are typo, but I didn't understand their meaning. At least one of them is typo

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1978,7 +1978,7 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
                   // runs inner stream
                   // each stream is forked.
                   // terminates when killSignal is true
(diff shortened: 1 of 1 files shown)


ikr0m:
Initially, I also thought I might be an `enqueue` and tried to grasp the meaning of both variants:
1. `if fails will end in queue failure`
2. 'if fails will enqueue in queue failure'
First one seemed more correct. I've searched `enq` from project if there are other usages but couldn't found. I'll change it to long form as you suggested

SystemFw:
Actually, on rereading the whole comment, I think that line (after `if fails`) is just out of date. It's not using the queue for failures anymore, but a Signal

ikr0m:
Should it be `if fails will enqueue in signal failure`?

SystemFw:
I think it should be something like "failures will be propagated through `done` Signal" or something like that

SystemFw:
Thank you!

ikr0m:
Thank you for your help!

PR #1524, opened 2019-06-24
Revert "Applied workaround for scala/bug 11303 from @joroKr21"

joroKr21:
This reverts commit e0eb895113352ce429560b0b2170f0d0cf4bffd4.

Forgot to revert this after the bug was fixed in Scala 2.13. Do we want to?


mpilquist:
Thanks!

PR #1528, opened 2019-06-28
FreeC.Bind: use method for continuations

diesalbla:
In the FreeC datatype, the Bind constructor represents the case of a `bind` or `flatMap` operation in a free monad constructor: it takes the step `FreeC[F, A]`, and the continuation of what to do with the result. Currently, the implementation represents that continuation using a value of type `Function1`. In Scala there is another way to represent that a record has a function attached: by declaring an abstract method, and using inheritance as a form of parameterisation.

- This commit changes the Bind case class to an abstract class with an abstract `cont` method, which represents that continuation. This may avoid the allocation and indirection of `Function1` instances.

- Because we are using method calls instead of functions, and because a delegated call from a `Bind` object to another means introducing a stack frame in every call, the naive translation of the `Bind` case in `mk` was causing several `StackOverflow`. To solve this problem, we add to the `Bind` class a `delegate` field, to another `Bind` on the same types: when creating a `Bind` that would delegate unto other, we use the source's own delegate. Inductively, this ensures that the chain of delegates from every object would always point either to another non-delegated `Bind`, or to itself.

- We modify the `View` class in line with the changes to `Bind`, so we turn `View` into an abstract class, we turn the `next` field (a `Function1`) into an abstract method, and instead of a `pureContinuation` function, we use a `PureView` subclass, private to the scope of the `ViewL` object.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/FreeC.scala
+++ b/core/shared/src/main/scala/fs2/internal/FreeC.scala
@@ -186,9 +186,9 @@ private[fs2] object FreeC {
     override def toString: String = s"FreeC.Eval($fr)"
   }
 
-  abstract class Bind[F[_], X, R](val step: FreeC[F, X]) extends FreeC[F, R] {
+  abstract class Bind[F[_], X, R](val step: FreeC[F, X]) extends FreeC[F, R] { self =>
     def cont(r: Result[X]): FreeC[F, R]
-
+    val delegate: Bind[F, X, R] = self
     override def toString: String = s"FreeC.Bind($step)"
   }
 
@@ -222,15 +222,15 @@ private[fs2] object FreeC {
             case Result(r) => mk(b.cont(r))
             case Eval(fr)  => ViewL.View(fr, b.cont)
             // case Bind(w, g) => mk(Bind(w, (e: Result[Any]) => Bind(g(e), b.f)))
-            case bb: FreeC.Bind[F, z, y] =>
-              mk {
-                new Bind[F, z, R](bb.step) {
-                  def cont(zr: Result[z]): FreeC[F, R] =
-                    new Bind[F, y, R](bb.cont(zr)) {
-                      def cont(yr: Result[y]): FreeC[F, R] = b.cont(yr)
-                    }
-                }
+            case bb: FreeC.Bind[F, z, _] =>
+              val nb = new Bind[F, z, R](bb.step) {
+                def cont(zr: Result[z]): FreeC[F, R] =
+                  new Bind[F, y, R](bb.cont(zr)) { self =>
+                    override val delegate: Bind[F, y, R] = b.delegate
+                    def cont(yr: Result[y]): FreeC[F, R] = delegate.cont(yr)
+                  }
               }
+              mk(nb)
           }
         case r @ Result.Pure(_)           => r
         case r @ Result.Fail(_)           => r
--- a/core/shared/src/main/scala/fs2/internal/FreeC.scala
+++ b/core/shared/src/main/scala/fs2/internal/FreeC.scala
@@ -55,7 +55,6 @@ private[fs2] sealed abstract class FreeC[F[_], +R] {
           catch { case NonFatal(e) => FreeC.Result.Fail(e) }
         case other => other.asFreeC[F]
       }
-
     }
 
   def asHandler(e: Throwable): FreeC[F, R] = ViewL(this) match {
@@ -63,7 +62,7 @@ private[fs2] sealed abstract class FreeC[F[_], +R] {
     case Result.Fail(e2) => Result.Fail(CompositeFailure(e2, e))
     case Result.Interrupted(ctx, err) =>
       Result.Interrupted(ctx, err.map(t => CompositeFailure(e, t)).orElse(Some(e)))
-    case ViewL.View(_, k) => k(Result.Fail(e))
+    case v @ ViewL.View(_) => v.next(Result.Fail(e))
   }
 
   def viewL[R2 >: R]: ViewL[F, R2] = ViewL(this)
@@ -192,9 +191,6 @@ private[fs2] object FreeC {
     override def toString: String = s"FreeC.Bind($step)"
   }
 
-  def pureContinuation[F[_], R]: Result[R] => FreeC[F, R] =
-    _.asFreeC[F]
-
   def suspend[F[_], R](fr: => FreeC[F, R]): FreeC[F, R] =
     new Bind[F, Unit, R](unit[F]) {
       def cont(r: Result[Unit]): FreeC[F, R] = fr
@@ -208,20 +204,27 @@ private[fs2] object FreeC {
   object ViewL {
 
     /** unrolled view of FreeC `bind` structure **/
-    final case class View[F[_], X, R](step: F[X], next: Result[X] => FreeC[F, R])
-        extends ViewL[F, R]
+    sealed abstract case class View[F[_], X, R](step: F[X]) extends ViewL[F, R] {
+      def next(r: Result[X]): FreeC[F, R]
+    }
+
+    private[ViewL] final class EvalView[F[_], R](step: F[R]) extends View[F, R, R](step) {
+      def next(r: Result[R]): FreeC[F, R] = r.asFreeC[F]
+    }
 
     private[fs2] def apply[F[_], R](free: FreeC[F, R]): ViewL[F, R] = mk(free)
 
     @tailrec
     private def mk[F[_], R](free: FreeC[F, R]): ViewL[F, R] =
       free match {
-        case Eval(fx) => View(fx, pureContinuation[F, R])
+        case e: Eval[F, R] => new EvalView[F, R](e.fr)
         case b: FreeC.Bind[F, y, R] =>
           b.step match {
             case Result(r) => mk(b.cont(r))
-            case Eval(fr)  => ViewL.View(fr, b.cont)
-            // case Bind(w, g) => mk(Bind(w, (e: Result[Any]) => Bind(g(e), b.f)))
+            case Eval(fr) =>
+              new ViewL.View[F, y, R](fr) {
+                def next(r: Result[y]): FreeC[F, R] = b.cont(r)
+              }
             case bb: FreeC.Bind[F, z, _] =>
               val nb = new Bind[F, z, R](bb.step) {
                 def cont(zr: Result[z]): FreeC[F, R] =
@@ -246,9 +249,9 @@ private[fs2] object FreeC {
         case Result.Pure(r)             => F.pure(Some(r))
         case Result.Fail(e)             => F.raiseError(e)
         case Result.Interrupted(_, err) => err.fold[F[Option[R]]](F.pure(None)) { F.raiseError }
-        case ViewL.View(step, next) =>
+        case v @ ViewL.View(step) =>
           F.flatMap(F.attempt(step)) { r =>
-            next(Result.fromEither(r)).run
+            v.next(Result.fromEither(r)).run
           }
       }
   }
--- a/core/shared/src/main/scala/fs2/internal/FreeC.scala
+++ b/core/shared/src/main/scala/fs2/internal/FreeC.scala
@@ -185,9 +185,9 @@ private[fs2] object FreeC {
     override def toString: String = s"FreeC.Eval($fr)"
   }
 
-  abstract class Bind[F[_], X, R](val step: FreeC[F, X]) extends FreeC[F, R] { self =>
+  abstract class Bind[F[_], X, R](val step: FreeC[F, X]) extends FreeC[F, R] {
     def cont(r: Result[X]): FreeC[F, R]
-    val delegate: Bind[F, X, R] = self
+    def delegate: Bind[F, X, R] = this
     override def toString: String = s"FreeC.Bind($step)"
   }
 
@@ -227,9 +227,10 @@ private[fs2] object FreeC {
               }
             case bb: FreeC.Bind[F, z, _] =>
               val nb = new Bind[F, z, R](bb.step) {
+                private[this] val bdel = b.delegate
                 def cont(zr: Result[z]): FreeC[F, R] =
-                  new Bind[F, y, R](bb.cont(zr)) { self =>
-                    override val delegate: Bind[F, y, R] = b.delegate
+                  new Bind[F, y, R](bb.cont(zr)) {
+                    override val delegate: Bind[F, y, R] = bdel
                     def cont(yr: Result[y]): FreeC[F, R] = delegate.cont(yr)
                   }
               }
--- a/core/shared/src/main/scala/fs2/internal/FreeC.scala
+++ b/core/shared/src/main/scala/fs2/internal/FreeC.scala
@@ -25,36 +25,38 @@ import scala.util.control.NonFatal
 private[fs2] sealed abstract class FreeC[F[_], +R] {
 
   def flatMap[R2](f: R => FreeC[F, R2]): FreeC[F, R2] =
-    Bind[F, R, R2](
-      this,
-      e =>
-        e match {
-          case Result.Pure(r) =>
-            try f(r)
-            catch { case NonFatal(e) => FreeC.Result.Fail(e) }
-          case Result.Interrupted(scope, err) => FreeC.Result.Interrupted(scope, err)
-          case Result.Fail(e)                 => FreeC.Result.Fail(e)
+    new Bind[F, R, R2](this) {
+      def cont(e: Result[R]): FreeC[F, R2] = e match {
+        case Result.Pure(r) =>
+          try f(r)
+          catch { case NonFatal(e) => FreeC.Result.Fail(e) }
+        case Result.Interrupted(scope, err) => FreeC.Result.Interrupted(scope, err)
+        case Result.Fail(e)                 => FreeC.Result.Fail(e)
       }
-    )
+    }
 
   def transformWith[R2](f: Result[R] => FreeC[F, R2]): FreeC[F, R2] =
-    Bind[F, R, R2](this,
-                   r =>
-                     try f(r)
-                     catch { case NonFatal(e) => FreeC.Result.Fail(e) })
+    new Bind[F, R, R2](this) {
+      def cont(r: Result[R]): FreeC[F, R2] =
+        try f(r)
+        catch { case NonFatal(e) => FreeC.Result.Fail(e) }
+    }
 
   def map[R2](f: R => R2): FreeC[F, R2] =
-    Bind[F, R, R2](this, Result.map(_)(f).asFreeC[F])
+    new Bind[F, R, R2](this) {
+      def cont(e: Result[R]): FreeC[F, R2] = Result.map(e)(f).asFreeC[F]
+    }
 
   def handleErrorWith[R2 >: R](h: Throwable => FreeC[F, R2]): FreeC[F, R2] =
-    Bind[F, R2, R2](this,
-                    e =>
-                      e match {
-                        case Result.Fail(e) =>
-                          try h(e)
-                          catch { case NonFatal(e) => FreeC.Result.Fail(e) }
-                        case other => other.asFreeC[F]
-                    })
+    new Bind[F, R2, R2](this) {
+      def cont(e: Result[R2]): FreeC[F, R2] = e match {
+        case Result.Fail(e) =>
+          try h(e)
+          catch { case NonFatal(e) => FreeC.Result.Fail(e) }
+        case other => other.asFreeC[F]
+      }
+
+    }
 
   def asHandler(e: Throwable): FreeC[F, R] = ViewL(this) match {
     case Result.Pure(_)  => Result.Fail(e)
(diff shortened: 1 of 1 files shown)


pchlupacek:
@diesalbla do you have any proves that this saves allocations? Any reasonable measurement ? Thanks

diesalbla:
@pchlupacek Reading at the code: at every point at which a `Bind` object is being created (methods `flatMap`, `transformWith`, `map`, `handleErrorWith`, `suspend`, and the `ViewL.View` case of `translate`), instead of creating an anonymous function (as a lambda expression) and the `Bind` case class, the modified code only creates the `Bind` object and writes the logic of the code in the `cont` method. Therefore, in each of these methods you need one less allocation. In the same way, when a `View` is created from a plain `Eval`, it no longer need to create a `pureContinuation`, and just uses the class instead. In the other case, when creating a `View`, while the previous code was not _creating_ a new function and reusing the one from a `Bind` instead, the new code is just doing the same reuse through method-to-method calls. The use of delegates keep that chain short. 

Since these optimisations sit on all of the API of `FreeC` and therefore in the path of  every `Stream` or `Pull` operation, as well as in the view-unrolling loop, this is very likely to reduce allocations in any fs2 program. 

This is not an unusual way of reasoning about performance. For example,  have found it used (as a secondary point) in a previous PR https://github.com/functional-streams-for-scala/fs2/pull/1204. As for experiments, I am afraid I have not found much howto information on the fs2-benchmarks project, nor in a few previous pull requests I have looked into. I wonder: could you perhaps ellaborate as to how do you usually run those benchmarks?

pchlupacek:
I think I understood the concept, however I also know that often certain performance optimisations fail to deliver improvement, and sometimes backfire in other way. 

I think we shall not enhance/improve the code where we can only be theoretic about the impacts of the change. If the change couldn't be measured to have theoretical impact, it is hard as well to prove that it may not harm the performance in other way. 

I am not expert in measurement of performance, but yes if we say there will be like half of allocation like this seems to promise, the standard java profiler shall prove this. Also we shall be able to see performance of have binds/viewL cycles  to be reduced. It may not be large number, but if we just be better by measurable notch, there is good reason to make the change.

djspiewak:
I agree with @diesalbla that this is likely to be a very noticeable win. At the very least, there's effectively zero loss. We should be careful when profiling though as certain optimizations surrounding `invokedynamic` will be turned off in those modes. Flight Recorder should give us some object allocation and heap pressure metrics without shifting optimization modes, so we should start there.

I suspect even preliminary measurements will show a relatively dramatic shift.

pchlupacek:
@djspiewak ok lets run and collect some numbers with FlightRecorder. Just to doublecheck. I think if we just run few programs operating heavy on FreeC, we shall see this noticeably imho.

diesalbla:
@djspiewak @pchlupacek I am looking into how to run these benchmarks, and will get back to you as soon as I have the results.

djspiewak:
@pchlupacek Yeah I think even the wall clock differences should be pretty noticeable.

@diesalbla JFR isn't necessarily the hardest thing to work with, though the documentation is deceptive. It's worth noting that it hasn't been backported into openjdk 8, so you'll have to use jdk 11 more than likely.

diesalbla:
I have run the `FreeCBenchmarks` using a configuration of 13 iterations and 13 warmup iterations, with both the `series 1.0` and this branch: 

##### series/1.0
```Bash
Series 1.0

`jmh:run -i 13 -wi 13 -f1 -t1 .*FreeCBenchmarks*. -prof gc -jvmArgs -XX:MaxInlineLevel=18`

[info] Benchmark                                                        Mode  Cnt          Score         Error   Units
[info] FreeCBenchmark.nestedFlatMaps                                   thrpt   13         22.150 ±       0.149   ops/s
[info] FreeCBenchmark.nestedFlatMaps:·gc.alloc.rate                    thrpt   13       3862.707 ±      26.063  MB/sec
[info] FreeCBenchmark.nestedFlatMaps:·gc.alloc.rate.norm               thrpt   13  191997933.665 ±      16.007    B/op
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Eden_Space           thrpt   13       3867.819 ±     131.778  MB/sec
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Eden_Space.norm      thrpt   13  192245011.020 ± 6087279.526    B/op
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Old_Gen              thrpt   13          0.402 ±       0.020  MB/sec
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Old_Gen.norm         thrpt   13      19959.181 ±     953.470    B/op
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Survivor_Space       thrpt   13          0.555 ±       0.063  MB/sec
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Survivor_Space.norm  thrpt   13      27601.040 ±    3092.826    B/op
[info] FreeCBenchmark.nestedFlatMaps:·gc.count                         thrpt   13        222.000                counts
[info] FreeCBenchmark.nestedFlatMaps:·gc.time                          thrpt   13       1216.000                    ms

[info] FreeCBenchmark.nestedMaps                                       thrpt   13         22.255 ±       0.202   ops/s
[info] FreeCBenchmark.nestedMaps:·gc.alloc.rate                        thrpt   13       3880.829 ±      35.501  MB/sec
[info] FreeCBenchmark.nestedMaps:·gc.alloc.rate.norm                   thrpt   13  191997933.624 ±      15.916    B/op
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Eden_Space               thrpt   13       3887.336 ±      99.289  MB/sec
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Eden_Space.norm          thrpt   13  192318150.433 ± 4442236.601    B/op
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Old_Gen                  thrpt   13          0.401 ±       0.031  MB/sec
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Old_Gen.norm             thrpt   13      19853.853 ±    1609.627    B/op
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Survivor_Space           thrpt   13          0.556 ±       0.063  MB/sec
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Survivor_Space.norm      thrpt   13      27483.220 ±    3016.077    B/op
[info] FreeCBenchmark.nestedMaps:·gc.count                             thrpt   13        223.000                counts
[info] FreeCBenchmark.nestedMaps:·gc.time                              thrpt   13       1945.000                    ms
[success] Total time: 556 s, completed 18 Aug 2019, 17:33:18
```

##### PR Branch

```Bash
[info] Benchmark                                                        Mode  Cnt          Score         Error   Units
[info] FreeCBenchmark.nestedFlatMaps                                   thrpt   13          4.448 ±       0.322   ops/s
[info] FreeCBenchmark.nestedFlatMaps:·gc.alloc.rate                    thrpt   13        646.810 ±      46.812  MB/sec
[info] FreeCBenchmark.nestedFlatMaps:·gc.alloc.rate.norm               thrpt   13  159997938.870 ±      74.177    B/op
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Eden_Space           thrpt   13        648.058 ±      46.774  MB/sec
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Eden_Space.norm      thrpt   13  160315194.546 ± 1989262.738    B/op
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Old_Gen              thrpt   13          0.015 ±       0.035  MB/sec
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Old_Gen.norm         thrpt   13       3797.986 ±    8995.128    B/op
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Survivor_Space       thrpt   13          8.517 ±       1.984  MB/sec
[info] FreeCBenchmark.nestedFlatMaps:·gc.churn.G1_Survivor_Space.norm  thrpt   13    2098804.864 ±  427678.645    B/op
[info] FreeCBenchmark.nestedFlatMaps:·gc.count                         thrpt   13        536.000                counts
[info] FreeCBenchmark.nestedFlatMaps:·gc.time                          thrpt   13     108044.000                    ms

[info] FreeCBenchmark.nestedMaps                                       thrpt   13          4.422 ±       0.234   ops/s
[info] FreeCBenchmark.nestedMaps:·gc.alloc.rate                        thrpt   13        643.026 ±      34.107  MB/sec
[info] FreeCBenchmark.nestedMaps:·gc.alloc.rate.norm                   thrpt   13  159997938.841 ±      74.063    B/op
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Eden_Space               thrpt   13        643.999 ±      34.647  MB/sec
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Eden_Space.norm          thrpt   13  160238408.375 ± 1432304.351    B/op
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Old_Gen                  thrpt   13          0.005 ±       0.013  MB/sec
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Old_Gen.norm             thrpt   13       1400.377 ±    3310.890    B/op
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Survivor_Space           thrpt   13          8.088 ±       1.889  MB/sec
[info] FreeCBenchmark.nestedMaps:·gc.churn.G1_Survivor_Space.norm      thrpt   13    2015147.480 ±  483798.264    B/op
[info] FreeCBenchmark.nestedMaps:·gc.count                             thrpt   13        531.000                counts
[info] FreeCBenchmark.nestedMaps:·gc.time                              thrpt   13     107413.000                    ms
```

djspiewak:
Holy shit…

diesalbla:
I am trying to figure out why the difference on the first measure (ops/s) since I suspect (not a benchmark expert) that it means that the new one is producing less garbage only because it has less throughput. 

I must admit I just used some parameters that Jason Zaugg was using [here](https://github.com/scala/scala/pull/8335#issue-307558823), but I am not sure if the `-jvmArgs -XX:MaxInlineLevel=18`, where that value indicates "_maximum number of nested calls that are inlined_", would be necessary here. I was intending to re-run without it (takes 20 minutes on my laptop).

djspiewak:
> I am trying to figure out why the difference on the first measure (ops/s) since I suspect (not a benchmark expert) that it means that the new one is producing less garbage only because it has less throughput.

Yeah I initially read the benchmarks backwards (ms/op instead of ops/sec) and so I got really excited. It is indeed 5 times slower. I don't really know why that would be though.

pchlupacek:
@diesalbla this is interesting result.

What you can do is to deep into profile and see what takes most of the time in either scenario. I think top 10s entries will be different and can give us the hint. 

Seems however that ration between allocation and throughput is similar in both cases.

diesalbla:
Not quite similar...

```
[info] FreeCBenchmark.nestedFlatMaps                                   thrpt   13         22.150 ±       0.149   ops/s
[info] FreeCBenchmark.nestedFlatMaps:·gc.alloc.rate                    thrpt   13       3862.707 ±      26.063  MB/sec
[info] FreeCBenchmark.nestedFlatMaps                                   thrpt   13         22.150 ±       0.149   ops/s
[info] FreeCBenchmark.nestedFlatMaps:·gc.alloc.rate                    thrpt   13       3862.707 ±      26.063  MB/sec

[info] FreeCBenchmark.nestedFlatMaps                                   thrpt   13          4.448 ±       0.322   ops/s
[info] FreeCBenchmark.nestedFlatMaps:·gc.alloc.rate                    thrpt   13        646.810 ±      46.812  MB/sec
[info] FreeCBenchmark.nestedMaps                                       thrpt   13          4.422 ±       0.234   ops/s
[info] FreeCBenchmark.nestedMaps:·gc.alloc.rate                        thrpt   13        643.026 ±      34.107  MB/sec
```
Comparing the rations of allocations per ops, which is to say `(646.81 / 4.448) / (3862.707 / 22.15) * 100`, gives you about "83.4", which, if I understand correctly (and have not made a mistake with numbers), that means that the branch takes only 83% of the allocations per "op" (execution of the benchmark). The result is similar to the `nestedMaps`.

Still, it may be a bit too obscure to be sure.

core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -200,20 +204,36 @@ private[fs2] object FreeC {
...
+            case Result(r) => mk(b.cont(r))
+            case Eval(fr) =>
+              new ViewL.View[F, y, R](fr) {
+                def next(r: Result[y]): FreeC[F, R] = b.cont(r)
+              }
+            case bb: FreeC.Bind[F, z, _] =>
+              val nb = new Bind[F, z, R](bb.step) {
+                def cont(zr: Result[z]): FreeC[F, R] =
+                  new Bind[F, y, R](bb.cont(zr)) { self =>
+                    override val delegate: Bind[F, y, R] = b.delegate
```
diesalbla: @djspiewak @pchlupacek  I am guessing that this was the cause for the strange behaviour that I was seeing in the benchmarks. 

Because this `b.delegate` is inside a `def cont`, this line was creating a back-reference to the `b` object, which prevented the garbage collector from cleaning up that `b` object, and which may have forced the evaluation of this line to reload an object. 

That would explain that, when running those benchmarks on the PR branch, we were seeing less allocations per _op_ (memory efficient); but there were also less throughput (ops per second), and more calls to the garbage collector (and more time in GC), because the GC could collect fewer objects in each run. 

I am pushing a commit with the correction to this, and the results of rerunning the benchmarks.

diesalbla:
These are the benchmarks with the corrected branch. I have run them using the following command: 
`jmh:run -i 13 -wi 13 -f1 -t1 .*FreeCBenchmarks*. -prof gc`

#### series/1.0

```
Benchmark                                             Mode  Cnt          Score         Error   Units

nestedFlatMaps                                        thrpt   13          2.364 ±       0.031   ops/s
nestedFlatMaps:·gc.alloc.rate                         thrpt   13        412.427 ±       5.112  MB/sec
nestedFlatMaps:·gc.alloc.rate.norm                    thrpt   13  191997980.538 ±     148.653    B/op
nestedFlatMaps:·gc.churn.Compressed_Class_Space       thrpt   13         ≈ 10⁻⁴                MB/sec
nestedFlatMaps:·gc.churn.Compressed_Class_Space.norm  thrpt   13         46.667 ±       0.001    B/op
nestedFlatMaps:·gc.churn.G1_Eden_Space                thrpt   13        413.847 ±       5.105  MB/sec
nestedFlatMaps:·gc.churn.G1_Eden_Space.norm           thrpt   13  192659035.897 ±   50268.692    B/op
nestedFlatMaps:·gc.churn.G1_Old_Gen                   thrpt   13        199.445 ±       2.361  MB/sec
nestedFlatMaps:·gc.churn.G1_Old_Gen.norm              thrpt   13   92848589.256 ±  153635.719    B/op
nestedFlatMaps:·gc.churn.G1_Survivor_Space            thrpt   13         13.500 ±       0.175  MB/sec
nestedFlatMaps:·gc.churn.G1_Survivor_Space.norm       thrpt   13    6284734.359 ±   19648.422    B/op
nestedFlatMaps:·gc.churn.Metaspace                    thrpt   13         ≈ 10⁻³                MB/sec
nestedFlatMaps:·gc.churn.Metaspace.norm               thrpt   13        169.333 ±       0.001    B/op
nestedFlatMaps:·gc.count                              thrpt   13       2496.000                counts
nestedFlatMaps:·gc.time                               thrpt   13     111953.000                    ms

nestedMaps                                            thrpt   13          2.335 ±       0.072   ops/s
nestedMaps:·gc.alloc.rate                             thrpt   13        407.181 ±      12.467  MB/sec
nestedMaps:·gc.alloc.rate.norm                        thrpt   13  191997980.569 ±     148.402    B/op
nestedMaps:·gc.churn.Compressed_Class_Space           thrpt   13         ≈ 10⁻⁴                MB/sec
nestedMaps:·gc.churn.Compressed_Class_Space.norm      thrpt   13         47.147 ±       1.423    B/op
nestedMaps:·gc.churn.G1_Eden_Space                    thrpt   13        408.601 ±      13.088  MB/sec
nestedMaps:·gc.churn.G1_Eden_Space.norm               thrpt   13  192663226.694 ±  596981.711    B/op
nestedMaps:·gc.churn.G1_Old_Gen                       thrpt   13        197.261 ±       5.084  MB/sec
nestedMaps:·gc.churn.G1_Old_Gen.norm                  thrpt   13   93034874.267 ± 1871479.818    B/op
nestedMaps:·gc.churn.G1_Survivor_Space                thrpt   13         13.773 ±       0.425  MB/sec
nestedMaps:·gc.churn.G1_Survivor_Space.norm           thrpt   13    6500434.742 ±  348929.632    B/op
nestedMaps:·gc.churn.Metaspace                        thrpt   13         ≈ 10⁻³                MB/sec
nestedMaps:·gc.churn.Metaspace.norm                   thrpt   13        171.078 ±       5.164    B/op
nestedMaps:·gc.count                                  thrpt   13       2514.000                counts
nestedMaps:·gc.time                                   thrpt   13     112893.000                    ms
```

#### This Branch

```

nestedFlatMaps                                        thrpt   13          2.853 ±        0.134   ops/s
nestedFlatMaps:·gc.alloc.rate                         thrpt   13        394.087 ±       18.426  MB/sec
nestedFlatMaps:·gc.alloc.rate.norm                    thrpt   13  151997954.336 ±      119.197    B/op
nestedFlatMaps:·gc.churn.Compressed_Class_Space       thrpt   13         ≈ 10⁻⁵                 MB/sec
nestedFlatMaps:·gc.churn.Compressed_Class_Space.norm  thrpt   13          2.971 ±       12.827    B/op
nestedFlatMaps:·gc.churn.G1_Eden_Space                thrpt   13        395.557 ±       18.343  MB/sec
nestedFlatMaps:·gc.churn.G1_Eden_Space.norm           thrpt   13  152567838.904 ±   619569.231    B/op
nestedFlatMaps:·gc.churn.G1_Old_Gen                   thrpt   13        125.445 ±       37.692  MB/sec
nestedFlatMaps:·gc.churn.G1_Old_Gen.norm              thrpt   13   48199963.004 ± 13821799.771    B/op
nestedFlatMaps:·gc.churn.G1_Survivor_Space            thrpt   13         15.871 ±        1.547  MB/sec
nestedFlatMaps:·gc.churn.G1_Survivor_Space.norm       thrpt   13    6136677.983 ±   748909.664    B/op
nestedFlatMaps:·gc.churn.Metaspace                    thrpt   13         ≈ 10⁻⁵                 MB/sec
nestedFlatMaps:·gc.churn.Metaspace.norm               thrpt   13         10.780 ±       46.545    B/op
nestedFlatMaps:·gc.count                              thrpt   13       2095.000                 counts
nestedFlatMaps:·gc.time                               thrpt   13     113441.000                     ms

nestedMaps                                            thrpt   13          2.903 ±        0.055   ops/s
nestedMaps:·gc.alloc.rate                             thrpt   13        400.880 ±        7.701  MB/sec
nestedMaps:·gc.alloc.rate.norm                        thrpt   13  151997954.259 ±      118.647    B/op
nestedMaps:·gc.churn.Compressed_Class_Space           thrpt   13         ≈ 10⁻⁵                 MB/sec
nestedMaps:·gc.churn.Compressed_Class_Space.norm      thrpt   13          2.872 ±       12.400    B/op
nestedMaps:·gc.churn.G1_Eden_Space                    thrpt   13        402.262 ±        8.257  MB/sec
nestedMaps:·gc.churn.G1_Eden_Space.norm               thrpt   13  152520339.308 ±   660049.443    B/op
nestedMaps:·gc.churn.G1_Old_Gen                       thrpt   13        128.005 ±       27.614  MB/sec
nestedMaps:·gc.churn.G1_Old_Gen.norm                  thrpt   13   48515519.547 ± 10266998.636    B/op
nestedMaps:·gc.churn.G1_Survivor_Space                thrpt   13         16.276 ±        1.268  MB/sec
nestedMaps:·gc.churn.G1_Survivor_Space.norm           thrpt   13    6172691.556 ±   500909.760    B/op
nestedMaps:·gc.churn.Metaspace                        thrpt   13         ≈ 10⁻⁵                 MB/sec
nestedMaps:·gc.churn.Metaspace.norm                   thrpt   13         10.421 ±       44.994    B/op
nestedMaps:·gc.count                                  thrpt   13       2145.000                 counts
nestedMaps:·gc.time                                   thrpt   13     112667.000                     ms

```

_Note_ I have removed the inlining option becuase, when I first tried to run it (on my laptop), because not only the JVM but also the CPU was quite warm, I was seeing huge variability in throughput.

djspiewak:
> Throughput improved by 20% (nestedFlatMap) and 25% (nestedMap). Allocation rate per operation reduced to 74% of the previous rates, and number gc.count reduced to 83% or 85% of of previous level.

Noooooiiiiiiiccceeee.

This is roughly order-of-magnitude in line with what I was expecting from these changes, so confirmation bias is appeased! Excellent sleuthing.

retronym:
`gc.alloc.rate.norm` is the counter you want to pay most attention to in jmh's `-prof gc`. That's the number of bytes allocated per operation.

retronym:
IIRC @tpolecat has some optimization experience with similar patterns of interpreters in Doobie.

retronym:
If you're using `sbt-jmh`, you can also run `jmh:run. ... -prof jmh.extras.JFR` to generate a Flight Recorder profile dump for the benchmark, which you can look at with [JDK Mission Control](https://www.azul.com/products/zulu-mission-control/)

pchlupacek:
@diesalbla good work. Thats great news :-)

diesalbla:
@pchlupacek Thanks. The results of the benchmarks are clear, but perhaps `FreeCBenchmarks` is too focused and synthetic. I was thinking of running another suite. Is there any one that may be more relevant?

pchlupacek:
Not sure. I think we have enough evidence this is non negligible improvement LGTM

PR #1529, opened 2019-06-29
Optimisations to the `viewL` Loop.

diesalbla:
This Commit introduces some optimisations to the View unrolling loop. The View Unrolling loop takes an input `FreeC` and expands-evaluates to remove all Binds on the left hand side, to end up with either a Result, or a pair of an Eval and a continuation.

When the loop finds a `Bind` inside a `Bind`, it 1) creates a new `Bind`, with the head of the inner one and the composition of the continuations `Bind`; and 2) recurs on that `Bind` object, which immediately deconstructed it. So, this created a `Bind` object that it would throw away in the next iteration. To remedy this, we introduce an inner loop (`@tailrec` function) that takes two parameters, the continuations and the heads of the binds. When a `Bind-Bind` case happens, it just recurs on the left spine of the structure.

In the inner loop, the loop could sometimes compose a continuation (a `Bind`-in-`Bind`), and in a next step apply that continuation to an intermediate `Result`, to obtain a `FreeC` that may also be a `Bind`. In which case, it immediately deconstructs that `Bind` object that the composed continuation created. To avoid this, we introduce a special sub-class of Function, called `>=>` (a composition operation for monadic functions). When the loop meets an intermediate Result, it checks to see if the continuation is a  `>=>`, and if it is, it recurs in the inner loop.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/FreeC.scala
+++ b/core/shared/src/main/scala/fs2/internal/FreeC.scala
@@ -203,6 +203,12 @@ private[fs2] object FreeC {
     final case class View[F[_], X, R](step: F[X], next: Result[X] => FreeC[F, R])
         extends ViewL[F, R]
 
+    final case class >=>[F[_], X, Y, Z](fst: Result[X] => FreeC[F, Y],
+                                        snd: Result[Y] => FreeC[F, Z])
+        extends (Result[X] => FreeC[F, Z]) {
+      def apply(r: Result[X]): FreeC[F, Z] = Bind[F, Y, Z](fst(r), snd)
+    }
+
     private[fs2] def apply[F[_], R](free: FreeC[F, R]): ViewL[F, R] = mk(free)
 
     @tailrec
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -203,17 +203,44 @@ private[fs2] object FreeC {
...
-            case Result(r)  => mk(b.f(r))
-            case Eval(fr)   => ViewL.View(fr, b.f)
-            case Bind(w, g) => mk(Bind(w, (e: Result[Any]) => Bind(g(e), b.f)))
+          @tailrec
+          def mk2[Y, Z](free: FreeC[F, Y],
+                        cont: Result[Y] => FreeC[F, Z]): Either[FreeC[F, Z], ViewL[F, Z]] =
+            free match {
+              case b: FreeC.Bind[F, x, Y] => mk2[x, Z](b.fx, >=>(b.f, cont))
+              case Eval(fx)               => Right(View(fx, cont))
+              case r @ Result.Pure(_) =>
```
diesalbla: I did try extracting the `Result` cases in a single function, but then the `@tailrec` annotation would not work.

diesalbla:
It would seem this `@tailrec` does not work ok in Scala `2.11` 😞

diesalbla:
I am closing this one in favour of #1528.

PR #1532, opened 2019-07-04
Add Queue#push1 and Queue#push operators

sloshy:
This is an operator I've implemented to make certain Queue scenarios a little easier and less error-prone to handle. The basic idea is that I wanted a transactional way to "push" an element out of the queue by force, sort of a `getAndSet`-style operation, but for single queue elements. You can implement something sort-of similar using this:

```scala
val q: Queue[IO, Int] = ???
q.offer1(elem).flatMap { offerWorked =>
  if (offerWorked) {
    Applicative[F].pure(None) //Queue is not full, so return nothing
  } else {
    q.dequeue1.flatMap(a => q.enqueue1(elem) >> Applicative[F].pure(Some(a))) //Queue is full, dequeue and enqueue again
  }
}
```

The problem with this technique is that if the queue is shared concurrently then the lack of enqueueing/dequeueing as a transaction could cause determinism issues, where the state of the queue will be changing underneath our feet.

This implementation goes down to the `PubSub` level to do it in a single `update` so that it should work concurrently as well as locally. At the `Queue` level there are two new methods: `push1` which works for individual elements, and `push` which is the streaming version.

If anything looks weird or fishy about the implementation or docs, feel free to point them out. Thanks in advance for looking at this PR!

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/PubSub.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/PubSub.scala
@@ -69,7 +69,18 @@ private[fs2] trait Subscribe[F[_], A, Selector] {
   def unsubscribe(selector: Selector): F[Unit]
 }
 
-private[fs2] trait PubSub[F[_], I, O, Selector] extends Publish[F, I] with Subscribe[F, O, Selector]
+private[fs2] trait PubSub[F[_], I, O, Selector]
+    extends Publish[F, I]
+    with Subscribe[F, O, Selector] {
+
+  /**
+    * Forcibly publishes one element by retrieving the element specified by the selector first, if necessary.
+    *
+    * @param i The element to always publish.
+    * @param s The selector for the element to retrieve if the element to publish cannot be accepted yet.
+    */
+  def push(i: I, s: Selector): F[Option[O]]
+}
 
 private[fs2] object PubSub {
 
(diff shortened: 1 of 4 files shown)


pchlupacek:
@sloshy I am not really sure that I follow to full extend usage of the primitive here. Isn't this something you will traditionally achieve with Ref only? Could you elaborate why queue is needed here? 

Thank you.

sloshy:
@pchlupacek So the reason I am basing this on `Queue` is because I want to have queue-like semantics with regards to flushing the queue. So say I am using this queue as a buffer where N elements are held inside of it, and every extra element that comes in "pushes" one element out, resulting in a stream of all elements except the last N. I want to process these last N elements differently from the rest of them, sort-of like the opposite of `Pull.take`. The following pseudocode should give you a general idea:
```scala
val listIO = Queue.bounded[IO, Int](2).flatMap { q =>
  val streamA = Stream.emits(List(1, 2, 3)).covary[IO]
  (q.push(streamA).map(_ + 3) ++ Stream.repeatEval(q.tryDequeue1).unNoneTerminate).compile.toList
}
listIO.flatMap(list => IO(println(list))).unsafeRunSync // List(4, 2, 3)
```

I think this specific use case could be implemented as a layer on top of queue using a `Ref`-based implementation if that is what you mean. I thought about how I would personally solve the issue first, and then I was thinking that it might be desirable to have an atomic `enqueue1 + dequeue1` operation for general queue use-cases, but if you think this might be something better done in a separate construct (or separate library) that is fine with me as well.

sloshy:
@mpilquist and @pchlupacek Hey I haven't followed up on this one for a while for various reasons, sorry about that. That said, don't bother reopening this one, if you were going to get around to it as mentioned on Gitter. I've had a lack of interest in pursuing this specific version of this idea for now. I also think my initial ideas, while they seemed to work as implemented, might be better expressed in a different way or using a different set of primitives as mentioned. Thanks for taking a look regardless and maybe someday I'll revisit this idea.

PR #1536, opened 2019-07-12
Adding the Partially-Applied trick to Stream.fromIterator

BalmungSan:
Now, when calling `Stream.fromIterator` one only has to specify the effect type. Instead of also having to specify the type of the elements inside the iterator, which IMHO was very pointless.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3124,7 +3124,9 @@ object Stream extends StreamLowPriority {
     now.flatMap(go)
   }
 
-  final class PartiallyAppliedFromEither[F[_]] {
+  private[fs2] final class PartiallyAppliedFromEither[F[_]](
+      private val dummy: Boolean
+  ) extends AnyVal {
     def apply[A](either: Either[Throwable, A])(implicit ev: RaiseThrowable[F]): Stream[F, A] =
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3124,7 +3124,9 @@ object Stream extends StreamLowPriority {
     now.flatMap(go)
   }
 
-  final class PartiallyAppliedFromEither[F[_]] {
+  private[fs2] final class PartiallyAppliedFromEither[F[_]](
+      private val dummy: Boolean
```
mpilquist: What's this boolean for?

mpilquist: Oh, to allow `AnyVal` use?

BalmungSan: Yes, just for that.
I believe a Boolean is the smallest object one can use. Probably the JIT would inline all that.

The pattern is copied as it is from the [**cats guidelines**](https://typelevel.org/cats/guidelines.html) _(with the only difference that they use a default value)_.

mpilquist: Gotcha, thanks! Will wait for feedback from Fabio but this LGTM.

Issue #1533, opened 2019-07-04
Add more operations to fs2-io to make it a fully featured file library


LukaJCB:
We talked about this on gitter at some point and the lack of a fully featured file library in the cats-effect ecosystem is something that should be addressed.

Below are some features I'd like to see added to `fs2-io`, note that a lot of those aren't streaming operations, but I feel like they're still very much relevant here.

- moving files 
- deleting files
- copying files
- create directories
- list files and folders at given path
- check if file or folder exists
- read filesystem metadata
- symlinks
- permissions
- temp files

I can start working on some of these soon if no one objects :)

ybasket:
I like the idea and would be in for contributing some code - did you start working on it already?

LukaJCB:
Yes, I can push a WIP branch :)

ybasket:
Just opened a PR for listing files, complementing your WIP, but there are still operations missing in case someone wants to contribute :)

r-glyde:
Hi. Is there still an interest in adding more of the functionality above and is anyone else working on them? I think some towards the bottom of the list are not done yet. I was going to look at permissions unless anyone else plans to.

r-glyde:
Hi @LukaJCB, just wondering what sort of implementation you had in mind for the temporary files (and I guess directories as well)? I was initially assuming a stream from a resource that would delete the file during the release? Or were you thinking more just create a wrapper for `nio.Files.createTempFile` and leave deletion for the user at some other point?

diesalbla:
Going through the list of items on the opening message, it looks as if the `Files[F[_]]` trait, together with `FileHandle[F[_]]`, are implementing most of the features requested on this Issue. This, together with the lack of activity for a year, is good  reason to close this issue.

PR #1540, opened 2019-07-16
Add more file IO operations

LukaJCB:
Just a WIP for #1533


mpilquist:
@LukaJCB Any progress on this on? I'd love to get it in 1.1.0 release that's coming up soon.

LukaJCB:
I'll have time to work on this tonight :)

LukaJCB:
This should be good to review now, it doesn't comprise the full feature set, but it's a good start I think

io/src/test/scala/fs2/io/file/FileSpec.scala
```diff
@@ -115,6 +116,110 @@ class FileSpec extends BaseFileSpec {
...
+
+  "createDirectories" - {
+    "should return in an existing path" in {
+      Stream
+        .resource(Blocker[IO])
+        .evalMap(
+          blocker =>
+            file
+              .createDirectories[IO](Paths.get("temp/inner"))
+              .bracket(file.exists[IO](_))(file.deleteIfExists[IO](blocker, _).void))
```
ybasket: Haven't tried, but looks to me like it would delete `inner`, but not `temp`? Maybe it would be better to run this test and the one above in temporary directories (like the other tests here) to not leave garbage behind or get trouble when executing the tests in parallel?

io/src/main/scala/fs2/io/file/file.scala
```diff
@@ -123,6 +127,72 @@ package object file {
...
+  /**
+    * Checks if a file exists
+    *
+    * Note that the result of this method is immediately outdated. If this
+    * method indicates the file exists then there is no guarantee that a
+    * subsequence access will succeed. Care should be taken when using this
+    * method in security sensitive applications.
+    */
+  def exists[F[_]: Sync](path: Path, flags: Seq[LinkOption] = Seq.empty): F[Boolean] =
+    Sync[F].delay(Files.exists(path, flags: _*))
```
mpilquist: Should this use `Blocker` since it's doing FS access?

LukaJCB: I wasn't sure in cases like these, since in general this should be super quick in most cases, but sure, I'll add it :)

mpilquist: Yeah, I've felt the same way. Flip side is that pretty much everything in `Files` API grabs a global mutex when looking up a file system.

LukaJCB: Very good point 👍

djspiewak: I think it's important to use `Blocker`, even here. Think about checking existence on an NFS mount, or even something more benign like a FUSE mount.

PR #1541, opened 2019-07-19
Add streams for entries of directories & file trees

ybasket:
Wrap up `getDirectoryStream(...)` and `walk(...)`  `java.nio.file.Files` into nice streams while taking care of resource management, blocking I/O and exceptions.

Implements parts of #1533, complementary to #1540.

### DIFF
--- a/io/src/main/scala/fs2/io/file/file.scala
+++ b/io/src/main/scala/fs2/io/file/file.scala
@@ -1,11 +1,14 @@
 package fs2
 package io
 
+import java.nio.file._
+import java.util.stream.{Stream => JStream}
+
+import scala.collection.JavaConverters._
 import scala.concurrent.duration._
 
-import java.nio.file.{Path, StandardOpenOption, WatchEvent}
-
 import cats.effect.{Blocker, Concurrent, ContextShift, Resource, Sync, Timer}
+import cats.implicits._
 
 /** Provides support for working with files. */
 package object file {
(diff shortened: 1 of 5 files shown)


io/src/main/scala/fs2/io/file/file.scala
```diff
@@ -119,4 +122,81 @@ package object file {
...
+  private def _runJavaCollectionResource[F[_]: Sync: ContextShift, C <: AutoCloseable](
+      blocker: Blocker,
+      javaCollection: F[C],
+      collectionIterator: C => Iterator[Path]): Stream[F, Path] =
+    Stream
+      .resource(Resource.fromAutoCloseable(javaCollection))
+      .flatMap(ds => _runIteratorOnBlocker[F, Path](blocker, collectionIterator(ds)))
+
+  // Like Stream.fromIterator, but on the given Blocker
+  private def _runIteratorOnBlocker[F[_]: ContextShift, A](blocker: Blocker, iterator: Iterator[A])(
```
mpilquist: Thoughts on making this `Stream.fromBlockingIterator`?

io/src/main/scala/fs2/io/file/file.scala
```diff
@@ -119,4 +122,81 @@ package object file {
     Stream
       .resource(Watcher.default(blocker))
       .flatMap(w => Stream.eval_(w.watch(path, types, modifiers)) ++ w.events(pollTimeout))
+
+  /**
+    * Creates a stream of [[Path]]s inside a directory.
+    */
+  def streamDirectory[F[_]: Sync: ContextShift](blocker: Blocker, path: Path): Stream[F, Path] =
```
mpilquist: `s/streamDirectory/directoryStream/`? Or `directoryContents`? I'm not keen on `streamDirectory` as it reads like a command to me.

mpilquist:
This looks great! Couple of nitpicky comments.

io/src/main/scala/fs2/io/file/file.scala
ybasket: Done

io/src/main/scala/fs2/io/file/file.scala
ybasket: I went with `directoryStream` as this is closer to the underlying Java naming.

Issue #1538, opened 2019-07-14
Behaviour of Chunk.scanLeft does not match standard collections


culturedsys:
The behaviour of the scanLeft method in Chunk does not match the behaviour of the method in the standard collections - specifically, the Chunk implementation does not include the initial value, and includes the final value twice, e.g:

```scala
Chunk(List(1, 2, 3)).scanLeft(0)(_ + _).toList == List(1, 3, 6, 10, 10)
```

Whereas

```scala
List(1, 2, 3).scanLeft(0)(_ + _) == List(0, 1, 3, 6, 10)
```

I'm not sure if this is intended or not. If it is intended, it seems like it would be good to mention it  in the documentation. If it isn't intended, there appears to be a simple fix - I can submit a PR.

mpilquist:
@culturedsys Yeah, this looks broken to me. A PR to fix would be great.

mpilquist:
Fixed by #1544

PR #1544, opened 2019-07-20
Change Chunk scanLeft behaviour to match standard collections

culturedsys:
Fix for #1538

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -213,17 +213,18 @@ abstract class Chunk[+O] extends Serializable { self =>
   def scanLeftCarry[O2](z: O2)(f: (O2, O) => O2): (Chunk[O2], O2) =
     scanLeft_(z, false)(f)
 
-  protected def scanLeft_[O2](z: O2, emitFinal: Boolean)(f: (O2, O) => O2): (Chunk[O2], O2) = {
+  protected def scanLeft_[O2](z: O2, emitZero: Boolean)(f: (O2, O) => O2): (Chunk[O2], O2) = {
     val b = collection.mutable.Buffer.newBuilder[O2]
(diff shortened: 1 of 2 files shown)


mpilquist:
Looks like this changed the behavior of `Stream#scan`, which causes the `pause` test to fail.

culturedsys:
The `Stream` methods `repartition`, `scan`, and `sliding` all use `Chunk#scanLeftCarry`, which depends on the same underlying implementation as `scanLeft`, so was changed by my patch. As the existing `Stream` methods depend on this behaviour, that suggests it's probably intended. I have attempted to re-implement the `Stream` methods to work with my implementation, and the existing implementations are probably more elegant (specifically, the existing behaviour of `scanLeftCarry` not including the zero element makes it easier to implement the correct behaviour for the `Stream` methods when they are called on an empty `Stream`, in which case there are no `Chunks` to operate on, and the zero element has to be manually processed).

So it may be best to leave the `Chunk` behaviour as it is. I can update my PR to include the re-implementations of the `Stream` functions if you think that's helpful; otherwise, I can submit a new PR clarifying the existing behaviour in the scaladoc.

mpilquist:
> Another alternative would be to leave scanLeftCarry as it is, but change scanLeft to match the behaviour of List et al.

I think I like this the best -- definitely if backwards compatibility wasn't a concern. Let's see what @systemfw thinks.

SystemFw:
> Another alternative would be to leave scanLeftCarry as it is, but change scanLeft to match the behaviour of List et al.

I like this the best as well, it's really weird to have `Chunk.scan`, which is a very well known name and concept, behave differently. Otoh `scanLeftCarry` it's our own thing and can have the most convenient semantics, which in this case would be to leave it as is since all the other methods rely on it

culturedsys:
I've pushed a new version, which maintains the existing `scanLeftCarry` behaviour.

mpilquist:
Looks great, thanks!

PR #1549, opened 2019-08-05
Additional socket options

Qthunder:
Addresses
https://github.com/functional-streams-for-scala/fs2/issues/1546

### DIFF
--- a/io/src/main/scala/fs2/io/tcp/tcp.scala
+++ b/io/src/main/scala/fs2/io/tcp/tcp.scala
@@ -1,5 +1,9 @@
 package fs2
 package io
 
+import java.net.SocketOption
+
 /** Provides support for TCP networking. */
-package object tcp
+package object tcp {
+  type SocketMapping[A] = (SocketOption[A], A)
+}
--- a/io/src/main/scala/fs2/io/tcp/tcp.scala
+++ b/io/src/main/scala/fs2/io/tcp/tcp.scala
@@ -1,9 +1,5 @@
 package fs2
 package io
 
-import java.net.SocketOption
-
 /** Provides support for TCP networking. */
-package object tcp {
-  type SocketMapping[A] = (SocketOption[A], A)
-}
+package object tcp
--- a/io/src/main/scala/fs2/io/tcp/SocketGroup.scala
+++ b/io/src/main/scala/fs2/io/tcp/SocketGroup.scala
@@ -47,7 +47,7 @@ final class SocketGroup(channelGroup: AsynchronousChannelGroup, blocker: Blocker
       receiveBufferSize: Int = 256 * 1024,
       keepAlive: Boolean = false,
       noDelay: Boolean = false,
-      additionalSocketOptions: List[SocketMapping[_]] = List.empty
+      additionalSocketOptions: List[SocketOptionMapping[_]] = List.empty
   )(implicit F: Concurrent[F], CS: ContextShift[F]): Resource[F, Socket[F]] = {
 
     def setup: F[AsynchronousSocketChannel] = blocker.delay {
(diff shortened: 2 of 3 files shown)


io/src/main/scala/fs2/io/tcp/tcp.scala
```diff
@@ -1,5 +1,9 @@
 package fs2
 package io
 
+import java.net.SocketOption
+
 /** Provides support for TCP networking. */
-package object tcp
+package object tcp {
+  type SocketMapping[A] = (SocketOption[A], A)
```
mpilquist: Minor style thing -- I'd either make this a case class or remove entirely. We use type aliases very sparingly as they introduce a new top level concept without providing additional type safety.

In this case, I think we can remove entirely and just pass `List[(SocketOption[A], A)]`. One thing that would sway me the other way is if we wanted to add some convenience constructors -- e.g. `SocketOptionValue.boolean(StandardSocketOptions.SO_BROADCAST, true)` -- that handled boxing to `java.lang.Boolean` etc.

Qthunder: I mentioned on the ticket but forgot to say here : I'm equally happy with a case class.

One reason I don't really like `List[(SocketOption[A], A)]` is that since the list is actually meant to be heterogenous in the general case, `A` will end up being `Any` in a lot of cases . That's why I liked the idea of the wildcard type combined with an alias/class that forces proper typing between the key (`SocketOption[A]`) and value (`A`) of each pair.

Another option, which achieves that objective without introducing a new type or type alias is having `List[(SocketOption[A], A) forSome { type A }]` as the type of `additionalSocketOptions` - how do you feel that? The flip-side is that you now have existential types in the signature.

mpilquist: Yep makes sense. Is that the deprecated form of existential? I'm good with a case class.

Qthunder: To summarize the above, I have 2 options in mind

1) ` additionalSocketOptions: List[(SocketOption[A], A) forSome { type A }]`
2) ```
 final case class SocketMapping[A](key: SocketOption[A], value: A)
  object SocketMapping {
    // convenience constructors
  }
 additionalSocketOptions: List[SocketMapping[_]]
```

Comment (mpilquist) [reply io/src/main/scala/fs2/io/tcp/tcp.scala]: Let's go with the case class. To bike shed the name a bit, I prefer `SocketOptionMapping` or `SocketOptionValue` or something like that. `SocketMapping` doesn't tell me it's related to options.

Comment (Qthunder) [reply io/src/main/scala/fs2/io/tcp/tcp.scala]: > Yep makes sense. Is that the deprecated form of existential? I'm good with a case class.

It's not deprecated, I would say it's a more general version of existential compared to the wildcard `_` which only works for certain use cases - same as the placeholder `_` arguments in anonymous functions or higher kinds expressed using kind-projector. (This use case is a good example, wildcard necessitates a type alias to express the same constraint)

Anyway - I'd be using a case class.

Comment (mpilquist) [reply io/src/main/scala/fs2/io/tcp/tcp.scala]: I meant deprecated in Scala 3, like shown below:

```scala
➜  dotr -version
Starting dotty REPL...
Dotty compiler version 0.17.0-RC1 -- Copyright 2002-2019, LAMP/EPFL
scala> case class SocketOption[A](name: String)
// defined case class SocketOption

scala> def configure(options: List[(SocketOption[A], A) forSome { type A }]) = 0
1 |def configure(options: List[(SocketOption[A], A) forSome { type A }]) = 0
  |                                                 ^^^^^^^
  |                               Existential types are no longer supported -
  |                               use a wildcard or dependent type instead
```

Qthunder: Ah, I see. Have pushed the case class solution - have added the convenience methods, although I somewhat question their utility now. For instance, the below code compiles successfully (I suppose it provides some help with type inference)

`val socketOptionMapping: SocketOptionMapping[java.lang.Boolean] = SocketOptionMapping(StandardSocketOptions.TCP_NODELAY, true)`

PR #1556, opened 2019-08-20
Move InvariantOps to benchmarks.

diesalbla:
The internal `FreeC` includes an implicit extension class `InvariantOps`, with a method `run` for running a `FreeC`. As it turns out, this method is only ever used in the benchmarks, not to implement Stream or Pull. Therefore, we can move it to that package.


pchlupacek:
Seems we need to resolve the conflicts here ?

pchlupacek:
@diesalbla thanks

PR #1577, opened 2019-09-02
Store the instance of Sync for FreeC of Algebra in Lazy Val.

diesalbla:
_Note_: merge only if #1578 is declined.


mpilquist:
Merged #1578 so closing this one.

PR #1579, opened 2019-09-03
Move append operation to FreeC abstract class.

diesalbla:
Since  `++` is a common Stream operator, it may make sense to _optimise_ and add a specialised operation in FreeC for it, as a variant of `transformWith`. 
This may save the cost of creating a PartialFunction for any use of `++`.
It also removes a reference to FreeC internals from `Stream`.


core/shared/src/main/scala/fs2/internal/FreeC.scala
```diff
@@ -35,6 +35,15 @@ private[fs2] sealed abstract class FreeC[F[_], +R] {
       }
     }
 
+  def append[R2](post: => FreeC[F, R2]): FreeC[F, R2] =
```
mpilquist: What do you think about calling this `andThen`?

diesalbla: Well, I would certainly avoid that word, since that is the composition method from the `Function` class of the Scala library. After which the similar method in `cats.data.Kleisli` takes its cue.

PR #1582, opened 2019-09-05
Add Stream.evals, evalSeq

### LINKED ISSUE #1571: Allow creating stream from sequence in effect
I think an operator that would combine `Stream.eval(fseq).flatMap(Stream.emits)` (similar to `Stream.evalUnChunk`, but for sequences) would be very nice to have to reduce some boilerplate when working with third-party functions that return results in batches.

kubukoz:
Closes #1571.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -185,7 +185,8 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
     * first successful attempt. Also see `retry`.
     */
   def attempts[F2[x] >: F[x]: Timer](
-      delays: Stream[F2, FiniteDuration]): Stream[F2, Either[Throwable, O]] =
+      delays: Stream[F2, FiniteDuration]
+  ): Stream[F2, Either[Throwable, O]] =
     attempt ++ delays.flatMap(delay => Stream.sleep_(delay) ++ attempt)
 
   /**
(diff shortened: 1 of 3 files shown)


kubukoz:
I can revert the changes scalafmt has made, if it'll make anything easier, but it's a separate commit anyway so should be easy for review

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3068,6 +3141,14 @@ object Stream extends StreamLowPriority {
   def evalUnChunk[F[_], O](fo: F[Chunk[O]]): Stream[F, O] =
     fromFreeC(Algebra.eval(fo).flatMap(Algebra.output))
 
+  /** Like `eval`, but lifts a foldable structure. **/
+  def evals[F[_], S[_]: Foldable, O](fo: F[S[O]]): Stream[F, O] =
+    eval(fo).flatMap(_.foldMap(Stream.emit))
```
kubukoz: I went with `foldMap`, but also considered `Stream.emits(foo.toList)`. Should I benchmark whether one would have a significant performance improvement over the other?

kubukoz:
What about the Seq variant? Maybe I should use fromIterator in it. I'll give that a try and maybe do a quick benchmark.

mpilquist:
Okay cool -- I suspect the `emits` will version will be faster, at least for lists. Though all I'm certain of is that benchmarks always surprise me. :)

mpilquist:
I'm going to merge as is for now, feel free to open a new PR if benchmark results change implementation. I'm getting the same Scalafmt results and want to avoid conflicts.

kubukoz:
`foldMap` seems to be consistently slower (by at least one order of magnitude) than `emits` when tested with List, scala.Stream and Chain. I'll replace it with `emits` in the subsequent PR ;)

The iterator solution is also massively slower than `emits`, so I'll leave `evalSeq` as it is for now.

If anyone wants to optimize for a specific collection/size, they can always use the more low-level operators themselves.

PR #1583, opened 2019-09-05
Add `evalFilter` and `evalFilterAsync`

### LINKED ISSUE #1570: Effectful filtering of streams operators
I often find myself looking for an operator that would allow filtering the stream with an effectful function.

Something like `def evalFilter(f: A => F[Boolean]): Stream[F, A]`, maybe also `def evalMapOption[B](f: A => F[Option[B]]): Stream[F, B]` (similar to how `TraverseFilter`'s methods look like).

I would like to contribute these, if it looks like a good idea.

kubukoz:
Creating a draft PR to discuss the signatures and implementations before going too deep in.

I have a few suggestions for the implementation of `evalFilter`:

- `evalMap` + `unNone` (current implementation in this PR)
- `flatMap` + `Stream.eval(f(elem)).map(_.guard[Option].foldMap(Stream.emits))`
- manual `pull.uncons` and operating on chunks (might require `Applicative[F]` I'm afraid)

May eventually close #1570

### DIFF
--- a/core/shared/src/test/scala/fs2/StreamSpec.scala
+++ b/core/shared/src/test/scala/fs2/StreamSpec.scala
@@ -914,6 +914,52 @@ class StreamSpec extends Fs2Spec {
       }
     }
 
+    "evalFilter" - {
+      "with effectful const(true)" in forAll { (s: Stream[Pure, Int]) =>
+        withClue("takes all elements") {
+          s.evalFilter(_ => IO.pure(true)).compile.toList.asserting(_ shouldBe s.toList)
+        }
+      }
+
+      "with effectful const(false)" in forAll { (s: Stream[Pure, Int]) =>
+        withClue("takes no elements") {
+          s.evalFilter(_ => IO.pure(false)).compile.toList.asserting(_ shouldBe empty)
+        }
+      }
+
+      "with function that filters out odd elements" in {
+        Stream
+          .range(1, 10)
+          .evalFilter(e => IO(e % 2 == 0))
+          .compile
+          .toList
+          .asserting(_ shouldBe List(2, 4, 6, 8))
+      }
+    }
+
+    "evalFilterAsync" - {
+      "with effectful const(true)" in forAll { (s: Stream[Pure, Int]) =>
+        withClue("takes all elements") {
+          s.covary[IO].evalFilterAsync(5)(_ => IO.pure(true)).compile.toList.asserting(_ shouldBe s.toList)
+        }
+      }
+
+      "with effectful const(false)" in forAll { (s: Stream[Pure, Int]) =>
+        withClue("takes no elements") {
+          s.covary[IO].evalFilterAsync(5)(_ => IO.pure(false)).compile.toList.asserting(_ shouldBe empty)
+        }
+      }
+
+      "with function that filters out odd elements" in {
+        Stream
+          .range(1, 10)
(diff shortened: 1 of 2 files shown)


mpilquist:
`_.guard[Option].as(o)` is a neat pattern, I'll be using that in the future.

Signatures look good to me. I don't think this operation needs to be optimized for chunks since each element results in effect evaluation, which should dwarf chunk concerns. We could document in ScalaDoc that singleton chunks are emitted and consider buffering as necessary.

kubukoz:
Okay, I'll continue with the current approach then. And add that to ScalaDoc.

I'll also do some quick benchmarks to choose between `unNone` and `flatMap`.

kubukoz:
Ready for a look @mpilquist

core/shared/src/test/scala/fs2/StreamSpec.scala
```diff
@@ -914,6 +914,91 @@ class StreamSpec extends Fs2Spec {
...
+      "with function that filters out odd elements" in {
+        Stream
+          .range(1, 10)
+          .evalFilterAsync[IO](5)(e => IO(e % 2 == 0))
+          .compile
+          .toList
+          .asserting(_ shouldBe List(2, 4, 6, 8))
+      }
+
+      "filters up to N items in parallel" in {
```
kubukoz: This one was tough... I wanted to avoid using `sleep` to ensure some parallelism is achieved, but I couldn't figure out a better way.

Possibly I could avoid using the semaphore (by failing from `sig.modify` if the value is too large) or the signal (by getting the amount of available permits, but that'd require looping the check in `Stream.repeatEval`). But I guess since this is test code we don't need to care about perf *that* much.

Unless there's a reliable way to ensure that the filtering action waits for a few elements to start processing - that'd speed up the test and probably ensure its determinism as well...

kubukoz:
Oops, still failing. I'll take a look...

kubukoz:
Okay, I just forgot `parJoin` is unordered. Back to `parEvalMap`.

kubukoz:
Oh come on, even a `Stream[SyncIO, ?]` can't be `unsafeRunSync`d on JS? That's not cool.

I'll use something different for the property tests then...

core/shared/src/test/scala/fs2/StreamSpec.scala
```diff
@@ -914,6 +914,99 @@ class StreamSpec extends Fs2Spec {
       }
     }
 
+    "evalFilter" - {
+      "with effectful const(true)" in {
+        val s = pureStreamGenerator[Int].sample
```
kubukoz: Not decided on the approach - taking a single sample is hardly a nice test, but I can't use `forAll` on a stream with effects because the JS build complains about `unsafeRunSync()` (even when `SyncIO` is used!). I guess using a concrete example (or a few) will be better...

mpilquist: Switch to `IO` instead of `SyncIO` and the tests should pass on Scala.js. You should also be able to use generators, just don't call `unsafeRunSync` -- `forAll { /* return F[Assertion] */ }`

kubukoz: I did that, and (locally) I'm still getting the same issue... `cannot synchronously await result on JavaScript; use runAsync or unsafeRunAsync` (even though I never used anything with `unsafe` in these tests)

PR #1593, opened 2019-09-07
Added bracketWeak, bracketCaseWeak, and onFinalizeWeak to allow scope-less resource management

mpilquist:
These three combinators behave like `bracket`, `bracketCase`, and `onFinalize` did in 1.0.5. One nice property of having these three operations is that we won't run in to a situation where someone was relying on the 1.0.5 behavior and there's no way to get it back in 1.1.0.


mpilquist:
Oh one note about this -- maybe since this allows registration to the current scope, we should make `Stream#scope` public again to allow manual scope creation.

mpilquist:
I restored `Stream#scope` and added scope-less `bracket`/`bracketCase`. I did NOT add equivalent scope-less methods for the cancelable variants of bracket as I'll be removing them in the next PR, along with the `Pull.acquireCancellable` stuff.

PR #1594, opened 2019-09-07
Add combinator unfoldLoopEval

ChristopherDavenport:
Just a convenience to allow people to work with streams rather than with `Pull` when they really just want an easy stream constructor

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3520,6 +3520,17 @@ object Stream extends StreamLowPriority {
     suspend(go(s))
   }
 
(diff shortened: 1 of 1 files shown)


mpilquist:
Add ScalaDoc too? Should we also add `unfoldLoop` -- non-F one?

kubukoz:
So this is like `unfold` but at least one element is produced? (since the Option is only on the "next" value we're iterating on, not on that and the produced element)

ChristopherDavenport:
Yes. It always produces a value, but may or may not iterate. This replaces the paradigm with unfold where you unnecessarily make that initial value an option, just to iterate each layer, and pass None as the final one. This makes behavior like pagination very easy to represent.

kubukoz:
Indeed, pagination is what I thought about. I like it :)

PR #1595, opened 2019-09-07
# Do not modify or remove the line above. # Everything below it will be ignored.

### LINKED ISSUE #1335: Stack-overflow occurring on 1.0.x (Works in 0.10.x)

### LINKED ISSUE #1357: java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
```scala
java.lang.NoSuchMethodError: java.nio.ByteBuffer.flip()Ljava/nio/ByteBuffer;
    at fs2.io.udp.AsynchronousSocketGroup$$anon$1.fs2$io$udp$AsynchronousSocketGroup$$anon$$read1(AsynchronousSocketGroup.scala:214)
    at fs2.io.udp.AsynchronousSocketGroup$$anon$1$$anon$2.run(AsynchronousSocketGroup.scala:344)
```

This is caused by compiling with Java 9+ and running on Java 8. See https://jira.mongodb.org/browse/JAVA-2559 for a detailed description. We could manually upcast before calling these methods. Could be hard to prevent such errors from reoccurring though.

### LINKED ISSUE #1371: `io.file.writeAll` does not append with `StandardOpenOption.APPEND`
It seems that `StandardOpenOption.APPEND` is being ignored, instead `io.file.writeAll` will just overrwrite the file.

The below example results in two files; `test-fs2.out` and `test-java.out`. They both should contain the text:
```
a
b
c
```
However `test-fs2.out` only contains the letter `c`.

```scala
import cats.effect._

import fs2._
import fs2.io

import java.nio.file._

import scala.language.higherKinds

object Scratch {
  def writeWithStream[T[_]](p: Path, s: String)
      (implicit F: ConcurrentEffect[T], cs: ContextShift[T]): T[Unit] = {
    Stream(s)
      .covary[T]
      .through(text.utf8Encode)
      .through(
        io.file.writeAll(
          p,
          scala.concurrent.ExecutionContext.global,
          Seq(StandardOpenOption.CREATE, StandardOpenOption.APPEND)
        )
      )
      .compile
      .drain
  }

    implicit val executionContext: ExecutionContext =
      scala.concurrent.ExecutionContext.Implicits.global

    implicit val contextShift: ContextShift[IO] =
      IO.contextShift(executionContext)

    val outPathfs2: Path = Paths.get("E:/Documents/test-fs2.out")
    val outPathJava: Path = Paths.get("E:/Documents/test-java.out")

    val fs2write = for {
      _ <- writeWithStream[IO](outPathfs2, "a\n")
      _ <- writeWithStream[IO](outPathfs2, "b\n")
      _ <- writeWithStream[IO](outPathfs2, "c\n")
    } yield ()

    fs2write.unsafeRunSync()

    Files.write(outPathJava, "a\n".getBytes(), StandardOpenOption.CREATE, StandardOpenOption.APPEND)
    Files.write(outPathJava, "b\n".getBytes(), StandardOpenOption.CREATE, StandardOpenOption.APPEND)
    Files.write(outPathJava, "c\n".getBytes(), StandardOpenOption.CREATE, StandardOpenOption.APPEND)

    println("done")
  }
}
```

### LINKED ISSUE #1404: groupedWithin emits empty groups

### LINKED ISSUE #1411: Seq inconsistency between 2.12 and 2.13

### LINKED ISSUE #1431: Properly handle blocking IO in watch API

### LINKED ISSUE #1443: Stream#translate forgets ability to interrupt effect evaluation

### LINKED ISSUE #1447: Closing/flushing in `writeOutputStream`
[Current implementation](https://github.com/functional-streams-for-scala/fs2/blob/4dde73e80b1828406e60460596aa19cefc6b2778/io/src/main/scala/fs2/io/io.scala#L111-111) of `writeOutputStream` has two problems.

1. It doesn't `.flush` the `OutputStream`. While *most* `OutputStream` implementations `.flush` on `.close`, this is neither documented nor guaranteed. We shouldn't rely on that.
1. It closes the stream on a CPU-bound pool. However, `.close` *might* flush, which *might* block!

So the relevant line should be replaced with something like this:
```scala
Stream.bracket(fos) { os =>
  cs.evalOn(blockingExecutionContext)(F.delay(os.close()))
}.flatMap { os =>
  useOs(os).scope ++ cs.evalOn(blockingExecutionContext)(F.delay(os.flush()))
)
```

kubukoz:
Requesting a pull to functional-streams-for-scala:series/1.0 from kubukoz:evals-perf

Write a message for this pull request. The first block
of text is the title and the rest is the description.

Changes:

5d4f67ed (Jakub Kozłowski, 55 seconds ago)
   Optimize `evals` for lists

51bec163 (Paul Cleary, 31 hours ago)
   Adding vinyldns to README

b7ee0ad8 (Michael Pilquist, 33 hours ago)
   Scalafmt updates

c9c7a96b (Michael Pilquist, 2 days ago)
   Setting version to 1.1.0-SNAPSHOT

87ed893e (Michael Pilquist, 2 days ago)
   Setting version to 1.1.0-M2

337bb293 (Michael Pilquist, 2 days ago)
   Upgraded to latest sbt-scalafmt release and removed coursier since it is
   now built in to sbt

29566c13 (Jakub Kozłowski, 2 days ago)
   Add `evals`, `evalSeq`

d2101a37 (Jakub Kozłowski, 2 days ago)
   Set scalafmt version + format Stream.scala

4d42f0c9 (Scala Steward, 3 days ago)
   Update sbt to 1.3.0

b49a7dba (Scala Steward, 4 days ago)
   Update sbt-sonatype to 2.6

dc7b4216 (Michael Pilquist, 4 days ago)
   Adjusted compile.resource scoping such that only the *last* top-level scope
   is extended to resource lifecycle

813f1728 (Diego E. Alonso-Blas, 4 days ago)
   Move append operation to FreeC abstract class.

   Given that append is probably very used Stream operator, it may make sense
   to add a specialised operation to the FreeC class itseld, as a variant of
   transformWith.

   This may save the cost of creating a PartialFunction for any call. It also
   removes a use of FreeC internals from `Stream`.

796704cc (Diego E. Alonso-Blas, 5 days ago)
   Unfold-expand recoverWith

   The method Result.recoverWith is only used in one single place, for the
   bracket method, so it can be unfolded. This avoids creating the
   PartialFunction object that is passed.

04b893eb (Michael Pilquist, 4 days ago)
   Fixed cancelation bug with nested scopes

ef15417c (Diego E. Alonso-Blas, 5 days ago)
   Remove the instance of Sync for FreeC.

   Since this is a private type, its type-class instances are of no use. All
   we need is the bracketCase method, because it is used in Pull.

3f8d4b9c (Scala Steward, 5 days ago)
   Update sbt-microsites to 0.9.4

2e203064 (Michael Pilquist, 5 days ago)
   Removed scope insertion when converting pull to stream

c88e4195 (Michael Pilquist, 5 days ago)
   Chanegd scoping behavior to better align with expectations

05c22b2e (Jakub Kozłowski, 6 days ago)
   Restore Google Cloud

4ef07cf9 (Jakub Kozłowski, 6 days ago)
   Add fs2-ssh

7b53033f (Scala Steward, 10 days ago)
   Update sbt-microsites to 0.9.3

4290aaf0 (Fabio Labella, 12 days ago)
   Fix typo in CONTRIBUTORS

ee49c75f (Scala Steward, 12 days ago)
   Update cats-core, cats-laws to 2.0.0-RC2

4a7d81d7 (heraklos, 12 days ago)
   Fix small typo

7e7afe97 (Diego E. Alonso Blas, 12 days ago)
   Mention

17e17f52 (Diego E. Alonso-Blas, 12 days ago)
   FreeC.raiseError: remove unneeded type parameter

8e51c373 (Scala Steward, 2 weeks ago)
   Update reactive-streams, ... to 1.0.3

ff67525c (Scala Steward, 2 weeks ago)
   Update sbt-mima-plugin to 0.6.0

36b1279c (Diego E. Alonso-Blas, 3 weeks ago)
   Algebra: replace the R class with a CompileEnd trait

   The `compileLoop` method has a system of nested executions, which
   correspond to the Step algebra operations, in which an inner stream is
   executed with a different CompileScope. This was implemented with an inner
   ADT, using the word R
   (for Result) and three cases:

   - The compilation on a certain Scope is done, without result.
   - The compilation is stopped with some output values, and the
    rest of the stream; or
   - The compilation was interrupted.

   This ADT is only used as a way of messaging from one part of the method to
   another: objects created in one part are deconstructed in another part.

   This commit replaces the use of that ADT by a "object algebra", a trait
   called "CompileEnd", whose methods correspond to the three cases of that
   ADT.

   The CompileEnd acts as a continuation in CPS: it indicates how the 
   execution of the stream should proceed when it reaches an end-point.

   There are two (sub)-types of CompileEnd: the first one is used for
   returning from the `compileLoop` method. the second one is used when
   dealing with a Step instruction.

7266a01d (Diego E. Alonso-Blas, 3 weeks ago)
   FreeC.InvariantOps: move to benchmarks.

   The internal `FreeC` includes an implicit extension class
   `InvariantOps`, with a method `run` for running a FreeC. As it turns out,
   this method is only ever used in the benchmarks, not to implement Stream or
   Pull. Therefore, we can move it to that package.

ba51708b (Diego E. Alonso-Blas, 3 weeks ago)
   Use back-to-front letters for intermediate Result types

   The use of the `R` as the return type of the `mk` was a bit confusing when
   dealing with the Bind-on-Bind case, in which we end up needing two extra
   variables (for types of intermediate results of Binds). Using letters Z, Y,
   and X, conveys the intuition of sequencing better.

34a2aee8 (Scala Steward, 3 weeks ago)
   Update sbt-scalafix to 0.9.6

bafe86f1 (Diego E. Alonso-Blas, 3 weeks ago)
   Force evaluation/access to the delegate

   Before this commit, in the Bind-on-Bind case we were creating an outer Bind
   (`nb`) whose `cont` method would create an inner `Bind`, and the
   `cont` method of the inner `Bind` would call to b.delegate. However,
   because the `delegate` is a value, this effectively meant that a reference
   to the `b` object was kept in the closure of `cont`. This was preventing
   the Garbage collector for reclaiming the object
   `b`, and making it not possible to optimise that call, even though the 
   delegate is an immutable field.

   In this commit, we force the evaluation of `b.delegate` expression, and
   kept its value inside the new outer `Bind`, which allows us to discard the
   `b`. This helps improve the performance in benchmarks.

e3cfbcc0 (Luka Jacobowitz, 3 weeks ago)
   Add Blocker to all functions

2d60a14c (Luka Jacobowitz, 4 weeks ago)
   Run directory tests in temp dirs

89a97475 (Luka Jacobowitz, 4 weeks ago)
   Add tests and scaladocs

a0a9da50 (Diego E. Alonso-Blas, 4 weeks ago)
   No Sync for compileLoop

af68847c (Scala Steward, 4 weeks ago)
   Update cats-effect, cats-effect-laws to 2.0.0-RC1

a1da5bf3 (GabrielAsman, 5 weeks ago)
   Convenience constructors.

c4ebfdd4 (GabrielAsman, 5 weeks ago)
   Replace type alias with case class.

3f7107b2 (Scala Steward, 5 weeks ago)
   Update cats-core, cats-laws to 2.0.0-RC1

ed9eccaf (GabrielAsman, 5 weeks ago)
   Remove extra new line

ddd7a530 (GabrielAsman, 5 weeks ago)
   Additional socket options.

e4f47914 (Scala Steward, 5 weeks ago)
   Update cats-effect, cats-effect-laws to 2.0.0-M5

e5df67a3 (Andrew Bushmin, 5 weeks ago)
   remove extra space in guide.md

1b6681df (Michael Pilquist, 6 weeks ago)
   Removed another ZIO dead link

30435066 (Michael Pilquist, 6 weeks ago)
   Removed dead link to ZIO stream

9daff9e8 (Tim Fisken, 6 weeks ago)
   Make Chunk#scanLeft behaviour match standard collections

   Changes Chunk#scanLeft to match the behaviour of the scanLeft method on the
   standard collections, by including the zero element and only including the
   final element once, while maintining the existing behaviour of
   Chunk#scanLeftCarry.

   Also adds tests for this behaviour, by comparing the result of 
   Chunk#scanLeft with List#scanLeft.

1e0ed42c (Tim Fisken, 6 weeks ago)
   Add tests for Chunk#scanLeftCarry

   Adds explicit tests for Chunk#scanLeftCarry for Chunks of 0, 1, and 2 Int
   elements, and a property based test based in terms of List#scanLeft.

38b4723e (Vitaly Lavrov, 7 weeks ago)
   Fix SocketGroup.open mehtod docs

29d3be2a (Yannick Heiber, 7 weeks ago)
   Add Stream.fromBlockingIterator

85b55884 (Yannick Heiber, 7 weeks ago)
   Rename streamDirectory to directoryStream

0f79f085 (Lucas Satabin, 7 weeks ago)
   Add reference to fs2-data

4688a66d (Yannick Heiber, 7 weeks ago)
   Add tests for streamDirectory & walk

0ea7eeca (Yannick Heiber, 7 weeks ago)
   Add streamDirectory and walk methods on file

f94adf69 (Luka Jacobowitz, 8 weeks ago)
   WIP add more file IO operations

0bd6d86e (Michael Pilquist, 8 weeks ago)
   Upgraded to mima 0.5.0

6e636009 (Luis Miguel Mejía Suárez, 8 weeks ago)
   Adding the Partially-Applied trick to Stream.fromIterator

87904a50 (Ikrom, 10 weeks ago)
   typo fixed

bd1052f3 (Diego E. Alonso-Blas, 10 weeks ago)
   ViewL: avoid using Function1 field, use method instead.

   In line with the recent changes to Bind, we now change the View class,
   (intended to represent an Eval + a continuation), so as to turn the `next`
   function into a class method.

   To handle the special case of an Eval, for which a "pureContinuation" was
   introduced, we add a special subclass of View, PureView, which uses the
   `._asFreeC` call. This subclass is kept private to the View, so as to not
   raise the exhaustiveness check.

3331d800 (Diego E. Alonso-Blas, 10 weeks ago)
   Fix the Stack Overflows in creating a Bind-Bind.

   The previous commit was causing some stack overflows in the inner cont 
   method, which was delegating to the `b.cont` method of other Bind. Unlike
   functions, where the reference to the continuation was passed, when using
   objects method calls are not "aliviated", and instead create cost in the
   Stack.

   To solve this problem, we add to the Bind class a new field, a
   `delegate`, which points to another Bind object of same types. This allows
   us to simulate a delegated call: when the main cont method is called, it
   follows the chain of delegation (in a loop), without adding extra Stack
   frames, until it reaches the non-delegated one.

   Instead of unrolling the delegate fields when we call to the `cont` method,
   we can directly unroll the first step upon creating the object. In which
   case, every Bind created during the mk loop will be referring to the first
   non-delegated bind.

c0073536 (Diego E. Alonso-Blas, 2 months ago)
   FreeC.Bind: use a method to represent a function.

   In the FreeC datatype, the Bind constructor represents the case of a bind
   or flatMap operation in a free monad constructor: it takes the step
   FreeC[A], and the continuation of what to do with the result. Currently,
   the implementation represents that continuation using a value of type
   `Function1`.

   However, in Scala there is another way to represent that a record has a
   function attached: by declaring an abstract method, and using inheritance
   as a form of parameterisation.

   This commit changes the Bind case class to an abstract class with an
   abstract `cont` method, which represents that continuation. This avoids the
   allocation and indirection of `Function1` instances,

cd0cd1b4 (Scala Steward, 2 months ago)
   Update sbt-scalajs-crossproject to 0.6.1

b0e5f526 (Scala Steward, 2 months ago)
   Update sbt-microsites to 0.9.2

e504b9b9 (Georgi Krastev, 8 months ago)
   Revert "Applied workaround for scala/bug 11303 from @joroKr21"

   This reverts commit e0eb895113352ce429560b0b2170f0d0cf4bffd4.

d88c7cd4 (Ikrom, 3 months ago)
   text comment changed

4156354e (Ikrom, 3 months ago)
   text comment changed

3474a7e2 (Ikrom, 3 months ago)
   possibly typo fixed

e8cb192e (Ikrom, 3 months ago)
   typo fixed

dc8e01f6 (Scala Steward, 3 months ago)
   Update sbt-doctest to 0.9.5

0277acf1 (Scala Steward, 3 months ago)
   Update sbt-jmh to 0.3.7

35244dda (Michael Pilquist, 3 months ago)
   Added Alternative[Chunk] and TraverseFilter[Chunk] instances

c0a55138 (Michael Pilquist, 3 months ago)
   Updated Stream.emits to take a scala.collection.Seq to match 2.12
   Stream.emits and 2.13 Chunk.seq behavior

d9ebb572 (Fabio Labella, 3 months ago)
   Add conversions to BitVector and ByteVector

006ea6fb (Michael Pilquist, 3 months ago)
   Changed Pull#{stream,streamNoScope} to require Unit result type

d8234184 (Diego E. Alonso Blas, 3 months ago)
   Replace a Option.fold with an Option.isEmpty

   It may save allocating a lambda. Even if it can be optimized away, 
 the
   first-order code in this case may be easier to read.
bcb1d1db (Michael Pilquist, 3 months ago)
   Updated ScalaDoc of file.tail

5472dae9 (shn-amn, 3 months ago)
   `file.keepReading` renamed to `file.tail` with an optional `offset` param.

7c5e9be7 (shn-amn, 3 months ago)
   Using `fs2.io.file.writeAll` rather than `java.nio.Files.write`.

53bc4e0a (shn-amn, 3 months ago)
   Much clearer!

40b9f91f (zakolenko, 3 months ago)
   fixed doctest (s.compile.toMap)

6238c969 (shn-amn, 3 months ago)
   Avoiding side effects.

520a53e8 (shn-amn, 3 months ago)
   Delay must be evaluated!

2ce2c2f9 (Michael Pilquist, 3 months ago)
   Setting version to 1.1.0-SNAPSHOT

681fd90a (Michael Pilquist, 3 months ago)
   Setting version to 1.1.0-M1

f95efb88 (zakolenko, 3 months ago)
   Added stream.compile.toMap syntax for compiling to a Map

f5af03e0 (Michael Pilquist, 3 months ago)
   Update docs/src/ReadmeExample.md

   Co-Authored-By: Bjørn Madsen <bm@aeons.dk>
1c0f76ad (Michael Pilquist, 3 months ago)
   Update docs/ReadmeExample.md

   Thanks!

   Co-Authored-By: Bjørn Madsen <bm@aeons.dk>
6321e42a (Bjørn Madsen, 3 months ago)
   Scala 2.13

df3a6408 (AdrianRaFo, 3 months ago)
   minor typo

b645b3c8 (AdrianRaFo, 3 months ago)
   address comments

9425be93 (AdrianRaFo, 3 months ago)
   restore ops

9378dd24 (Michael Pilquist, 3 months ago)
   Upgraded scodec-bits

bfa3bd1d (shn-amn, 3 months ago)
   Test for file.keepReading.

ece2bd40 (shn-amn, 3 months ago)
   file.keepReading reads a file as it's being appended.

b41c50b1 (Michael Pilquist, 3 months ago)
   Setting version to 1.0.6-SNAPSHOT

0f8faf32 (Michael Pilquist, 3 months ago)
   Setting version to 1.0.5

6082b433 (Michael Pilquist, 3 months ago)
   Marked an intermitting failing test pending

9dd569c6 (Michael Pilquist, 3 months ago)
   Fixed resource safety test that failed often

8518390a (AdrianRaFo, 3 months ago)
   avoid extra type parameter

4447a5ad (Michael Pilquist, 3 months ago)
   Rolled back cats-core to 1.6.1 in prep for 1.0.5 release

1c5a977d (Michael Pilquist, 3 months ago)
   Made utf8Bom private

a97e3ac6 (Michael Pilquist, 3 months ago)
   Updated UTF8 decoding to ignore the optional byte order mark - fixes

79c05307 (AdrianRaFo, 3 months ago)
   flatTap comes from cats

db616a44 (AdrianRaFo, 3 months ago)
   evalMap function doesn't need to return a mandatory Unit

6ed17f89 (AdrianRaFo, 3 months ago)
   add flatTap to Stream

44ae957f (Dmitry Polienko, 3 months ago)
   Fix Chunk.Bytes offset handling

60f2d18e (Dmitry Polienko, 3 months ago)
   Remove a custom generator from JavaInputOutputStreamSpec

c42170e9 (Michael Pilquist, 3 months ago)
   Replaced blockingExecutionContext with Blocker

a666a4fd (Michael Pilquist, 3 months ago)
   Removed hard-coded pool size in SocketGroup constructor

39fdf586 (Michael Pilquist, 3 months ago)
   Removed hard-coded pool size in SocketGroup constructor

0ff275d8 (Michael Pilquist, 3 months ago)
   Removed blocking call from io.scala

7468892f (Scala Steward, 3 months ago)
   Update tut-plugin to 0.6.12

70e63c5c (Michael Pilquist, 3 months ago)
   Removed blocking/BlockContext

35d41555 (Michael Pilquist, 3 months ago)
   Fixed a few blocking calls in file support

c7fec3f6 (Michael Pilquist, 3 months ago)
   Remove old AsyncYield type class

5a5fe7f9 (Michael Pilquist, 3 months ago)
   Updated io.file.Watcher to use a blocking execution context - fixes #1431

acbe3400 (Michael Pilquist, 3 months ago)
   Removed unsafe AsynchronousSocketGroup from UDP API

766fda21 (Michael Pilquist, 3 months ago)
   Removed unsafe AsynchronousChannelGroup from tcp API

78bbab28 (Michael Pilquist, 3 months ago)
   Removed deprecated io methods and shifted blocking TCP tasks to a blocking
   execution context

a8da2992 (Michael Pilquist, 3 months ago)
   Adjusted Travis for new branch

1b85bec3 (Michael Pilquist, 3 months ago)
   Revert "Dropped 2.13 build from 1.0 release series"

   This reverts commit 8770c1113a46c20d18b80087a022673444f0df2a.

8770c111 (Michael Pilquist, 3 months ago)
   Dropped 2.13 build from 1.0 release series

3a93cc21 (Michael Pilquist, 3 months ago)
   Updated version to 1.1.0-SNAPSHOT

be983cb0 (Michael Pilquist, 3 months ago)
   Yield CPU on Scala.js

b1ee75a1 (Michael Pilquist, 3 months ago)
   Removed erroneous ScalaDoc

9b4d6f82 (Michael Pilquist, 3 months ago)
   Flush in writeOutputStream - fixes #1447

20ac0970 (Michael Pilquist, 3 months ago)
   Added translateInterrupt - fixes #1443

01225ce7 (Michael Pilquist, 3 months ago)
   Updated comment to be more accurate

72e6e5e1 (Michael Pilquist, 3 months ago)
   Stick with global EC on Scala.js for tests

06a7f302 (Michael Pilquist, 3 months ago)
   Restored default ScalaTest execution context in tests to avoid intermittent
   ConcurrentModificationExceptions

cc710939 (Adam Chlupacek, 3 months ago)
   Propagate error from inner stream of parJoin before continue on with ++.

0e2a5347 (Mark Tomko, 3 months ago)
   Replace fixed thread pool with cached thread pool for example blocking
   execution context

c8a8d684 (Michael Pilquist, 3 months ago)
   Changd Topic to use SizedQueue

c14986f1 (Michael Pilquist, 3 months ago)
   Fixed compilation errors in experimental module

e200d9eb (Michael Pilquist, 3 months ago)
   Removed unneeded PubSub.withSize

a385ca32 (Michael Pilquist, 3 months ago)
   Moved queue + size tuple in to a new type which results in O(1) size for
   all queue types (e.g. circular buffers and unbounded inspectable queues)

a0f72b09 (Michael Pilquist, 3 months ago)
   Updated experimental module

1ab754e7 (Akhtyam Sakaev, 3 months ago)
   fix tailrec annotations

289869eb (Alex Berezovskiy, 3 months ago)
   Remove API links from documentation page

fece00b3 (Alex Berezovskiy, 3 months ago)
   Update API links in README

c2b942f9 (Alex Berezovskiy, 3 months ago)
   Update sbt and plugins

b35a5a05 (Alex Berezovskiy, 3 months ago)
   Update Scala 2.12 to 2.12.8

696cef01 (Fabio Labella, 3 months ago)
   Avoid executing function twice in `mapFilter`

5c268a8b (Michael Pilquist, 3 months ago)
   Fixed 4 test failures on Scala.js

6ee0eff6 (Michael Pilquist, 3 months ago)
   Corrected scalatestplus-testng version for 2.13.0-RC2

0801493b (Michael Pilquist, 3 months ago)
   Made a test pending on Travis

e13ba359 (Michael Pilquist, 3 months ago)
   Corrected scalatestplus-scalacheck version for 2.13.0-RC2

aafbb21e (gchew, 3 months ago)
   Modified fs2.test.verbose value to be true to fix test execution issues

cd80659b (gchew, 3 months ago)
   Added code to build.sbt to enable the fs2.test.verbose system property to
   be reflected in the test javaOptions. This fixes the issue where
   flickersOnTravis in Fs2Spec was always a no-op. Marked the test 'inner
   stream finalizer always runs before switching' as pending because it was
   observed hanging both in the Travis environment and locally. Re-enabled 3
   tests that should run in the series/1.0 branch

d61f6e63 (gchew, 3 months ago)
   Reverted three tests from the topic/fs3 branch that were originally marked
   as pending back to pending

03c92cda (gchew, 3 months ago)
   Added flickersOnTravis for repeatN 5

bcba7bab (gchew, 3 months ago)
   Changed timeout to 60 seconds

61142bc6 (Ross A. Baker, 4 months ago)
   Upgrade to kind-projector-0.10.1

7e5b3340 (Ross A. Baker, 4 months ago)
   Enough scalatest and scalatestplus to compile on 2.12

3a0b30c7 (Ross A. Baker, 4 months ago)
   WIP: Support Scala 2.13.0-RC2

97463e85 (gchew, 4 months ago)
   Disable fatal warnings for test compilation because sbt-doctest generated
   tests generate warnings which lead to test failures

4f60a02f (Tim McIver, 4 months ago)
   Use more idiomatic method to patch renamed Sink.

d960793a (Tim McIver, 4 months ago)
   Use a match on `name` rather than `if`.

eb2371d8 (Diego E. Alonso-Blas, 4 months ago)
   CompileScope.open: rearrange operations

   We reorganise the operations of the `open` method of the
   `CompileScope` class.

   We change `createCompileScope` from being an `Option` of the
   `InterruptContext` to directly building the new scope. This hides the
   `newScopeId` in this F-value.

   This operation saves the allocations of a `Some` and a Tuple2 objects, 
   although now the `CompileScope` is now built regardless of path. Not a big
   loss, since this is likely the frequent case.

9e0105ea (Tim McIver, 4 months ago)
   Got tests passing including Sink rename.

d5b1f8fc (Scala Steward, 4 months ago)
   Update sbt-scalajs to 0.6.28

9ff06c43 (Tim McIver, 4 months ago)
   Failing attempt at matching `Sink` using a `SymbolMatcher`.

e8d11c66 (Akhtyam Sakaev, 4 months ago)
   fix recursive aux functions naming differences

f8d61649 (gchew, 4 months ago)
   Added reactive-streams tests. Restored imports to match code in topic/fs3
   after they were modified by IntelliJ. Enabled 3 tests that were pending
   that consistently succeed on the series/1.0 branch.

5b54043b (Mike Gibson, 4 months ago)
   Adding Chromaprint.scala to projects using fs2

0adab092 (gchew, 4 months ago)
   Added fs2 io tests

22644dc9 (gchew, 4 months ago)
   Initial round of changes. More work needed

6a66d0a3 (Damien Bailly, 4 months ago)
   Track queue size of bounded queues

745459b6 (Adam Chlupacek, 4 months ago)
   Fix subscriber unregister under heavy concurrent load.

3a8c0c4f (Gabriel Volpe, 4 months ago)
   Updating links

672b16c7 (Michael Pilquist, 4 months ago)
   Simplified chunkMin implementation to emit eagerly -- i.e., as soon as it
   has accumulated a chunk of the minimum size

c7b49996 (Travis Brown, 4 months ago)
   Cogen.apply interacts badly with SAM syntax backport on 2.11

d1a73e35 (Jakub Kozłowski, 4 months ago)
   Build for 2.13.0-RC1

7dd748fd (Travis Brown, 4 months ago)
   Format

fb7bab40 (Travis Brown, 4 months ago)
   Fix #1454 on 2.13

01105cc3 (Travis Brown, 4 months ago)
   Revert "kind-projector has moved to typelevel org"

   This reverts commit 44cc64471a4a59c95567427ecf0322ec97012bd1.

615e42cd (Travis Brown, 4 months ago)
   Replace deprecated ⇒

05a253c7 (Travis Brown, 4 months ago)
   Remove scala-collection-compat dependency

44cc6447 (Seth Tisue, 4 months ago)
   kind-projector has moved to typelevel org

448854f2 (Paulius Imbrasas, 4 months ago)
   Add fs2-google-pubsub as related project

b065d583 (Scala steward, 4 months ago)
   Update cats-effect, cats-effect-laws to 1.3.0

e4514016 (Tim McIver, 5 months ago)
   Change a mistaken Pipe to Sink.

2abab664 (Diego E. Alonso-Blas, 5 months ago)
   CompileScope.State: Small cleanup and optimisation

   - Optimise `unregisterChild`: do not copy if there was no child.
   - Remove method `unregisterResource` which is unused.
   - Unify methods "findAncestor" and "findSelfOrAncestor",
    since one was a special case of the other.

   - Privatise other methods of the State class.

7633eb9c (Scala steward, 5 months ago)
   Update sbt-jmh to 0.3.6

a113b606 (Chris Davenport, 5 months ago)
   Use Chunk Queue to avoid concatenation

071a27f6 (Oleg Pyzhcov, 5 months ago)
   Add dual constructors for fs2.concurrent datatypes

99643190 (rhong, 5 months ago)
   Avoid calculate size of emits

0393a079 (Scala steward, 5 months ago)
   Update tut-plugin to 0.6.11

14fb7314 (Brian P. Holt, 5 months ago)
   add FunctionK instances for Stream and Pull

ffabfa04 (Christopher Davenport, 5 months ago)
   Cleanup Remnants of shifting test positions

edaaede2 (Christopher Davenport, 5 months ago)
   Additional Tests that should hold

064a73a0 (Chris Davenport, 5 months ago)
   Add Tests for Behavior

7c47c5ef (Scala steward, 5 months ago)
   Update sbt-scalajs, scalajs-compiler to 0.6.27

765fbad1 (Chris Davenport, 5 months ago)
   Fix Example Use

5160bc51 (Chris Davenport, 5 months ago)
   Add chunkMin Combinator

fe3f08ee (Scala steward, 5 months ago)
   Update sbt-osgi to 0.9.5

d9521940 (Scala steward, 5 months ago)
   Update sbt-jmh to 0.3.5

e0e81ed3 (Scala steward, 6 months ago)
   Update sbt-travisci to 1.2.0

11f8fac9 (Thrasos Thrasyvoulou, 6 months ago)
   Update v1.0.3 to v1.0.4 on README.md

b6f6c61c (Tim McIver, 6 months ago)
   Account for `Stream.to` as non extension method.

7e41f798 (Tim McIver, 6 months ago)
   Tests for SinkToPipeRules passing.

6e4f1258 (Fabio Labella, 6 months ago)
   Try to avoid race condition in tests

55ee1f3a (Fabio Labella, 6 months ago)
   Revert "Remove short timeout for CI"

   This reverts commit cd1d38b53e8fe3f04e4db1a7a9ebdde5ff12a5ad.

cd1d38b5 (Fabio Labella, 6 months ago)
   Remove short timeout for CI

63e95e64 (Fabio Labella, 6 months ago)
   Make resource compilation cancelable

0b175de9 (Fabio Labella, 6 months ago)
   Add test for cancelability of the Resource compiler

a34ed54d (Michael Pilquist, 6 months ago)
   Adjusted rechunkRandomlyWithSeed so that it does not require Sync
   constraint

08c3c440 (Michael Pilquist, 6 months ago)
   Minor ScalaDoc adjustment

f64eabf9 (Arnau Abella Gassol, 6 months ago)
   Chunkiness test

1f1a8913 (Arnau Abella Gassol, 6 months ago)
   First proposal

024c4266 (Arnau Abella Gassol, 6 months ago)
   stdoutLines and stdinLines

0056398f (wedens, 6 months ago)
   add `toInputStreamResource` and implement `toInputStream` in terms of it

cc64d9b4 (Arnau Abella Gassol, 6 months ago)
   Small changes for correctness.

ca69f3e5 (Arnau Abella Gassol, 6 months ago)
   Remove parJoin from PubSub example in docs

08d4234c (Piotr Gabara, 6 months ago)
   Add Chunk.toNel

927f1098 (Scala steward, 6 months ago)
   Update scalatest to 3.0.6

0af60b53 (Michael Pilquist, 6 months ago)
   Setting version to 1.0.5-SNAPSHOT

0fee7772 (Michael Pilquist, 6 months ago)
   Setting version to 1.0.4

c9c317d5 (Fabio Labella, 7 months ago)
   Fix scope leasing in parJoin

d0d36fdb (Fabio Labella, 7 months ago)
   Add test for interruptible eval

5970e00b (Fabio Labella, 7 months ago)
   Remove obsolete comment

f3c93d60 (Fabio Labella, 7 months ago)
   MonoidK instance for Stream

492609a4 (Fabio Labella, 7 months ago)
   Wip: Alternative instance

e937c2d2 (Michael Pilquist, 7 months ago)
   Fixed 2.13.0-M5 build by working around scalac crash

8ea5ccdd (Michael Pilquist, 7 months ago)
   Fixed failures in TCP tests caused by assumption of early finalizer
   execution

17b97873 (Michael Pilquist, 7 months ago)
   Removed Algebra.Release

3e41d078 (Brian P. Holt, 7 months ago)
   add FunctorFilter instances for Stream and Chunk

5dea53b6 (Michael Pilquist, 7 months ago)
   Adjusted unmanagedSourceDirectories to match naming convention from cats

6aeff3ac (Michael Pilquist, 7 months ago)
   Updated Chunk.indexedSeq to work with both mutable and immutable
   IndexedSeqs on 2.13

9ad0d7fb (Jakub Kozłowski, 7 months ago)
   `to` -> `through`, `class` -> `object` in doc

273ca0e8 (Michael Pilquist, 7 months ago)
   Skip some of the Chunk.apply optimization checks on Scala.js

df9c73d7 (Michael Pilquist, 7 months ago)
   Added comment explaining the fallback assertion in ChunkSpec

1435d582 (Michael Pilquist, 7 months ago)
   Made Chunk.apply optimizations work for 2.13 and biased Chunk constructors
   more to singleton chunks

2d5eb631 (Michael Pilquist, 7 months ago)
   Upgraded to latest scala-collection-compat

e60cc82e (Michael Pilquist, 7 months ago)
   Added a special case for calling Chunk.seq with a buffer

0d0c8363 (Michael Pilquist, 7 months ago)
   Fixed ScalaDoc on Chunk.seq

8867a98a (Michael Pilquist, 7 months ago)
   Removed trailing comma to fix 2.11 build

a9b14a3a (Michael Pilquist, 7 months ago)
   Fixed #1411 - updated Chunk to use scala.collection.Seq

fa6dd74a (Michael Pilquist, 7 months ago)
   Added specialized chunk generators

e44d6a91 (Michael Pilquist, 7 months ago)
   Moved Chunk generators to standalone trait and increased coverage of
   constructors

992cf606 (Scala steward, 7 months ago)
   Update cats-core, cats-laws to 1.6.0

978aae58 (Michael Pilquist, 7 months ago)
   Fixed test compilation failures due to ScalaTest renames

1a52d2d8 (Michael Pilquist, 7 months ago)
   Upgraded to latest scalatest

d5d43744 (Michael Pilquist, 7 months ago)
   Updated versions in README

b111c8f3 (Sebastian Bruckner, 7 months ago)
   Fix groupWithin to not emit empty groups, resolves #1404

fa7efa5b (Tamer Abdulradi, 7 months ago)
   Add Tamer Abdulradi as contributer

5985e713 (Tamer Abdulradi, 7 months ago)
   Add scanMap and scanMonoid

9b0e4720 (Michael Pilquist, 7 months ago)
   Setting version to 1.0.4-SNAPSHOT

179728e5 (Michael Pilquist, 7 months ago)
   Setting version to 1.0.3

1273cc5e (Michael Pilquist, 7 months ago)
   Corrected minor doc typo

7a387c85 (Fabio Labella, 8 months ago)
   Add holdResource and holdOptionResource

45d02c9e (Fabio Labella, 8 months ago)
   Add scaladoc for resource compilation

3165714a (Fabio Labella, 8 months ago)
   Change syntax to `compile.resource`

84258555 (Fabio Labella, 8 months ago)
   Add explanation for CompileScope.release

a64640ee (Fabio Labella, 8 months ago)
   Refactor tests

52a310c9 (amarrella, 8 months ago)
   Better import for other concurrent data types

00b4e2bd (amarrella, 8 months ago)
   Updated fix to fix Chunk bug

e80ab721 (Scala steward, 8 months ago)
   Update cats-effect, cats-effect-laws to 1.2.0

af9b00ed (Diego E. Alonso-Blas, 8 months ago)
   Small improvement touches to the debounce method.

   We make some small improvements to the debounce method:
   - We fold the flatMap on chunks into the evalMap below, saving some
    inner stream allocations
   - We extend the API documentation for this use case.

ad972461 (Scala steward, 8 months ago)
   Update scodec-bits to 1.1.9

e9a12402 (Scala steward, 8 months ago)
   Update scodec-bits to 1.1.8

dc637086 (Tim McIver, 8 months ago)
   Fix typo.

06685c61 (erlangxk, 8 months ago)
   create Text message with different milliseconds

3245cbad (Diego E. Alonso-Blas, 8 months ago)
   Little optimisation for Pull.last

   The last element of a pull should be the last element of the current chunk.
   No need to loop through all elements in the chunk.

860b97c3 (Michal, 8 months ago)
   fix readme for correct library versions in example

7634ef8b (Fabio Labella, 8 months ago)
   Release resources in parent scope through scope.close

b0c1fbc7 (Fabio Labella, 8 months ago)
   Run outer scope in Resource

8e359be7 (Fabio Labella, 8 months ago)
   Move top level compile loop into Compiler

fcf1e13c (Fabio Labella, 8 months ago)
   Pure constructor for CompileScope

c5f4c648 (Fabio Labella, 8 months ago)
   Add spec for compile to resource behaviour

1531cc85 (Fabio Labella, 8 months ago)
   Syntactic infrastucture for Resource compilation

d26d3f64 (Michael Pilquist, 8 months ago)
   Fixed 2.11 and 2.12 build due to merge

02a8344a (Michael Pilquist, 8 months ago)
   Removed Sync[F] constraint on gzip/gunzup

710f456c (Diego E. Alonso-Blas, 8 months ago)
   Remove third type parameter from Algebra.raiseError

4899d34e (Diego E. Alonso-Blas, 8 months ago)
   Remove Type Parameters in FreeC and subclasses.

   Using the fact that `FreeC[F[_], +R]` is covariant in O, we modify its
   sub-classes `Fail` and `Interrupted`, so as to set `R` to INothing, and
   then we can remove the type parameter from the class. We apply the same
   change to some of the factory methods.

   We also can remove the covery method.

23926716 (Michael Pilquist, 8 months ago)
   Deprecated Sink in favor of Pipe

a69a6283 (Michael Pilquist, 8 months ago)
   Added a mima exclusion for compiler#apply change

dc5e3074 (Scala steward, 8 months ago)
   Update sbt-release to 1.0.11

0807483b (Diego E. Alonso-Blas, 8 months ago)
   Remove MonadInstance object

   The `Result.monadInstance` non-implicit val was only used at one place, for
   the `map` function (through `flatMap`), explicitly. We remove that instance
   and add a function `map` on result, directly.

c6b7e15c (Michael Pilquist, 8 months ago)
   Replaced by-name with explicit thunk in compiler

601ba5c3 (Michael Pilquist, 8 months ago)
   Added s.compile.string syntax for efficiently compiling to a string

105e642d (Diego E. Alonso-Blas, 8 months ago)
   Remove unneeded changes

   We remove some of the changes first introduced, so as to reduce diff noise.

c858d4a6 (Michael Pilquist, 8 months ago)
   Removed use of cats.data.Eval from stream compilation

a94e0b99 (Michael Pilquist, 8 months ago)
   Fixed #1371 - io.file.writeAll now respects StandardOpenOption.APPEND

898bef3f (Michael Pilquist, 8 months ago)
   Default sbt to 2.12 build

179810ca (Michael Pilquist, 8 months ago)
   Removed extraneous trailing comma in build.sbt

17a1fd52 (Michael Pilquist, 8 months ago)
   Add 2.13.0-M5 support (#1378)

   * Minimal upgrade to 2.13.0-M5
   
   * Applied workaround for scala/bug 11303 from @joroKr21
   
   * Tests passing on 2.13
   
   * Restored Chunk.seq optimization
   
   * Fixed scala-collection-compat dependency for JS
   
   * Fixed benchmark compilation
   
   * Fixed cross compilation of coreJVM
   
   * Fixed compilation issue in reactive streams
   
   * Fixed mima on 2.13 and compilation on 2.11
   
   * Fixed travis failure
   
   * Fixed bad merge in Chunk

a298df0e (Diego E. Alonso-Blas, 8 months ago)
   Remove Type Parameter

   Since the `Step` case class is fully private to the Algebra, one of its
   type parameters can be made into "INothing".

a733166d (Diego E. Alonso-Blas, 8 months ago)
   Fix typos and wrong links noted in code review.

25456e4e (Diego E. Alonso-Blas, 8 months ago)
   Ecosystem section

   We introduce a special Microsite page for the library "ecosystem", which is
   to say the libraries on which FS2 is built upon, and the libraries built on
   top of FS2.

e04ac498 (Diego E. Alonso-Blas, 9 months ago)
   References Microsite Page.

   Adds an overview paragraph at the top of the Rreadme. Adds to the microsite
   a section for talks, blog posts, related libraries and othere reference
   material on FS2.

08faa411 (Gabriel Volpe, 9 months ago)
   Moving CNAME into standard tut directory

4e1a2f2e (Gabriel Volpe, 9 months ago)
   Removing specification of custom tut dir.

00b47990 (Gabriel Volpe, 9 months ago)
   Adding new logo to microsite. Re-organizing tut docs.

8025171e (Diego E. Alonso-Blas, 9 months ago)
   Some more operations can be private too

71b78df1 (Diego E. Alonso-Blas, 9 months ago)
   Make all Algebra subclasses private.

   This PR modifies the Algebra object, to hide all the sub-classes of that
   algebra. Thus, the implementation are not visible outside it, and only in
   the algebra are those constructed/deconstructed.

   To allow this, we cut from the implementation of `Pull.mapOutput` the part
   that  generates a natural transformation, to a Algebra.mapOutput operation.

bfd15970 (Diego E. Alonso-Blas, 9 months ago)

   The methods translate0 and translateStep were introduced in previous 
   versions, to deal with different variants of the translations.
   - translate0 was introduced to remove translation from the Algebra,
   - translateStep was introduced to support the new compiler-interpreter.

   Looking at them, the differences between these methods were:
   - That translate0 was not parameterised on the output type X.
   - That, for the algebra operation of output, the translate0 had a
    recursive call to translate0, whereas the translateStep method
    was performing a Cast to a result of the parametric type X.
   - That `translate0` first recurs with itself, until it resorts to
    `translateStep`, as in some layered approach.

   Thus, we have unified these methods, by replacing the initial call to
   `translate0` as a call to `translateStep` with `X = O`. We add a boolean
   flag, `isMainLevel`, to indicate if it is the main level
   (formerly `translate0`) or the lower levels (`translateStep`).

   We also wrap the `translateStep` inside a `translateAux` function, which is
   the main entry method of the recursive loop. Since neither the `fK` nor the
   `Concurrent[G]` change in the recursive calls of `translateStep`, we use a
   pattern useful in recursive functions, of not passing in the recursive
   calls the parameters, either type or value, that never change.

d2aa865a (Yannick Heiber, 9 months ago)
   Add repeatN combinator to Stream and a spec for repeat and repeatN

69e2297c (Michael Pilquist, 9 months ago)
   Removed dead code path in Stream.bracket

ad816ede (amarrella, 9 months ago)
   StreamApp class minimized too

af14d3ed (amarrella, 9 months ago)
   Reduced scalafix impact on streamapp object

71e3d8d7 (amarrella, 9 months ago)
   Smaller diff in replacing StreamApp

aca29217 (amarrella, 9 months ago)
   Updated scalafix version

e666c6ad (Fabio Labella, 9 months ago)
   Add MiMa exclusions for Pipe.joinQueued

5f7832f7 (Joe Ferris, 9 months ago)
   Add recommendations for blockingExecutionContext

   * Document that the io implementations do, in fact, block
   * Recommend an unbounded thread pool
   * Reference Cats Effect docs on blocking threads

72fd820d (fabio-labella, 9 months ago)
   Express Pipe.join with broadcast

4781e6f2 (Scala steward, 9 months ago)
   Update tut-plugin to 0.6.10

5c514193 (Michael Pilquist, 9 months ago)
   Setting version to 1.0.3-SNAPSHOT

aa49f808 (Michael Pilquist, 9 months ago)
   Setting version to 1.0.2

95ec9a9f (Michael Pilquist, 9 months ago)
   Fixed #1357

010d1aef (Michael Pilquist, 9 months ago)
   Helping 2.13.0-M5 type inference along

172e5079 (Michael Pilquist, 9 months ago)
   Updated README for 1.0.1 release

cf6f153f (Michael Pilquist, 9 months ago)
   Setting version to 1.0.2-SNAPSHOT

9df57ae4 (Michael Pilquist, 9 months ago)
   Setting version to 1.0.1

942294e6 (Michael Pilquist, 9 months ago)
   Setting version to 1.0.2-SNAPSHOT

0abc2a6f (Michael Pilquist, 9 months ago)
   Setting version to 1.0.1

521361e7 (Michael Pilquist, 9 months ago)
   Setting version to 1.0.2-SNAPSHOT

bbe6fcdd (Michael Pilquist, 9 months ago)
   Setting version to 1.0.1

89a90ce0 (Michael Pilquist, 9 months ago)
   Adjusted mima exclusions

74422317 (Michael Pilquist, 9 months ago)
   Changed io.{udp,tcp} to require a ContextShift[F] to ensure shifting occurs
   after callbacks

c9823149 (Diego E. Alonso-Blas, 9 months ago)
   Remove the `O` parameter from CompileScope.

   This commit removes the type parameter `O`, which is usually associated to
   the output of a `Stream[F, O]`, from several internal classes and types,
   such as `CompileScope`, its internal `State`, the `InterruptContext, and
   the algebra operation `GetScope`. We also remove it from the definitions of
   several auxiliary methods.

   It turns out that `CompileScope` is not really concerned itself with the
   emission of values of type `O`, but merely with the correct handling of the
   shared resources in a common scope. Removing the parameter `O` helps make
   this clearer.

8e7c8416 (Michael Pilquist, 9 months ago)
   Upgraded to cats-effect 1.1.0

b3f352d6 (Diego E. Alonso-Blas, 10 months ago)
   Changes and Syntactic Sugaring in the Internals

   In the Algebra GADT, we make the type parameter `O` (which is the return
   type of a `Stream` covariant. This allows us to define the sub-trait
   `AlgEffect` as Algebra operations with `O=INothing`, in other words,
   operations that are not concerned with output. As a consequence, for those
   operations we can remove the middle type parameter.

a3177936 (Scala steward, 9 months ago)
   Update cats-core, cats-laws to 1.5.0

a51778fe (Michael Pilquist, 9 months ago)
   Bumped to latest sbt version

2a820084 (Michael Pilquist, 9 months ago)
   Removed parallel collections from build

6decf700 (Michael Pilquist, 9 months ago)
   Updated travis to openjdk11

15694d36 (Michael Pilquist, 9 months ago)
   Removed dependency on parallel collections

5fc4d10c (Scala steward, 9 months ago)
   Update tut-plugin to 0.6.9

73cb339c (Scala steward, 9 months ago)
   Update sbt-scalajs, scalajs-compiler, scalajs-library... to 0.6.26

b1e9c045 (Scala steward, 9 months ago)
   Update sbt-release to 1.0.10

0e9ad7ab (Scala steward, 9 months ago)
   Update sbt-jmh, sbt-jmh-extras to 0.3.4

f6c8bd3b (Scala steward, 9 months ago)
   Update kind-projector to 0.9.9

9aeeb28e (Scala steward, 9 months ago)
   Update scodec-bits to 1.1.7

0b4582fc (Scala steward, 9 months ago)
   Update sbt-osgi to 0.9.4

1c2d8c92 (Scala steward, 9 months ago)
   Update sbt-mima-plugin to 0.3.0

9e51c89e (Michael Pilquist, 9 months ago)
   Fixed some exceptions in JMH benchmarks

593a54e3 (Michael Pilquist, 9 months ago)
   Adjusted fix for 1335 to be more universal - translate on FreeC now
   suspends on Eval nodes

a66f7672 (Michael Pilquist, 10 months ago)
   Fixed 2.11 build

b0d9b87d (Michael Pilquist, 10 months ago)
   Fix #1335 - stack safety of map

605933d6 (Michael Pilquist, 10 months ago)
   Auto-formatted ScalaDoc changes

be3cd631 (Daniel Spiewak, 10 months ago)
   Fixed binary compatibility issues

147a429d (Daniel Spiewak, 10 months ago)
   Fixed formatting issues from series/1.0

ac559014 (Daniel Spiewak, 10 months ago)
   Copied over (mostly) implementation from fs2-gzip

1de2c5d6 (Michael Pilquist, 10 months ago)
   Inlined a few functions

230b1a81 (Michael Pilquist, 10 months ago)
   Replaced an occurrence of *> with >>

b0fddd21 (Michael Pilquist, 10 months ago)
   Fixed compilation issue in StreamSubscriber

2f9b2443 (Diego E. Alonso-Blas, 10 months ago)
   Attend some code review comments.

95044827 (Michael Pilquist, 10 months ago)
   Cleaned up the implementation of JavaInputOutputStream

f60f1241 (Michael Pilquist, 10 months ago)
   Fix build compilation issue

d6cfba19 (fabio-labella, 10 months ago)
   Replace *> with >> to avoid memory issues

58ca5585 (Diego E. Alonso-Blas, 10 months ago)
   MiMa: add filter for EitherSyntax

   The MigrationManager plugin is detecting that the `EitherSyntax` is 
   missing, so we add a filter to indicate it is safe. Note that this class
   was private to `fs2`, and therefore should not be part of any client using
   the library.

f88e8f4a (Luciano, 10 months ago)
   Link three new projects in README

   Added fs2-columns, mongosaur and scarctic to "Projects using FS2"
54ffaf06 (Diego E. Alonso-Blas, 10 months ago)
   Replace .swap.right with .left, as suggested

aa2c0959 (Diego E. Alonso-Blas, 10 months ago)
   Remove the EitherSyntax class

   We remove the EitherSyntax implicit class, used to provide a right-biased
   flatMap for `Either`. Although fs2 still supports the Scala 2.11, in truth
   the special right-biased flatMap is used nowhere in `fs2`. The only method
   being used was the `toOption`; and since that was followed with
   `Option.get`, we can just use `.right.get` instead by way of the
   Either.RightProjection class.

7435505e (Diego E. Alonso Blas, 10 months ago)
   Update Algebra.scala

81d8c154 (Diego E. Alonso-Blas, 10 months ago)
   Internal changes to the implementation of `translate`.

   We make some changes to the internal Stream implementation, in the Algebra
   file, to avoid using implicit classes.

   - We remove the syntax extension implicit class `AlgScopeSyntax`,
    which was not being used anywhere.
   - We remove the syntax extension `AlgEffectSyntax`, and extract its
    only method as an internal private function.
    We replace the only two uses needed for this function.

   After removing the `AlgScopeSyntax` syntax extension, it turns out that
   there is no place in the code in which the `AlgScope` sub-trait of
   `Algebra` is used, so we can remove that one as well.

97f59c32 (Michael Pilquist, 10 months ago)
   Made ConcurrentEffect changes to tcp/udp support binary compatible with 1.0

d57a00d7 (Diego E. Alonso-Blas, 10 months ago)
   Expands the ScalaDocs for the Stream class.

   We extend the ScalaDocs description of the Stream class.
   - We adds some descriptions for the notions of pull-based evaluation, and
   laziness. We show some special cases of Streams.
   - We add a section describing the equivalence between some of the methods
   of the Stream class, and similar methods in the List class. This
   equivalences hold true for finite and pure streams.
   - We add into the docs of some of methods a description of how the methods
   handles non-terminating streams.

80ed4d1a (Jakub Kozłowski, 10 months ago)
   Reorganize Converter

8339b131 (Jakub Kozłowski, 10 months ago)
   Use Stream.resource, update ReadmeExample

24331424 (Jakub Kozłowski, 10 months ago)
   Fix thread pool leak in README

1e028527 (Michael Pilquist, 10 months ago)
   Guarantee yieldBack

29977970 (Michael Pilquist, 10 months ago)
   Changed yield back strategy to use Concurrent only

e93d1dd8 (Michael Pilquist, 10 months ago)
   Removed ConcurrentEffect from tcp/udp packages

3e7b42a3 (amarrella, 10 months ago)
   Removed unnecessary edit

fd858661 (amarrella, 10 months ago)
   Revert "Delete .scalafmt.conf"

   This reverts commit 4ac431e33e9ac574130d53f3c8c20d471f27acfd.

4ac431e3 (Alessandro Marrella, 10 months ago)
   Delete .scalafmt.conf

c1dd4c08 (amarrella, 10 months ago)
   Added usability renames

209cea33 (amarrella, 10 months ago)
   Remove fs2.async imports

648f8e6d (amarrella, 10 months ago)
   Added chunk rules

672ea4f9 (amarrella, 10 months ago)
   only replace 'stream' method

04519d71 (amarrella, 10 months ago)
   fixed compileerror

8d555f00 (amarrella, 10 months ago)
   using less .get

3840206a (amarrella, 10 months ago)
   renamed file

500a8432 (amarrella, 10 months ago)
   renamed rule

a1d35e2a (amarrella, 10 months ago)
   renamed rule file

c71c8786 (amarrella, 10 months ago)
   Moved all rules to one file

b7b6068b (amarrella, 10 months ago)
   Refactored scalafix to be github based

55ae424e (amarrella, 10 months ago)
   removed TODO :)

9ec7fa49 (amarrella, 10 months ago)
   Queue and Topic

ad877950 (amarrella, 10 months ago)
   Signal

c82a68b3 (amarrella, 10 months ago)
   semaphore fix

223a8d13 (amarrella, 10 months ago)
   Semaphore

059c733a (amarrella, 10 months ago)
   Refactored concurrent rules and did deferred

e179b14b (amarrella, 10 months ago)
   Concurrent rules, no imports

a9cefb4d (amarrella, 10 months ago)
   Concurrent (Ref) renames

ba16b95b (amarrella, 10 months ago)
   Fixed imports

dd000463 (amarrella, 10 months ago)
   Using actual effect in delayCancellable

37f37a4e (amarrella, 10 months ago)
   scheduler done

cc1b6d13 (amarrella, 10 months ago)
   Fixed bracket rules and refactored unapply -> apply

f3d65a01 (amarrella, 10 months ago)
   Started scheduler and broken bracket (for good reasons)

d616c530 (amarrella, 10 months ago)
   Added bracket

11d5e83f (amarrella, 10 months ago)
   fixed build stuff

e37698fe (amarrella, 10 months ago)
   Refactored to match fs2 build

2958da41 (amarrella, 10 months ago)
   instead of deleting object, refactor

874997d1 (amarrella, 10 months ago)
   Effect in body is replaced

dda842fa (amarrella, 10 months ago)
   WIP smarter logic

550e6909 (amarrella, 10 months ago)
   StreamApp scalafix

66d7935d (Christopher Davenport, 11 months ago)
   Add isOpen to Socket Interface


kubukoz:
Oops, `hub` failed me. Sorry about that.

PR #1598, opened 2019-09-07

diesalbla:
This Pull Requests unifies the two major ADT at the internal implementation of the Stream: `FreeC[F[_], R]`, the Free monad with catch and interruption; and `Algebra[F[_], O, R]`, a representation of the instructions in that FreeC program. These two ADT were combined to form a stream, using the approximate type alias of `PullF, O, R]` as `FreeC[Algebra[F, O, ?], Unit]`. Notably, the only way in which an object of type `Algebra[F, O, R]` could appear in a FreeC was through a special sub-class of `FreeC`, that is `Eval`. From this, we unify these two ADTs by doing the following:

- We add the output parameter O, representing the output data of a stream, as an extra covariant parameter to the `FreeC` ADT, and its subtracts. 
- We turn the Eval[F[_], A] into an abstract class, which is equivalent to the old `Algebra` trait. We therefore turn all the sub-classes of Algebra into subclasses of `FreeC.Eval`. Because FreeC and Algebra classes are in different files, we need to unseal it. This could be solved by merging the two files, but that would be a bigger change.
- We modify the `ViewL` trait (used for loop unrolling) by also adding the extra third parameter, and by changing some of its subclasses from using a generic `F[R]` to having an `Eval[F, R]`. Like the `FreeC`, the previous `ViewL[F, R]` always had that F to an `Algebra`. 

As a result, a Pull is no longer represented as a FreeC[Algebra[F, O, ?], Unit], but as a FreeC[F, O, R] instead. 

With the previous 2-layered representation, there was a need to have in the `Algebra` object a few methods to lift simple data and fill the type parameters. With the new representations, those methods are not needed and we can remove them in favour of the object they build. That is the case of the private methods `openScope` and `step`.  

#### Translate to mapOutput

Before this change, the implementation of Stream[F, A].map[B](f: A => B) had to work at two levels: one being a `FunctionK` between two type-instances of Algebra, which would differ on the output type. To apply this transformation through the Stream, the `FreeC` ADT had a `translate` method. After the changes, now we do not need that `FunctionK` wrapper, and we can just replace the translate method of FreeC with a mapOutput method.

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/FreeCBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/FreeCBenchmark.scala
@@ -16,7 +16,7 @@ class FreeCBenchmark {
   @Benchmark
   def nestedMaps = {
     val nestedMapsFreeC =
-      (0 to N).foldLeft(FreeC.pure[IO, Int](0): FreeC[IO, Int]) { (acc, i) =>
+      (0 to N).foldLeft(FreeC.pure[IO, Int](0): FreeC[IO, INothing, Int]) { (acc, i) =>
         acc.map(_ + i)
       }
     run(nestedMapsFreeC)
@@ -25,21 +25,20 @@ class FreeCBenchmark {
   @Benchmark
   def nestedFlatMaps = {
     val nestedFlatMapsFreeC =
-      (0 to N).foldLeft(FreeC.pure[IO, Int](0): FreeC[IO, Int]) { (acc, i) =>
+      (0 to N).foldLeft(FreeC.pure[IO, Int](0): FreeC[IO, INothing, Int]) { (acc, i) =>
         acc.flatMap(j => FreeC.pure(i + j))
       }
     run(nestedFlatMapsFreeC)
   }
 
-  private def run[F[_], R](self: FreeC[F, R])(implicit F: MonadError[F, Throwable]): F[Option[R]] =
+  private def run[F[_], O, R](
+      self: FreeC[F, O, R]
+  )(implicit F: MonadError[F, Throwable]): F[Option[R]] =
     self.viewL match {
       case Result.Pure(r)             => F.pure(Some(r))
       case Result.Fail(e)             => F.raiseError(e)
       case Result.Interrupted(_, err) => err.fold[F[Option[R]]](F.pure(None)) { F.raiseError }
-      case v @ ViewL.View(step) =>
-        F.flatMap(F.attempt(step)) { r =>
-          run(v.next(Result.fromEither(r)))
-        }
+      case v @ ViewL.View(_)          => F.raiseError(new RuntimeException("Never get here)"))
     }
 
 }
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -22,8 +22,7 @@ import fs2.internal._
   * `raiseError` is caught by `handleErrorWith`:
   *   - `handleErrorWith(raiseError(e))(f) == f(e)`
   */
-final class Pull[+F[_], +O, +R] private (private val free: FreeC[Nothing, Nothing, R])
-    extends AnyVal {
+final class Pull[+F[_], +O, +R] private (private val free: FreeC[Nothing, O, R]) extends AnyVal {
 
   private[fs2] def get[F2[x] >: F[x], O2 >: O, R2 >: R]: FreeC[F2, O2, R2] =
     free.asInstanceOf[FreeC[F2, O2, R2]]
@@ -89,7 +88,7 @@ final class Pull[+F[_], +O, +R] private (private val free: FreeC[Nothing, Nothin
 object Pull extends PullLowPriority {
 
   @inline private[fs2] def fromFreeC[F[_], O, R](free: FreeC[F, O, R]): Pull[F, O, R] =
-    new Pull(free.asInstanceOf[FreeC[Nothing, Nothing, R]])
+    new Pull(free.asInstanceOf[FreeC[Nothing, O, R]])
 
   /** Result of `acquireCancellable`. */
   sealed abstract class Cancellable[+F[_], +R] {
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -22,11 +22,11 @@ import fs2.internal._
   * `raiseError` is caught by `handleErrorWith`:
   *   - `handleErrorWith(raiseError(e))(f) == f(e)`
   */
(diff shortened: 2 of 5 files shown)


diesalbla:
@mpilquist If I recall correctly, from the [slides](https://speakerdeck.com/mpilquist/fs3-evolving-a-streaming-platform?slide=14) of your [talk](https://www.youtube.com/watch?v=mStYwML3JZk) in the Northeast Scala symposium, one major goal was to "Improve readability of internals by replacing FreeC+Algebra with a more direct encoding". I humbly believe that the unification done in this PR, while it may not be the end-destination, could be a small step towards that.

pchlupacek:
Not sure about the views of others, but this seems to me like marriage between two distinct purpose algebras, and as a readability is concerned, I am not sure if this makes an improvement. 

I liked the FreeC to be responsible for generic program flow, whereas the Algebra for interpretation logic. 

@diesalbla Is there any noticeable performance improvement? I bet we may save some allocation in certain cases....

mpilquist:
I'd love to see some benchmarks if possible. I agree that the nice separation of `FreeC` and `Algebra` is gone with this approach, making the code less modular, though the FS3 branch has that same issue and I made that point in the talk at NEScala.

Curious if we could go a step further here and unify `Algebra` with `Pull`, leaving just `Pull` as the main engine and `Stream` as a newtype over it.

pchlupacek:
@mpilquist @diesalbla I like the idea of getting rid of Pull. Perhaps Pull can replace either Of (FreeC/Algebra) and by doing so perhaps we emerge in something there.

diesalbla:
> I liked the FreeC to be responsible for generic program flow, whereas the Algebra for interpretation logic.

Note that, as a matter of code, you still have all of that generic control-flow in the `FreeC.scala` file, in the `FreeC` abstract class and the `FreeC` object. I am assuming that by "generic control flow" you mean the methods on the `FreeC` abstract class (`flatMap`, `transformWith`, `asHandler`, etc); as well as the the unrolling/tree-rotation operation of `ViewL.apply`, to turn left-handled bind-on-bind associations into right-handed bind-on-bind associations. Note also that all of this generic control-flow operations is still implemented without any reference to the sub-classes of the `Algebra` object.

True, the `FreeC` file in itself would no longer define a usable generic free monad with errors and interruption; but I cannot see that as a loss, since `FreeC` was never exported. 

> I agree that the nice separation of FreeC and Algebra is gone with this approach, making the code less modular

Well, you are still keeping them separate, albeit in a different way: whereas it currently goes through the type parameter `F[_]` of `FreeC`, and through a field of `FreeC.Eval`, now it goes through subtyping of an open abstract type `FreeC.Eval`. 

Note also that, although `FreeC` was defined without without reference to `Algebra`, the  `Algebra` depends on `FreeC` (through the `Step` class), so the existing modularity is was one-sided. The new structure would keep this: `FreeC` object know nothing of the classes in the `Algebra` object. The one notable change to the `FreeC` ADT, from before, is that it now has an extra "passenger" parameter, the `O`, which is just nullified in the `Result` cases. 

> and as a readability is concerned, I am not sure if this makes an improvement

I would say that having no nested types, no kind-projection `?` symbol, no need for "smart constructors", no need to use a `FunctionK` for mapping the stream outputs, and the extra consistency of `F[_]` now meaning the stream's effect type everywhere, are a few upsides.  Against that, I can see the downsides of adding an extra type-parameter `O` to the `FreeC` ADT, and of replacing in some places .

Note also that avoiding kind-projection and nested types is not just, IMHO, easier to read for people. It may also to read and reason about for the Scala compiler.

There is also the slight negative in `FreeC` no longer being a `sealed` abstract class. That could  be fixed by merging the two files, at the cost of losing the modular separation recalled above. 

> Curious if we could go a step further here and unify Algebra with Pull, leaving just Pull as the main engine and Stream as a newtype over it.

I am not sure if I understand what you mean. I thought of renaming `FreeC` to `Pull`, given that it now has the same type parameters. However, there is still a difference: the `Pull` is covariant in its first parameter while `FreeC` is invariant. When I changed `FreeC` to be covariant (that is `FreeC[+F[_], +O, +R]` there were some compile errors. I also made a few tries, in both `Stream` and `Pull`, to see if we could avoid the use of `FreeC[Nothing, Nothing, X]`, and the need for conversions `get`/`fromFreeC` in both Stream and Pull, but there were a couple compile errors. By that time, since this PR was already getting large enough (hundred-lines), I hoped to delay that matter for a subsequent PR.

> I'd love to see some benchmarks if possible.  

Yeah, I was intending to try a few. I once runned the `StreamBenchmarks`, but they seem to take a bit long to run and wanted to find a little bit more about benchmarking first.

diesalbla:
> Is there any noticeable performance improvement? I bet we may save some allocation in certain cases....

Well, I think that should also be an upside of this change: now that every _algebra_ instruction is a FreeC itself, we no longer need to wrap them in the `FreeC.eval` as we were doing before. Thus, any different single instruction, such as getting a scope, opening a scope, or outputting some values, will now be represented with one object in the `FreeC` instead of two, because they _are_ now `FreeC` objects. 

Looking for benchmarks to confirm, though...

diesalbla:
I have been running some of the **ConcurrentBenchmarks.**  For brevity, I removed some of the numbers generated and left 1,4,16,64,256.

- JMH command: `jmh:run -i 10 -wi 10 -f2 -t4 .*ConcurrentBenchmark*. -prof gc`
- JVM: Java HotSpot(TM) 64-Bit Server VM (build 12.0.2+10, mixed mode, sharing)

The results, detailed below, indicate a reduction of the `gc.alloc.rate.norm` of about 2.5% (for the small number) to 1.5 % (for N=256).

[ConcurrentBenchmark.ods.zip](https://github.com/functional-streams-for-scala/fs2/files/3589770/ConcurrentBenchmark.ods.zip)

mpilquist:
> However, there is still a difference: the Pull is covariant in its first parameter while FreeC is invariant. When I changed FreeC to be covariant (that is FreeC[+F[_], +O, +R] there were some compile errors.

Yeah, this is what I meant -- making `FreeC` covariant and moving it in to `Pull` directly.

diesalbla:
Well, I have been able to re-make the internal `FreeC[Nothing, Nothing, R]` (resp. `FreeC[Nothing, Nothing, Unit]`) of Pull (resp Stream) into a `FreeC[Nothing, O, R]` (resp Unit), so that gets it a little bit closer than it was.

diesalbla:
@mpilquist I have started some attempts to do that unification of Pull and FreeC, on this PR here (on top of this branch). At present, I am still getting a few compile errors, and _compiler_ errors as well. https://github.com/diesalbla/fs2/pull/1

PR #1603, opened 2019-09-08
Removed Stream.bracketCancellable and Pull.acquire*

mpilquist:
`Stream.bracketCancellable` and `Stream.bracketCaseCancellable` allowed early, manual finalization of resources. Upon invoking the finalizer, the internal resource would be cleaned up, ensuring we don't leak references to already invoked finalizers. Now that we introduce a scope for each `bracket`, folks would expect that invoking cancelation of this case also closes the dedicated scope -- i.e., we don't leak scopes either. Instead of implementing this, I'd like to remove these operations instead. There were no tests for these operations and they were very rarely used. Searching GitHub shows no GH hosted open source projects use either operation.

Similarly, `Pull.acquireCancellable` provided a mechanism to manually invoke finalizers. I added this back in #662 to fix resource leaks, but the motivating use cases at the time are all addressed now by using effect level bracketing.

Removal of `Pull.acquire` is perhaps the most controversial. When `Pull` was the primary mechanism in which scopes were created, acquiring a resource via `Pull` made more sense than it does now.

This PR also introduces a `s.pull.headOrError` method, that converts a `Stream[F, O]` to a `Pull[F, INothing, O]`. Uses of `Pull.acquire` can be replaced with `Stream.bracket(acquire)(release).pull.headOrError`.

Note: the only place where `Pull.Cancellable` was used was in the constructors for `FileHandle` in `fs2.io.file.pulls`. These were vestiges of a time before `cats.effect.Resource`. These constructors were removed in favor of new constructors on `FileHandle` companion that return `Resource[F, FileHandle[F]]`.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1912,7 +1912,7 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Algebra[Nothing, 
       f: ExitCase[Throwable] => F2[Unit]
   )(implicit F2: Applicative[F2]): Stream[F2, O] =
     Stream.fromFreeC(
-      Algebra.acquire[F2, O, Unit](().pure[F2], (_, ec) => f(ec)).flatMap(get[F2, O])
+      Algebra.acquire[F2, O, Unit](().pure[F2], (_, ec) => f(ec)).flatMap(_ => get[F2, O])
     )
 
   /**
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3965,6 +3965,17 @@ object Stream extends StreamLowPriority {
     }
 
     /**
+      * Takes the first value output by this stream and returns it in the result of a pull.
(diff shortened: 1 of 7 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4042,6 +3995,17 @@ object Stream extends StreamLowPriority {
       go(init, self)
     }
 
+    /**
+      * Takes the first value output by this stream and returns it in the result of a pull.
+      * If no value is output before the stream terminates, the pull is failed with a `NoSuchElementException`.
+      * If more than 1 value is output, everything beyond the first is ignored.
+      */
```
SystemFw: I wonder if there is any particular reason for the inconsistency with `compile.lastOrError`

mpilquist: What's the inconsistency, just the name? I'll rename to `headOrError`.

SystemFw: Just the `last` vs `first` behaviour. I can see how `Pull` is more `first`-y (`uncons`), but it's inconsistent with `lastOrError`. Not a big deal though

PR #1608, opened 2019-09-10
Stream: reimplement range without fold

diesalbla:
The `Stream.range` operation, for creating a stream of integers from
an initial value to a final value in a step-by-step way, was
implemented using `unfold`. An `unfold` is too general and too lazy for what we need to output integer values (which has no side effects). Thus, this commit makes a lighter implementation of `Stream.range`, by unfolding the `unfold` call and removing the extra `Option` and `Tuple` objects.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3408,13 +3408,14 @@ object Stream extends StreamLowPriority {
     * res0: List[Int] = List(10, 12, 14, 16, 18)
     * }}}
     */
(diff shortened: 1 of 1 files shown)


SystemFw:
This is a micro optimisation imho. I don't really see any problems with the current implementation.

> An unfold is too general and too lazy for what we need to output integer values (which has no side effects).

`unfold` is not about side effects and it's _less_ general than recursion, recursion is the maximally general strategy you can use

diesalbla:
Perhaps I should rephrase: 
- The use of a `(current-next seed)` tuple, as done in `unfold`, is a bit more general than what is needed here, where `next` is just `current + by`. 

I cannot see the new code being much more complex than the previous one, based on fold, but I suppose I can run some benchmarks to see how effective it is, anyway.

PR #1610, opened 2019-09-10

diesalbla:
Continuation of https://github.com/functional-streams-for-scala/fs2/pull/1598, which was automatically closed by Github  due to the deletion of the `series/1.1` branch.

------------------

This unifies the two ADT in the internal stream implementation: `FreeC[F[_], R]`, the Free monad with catch and interruption; and `Algebra[F[_], O, R]`, for the scope-handling and output instructions. These are combined using the approximate type aliae of `PullF, O, R]` as `FreeC[Algebra[F, O, ?], R]`. Notably, an object of type `Algebra[F, O, R]` can only appear in a FreeC through the `FreeC.Eval` class. This allows us to unify these two ADTs by doing the following:

- Add an output parameter `O` (stream output) to the `FreeC` ADT, and its subclasses; as well as to the `ViewL` trait (for unrolling).
- Turn `FreeC.Eval` into an abstract class that replaces the previous `Algebra` trait, so that all subclasses of Algebra become subclasses of `FreeC.Eval`. Because FreeC and Algebra ADTs are in separate files, we unseal it. 
- We replace the `ViewL` subclasses from containing a `F[R]` to containing an `Eval[F, R]`. 

As a result, a Pull is no longer an alias for `FreeC[Algebra[F, O, ?], Unit]`, but for `FreeC[F, O, R]` instead. 

With the previous 2-layered representation, there was a need to have in the `Algebra` object a few methods to lift simple data and fill the type parameters. With the new representations, those methods are not needed and we can remove them in favour of the object they build. That is the case of the private methods `openScope` and `step`.  

#### Translate to mapOutput

Before this change, the implementation of Stream[F, A].map[B](f: A => B) had to work at two levels: one being a `FunctionK` between two type-instances of Algebra, which would differ on the output type. To apply this transformation through the Stream, the `FreeC` ADT had a `translate` method. After the changes, now we do not need that `FunctionK` wrapper, and we can just replace the translate method of FreeC with a mapOutput method.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -8,27 +8,42 @@ import fs2.internal.FreeC.{Result, ViewL}
 
 import scala.util.control.NonFatal
 
-/* `Algebra[F[_], O, R]` is a Generalised Algebraic Data Type (GADT)
+/* `Eval[F[_], O, R]` is a Generalised Algebraic Data Type (GADT)
  * of atomic instructions that can be evaluated in the effect `F`
  * to generate by-product outputs of type `O`.
  *
  * Each operation also generates an output of type `R` that is used
  * as control information for the rest of the interpretation or compilation.
  */
-private[fs2] sealed trait Algebra[F[_], +O, R]
-
 private[fs2] object Algebra {
 
-  private[this] final case class Output[F[_], O](values: Chunk[O]) extends Algebra[F, O, Unit]
+  private[this] final case class Output[F[_], O](values: Chunk[O]) extends FreeC.Eval[F, O, Unit] {
+    override def mapOutput[P](f: O => P): FreeC[F, P, Unit] =
+      FreeC.suspend {
+        try Output(values.map(f))
+        catch { case NonFatal(t) => Result.Fail[F](t) }
+      }
+  }
 
-  private[this] final case class Step[F[_], X](
-      stream: FreeC[Algebra[F, X, ?], Unit],
-      scope: Option[Token]
-  ) extends Algebra[F, INothing, Option[(Chunk[X], Token, FreeC[Algebra[F, X, ?], Unit])]]
+  /**
+    * Steps through the stream, providing either `uncons` or `stepLeg`.
+    * Yields to head in form of chunk, then id of the scope that was active after step evaluated and tail of the `stream`.
+    *
+    * @param stream             Stream to step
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/fs2/internal/Algebra.scala
```diff
@@ -49,94 +64,50 @@ private[fs2] object Algebra {
...
-
-  def mapOutput[F[_], A, B](fun: A => B): Algebra[F, A, ?] ~> Algebra[F, B, ?] =
-    new (Algebra[F, A, ?] ~> Algebra[F, B, ?]) {
-      def apply[R](alg: Algebra[F, A, R]): Algebra[F, B, R] = alg match {
-        case o: Output[F, A] => Output[F, B](o.values.map(fun))
-        case _               => alg.asInstanceOf[Algebra[F, B, R]]
-      }
-    }
+  ): FreeC[F, O, R] =
+    Acquire(resource, release)
```
diesalbla: @mpilquist For now, I have only unfolded/removed the factory methods that were private to this object. However, since the whole `Algebra` is private to `fs2`, would it make sense to open  the subclasses and remove this method, as well as `eval` and `output` above?

diesalbla:
The attached file contains (compressed) a spreadsheet with the results of some more benchmarking results.

I used the suite `StreamBenchmarks`, with modifications. In each single benchmark, in the `@GenerateN` annotation I reduced the range `1, 10, 100, 1000, 10000, 100000` to use only the values `10`, 1000`, and `100000`. I am also attaching a diff file with the changes. For running them, I used the following command: 
`jmh:run -f 1 -i 10 -wi 5 .*StreamBenchmark*. -prof gc`

The results indicate, across most benchmarks, a reduction in memory consumption of 2.5% (the `eval` benchmark), 3% (`leftAssocConcat`), 5% (`leftAssocFlatmap`), 3.25% (`rightAssocConcat`), and 5% (`rightAssocFlatmap`). The one micro-benchmark for which no improvement is observed is in `toVector`, but that is understandable given that most allocations there go in creating the Vector itself.


[Diff file (a patch)](https://github.com/functional-streams-for-scala/fs2/files/3603219/benchmarking_mod.patch.txt)


[StreamBenchmarks_UnifyFreecAlgebra.ods.zip](https://github.com/functional-streams-for-scala/fs2/files/3603205/StreamBenchmarks_UnifyFreecAlgebra.ods.zip)

pchlupacek:
@diesalbla seems promissing, thanks for this

PR #1615, opened 2019-09-12
FreeC and subclasses: make the F parameter Covariant

diesalbla:
**WIP** Currently, this branch only compiles main code, and only under Scala 2.13.0. In Scala 2.12, it runs into several compiler bugs and crashes the compiler.

Built on top of #1614.

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/FreeCBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/FreeCBenchmark.scala
@@ -15,7 +15,7 @@ class FreeCBenchmark {
   @Benchmark
   def nestedMaps = {
     val nestedMapsFreeC =
-      (0 to N).foldLeft(Result.Pure[IO, Int](0): FreeC[IO, INothing, Int]) { (acc, i) =>
+      (0 to N).foldLeft(Result.Pure[Int](0): FreeC[IO, INothing, Int]) { (acc, i) =>
         acc.map(_ + i)
       }
     run(nestedMapsFreeC)
@@ -24,7 +24,7 @@ class FreeCBenchmark {
   @Benchmark
   def nestedFlatMaps = {
     val nestedFlatMapsFreeC =
-      (0 to N).foldLeft(Result.Pure[IO, Int](0): FreeC[IO, INothing, Int]) { (acc, i) =>
+      (0 to N).foldLeft(Result.Pure[Int](0): FreeC[IO, INothing, Int]) { (acc, i) =>
         acc.flatMap(j => Result.Pure(i + j))
       }
     run(nestedFlatMapsFreeC)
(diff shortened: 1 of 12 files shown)


mpilquist:
Looks promising! We'll need to figure out 2.11/2.12 support so a branch that keeps those versions enabled would be good -- something where we can minimize the compiler issues and look for workarounds. Assuming we get through that, we should be able to get rid of `FreeC` altogether, right?

diesalbla:
> We'll need to figure out 2.11/2.12 support so a branch that keeps those versions enabled would be good -- something where we can minimize the compiler issues and look for workarounds. 

The bugs being hit by this branch are not trivial: they concern the interaction of higuer-kinds and variance. The Scala bug https://github.com/scala/bug/ has hundreds of these. We could try to look for the ones that may be involved, and see if we can back-port their fixes to Scala 2.12. 

However, given that Lightbend no longer maintains 2.11, and even 2.12 bugs are now taken with caution, I am not sure how much that effort would be worth. It may be that a 2.13-only or 2.13-first branch would be easier going forward.

> Assuming we get through that, we should be able to get rid of FreeC altogether, right?

I would not be sure about that. I can see a lot of value in the current key design idea, that the public API of `Stream` and `Pull` encapsulates and thus decouples the users of the library from their internal implementation of `FreeC / Algebra` ADT. That separation has allowed it to evolve without compatibility problems.

The current PR, as it is, already has some benefits in removing some of the "variance"/"type-casting" patches. It also has some type-safety, and it detected the unsound type-cast error that @SystemFw solved last night in https://github.com/functional-streams-for-scala/fs2/pull/1617

mpilquist:
> I can see a lot of value in the current key design idea, that the public API of Stream and Pull encapsulates and thus decouples the users of the library from their internal implementation of FreeC / Algebra ADT.

Right, I wouldn't give up on that. I want to get rid of `FreeC` (and `Algebra`) and what's currently API of those end up as `private`/`private[fs2]` members of `Pull`.

core/shared/src/main/scala/fs2/fs2.scala
```diff
@@ -1,5 +1,9 @@
 package object fs2 {
 
+  /** A stream transformation
+    */
+  type VarPipe[-F[_], +G[_], -I, +O] = Stream[F, I] => Stream[G, O]
```
mpilquist: I don't want to provide this type alias. In general, we need less type aliases instead of more.

diesalbla: Oh, this is something I forgot behind when I was trying to fix a failing test in `TextSpec`, which was caused by the use of _covariance_ functions. I tried defining a `PureIn`, the opposite of `Pure` (an alias for `Any`) to type streams that appeared as inputs... but that did not work well. 

This is a matter for which we should also have some design discussions: how should we define Pipes by default if the `F` is co-variant.

diesalbla:
I have found a "hacky solution" for the 2.12 problem. I am continuing in a different Pull Request.

PR #1619, opened 2019-09-13
Stream.evalFilter

kubukoz:
Reincarnation of #1583.


mpilquist:
The tests were failing on the call to `s.toList` in `asserting(_ shouldBe s.toList)`. The call to `s.toList` calls `unsafeRunSync` on an internal `IO` value that's guaranteed to be synchronous. However, the internals of IO have a bug where a separate IO compilation can confuse the check -- the workaround is here: https://github.com/typelevel/cats-effect/issues/529

kubukoz:
That looks slightly better, thanks. Still getting some weird JS errors...

mpilquist:
@kubukoz I got it passing. Ready to merge if you are.

kubukoz:
Good to go!

PR #1630, opened 2019-09-21
Solves memory Leak

### LINKED ISSUE #1625: Memory leak
Hi.

I have written a `Pipe2` that computes a diff between two sorted streams, and it seems to cause a memory leak that in some cases leads to the heap blowing up.

To ease everyone's life, I have created [this GitHub project](https://github.com/epellizzer/showcase-fs2-diff) that contains the commented code as well as the unit tests that show the problem.

It was the first time I wrote a `Pipe2`, so I hope my implementation is not too naive, but even so, FS2 should probably not react that way.

diesalbla:
Fixes #1625. Reverts changes in commit 36b1279cda415ebe8bb3faa33cb9aa3e6c6697d9.

This commit changed the Compilation Loop was modified in a recent PR to use a continuation and a stack of continuations. However, this introduced a memory leak that could be seen running a simple program, as described in #1625.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Algebra.scala
+++ b/core/shared/src/main/scala/fs2/internal/Algebra.scala
@@ -138,13 +138,6 @@ private[fs2] object Algebra {
         F.pure(init)
     }
 
-  private[this] trait CompileCont[F[_], X, Y] {
-    final type Out = Y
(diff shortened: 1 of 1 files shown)


mpilquist:
Thanks!

PR #1631, opened 2019-09-21
Re-enabled mima on master and cleaned up exclusion list

mpilquist:
I was getting ready to do a 2.0.1 release to get #1630 out when I noticed recent changes on master broke binary compatibility. These snuck through b/c of some holes in the Mima configuration. This PR fixes those holes. I'll create another PR to roll back recent changes that broke binary compatibility.


mpilquist:
This PR disables checking changes to `ScalaSignature` -- see lightbend/mima#361 for details on this. We don't make any bincompat guarantees w.r.t. reflective access to methods, only jar swapping (e.g. running lib built against 2.0.0 with 2.1.0).

PR #1634, opened 2019-09-22
Benchmarks: add recent memory leak

diesalbla:
There recently was a memory leak https://github.com/functional-streams-for-scala/fs2/issues/1625 introduced by some changes to the `compileLoop` method. We add as a benchmark the core part of that memory leak, in order to guide future development.


mpilquist:
Might be better to add this to the `MemorySanityChecks` file. Neither the benchmarks nor `MemorySanityChecks` are run routinely though. I'd really like to create a ScalaTest runner that runs each of the programs in `MemorySanityChecks` for a short while in a JVM forked with a small amount of memory and then integrate that with Travis build. I haven't gotten to that for years now though.

PR #1635, opened 2019-09-24
Add `evalFilterNot` and `evalFilterNotAsync`

CremboC:
A recent PR #1619 added `evalFilter` and `evalFilterAsync`, here I'm added their cousins `evalFilterNot` and `evalFilterNotAsync`.

This is an optimisations instead of doing something like
```
_.evalFilter(foo.map(!_))
```
we simply do the more obvious
```
_.evalFilterNot(foo)
```

Only issue I see is the name of `evalFilterNotAsync` being bad. But it's follows the same pattern so not sure what else to call it...

P.S. I am expecting the tests to fail on this first commit. Specifically hoping only one of them fails and I'll fix it once it does.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1039,6 +1039,26 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Nothing, O, Unit]
     }.flatten
 
   /**
+   * Like `filterNot`, but allows filtering based on an effect.
+   *
+   * Note: The result Stream will consist of chunks that are empty or 1-element-long.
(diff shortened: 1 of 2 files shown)


mpilquist:
I recently noticed we don't have `filterNot` on `Stream` -- want to add that too?

SystemFw:
I'm going to say it: I think both `evalFilterNotAsync` and `evalFilterAsync` are overly specific :)

CremboC:
@SystemFw I don't need the `Async` myself, but since it's there, I added the `Not` version for symmetry 🤷‍♂

SystemFw:
Yeah I would remove both honestly :P

PR #1639, opened 2019-09-27
Add foldEvery

TimWSpence:
- Windows the stream into fixed widths and runs a fold incrementally
  on each window, emitting the result of the fold

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3024,6 +3024,39 @@ object Stream extends StreamLowPriority {
         .eval(timer.clock.monotonic(NANOSECONDS).map(now => (now - start).nanos))
     }
 
+  def foldEvery[F[_], A, B](
+      d: FiniteDuration
+  )(init: B)(f: (B, A) => B)(implicit F: Sync[F]): Pipe[F, A, B] = { in =>
+    {
+      val width = d.toMillis
+
+      val signal = Stream.eval(F.delay(System.currentTimeMillis())).repeat.zipWithPrevious.map {
+        case (None, _) => false
+        case (Some(t1), t2) => {
+          val s1 = t1 / width
+          val s2 = t2 / width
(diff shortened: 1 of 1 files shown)


TimWSpence:
Is there any interest in me cleaning this up? A colleague challenged me to write something like this recently and I thought it seemed a reasoanably useful and generic function

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3024,6 +3024,39 @@ object Stream extends StreamLowPriority {
         .eval(timer.clock.monotonic(NANOSECONDS).map(now => (now - start).nanos))
     }
 
+  def foldEvery[F[_], A, B](
+      d: FiniteDuration
+  )(init: B)(f: (B, A) => B)(implicit F: Sync[F]): Pipe[F, A, B] = { in =>
+    {
+      val width = d.toMillis
+
+      val signal = Stream.eval(F.delay(System.currentTimeMillis())).repeat.zipWithPrevious.map {
```
Daenyth: This should use `Timer[F]` rather than `System`

TimWSpence: For sure, that would definitely be part of cleanup! I just thought I'd get some feedback on whether people would like it to be part of fs2, before I invest time in cleaning it up, writing tests, etc

mpilquist:
Is `s.foldEvery(d)(z)(f) == s.groupWithin(Int.MaxValue, d).map(_.fold(z)(f))`?

CremboC:
@mpilquist in theory yes, but the key difference is the stream nature.
`groupWithin` will basically block for either `n` elements, or `d` time and _then_ do the aggregation afterwards (basically wasting time, blocking the upstream then)
while `foldEvery` will aggregate in a streaming fashion, without blocking upstream. Does this make sense?

We also implemented `evalFoldEvery` with @TimWSpence which I think is worth adding to this PR.

PR #1644, opened 2019-09-30
Add a section on `interruptWhen` to guide.md

poliez:


site/src/main/tut/guide.md
```diff
@@ -412,6 +413,81 @@ type Pipe2[F[_],-I,-I2,+O] = (Stream[F,I], Stream[F,I2]) => Stream[F,O]
 def mergeHaltBoth[F[_]:Concurrent,O]: Pipe2[F,O,O,O] = (s1, s2) => ???
 ```
 
+### Interruption
+
+Sometimes some tasks have to run only when some conditions are met or until some other task completes. Luckily for us, `Stream` defines some really useful methods that let us accomplish this.
+In the following example we will see how `interruptWhen` helps us to describe such cases. we will describe a program composed by two concurrent streams: the first will print the current time to the console every second, the second will stop the first.
```
mpilquist: s/we/We/

site/src/main/tut/guide.md
```diff
@@ -412,6 +413,81 @@ type Pipe2[F[_],-I,-I2,+O] = (Stream[F,I], Stream[F,I2]) => Stream[F,O]
...
+program.compile.drain.unsafeRunSync
+```
+
+Let's take this line by line now, so we can understand what's going on.
+```scala
+val program = 
+  Stream.eval(Deferred[IO, Unit]).flatMap { switch =>
+```
+
+Here we create a `Stream[IO, Deferred[IO, Unit]]`. [`Deferred`](https://typelevel.org/cats-effect/concurrency/deferred.html) is a concurrency primitive that represents an unset promise. we will use the emitted istance of `Deferred[IO, Unit]` as a mechanism to signal the completion of a task. Given this purpose, we call this instance `switch`.
```
mpilquist: s/we/We/

SystemFw: I would only change "unset promise" to "a condition yet to be fulfilled" possibly

poliez:
> Great example, very clear!

Thanks! I'll commit the requested changes in a few minutes. Thanks for accepting this :smile:

PR #1648, opened 2019-10-03
Add `readOutputStream` combinator in fs2-io

### LINKED ISSUE #1641: Add a combinator for (OutputStream => Unit) => Stream[F, Byte]
Some java apis accept an `OutputStream` as an input parameter, which they then write to. A combinator that exposes the results of those writes as an fs2 stream is desirable for wrapping such libraries; `(OutputStream => Unit) => Stream[F, Byte]`

I'll PR this to `fs2-io` as I have time, based on my previous work:  https://gist.github.com/Daenyth/9a569040a4333595b2be4019d1bc6b91

Daenyth:
I still need to finalize the unit tests - I copied what was in my gist with minor adaptations, needs cleaning still

Close #1641

### DIFF
--- a/io/src/main/scala/fs2/io/io.scala
+++ b/io/src/main/scala/fs2/io/io.scala
@@ -1,10 +1,10 @@
 package fs2
 
-import cats.effect.{Async, Blocker, ConcurrentEffect, ContextShift, Resource, Sync}
+import cats.effect.{Async, Blocker, Concurrent, ConcurrentEffect, ContextShift, Resource, Sync}
 
 import cats._
 import cats.implicits._
-import java.io.{InputStream, OutputStream}
+import java.io.{InputStream, OutputStream, PipedInputStream, PipedOutputStream}
 import java.nio.charset.Charset
 
 /**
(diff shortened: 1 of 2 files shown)


mpilquist:
@Daenyth Anything left on this one? It's still marked as a draft.

Daenyth:
I swear the build was failing, but it's green now and I haven't changed anything. I have no idea what's up there.

Daenyth:
@mpilquist should be good to go now that it's green

nigredo-tori:
About the failing tests, I'll repeat it here so it doesn't get lost: it looks like the first test can deadlock on a single-thread `ContextShift`, since it writes without shifting to `blocker`. This can be flaky, since for a deadlock it needs to overflow the pipe, and that should only happen with a sufficiently large gnerated array (IIRC, over 2KB).

io/src/main/scala/fs2/io/io.scala
```diff
@@ -112,6 +112,47 @@ package object io {
...
+          Sync[F].delay {
+            // Piped(I/O)Stream implementations cant't throw on close, no need to nest the handling here.
+            ois._2.close()
+            ois._1.close()
+          }
+      )
+
+    Stream.resource(mkOutput).flatMap {
+      case (os, is) =>
+        val write = f(os) *> Sync[F].delay(os.close())
```
mpilquist: Use `blocker.delay` here instead

io/src/test/scala/fs2/io/IoSpec.scala
```diff
@@ -25,6 +28,64 @@ class IoSpec extends Fs2Spec {
...
   }
 
+  "readOutputStream" - {
+    "writes data and terminates when `f` returns" in forAll(
+      arrayGenerator[Byte],
+      intsBetween(1, 20)
+    ) { (bytes: Array[Byte], chunkSize: Int) =>
+      Blocker[IO].use { blocker =>
+        readOutputStream[IO](blocker, chunkSize)(
+          (os: OutputStream) => IO(os.write(bytes))
```
mpilquist: Use `blocker.delay[IO]` instead (h/t @nigredo-tori)

Issue #1527, opened 2019-06-28
Add `unconsNonEmpty` combinator to `Pull`


Daenyth:
It's often useful to write pull with `uncons`, and skip over empty chunks by pulling the next chunk. This combinator makes sense to be in the library

example impl from @mpilquist: https://gitter.im/functional-streams-for-scala/fs2?at=5d162e628dff05627b84bebe

```scala
    def unconsNonEmpty: Pull[F, INothing, Option[(Chunk[O], Stream[F, O])]] =
      self.uncons.flatMap {
        case Some((hd, tl)) =>
          if (hd.isEmpty) tl.pull.unconsNonEmpty
          else Pull.pure(Some((hd, tl)))
        case None => Pull.pure(None)
      }
```

monadplus:
I was wondering in which use cases you will be pulling empty chunks 🤔

PR #1650, opened 2019-10-03
Add unconsNonEmpty (#1527)

valydia:
I can't see any unit test related to the `uncons*` method? Shall this one follow the rule as well

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3847,7 +3847,7 @@ object Stream extends StreamLowPriority {
       else go(Nil, n, self)
     }
 
-    /** Like [[uncons]] but skips over the empty chunk. */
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3847,6 +3847,15 @@ object Stream extends StreamLowPriority {
       else go(Nil, n, self)
     }
 
+    /** Like [[uncons]] but skips over the empty chunk. */
```
Daenyth:
```suggestion
    /** Like [[uncons]] but skips over empty chunks, pulling until it can emit the first non-empty chunk. */
```

What do you think about this alternate wording? I think it's helpful to be very explicit in the method docs since fs2 has so many operations

mpilquist:
Thanks!

PR #1660, opened 2019-10-14
Fixed deadlock race condition in readOutputStream

mpilquist:
cc @Daenyth @nigredo-tori

### DIFF
--- a/io/src/main/scala/fs2/io/io.scala
+++ b/io/src/main/scala/fs2/io/io.scala
@@ -1,8 +1,18 @@
 package fs2
 
 import cats._
-import cats.effect.{Async, Blocker, Concurrent, ConcurrentEffect, ContextShift, Resource, Sync}
-import cats.effect.concurrent.Deferred
+import cats.effect.{
+  Async,
+  Blocker,
+  Concurrent,
+  ConcurrentEffect,
+  ContextShift,
+  ExitCase,
+  Resource,
+  Sync
+}
+import cats.effect.implicits._
+import cats.effect.concurrent.Ref
 import cats.implicits._
 import java.io.{InputStream, OutputStream, PipedInputStream, PipedOutputStream}
 import java.nio.charset.Charset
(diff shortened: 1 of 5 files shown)


io/src/main/scala/fs2/io/io.scala
```diff
@@ -146,9 +156,24 @@ package object io {
...
+            ec =>
+              blocker.delay(os.close()) *> ref.set(ec match {
+                case ExitCase.Error(t) => Some(t); case _ => None
+              })
+          )
+          val read = readInputStream(is.pure[F], chunkSize, blocker, closeAfterUse = false)
+          read.concurrently(Stream.eval(write)) ++ Stream.eval(ref.get).flatMap {
+            case None    => Stream.empty
+            case Some(t) => Stream.raiseError[F](t)
+          }
```
nigredo-tori: This is messy... How about something like this:
```scala
val mkIn = Resource.fromAutoCloseable(Sync[F].delay { new PipedInputStream })
def mkOut(is: PipedInputStream) = Resource.fromAutoCloseable(Sync[F].delay { new PipedOutputStream(is) })

Stream.resource(mkIn).flatMap { is =>
  Deferred.of[F, Unit].flatMap { connected =>
    val write = Stream.eval_(mkOut(is).evalTap(connected.complete(())).use(f))
    // We can't read from the pipe until we are sure it will definitely get closed.
    val read = Stream.eval_(connected.get) ++ readInputStream(is.pure[F], chunkSize, blocker, closeAfterUse = false)
    read.concurrently(write)
  }
}
```
This still has annoying synchronization (I don't see how we can get rid of it), but no explicit closing or error handling. It's relatively clear that reading from the pipe doesn't prevent the `PipedOutputStream` from closing. I can't see a deadlock, either - although it's already night here, so I'm not at my best.

mpilquist: Yeah, I like yours better. Want to open a PR and I'll close this one?

nigredo-tori: I'd rather it was incorporated into yours in any way you see fit, if that's all right with you.

mpilquist: :+1: Just pushed an update.

mpilquist: I had to revert it as the new implementation still has a race at shutdown time:

```
- fails when `f` fails *** FAILED *** (90 milliseconds)
[info]   GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation. (IoSpec.scala:54)
[info]     Falsified after 10 successful property evaluations.
[info]     Message: Right(Vector()) was not equal to Left(java.lang.Exception: boom)
[info]     Location: (IoSpec.scala:58)
[info]     Occurred when passed generated values (
[info]       10
[info]     )
[info]     Init Seed: 1571079680891
...
Cause: org.scalatest.exceptions.TestFailedException: Right(Vector()) was not equal to Left(java.lang.Exception: boom)
```

mpilquist:
I ran IoSpec over 10,000 times (by bumping up the ScalaCheck iteration count) and was able to reproduce this reliably. This current implementation passed all 10k iterations. Merging on green.

io/src/main/scala/fs2/io/io.scala
```diff
@@ -137,7 +147,7 @@ package object io {
         (os: OutputStream, is: InputStream)
       })(
         ois =>
-          Sync[F].delay {
+          blocker.delay {
```
Daenyth: FWIW this probably isn't needed; the reason I didn't do it before was to avoid the thread shift. Close in P(I/O)Stream just sets some internal state

io/src/main/scala/fs2/io/io.scala
```diff
@@ -146,9 +156,25 @@ package object io {
 
     Stream.resource(mkOutput).flatMap {
       case (os, is) =>
-        val write = f(os) *> Sync[F].delay(os.close())
-        val read = readInputStream(is.pure[F], chunkSize, blocker, closeAfterUse = false)
-        read.concurrently(Stream.eval(write))
+        Stream.eval(Deferred[F, Option[Throwable]]).flatMap { err =>
+          // We need to close the output stream regardless of how `f` finishes
```
Daenyth: Good find. I think I never ran into this because of the way I used the combinator in practice.

PR #1665, opened 2019-10-15
Simplify readOutputStream

nigredo-tori:
This replaces `readOutputStream` implementation with something less low-level, as proposed [here](https://github.com/functional-streams-for-scala/fs2/pull/1660/files#r334579468).

I have failed to reproduce the `Right(Vector()) was not equal to Left(java.lang.Exception: boom)` thing in my environment (Linux, Java 11, 8 cores) - could it have been fixed by one of the latest commits? Instead, I have been reliably seeing `InterruptedException`s in the same test. Changing `concurrently` to `mergeHaltR` has solved this, which is worrying, since they are supposed to be equivalent. I'll try and shrink that to a testable example tomorrow.

I have ran the `IoSpec` a few times with `maxSuccessful=10000` and different `-XX:ActiveProcessorCount` values, and had no failures. @mpilquist , @Daenyth, could you please run the tests in whatever environment you ran them in when you had problems with this approach and/or the original implementation?

### DIFF
--- a/io/src/main/scala/fs2/io/io.scala
+++ b/io/src/main/scala/fs2/io/io.scala
@@ -1,17 +1,7 @@
 package fs2
 
 import cats._
-import cats.effect.{
-  Async,
-  Blocker,
-  Concurrent,
-  ConcurrentEffect,
-  ContextShift,
-  ExitCase,
-  Resource,
-  Sync
-}
-import cats.effect.implicits._
+import cats.effect.{Async, Blocker, Concurrent, ConcurrentEffect, ContextShift, Resource, Sync}
 import cats.effect.concurrent.Deferred
 import cats.implicits._
 import java.io.{InputStream, OutputStream, PipedInputStream, PipedOutputStream}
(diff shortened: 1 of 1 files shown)


io/src/main/scala/fs2/io/io.scala
```diff
@@ -139,44 +129,26 @@ package object io {
...
-            case Some(t) => Stream.raiseError[F](t)
-          }
-        }
+    val mkIn = Resource.fromAutoCloseable(Sync[F].delay { new PipedInputStream })
+    def mkOut(is: PipedInputStream) =
+      Resource.fromAutoCloseable(Sync[F].delay { new PipedOutputStream(is) })
+
+    Stream.resource(mkIn).flatMap { is =>
+      Stream.eval(Deferred[F, Unit]).flatMap { connected =>
+        val write = Stream.eval_(mkOut(is).evalTap(_ => connected.complete(())).use(f))
```
Daenyth: I think we want to `blocker.blockOn(write)`. While it's true that the caller *can* do this, I feel like the implementation should, because it's required 100% of the time.

mpilquist:
Issue reproduced for me on second run of 10000 iterations per run:

```
Right(Vector()) was not equal to Left(java.lang.Exception: boom)
```

Furthermore, I don't see how this change addresses the race condition here. The original issue had 2 changes needed:
1) Ensure that when the writer terminated with an error, we closed the `PipedOutputStream` so that any pending read on the `PipedInputStream` terminates with `-1`.
2) In such a case, preventing a race condition where we have this logical interleaving of events:
  a) Writer terminates with an error and closes output stream
  b) Reader comes off a pending read with a -1, indicating no more to read
  c) Reader terminates gracefully
  d) Reader side of `concurrently` requests termination of writer side
  e) Writer side finishes shutting down from (a) but b/c reader pre-empted the shutdown, the error isn't propagated back to overall stream.

AFAICT, interleavings like the above, where shutdown of the `PipedOutputStream` cause the reader side of `concurrently` to terminates successfully before the failure is propagated, are the root cause of this test failure.

PR #1668, opened 2019-10-18
Add canoe to the list of projects

augustjune:
I would like a library of mine to be added to the list of projects which use fs2 library.


mpilquist:
Thanks!!

PR #1670, opened 2019-10-19
Request: Add neotypes to the library ecosystem

BalmungSan:
I am proposing the addition of **neotypes** to the library list on the ecosystem section.

Even though **neotypes** is not directly built using **cats-effect** nor **fs2**, it does provides two modules to integrate with these two.

Let me know if I should make something else.


BalmungSan:
I just noticed that the list was sorted alphabetically, I thought I should have added it at the bottom _(not sure why `¯\_(ツ)_/¯`)_.

I have already moved it to the correct place.

mpilquist:
Thanks!

Issue #1511, opened 2019-06-15
Write scalafix upgrade for 1.0 to 1.1/2.0


mpilquist:
The 1.1.0-M1 (soon to be relabeled as 2.0) release is mostly source compatible with 1.0.x, with a handful of exceptions:
- replaced use of blocking execution contexts with `cats.effect.Blocker`
- changed TCP/UDP support to use `SocketGroup`
- maybe changed `Pull#stream` and `Pull#streamNoScope` to require result type to be `Unit`

A scalafix upgrade script that automates as much of those as possible would be fantastic.

fthomas:
It would be twice as fantastic if the Scalafix migration would then be added to [Scala Steward](https://github.com/fthomas/scala-steward/blob/master/docs/scalafix-migrations.md) before the final release. :-)

PR #1675, opened 2019-10-25
WIP: Scalafix rule to upgrade to v2.0 (#1511)

valydia:
This is a WIP to implement the scalafix rule to upgrade from v1.0.5 to v2.0.0. 

>replaced use of blocking execution contexts with cats.effect.Blocker

Changes for this are mostly inspired from [this commit](https://github.com/functional-streams-for-scala/fs2/commit/c42170e95baa1e796886ad9a15856b4c34fdc269)

>changed TCP/UDP support to use SocketGroup

UDP socket group has been implemented - for the TCP, I haven't worked out a solutions yet

>maybe changed Pull#stream and Pull#streamNoScope to require result type to be Unit

Haven't done that part yet.

If you want help / give guidance with the part that still needs to be done, please shout at me. Or if you can think of more use case that should be included let me know as well 👍


scalafix/build.sbt
```diff
@@ -57,3 +64,47 @@ lazy val tests = project
   )
   .dependsOn(rules)
   .enablePlugins(ScalafixTestkitPlugin)
+
```
valydia: Created other sets of module for the scalafix v2, as it uses different version of fs2. To run the v2 tests, run `sbt tests2/test`

scalafix/v2/rules/src/main/scala/fix/v2.scala
```diff
@@ -0,0 +1,370 @@
...
+                      List()
+                    )
+                  )
+              )
+            )
+          ):: params
+        Patch.replaceTree(ap, Term.Apply(fun, newParams).toString)
+
+      case fromPathMatcher(ap @ Term.Apply(fun, List(_, bec, _))) =>
+        replaceBlockingExecutionContextTermAndRenameFunction(bec, ap, resourceFMap)(fun, s"io.file.FileHandle.fromPath")
```
valydia: This transformation will change the type, indeed in [this commit](https://github.com/functional-streams-for-scala/fs2/commit/bdcbb5b1c69d80a62d84c531bf7519aee13ac7f0#diff-88c75a27cf772b8d90dd50ba68691626L108):
```scala
io.file.pulls.fromPath(...): Pull[F, Nothing, Pull.Cancellable[F, FileHandle[F]]]
``` 
to
```scala
io.file.FileHandle.fromPath(...): Resource[F, FileHandle[F]]
```
But the user can use the compiler error to sort that further... Unless someone know a way to convert `Pull[F, Nothing, Pull.Cancellable[F, FileHandle[F]]]` into a `Resource[F, FileHandle[F]]` in which case the rule can be improved

scalafix/v2/rules/src/main/scala/fix/v2.scala
```diff
@@ -0,0 +1,370 @@
...
+
+
+      case fromFileChannelMatcher(ap @ Term.Apply(fun, List(_, bec))) =>
+        replaceBlockingExecutionContextTermAndRenameFunction(bec, ap, resourceFMap)(fun, "io.file.FileHandle.fromFileChannel")
```
valydia: Same comment as above

scalafix/v2/rules/src/main/scala/fix/v2.scala
```diff
@@ -0,0 +1,370 @@
...
+                  Term.Select(
+                    ap @ Term.Apply(
+                      Term.Select(Term.Name("Stream"), Term.Name("resource")),
+                      List(bec)
+                    ),
+                  Term.Name("flatMap") | Term.Name("map")
+                  ),
+                _)  if secondType(getType(bec.symbol)).exists(isTcpSocketGroup) =>
+
+      // TODO
```
valydia: This is were the Patch for TCP SocketGroup would come

mpilquist:

PR #1681, opened 2019-11-08
More flexible `readOutputStream` signature

wedens:
My current use case for this change:

I have the following definitions:
```scala
def createEventWriter(os: OutputStream): Resource[IO, XMLEventWriter]
def writeEvent(writer: XMLEventWriter): XMLEvent => IO[Unit]
def encodeEvents(blocker: Blocker, chunkSize: Int): Pipe[IO, XMLEvent, Byte] 
```
My goal is to implement `encodeEvents`. 

What I have to do with the current signature:
```scala
  def writeJavaEvents(blocker: Blocker, chunkSize: Int): Pipe[IO, XMLEvent, Byte] = events => {
    fs2.io.readOutputStream[IO](blocker, chunkSize) { os =>
      createEventWriter(os).use { writer =>
        events.evalMap(event => IO(writer.add(event))).compile.drain
      }
    }
  }
```

For me, having to call `compile` in the middle of the code involving streams is a red flag. What I'd prefer to write is:

```scala
  def writeJavaEvents(blocker: Blocker, chunkSize: Int): Pipe[IO, XMLEvent, Byte] = events => {
    fs2.io.readOutputStream[IO](blocker, chunkSize) { os =>
      Stream.resource(createEventWriter(os)).flatMap { writer =>
        events.evalMap(writeEvent(writer))
      }
    }
  }
```

I'm open to being convinced that that the current signature is preferable, if this is the case ;)

### DIFF
--- a/io/src/main/scala/fs2/io/io.scala
+++ b/io/src/main/scala/fs2/io/io.scala
@@ -137,7 +137,7 @@ package object io {
       blocker: Blocker,
       chunkSize: Int
   )(
-      f: OutputStream => F[Unit]
+      f: OutputStream => Stream[F, Unit]
   ): Stream[F, Byte] = {
     val mkOutput: Resource[F, (OutputStream, InputStream)] =
       Resource.make(Sync[F].delay {
@@ -161,7 +161,7 @@ package object io {
           // In such a case, there's a race between completion of the read
           // stream and finalization of the write stream, so we capture the error
           // that occurs when writing and rethrow it.
-          val write = f(os).guaranteeCase(
+          val write = f(os).onFinalizeCase(
             ec =>
               blocker.delay(os.close()) *> err.complete(ec match {
                 case ExitCase.Error(t) => Some(t)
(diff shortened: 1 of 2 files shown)


mpilquist:
No objection here but would like to hear from @Daenyth. I'm surprised MiMa didn't flag the change as binary incompatible. I guess it's due to erasure of the Function1 argument. We'll need to have a `readOutputStreamS` or something with the new proposed signature.

Daenyth:
I'm not convinced that the code example using `compile` is a "red flag". For me, it exactly matches the shape of computation that you want to use. I'm reluctant to widen the type signature to `Stream`, because it's so much more powerful. It makes it more likely this gets used in a surprising way, I would think, because of that extra power. I'd be more convinced if it caused a function to require `Sync` for compile when it didn't require `Sync` otherwise, but because we already require a more strict constraint, it doesn't make a difference in this case.

If we decide to merge it, I think we'll want to have a new test that covers invoking this with a stream that emits `Unit` multiple times, just to cover the bases.

👍 on making this a new name, both for binary compatibility safety as well as for guiding the user toward correct usage.

mpilquist:

PR #1682, opened 2019-11-11
preserve CompositeFailure in parJoin

ingarabr:
The bug here is that we're wrapping CompositeFailure into CompositeFailure.
It makes the CompositeFailure structure a bit useless. It preserves the order of the first error but not the remaining/tailing to prevent appending elements to the list.


mpilquist:
Thanks! There appears to be other cases that need similar guards -- basically anywhere we are creating a new `CompositeFailure`, both in `FreeC` and `Algebra`. I wonder if we should put some logic in `CompositeFailure.apply` that flattens any top level `CompositeFailure` instances.

ingarabr:
Yeah, I was thinking about the same. Is 7a3e8c3 in line with what you had in mind? I didn't find any other use cases of `new CompositeFailure` so it should now be caught in the `apply` method.

core/shared/src/main/scala/fs2/CompositeFailure.scala
```diff
@@ -20,7 +20,18 @@ object CompositeFailure {
...
       rest: List[Throwable] = List.empty
   ): CompositeFailure =
-    new CompositeFailure(first, NonEmptyList(second, rest))
+    apply(first, NonEmptyList(second, rest))
+
+  def apply(
+      first: Throwable,
+      rest: NonEmptyList[Throwable]
+  ): CompositeFailure =
+    first match {
```
mpilquist: Seems like we need to check each member of `rest` too.

mpilquist:
Yep, see the one comment I made about expectation of `apply` behavior. Otherwise looks good.

ingarabr:
@mpilquist , thanks!

PR #1684, opened 2019-11-18
Mention Scala 2.13 in README

mgibowski:
README mentions other Scala versions but not 2.13. Somebody could be mislead that it is not supported, hence these small changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ FS2: Functional Streams for Scala
 
 FS2 is a library for purely functional, effectful, and polymorphic stream processing library in the [Scala programming language](https://scala-lang.org). Its design goals are compositionality, expressiveness, resource safety, and speed. The name is a modified acronym for **F**unctional **S**treams for **Scala** (FSS, or FS2).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,7 +37,7 @@ The [1.0 migration guide](https://github.com/functional-streams-for-scala/fs2/bl
 summarizes the differences between 1.0 and 0.10. To get 1.0.x, add the following to your SBT build:
 
 ```
-// available for Scala 2.11, 2.12
+// available for Scala 2.11, 2.12, 2.13
 libraryDependencies += "co.fs2" %% "fs2-core" % "2.0.0" // For cats 1.5.0 and cats-effect 1.2.0
```
Daenyth: This seems odd also. 2.0.0 depends on cats-effect 2.x, not 1.2.0. 

The "Where to get the latest version" section above also appears outdated.

mpilquist:
Thanks for fixing this! Like @Daenyth mentioned, there are a few other things out of date here, like references to upstream versions and even mentioning the latest 2.1.0 release. Let's merge this as-is and fix the other stuff in a follow up.

Issue #1602, opened 2019-09-08
Split Stream.scala and StreamSpec.scala into smaller files


kubukoz:
I imagine Stream's companion object could extend a trait that'd be in a separate file, for starters. Then it could be split into StreamInstances and StreamFunctions.

StreamSpec could be split much easier, by groups of related tests (e.g. RegressionSpec, EvalMapTests etc.).

The reasons for this suggestion:

- incremental compilation is slow due to having to recompile the whole file which is thousands of lines long
- formatting takes a looot of time
- `testOnly` is very slow if you can't specify specific test names (i.e. in bloop)
- (probably not so relevant) the files are impossible to browse on GH with mobile browsers as they hang the whole application.

mpilquist:
Re: tests, try this: `testOnly StreamSpec -- -z evalMap` -- this will run all tests with the word `evalMap` in test name.

It would be nice to improve incremental compilation and speed up formatting. We could move formatting to a CI check instead of formatting on save perhaps. I'm not a big fan of `StreamInstances` and `StreamFunctions` as I find it difficult to find stuff and difficult to move various constructors and functions around so it's accessible everywhere it's needed to make the file modularization work.

SystemFw:
> I'm not a big fan of StreamInstances and StreamFunctions as I find it difficult to find stuff and difficult to move various constructors and functions around so it's accessible everywhere it's needed to make the file modularization work.

There is definitely some level of personal preference here, but I agree with Michael. I really care about having  a tiny top level scope (in terms of files), and quickly navigate the structure of the library, at the cost of bigger files. Compare the top level of fs2 0.9 and what we have now and you'll see this tendency at work.
 Others (for example Alex) prefer having many small files (you can see that in cats-effect `internal`, one file per op)

kubukoz:
> Re: tests, try this: `testOnly StreamSpec -- -z evalMap` -- this will run all tests with the word `evalMap` in test name.

It works in sbt, but not bloop :/.

> We could move formatting to a CI check instead of formatting on save perhaps

A CI *check* will be frustrating. A CI auto-commit to upstream that does formatting would be much better. But I haven't seen that work so far.

I can see some benefit in having everything in a single place, but in the age of IDEs and Metals it's also easy to search for a symbol and navigate to it. It doesn't help discoverability though.

[...]

PR #1688, opened 2019-11-22
Disabled scalafmt on save, added fmt alias, configured travis to run fmtCheck

mpilquist:
Folks will have to remember to run `fmt` prior to creating a PR, similar to the Cats PR process.


SystemFw:
What's the motivation behind this? Has it become an annoyance for you?

mpilquist:
Motivation was increasing incremental compile time - #1602.

SystemFw:
That makes sense, and I had missed that you added the `check` in `CI`, my main objection was me forgetting to run formatting before merge :P

PR #1703, opened 2019-12-06
Drop support for Scala 2.11

diesalbla:
Lightbend released the last version of Scala 2.11 almost two years ago, and both `cats` and `cats-effect` are also dropping it.

https://github.com/typelevel/cats-effect/pull/713
https://github.com/typelevel/cats-effect/issues/667
https://github.com/typelevel/cats/pull/3051

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: scala
 
 scala:
-  - 2.11.12
   - 2.12.9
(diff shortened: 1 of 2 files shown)


mpilquist:
Thanks -- I'll bump version.sbt to 2.2.0 after merging.

PR #1708, opened 2019-12-10
Small typo in comment

labra:
I changed from "a stream that will emits a single..." to "a stream that will emit a single..."


mpilquist:
Thanks!

PR #1718, opened 2019-12-22
Replaced tut with mdoc and fixed microsite publishing

mpilquist:


build.sbt
```diff
@@ -384,6 +384,9 @@ lazy val microsite = project
     micrositeGithubOwner := "functional-streams-for-scala",
     micrositeGithubRepo := "fs2",
     micrositeBaseUrl := "",
+    micrositeTheme := "pattern",
+    micrositeHighlightTheme := "solarized-light",
+    micrositeExtraMdFilesOutput := resourceManaged.value / "main" / "jekyll",
```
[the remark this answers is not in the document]
mpilquist: Without the `main`, the `index.md` file was generated in the wrong directory and wasn't included in the generated site output.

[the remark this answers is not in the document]
gvolpe: Interesting. I had to add this too but without that `"main"` to get the `index.md` in the right place...

docs/ReadmeExample.md
```diff
@@ -66,68 +58,73 @@ def fahrenheitToCelsius(f: Double): Double =
...
+// written: Stream[IO, Unit] = Stream(..)
 ```
 
 There are a number of ways of interpreting the stream. In this case, we call `compile.drain`, which returns a val value of the effect type, `IO`. The output of the stream is ignored - we compile it solely for its effect.
 
 ```scala
-scala> val task: IO[Unit] = written.compile.drain
-task: cats.effect.IO[Unit] = <function1>
+val task: IO[Unit] = written.compile.drain
+// task: IO[Unit] = Map(
```
olafurpg: FYI, there is a `mdoc:to-string` modifier to get the same formatted output as tut

olafurpg: https://scalameta.org/mdoc/docs/modifiers.html#to-string

mpilquist: Thanks!

gvolpe:
@mpilquist opting for `micrositeTheme := "pattern"` was intentional? The new default design of Mdoc's microsites are quite nice but I understand if you prefer to stick to the old fashion one :)

mpilquist:
@gvolpe Only because the new default theme resulted in losing the FS2 logo. I started down the path of adjusting image names but decided to take the path of least resistance and revert to the old default.

gvolpe:
I imagined it was because of that because the same happened to me! :smile: 

Maybe I can help to get the required 36x36 SVG logo to pick the new style, if you are okay with it :)

I managed to do it with Fs2Rabbit e.g. https://fs2-rabbit.profunktor.dev/

mpilquist:
Yep definitely okay with it. Would be very much appreciated!

gvolpe:
https://github.com/functional-streams-for-scala/fs2/pull/1722

PR #1722, opened 2019-12-30
Update Microsite style

gvolpe:
This is how it looks now:

one

two

I also had to resize the `favicon` to 32x32 otherwise it doesn't render.


mpilquist:
Thanks!

Zelenya:
@gvolpe the pngs didn't work anymore? Does it require svgs now? I've changed the icons to a cleaned up version. Now there is a mix of old and new logos. I'd like to fix it

gvolpe:
@Zelenya right, the new microsite used Mdoc, which requires the logo to be an SVG file of 36x36 (two files, actually) and the `favicon` should be a PNG of 32x32.

It would be great if you could open a PR with better quality logos!

Zelenya:
@gvolpe :+1:

PR #1723, opened 2019-12-31
Update microsite svgs

Zelenya:


Zelenya:
@gvolpe Could you please double check if it works where it should

PR #1724, opened 2020-01-03
Enabled fatal-warnings, fixed 2.13 compiler warnings, removed benchmark macros, and fixed concat issue in Chunk

mpilquist:
Some notes:
 - I removed the benchmark macros b/c Scala 3 doesn't support them and we can get approximately the same behavior using standard JMH features
 - Bloop or one of its dependencies was getting confused by the `+` in the `2.13+` directory name -- likely treating it as a regex char or something.


benchmark/src/main/scala/fs2/benchmark/StreamBenchmark.scala
```diff
@@ -2,20 +2,29 @@ package fs2
...
+  Param,
+  Scope,
+  State
+}
 import java.util.concurrent.TimeUnit
 
 @State(Scope.Thread)
 class StreamBenchmark {
-  @GenerateN(1, 4096, 665536, 67108863)
+  @Param(Array("10", "100", "1000", "10000"))
```
pauljamescleary: this array (10, 100, 1000...) is different than the old.  Looks like the old was on specific boundaries, i wonder if there was a specific test in mind.

Certainly, 10,000 will be different than 67,108,863 as a max upper bounds?

benchmark/src/main/scala/fs2/benchmark/ConcurrentBenchmark.scala
```diff
@@ -2,20 +2,22 @@ package fs2
...
-import org.openjdk.jmh.annotations.{Benchmark, Scope, State}
+import org.openjdk.jmh.annotations.{Benchmark, Param, Scope, State}
 
 @State(Scope.Thread)
 class ConcurrentBenchmark {
   implicit val cs: ContextShift[IO] =
     IO.contextShift(scala.concurrent.ExecutionContext.Implicits.global)
 
-  @GenerateN(1, 2, 4, 7, 16, 32, 64, 128, 256)
+  @Param(Array("1", "16", "256"))
```
pauljamescleary: Curious as to why you lowered the number of possibilities here

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -1466,8 +1471,13 @@ object Chunk extends CollectorK[Chunk] {
...
     concatBooleans(chunks, chunks.foldLeft(0)(_ + _.size))
 
+  /**
+    * Concatenates the specified sequence of boolean chunks in to a single chunk.
+    * The `totalSize` parameter must be equal to the sum of the size of each chunk or
+    * otherwise an exception may be thrown.
+    */
   def concatBooleans(chunks: GSeq[Chunk[Boolean]], totalSize: Int): Chunk[Boolean] =
-    if (chunks.isEmpty) Chunk.empty
+    if (totalSize == 0) Chunk.empty
```
pauljamescleary: the validation check / code pattern repeats itself a lot, does it make sense to have a HOF to wrap each of the concats in?

- totalSize check is only to 0, any need to check a negative value?
- are there tests on this?  (this is totes a newb question)

benchmark/src/main/scala/fs2/benchmark/ConcurrentBenchmark.scala
mpilquist: Running the JMH suite takes hours and I don't typically look at many of the intermediate results -- just compare the individual cases enough to get a feel for asymptotic growth.

core/shared/src/main/scala/fs2/Chunk.scala
matthughes: Yeah I don't know how ugly you consider ClassTag or the cost of this indirection but you could dedupe all these methods with this private method:

```scala
  def concatBooleans(chunks: GSeq[Chunk[Boolean]], totalSize: Int): Chunk[Boolean] =
    concatGen[Boolean](chunks, totalSize, Chunk.booleans)
  
  private def concatGen[F: ClassTag](chunks: GSeq[Chunk[F]], totalSize: Int, chunkBuilder: Array[F] => Chunk[F]): Chunk[F] =
    if (chunks.isEmpty) Chunk.empty
    else {
      val arr = new Array[F](totalSize)
      var offset = 0
      chunks.foreach { c =>
        if (!c.isEmpty) {
          c.copyToArray(arr, offset)
          offset += c.size
        }
      }
      chunkBuilder(arr)
    }
```

benchmark/src/main/scala/fs2/benchmark/StreamBenchmark.scala
mpilquist: Good catch, fixed

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: No need to check totalSize as it's passed to `new Array(..)` and that will handle it.

mpilquist: I consolidated those primitive methods to a single private concatUnboxed

benchmark/src/main/scala/fs2/benchmark/RangeFoldBenchmark.scala
```diff
@@ -0,0 +1,19 @@
...
+
+@State(Scope.Thread)
+class RangeFoldBenchmark {
+  @Param(Array("1", "4096", "665536", "67108863"))
+  var n: Int = _
+
+  @Benchmark
+  def rangeFold(): Option[Long] =
+    Stream.range(0, n).fold(0L)(_ + _.toLong).compile.last
+}
```
pauljamescleary: There has to be a `gif` for newline characters by this time.  What is the Github method of invoking a `gif`?

benchmark/src/main/scala/fs2/benchmark/StreamBenchmark.scala
```diff
@@ -48,10 +51,9 @@ class StreamBenchmark {
       .unsafeRunSync
       .get
 
-  @GenerateN(1, 10, 100, 1000, 10000, 100000)
   @Benchmark
-  def rightAssocFlatMap(N: Int): Int =
-    (0 until N)
+  def rightAssocFlatMap(n: Int): Int =
```
pauljamescleary: shouldn't this benchmark use the member Param `n` as well?

benchmark/src/main/scala/fs2/benchmark/StreamBenchmark.scala
```diff
@@ -61,48 +63,26 @@ class StreamBenchmark {
...
-  def emitsThenFlatMap(N: Int): Vector[Int] =
-    Stream.emits(0 until N).flatMap(Stream(_)).toVector
+  def emitsThenFlatMap(): Vector[Int] =
+    Stream.emits(0 until n).flatMap(Stream(_)).toVector
 
-  @GenerateN(1, 10, 100, 1000, 10000)
   @Benchmark
-  def sliding(N: Int) =
-    Stream.emits(0 until 16384).sliding(N).covary[IO].compile.drain.unsafeRunSync
+  def sliding(n: Int) =
```
pauljamescleary: and the following as well?

benchmark/src/main/scala/fs2/benchmark/RangeFoldBenchmark.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import org.openjdk.jmh.annotations.{Benchmark, Param, Scope, State}
+
+@State(Scope.Thread)
+class RangeFoldBenchmark {
+  @Param(Array("1", "4096", "665536", "67108863"))
+  var n: Int = _
+
+  @Benchmark
+  def rangeFold(): Option[Long] =
```
pauljamescleary: General style question, is it better to add another `@Param` to the existing benchmark that is on `Stream` which organizes benchmarks by type, or separate it out like this>

benchmark/src/main/scala/fs2/benchmark/StreamBenchmark.scala
mpilquist: Yep, fixed

benchmark/src/main/scala/fs2/benchmark/RangeFoldBenchmark.scala
mpilquist: When adding an additional param, JMH will run the cross product -- e.g., n = 1, 2, 3 and m = 4, 5, 6 will cause JMH to run 9 cases (1, 4), (1, 5), (1, 6), (2, 4), ...

pauljamescleary: Ahh, boooo

PR #1725, opened 2020-01-03
Fixed #1709 - .to(Collector) syntax fix for 2.12

### LINKED ISSUE #1709: The new collector syntax doesn't work with Scala 2.12 when the result type is explicitly specified
I'm not sure whether this should be treated as an fs2 issue or a Scala issue... Probably, both.

```scala
scala> Stream.empty[Id].covaryOutput[Byte].compile.to(Chunk)
res1: fs2.Chunk[Byte] = empty

scala> Stream.empty[Id].covaryOutput[Byte].compile.to(Chunk): Chunk[Byte]
<console>:21: error: type mismatch;
 found   : fs2.Chunk.type
 required: fs2.internal.Factory[Byte,fs2.Chunk[Byte]]
    (which expands to)  scala.collection.generic.CanBuildFrom[Nothing,Byte,fs2.Chunk[Byte]]
       Stream.empty[Id].covaryOutput[Byte].compile.to(Chunk): Chunk[Byte]
                                                      ^

scala> val c: Collector.Aux[Byte, Chunk[Byte]] = Chunk
c: fs2.Collector.Aux[Byte,fs2.Chunk[Byte]] = fs2.CollectorK$$anon$6@2860828f

scala> Stream.empty[Id].covaryOutput[Byte].compile.to(c): c.Out
<console>:22: error: type mismatch;
 found   : fs2.Collector.Aux[Byte,fs2.Chunk[Byte]]
    (which expands to)  fs2.Collector[Byte]{type Out = fs2.Chunk[Byte]}
 required: fs2.internal.Factory[Byte,fs2.Chunk[Byte]]
    (which expands to)  scala.collection.generic.CanBuildFrom[Nothing,Byte,fs2.Chunk[Byte]]
       Stream.empty[Id].covaryOutput[Byte].compile.to(c): c.Out
                                                      ^
```

I've managed to find this magic workaround:

```scala
scala> val s = Stream.empty[Id].covaryOutput[Byte].compile.to(Chunk)
s: fs2.Chunk[Byte] = empty

scala> s: Chunk[Byte]
res6: fs2.Chunk[Byte] = empty
```

Adding a non-overloaded synonym for `to(Collector[A])` (as `toCollector` or something else) might be enough to fix this (I haven't tried it yet).

mpilquist:
`.to` was overloaded -- one took the `Collector` and the other took an implicit factory. When the target type was known, the 2.12 compiler was picking the factory overload. To fix, I removed the old overload -- or rather, made it `private[Stream]` so that it cannot be called but it stays around for binary compatibility reasons.

Note that in practice, any `s.compile.to[C]` can be written as `s.compile.to(C)` b/c `Collector` can be implicitly instantiated from any `Factory`. Only case where that's not true is when `C` doesn't implement `Factory` directly and instead, there's some implicit factory in scope.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3729,7 +3729,7 @@ object Stream extends StreamLowPriority {
     @inline private def to_(c: Collector[O]): c.Out =
       self.covary[SyncIO].compile.to(c).unsafeRunSync
 
-    /** Runs this pure stream and returns the emitted elements in a collection of the specified type. Note: this method is only available on pure streams. */
+    // TODO Delete this in 3.0
     private[Stream] def to[C[_]](implicit f: Factory[O, C[O]]): C[O] = to_(f)
 
     /** Runs this pure stream and returns the emitted elements in a chunk. Note: this method is only available on pure streams. */
@@ -3788,7 +3788,7 @@ object Stream extends StreamLowPriority {
     @inline private def to_(c: Collector[O]): Either[Throwable, c.Out] =
       lift[SyncIO].compile.to(c).attempt.unsafeRunSync
 
-    /** Runs this fallible stream and returns the emitted elements in a collection of the specified type. Note: this method is only available on fallible streams. */
+    // TODO Delete this in 3.0
     private[Stream] def to[C[_]](implicit f: Factory[O, C[O]]): Either[Throwable, C[O]] = to_(f)
 
     /** Runs this fallible stream and returns the emitted elements in a chunk. Note: this method is only available on fallible streams. */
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2636,13 +2636,6 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Nothing, O, Unit]
   )(f: (Stream[F, O], Stream[F2, O2]) => Stream[F2, O3]): Stream[F2, O3] =
     f(this, s2)
 
-  /**
-    * Applies the given sink to this stream.
-    */
-  @deprecated("Use .through instead", "1.0.2")
-  private[fs2] def to[F2[x] >: F[x]](f: Stream[F, O] => Stream[F2, Unit]): Stream[F2, Unit] =
```
mpilquist: This is the reason for the mima exclusion. Not related to #1709 but noticed we can remove this safely now.

PR #1727, opened 2020-01-04
Added debug and debugChunks combinators

### LINKED ISSUE #1569: `showLinesStdOut` and friends could keep the elements unchanged
Currently `showLinesStdOut` uses `evalMap` to print each element, discarding its value. I think these operators would be much more useful if they kept the original elements (so that there's a nice and easy way to debug the values of a stream at a stage without having to reimplement `showLinesStdOut` with `evalTap`).

mpilquist:
Fixes #1569


mpilquist:
@kubukoz How about this?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -656,6 +656,69 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Nothing, O, Unit]
...
+    * @example {{{
+    * scala> Stream(1, 2).append(Stream(3, 4)).debug("a").toList
+    * a: 1
+    * a: 2
+    * a: 3
+    * a: 4
+    * res0: List[Int] = List(1, 2, 3, 4)
+    * }}}
+    */
+  def debug(tag: String = "", logger: String => Unit = println(_)): Stream[F, O] =
```
kubukoz: Maybe unsafeLogger?

kubukoz:
Looks very nice, I'll see what it looks like in some examples :) Although I'm starting to get worried about having so many methods (4 of them being for debugging). Maybe it's a good time to think about putting them in some importable object (as pipes), or an implicit class...

mpilquist:
Yeah, it's a tricky balance. I could see getting rid of the format varieties, leaving two methods. And moving them to somewhere else kind of defeats the point of having a quick / easily accessible way of debugging.

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: Hm, the call site seems clunkier then -- `s.debug(unsafeLogger = println)`

kubukoz: Just thinking about side effects in the API. Arguably `debug` itself suggests that there will be something shady going on (e.g. `toString` for formatting),  so maybe `logger` will also be an acceptable trade-off :)

kubukoz:
Good point, it's probably a case of something you want to access quickly... but the API growing might eventually be a problem so we can think about it next time when adding new combinators :) 

Just checked the new functions, everything seemed fine - 👍 from me

SystemFw:
How about only leaving the `format*` signatures instead? `_ => "myTag"` does not seem very problematic if `O` infers

kubukoz:
@SystemFw it's more like `"myTag: " + _`, but fair enough

PR #1730, opened 2020-01-05
FreeC: make the F parameter covariant 

diesalbla:
Follows on the early work on https://github.com/functional-streams-for-scala/fs2/pull/1615.

In the internal implementation of streams, we change the definition of the `FreeC[F[_], +O, +R]` generalised abstract data type (GADT), to turn the `F` into a covariant parameter of higuer-kind, that is, `FreeC[ + F[_], +O, +R ]`.

- _Pure_ subclasses: those subclasses of `FreeC` that do not involve `F`, such as the `Result` and the `Output`, can be specialised with `F = Pure` (an alias for `Nothing` of the `F` kind). We can then integrate `Result` into the `FreeC` GADT, and drop `asFreeC`.  

- The classes `Pull` and `Stream` no longer need to resort to the `Nothing` parameter and use type-casts. Instead, the type-parameters of `Stream` and `FreeC` are no aligned. As a consequence, we will no longer need to use the `covary` methods of `Stream` or `Pull`. 

### Broken edge for Scala 2.12.

When we made the first changes, we noticed that pattern-matching on the algebra classes `Step` and `OpenScope` was causing a _compiler crash_. This problem has been fixed in Scala 2.13.  However, in order to keep support for 2.12, we need to use a small hack for these two case classes: 
- In `Step`, we have to replace an appearance of `F` by `Any`, on the definition, and use some type-casts (`asInstanceOf`) when using that field.
- In `OpenScope`, we use type parameter `G[_]` for the `Concurrent` instance, different to the `F` from the `FreeC`. Of course, we know from the rest of the program that the `F` should be the `G`.
 
### Changes for a subsequent Pull Request 

To keep this Pull Request small, there are some pending changes to be done later on: 
- Remove any use of the internal `covary`, `get`, and `fromFreeC` methods in `Pull` and `Sstream`
- Consider turning `Pull` into a type alias (not just an `AnyVal` wrapper) of `FreeC`, closer to the vision of the `fs3` branch.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -205,6 +205,16 @@ lazy val mimaSettings = Seq(
     }.toSet
   },
   mimaBinaryIssueFilters ++= Seq(
+    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.PurePipeOps"),
+    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.PurePipe2Ops"),
+    ProblemFilters.exclude[MissingClassProblem]("fs2.Stream$PurePipeOps$"),
+    ProblemFilters.exclude[MissingClassProblem]("fs2.Stream$PurePipe2Ops$"),
+    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.PurePipe2Ops"),
+    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.PurePipeOps"),
+    // PurePipeOpe removed
+    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.covaryPurePipe"),
+    ProblemFilters.exclude[MissingClassProblem]("fs2.Stream$PurePipeOps"),
+    ProblemFilters.exclude[MissingClassProblem]("fs2.Stream$PurePipe2Ops"),
     // No bincompat on internal package
     ProblemFilters.exclude[Problem]("fs2.internal.*"),
     // Mima reports all ScalaSignature changes as errors, despite the fact that they don't cause bincompat issues when version swapping (see https://github.com/lightbend/mima/issues/361)
--- a/build.sbt
+++ b/build.sbt
@@ -205,16 +205,6 @@ lazy val mimaSettings = Seq(
     }.toSet
   },
   mimaBinaryIssueFilters ++= Seq(
-    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.PurePipeOps"),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.PurePipe2Ops"),
-    ProblemFilters.exclude[MissingClassProblem]("fs2.Stream$PurePipeOps$"),
-    ProblemFilters.exclude[MissingClassProblem]("fs2.Stream$PurePipe2Ops$"),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.PurePipe2Ops"),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.PurePipeOps"),
-    // PurePipeOpe removed
-    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.covaryPurePipe"),
-    ProblemFilters.exclude[MissingClassProblem]("fs2.Stream$PurePipeOps"),
-    ProblemFilters.exclude[MissingClassProblem]("fs2.Stream$PurePipe2Ops"),
     // No bincompat on internal package
     ProblemFilters.exclude[Problem]("fs2.internal.*"),
     // Mima reports all ScalaSignature changes as errors, despite the fact that they don't cause bincompat issues when version swapping (see https://github.com/lightbend/mima/issues/361)
--- a/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
+++ b/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
@@ -52,24 +52,24 @@ object ResourceTrackerSanityTest extends App {
 }
 
 object RepeatPullSanityTest extends App {
-  def id[A]: Pipe[Pure, A, A] = _.repeatPull {
+  def id[F[_], A]: Pipe[F, A, A] = _.repeatPull {
     _.uncons1.flatMap {
       case Some((h, t)) => Pull.output1(h).as(Some(t));
       case None         => Pull.pure(None)
     }
   }
-  Stream.constant(1).covary[IO].through(id[Int]).compile.drain.unsafeRunSync()
+  Stream.constant(1).covary[IO].through(id[IO, Int]).compile.drain.unsafeRunSync()
 }
 
 object RepeatEvalSanityTest extends App {
-  def id[A]: Pipe[Pure, A, A] = {
-    def go(s: Stream[Pure, A]): Pull[Pure, A, Unit] =
+  def id[F[_], A]: Pipe[F, A, A] = {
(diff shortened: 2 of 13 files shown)


build.sbt
```diff
@@ -205,6 +205,16 @@ lazy val mimaSettings = Seq(
     }.toSet
   },
   mimaBinaryIssueFilters ++= Seq(
+    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.PurePipeOps"),
```
mpilquist: To preserve binary compatibility, we should deprecate this stuff until 3.0 release -- I think that will coincide with cats-effect 3, probably in the 6+ month timeframe.

core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
```diff
@@ -52,24 +52,24 @@ object ResourceTrackerSanityTest extends App {
...
     }
   }
-  Stream.constant(1).covary[IO].through(id[Int]).compile.drain.unsafeRunSync()
+  Stream.constant(1).covary[IO].through(id[IO, Int]).compile.drain.unsafeRunSync()
 }
 
 object RepeatEvalSanityTest extends App {
-  def id[A]: Pipe[Pure, A, A] = {
-    def go(s: Stream[Pure, A]): Pull[Pure, A, Unit] =
+  def id[F[_], A]: Pipe[F, A, A] = {
```
mpilquist: Did these have to change? If so, could you update the PR with a summary of public facing source compatibility API changes?

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -161,8 +160,8 @@ object Pull extends PullLowPriority {
     * Returns a pull that evaluates the supplied by-name each time the pull is used,
     * allowing use of a mutable value in pull computations.
     */
-  def suspend[F[x] >: Pure[x], O, R](p: => Pull[F, O, R]): Pull[F, O, R] =
-    fromFreeC(FreeC.suspend(p.get))
+  def suspend[F[_], O, R](p: => Pull[F, O, R]): Pull[F, O, R] =
```
mpilquist: The `F[x] >: Pure[x]` bound is to prevent inference of `Nothing`. Any reason this changes here?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2982,10 +2981,10 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Nothing, O, Unit]
 
 object Stream extends StreamLowPriority {
   @inline private[fs2] def fromFreeC[F[_], O](free: FreeC[F, O, Unit]): Stream[F, O] =
-    new Stream(free.asInstanceOf[FreeC[Nothing, O, Unit]])
+    new Stream(free)
 
   /** Creates a pure stream that emits the supplied values. To convert to an effectful stream, use `covary`. */
-  def apply[F[x] >: Pure[x], O](os: O*): Stream[F, O] = emits(os)
+  def apply[O](os: O*): Stream[Pure, O] = emits(os)
```
mpilquist: I think this one has the potential to impact lots of folks, though maybe not -- perhaps folks typically use `Stream(...).covary[F]`

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: Do you have an example code snippet where lack of this `>: Pure[x]` was causing a problem?

In general, I would think that covariance of both `Pull` and `FreeC` would make locally inferring `F[A] = Nothing` not a problem, since it can up-cast it naturally when used in another context.

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: Note that, with a covariant `F`, we can deprecate the `covary` methods in favour of type-ascriptions, such as ```Scala
Stream(1): Stream[IO, Int]
```

Comment (kubukoz) [reply core/shared/src/main/scala/fs2/Stream.scala]: I'm 👎 on deprecating `covary`, it's much more handy than type ascriptions (especially because you only need to type in the effect type and the element type is inferred).

I rarely use `apply` with an explicit type though, so I'm okay with this particular change :)

File: core/shared/src/main/scala/fs2/Pull.scala (hunk 3 above)
Comment (mpilquist) [reply core/shared/src/main/scala/fs2/Pull.scala]: Right, but `Stream` and `Pull` were already covariant in `F`. I can try to come up with a failing example later, but take a look at this blog post which documents why we had to create `Pure` in the first place: https://mpilquist.github.io/blog/2018/07/04/fs2/. Basically, there are cases where inferring `F = Nothing` causes scalac to do the wrong thing.

File: core/shared/src/main/scala/fs2/Stream.scala (hunk 4 above)
Comment (mpilquist) [reply core/shared/src/main/scala/fs2/Stream.scala]: Yeah I much prefer `covary[F]` over type ascription as I don't have to repeat the output type -- and it's worse with pull where you have to specify both output and result types.

File: core/jvm/src/test/scala/fs2/MemorySanityChecks.scala (hunk 2 above)
Comment (diesalbla) [reply core/jvm/src/test/scala/fs2/MemorySanityChecks.scala]: So, at some point in writing these changes, I thought I may need to change it. It 

Now that the `F` is covariant and not invariant, I am not sure if there is still much use in keeping the idea of a `Pipe` with a same parameter for the input and output stream. Ideally, you may want the pipe to be applied to any sub-F-type of the input, and be compatible with any super-F-type of the output (just like any function).

Comment (mpilquist) [reply core/jvm/src/test/scala/fs2/MemorySanityChecks.scala]: `F` in `Stream` and `Pull` have always been covariant though, so changing internals shouldn't cause changes to the API/usage.

File: core/shared/src/main/scala/fs2/Stream.scala [hunk 5]
```diff
@@ -2982,10 +2981,10 @@ final class Stream[+F[_], +O] private (private val free: FreeC[Nothing, O, Unit]
 
 object Stream extends StreamLowPriority {
   @inline private[fs2] def fromFreeC[F[_], O](free: FreeC[F, O, Unit]): Stream[F, O] =
-    new Stream(free.asInstanceOf[FreeC[Nothing, O, Unit]])
+    new Stream(free)
 
   /** Creates a pure stream that emits the supplied values. To convert to an effectful stream, use `covary`. */
-  def apply[F[x] >: Pure[x], O](os: O*): Stream[F, O] = emits(os)
+  def apply[F[x] >: Pure[x], O](os: O*): Stream[Pure, O] = emits(os)
```
mpilquist: Should return `Stream[F, O]`

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4585,7 +4582,7 @@ object Stream extends StreamLowPriority {
   }
 
   /** Provides operations on pure pipes for syntactic convenience. */
-  implicit final class PurePipeOps[I, O](private val self: Pipe[Pure, I, O]) extends AnyVal {
+  implicit final class PurePipeOpsf[I, O](private val self: (Stream[Pure, I] => Stream[Pure, O])) extends AnyVal {
```
mpilquist: Typo: trailing `f` on `PurePipeOps`

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: I have undone the change to the `apply` method. In general, I am undoing any change to the public API that is not strictily necessary to introduce the covariance in this Pull Request.

mpilquist:
I reverted a few other unnecessary changes in StreamSpec, CompilationTest, and the README. Merging on successful Travis build. Thanks for all the work on this one, it looks great!

PR #1731, opened 2020-01-07
Stream/Pull Internals: purge fromFreeC/get converters.

diesalbla:
The internal methods `fromFreeC` and `get` of Pull and Stream were needed at first to convert between the wrapped `FreeC` and them. This conversion was needed due to the difference between `FreeC` invariant type-parameter `F[_]`, and Stream/Pull covariant `+F[_]`. Now that `FreeC` is also covariant on F[_], we can drop these converters. Also, a few uses of `covary` and extra type parameters can be simplified as well.

Since these methods were private to fs2, but not entirely private, it turns out that Mima detects them as missing methods. We have to mark them as exceptions in the build

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -24,16 +24,15 @@ import fs2.internal.Algebra.Eval
   * `raiseError` is caught by `handleErrorWith`:
   *   - `handleErrorWith(raiseError(e))(f) == f(e)`
   */
-final class Pull[+F[_], +O, +R] private (private[fs2] val free: FreeC[F, O, R]) extends AnyVal {
-  private[fs2] def get[F2[x] >: F[x], O2 >: O, R2 >: R]: FreeC[F2, O2, R2] =
-    free
+final class Pull[+F[_], +O, +R] private[fs2] (private[fs2] val free: FreeC[F, O, R])
+    extends AnyVal {
 
   /** Alias for `_.map(_ => o2)`. */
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -116,20 +113,19 @@ object Pull extends PullLowPriority {
...
   def output[F[x] >: Pure[x], O](os: Chunk[O]): Pull[F, O, Unit] =
-    if (os.isEmpty) Pull.done else fromFreeC(Algebra.Output[O](os))
+    if (os.isEmpty) Pull.done else new Pull(Algebra.Output[O](os))
 
   /** Pull that outputs nothing and has result of `r`. */
   def pure[F[x] >: Pure[x], R](r: R): Pull[F, INothing, R] =
-    fromFreeC[F, INothing, R](Result.Pure(r))
+    new Pull(Result.Pure(r))
+
+  val unit: Pull[Pure, INothing, Unit] = new Pull(Result.unit)
```
mpilquist: This is redundant with `Pull.done` -- I prefer the name `Pull.unit` but `done` goes back to the old 0.9 API. I'm mildly against adding an alias since I don't like to provide multiple ways of doing the same thing but maybe this is a good exception to that rule considering `done` is kind of a strange name. Thoughts?

mpilquist: Oh and if we keep this, can you add ScalaDoc similar to `done`?

diesalbla: I think I am going to drop this. I have other changes in mind for `Pull`.

PR #1733, opened 2020-01-08
Rename some of the Classes of FreeC/Algebra.

diesalbla:
- We rename the "Result.Pure" class, which just wraps a value of the
  type `R`, to `Result.Value`, because it's just a value of type R.

- We rename the `Eval` class, which takes an `F[R]` into an `R`,
  as Lift: we are lifting an effect-ful `F[R]` into the "Stream"

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/FreeCBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/FreeCBenchmark.scala
@@ -15,7 +15,7 @@ class FreeCBenchmark {
   @Benchmark
   def nestedMaps = {
     val nestedMapsFreeC =
-      (0 to N).foldLeft(Result.Pure[Int](0): FreeC[IO, INothing, Int]) { (acc, i) =>
+      (0 to N).foldLeft(Result.Value[Int](0): FreeC[IO, INothing, Int]) { (acc, i) =>
         acc.map(_ + i)
       }
     run(nestedMapsFreeC)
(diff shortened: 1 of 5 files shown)


mpilquist:
:+1: on Value, :-1: on Lift -- we use "eval" a lot in the public API so I'd rather be consistent in `FreeC`

diesalbla:
@mpilquist So, at present there is a collision between two classes called `Eval`: 

- `final case class Algebra.Eval[F[_], R](fr: F[R])`, which I was suggesting to rename as `Lift`.
- `abstract FreeC.Eval[+F[_], +O, +R]`, the branch of `FreeC` that is neither `Bind` nor `Result`, and from which all classes inside `Algebra` extend from. In essence, this subclass adds little, and is only used inside the `View` sub-classes to mark "this is neither result nor bind".

The goal would be to avoid having this name duplication. I agree that the former is coherent with the `eval` and `eval_` methods. Could you perhaps suggest a new name for the latter?

mpilquist:
Oh thanks for that clarification. `cats.free.Free` uses `Suspend` for what is currently `FreeC.Eval`. What do you think of that?

diesalbla:
I am guessing that `cats.free.Free` is a good inspiration. However, since there is a `suspend` method in  `FreeC`, `Stream`, and `Pull`, would that add some confusion? 
It is an improvement, though.

mpilquist:
Yeah maybe that's confusing. I'm fine with either FreeC.Lift or FreeC.Suspend - whatever you prefer.

SystemFw:
I actually prefer namespacing, which is to say `FreeC.Eval` and `Algebra.Eval`, to renaming (which is kinda like custom name mangling).
I would also like to humbly suggest some moderation with all the renames if there are no semantic changes, I think it makes reading the history of the changes in a given file needlessly confusing (when we want to reason about semantics changes to e.g. scoping)

diesalbla:
The reason behind renaming `Algebra.Eval` is to be able to merge the two files `FreeC` and `Algebra` into a single module, so as to _seal_ this GADT again. Currently, the `FreeC` GADT is not sealed, but on the other hand it is also closed by being `private[fs2]`.

PR #1734, opened 2020-01-09
Unify FreeC and Pull: first attempt.

diesalbla:
After removing `fromFreeC` and `get`, we have observed that many of the operations in Pull were directly mapped to the inner FreeC, whereas the others were based on those operations.
Also, given that `FreeC` and `Pull` have effectively the same parameters, we can unify them.

this commit tries to unify `Pull` and `FreeC` under the hood, by turning `Pull` into an abstract class (with abstract methods) and `FreeC` into the only direct subclass of Pull. One inconvenience is that, in some methods that are defined in `Pull` (and thus are public), to implement them in `FreeC`, we need to down-cast them using an `asFreeC` method. 

The reason for using this way is to move slowly to the FS3 design: Pull as FreeC, Stream as wrapper on a Pull.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -24,15 +24,29 @@ import fs2.internal.Algebra.Eval
   * `raiseError` is caught by `handleErrorWith`:
   *   - `handleErrorWith(raiseError(e))(f) == f(e)`
   */
-final class Pull[+F[_], +O, +R] private[fs2] (private[fs2] val free: FreeC[F, O, R])
-    extends AnyVal {
+abstract class Pull[+F[_], +O, +R] private[fs2] () {
+  private[fs2] def asFreeC: FreeC[F, O, R] = this.asInstanceOf[FreeC[F, O, R]]
+
+  /** Applies the resource of this pull to `f` and returns the result. */
+  def flatMap[F2[x] >: F[x], O2 >: O, R2](f: R => Pull[F2, O2, R2]): Pull[F2, O2, R2]
+
+  /** Applies the resource of this pull to `f` and returns the result in a new `Pull`. */
+  def map[R2](f: R => R2): Pull[F, O, R2]
+
+  /** Applies the outputs of this pull to `f` and returns the result in a new `Pull`. */
+  def mapOutput[O2](f: O => O2): Pull[F, O2, R]
+
+  /** If `this` terminates with `Pull.raiseError(e)`, invoke `h(e)`. */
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -24,15 +24,29 @@ import fs2.internal.Algebra.Eval
   * `raiseError` is caught by `handleErrorWith`:
   *   - `handleErrorWith(raiseError(e))(f) == f(e)`
   */
-final class Pull[+F[_], +O, +R] private[fs2] (private[fs2] val free: FreeC[F, O, R])
-    extends AnyVal {
+abstract class Pull[+F[_], +O, +R] private[fs2] () {
```
mpilquist: Can constructor be totally private now instead of `private[fs2]`?

mpilquist:
This looks great. Will be even better when Pull & FreeC are merged completely. The mima errors are going to be a problem though -- we will likely need to delay the unification until FS2 3.0 unless there's some workaround.

diesalbla:
So, looking at the `Mima` problems, I find the following: 
```Scala
[error]  * extension method scan_$extension(fs2.internal.FreeC,java.lang.Object,scala.Function2)fs2.internal.FreeC in object fs2.Stream has a different result type in current version, where it is fs2.Pull rather than fs2.internal.FreeC

[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("fs2.Stream.scan_$extension")

[error]  * method stepLeg()fs2.internal.FreeC in class fs2.Stream#StepLeg has a different result type in current version, where it is fs2.Pull rather than fs2.internal.FreeC

[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("fs2.Stream#StepLeg.stepLeg")
```

From what I see, my guess is that currently, because both `Pull` and `Stream` are just wrappers on `FreeC`, the binary code (the generated .class) is actually referring to `FreeC` in most of its methods, not to `Pull` or `Stream`, which may not exist as classes in the bytecode. 
Note, also, that the `private[fs2]` modifier on `FreeC` is only relevant at the source level: at binary level, `FreeC` is a public class and most of its methods are public.

If we want to keep binary-compatibility, but at the same time bring `Pull` and `FreeC` more closely,  it may be easier to make `Pull` a type alias of `FreeC`; however, we may need to drop some source-compatibility on the `Pull` api.

mpilquist:
I think to maintain binary compatibility, we'll be forced to keep `Pull` as an `AnyVal` wrapper of `FreeC`. If we remove the wrapper, we won't be able to reproduce the boxing decisions that scalac makes with `AnyVal`s.

PR #1735, opened 2020-01-11
Unify FreeC, Algebra,

diesalbla:
This is a binary and source-incompatible change to FS2.

This is a draft pull request, intended to try out (with the CI) how things work, and get an overview of the structure with them.
______________________________

- We define Pull as the publicly-available top-level class. This class contains (in abstract form) four methods,  to `map`, `flatMap`, `mapOutput`, and `handleErrorWith`. This Pull class is sealed, to prevent other extensions to it.

- We add an object `PullImpl`, which is not publicly available, and which contains all the sub-classes of `Pull`, pulled from `FreeC` and `Algebra` files. At the top of these is a class `PullC`, which is similar to the previous `FreeC`.

- The new `PullImpl` object differs from the old  `FreeC` and `Algebra` in that 1) we rename the `Pure` result class to `Value`, and `FreeC.Eval` to `Action` (like suggested in #1733). Also, we remove the `Result` object.

#### Possible changes/adaptations: 

- The `Pull` class contains some of the methods from `FreeC`, marked as `private[fs2]` which we may want to turn into private functions of `PullImpl`.


mpilquist:
I'd like to see `PullImpl` merged in to `Pull` entirely instead of being in a separate object. For example, inlining the methods of `PullC` in to `Pull`, moving subtypes to `Pull` companion.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -65,46 +74,72 @@ final class Pull[+F[_], +O, +R] private[fs2] (private[fs2] val free: FreeC[F, O,
...
+      def cont(e: Result[R2]): Pull[F2, O2, R2] = e match {
+        case Fail(e) =>
+          try h(e)
+          catch { case NonFatal(e) => Fail(e) }
+        case other => other
+      }
+    }
 
   /** Discards the result type of this pull. */
   def void: Pull[F, O, Unit] = as(())
```
diesalbla: @mpilquist BTW, and off-main topic: Should we remove the definitions of `Pull.as` and `Pull.void`, since they are given _for free_ by the Functor syntax of `cats`?

mpilquist:
I’d rather keep them for discoverability

PR #1737, opened 2020-01-11
Unify instances of Sync and Monad for Pull.

diesalbla:
It seems that the implementation of the Sync type-class for Pull does
not require anything extra on the F or O, so it has same parameters as
the Monad instance, but same code. Chance to DRY it up.

This change should keep backwards-, but may break forwards-compatibility.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -161,26 +161,10 @@ object Pull extends PullLowPriority {
   /** `Sync` instance for `Pull`. */
   implicit def syncInstance[F[_], O](
       implicit ev: ApplicativeError[F, Throwable]
-  ): Sync[Pull[F, O, ?]] =
-    new Sync[Pull[F, O, ?]] {
-      def pure[A](a: A): Pull[F, O, A] = Pull.pure(a)
-      def handleErrorWith[A](p: Pull[F, O, A])(h: Throwable => Pull[F, O, A]) =
-        p.handleErrorWith(h)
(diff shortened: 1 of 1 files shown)


mpilquist:
The `ApplicativeError[F, Throwable]` constraint is to limit use of `raiseError` -- otherwise we claim to support `Sync[Pull[Pure, O, ?]]`. The evidence param is used to materialize a `RaiseThrowable[F]` so that we can call `Pull.raiseError` in the implementation of `ApplicativeError.raiseError`. In this refactoring, this issue is avoided by constructing the `Fail` result directly but I'm not sure we should do that.

diesalbla:
On a pragmatic side: note that this class-instance is hidden to fs2, not accessible outside, whereas the publicly accessible method does require that instance of Applicative-Error. 

On the theory-oriented (but not too rigurous not certain) side, this class indicates is that we do not need the evidence for `ApplicativeError[F, Throwable]` to raise an error _into the Pull_ datatype. Only when we interpret (`compile`) it, we need the ability to map a `Fail(exc)` into `F.raiseError(exc)`. But this is a bit like the `Monad[Pull[F, O, ?]]` instance not requiring an `F[_]: Monad` instance, just because we need it to map a `Result.Pure(x)` into a `F.pure(x)`. It is the `compile` method that asks for the `MonadError[F, O]` evidence. In other words, the `Pull` seems to be like a _free_-Sync data type.

mpilquist:
Yeah, sharing the implementation is fine. I was pointing out why we take the constraint.

Issue #1587, opened 2019-09-06
base64 text pipe


ChristopherDavenport:
I was interested recently in a large file contents and translating it from its base64 to the string representation.

I think it would be fairly straightforward to implement but wanted to see if there was an interest.

mpilquist:
I'd love to be able to reuse the methods on `ByteVector` for this -- perhaps with some small changes to scodec-bits to handle carry?

mpilquist:
@ChristopherDavenport How are these signatures for the `fs2.text` object?

```scala
object text {

  def base64Decode[F[_]]: Pipe[F, String, Byte] =
    _.through(base64DecodeC).flatMap(Stream.chunk)

  def base64DecodeC[F[_]]: Pipe[F, String, Chunk[Byte]] = ??? 

  def base64Encode[F[_]]: Pipe[F, Byte, String] =
    _.chunks.through(base64EncodeC)

  def base64EncodeC[F[_]]: Pipe[F, Chunk[Byte], String] = ???
}
```

djspiewak:
Is there a reason to have the separate chunk variants at all rather than just ensuring that the primary variants operate on chunks rather than individual bytes?

mpilquist:
@djspiewak I think we can skip the chunk variants. I looked up why we have the utf8 chunk variants and found this: https://github.com/functional-streams-for-scala/fs2/pull/609

djspiewak:
Ah interesting. It's worth noting that it would *probably* be more efficient, even in the `utf8` codecs, to exploit the chunkiness more explicitly. But, given that it's `String`s, I'm guessing it wouldn't make a massive difference. It would definitely be much more noticeable with `Chunk[Byte]`.

PR #1742, opened 2020-01-17
Added text.base64Encode and text.base64Decode

mpilquist:
Adapted the implementations from scodec-bits for streaming / continuation.

### DIFF
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -275,21 +275,15 @@ object text {
               case 0 =>
                 buffer = (cidx & 0x3f)
                 mod += 1
-              case 1 =>
-                buffer = (buffer << 6) | (cidx & 0x3f)
-                mod += 1
-              case 2 =>
+              case 1 | 2 =>
                 buffer = (buffer << 6) | (cidx & 0x3f)
                 mod += 1
               case 3 =>
                 buffer = (buffer << 6) | (cidx & 0x3f)
                 mod = 0
-                val c = buffer & 0x0ff
-                val b = (buffer >> 8) & 0x0ff
-                val a = (buffer >> 16) & 0x0ff
-                acc(bidx) = a.toByte
-                acc(bidx + 1) = b.toByte
-                acc(bidx + 2) = c.toByte
+                acc(bidx) = (buffer >> 16).toByte
+                acc(bidx + 1) = (buffer >> 8).toByte
+                acc(bidx + 2) = buffer.toByte
                 bidx += 3
             }
         }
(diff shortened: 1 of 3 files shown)


mpilquist:
Fixes #1587

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -201,4 +204,219 @@ object text {
...
+                    s"Unexpected character '$c' at index $idx after padding character; only '=' and whitespace characters allowed after first padding character"
+                  )
+                }
+              }
+            }
+            mod match {
+              case 0 =>
+                buffer = (cidx & 0x3f)
+                mod += 1
+              case 1 =>
```
matthughes: `case 1 | 2` might make intent a little clearer

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -201,4 +204,219 @@ object text {
...
+    }
+
+    def finish(state: State): Either[String, Chunk[Byte]] =
+      if (state.padding != 0 && state.mod != 0) paddingError
+      else
+        state.mod match {
+          case 0 => Right(Chunk.empty)
+          case 1 => Left("Final base 64 quantum had only 1 digit - must have at least 2 digits")
+          case 2 =>
+            Right(Chunk(((state.buffer >> 4) & 0x0ff).toByte))
```
matthughes: Is there a reason this is `0x0ff` vs `0xff`?

core/shared/src/test/scala/fs2/TextSpec.scala
```diff
@@ -230,5 +234,28 @@ class TextSpec extends Fs2Spec {
...
+          .through {
+            if (unchunked) _.unchunk
+            else _.rechunkRandomlyWithSeed(0.1, 2.0)(rechunkSeed)
+          }
+          .through(text.base64Decode[Fallible])
+          .compile
+          .to(ByteVector) shouldBe
+          Right(bs.map(ByteVector.view(_)).foldLeft(ByteVector.empty)(_ ++ _))
+      }
+    }
```
matthughes: Maybe repeat the encode/decode test but add random whitespace at beginning and end?

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -201,4 +204,219 @@ object text {
...
+              val out = new String(
+                Array(
+                  alphabet.toChar(first),
+                  alphabet.toChar(second),
+                  alphabet.toChar(third),
+                  alphabet.pad
+                )
+              )
+              Pull.output1(out)
+            case _ => sys.error("carry must be size 0, 1, or 2")
```
matthughes: Consider adding the invalid carry size in the error

core/shared/src/test/scala/fs2/TextSpec.scala
matthughes: Could also add test feeding the decode invalid base64 characters and assert it fails as expected (not throws exception).

matthughes: Or invalid lengths.

core/shared/src/main/scala/fs2/text.scala
mpilquist: No, old habit. Fixing.

mpilquist:
@matthughes Want to take a quick look at updates from review?

PR #1746, opened 2020-01-19
Fixed ClassCastException in Chunk#toArray, fixes #1745

### LINKED ISSUE #1745: fs2 upgrade 2.2.0 - ClassCastException
After upgrading `fs2:2.0.1` to `fs2:2.2.0` im getting this stacktrace, it looks there is some optimisation have been done with `Chunk` 

I tried with fs2:2.1.0 and everything work fines

```
[info] - should readFile *** FAILED ***
[info]   java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [B ([Ljava.lang.Object; and [B are in module java.base of loader 'bootstrap')
[info]   at fs2.Chunk.toBytes(Chunk.scala:276)
[info]   at fs2.io.file.FileHandle$$anon$1.$anonfun$write$1(FileHandle.scala:148)
[info]   at scala.runtime.java8.JFunction0$mcI$sp.apply(JFunction0$mcI$sp.scala:17)
[info]   at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:87)
[info]   at cats.effect.internals.IORunLoop$.startCancelable(IORunLoop.scala:41)
[info]   at cats.effect.internals.IOBracket$BracketStart.run(IOBracket.scala:86)
[info]   at cats.effect.internals.Trampoline.cats$effect$internals$Trampoline$$immediateLoop(Trampoline.scala:70)
[info]   at cats.effect.internals.Trampoline.startLoop(Trampoline.scala:36)
[info]   at cats.effect.internals.TrampolineEC$JVMTrampoline.super$startLoop(TrampolineEC.scala:93)
[info]   at cats.effect.internals.TrampolineEC$JVMTrampoline.$anonfun$startLoop$1(TrampolineEC.scala:93)
```

if you want to see the whole code it is here =>
https://github.com/regis-leray/fs2-ftp/pull/11

mpilquist:
@CremboC Could you review? Trick here is to avoid specializing `toArray` on `Boxed` -- it was making a bad assumption that the `ClassTag[O2]` passed to `toArray` was the same as used to create the underlying array.

### DIFF
--- a/core/shared/src/test/scala/fs2/ChunkSpec.scala
+++ b/core/shared/src/test/scala/fs2/ChunkSpec.scala
@@ -179,4 +179,26 @@ class ChunkSpec extends Fs2Spec {
     "longs" in testEmptyConcat(Chunk.concatLongs)
     "chars" in testEmptyConcat(Chunk.concatChars)
   }
+
+  "map andThen toArray" in {
+    val arr: Array[Int] = Chunk(0, 0).map(identity).toArray 
+    arr shouldBe Array(0, 0)
+  }
+
+  "mapAccumulate andThen toArray" in {
+    val arr: Array[Int] = Chunk(0, 0).mapAccumulate(0)((s, o) => (s, o))._2.toArray 
+    arr shouldBe Array(0, 0)
+  }
+
+  "scanLeft andThen toArray" in {
+    val arr: Array[Int] = Chunk(0, 0).scanLeft(0)((_, o) => o).toArray 
+    arr shouldBe Array(0, 0, 0)
+  }
+
+  "zip andThen toArray" in {
+    val arr: Array[(Int, Int)] = Chunk(0, 0).zip(Chunk(0, 0)).toArray
+    arr shouldBe Array((0, 0), (0, 0))
+    val arr2: Array[Int] = Chunk(0, 0).zip(Chunk(0, 0)).map(_._1).toArray
+    arr2 shouldBe Array(0, 0)
+  }
 }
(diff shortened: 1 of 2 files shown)


core/shared/src/test/scala/fs2/ChunkSpec.scala
```diff
@@ -179,4 +179,26 @@ class ChunkSpec extends Fs2Spec {
     "longs" in testEmptyConcat(Chunk.concatLongs)
     "chars" in testEmptyConcat(Chunk.concatChars)
   }
+
+  "map andThen toArray" in {
+    val arr: Array[Int] = Chunk(0, 0).map(identity).toArray
```
CremboC: I think it's worth adding test similar to the actual issue, i.e. `Chunk(0, 0).map(_.toByte).toArray`, since I think the issue is that it' inferred as `Int` initially, but then we're basically casting it into a byte, so the types change. I don't think identity will do anything like that in this case.

mpilquist: The `map(_.toByte)` was a red herring -- the issue reproduces with `map(identity)` as well. I added another test that more directly captures the issue:

```
// This one works fine b/c the underlying `Array[Any]` is returned
Chunk.Boxed(Array[Any](0)).asInstanceOf[Chunk[Int]].toArray[Any]
// This one throws b/c the underlying `Array[Any]` is returned and is stored in a reference to an `Array[Int]`, which causes the `ClassCastException`
Chunk.Boxed(Array[Any](0)).asInstanceOf[Chunk[Int]].toArray[Int]
```

PR #1750, opened 2020-01-19
Fixed bincompat issue with Pull.free

mpilquist:


mpilquist:
I don't think so -- mima didn't pick these up.

PR #1751, opened 2020-01-21
Merge pull request #1751 from r-glyde/file-permissions

r-glyde:
Part of #1533 allowing getting and setting file permissions

### DIFF
--- a/io/src/test/scala/fs2/io/file/FileSpec.scala
+++ b/io/src/test/scala/fs2/io/file/FileSpec.scala
@@ -9,7 +9,10 @@ import cats.effect.concurrent.Ref
 import cats.implicits._
 
 import scala.concurrent.duration._
-import java.nio.file.Paths
+import java.nio.file.{Files, Paths}
+import java.nio.file.attribute.PosixFilePermissions
+
+import CollectionCompat._
 
 class FileSpec extends BaseFileSpec {
   "readAll" - {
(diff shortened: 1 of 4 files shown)


io/src/test/scala/fs2/io/file/FileSpec.scala
```diff
@@ -127,6 +130,59 @@ class FileSpec extends BaseFileSpec {
...
+        .unsafeRunSync()
+        .isLeft shouldBe true
+    }
+    "should return permissions for existing file" in {
+      val permissions = PosixFilePermissions.fromString("rwxrwxr-x")
+      Blocker[IO]
+        .use { b =>
+          tempFile
+            .evalMap { p =>
+              IO.delay(Files.setPosixFilePermissions(p, permissions)) >> file.permissions[IO](b, p)
```
kubukoz: I think it's fine to use the fs2 method for setting permissions - that way we'll verify that the methods interact well with each other. I'd wait for someone else to state their opinion on this, though.

r-glyde: Yeah, I think that should be fine as well. It felt a bit odd when I first did it that way but will just be the same in the end and a bit tidier as well. I'll wait for someone to confirm before pushing a change.

mpilquist: Yeah I think it's fine to use fs2 method here

r-glyde: Updated to just use the fs2 methods in both tests

kubukoz:
Thanks!

PR #1752, opened 2020-01-21
Merge pull request #1752 from functional-streams-for-scala/no-211

kubukoz:


README.md
```diff
@@ -7,9 +7,14 @@ FS2: Functional Streams for Scala
 
 ### Overview
 
-FS2 is a library for purely functional, effectful, and polymorphic stream processing library in the [Scala programming language](https://scala-lang.org). Its design goals are compositionality, expressiveness, resource safety, and speed. The name is a modified acronym for **F**unctional **S**treams for **Scala** (FSS, or FS2).
+FS2 is a library for purely functional, effectful, and polymorphic stream processing library in the [Scala programming language](https://scala-lang.org).
+Its design goals are compositionality, expressiveness, resource safety, and speed.
```
diesalbla: _Note_: in Markdown, each newline in the source introduces a newline in the formatted text (unlike LaTeX), which sometimes makes the result not quite what one expects, for instance, poorly justified.

kubukoz: I don't think that's true on GitHub. I've used newlines before to have smaller diffs when changing just a word or two in the future...

If you use the rich preview for this file, you can see that it doesn't introduce newlines :) double `\n` would do it, for sure.

diesalbla: Oh, good to know that Github corrects this. Thanks.

Issue #734, opened 2016-09-13
s.to(sink) and s.tov(sink) should return Stream[F,Nothing]


mpilquist:
Currently, the `to` and `tov` functions return a `Stream[F,Unit]` but both drain the stream. We should return `Stream[F,Nothing]` here to make it clear that it's impossible to get a value out after running these.

Deferring fix to 0.10 since it breaks compatibility. We'll have to live with this for a while. :)

mpilquist:
See conversation in #752

PR #1754, opened 2020-01-21
Make writeAll return a Nothing-stream

kubukoz:
`writeAll` returns a stream of zero elements, so it might as well return `Nothing` to avoid confusion (someone mentioned to me recently that they were surprised they didn't get even one element from a stream piped through it).

I think it'll be binary compatible because of type erasure (plus, the stream was empty anyway, and the Unit type matched only thanks to covariance).

### DIFF
--- a/io/src/main/scala/fs2/io/file/file.scala
+++ b/io/src/main/scala/fs2/io/file/file.scala
@@ -68,12 +68,14 @@ package object file {
     * Writes all data to the file at the specified `java.nio.file.Path`.
     *
     * Adds the WRITE flag to any other `OpenOption` flags specified. By default, also adds the CREATE flag.
+    *
+    * The returned stream contains no elements.
     */
   def writeAll[F[_]: Sync: ContextShift](
       path: Path,
       blocker: Blocker,
       flags: Seq[StandardOpenOption] = List(StandardOpenOption.CREATE)
-  ): Pipe[F, Byte, Unit] =
+  ): Pipe[F, Byte, Nothing] =
     in =>
       Stream
         .resource(WriteCursor.fromPath(path, blocker, flags))
--- a/io/src/test/scala/fs2/io/file/BaseFileSpec.scala
+++ b/io/src/test/scala/fs2/io/file/BaseFileSpec.scala
@@ -36,7 +36,7 @@ class BaseFileSpec extends Fs2Spec {
   protected def modify(file: Path): Stream[IO, Unit] =
     Stream.eval(IO(Files.write(file, Array[Byte](0, 1, 2, 3))).void)
 
-  protected def modifyLater(file: Path, blocker: Blocker): Stream[IO, Unit] =
+  protected def modifyLater(file: Path, blocker: Blocker): Stream[IO, Nothing] =
     Stream
       .range(0, 4)
       .map(_.toByte)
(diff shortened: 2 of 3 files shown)


kubukoz:
Ongoing discussion on [gitter](https://gitter.im/functional-streams-for-scala/fs2-dev?at=5e27710d6b64bc54d12a6f27)

mpilquist:
I think we should either:
- stick with `Stream[F, Unit]` and provide a guarantee that a single `Unit` is emitted at completion of the stream
- change to `Stream[F, Nothing]`
Whatever we do, I think we should do it for all "sinks", not just `writeAll`.

My preference is changing to `Stream[F, Nothing]`, though that's not surprising given #734 / #752. Funny how we ended up removing `Sink` years later. Note, I'm linking those only for reference. I don't think we should consider this a settled issue and instead should discuss what we think is best now that we have years of experience of usage.

SystemFw:
Heh, the linked issue is pretty much identical to the conversation we had on Gitter :) 
Tbh I think it's one of those things where there isn't any decisive argument, I lean towards making it emit something, just because the association `Nothing => broken for` is obviously clear to us, but I've seen several people getting bitten by it

mpilquist:
One practical reason for preferring nothing is it cleans up expressions like `s.drain ++ t` to just `s ++ t`. This most frequently comes up when working with sockets & files.

I guess I'm not strongly motivated by for-comprehensions b/c they feel clunky with streams. They put the focus on `flatMap`, but streams tend to focus more on combinators and `++`.

CremboC:
I think we should be consistent - if we got a `Stream[F, Unit]` then it should emit, if we got a `Stream[F, Nothing]` then it shouldn't emit. At least then we have a pattern, although not obvious, we could put a caveat somewhere in docs about this.

SystemFw:
>  if we got a Stream[F, Unit] then it should emit, if we got a Stream[F, Nothing] then it shouldn't emit

I think you mean "if it doesn't emit, it should be a Stream[F, Nothing]" (the reverse is necessarily true unless you `asInstanceOf` . I agree with the consistency, do you have any opinion on which one of the two is preferable for this case?

@mpilquist I think both your arguments are solid btw :)

CremboC:
@SystemFw just to clarify, currently the following is true:
- Signature `Stream[F, Unit]` can mean anything - 0 or more elements emitted which are all `()`
- Signature `Stream[F, Nothing]` guarantees 0 elements emitted (unless, as you said, you use `asInstanceOf`, but ignore that!)

If that's true, since this method emits 0 elements, it should be `Stream[F, Nothing]` as well. Also I think we should then update other methods to follow the same pattern - if 0 elements, then return type should be `Stream[F, Nothing]`.

SystemFw:
> If that's true, since this method emits 0 elements, it should be Stream[F, Nothing] as well.


yeah, but the issue is: should we change it to emit 1 Unit instead (and keep the current type), or keep the behaviour as is (and change the type of be `Stream[F, Nothing]`)?

SystemFw:
Anyway, if the consensus is to change it to `Stream[F, Nothing]`, I won't stand in the way, since it's indeed very reasonable.

I do know however of cases where teams I work with spent ages chasing what ended up being an empty stream, and the implication `Stream[F, Nothing] ==> dead code after flatMap`, although very obvious if you already know, it's not immediate to people (especially because they might overlook the return type altogether, since it will always compile), so we should probably add that somewhere in the docs

kubukoz:
Superseded by #1913

PR #1755, opened 2020-01-21
Merge pull request #1755 from functional-streams-for-scala/remove-sbt-script

kubukoz:
It seems to be unused, at least...


kubukoz:
@lukajcb

image

kubukoz:
Looks like refined-github decided to change the title for me... I need to disable that.

Issue #1679, opened 2019-11-05
Scope lookup failure when using concurrently/merge


mpilquist:
Brought up by @augustjune on Gitter today:

```scala
@ def holdLater[F[_]: Concurrent, A](stream: Stream[F, A]): Stream[F, Signal[F, A]] = {
    def uncons1(stream: Stream[F, A]): Stream[F, (A, Stream[F, A])] =
      stream.pull.uncons1.flatMap {
        case Some((a, rest)) => Pull.output1(a -> rest)
        case None            => Pull.done
      }.stream

    uncons1(stream).flatMap { case (h, t) =>
      Stream.eval(SignallingRef[F, A](h)).flatMap { sig =>
        Stream(sig).concurrently(t.evalMap(sig.set))
      }
    }
  }
defined function holdLater

@ holdLater(Stream(1, 2, 3).covary[IO].metered(1.second)).flatMap(_.discrete).take(3).compile.toList.unsafeToFuture
res19: scala.concurrent.Future[List[Int]] = Future(<not completed>)

@ res19
res20: scala.concurrent.Future[List[Int]] = Future(Failure(java.lang.Throwable: Fail to find scope for next step: current: Token(7e32dc50), step: Step(FreeC.Bind(FreeC.Bind(Step(FreeC.Bind(Eval(IO$1895120358)),None))),Some(Token(74cab720)))))
```

Replacing `concurrently` with `merge` or `parJoinUnbounded` results in the same error.

PR #1756, opened 2020-01-22
Add peek method to Stream

bplommer:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2265,6 +2265,26 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
...
+    * that emits the same elements as this stream, with the peeked element memoized.
+    *
+    * @example {{{
+    * scala> val outer =  Stream(1, 2, 3).peek.toList.head
+    * outer: (Int, Stream[Pure, Int]) = (1, Stream(..))
+    * scala> outer._2.toList
+    * res1: List[Int] = List(1, 2, 3)
+    * }}}
+    */
+  def peek: Stream[F, (O, Stream[F, O])] =
```
diesalbla: Perhaps `uncons1` would be a better name for this. That name indicates that you are deconstructing the head and tail of the list (contrary to what a `cons` does), and is similar to what a similar method for `Pull` does.

kubukoz: From `peek1` on ToPull:

`/** Like [[uncons1]] but does not consume the element (i.e., the element is pushed back). */`

It makes sense to call it `uncons1`, IMO. With the `1` because we don't emit the whole chunk.

kubukoz:
I like this, would like to see some tests and other maintainers' opinion :)

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: In this case, `peek1` seems like the right name. One thing I'm wary about is scoping bugs -- e.g., take a look at #1679 and look at similarity of the `uncons1` function.

SystemFw: I was thinking the same, I think it would be nice to define clearer rules around this: i.e. do we permit to return the tail of the Stream from a `Pull` (and #1679 is a bug) or do we not?

mpilquist:

PR #1757, opened 2020-01-22
A Stream can pull anything.

diesalbla:
A `Stream[F, O]` is defined as a wrap on a `FreeC[F, O, Unit]`. The Unit type indicates that the "Result" of the `FreeC` computation that produces the stream has no relevant information: all that matters of the Stream is the output. 

We can also express that idea by marking that return type as an `Any`, the top type, which is another way to indicate that no useful information can come off the Result of the inner `FreeC`. One benefit is that we do not need the spurious conversions of the `FreeC` with `void`, just to wrap it back into a Stream.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -38,13 +38,10 @@ final class Pull[+F[_], +O, +R] private[fs2] (private val free: FreeC[F, O, R]) 
   /**
     * Interpret this `Pull` to produce a `Stream`.
     *
-    * May only be called on pulls which return a `Unit` result type. Use `p.void.stream` to explicitly
+    * May only be called on pulls which return a `Unit` result type. Use `p.stream` to explicitly
     * ignore the result type of the pull.
     */
-  def stream(implicit ev: R <:< Unit): Stream[F, O] = {
-    val _ = ev
-    new Stream(free.asInstanceOf[FreeC[F, O, Unit]])
-  }
+  def stream: Stream[F, O] = new Stream(free)
 
   /** Applies the resource of this pull to `f` and returns the result. */
   def flatMap[F2[x] >: F[x], O2 >: O, R2](f: R => Pull[F2, O2, R2]): Pull[F2, O2, R2] =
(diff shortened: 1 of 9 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -38,13 +38,10 @@ final class Pull[+F[_], +O, +R] private[fs2] (private val free: FreeC[F, O, R])
...
     *
-    * May only be called on pulls which return a `Unit` result type. Use `p.void.stream` to explicitly
+    * May only be called on pulls which return a `Unit` result type. Use `p.stream` to explicitly
     * ignore the result type of the pull.
     */
-  def stream(implicit ev: R <:< Unit): Stream[F, O] = {
-    val _ = ev
-    new Stream(free.asInstanceOf[FreeC[F, O, Unit]])
-  }
+  def stream: Stream[F, O] = new Stream(free)
```
mpilquist: I'm generally :-1: on removing the `R <:< Unit` requirement here as requiring that evidence has exposed bugs where the remainder was not dealt with correctly.

core/shared/src/main/scala/fs2/fs2.scala
```diff
@@ -34,4 +34,10 @@ package object fs2 {
     * Alias for `Nothing` which works better with type inference.
     */
   type INothing <: Nothing
+
+  /**
+    * Alias for `Any` which may work better with type inference.
+    */
+  type IAny >: Any
```
mpilquist: If we were to go this way, I don't think we need `IAny` as scalac doesn't treat `Any` special from type inference perspective.

mpilquist:
Not sure about this one. Would like to hear from other contributors. API-wise, I'm concerned about the ability to convert any pull to a stream, which seems like a regression considering we [recently added the `R <:<: Unit` constraint](006ea6fb77d62dd5ce321eaf8d30f499e117ceb2). It's like the pull version of `-Ywarn-value-discard`.

SystemFw:
This is a torny issue in general (consider `*>` on one hand, and `-Ywarn-value discard` on the other).
Overall, I think I'm in favour of keeping `R <:< Unit`, because it really drives home the difference between `O` and `R` in `Pull`.

Also, I would only use `Any` when nothing else will do (e.g. in the implementation of `Resource` and `IO`) to really mean _anything_,  because according to this argument, all uses of `Unit` could be replaced by `Any` (which in a way is the definition of value discarding, which I don't think is a good idea)

diesalbla:
>  I'm generally 👎 on removing the R <:< Unit requirement here as requiring that evidence has exposed bugs where the remainder was not dealt with correctly.

All right. That makes this PR no longer useful.

PR #1763, opened 2020-01-26
Add temporary files and directories to io module

r-glyde:
Part of #1533. 

~Just a draft PR for now.~ I wasn't completely sure if this was the sort of intended implementation or if should just be a wrapper around the `nio` functions and leave removing files/directories to the user (or something else). Happy to close if this is quite far off from what was intended.

### DIFF
--- a/io/src/main/scala/fs2/io/file/file.scala
+++ b/io/src/main/scala/fs2/io/file/file.scala
@@ -181,6 +181,8 @@ package object file {
 
   /**
     * Get file permissions as set of [[PosixFilePermission]]
+    *
+    * This will only work for POSIX supporting file systems i.e. not Windows
     */
   def permissions[F[_]: Sync: ContextShift](
       blocker: Blocker,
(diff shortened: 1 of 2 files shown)


io/src/main/scala/fs2/io/file/file.scala
```diff
@@ -227,6 +231,32 @@ package object file {
...
+            Files.deleteIfExists(path)
+            FileVisitResult.CONTINUE
+          }
+          override def postVisitDirectory(path: Path, e: IOException): FileVisitResult = {
+            Files.deleteIfExists(path)
+            FileVisitResult.CONTINUE
+          }
+        }
+      )
+    }
```
r-glyde: I think this could also be done with something like:
```scala
walk(blocker, path, options)
  .compile
  .toVector
  .flatMap(_.reverse.traverse_(deleteIfExists[F](blocker, _)))
```
but thought it might have issues with very largely nested directory structures?

PR #1774, opened 2020-01-31
Replace ShouldMatchers in tests with assertions

### LINKED ISSUE #1764: Style change: replace ShouldMatchers in tests with assertions
Example rewrites:

|Before|After|
|----|----|
|`x shouldBe y`|`assert(x == y)`|
|`x should be < y`|`assert(x < y)`|
|`an[Exception] shouldBe thrown by f`|`assertThrows[Exception](f)`|

I recently made this same change to scodec-bits (in topic/dotty branch for now) and found the result much simpler to read. Note that `assert` is macro-powered and if the assertion fails, you still get details on the cause of the failure:

```scala
scala> val xs = List(1, 2, 3)
xs: List[Int] = List(1, 2, 3)

scala> assert(xs.reverse == List(3, 2, 1, 0))
org.scalatest.exceptions.TestFailedException: List(3, 2, 1) did not equal List(3, 2, 1, 0)
```

mgibowski:
Fixes #1764

This PR replaces all `should` matchers in all tests at one go.
It may be a lot of changes to review... If necessary, I could break this PR into few smaller ones.

### DIFF
--- a/core/jvm/src/test/scala/fs2/CompressSpec.scala
+++ b/core/jvm/src/test/scala/fs2/CompressSpec.scala
@@ -89,7 +89,7 @@ class CompressSpec extends Fs2Spec {
         .through(compress.inflate())
         .compile
         .toVector
-        .asserting(_ == s.toVector)
+        .assert(_ == s.toVector)
     }
 
     "deflate.compresses input" in {
@@ -111,7 +111,7 @@ class CompressSpec extends Fs2Spec {
         .through(compress.gunzip[IO](8192))
         .compile
         .toVector
-        .asserting(_ == s.toVector)
+        .assert(_ == s.toVector)
     }
 
     "gzip |> gunzip ~= id (mutually prime chunk sizes, compression larger)" in forAll {
@@ -121,7 +121,7 @@ class CompressSpec extends Fs2Spec {
           .through(compress.gunzip[IO](509))
           .compile
           .toVector
-          .asserting(_ == s.toVector)
+          .assert(_ == s.toVector)
     }
 
     "gzip |> gunzip ~= id (mutually prime chunk sizes, decompression larger)" in forAll {
@@ -131,7 +131,7 @@ class CompressSpec extends Fs2Spec {
           .through(compress.gunzip[IO](1031))
           .compile
           .toVector
-          .asserting(_ == s.toVector)
+          .assert(_ == s.toVector)
     }
 
     "gzip |> GZIPInputStream ~= id" in forAll { s: Stream[Pure, Byte] =>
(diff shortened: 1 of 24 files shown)


gvolpe:
How about using Cats equality instead of universal equality?

mpilquist:
I don't think that will play nicely with the ScalaTest `assert` macro -- we'll just get true/false I think.

gvolpe:
Not sure I follow. `Eq[A].eqv(a1, a2)` returns a `Boolean`, so we could have sth like

```scala
assert(a1.eqv(a2))
```

The only issue I fiu d was when trying to use syntax `===`, which clashes with the Scalactic one brought by Scalatest, depending on the interface being mixed in.

mpilquist:
:sparkles: macros :sparkles:

Take a look at the exception messages:

```scala
@ import $ivy.`org.scalatest::scalatest:3.1.0`, org.scalatest.Assertions._
import $ivy.$                               , org.scalatest.Assertions._

@ import $ivy.`org.typelevel::cats-effect:1.0.0`, cats._, cats.implicits._, cats.effect._, cats.data._
import $ivy.$                                 , cats._, cats.implicits._, cats.effect._, cats.data._

@ val x = 1
x: Int = 1

@ assert(x == 2)
org.scalatest.exceptions.TestFailedException: 1 did not equal 2
  org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:472)
  org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:471)
  org.scalatest.Assertions$.newAssertionFailedException(Assertions.scala:1231)
  org.scalatest.Assertions$AssertionsHelper.macroAssert(Assertions.scala:1295)
  ammonite.$sess.cmd6$.<init>(cmd6.sc:1)
  ammonite.$sess.cmd6$.<clinit>(cmd6.sc)

@ assert(Eq[Int].eqv(x, 2))
org.scalatest.exceptions.TestFailedException: cats.`package`.Eq.apply[Int](cats.implicits.catsKernelStdOrderForInt).eqv(ammonite.$sess.cmd5.x, 2) was false
  org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:472)
  org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:471)
  org.scalatest.Assertions$.newAssertionFailedException(Assertions.scala:1231)
  org.scalatest.Assertions$AssertionsHelper.macroAssert(Assertions.scala:1295)
  ammonite.$sess.cmd7$.<init>(cmd7.sc:1)
  ammonite.$sess.cmd7$.<clinit>(cmd7.sc)
```

gvolpe:
Oh that's what you meant, damn. I wonder if there's any better way to reconcile assertions with Cats equality? 🤔

Probably work for another PR.

PR #1781, opened 2020-02-03
Prevent concurrent writes to fs2-io TCP socket from causing WritePendingException 

bplommer:
I had to tune the test parameters (message size and write buffer size) to make the test consistently fail on my machine without this change - hopefully it does so on other machines too 🙂

### DIFF
--- a/io/src/main/scala/fs2/io/tcp/SocketGroup.scala
+++ b/io/src/main/scala/fs2/io/tcp/SocketGroup.scala
@@ -212,7 +212,7 @@ final class SocketGroup(channelGroup: AsynchronousChannelGroup, blocker: Blocker
   private def apply[F[_]](
       ch: AsynchronousSocketChannel
   )(implicit F: Concurrent[F], cs: ContextShift[F]): Resource[F, Socket[F]] = {
-    val socket = (Semaphore[F](1), Semaphore[F](1)).flatMap { semaphores =>
+    val socket = (Semaphore[F](1), Semaphore[F](1)).tupled.flatMap { semaphores =>
       val (readSemaphore, writeSemaphore) = semaphores
       Ref.of[F, ByteBuffer](ByteBuffer.allocate(0)).map { bufferRef =>
         // Reads data to remaining capacity of supplied ByteBuffer
--- a/io/src/main/scala/fs2/io/tcp/SocketGroup.scala
+++ b/io/src/main/scala/fs2/io/tcp/SocketGroup.scala
@@ -212,9 +212,8 @@ final class SocketGroup(channelGroup: AsynchronousChannelGroup, blocker: Blocker
   private def apply[F[_]](
       ch: AsynchronousSocketChannel
   )(implicit F: Concurrent[F], cs: ContextShift[F]): Resource[F, Socket[F]] = {
-    val socket = (Semaphore[F](1), Semaphore[F](1)).tupled.flatMap { semaphores =>
-      val (readSemaphore, writeSemaphore) = semaphores
-      Ref.of[F, ByteBuffer](ByteBuffer.allocate(0)).map { bufferRef =>
+    val socket = (Semaphore[F](1), Semaphore[F](1), Ref[F].of(ByteBuffer.allocate(0))).mapN {
+      (readSemaphore, writeSemaphore, bufferRef) =>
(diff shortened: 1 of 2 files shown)


io/src/main/scala/fs2/io/tcp/SocketGroup.scala
```diff
@@ -212,7 +212,8 @@ final class SocketGroup(channelGroup: AsynchronousChannelGroup, blocker: Blocker
   private def apply[F[_]](
       ch: AsynchronousSocketChannel
   )(implicit F: Concurrent[F], cs: ContextShift[F]): Resource[F, Socket[F]] = {
-    val socket = Semaphore[F](1).flatMap { readSemaphore =>
+    val socket = (Semaphore[F](1), Semaphore[F](1)).tupled.flatMap { semaphores =>
```
kubukoz: Personally I'd use a pattern match like `flatMap { case (readSemaphore, writeSemaphore) =>`.

kubukoz:
Thanks! I'll see how it behaves on my device.

io/src/test/scala/fs2/io/tcp/SocketSpec.scala
```diff
@@ -127,5 +128,43 @@ class SocketSpec extends Fs2Spec {
           .get
       result shouldBe sizes
     }
+
+    "write - concurrent calls do not cause WritePendingException" in {
```
kubukoz: Could we simulate the conditions leading to the exception by connecting to a (fs2-io tcp-based) server that pulls bytes from the socket reaaally slow?

io/src/main/scala/fs2/io/tcp/SocketGroup.scala
bplommer: I did that at first - the only reason I changed it to this was because the extra indentation blew up the diff, but happy to change it

kubukoz: Fair enough :) let's leave it be

io/src/test/scala/fs2/io/tcp/SocketSpec.scala
```diff
@@ -127,5 +128,43 @@ class SocketSpec extends Fs2Spec {
...
+      val message = Chunk.bytes(("123456789012345678901234567890" * 10000).getBytes)
+
+      val localBindAddress =
+        Deferred[IO, InetSocketAddress].unsafeRunSync()
```
kubukoz: I think we can just flatMap on it (passing the Deferred as a parameter to `doNothingServer` will also be nice for the nesting problem)

io/src/test/scala/fs2/io/tcp/SocketSpec.scala
kubukoz: really slow, as in "waits for a promise that's fulfilled some time around when the second write attempt is started" :)

bplommer: At the moment we're connecting to a server that doesn't pull bytes at all, which I guess is the limiting case of pulling them really slow. I've pushed a change to get the exception without changing the write buffer size.

bplommer: I’m not quite sure I’m following. What’s the issue with the test as it stands?

kubukoz: I suppose I don't like that we rely on an arbitrary amount for the size of the buffer (I think the server one can be configured, maybe the client one can too) and the amount of concurrent requests to write. But I won't stand in the way if other maintainers are okay with it :)

io/src/test/scala/fs2/io/tcp/SocketSpec.scala
```diff
@@ -127,5 +128,44 @@ class SocketSpec extends Fs2Spec {
...
+                      Stream.eval(sock.write(message)).repeatN(10L)
+                    ).repeatN(2L)
+                  )
+                  .parJoinUnbounded
+              }
+          }
+          .compile
+          .drain
+          .attempt
+          .unsafeRunTimed(timeout)
```
kubukoz: I'd suggest using `timeout,` instead. `unsafezrunTimed` limits the time between async boundaries, not the whole IO.

PR #1783, opened 2020-02-05
Enhancements to fs2.compress

greg-a-atkinson:
Enhance fs2.compress with RFC 1952 gzip/gunzip implementation and remove ArrayBuffer from deflate/inflate implemtation.

Gzip/gunzip enhancements support header properties, header CRC validation, content CRC validation and content length validation. Deflate/inflate enhancements are mainly for the purpose of replacing the previous ArrayBuffer-based implementation in favour of a chunk-based implementation to avoid excessive array concatenation and copying during ArrayBuffer resizing highlighted by application profiling.

Existing public interfaces are unchanged. Although, it would have been preferable to specify the type of F to allow the Inflater and Deflater resources to be bracketed rather than closed in a trailing empty stream - a solution to this would be appreciated.

Additional tests have been added, including tests that verify the updated implementation against the previous implementation and tests that can be used to show performance is not degraded.

Feedback for improvement to the code will be appreciated.

Thanks

### DIFF
--- a/core/jvm/src/main/scala/fs2/compress.scala
+++ b/core/jvm/src/main/scala/fs2/compress.scala
@@ -9,7 +9,7 @@ import scala.annotation.tailrec
 import scala.collection.mutable.ArrayBuffer
 
 /** Provides utilities for compressing/decompressing byte streams. */
-@deprecated("Use fs2.compress2 instead", "2.3.0")
+@deprecated("Use fs2.compression instead", "2.3.0")
 object compress {
 
   /**
--- a/core/jvm/src/main/scala/fs2/compress.scala
+++ b/core/jvm/src/main/scala/fs2/compress.scala
@@ -9,6 +9,7 @@ import scala.annotation.tailrec
 import scala.collection.mutable.ArrayBuffer
 
 /** Provides utilities for compressing/decompressing byte streams. */
+@deprecated("Use fs2.compress2 instead", "2.3.0")
 object compress {
 
   /**
--- a/core/jvm/src/main/scala/fs2/compress2.scala
+++ b/core/jvm/src/main/scala/fs2/compress2.scala
@@ -1,0 +1,863 @@
+package fs2
+
+import java.io.EOFException
+import java.nio.charset.StandardCharsets
+import java.time.Instant
+import java.util.zip._
+
+import cats.effect.Sync
+
+/** Provides utilities for compressing/decompressing byte streams. */
+object compress2 {
+
+  /**
+    * Returns a `Pipe` that deflates (compresses) its input elements using
+    * a `java.util.zip.Deflater` with the parameters `level`, `nowrap` and `strategy`.
+    * @param level the compression level (0-9)
+    * @param nowrap if true then use GZIP compatible compression
+    * @param bufferSize size of the internal buffer that is used by the
+    *                   compressor. Default size is 32 KB.
+    * @param strategy compression strategy -- see `java.util.zip.Deflater` for details
+    */
+  def deflate[F[_]](
+      level: Int = Deflater.DEFAULT_COMPRESSION,
+      nowrap: Boolean = false,
+      bufferSize: Int = 1024 * 32,
+      strategy: Int = Deflater.DEFAULT_STRATEGY
+  )(implicit SyncF: Sync[F]): Pipe[F, Byte, Byte] =
+    stream =>
+      Stream
+        .bracket(
+          SyncF.delay {
+            val deflater = new Deflater(level, nowrap)
+            deflater.setStrategy(strategy)
+            deflater
+          }
+        )(deflater => SyncF.delay(deflater.end()))
+        .flatMap(deflater => _deflate(deflater, bufferSize, crc32 = None)(stream))
+
+  private def _deflate[F[_]](
+      deflater: Deflater,
+      bufferSize: Int,
+      crc32: Option[CRC32]
+  ): Pipe[F, Byte, Byte] =
+    _deflate_stream(deflater, bufferSize.max(minBufferSize), crc32)(_).stream
+
+  private def _deflate_chunk[F[_]](
+      chunk: Chunk[Byte],
+      deflater: Deflater,
+      deflatedBufferSize: Int,
+      isFinalChunk: Boolean,
+      crc32: Option[CRC32]
+  ): Pull[F, Byte, Unit] = {
+    val bytesChunk = chunk.toBytes
+    deflater.setInput(
+      bytesChunk.values,
+      bytesChunk.offset,
+      bytesChunk.length
+    )
+    if (isFinalChunk)
+      deflater.finish()
+    crc32.foreach(_.update(bytesChunk.values, bytesChunk.offset, bytesChunk.length))
+
+    def isDone: Boolean =
+      (isFinalChunk && deflater.finished) || (!isFinalChunk && deflater.needsInput)
+
+    def deflateInto(deflatedBuffer: Array[Byte]): Int =
+      if (isDone) 0 else deflater.deflate(deflatedBuffer)
+
+    def pull(): Pull[F, Byte, Unit] = {
+      val deflatedBuffer = new Array[Byte](deflatedBufferSize)
+      val deflatedBytes = deflateInto(deflatedBuffer)
+      if (isDone) {
+        Pull.output(asChunkBytes(deflatedBuffer, deflatedBytes))
+      } else {
+        Pull.output(asChunkBytes(deflatedBuffer, deflatedBytes)) >> pull()
+      }
+    }
+
+    pull()
+  }
+
+  private def _deflate_stream[F[_]](
+      deflater: Deflater,
+      bufferSize: Int,
+      crc32: Option[CRC32]
+  ): Stream[F, Byte] => Pull[F, Byte, Unit] =
+    _.pull.unconsNonEmpty.flatMap {
+      case Some((inflatedChunk, inflatedStream)) =>
+        _deflate_chunk(inflatedChunk, deflater, bufferSize, isFinalChunk = false, crc32) >> _deflate_stream(
+          deflater,
+          bufferSize,
+          crc32
+        )(inflatedStream)
+      case None =>
+        _deflate_chunk(Chunk.empty[Byte], deflater, bufferSize, isFinalChunk = true, crc32)
+    }
+
+  /**
+    * Returns a `Pipe` that inflates (decompresses) its input elements using
+    * a `java.util.zip.Inflater` with the parameter `nowrap`.
+    * @param nowrap if true then support GZIP compatible decompression
+    * @param bufferSize size of the internal buffer that is used by the
+    *                   decompressor. Default size is 32 KB.
+    */
+  def inflate[F[_]](nowrap: Boolean = false, bufferSize: Int = 1024 * 32)(
+      implicit SyncF: Sync[F]
+  ): Pipe[F, Byte, Byte] =
+    stream =>
+      Stream
+        .bracket(SyncF.delay(new Inflater(nowrap)))(inflater => SyncF.delay(inflater.end()))
+        .flatMap(inflater => _inflate(bufferSize, inflater, crc32 = None)(SyncF)(stream))
+
+  private def _inflate[F[_]](bufferSize: Int, inflater: Inflater, crc32: Option[CRC32])(
+      implicit SyncF: Sync[F]
+  ): Pipe[F, Byte, Byte] =
+    _.pull.unconsNonEmpty.flatMap {
+      case Some((deflatedChunk, deflatedStream)) =>
+        _inflate_chunk(deflatedChunk, inflater, bufferSize.max(minBufferSize), crc32) >> _inflate_stream(
+          inflater,
+          bufferSize.max(minBufferSize),
+          crc32
+        )(SyncF)(deflatedStream)
+      case None =>
+        Pull.done
+    }.stream
+
+  private def _inflate_chunk[F[_]](
+      chunk: Chunk[Byte],
+      inflater: Inflater,
+      inflatedBufferSize: Int,
+      crc32: Option[CRC32]
+  ): Pull[F, Byte, Unit] = {
+    val bytesChunk = chunk.toBytes
+    inflater.setInput(
+      bytesChunk.values,
+      bytesChunk.offset,
+      bytesChunk.length
+    )
+    def inflateInto(inflatedBuffer: Array[Byte]): Int =
+      if (inflater.finished()) -2
+      else if (inflater.needsInput()) -1
+      else {
+        val byteCount = inflater.inflate(inflatedBuffer)
+        crc32.foreach(_.update(inflatedBuffer, 0, byteCount))
+        byteCount
+      }
+
+    def pull(): Pull[F, Byte, Unit] = {
+      val inflatedBuffer = new Array[Byte](inflatedBufferSize)
+
+      inflateInto(inflatedBuffer) match {
+        case inflatedBytes if inflatedBytes <= -2 =>
+          Pull.output(chunk)
+        case inflatedBytes if inflatedBytes == -1 =>
+          Pull.done
+        case inflatedBytes if inflatedBytes < inflatedBuffer.length =>
+          if (inflater.finished()) {
+            inflater.getRemaining match {
+              case bytesRemaining if bytesRemaining > 0 =>
+                Pull.output(asChunkBytes(inflatedBuffer, inflatedBytes)) >>
+                  Pull.output(
+                    Chunk.Bytes(
+                      bytesChunk.values,
+                      bytesChunk.offset + bytesChunk.length - bytesRemaining,
+                      bytesRemaining
+                    )
+                  )
+              case _ =>
+                Pull.output(asChunkBytes(inflatedBuffer, inflatedBytes))
+            }
+          } else Pull.output(asChunkBytes(inflatedBuffer, inflatedBytes))
+        case inflatedBytes =>
+          Pull.output(asChunkBytes(inflatedBuffer, inflatedBytes)) >> pull()
+      }
+    }
+
+    pull()
+  }
+
+  private def _inflate_stream[F[_]](
+      inflater: Inflater,
+      inflatedBufferSize: Int,
+      crc32: Option[CRC32]
+  )(implicit SyncF: Sync[F]): Stream[F, Byte] => Pull[F, Byte, Unit] =
+    _.pull.unconsNonEmpty.flatMap {
+      case Some((deflatedChunk, deflatedStream)) =>
+        _inflate_chunk(deflatedChunk, inflater, inflatedBufferSize, crc32) >> _inflate_stream(
+          inflater,
+          inflatedBufferSize,
+          crc32
+        )(SyncF)(deflatedStream)
+      case None =>
+        if (!inflater.finished)
+          Pull.raiseError[F](new DataFormatException("Insufficient data"))
+        else {
+          Pull.done
+        }
+    }
+
+  /**
+    * Returns a pipe that incrementally compresses input into the GZIP format
+    * as defined by RFC 1952 at https://www.ietf.org/rfc/rfc1952.txt. Output is
+    * compatible with the GNU utils `gunzip` utility, as well as really anything
+    * else that understands GZIP. Note, however, that the GZIP format is not
+    * "stable" in the sense that all compressors will produce identical output
+    * given identical input. Part of the header seeding is arbitrary and chosen by
+    * the compression implementation. For this reason, the exact bytes produced
+    * by this pipe will differ in insignificant ways from the exact bytes produced
+    * by a tool like the GNU utils `gzip`.
+    *
+    * @param bufferSize The buffer size which will be used to page data
+    *                   into chunks. This will be the chunk size of the
+    *                   output stream. You should set it to be equal to
+    *                   the size of the largest chunk in the input stream.
+    *                   Setting this to a size which is ''smaller'' than
+    *                   the chunks in the input stream will result in
+    *                   performance degradation of roughly 50-75%. Default
+    *                   size is 32 KB.
+    * @param deflateLevel     level the compression level (0-9)
+    * @param deflateStrategy  strategy compression strategy -- see `java.util.zip.Deflater` for details
+    * @param modificationTime optional file modification time
+    * @param fileName         optional file name
+    * @param comment          optional file comment
+    */
+  def gzip[F[_]](
+      bufferSize: Int = 1024 * 32,
+      deflateLevel: Option[Int] = None,
+      deflateStrategy: Option[Int] = None,
+      modificationTime: Option[Instant] = None,
+      fileName: Option[String] = None,
+      comment: Option[String] = None
+  )(implicit SyncF: Sync[F]): Pipe[F, Byte, Byte] =
+    stream =>
+      Stream
+        .bracket(
+          SyncF.delay {
+            val deflater = new Deflater(deflateLevel.getOrElse(Deflater.DEFAULT_COMPRESSION), true)
+            deflater.setStrategy(deflateStrategy.getOrElse(Deflater.DEFAULT_STRATEGY))
+            (deflater, new CRC32())
+          }
+        ) { case (deflater, _) => SyncF.delay(deflater.end()) }
+        .flatMap {
+          case (deflater, crc32) =>
+            _gzip_header(modificationTime, deflateLevel, fileName, comment) ++
+              _deflate(deflater, bufferSize, Some(crc32))(stream) ++
+              _gzip_trailer(deflater, crc32)
+        }
+
+  private def _gzip_header[F[_]](
+      modificationTime: Option[Instant],
+      deflateLevel: Option[Int],
+      fileName: Option[String],
+      comment: Option[String]
+  ): Stream[F, Byte] = {
+    // See RFC 1952: https://www.ietf.org/rfc/rfc1952.txt
+    val secondsSince197001010000: Long =
+      modificationTime.map(_.getEpochSecond).getOrElse(0)
+    val header = Array[Byte](
+      gzipMagicFirstByte, // ID1: Identification 1
+      gzipMagicSecondByte, // ID2: Identification 2
+      gzipCompressionMethod.DEFLATE, // CM: Compression Method
+      (gzipFlag.FHCRC + // FLG: Header CRC
+        fileName.map(_ => gzipFlag.FNAME).getOrElse(zeroByte) + // FLG: File name
+        comment.map(_ => gzipFlag.FCOMMENT).getOrElse(zeroByte)).toByte, // FLG: Comment
+      (secondsSince197001010000 & 0xff).toByte, // MTIME: Modification Time
+      ((secondsSince197001010000 >> 8) & 0xff).toByte,
+      ((secondsSince197001010000 >> 16) & 0xff).toByte,
+      ((secondsSince197001010000 >> 24) & 0xff).toByte,
+      deflateLevel match { // XFL: Extra flags
+        case Some(Deflater.BEST_COMPRESSION) => gzipExtraFlag.DEFLATE_MAX_COMPRESSION_SLOWEST_ALGO
+        case Some(Deflater.BEST_SPEED)       => gzipExtraFlag.DEFLATE_FASTEST_ALGO
+        case _                               => zeroByte
+      },
+      gzipOperatingSystem.THIS
+    ) // OS: Operating System
+    val crc32 = new CRC32()
+    crc32.update(header)
+    val fileNameEncoded = fileName.map { string =>
+      val bytes = string.replaceAll("\u0000", "_").getBytes(StandardCharsets.ISO_8859_1)
+      crc32.update(bytes)
+      crc32.update(zeroByte)
+      bytes
+    }
+    val commentEncoded = comment.map { string =>
+      val bytes = string.replaceAll("\u0000", " ").getBytes(StandardCharsets.ISO_8859_1)
+      crc32.update(bytes)
+      crc32.update(zeroByte)
+      bytes
+    }
+    val crc32Value = crc32.getValue
+    val crc16 = Array[Byte](
+      (crc32Value & 0xff).toByte,
+      ((crc32Value >> 8) & 0xff).toByte
+    )
+    Stream.chunk(asChunkBytes(header)) ++
+      fileNameEncoded
+        .map(bytes => Stream.chunk(asChunkBytes(bytes)) ++ Stream.emit(zeroByte))
+        .getOrElse(Stream.empty) ++
+      commentEncoded
+        .map(bytes => Stream.chunk(asChunkBytes(bytes)) ++ Stream.emit(zeroByte))
+        .getOrElse(Stream.empty) ++
+      Stream.chunk(asChunkBytes(crc16))
+  }
+
+  private def _gzip_trailer[F[_]](deflater: Deflater, crc32: CRC32): Stream[F, Byte] = {
+    // See RFC 1952: https://www.ietf.org/rfc/rfc1952.txt
+    val crc32Value = crc32.getValue
+    val bytesIn = deflater.getTotalIn
+    val trailer = Array[Byte](
+      (crc32Value & 0xff).toByte, // CRC-32: Cyclic Redundancy Check
+      ((crc32Value >> 8) & 0xff).toByte,
+      ((crc32Value >> 16) & 0xff).toByte,
+      ((crc32Value >> 24) & 0xff).toByte,
+      (bytesIn & 0xff).toByte, // ISIZE: Input size
+      ((bytesIn >> 8) & 0xff).toByte,
+      ((bytesIn >> 16) & 0xff).toByte,
+      ((bytesIn >> 24) & 0xff).toByte
+    )
+    Stream.chunk(asChunkBytes(trailer))
+  }
+
+  final case class GunzipResult[F[_]](
+      content: Stream[F, Byte],
+      modificationTime: Option[Instant] = None,
+      fileName: Option[String] = None,
+      comment: Option[String] = None
+  )
+
+  /**
+    * Returns a pipe that incrementally decompresses input according to the GZIP
+    * format as defined by RFC 1952 at https://www.ietf.org/rfc/rfc1952.txt. Any
+    * errors in decompression will be sequenced as exceptions into the output
+    * stream. Decompression is handled in a streaming and async fashion without
+    * any thread blockage.
+    *
+    * The chunk size here is actually really important. Matching the input stream
+    * largest chunk size, or roughly 8 KB (whichever is larger) is a good rule of
+    * thumb.
+    *
+    * @param bufferSize The bounding size of the input buffer. This should roughly
+    *                   match the size of the largest chunk in the input stream.
+    *                   This will also be the chunk size in the output stream.
+    *                    Default size is 32 KB.
+    * @return    Returns a stream containing a single GzipResult containing the
+    *            properties of the content stream. Continue to stream the content
+    *            by flat mapping it:
+    *            stream
+    *             .through(gunzip[IO]())
+    *             .flatMap { gunzipResult =>
+    *               // Access properties here.
+    *               gunzipResult.content
+    *             }
+    */
+  def gunzip[F[_]](
+      bufferSize: Int = 1024 * 32
+  )(
+      implicit SyncF: Sync[F]
+  ): Stream[F, Byte] => Stream[F, GunzipResult[F]] =
+    stream =>
+      Stream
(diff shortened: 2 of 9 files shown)


core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +105,658 @@ object compress {
...
-    * by delegating to `java.util.zip.GZIPOutputStream`. Output is compatible
-    * with the GNU utils `gunzip` utility, as well as really anything else that
-    * understands GZIP. Note, however, that the GZIP format is not "stable" in
-    * the sense that all compressors will produce identical output given
-    * identical input. Part of the header seeding is arbitrary and chosen by
-    * the compression implementation. For this reason, the exact bytes produced
-    * by this pipe will differ in insignificant ways from the exact bytes produced
-    * by a tool like the GNU utils `gzip`.
+    * Returns a `Pipe` that gzips (compresses) its input elements as defined by RFC 1952:
+    * https://www.ietf.org/rfc/rfc1952.txt
```
djspiewak: This is actually clearer than what was there before? Shouldn't we just add the RFC link to what's there?

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +105,658 @@ object compress {
...
     *
-    * @param bufferSize The buffer size which will be used to page data
-    *                   from the OutputStream back into chunks. This will
-    *                   be the chunk size of the output stream. You should
-    *                   set it to be equal to the size of the largest
-    *                   chunk in the input stream. Setting this to a size
-    *                   which is ''smaller'' than the chunks in the input
-    *                   stream will result in performance degradation of
-    *                   roughly 50-75%.
+    * @param bufferSize size of the internal buffer that is used by the compressor. Default size is 32 KB.
```
djspiewak: Ditto above.

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +105,658 @@ object compress {
...
-    * The chunk size here is actually really important. If you set it to be too
-    * small, then there will be insufficient buffer space for `GZIPInputStream` to
-    * read the GZIP header preamble. This can result in repeated, non-progressing
-    * async suspensions. This case is caught internally and will be raised as an
-    * exception (`NonProgressiveDecompressionException`) within the output stream.
-    * Under normal circumstances, you shouldn't have to worry about this. Just, uh,
-    * don't set the buffer size to something tiny. Matching the input stream largest
-    * chunk size, or roughly 8 KB (whichever is larger) is a good rule of thumb.
+    * Returns a `Pipe` that gunzips (decompresses) its input elements as defined by RFC 1952:
+    * https://www.ietf.org/rfc/rfc1952.txt
```
djspiewak: Ditto above

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +105,658 @@ object compress {
...
+    * @param bufferSize size of the internal buffer that is used by the decompressor. Default size is 32 KB.
+    * @param modificationTimeCallback callback to provide the optional file modificationt time, if included in the gzip stream
+    * @param fileNameCallback callback to provide the optional file name, if included in the gzip stream.
+    * @param commentCallback callback to provide the optional file comment, if included in the gzip stream.
+    */
+  def gunzip[F[_]: RaiseThrowable](
+      bufferSize: Int = 1024 * 32,
+      modificationTimeCallback: Instant => Unit = _ => (),
+      fileNameCallback: String => Unit = _ => (),
+      commentCallback: String => Unit = _ => ()
```
djspiewak: In general, I think callbacks like this are not great. At the very least, they should return `F[Unit]` so that effects can be properly controlled and sequenced. But aside from that, I would definitely prefer if there were a way to sequence this into the stream more properly.

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +105,658 @@ object compress {
...
+              Pull.raiseError(new ZipException("Not in gzip format"))
           }
-          val pushed = abis.push(arr)
-          if (!pushed) throw NonProgressiveDecompressionException(bufferSize)
-        }
+        case None =>
+          Pull.raiseError(new EOFException())
+      }
+      .stream
+      .flatten
```
djspiewak: Not to be That Guy™, but it's probably worth tossing together a quick benchmark to show that this implementation (which is inarguably nicer than what was there before!) doesn't cause any performance regressions. The previous implementation delegated all of the state management work to `GZIPOutputStream`, which has a reasonably well-optimized internal loop. This implementation is a lot nicer, but also does quite a bit more effect sequencing, which is usually very slow, and it's not entirely clear to me how much of it is in the hot path. Benchmarks would show for sure.

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +105,658 @@ object compress {
...
+      case name if name.indexOf("mac") >= 0 => MACINTOSH
+      case _                                => UNKNOWN
+    }
+  }
+  private val gzipInputCrcBytes = 4
+  private val gzipInputSizeBytes = 4
+  private val gzipTrailerBytes = gzipInputCrcBytes + gzipInputSizeBytes
+
+  private val zeroByte: Byte = 0
+  private val emptyByteChunk: Chunk.Bytes = Chunk.Bytes(Array.empty[Byte], 0, 0)
```
djspiewak: `Chunk.empty[Byte]`?

core/jvm/src/test/scala/fs2/CompressSpec.scala
```diff
@@ -73,93 +85,426 @@ class CompressSpec extends Fs2Spec {
...
+        .rechunkRandomlyWithSeed(0.1, 2)(System.nanoTime())
+        .through(compress.deflate())
+        .rechunkRandomlyWithSeed(0.1, 2)(System.nanoTime())
+        .through(classic.inflate())
+        .compile
+        .toVector
+        .asserting(_ shouldBe s.toVector)
+    }
+
+    "deflate performance" - {
```
djspiewak: This really isn't a very good way to measure performance of *either* implementation, just due to the way the JVM works. JMH is a considerably better tool for this, and I believe there's already a **benchmarks** subproject.

core/jvm/src/test/scala/fs2/CompressSpec.scala
```diff
@@ -73,93 +85,426 @@ class CompressSpec extends Fs2Spec {
...
+              comment = Some(comment)
+            )
+          )
+          .compile
+          .to(Array)
 
-      val bis = new ByteArrayInputStream(bytes)
-      val gzis = new GZIPInputStream(bis)
+        val bis = new ByteArrayInputStream(bytes)
+        val gzis = new GZIPInputStream(bis)
```
djspiewak: Did scalafmt really double-indent this?

core/jvm/src/test/scala/fs2/CompressSpec.scala
```diff
@@ -168,5 +513,454 @@ class CompressSpec extends Fs2Spec {
...
+        .toVector
+        .asserting { vector =>
+          actualFilename shouldBe expectedFilename
+          actualComment shouldBe expectedComment
+          actualModificationTime shouldBe expectedModificationTime
+          new String(vector.toArray, StandardCharsets.US_ASCII) shouldBe expectedContent
+        }
+    }
+
+    "gzip performance" - {
```
djspiewak: Ditto above w.r.t. JMH and performance measurement.

core/jvm/src/test/scala/fs2/CompressSpec.scala
```diff
@@ -168,5 +513,454 @@ class CompressSpec extends Fs2Spec {
...
+          Pull.output(Chunk.bytes(Array.fill(chunkSize)(byte))) >> go(nextBytesRemaining)
+        else Pull.output(Chunk.bytes(Array.fill(bytesRemaining.toInt)(byte)))
+      }
+      go(bytes).stream
+    }
+
+  }
+}
+
+object classic {
```
djspiewak: This seems unnecessary. The tests as they are written without `classic` are sufficient coverage. It would be *nice* to have an optional test which delegates to the `gzip` command if it is present on the `PATH` (we did this for fs2-gzip originally, but representing it in ScalaTest is harder than in Specs2).

djspiewak:
This is exciting! It is much, *much* nicer than @wemrysi's and my original hacky continuation suspension based implementation.

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +105,658 @@ object compress {
...
+      bytesChunk.offset,
+      bytesChunk.length
+    )
+    def inflateInto(inflatedBuffer: Array[Byte]): Int =
+      if (inflater.finished()) -2
+      else if (inflater.needsInput()) -1
+      else {
+        val byteCount = inflater.inflate(inflatedBuffer)
+        crc32.foreach(crc =>
+          crc.synchronized {
```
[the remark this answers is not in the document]
greg-a-atkinson: Agreed. Synchronization removed.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Original documentation restored with the addition of the RFC link.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Original documentation restored.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Original documentation restored with the addition of the RFC link and the removal of the references to GZIPInputStream.

This gunzip implementation doesn't appear to as sensitive to small buffer sizes. I've had success with chunk sizes of as little as 64 bytes, but it's difficult to determine the minimum buffer size for decompressing all possible input. With a quick online search I managed to find some C# documentation that recommended a minimum 128 byte buffer, but this is tenuous at best and I wouldn't trust much other than actual implementation documentation.

Furthermore, I found that for this implementation it didn't make much different to double the size of the input buffer as the size of the output buffer because the output is chunked out and it's difficult to predict the input compression ratio.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Agreed, I don't love this callback approach either, but don't have a solution for this yet.

Regarding wrapping the the header property values in F, I'm not sure I know how to do this. F is currently only a RaiseThrowable, how F.unit or F.pure be used? There is a similar problem preventing the Inflater and Deflater instances from being bracketed. Is it possible to further constrain F while maintaining compatibility - what am I missing?

Regarding an alternative to callback parameters and sequencing the stream, what do you think about removing the callback parameters and changing the return type to return a stream of a single tuple (or an appropriate structure) that would return the materialised header properties and the remaining stream? I'd have to tinker with it, but this should be possible, unless there is a better idea. An additional benefit of this approach would be that it would be possible to discard the compressed content based on the content of the header properties.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Agreed. Replaced and inlined.

core/jvm/src/test/scala/fs2/CompressSpec.scala
greg-a-atkinson: Yes, scalafmt made the change and it looks correct upon inspection.

core/jvm/src/test/scala/fs2/CompressSpec.scala
greg-a-atkinson: Agreed, this is not a proper benchmark. I added these for a rough comparison and sanity check. The real before and after performance metrics were gathered by profiling with Oracle Java Mission Control's Flight Recorder. At one point compress.inflate was on of the hottest areas of code in a application heavy on streaming compressed json from networked sources.

Would you prefer I just remove these false performance tests along with the classic implementation?

core/jvm/src/test/scala/fs2/CompressSpec.scala
greg-a-atkinson: As above. Would you prefer I just remove these false performance tests along with the classic implementation?

core/jvm/src/test/scala/fs2/CompressSpec.scala
greg-a-atkinson: Agreed, this can be removed. Retaining the original implementation helped sanity check compatibility.

Integrating with the system's gzip/gunzip if on the path is a nice idea and much more powerful that the one test I added to decompress the rudimentary unix.gzip file I created and copied as hex bytes into the tests.

core/jvm/src/main/scala/fs2/compress.scala
kubukoz: > F is currently only a RaiseThrowable, how F.unit or F.pure be used?

I think you could go with a "partially applied" method, so this one wouldn't take callbacks (it would take `Applicative[F]`) but would create an instance of a class that has an `apply` that takes callbacks (in there, you could use the Applicative instance in default parameters).

greg-a-atkinson:
@djspiewak thanks your review and comments. I've replied to each comment having made most of the changes and with some ideas for a couple. Let me know what you think?

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -25,45 +22,74 @@ object compress {
...
       strategy: Int = Deflater.DEFAULT_STRATEGY
-  ): Pipe[F, Byte, Byte] = { in =>
-    Pull.suspend {
-      val deflater = new Deflater(level, nowrap)
-      deflater.setStrategy(strategy)
-      val buffer = new Array[Byte](bufferSize)
-      _deflate_stream(deflater, buffer)(in)
-    }.stream
+  ): Pipe[F, Byte, Byte] = stream => {
+    val deflater = new Deflater(level, nowrap)
```
mpilquist: Looks like this needs a `Stream.suspend` to ensure each use gets its own `deflater` instance. E.g., consider `val p = compress.deflate(); val x = s.through(p); val y = x.toList, val z = x.toList`

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -25,45 +22,74 @@ object compress {
...
+    deflater.setStrategy(strategy)
+    _deflate(deflater, bufferSize, crc32 = None)(stream) ++
+      Stream.suspend {
+        deflater
+          .end() // If F is narrowed to Applicative then deflater.end() can be bracketed rather than appended in an empty stream.
```
mpilquist: What happens if this never gets called? E.g., `src.through(compress.deflate).take(10)`

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +101,680 @@ object compress {
     */
   def inflate[F[_]](nowrap: Boolean = false, bufferSize: Int = 1024 * 32)(
       implicit ev: RaiseThrowable[F]
+  ): Pipe[F, Byte, Byte] = stream => {
+    val inflater = new Inflater(nowrap)
```
mpilquist: Same as with `deflater` - wrap with `Stream.suspend`

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +101,680 @@ object compress {
...
+            inflater.getRemaining match {
+              case bytesRemaining if bytesRemaining > 0 =>
+                Pull.output(asChunkBytes(inflatedBuffer, inflatedBytes)) >>
+                  Pull.output(
+                    Chunk.Bytes(
+                      bytesChunk.values,
+                      bytesChunk.offset + bytesChunk.length - bytesRemaining,
+                      bytesRemaining
+                    )
+                  ) >> Pull.done
```
mpilquist: Shouldn't need the `>> Pull.done` or below

core/jvm/src/main/scala/fs2/compress.scala
```diff
@@ -75,197 +101,680 @@ object compress {
...
+        .getOrElse(Stream.empty) ++
+      Stream.chunk(asChunkBytes(crc16))
+  }
 
-        val trailer: Stream[F, Byte] = Stream.suspend {
-          gzos.close()
-          slurpBytes
-        }
+  private def _gzip_trailer[F[_]](deflater: Deflater, crc32: CRC32): Stream[F, Byte] =
+    Stream.suspend {
```
mpilquist: No need for suspend here as I don't see anything mutable going on

core/jvm/src/main/scala/fs2/compress.scala
mpilquist: :+1: for returning a case class with the header properties and the body stream.

core/jvm/src/main/scala/fs2/compress.scala
mpilquist: Looks like openjdk allocates native memory for both `Inflater` / `Deflater` so we really need to bracket this somehow in order to ensure that native memory is released.

greg-a-atkinson: Agreed. That's correct, not calling deflater.end() or inflater.end() risks (will) leaking native memory - an issue present in the existing implementation and why I'd like to constrain F to Applicative so that the deflate/inflate/gzip/gunzip streams can be bracketed, but this introduces a binary incompatibility.

mpilquist: We'd need a full `F[_]: Sync` -- we need the `delay` constructor.

We could leave the old implementation in place and deprecate it and then define the new implementation to require the `Sync` constraint.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Agreed. Stream.suspend removed.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Agreed. Both removed. Thank you.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: I'm not sure. Deflate and inflate return a function so a new Deflater is created for each use. I've added tests "deflate and inflate are reusable" and "gzip and gunzip are reusable." Let me know if I've misunderstood. Thanks

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: I'm not sure. Deflate and inflate return a function so a new Deflater is created for each use. I've added tests "deflate and inflate are reusable" and "gzip and gunzip are reusable." Let me know if I've misunderstood. Thanks

mpilquist: In the example, the function is only called once `val x = s.through(p)` and then "run" twice. Here's a simplified example:

```scala
@ def pipe[F[_]]: Pipe[F, Int, Int] = in => {
    var cnt = 0
    in.map { i => cnt += 1; i } ++ Stream(cnt)
  }
defined function pipe

@ Stream(1, 1, 1).through(pipe).toList
res2: List[Int] = List(1, 1, 1, 3)

@ val x = Stream(1, 1, 1).through(pipe)
x: Stream[Nothing, Int] = Stream(..)

@ x.toList
res4: List[Int] = List(1, 1, 1, 3)

@ x.toList
res5: List[Int] = List(1, 1, 1, 6)

@ x.toList
res6: List[Int] = List(1, 1, 1, 9)
```

core/jvm/src/test/scala/fs2/CompressSpec.scala
djspiewak: > Would you prefer I just remove these false performance tests along with the classic implementation?

I think that would definitely be better. Also I'm curious as to your JFR results!

core/jvm/src/test/scala/fs2/CompressSpec.scala
djspiewak: Yep! Obviously better to convert them into real JMH measurements, but I'm cool with just nuking them from orbit.

core/jvm/src/main/scala/fs2/compress.scala
djspiewak: Can it be done with a `Stream.bracket` instead?

core/jvm/src/main/scala/fs2/compress.scala
djspiewak: > F is currently only a RaiseThrowable, how F.unit or F.pure be used

`F` is `RaiseThrowable` *here*, but the call site probably knows something more about it! For example, the call site may know that `F` is actually `IO`, which would allow it to wrap side effects into `F[Unit]` values. But either way the case class idea is way better.

core/jvm/src/main/scala/fs2/compress.scala
mpilquist: @djspiewak Yeah, it would be done via `Stream.bracket` -- the reason we need a `Sync[F]` is so we can lift creation (e.g. `Sync[F].delay(new Inflater)`) and disposal (e.g. `Sync[F].delay(inflater.end())`) in an effect. I assume the `Applicative[F]` suggestion was in order to get access to `pure` but that's not enough -- we'd need `delay`.

core/jvm/src/main/scala/fs2/compress.scala
mpilquist: The reply [here](https://github.com/functional-streams-for-scala/fs2/pull/1783#discussion_r376115069) was intended to be in response to this question. Sorry for the confusion!

core/jvm/src/main/scala/fs2/compress.scala
djspiewak: Ahhh, I see what you mean.

core/jvm/src/test/scala/fs2/CompressSpec.scala
greg-a-atkinson: fs2_inflate

Of course. Attached is a screenshot of the JFR hot methods trace that started this effort, namely attempting to avoid the mutable ArrayBuffer used to accumulate the bytes produced by Inflater. The context was streaming multi-megabyte and multi-gigabyte compressed json documents.

In the spirit of fs2 and to draw on all your insight and improvements it would be good to feedback these enhancements that are already in use in this area.

While this PR relates to efficiently chunking the compressed byte stream, a bigger challenge was efficiently recombining the chunked byte stream with constant time complexity. If there's interest, I can raise a PR for BytesVector (signature below) with constant time append and Chunk implementations. 

`final case class Chunk.BytesVector(vector: Vector[Chunk.Bytes]) extends Chunk[Byte] with KnownElementType[Byte]`

The combination of these enhancements made it possible to build efficient streaming json parsing from wire to instance for very large objects. Chunking byte streams has historically been the domain of C-style array-length-offset arithmetic and mutable arrays, which is difficult to compete with from the perspective of raw performance. But in the context of integrating raw byte streams with complex and useful applications, like can be done by building streams with fs2, a better performing overall result can be achieved.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: You're right, thanks. I'm going to replace the callback parameters with a case class return...

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: I'll be happy to. Once the implementation review feedback has been completed I'll add benchmarks...

core/jvm/src/test/scala/fs2/CompressSpec.scala
greg-a-atkinson: Agreed. I'll replace these at the end with benchmarks...

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Ah, thanks for the second explanation. I expert this will be addressed by your suggestion to use Sync.delay, but I'll make sure this case is explicitly confirmed by the tests.

core/jvm/src/test/scala/fs2/CompressSpec.scala
greg-a-atkinson: `classic` implementation no longer required in the tests as it needs to remain remain to be deprecated for the necessary type changes to be implemented in fs2.compress2. (Opinions on naming are welcome.)

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Agreed. fs2.compress and CompressSpec have been restored to their original implementations, for the necessary type changes to be implemented in fs2.compress2 - naming opinions are welcome.

core/jvm/src/test/scala/fs2/CompressSpec.scala
mpilquist: Have you looked at `Chunk.Queue[Byte]`? It's a wrapper around a `scala.collection.immutable.Queue[Chunk[A]]` with efficient conversion to a single array backed chunk.

mpilquist: Another option is to use `Chunk.ByteVector` and just append `ByteVector`s as they are received.

djspiewak: > The combination of these enhancements made it possible to build efficient streaming json parsing from wire to instance for very large objects. Chunking byte streams has historically been the domain of C-style array-length-offset arithmetic and mutable arrays, which is difficult to compete with from the perspective of raw performance. But in the context of integrating raw byte streams with complex and useful applications, like can be done by building streams with fs2, a better performing overall result can be achieved.
 
Worth noting that this is basically *directly* [in my wheelhouse](https://github.com/slamdata/tectonic). Tectonic started as a fork of Jawn, and basically uses the same approach: byte chunks are "absorbed" into the main contiguous buffer, which is then parsed until some partial-failure is hit. Every partial-success (e.g. successfully parsing a key, a `:`, a `}`, etc) is a stateful safepoint within the parser, advancing the cursor and persisting (in registers) all the information required to resume parsing from that offset. Thus, when we run off the end with a partial result, we can just hard-fail with an exception (which is cached). The exception effectively just routes control flow back to the stream `Pull`, asynchronously suspending the parser until the next chunk comes in.

All parsing operations are done against a straight, contiguous byte array (which grows up to 1 MB). The downside to this approach is that it isn't possible to avoid paging the bytes fully into memory (e.g. for mmap'd files and such), but in practice `ByteBuffer` is insanely slow anyway, even when you play tricks to get monomorphic access.

Anyway, performance in practice saturates the IO bus, even when performing relatively complex transformations on the *results* of the parse.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Right, the functions fs2.compress2.(deflate|inflate|gzip|gunzip) are now bracketed.

core/jvm/src/main/scala/fs2/compress.scala
greg-a-atkinson: Right, callback parameters removed from fs2.compress2.gunzip and replaced with a returned case class containing the materialised header fields and a body stream. Opinions on naming welcome.

core/jvm/src/test/scala/fs2/CompressSpec.scala
greg-a-atkinson: Let's say the json document being streamed is 2GB. It's not ideal to materialise 65k x 32KB chunks into a single array and even if that is only done once it will still require at least 4GB of memory - so it's not ideal to use Chunk.Queue[Byte] in this context. Chunk.ByteVectorChunk's ByteVector documents logarithmic time complexity for most operations, but I'll run a comparison when I get a chance.

Performing streaming tokenization for streaming parsing of the large json documents doesn't require materialising the entire document into one array or one String as only localised concatenation is required for tokens that span sequential chunks, which can be achieved with a view (similar to ByteVector's view approach) that acts as a sliding window over the localised scope of the streaming tokenizer. The single class, Vector-backed implementation of Chunk[Byte] I mentioned is nowhere near the scope of what I read Tectonic to be, but I'll still be happy to share it so it can be improved or replaced.

djspiewak: > Let's say the json document being streamed is 2GB. It's not ideal to materialise 65k x 32KB chunks into a single array and even if that is only done once it will still require at least 4GB of memory

Tectonic only does it up to a maximum bound of 1 MB. It discards anything that it has already parsed, including fragments. In order to create a true overflow of the type you suggest, you would need a single *json string* which is 2 GB in size, which is quite rare. Basically, Tectonic is doing exactly the sliding window you're talking about, but with a raw array and width-sensitive to the exact token state.

> The single class, Vector-backed implementation of Chunk[Byte] I mentioned is nowhere near the scope of what I read Tectonic to be, but I'll still be happy to share it so it can be improved or replaced.

I agree. And more relevantly, Tectonic's internal byte paging machinery is quite complex and specialized. If what you've implemented is a more general mechanism that achieves the same thing with similar performance but without the manic complexity in Tectonic, that would be really wonderful.

greg-a-atkinson: A good discussion, thanks. So, a single json string of 2GB is commonplace for my use case, hence this PR suggestion. I would have to go look what the maximum currently is, but I suspect I would fine two or three times that size. Why single json strings of that size is a different discussion entirely.

The chunk implementation that I found useful to support streaming tokenisation into streaming parsing of multi-gig json objecs is here: https://github.com/functional-streams-for-scala/fs2/pull/1789 - I would need to add a bunch of tests to ensure nothing is broken in this incarnation, so for now treat the PR as worthy of discussion only.

This BytesVector implementation enables tokenisation without array copying or concatenation because the tokens can be backed by the original byte chunks of the underlying stream, even when those tokens straddle byte chunks or span multiple byte chunks. If found it useful to constrain the implementation to always have an effective offset of zero and a total length equal to the sum of the length of the underlying Chunk.Byte instances. This reduced the cost complexity of of implementing the Chunk interface and ensured that only the visible chunks and their visible bytes for the required offset and length were retained and the out-of-scope chunks ignored. This was implemented by rewriting the leading and trailing Chunk.Byte to maintain only the in-scope bytes.

Once tokenised, parsing is just a matter of scanning the token and rewriting the escaped characters, depending on the token type - but you're way more familiar with Jawn, and as you've mentioned Tectonic, than I am. But, 99% of the time tokens do not contain escaped characters or multi-byte characters and are exactly the length of the parsed node, which is a good hint at the parsed node size. I imagine that with a UTF-16 encoding constraint and a String implementation that didn't instantiate and copy into a new array it would be possible to back the parsed nodes with the original byte chunks, but this would introduce new challenges. (Not that I haven't see java.lang.String rewritten and replaced for similar challenges before.)

greg-a-atkinson: Thanks to the suggestion by @mpilquist, scodec.bits.ByteVector performed better for my use case. A rough, but reasonable performance comparison using a single 416MB json document with 38,831,305 tokens, was tokenised by the proposed Chunk.BytesVector in 11.894 seconds and by the recommended codec.bits.ByteVector in 7.871 seconds.

djspiewak: So around 423 Mbps, and your average token size is 11.23 bytes. Very impressive.

core/jvm/src/test/scala/fs2/CompressSpec.scala
greg-a-atkinson: jmh:run -i 1 -f 1 -wi 1 fs2.benchmark.CompressBenchmark fs2.benchmark.Compress2Benchmark

bytes = 1 MiB
bufferSize = 32 kiB
withRandomBytes => random bytes or zero bytes

```
[info] Benchmark                   (withRandomBytes)   Mode  Cnt     Score   Error  Units

[info] Compress2Benchmark.deflate               true  thrpt         29.175          ops/s
[info] Compress2Benchmark.deflate              false  thrpt        169.255          ops/s
[info] Compress2Benchmark.inflate               true  thrpt       1317.750          ops/s
[info] Compress2Benchmark.inflate              false  thrpt        385.535          ops/s
[info] Compress2Benchmark.gzip                  true  thrpt         27.654          ops/s
[info] Compress2Benchmark.gzip                 false  thrpt        172.131          ops/s
[info] Compress2Benchmark.gunzip                true  thrpt       2431.867          ops/s
[info] Compress2Benchmark.gunzip               false  thrpt        384.544          ops/s

[info] CompressBenchmark.deflate                true  thrpt          6.849          ops/s
[info] CompressBenchmark.deflate               false  thrpt        160.612          ops/s
[info] CompressBenchmark.inflate                true  thrpt          7.231          ops/s
[info] CompressBenchmark.inflate               false  thrpt          7.352          ops/s
[info] CompressBenchmark.gzip                   true  thrpt         26.660          ops/s
[info] CompressBenchmark.gzip                  false  thrpt        171.284          ops/s
[info] CompressBenchmark.gunzip                 true  thrpt       1045.854          ops/s
[info] CompressBenchmark.gunzip                false  thrpt        461.132          ops/s
```

core/jvm/src/main/scala/fs2/compress2.scala
```diff
@@ -0,0 +1,862 @@
...
+    *             .flatMap { gunzipResult =>
+    *               // Access properties here.
+    *               gunzipResult.content
+    *             }
+    */
+  def gunzip[F[_]](
+      bufferSize: Int = 1024 * 32
+  )(
+      implicit SyncF: Sync[F],
+      RaiseThrowableF: RaiseThrowable[F]
```
mpilquist: Note, we don't need `RaiseThrowable[F]` here since we have `Sync[F]`.

greg-a-atkinson: Agreed. Thanks for highlighting this. The change has been applied, the branch rebased and the pull request green.

core/jvm/src/main/scala/fs2/compress2.scala
```diff
@@ -0,0 +1,863 @@
...
+    crc32.foreach(_.update(bytesChunk.values, bytesChunk.offset, bytesChunk.length))
+
+    def isDone: Boolean =
+      (isFinalChunk && deflater.finished) || (!isFinalChunk && deflater.needsInput)
+
+    def deflateInto(deflatedBuffer: Array[Byte]): Int =
+      if (isDone) 0 else deflater.deflate(deflatedBuffer)
+
+    def pull(): Pull[F, Byte, Unit] = {
+      val deflatedBuffer = new Array[Byte](deflatedBufferSize)
```
mpilquist: Thoughts on reusing this array for each call to `pull` instead of allocating a new one each time? Similar question with inflation.

mpilquist:
Anyone have any other naming ideas besides `compress2`?

SystemFw:
> How about fs2.compression.{gunzip,gzip,inflate,deflate}?

I like that

gvolpe:
> How about fs2.compression.{gunzip,gzip,inflate,deflate}?

Those names are good :+1:

greg-a-atkinson:
The naming change looks good.

core/jvm/src/main/scala/fs2/compress2.scala
greg-a-atkinson: What happens if prefetchN is called on the stream? Would the stream be corrupted by the caller calling prefetchN if the underlying byte array was reused?

core/jvm/src/main/scala/fs2/compression.scala
```diff
@@ -0,0 +1,863 @@
...
+    crc32.foreach(_.update(bytesChunk.values, bytesChunk.offset, bytesChunk.length))
+
+    def isDone: Boolean =
+      (isFinalChunk && deflater.finished) || (!isFinalChunk && deflater.needsInput)
+
+    def deflateInto(deflatedBuffer: Array[Byte]): Int =
+      if (isDone) 0 else deflater.deflate(deflatedBuffer)
+
+    def pull(): Pull[F, Byte, Unit] = {
+      val deflatedBuffer = new Array[Byte](deflatedBufferSize)
```
jgogstad: I saw this in the previous implementation as well, isn't this a bottleneck?

If I stream `n` GiB of data, this is going to allocate and garbage collect `n` GiB of memory. What are your thoughts on using a single mutable `Array` as a buffer, instead of allocating and garbage collecting a new buffer on each pull?

djspiewak: It defaults to 32 KB in size. It also gets wrapped up and emitted as the output chunk, so it cannot be shared.

mpilquist: BTW, I implemented the shared buffer approach before merging and then realized exactly what Daniel pointed out -- the buffer is emitted to the output so if we use a shared buffer, then we'd need to copy it when creating a chunk.

PR #1785, opened 2020-02-05
Override Chunk.BufferChunk.iterator to return buffer.iterator

greg-a-atkinson:
Override Chunk.BufferChunk.iterator to return buffer.iterator.

Chunk.iterator reasonably assumes that the underlying collection is indexed and iterates through the chunk with sequential calls to apply(), but Chunk.BufferChunk can be created from a ListBuffer which has a non-constant time apply() implementation. Because Chunk.flatMap makes use of Chunk.iterator, flat mapping over a ListBuffer backed BufferChunk has high cost complexity. The simplest solution is to override BufferChunk's iterator implementation with the iterator implementation of the underlying buffer.

Alternatively or subsequently, there may be a gain in having different chunk implementations for Buffer and IndexedBuffer.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -624,6 +624,10 @@ object Chunk extends CollectorK[Chunk] {
       else if (n >= size) Chunk.empty
       else buffer(b.drop(n))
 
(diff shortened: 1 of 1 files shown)


greg-a-atkinson:
Thank you :)

PR #1786, opened 2020-02-06
Added flatMapOrDone and flatMapOrNone to Pull

mpilquist:
I don't feel strongly about adding these. What does everyone think?


SystemFw:
I'm a bit unsure of the names, sometimes when you need to be this literal it might be a sign that something it's not worth naming in the first place.

Still unsure about whether they should be included or not, but in general I like `flatMapOrDone` better that `flatMapOrNone`

mpilquist:
I'm leaning towards closing this one. While the pattern encapsulated here shows up a bunch, I think it's clearer as-is than with these adhoc combinators.

PR #1789, opened 2020-02-09
Add Chunk.BytesVector as a view into concatenated Chunk.Bytes instances.

greg-a-atkinson:
A chunk implementation representing concatenated Chunk.Bytes instances as an alternative to copying them into new arrays:

`final case class BytesVector private (chunks: Vector[Chunk.Bytes], length: Int) extends Chunk[Byte]`

PR raised for discussion and not yet ready to me merged without tests and further input.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -784,6 +784,268 @@ object Chunk extends CollectorK[Chunk] {
     def apply(values: Array[Byte]): Bytes = Bytes(values, 0, values.length)
   }
 
+  final case class BytesVector private (chunks: Vector[Chunk.Bytes], length: Int)
+      extends Chunk[Byte]
+      with KnownElementType[Byte] {
+
+    def elementClassTag: ClassTag[Byte] = ClassTag.Byte
+
+    def size: Int = length
+
+    def apply(i: Int): Byte = at(i)
+
+    val at: Int => Byte =
+      chunks.length match {
+        case 0 => _ => throw new IndexOutOfBoundsException()
+        case 1 => i => chunks(0).at(i)
+        case _ => i => at(i, chunks)
+      }
+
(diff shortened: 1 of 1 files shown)


mpilquist:
I'd like to see the performance of this compared to using `Chunk.ByteVector` directly, which wraps a single `scodec.bits.ByteVector`. I strongly suspect that using `scodec.bits.ByteVector` and its `++` operation along with its `view` constructor will generally beat using a `Vector[Chunk[Byte]]`. The `Chunk.ByteVector` class could probably be optimized some more like the implementation in this PR.

greg-a-atkinson:
> I'd like to see the performance of this compared to using `Chunk.ByteVector` directly, which wraps a single `scodec.bits.ByteVector`. I strongly suspect that using `scodec.bits.ByteVector` and its `++` operation along with its `view` constructor will generally beat using a `Vector[Chunk[Byte]]`. The `Chunk.ByteVector` class could probably be optimized some more like the implementation in this PR.

Agreed. This is something I’m going to try. Thanks for pointing it out.

greg-a-atkinson:
> I'd like to see the performance of this compared to using `Chunk.ByteVector` directly, which wraps a single `scodec.bits.ByteVector`. I strongly suspect that using `scodec.bits.ByteVector` and its `++` operation along with its `view` constructor will generally beat using a `Vector[Chunk[Byte]]`. The `Chunk.ByteVector` class could probably be optimized some more like the implementation in this PR.

A rough, but reasonable performance comparison using a single 416MB json document with 38,831,305 tokens, was tokenised by the proposed Chunk.BytesVector in 11.894 seconds and by the recommended codec.bits.ByteVector in 7.871 seconds. I’m happy with that and even more so that ByteVector was a drop-in replacement. Thanks!

mpilquist:
Awesome! Thanks for trying that experiment. `ByteVector` is fairly well optimized. ;)

Issue #1765, opened 2020-01-28
Add cats.Align instance for fs2.Stream


LukaJCB:
Should be fairly similar to the `zip` implementation

bobbyrauchenberg:
i'd be happy to pick this up

bobbyrauchenberg:
I'm new to FS2 so apologies if it's a stupid question, but is there a case where it's possible to return anything other than `Both[A, B]`? Or is returing Align.Left and Align.Right exactly the point of this work?

What would we do in the case where we have 

```scala
streamA.align(streamB)
```

where `streamA` is `Stream.constant("1")` and `streamB` is `Stream.empty` - looking at the docs for Align, we'd be returning an infinite Stream of `Align.Left(1)`

LukaJCB:
@bobbyrauchenberg good question! 
You are right that returning `Ior.Left` and `Ior.Right` in addition to `Ior.Both` is exactly the point of this work. To answer your question, yes you would get an infinite stream of `Left(1)`.

kubukoz:
Hey @bobbyrauchenberg, that seems to me like the correct behavior. The longer of the two streams will result in having values as `Left` xor `Right` at the end, but the prefix will always be Both. This could probably be made into a nice property test.

mpilquist:
This was fixed by #1791

PR #1791, opened 2020-02-12
cats.Align instance for Stream

bobbyrauchenberg:
This is for #1765

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -15,6 +15,8 @@ import java.io.PrintStream
 import scala.annotation.tailrec
 import scala.concurrent.TimeoutException
 import scala.concurrent.duration._
+import cats.data.Ior
+import cats.data.Ior.Both
 
 /**
   * A stream producing output of type `O` and which may evaluate `F` effects.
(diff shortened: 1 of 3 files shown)


core/shared/src/test/scala/fs2/StreamSpec.scala
```diff
@@ -3628,6 +3630,70 @@ class StreamSpec extends Fs2Spec with Matchers {
...
+        assert(
+          s.align(empty).take(3).toList == List(Ior.Left("A"), Ior.Left("B"), Ior.Left("C"))
+        )
+      }
+
+      "values in both sides" in {
+        val ones = Stream.constant("1")
+        val s = Stream("A", "B", "C")
+        assert(
+          s.align(ones).take(3).toList == List(
```
LukaJCB: Can we do `take(4)` and then add the `Ior.Right("1")` element?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4717,6 +4719,81 @@ object Stream extends StreamLowPriority {
...
 
+  /** `Align` instance for `Stream`.
+    * * @example {{{
+    * scala> import cats.implicits._
+    * scala> Stream(1,2,3).align(Stream("A","B","C","D","E")).toList
+    * res0: List[cats.data.Ior[Int,String]] = List(Both(1,A), Both(2,B), Both(3,C), Right(D), Right(E))
+    * }}}
+    *
+    */
+  implicit def alignInstance[F[_]](implicit F: Functor[Stream[F, ?]]): Align[Stream[F, ?]] =
```
kubukoz: Can't check now, but doesn't Stream have a Functor "just like that", without any parameters?

bobbyrauchenberg:
I guess I need to rebase #1792 into my branch once it's merged

PR #1803, opened 2020-02-27
Include fd4s/fs2-kafka in the list of projects that use fs2

usommerl:
There are two projects with the same name that provide Apache Kafka bindings for fs2. I feel like both should be included in the list of projects that use fs2 because that is something that puzzled me as a new user. I included the projects in lexicographical order, so the ordering should not be interpreted as a matter of personal preference.


mpilquist:
Thanks!

PR #1806, opened 2020-03-04
Updated Scala.js to 1.0.0

disblader:
Hello,

I attempted to upgrade to scala.js 1.0.0 release. 

With 1.0.0 scala.js removed the global `isScalaJSProject` setting in favour of using `crossProjectPlatform == JSPlatform` check. Unfortunately, `crossProjectPlatform` is only defined on crossProjects, so any non cross-built project did not have that setting key. As a result I separated the test settings (and subsequently - common settings) into 2 sets - a cross-project version and a non cross-project version.

Additionally, I had to upgrade some library versions to get their 1.0.0 artifacts.

One problem is currently outstanding: `scalatestplus-scalacheck` does not currently publish a scala.js 1.0.0 artifact. However I figured I'll submit these changes now for review, and update the pull request when the artifacts are available.


mpilquist:
Thanks so much for working on this! I added a commit which adjusts the ScalaTest/ScalaCheck versions. Unforutnately, we still need to wait for ScalaTest 3.3.0-SNAP1 to get published to Sonatype. Bill released it last week but something must have glitched in Sonatype publishing part b/c it's still not there. We'll also need to add environment variable support for building against Scala.js 0.6, like is done in the cats build (or scodec for a simpler example).

disblader:
No worries at all!

I added the env variable for scala.js, same as cats, to project.sbt. Is there anything else that needs to be done for this to be included into the publishing process etc?

PR #1809, opened 2020-03-06
Make .compile.foldMonoid chunk-aware for potential speedups

Billzabob:
Using `.combineAll` for the chunk allows for potential speedups. For example, this could make `foldMonoid` much more efficient for `Map`s since it can take advantage of the efficient `combineAll`: https://github.com/typelevel/cats/blob/8bd728339b5b1c6b1b2f40f3cc9ef97480e1f02b/kernel/src/main/scala/cats/kernel/instances/MapInstances.scala#L77-L87

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4395,7 +4395,7 @@ object Stream extends StreamLowPriority {
       * }}}
       */
     def foldSemigroup(implicit O: Semigroup[O]): G[Option[O]] =
-      fold(Option.empty[O])((acc, o) => acc.map(O.combine(_, o)).orElse(Some(o)))
+      foldChunks(Option.empty[O])(_ |+| _.combineAllOption)
 
     /**
       * Compiles this stream in to a value of the target effect type `F`,
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4380,7 +4380,7 @@ object Stream extends StreamLowPriority {
       * }}}
       */
     def foldMonoid(implicit O: Monoid[O]): G[O] =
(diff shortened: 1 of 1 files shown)


Billzabob:
The build seems to have failed from what I believe is an unrelated issue.

mpilquist:
Great idea. Can we use `combineAllOption` in `foldSemigroup` in a similar fashion?

Billzabob:
I'm not sure if this should actually be merged. I decided to try a really basic performance test and it actually seems to make things slower. Here's what I did:
```scala
object Main extends IOApp {

  val chunkSize = 256 * 16

  override def run(args: List[String]): IO[ExitCode] = {
    time(currentImpl).map(t => s"$t ms").flatMap(putStrLn) >> time(chunkedImpl).map(t => s"$t ms").flatMap(putStrLn).as(ExitCode.Success)
  }

  def time(io: IO[Unit])(implicit timer: Timer[IO]): IO[Long] = {
    val currentTime = timer.clock.monotonic(TimeUnit.MILLISECONDS)
    (currentTime, io, currentTime).mapN((a, _, b) => b - a)
  }

  val currentImpl = Stream.constant(Map("a" -> 1), chunkSize).take(100_000_000).covary[IO].compile.foldMonoid.void
  val chunkedImpl = Stream.constant(Map("a" -> 1), chunkSize).take(100_000_000).covary[IO].compile.foldChunks(Map.empty[String, Int])(_ |+| _.combineAll).void
}
```

No matter the chunk size I've tested it seems that this new implementation takes about 3 times as long. I'm not sure I understand why though...

mpilquist:
Probably boxing/unboxing that's avoided via `Chunk.foldLeft`?

SystemFw:
Yeah I had a suspicion that the benefits of Chunk folding would outweigh the combineAll, but definitely good to have it measured

PR #1817, opened 2020-03-19
Support for preemption-safe stream subscription

### LINKED ISSUE #1816: Reactive streams subscription cannot be handled uninterruptibly
Related to #1638, and also http4s/http4s#3261.

So the `StreamSubscriber` mechanism is quite clever, and the API correctly assumes that actually *registering* the subscriber with the `Publisher` is going to be something which is very use-case and API specific, and thus it isn't captured in the fs2 API whatsoever. That's good. Unfortunately, this paradigm opens up a tricky loophole.

For publishers where `cancel()` must be invoked to avoid leaking resources (which is most of them), it is impossible for users to *uninterruptably* register their subscription and register the finalizer for proper cleanup. As a simplified example:

```scala
StreamSubscriber[F, A] map { sub =>
  Stream.eval(F.delay(publisher.subscribe(sub)) >> sub.stream
}
```

This isn't an unusual paradigm. In fact, almost everyone who is *correctly* using `.stream` is going to do something very much like this. Unfortunately, it is wrong. If the fiber running the resulting `Stream` is canceled during the `>>` cancel boundary, it will leak the upstream stream since `cancel()` cannot be run. It is impossible for user-land code to resolve this issue, since we really need the `F[Unit]` subscription action to be part of the `bracket` which is *inside* of `stream`, but… we don't have access to it.

The solution here is to change `stream` to take an `F[Unit]` which manages the subscription. This keeps it generic so that users can control things, but fs2-reactive-streams itself can juggle the resource management and registration of the finalizer. This change can be done while preserving binary compatibility by overloading `.stream` and deprecating the zero-argument version.

Note that this doesn't *entirely* close the loopholes which surround handling these types of APIs, but it at least fixes the one that cannot be otherwise worked around in userland.

djspiewak:
Fixes #1816

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -228,7 +228,8 @@ lazy val mimaSettings = Seq(
     ProblemFilters.exclude[IncompatibleSignatureProblem]("*"),
     // .to(sink) syntax was removed in 1.0.2 and has been hidden in all 2.x releases behind private[fs2], hence it's safe to remove
     ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.to"),
-    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.to$extension")
+    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream.to$extension"),
+    ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.interop.reactivestreams.StreamSubscriber#FSM.stream")    // FSM is package private
   )
 )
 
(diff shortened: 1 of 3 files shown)


SystemFw:
MiMa says no :(

djspiewak:
It's spurious because the parent type is package-private. I'll add an exclusion rule for it.

djspiewak:
I am very confused about these build failures.

mpilquist:
Just pushed a fix -- needed to run `fmt` on root project.

PR #1832, opened 2020-04-03
Try to Fix #1691 Stream.merge non-termination for non-IO errors (solution for ZIO didn't work for EitherT...)

### LINKED ISSUE #1691: Stream.merge freezes forever on uncaught defect in ZIO
Reproduction:

```scala
import fs2.Stream
import zio.{Task, UIO}
import zio.interop.catz._
import zio.duration._

object App extends zio.App {
  override def run(args: List[String]) = {
    Stream
      .eval(Task.interrupt)
      .onFinalize(Task(println("L finalizer")))
      .mergeHaltBoth(Stream.eval(Task(println("R stream"))))
      .compile.drain
      .ensuring(UIO(println("fin")))
      .timeoutFail(new Exception("Timed out"))(2.seconds)
      .as(0).orDie
  }
}
```
scastie: https://scastie.scala-lang.org/xtrQQLpLQW2F9TNIEDzXXw

Expected:

```scala
R stream
L finalizer
fin
```

Actual:

```scala
R stream
L finalizer
Scastie: timed out after 30 seconds when running code
```

The hang probably happens on [this line](https://github.com/functional-streams-for-scala/fs2/blob/master/core/shared/src/main/scala/fs2/Stream.scala#L1826) in `merge`, where there is an assumption that `io.attempt.flatMap` will always execute. Maybe `bracketCase` could help?

neko-kai:


diesalbla:
@neko-kai It seems that this merging of branches  https://github.com/typelevel/fs2/pull/2004 passes the tests. 

However, in the `main` branch, if you keep the `Stream.scala` class unchanged, it still passes the new tests.

neko-kai:
@diesalbla Interesting, I'm not sure how that happens exactly, because `.attempt` in runStream must actually return to be able to signal interruption - and it won't for EitherT.left / ZIO.terminate / Fiber.cancel outcomes. EitherT behaved differently from ZIO for me as well, could you check if the original reproduction example in https://github.com/typelevel/fs2/issues/1691 with ZIO works too without changing Stream.scala?

djspiewak:
FWIW…

Part of what I've discovered while adjusting CE3's hierarchy to be more composable is that the existing CE2 hierarchy *fundamentally* forces types that aren't fundamentally `IO`-like to discard context in an astonishing number of cases, spanning across `bracket`, `async`, `join`, and more. This affects `EitherT`, `ZIO`, `BIO`, and… really almost everything. In other words, this particular ticket is the tip of a very, very large and very fundamental iceberg.

It's worth fixing, but ultimately I guarantee there are a lot more cases just like this that we simply haven't discovered yet. CE3 solves all of that; the issues cannot be resolved on CE2 without breaking bincompat.

mpilquist:

PR #1840, opened 2020-04-10
Fix relative document paths

ssanj:
Fix relative documentation paths by using a `.html` extension as opposed to `.md` one.

The relative `.md` extensions could work as outlined [here](https://github.com/functional-streams-for-scala/fs2/issues/1836) but that would involve updating a Jekyll instance with a new plugin which could get messy. 

I've based this fix on the Doobie [docs](https://tpolecat.github.io/doobie/docs/18-FAQ.html) which use the same approach.


mpilquist:
Thanks!

Issue #1831, opened 2020-04-03
gzip compression is not streaming well


svalaskevicius:
two issues (same problem / file :) )

1. `gzip` does not allow to pass SYNC_FLUSH to deflater

2. `gunzip` delays chunks by at least one in `.through(_gunzip_validateTrailer(contentCrc32, inflater))`  - which is particularly bad when inflating a real time stream (e.g. a streaming http response / long polling)

mpilquist:
Thanks for the report. Can you clarify if this is with `fs2.compress` or `fs2.compression`?

svalaskevicius:
SYNC_FLUSH is a problem for both, the 2nd problem is from the `fs2.compression`

mpilquist:
cc @greg-a-atkinson @djspiewak

djspiewak:
`SYNC_FLUSH` is probably an easy fix (and also totally sounds like something that we would have missed). I don't know enough about the `compression` implementation to speak to how easy it would be to fix the run-behind problem.

svalaskevicius:
Yeah, I guess an easy workaround would be to allow to opt out of the trailer validation - for infinite streams of gzipped content it's not of much use anyways

djspiewak:
> Yeah, I guess an easy workaround would be to allow to opt out of the trailer validation - for (hopefully) infinite streams of gzipped content it's not of much use anyways

Seems like something that should be configurable for exactly this reason.

greg-a-atkinson:
> two issues (same problem / file :) )
> 
> 1. `gzip` does not allow to pass SYNC_FLUSH to deflater
> 2. `gunzip` delays chunks by at least one in `.through(_gunzip_validateTrailer(contentCrc32, inflater))`  - which is particularly bad when inflating a real time stream (e.g. a streaming http response / long polling)

Hi, thanks for raising your query.

For the first point, try compression.gzip(deflateStrategy = Some(Deflater.SYNC_FLUSH)) for your use case? The deflate strategy defaults to Deflater.DEFLUAT_STRATEGY in the same way as the previous compress.gzip and Deflater itself.

[...]

PR #1842, opened 2020-04-10
Enhance fs2.compression to allow the configuration of the deflate flush mode.

greg-a-atkinson:
Enhance fs2.compression to allow the configuration of the deflate flush mode.

See: https://github.com/functional-streams-for-scala/fs2/issues/1831


greg-a-atkinson:
> Looks good! Thanks :)
> 
> should it be added to compress.scala too or is that one considered deprecated?

Yes, fs2.compress was deprecated.

mpilquist:
The addition of the additional parameter breaks binary compatibility. Mima should have complained but it didn't, which is b/c mima is only looking at 2.0.0. Once #1844 is merged, we should update this branch.

mpilquist:
Mima is correctly failing now:

```
[error] fs2-core: Failed binary compatibility check against co.fs2:fs2-core_2.12:2.3.0! Found 4 potential problems
6135[error]  * static method gzip(Int,scala.Option,scala.Option,scala.Option,scala.Option,scala.Option,cats.effect.Sync)scala.Function1 in class fs2.compression does not have a correspondent in current version
6136[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.compression.gzip")
6137[error]  * static method deflate(Int,Boolean,Int,Int,cats.effect.Sync)scala.Function1 in class fs2.compression does not have a correspondent in current version
6138[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.compression.deflate")
6139[error]  * method deflate(Int,Boolean,Int,Int,cats.effect.Sync)scala.Function1 in object fs2.compression does not have a correspondent in current version
6140[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.compression.deflate")
6141[error]  * method gzip(Int,scala.Option,scala.Option,scala.Option,scala.Option,scala.Option,cats.effect.Sync)scala.Function1 in object fs2.compression does not have a correspondent in current version
6142[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.compression.gzip")
6143[info] Done compiling.
6144[error] java.lang.RuntimeException: Failed binary compatibility check against co.fs2:fs2-core_2.12:2.3.0! Found 4 potential problems
```

greg-a-atkinson:
> The addition of the new parameter breaks binary compatibility. Mima should have complained but it didn't, which is b/c mima is only looking at 2.0.0. Once #1844 is merged, we should update this branch.

Thanks for the heads-up. I'll update the PR.

core/jvm/src/main/scala/fs2/compression.scala
```diff
@@ -229,6 +254,51 @@ object compression {
...
+    *                   performance degradation of roughly 50-75%. Default
+    *                   size is 32 KB.
+    * @param deflateLevel     level the compression level (0-9)
+    * @param deflateStrategy  strategy compression strategy -- see `java.util.zip.Deflater` for details
+    * @param modificationTime optional file modification time
+    * @param fileName         optional file name
+    * @param comment          optional file comment
+    * @param deflateFlushMode flush mode -- see `java.util.zip.Deflater` for details
+    */
+  def gzip[F[_]](
```
mpilquist, line 294: At the risk of overdoing it, instead of this overload, we could have a `GzipParameters` (like `TlsParameters`) class with nicely typed values (e.g., `DeflateLevel` enum instead of an `Int`) and good defaults. Then we'd be able to add params if necessary without breaking source or binary compatibility. Thoughts?

greg-a-atkinson: > At the risk of overdoing it, instead of this overload, we could have a `GzipParameters` (like `TlsParameters`) class with nicely typed values (e.g., `DeflateLevel` enum instead of an `Int`) and good defaults. Then we'd be able to add params if necessary without breaking source or binary compatibility. Thoughts?

No trouble, I'd be happy to do this. Parameter class - all good. Enumerations - I guess the underling java.util.zip.(Inflater|Deflater) implementations aren't changing so this should be fine. Enumerations would betters support the ability to swap out the underlying inflater/deflater implementations and I've been curious if a pure Scala implementation of these without their synchronization and JNI calls could compete.

greg-a-atkinson: Hi @mpilquist, is the updated implementation preferred for future compatibility? Are the enumerations and defaults satisfactory? Any further suggestions or thoughts? The addition of these parameter classes increases the public surface of this package object quite substantially, should it be broken up and the original four public methods aliased? Thanks.

greg-a-atkinson:
As an aside, StreamSpec appears to be a little flaky and I've seen it fail a couple of time. I'll have a look if I get a chance:

[error] Failed: Total 2478, Failed 1, Errors 0, Passed 2477, Pending 18
[error] Failed tests:
[error] 	fs2.StreamSpec
[error] (coreJVM / Test / test) sbt.TestsFailedException: Tests unsuccessful

https://travis-ci.org/github/functional-streams-for-scala/fs2/jobs/676742328

core/jvm/src/main/scala/fs2/compression.scala
```diff
@@ -10,9 +10,175 @@ import cats.effect.Sync
...
+  /**
+    * Deflate algorithm parameters for the selected implementation, as one of:
+    *   1. [[compression.JavaUtilZipDeflaterParams]] for [[java.util.zip.Deflater]]
+    */
+  sealed trait DeflateParams
+
+  /**
+    * Deflate algorithm parameters for the [[java.util.zip.Deflater]] implementation.
+    */
+  sealed trait JavaUtilZipDeflaterParams extends DeflateParams {
```
mpilquist, line 35: Is this `JavaUtilZipDeflaterParams` intended to account for a possible future where `deflate` is implemented without using `java.util.zip.Deflate` and hence may require different config? If so, I think I'd prefer to skip that and just merge this in to `DeflateParams`. Similar on inflate side.

Looks good overall though, what do you think?

greg-a-atkinson: Yes, I have looked at alternative implementations and found that they have different configuration. As one example, the java.util.zip implementation only supports some of the flush modes requested in #1831, in fact there are several things it doesn't support or supports differently (eg. compression levels), so it may not be possible to support alternative implementations with a single parameter set. The intention of the naming of these two Params classes is for the provided configuration to also determine the desired implementation, where one of multiple may be selected.

As I understand it, you would prefer that this not support multiple future implementations, but only one, where if it is changed it will need to support at least the same functionality? No problem, I'll reduce and refactor as suggested, and update the PR.

As an aside and explanation, and depending on the time I'll have to put into these late night experiments and musings, my gut feeling is that there may be a performance gain to be made here by replacing java.util.zip's buffering-, thread-safe and JNI-based implementation with a pure fs2-based implementation. This would work by directly streaming the deflated/inflated Huffman encoding/decodings without first copying and concatenating these into the fixed-size output buffer controlled by the bufferSize parameter, also avoiding synchronization and native calls. If this is possible it should increase performance by avoiding buffer-filling array copying and may have interesting memory characteristics if the original and immutable Huffman encoding/decoding elements are directly streamed.

greg-a-atkinson: @mpilquist - PR updated as suggested with simplified DeflateParams and InflateParams. Anything else? Thanks.

mpilquist:
This is great, thanks! A pure scala implementation would be great. Could maybe even not require an effect to do inflation/deflation.

PR #1855, opened 2020-04-24
Add Stream#chunkAll

### LINKED ISSUE #1834: Add chunkAll
A use case for this came up recently so I sketched this solution:

```scala
def chunkAll[F[_], O](s: Stream[F, O]): Stream[F, Chunk[O]] = {
    def loop(s: Stream[F, O], acc: Chunk.Queue[O]): Pull[F, Chunk[O], Unit] =
      s.pull.uncons.flatMap {
        case Some((hd, tl)) => loop(tl, acc :+ hd)
        case None => Pull.output1(acc.toChunk)
      }
    loop(s, Chunk.Queue.empty).stream
  }
```

I decided against opening a PR for a couple of reasons:
1) It's not safe on infinite streams. It wouldn't be our first operation with such an issue but maybe adding more such operations isn't a great idea.
2) It can be simulated with `chunkN(Int.MaxValue, true)` for all safe use cases.

Opening this issue to document the decision. I'm totally okay with closing this as won't fix, but would like some opinions.

mpilquist:
Fixes #1834

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -365,25 +365,25 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
     filterWithPrevious((o1, o2) => eq.neqv(f(o1), f(o2)))
 
   /**
-   * Collects all output chunks in to a single chunk and emits it at the end of the
-   * source stream. Note: if more than 2^32-1 elements are collected, this operation 
-   * will fail.
-   * 
+    * Collects all output chunks in to a single chunk and emits it at the end of the
+    * source stream. Note: if more than 2^32-1 elements are collected, this operation
+    * will fail.
+    *
     * @example {{{
     * scala> import cats.implicits._
     * scala> (Stream(1) ++ Stream(2, 3) ++ Stream(4, 5, 6)).chunkAll.toList
     * res0: List[Chunk[Int]] = List(Chunk(1, 2, 3, 4, 5, 6))
     * }}}
-   */
+    */
   def chunkAll: Stream[F, Chunk[O]] = {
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -364,6 +364,26 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
...
+    * @example {{{
+    * scala> import cats.implicits._
+    * scala> (Stream(1) ++ Stream(2, 3) ++ Stream(4, 5, 6)).chunkAll.toList
+    * res0: List[Chunk[Int]] = List(Chunk(1, 2, 3, 4, 5, 6))
+    * }}}
+    */
+  def chunkAll: Stream[F, Chunk[O]] = {
+    def loop(s: Stream[F, O], acc: Chunk.Queue[O]): Pull[F, Chunk[O], Unit] =
+      s.pull.uncons.flatMap {
+        case Some((hd, tl)) => loop(tl, acc :+ hd)
```
diesalbla, line 381: Two questions here: 

1. Would we gain something if we were to get the _chunks_ from the original `s.pull`, and iterate and append chunk by chunk, rather than each element? 
2. Would we benefit from having a mutable constructor of an immutable chunk, much in the same way as Scala collections have mutable builders for then-immutable structures (e.g. `ListBuffer` for `List`)?

mpilquist: Yep, this implementation does both already more or less. We're pulling full chunks from the source (via `uncons`, not `uncons1`) and we're appending those chunks to a `Chunk.Queue`, which preserves them as whole chunks during the building phase. Only when `toChunk` is called on that queue are they all concatenated in to a single chunk.

PR #1857, opened 2020-04-25
Initial skeleton for munit based tests

mpilquist:
Just getting started here but wanted to share. As part of this refactoring, we'd cover the following:
- move from ScalaTest to munit
- stick with ScalaCheck for property tests
- use `TestContext` in the shared tests instead of relying on running async tests on real execution contexts
- supplement the `TestContext` based tests with a small number of async tests that run on real contexts if necessary -- these can be JS/JVM if not defined as properties or JVM only if defined as properties
- small test source files - roughly one per category of tests from the old specs


core/shared/src/test/scala/fs2/Fs2Suite.scala
```diff
@@ -0,0 +1,14 @@
+package fs2
+
+import munit.ScalaCheckSuite
+import cats.effect.laws.util.TestContext
+
+abstract class Fs2Suite extends ScalaCheckSuite with TestPlatform with Generators {
+
+  override def scalaCheckTestParameters =
+    super.scalaCheckTestParameters
+      .withMinSuccessfulTests(if (isJVM) 25 else 5)
```
mpilquist, line 10: We'll likely be able to crank these numbers way up when using test context.

PR #1858, opened 2020-04-27
Add parZip

### LINKED ISSUE #936: Add yip/yipWith/yipAll - non-deterministic versions of zip/zipWith/zipAll
In the context of `PipeSpec`:
```scala
"zipWith termination" in {
  val never = Stream.eval(IO.async[Int](_ => ()))
  val s = Stream(1, 2, 3)
  // works:
  runLog(s.zipWith(s ++ never)(_ == _), timeout = 1 second) shouldBe Vector(true, true, true)
  // hangs:
  runLog((s ++ never).zipWith(s)(_ == _), timeout = 1 second) shouldBe Vector(true, true, true)
}
```

SystemFw:
Fixes #936.

This PR adds the elusive `yip` operator, under the more modern name of `parZip`, with the following semantics:
 - combines elements pairwise
 - executes effects concurrently
 - no side is allowed to get ahead by more than one element
 - terminates when either stream terminates
The key insight of the implementation is that `parZip` is a special case of `merge`, where the two sides wait on each other.

I've also added `parZipWith`, and I'm considering `parZipRight` and `parZipLeft`, though I would probably not add `parZipAll` since it requires changing the implementation and it's rare.

Testing has proved a bit challenging, I have added two alternatives for the same test (marked -1 and -2), I'd appreciate input on which one to keep

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2266,6 +2266,15 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
   ): Stream[F2, O2] =
     parJoin(Int.MaxValue)
 
+
+  def parZip[F2[x] >: F[x]: Concurrent, O2](that: Stream[F2, O2]): Stream[F2, (O, O2)] =
+    this zip that
+
+  def parZipWith[F2[x] >: F[x]: Concurrent, O2 >: O, O3, O4](
+      that: Stream[F2, O3]
+  )(f: (O2, O3) => O4): Stream[F2, O4] =
+    this.parZip(that).map(f.tupled)
+
   /** Like `interrupt` but resumes the stream when left branch goes to true. */
   def pauseWhen[F2[x] >: F[x]](
       pauseWhenTrue: Stream[F2, Boolean]
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2267,7 +2267,7 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
     parJoin(Int.MaxValue)
 
   def parZip[F2[x] >: F[x]: Concurrent, O2](that: Stream[F2, O2]): Stream[F2, (O, O2)] =
-    this.zip(that)
+    Stream.parZip(this, that)
 
   def parZipWith[F2[x] >: F[x]: Concurrent, O2 >: O, O3, O4](
       that: Stream[F2, O3]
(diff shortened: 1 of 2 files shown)


mpilquist:
Looks great. I find the `TestContext` stuff pretty hard to read but that's probably just a matter of me getting used to it.

SystemFw:
> I find the TestContext stuff pretty hard to read but that's probably just a matter of me getting used to it

yeah. I think the test itself is complex in its own right as well. Note that without TestContext is really hard to make the test reproducible without waiting for entire seconds and making it super slow.
Do you have any preference between tests -1 and -2 or are they equally ugly :P ?

The first one is pure, but it uses a _lot_ of fs2 stuff (Signal, discrete, changesBy) so I'm worried about test independence, a bug in any of those would affect this test as well. We do use `Ref` often in other tests, but that's significantly simpler than the things I've used in this one. The second one has some `var`s, and the assertion is (very very slightly) less precise, but it's self contained.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4690,6 +4723,62 @@ object Stream extends StreamLowPriority {
...
+          case None => Pull.done
+          case Some((leftValue, stream)) =>
+            Pull.eval {
+              Deferred[F, Unit].flatMap { awaitRight =>
+                state.modify {
+                  case Racing =>
+                    LeftFirst(leftValue, awaitRight) -> Pull.eval(awaitRight.get)
+                  case RightFirst(rightValue, awaitLeft) =>
+                    Racing -> { emit(leftValue, rightValue) >> Pull.eval(awaitLeft.complete(())) }
+                  case LeftFirst(_, _) => sys.error("fs2 parZip protocol broken by lhs. File a bug")
```
matthughes, line 4754: Would it be worth spitting out the left/right values in the error messages in these cases?

SystemFw: I don't know. Generally how would so in an error case, but this would not be an error, it would be a bug, since I'm pretty confident it cannot possibly happen, it only needs to be there for pattern matching completeness

SystemFw:
I'm going to add `parZipLeft` and `parZipRight`, and remove one the test alternatives.
I've decided to keep the one with `var`s, the other one uses `discrete`, `concurrently`, `changes` and `interruptAfter` and therefore I think it's just too tied to the rest of fs2 given that we have a viable alternative.

Daenyth:
I'd prefer the test that doesn't use `var`, as I think it provides a better realistic usage that way. I think it's more likely that the test might misuse impurity in a subtle way that invalidates the result, as opposed to the test using fs2 tools.

I think the chance of a false pass for the fs2-using test which doesn't result in any other tests in the codebase failing is pretty slim.

If it ever did break, then generally it should happen in a way we can TDD the bug out.

SystemFw:
I will go ahead and merge for now, we can discuss the text context usage again in a specific conversation later

PR #1859, opened 2020-05-02
adds zipWithIndex function on Chunk

DamienOReilly:
Adds zipWithIndex function on `Chunk`, similar to that of `Stream`.

e.g.
```
fs2.Chunk("The", "quick", "brown", "fox").zipWithIndex.toList

res0: List[(String, Int)] = List((The,0), (quick,1), (brown,2), (fox,3))
```

### DIFF
--- a/core/shared/src/test/scala/fs2/ChunkSpec.scala
+++ b/core/shared/src/test/scala/fs2/ChunkSpec.scala
@@ -200,8 +200,7 @@ class ChunkSpec extends Fs2Spec {
   }
 
   "zipWithIndex andThen toArray" in {
-    val chunk = fs2.Chunk("x", "y", "z")
-    assert(chunk.zipWithIndex.toArray === chunk.toArray.zipWithIndex)
+    forAll((chunk: Chunk[Int]) => assert(chunk.zipWithIndex.toArray === chunk.toArray.zipWithIndex))
   }
 
   "Boxed toArray - regression #1745" in {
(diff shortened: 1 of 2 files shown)


core/shared/src/test/scala/fs2/ChunkSpec.scala
```diff
@@ -199,6 +199,11 @@ class ChunkSpec extends Fs2Spec {
     assert(arr2 === Array(0, 0))
   }
 
+  "zipWithIndex andThen toArray" in {
+    val arr: Array[(Int, Int)] = Chunk(0, 0, 0).zipWithIndex.toArray
+    assert(arr === Array((0, 0), (0, 1), (0, 2)))
```
kubukoz, line 204: We could probably assert `chunk.zipWithIndex.toArray == chunk.toArray.zipWithIndex` and make it a property test :)

DamienOReilly: Good idea :-)

core/shared/src/test/scala/fs2/ChunkSpec.scala
```diff
@@ -199,6 +199,11 @@ class ChunkSpec extends Fs2Spec {
     assert(arr2 === Array(0, 0))
   }
 
+  "zipWithIndex andThen toArray" in {
+    val chunk = fs2.Chunk("x", "y", "z")
+    assert(chunk.zipWithIndex.toArray === chunk.toArray.zipWithIndex)
```
LukaJCB, line 204: Can you make this a property based test? :)

DamienOReilly: Oops I read above as "proper test", not property test 🤣

PR #1861, opened 2020-05-04
Add in Kafka4s under Projects using FS2

zmccoy:
This just adds another project that uses FS2 for Kafka in the README.


kubukoz:
Thanks ;)

PR #1870, opened 2020-05-18
Added collectWhile to Stream

Billzabob:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -499,6 +499,17 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
...
+  /**
+    * Like [[collect]] but terminates as soon as the partial function is undefined.
+    *
+    * @example {{{
+    * scala> Stream(Some(1), Some(2), Some(3), None, Some(4)).collectWhile { case Some(i) => i }.toList
+    * res0: List[Int] = List(1, 2, 3)
+    * }}}
+    */
+  def collectWhile[O2](pf: PartialFunction[O, O2]): Stream[F, O2] =
+    takeWhile(pf.isDefinedAt).collect(pf)
```
kubukoz, line 511: There's just one thing I don't like about this... if the `isDefinedAt` check is expensive, we'll do it twice. I think we could replace that `collect` with `map` and pass the partial function as a defined one (since we know it'll be defined for these inputs). So basically `map(pf).

Billzabob: Good point, I'll fix it

kubukoz: I just realised it will still probably do the `isDefinedAt` check in `pf.apply`, but I don't see how we can avoid it without implementing the whole thing with something more low level, like a pull. I'd say we can leave it as it is and improve if there are issues :)

PR #1872, opened 2020-05-19
Drop new keyword on case classes

janekdb:


janekdb:
Running the tests on my puny VM seems to crush it so hopefully tests run :)

mpilquist:
Thanks!

PR #1877, opened 2020-05-26
Fix #1873 - deadlock in TLSEngine

### LINKED ISSUE #1873: Deadlock in the TLSEngine
When multiple streams are concurrently reading/writing a `TLSSocket`, there is a reproducible deadlock in the `TLSEngine`. This is illustrated by the following [scastie](https://scastie.scala-lang.org/fcabestre/6TY4XKMeSoGCrEEozQPIsw/4). 

Let me quote a quick analysis from @mpilquist on the FS2 Gitter room: 
> Definitely a bug in TLSEngine -- it occurs when there's a read that blocks for data followed by a write and the write starts the TLS handshake. The read gets the first response data from the handshake and also tries to start handshaking separately, but gets blocked on the handshake semaphore.

mpilquist:
The key part of the fix is the guard in `TLSEngine.read` that checks if a session is active and if not, handshakes before reading. I think a similar deadlock can still occur if a mid-session renegotiation is requested while an application level read is pending. Though TLS1.3 has dropped support for renegotiation.

### DIFF
--- a/io/src/main/scala/fs2/io/tls/TLSEngine.scala
+++ b/io/src/main/scala/fs2/io/tls/TLSEngine.scala
@@ -113,7 +113,7 @@ private[tls] object TLSEngine {
                 }
               case None => Applicative[F].pure(None)
             },
-            write0(Chunk.empty, None) >> read0(maxBytes, timeout)
+            write(Chunk.empty, None) >> read0(maxBytes, timeout)
           )
 
       /** Performs an unwrap operation on the underlying engine. */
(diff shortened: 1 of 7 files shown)


io/src/main/scala/fs2/io/tls/TLSEngine.scala
```diff
@@ -56,51 +59,65 @@ private[tls] object TLSEngine {
...
+          .flatTap(result => log(s"wrap result: $result"))
           .flatMap { result =>
             result.getStatus match {
               case SSLEngineResult.Status.OK =>
-                doWrite(binding) >> {
+                doWrite(timeout) >> {
                   result.getHandshakeStatus match {
                     case SSLEngineResult.HandshakeStatus.NOT_HANDSHAKING =>
-                      Applicative[F].unit
+                      wrapBuffer.inputRemains.map(_ > 0).ifM(wrap(timeout), Applicative[F].unit)
```
gvolpe, line 79: Alternatively: `wrapBuffer.inputRemains.flatMap(x => wrap(timeout).whenA(x > 0))`

PR #1879, opened 2020-05-28
Omit redundant type-class constraint

diesalbla:
This would be binary-breaking, though

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1083,7 +1083,11 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
   /**
     * Alias for `evalMapChunk(o => f(o).as(o))`.
     */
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1084,7 +1084,7 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
   /**
     * Alias for `evalMapChunk(o => f(o).as(o))`.
     */
-  def evalTapChunk[F2[x] >: F[x]: Functor: Applicative](f: O => F2[_]): Stream[F2, O] =
+  def evalTapChunk[F2[x] >: F[x]: Applicative](f: O => F2[_]): Stream[F2, O] =
```
kubukoz, line 1087: We could work around the binary compatibility problem by adding an overload (with the old signature) and making it private[Stream]. This one would just proxy to it.

PR #1885, opened 2020-06-05
Fix scope leak with Stream.retry and in general, with pulls that do...

mpilquist:
…not consume all elements from source.

Some observations prior to this PR:

```scala
// No leak:
Stream.eval(IO.unit).attempt.repeat.compile.drain.unsafeRunSync()

// Leaks in 2.1.0+, does not leak in 2.0.1:
Stream.eval(IO.unit).attempts(Stream.constant(1.second)).head.repeat.compile.drain.unsafeRunSync()

// Leak fixed by inserting an explicit scope after head
Stream.eval(IO.unit).attempts(Stream.constant(1.second)).head.scope.repeat.compile.drain.unsafeRunSync()
```

The reason a leak was introduced in 2.1.0 is that `handleErrorWith` was modified to introduce a scope. The use of `head`, which is an alias for `take(1)`, is implemented with a pull that discards the remainder of the stream once a single output element has been output. This has the effect of discarding the release of the scope opened by `handleErrorWith`.

More generally, any pull which discards the remainder of the stream has the potential of causing a similar scope leak. Hence, this PR returns to the FS2 1.0.x design of introducing a scope every time a pull is converted to a stream. Doing so causes the scope introduced by `handleErrorWith` to be a child scope of the pull-to-stream scope, and hence, when the pull-to-stream scope is closed, all descendant scopes are closed too.

### DIFF
--- a/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
+++ b/core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
@@ -227,3 +227,18 @@ object ZipThenBindThenParJoin extends App {
     .drain
     .unsafeRunSync()
 }
+
+object RetrySanityTest extends App {
+  implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.Implicits.global)
+  implicit val timer: Timer[IO] = IO.timer(ExecutionContext.global)
+  import scala.concurrent.duration._
+
+  Stream.retry(IO.unit, 1.second, _ * 2, 10).repeat.compile.drain.unsafeRunSync()
+}
+
+object AttemptsThenPullSanityTest extends App {
+  implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.Implicits.global)
+  implicit val timer: Timer[IO] = IO.timer(ExecutionContext.global)
+  import scala.concurrent.duration._
+  Stream.eval(IO.unit).attempts(Stream.constant(1.second)).head.repeat.compile.drain.unsafeRunSync()
+}
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -36,12 +36,26 @@ final class Pull[+F[_], +O, +R] private[fs2] (private val free: FreeC[F, O, R]) 
     new Pull(free.map(r => Right(r)).handleErrorWith(t => Result.Pure(Left(t))))
 
   /**
-    * Interpret this `Pull` to produce a `Stream`.
+    * Interpret this `Pull` to produce a `Stream`, introducing a scope.
     *
     * May only be called on pulls which return a `Unit` result type. Use `p.void.stream` to explicitly
     * ignore the result type of the pull.
     */
   def stream(implicit ev: R <:< Unit): Stream[F, O] = {
+    val _ = ev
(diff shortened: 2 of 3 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1793,7 +1793,7 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
     * }}}
     */
   def map[O2](f: O => O2): Stream[F, O2] =
-    this.pull.echo.mapOutput(f).stream
+    this.pull.echo.mapOutput(f).streamNoScope
```
mpilquist, line 1796: Without this change, the test for `resourceWeak` fails b/c there's a `map` in the implementation of `resourceWeak`. It's okay to not have a scope here b/c this pull consumes the entire source stream. Note back in 1.0.5, we had this same definition of `map` and it was the only such invocation in the library. So while this does feel a bit ad-hoc, we also have the most mileage on it.

core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
```diff
@@ -227,3 +227,18 @@ object ZipThenBindThenParJoin extends App {
     .drain
     .unsafeRunSync()
 }
+
+object RetrySanityTest extends App {
```
gvolpe, line 231: Unrelated to this PR but I think this repetition (cs, timer, etc) could be extracted into a common `SanityApp` or smth, no?

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -36,12 +36,26 @@ final class Pull[+F[_], +O, +R] private[fs2] (private val free: FreeC[F, O, R])
...
 
   /**
-    * Interpret this `Pull` to produce a `Stream`.
+    * Interpret this `Pull` to produce a `Stream`, introducing a scope.
     *
     * May only be called on pulls which return a `Unit` result type. Use `p.void.stream` to explicitly
     * ignore the result type of the pull.
     */
   def stream(implicit ev: R <:< Unit): Stream[F, O] = {
+    val _ = ev
```
gvolpe, line 45: I guess we do this because of a compiler warning? Could we shut this warning in a better way? e.g., in Scala 2.13.2 we can use the `@nowarn` annotation and probably the silencer plugin in previous versions.

core/jvm/src/test/scala/fs2/MemorySanityChecks.scala
mpilquist: Yeah, these started as minimizations and hence each were as small as possible. IOApp has too much overhead due to the cancellation stuff. I’d like to see these all refactored in to an integration test suite that we can run periodically. Maybe run each test for a minute or something, checking heap growth.

core/shared/src/main/scala/fs2/Pull.scala
mpilquist: Yeah fatal-warnings and warn-unused. I’m wary of adding compiler plugins but if only for 2.12, that’s okay. Though we don’t have fatal-warnings enabled on 2.12 so probably not necessary. There’s a lot of these we can fix with the 2.13 warning override. Want to open an issue for that?

gvolpe: Will do that :+1:

PR #1886, opened 2020-06-07
Add stream and pull pic

Zelenya:
Added the same style as I did in cats-effect

09765C4B-6201-4738-9568-A565923690E2


Zelenya:
Thanks @SystemFw and @gvolpe for the code suggestions

mpilquist:
Site is updated -- thanks again, awesome work!

PR #1889, opened 2020-06-08
Port memory leak tests to integration tests, runnable via it:test

mpilquist:
Bad news: two tests are failing - `parJoin` and `signal continuous'.


SystemFw:
The silver lining is that this test strategy seems to be very useful :)

mpilquist:
Both leaks appear to have existed since 1.0.0 and didn't exist in 0.9.7 (this is definitely true for `parJoin` and I think it's true for the other, though I started focusing on `parJoin`).

I updated the PR to write a heap dump upon test failure. Analyzing the `parJoin` failure points to a leak in `PubSub.Strategy.Discrete` -- looks like the `outOfOrder` map is huge.

SystemFw:
btw I'm getting a bit worried about PubSub in general.
I appreciate the interface reusability that it gives, but we already know for sure that `Topic` has a contention issue that isn't immediately fixable with PubSub, and now this leak.

Hopefully Pavel can save the day :)
But if not, do you think it's something that needs addressing at the root, or should we just fix it as we go?

mpilquist:
I rolled `SignallingRef` back to its 0.9 implementation, which fixed both failing leak tests.

The discrete pubsub strategy definitely seems to have a leak where the `outOfOrder` map grows very very large, though I wasn't able to work out the exact case in which it happens. I find the pubsub stuff hard to reason about, but perhaps that's just unfamiliarity. I have no strong opinions on pubsub going forward, though I don't think we should be shy about writing custom concurrent data structures if we get stuff in exchange -- i.e., maximally reusing pubsub is not a goal.

SystemFw:
> i.e., maximally reusing pubsub is not a goal.

Right, cause we already know that for Topic is for sure a regression

PR #1893, opened 2020-06-09
Fix automated publishing of microsite

gvolpe:
Alright, this should fix it. I don't know why the $JAVA_HOME variable is set to the Zulu VM when I am specifically installing `openjdk11` but well...


gvolpe:
Please, squash the commits when merging :pray:

PR #1899, opened 2020-06-13
Collector support for ArraySeq

tmccarthy:
This PR adds a `Collector` for `scala.collection.immutable.ArraySeq` where a `ClassTag` is available.

Support for untagged `ArraySeq` already exists via `ArraySeq.untagged`, eg:

```scala
Stream.range(1, 5).compile.to(ArraySeq.untagged)
```

`ArraySeq`s created this way will fail to take advantage of the specialised implementations provided for arrays of primatives.

#### Why provide specialised support for `ArraySeq`?

The immutable `ArraySeq` is newly added in Scala 2.13. It is an immutable wrapper over a bare array with specialised implementations for primitives. As such it is well suited to the kinds of performance-sensitive uses for which fs2 is commonly used.

### DIFF
--- a/core/shared/src/main/scala-2.13/fs2/CollectorPlatform.scala
+++ b/core/shared/src/main/scala-2.13/fs2/CollectorPlatform.scala
@@ -1,6 +1,8 @@
 package fs2
 
+import scala.collection.immutable.ArraySeq
 import scala.collection.{Factory, IterableFactory, MapFactory}
+import scala.reflect.ClassTag
 
 private[fs2] trait CollectorPlatform { self: Collector.type =>
   implicit def supportsFactory[A, C[_], B](
@@ -13,6 +15,14 @@ private[fs2] trait CollectorPlatform { self: Collector.type =>
   implicit def supportsMapFactory[K, V, C[_, _]](f: MapFactory[C]): Collector.Aux[(K, V), C[K, V]] =
     make(Builder.fromMapFactory(f))
 
+  implicit def supportsTaggedArraySeq[A : ClassTag](a: ArraySeq.type): Collector.Aux[A, ArraySeq[A]] = {
+    val _ = a
+    make(implicitly[ClassTag[A]] match {
+      case ClassTag.Byte => Builder.byteArray.mapResult(ArraySeq.unsafeWrapArray).asInstanceOf[Builder[A, ArraySeq[A]]]
+      case _ => Builder.taggedArraySeq[A]
+    })
+  }
+
   private[fs2] trait BuilderPlatform { self: Collector.Builder.type =>
     def fromFactory[A, C[_], B](f: Factory[A, C[B]]): Builder[A, C[B]] =
       fromBuilder(f.newBuilder)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala-2.13/fs2/CollectorPlatform.scala
```diff
@@ -13,6 +15,20 @@ private[fs2] trait CollectorPlatform { self: Collector.type =>
   implicit def supportsMapFactory[K, V, C[_, _]](f: MapFactory[C]): Collector.Aux[(K, V), C[K, V]] =
     make(Builder.fromMapFactory(f))
 
+  /**
+    * Use `ArraySeq.untagged` to build a `Collector` where a `ClassTag` is not available.
+    */
```
tmccarthy, line 20: I'm open to feedback on whether there is a better place for this hint.

core/shared/src/main/scala-2.13/fs2/CollectorPlatform.scala
```diff
@@ -13,6 +15,20 @@ private[fs2] trait CollectorPlatform { self: Collector.type =>
...
+  implicit def supportsTaggedArraySeq[A: ClassTag](
+      a: ArraySeq.type
+  ): Collector.Aux[A, ArraySeq[A]] = {
+    val _ = a
+    make(implicitly[ClassTag[A]] match {
+      case ClassTag.Byte =>
+        Builder.byteArray.mapResult(ArraySeq.unsafeWrapArray).asInstanceOf[Builder[A, ArraySeq[A]]]
+      case _ => Builder.taggedArraySeq[A]
```
tmccarthy, line 28: This specialisation is basically copied from [`Collector.supportsArray`](https://github.com/functional-streams-for-scala/fs2/blob/v2.4.2/core/shared/src/main/scala/fs2/Collector.scala#L27).

tmccarthy:
The effect of this PR is that you can use collectors like `.to(ArraySeq)` if there is a `ClassTag` available. If none is available, the user must use `.to(ArraySeq.untagged)` to avoid a compilation error.

There is another way to do this. We can provide two implicit conversions from `ArraySeq.type` to `Collector`: 

* A high-priority conversion that requires a `ClassTag` and provides specialised `ArraySeq`s for primatives
* A low-priority conversion that does not require a `ClassTag` and does not provide specialised `ArraySeq`s

If we did it this way, users would be able to use `.to(ArraySeq)` whether or not a `ClassTag` is available, and the implicit resolution would specialise if possible. I'm happy to rework this PR to use that approach if it seems better.

mpilquist:
This looks great, thanks! Couple of quick comments:
- I think the current scheme makes sense as use cases where the parameter to `ArraySeq` is concrete result in the right behavior and cases where the type is unknown result in a compilation error, which encourages either addition of the class tag or usage of the untagged method.
- Any thought about adding conversions to/from `Chunk` as well? Perhaps in another PR?

tmccarthy:
Thanks @mpilquist!

With regards to `Chunk`, there is already support for efficient creation from `ArraySeq` instances. [`Chunk.iterable`](https://github.com/functional-streams-for-scala/fs2/blob/v2.4.2/core/shared/src/main/scala/fs2/Chunk.scala#L597) uses [`ArrayBackedSeq`](https://github.com/functional-streams-for-scala/fs2/blob/v2.4.2/core/shared/src/main/scala-2.13/fs2/internal/ArrayBackedSeq.scala) to efficiently wrap the underlying array. I would be happy to add a specialised `Chunk.arraySeq` conversion for clarity's sake, if you think that would be worthwhile.

I think it would be a good idea to add something like `chunk.toArraySeq` in another PR. I'm happy to do this too.

mpilquist:
Sounds good! `fromArraySeq` and `toArraySeq` also sound good.

PR #1900, opened 2020-06-15
Internals: merge files `FreeC` and `Algebra`.

diesalbla:
As part of the work towards the simplifed model of `fs3`, we unify the files and objects `FreeC` and `Algebra` into a single one. This allows us to re-seal the `FreeC` GADT, since it is in one file.

To reduce some collisions, we also do these smaller changes:

- For convenience of a smaller diff and preserving source history, we move lines from `FreeC.scala` to `Algebra.scala`, as there are fewer.
- Due to collision between `FreeC.Eval`, and `Algebra.Eval`, we rename the latter into `Lift`, in the sense of "lifting" an `F[A]` into a `Stream[F, A]`.
- To reduce changes in the `Stream` and `Pull` files, we keep `Algebra` as an internal alias of `FreeC`.

______________________

This Pull Request is related to https://github.com/functional-streams-for-scala/fs2/pull/1735 and others. However, the goal of this PR is to submit a change that should be binary-compatible.


mpilquist:
Looks good. I’d like to remove the type aliases though - I know the diff will get huge but I find aliases just obfuscate things. I recently was giving a walkthrough of FS2 internals and had forgotten about some of the existing optimizations which make Algebra a subtype of FreeC and got confused for a bit. The aliases will make that worse.

Also, I lean a bit to renaming the existing `Eval` to something else and keeping what’s called `Lift` in this PR named `Eval`, to keep parity with `Pull.eval` and `Stream.eval`.

diesalbla:
> Looks good. I’d like to remove the type aliases though - I know the diff will get huge but I find aliases just obfuscate things. 

I have removed the `val Algebra = FreeC` aliae, so now all access to the Algebra object now call to the `FreeC` object. 

> Also, I lean a bit to renaming the existing `Eval` to something else and keeping what’s called `Lift` in this PR named `Eval`, to keep parity with `Pull.eval` and `Stream.eval`.

Well, that is a bit of a hard problem :). At present, the `FreeC` GADT has a few strangely named. I went for `Action`, since it conveys the idea that a `FreeC` can be either a terminal (or `Result`), a single action in `F`, or a binding of a `FreeC` with a continuation. I would use `Step`, but that is used for the class that contains a (concurrent?) `FreeC`. 

If someone can think of a better name, that may be worth considering. This is not a major trouble, since the word is only used inside the `FreeC` object itself, so renaming is localised.

mpilquist:
`Action` sounds good to me. Sorry to nitpick but one more thing... now that `FreeC` and `Algebra` are one, how about we avoid using either name and go with a new name? Eventually, for FS2 3.0, that will just be `Pull` hopefully, but in the meantime, how about `Engine` or `Interpreter` or something like that? Basically avoid the whole free+algebra convention.

diesalbla:
> Sorry to nitpick but one more thing... now that FreeC and Algebra are one, how about we avoid using either name and go with a new name

Unless I am mistaken, that would break binary compatibility. Both `Pull` and `Stream` extend from `AnyVal` and wrap a `FreeC`. Thus, methods in the public API of `fs2`, which in Scala source take as parameters or return as result a `Pull` or a `Stream`; are compiled to Java Bytecode method that take or return a `FreeC`. Thus, the name `FreeC` is embedded into the binary API of `fs2`.

Only a full binary-incompatible change (like 3.0) would allow for that rename. The goal of this PR was to reduce the gap towards that, but in bin-compat way.

mpilquist:
Yep, you’re right. Those anyvals have really been a pain!

PR #1905, opened 2020-06-19
Chunk support for ArraySeq

tmccarthy:
This PR adds support for immutable and mutable `ArraySeq`s for `Chunk`. Conversions in both directions are provided. This PR follows on from #1899.

#### Constructing a `Chunk` from `ArraySeq`

The `Chunk.arraySeq` factory method constructs a `Chunk` by wrapping the underlying array. This reproduces the behaviour previously implemented in [`ArrayBackedSeq`](https://github.com/functional-streams-for-scala/fs2/blob/v2.4.2/core/shared/src/main/scala-2.13/fs2/internal/ArrayBackedSeq.scala).

```scala
import scala.collection

Chunk.arraySeq(mutable.ArraySeq(1, 2, 3))   // Create Chunk from mutable ArraySeq
Chunk.arraySeq(immutable.ArraySeq(1, 2, 3)) // Create Chunk from immutable ArraySeq
Chunk.iterable(immutable.ArraySeq(1, 2, 3)) // Delegates to the factory methods above
```

#### Constructing an `ArraySeq` from a `Chunk`

```scala
import scala.collection.immutable.ArraySeq

val chunk = Chunk(1, 2, 3)

chunk.toArraySeq         // Uses implicit ClassTag to specialise ArraySeq
chunk.toArraySeqUntagged // Creates untagged ArraySeq
```

#### Why provide specialised support for `ArraySeq`?

The immutable `ArraySeq` is newly added in Scala 2.13. It is an immutable wrapper over a bare array with specialised implementations for primitives. As such it is well suited to the kinds of performance-sensitive uses for which fs2 is commonly used.


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -594,12 +593,10 @@ object Chunk extends CollectorK[Chunk] {
...
   def iterable[O](i: collection.Iterable[O]): Chunk[O] =
     i match {
-      case ArrayBackedSeq(arr) =>
-        // arr is either a primitive array or a boxed array
-        // cast is safe b/c the array constructor will check for primitive vs boxed arrays
-        array(arr.asInstanceOf[Array[O]])
-      case v: Vector[O]                    => vector(v)
-      case b: collection.mutable.Buffer[O] => buffer(b)
+      case a: collection.mutable.ArraySeq[O]   => arraySeq(a)
+      case a: collection.immutable.ArraySeq[O] => arraySeq(a)
```
tmccarthy, line 597: I have replaced the previous implementation of this behaviour via `ArrayBackedSeq` (now deleted).

core/shared/src/main/scala-2.13/fs2/ChunkPlatform.scala
```diff
@@ -0,0 +1,41 @@
...
+import scala.collection.immutable.ArraySeq
+import scala.collection.{immutable, mutable}
+import scala.reflect.ClassTag
+
+trait ChunkPlatform[+O] { self: Chunk[O] =>
+
+  def toArraySeq[O2 >: O: ClassTag]: ArraySeq[O2] = {
+    val array: Array[O2] = new Array[O2](size)
+    copyToArray(array)
+    ArraySeq.unsafeWrapArray[O2](array)
```
tmccarthy, line 12: This seems to be the most efficient way to construct an `ArraySeq`, given it will often take advantage of `System.arraycopy`. I might be missing something, though, so I'm open to feedback if there are better ways to do this.

core/shared/src/main/scala-2.13/fs2/ChunkPlatform.scala
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  def toArraySeqUntagged: ArraySeq[O] =
+    self match {
+      case knownType: Chunk.KnownElementType[O] => toArraySeq[O](knownType.elementClassTag)
```
tmccarthy, line 17: We attempt to specialise the `ArraySeq` even if there is no `ClassTag` available.

core/shared/src/main/scala-2.13/fs2/ChunkPlatform.scala
```diff
@@ -0,0 +1,41 @@
...
+          i += 1
+        }
+        buf.result()
+    }
+
+}
+
+trait ChunkCompanionPlatform { self: Chunk.type =>
+
+  // The array casts below are safe because the array constructor will check for primative vs boxed arrays
```
tmccarthy, line 33: This comment replaces the now-deleted comment [here](https://github.com/functional-streams-for-scala/fs2/blob/v2.4.2/core/shared/src/main/scala/fs2/Chunk.scala#L599) in `Chunk`.

tmccarthy:
Removing the `ArrayBackedSeq` matcher has broken the build for `2.12`. I'll need to re-think the cross-version approach for the `.iterable` method 😕.

core/shared/src/main/scala-2.13/fs2/ChunkPlatform.scala
```diff
@@ -0,0 +1,45 @@
+package fs2
+
+import scala.collection.immutable.ArraySeq
+import scala.collection.immutable
+import scala.reflect.ClassTag
+
+trait ChunkPlatform[+O] { self: Chunk[O] =>
```
mpilquist, line 7: Can you make this trait and it’s companion `private[fs2]` both here and in 2.13 file? Otherwise looks ready for merge. Thanks!

tmccarthy:
Thanks @mpilquist! This one should be good to go once the build completes.

Issue #1909, opened 2020-06-19
Documentation for s.as[O2]


kamilkloch:
Hi all,

I just noticed a small inconsistency in `s.as[O2]` documentation, which reads:
`Alias for _.map(_ => o2)`. It is not true - `s.as(o2)` memoizes `o2` and `s.map` does not:

```scala
def rnd: Int = scala.util.Random.nextInt(3)
println(Stream(1, 2, 3).as(rnd).toList) // List(1, 1, 1)
println(Stream(1, 2, 3).map(_ => rnd).toList) // List(1, 0, 1)
```
Granted, it is a minor glitch and `as` should not be fed an impure function in the first place, but perhaps the doc should be changed to: Alias for `val o2Memoized = o2; _.map(_ => o2Memoized)`?

kubukoz:
Fixed in #1910

PR #1910, opened 2020-06-19
Update s.as[O2] documentation.

kamilkloch:
See #1909.


kubukoz:
Dzięki!

kamilkloch:
Managed to sneak a typo into a one-liner change...

kubukoz:
See, this is why we don't github at night

kamilkloch:
So true.

PR #1913, opened 2020-06-20
Changed existing sinks to emit Nothing

mpilquist:
This is a follow up to #1754, which changes all sinks from `Pipe[F, O, Unit]` to `Pipe[F, O, INothing]`. Before this PR, there was no guarantee as to how many values would be emitted. Some sinks emitted 0, some 1, and some multiple. This PR unifies all sinks to emit 0 elements.

This PR deprecates `Stream.eval_` and replaces it with `Stream.exec: F[Unit] => Stream[F, INothing]`. In cases where `Stream.eval_(fa)` was called with an `fa: F[A]` such that `A != Unit`, the new syntax is now either `Stream.exec(fa.void)` or `Stream.eval(fa).drain`. That is, we’re forcing folks to be explicit if they want to throw away the result of an `F[A]`.

This PR also introduces `foreach: Stream[F, O] => (O => F[Unit]) => Stream[F, INothing]`.

Finally, this PR makes it a compilation error to call `flatMap` or `>>` on a `Stream[F, Nothing]`. Hence, any cases where folks were flatmapping a `Stream[F, Unit]` and signatures have changed from `Stream[F, Unit]` to `Stream[F, Nothing]` will be caught at compilation time.


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3328,15 +3341,23 @@ object Stream extends StreamLowPriority {
...
     * @example {{{
     * scala> import cats.effect.IO
-    * scala> Stream.eval_(IO(println("Ran"))).covaryOutput[Int].compile.toVector.unsafeRunSync
+    * scala> Stream.evalAction(IO(println("Ran"))).covaryOutput[Int].compile.toVector.unsafeRunSync
     * res0: Vector[Int] = Vector()
     * }}}
     */
-  def eval_[F[_], A](fa: F[A]): Stream[F, INothing] =
-    new Stream(Eval(fa).map(_ => ()))
+  def evalAction[F[_]](action: F[Unit]): Stream[F, INothing] =
```
SystemFw, line 3359: do we need another name? I think that for a lot of code this is actually longer than doing `.drain` after, and that still achieves the underlying purpose of making it a bit clearer than things are getting discarded.

You could actually argue that `eval_` is more inline with that since the `_` is "weird" enough to signal something is happening, whereas `evalAction` is a very innocuous name. 

I could definitely picture a lot of `Stream.evalAction(fa).flatMap` in the will.

core/shared/src/main/scala/fs2/concurrent/Queue.scala
```diff
@@ -21,7 +21,7 @@ trait Enqueue[F[_], A] {
     * Enqueues each element of the input stream to this queue by
     * calling `enqueue1` on each element.
     */
-  def enqueue: Pipe[F, A, Unit] = _.evalMap(enqueue1)
+  def enqueue: Pipe[F, A, INothing] = _.foreach(enqueue1)
```
SystemFw, line 24: There is a fairly unambiguous "number of expected units here", so I don't see why we need to remove useful combinators by draining

core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -118,8 +118,8 @@ object Topic {
                   }
               }
 
-          def publish: Pipe[F, A, Unit] =
-            _.evalMap(publish1)
+          def publish: Pipe[F, A, INothing] =
```
SystemFw, line 121: Same as enqueue

Daenyth:
As someone with a lot of team members who are learning scala/fp/fs2 and have mixed skill levels - I don't think this is the needed change. I'm concerned that the `INothing` results lead people too easily toward locked streams, because the compiler really doesn't do a good job assisting in cases like `for` on a `Stream`.

I think a better change would be to have each "sink" style pipe document what it emits, and have regression tests for it. People rely on that behavior, and while this is binary compatible, it's not *semantically* compatible. It's possible people have code that relies on the number of elements emitted from some sink, and changing that sink to emit nothing instead is the the kind of thing that can break user code via a transitive dependency eviction. That kind of eviction is a lot more likely as well, due to scala-steward keeping all downstream-of-fs2 stuff updated. A user can easily pull in more than they mean to.

My proposal would be this:
- Sinks which have some numeric correspondence, emit that many elements (as with Fabio's point)
- Sinks which are "terminal" and have no clear completion status - emit one `()` element. This is unlikely to break downstream code, because `usedToEmitNothing.flatMap(f)` would have been pointless in that code, and because the sinks were typed as returning `Unit` elements, if they wanted to be sure it was empty, they can `.drain`.

I'd avoid return signatures of `INothing` except where it aids inference: things like `eval_`, `never`, `.drain`, and so on.

mpilquist:
I agree that the potential for breaking clients and the churn induced likely tip the scales towards closing this instead of merging it. If not for backwards compatibility, I lean the other way though. Some thoughts:

> As someone with a lot of team members who are learning scala/fp/fs2 and have mixed skill levels - I don't think this is the needed change. I'm concerned that the `INothing` results lead people too easily toward locked streams, because the compiler really doesn't do a good job assisting in cases like `for` on a `Stream`.

To be clear, the issue here is not `Nothing`/`INothing` as a return type but rather streams that emit no elements. We can be explicit about “this stream emits no elements” but specifying the output type as `Nothing`.

> I think a better change would be to have each "sink" style pipe document what it emits, and have regression tests for it. People rely on that behavior, and while this is binary compatible, it's not *semantically* compatible. It's possible people have code that relies on the number of elements emitted from some sink, and changing that sink to emit nothing instead is the the kind of thing that can break user code via a transitive dependency eviction. 

Folks should not be relying on the number of units emitted from a sink. Doing so is a bug, as our current sinks make absolutely zero claims as to the number of units that get emitted. At best, this is relying on undocumented behavior.

> My proposal would be this:
> 
> - Sinks which have some numeric correspondence, emit that many elements (as with Fabio's point)

Consider the difference of enqueuing individual elements vs enqueuing chunks of elements. Hypothetically, let’s say we were to optimize `Queue#enqueue` to work on chunks — would we then have to emit extra units to preserve the correspondence with the number of emitted elements?

> - Sinks which are "terminal" and have no clear completion status - emit one `()` element. This is unlikely to break downstream code, because `usedToEmitNothing.flatMap(f)` would have been pointless in that code, and because the sinks were typed as returning `Unit` elements, if they wanted to be sure it was empty, they can `.drain`.

As mentioned before, the motivation here seems to be avoiding empty streams instead of anything particular to sinks, just to support for-comprehensions. I don’t think the cost of trying to enforce this is worth the effort, especially given zero help from the compiler.

> I'd avoid return signatures of `INothing` except where it aids inference: things like `eval_`, `never`, `.drain`, and so on.

AFAIK, all uses of `INothing` in this PR help inference.

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: I have found method names that end in underscore, and are named almost as other methods, are a bad idea. They are easy to miss, since most typefaces make underscores barely visible).  Furthermore, in this specific case I would avoid using the word _eval_, or _evaluate_, for an operation (in stream) that returns no value. 

I would suggest using some short words like `do` or `act` to name this operation.

mpilquist: @diesalbla `Stream.exec`?

mpilquist:
Another option is introducing new methods that return `Nothing` - e.g., deprecate `Socket#writes` and introduce `Socket#writeAll`

core/shared/src/main/scala/fs2/Sink.scala
```diff
@@ -12,7 +12,7 @@ object Sink {
 
   /** Lifts a function `I => F[Unit]` to `Sink[F,I]`. */
   @deprecated("Use stream.evalMap(f) instead", "1.0.2")
```
matthughes, line 14: Shouldn't this deprecation be updated to say use foreach?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1333,6 +1333,19 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
...
+    * Like `evalMap` but discards the result of evaluation, resulting
+    * in a stream with no elements.
+    *
+    * @example {{{
+    * scala> import cats.effect.IO
+    * scala> Stream(1,2,3,4).foreach(i => IO(println(i))).compile.drain.unsafeRunSync
+    * res0: Unit = ()
+    * }}}
+    */
+  def foreach[F2[x] >: F[x]](f: O => F2[Unit]): Stream[F2, INothing] =
```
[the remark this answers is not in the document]
mpilquist: There's always `evalMap(f)` though, if folks want the units.

mpilquist:
Alright, I made this one less controversial by sticking with unit returns for the obvious unit per element cases. The sinks in `io` package are all changed to return `Nothing` though, as they were never emitting once per byte.

matthughes:
Regarding the map/flatMap argument, I don't suppose there are some type tricks you could use to make map/flatMap only visible on non-Nothing Streams?

matthughes:
In general, I'm in favor of this change though I'm not sure of whether it should go in 2.x or 3.0

For functions that return an empty Stream `Unit`s (broadcastTo, balanceTo, etc):
  * change existing method signatures in 2.x OR 
  * add transitional method signatures that return Nothing and change them in 3.0

For functions that return 0 or N `Unit`s to indicate number of effects processed (enqueue), I agree that we shouldn't throw that information out.  But as @mpilquist points out above, the information encoded by `Unit` is lossy and not at all intuitive.  If folks really want to use that information, perhaps these methods should return `Int` instead of `Unit`; while the current implementation would always return 1, it would at least support improvements in a backwards compatible way.  In addition, introduce drained versions of these functions for the (vast majority?) of folks who throw that information away. Or change the current version to return INothing and introduce tracked versions. 

Lastly, I do wish we could provide help to ensure users aren't flatMapping on empty streams but that seems out of scope of this change.

mpilquist:
Not sure how folks feel about making `flatMap`/`>>` take a `NotNothing` witness, but here's what it would look like: https://github.com/mpilquist/fs2/commit/f7f031cf97d7877d9b383928e1d4a1e53be66368

mpilquist:
OK I think this one is good for final review. I incorporated the `NotNothing` trick suggested by @matthughes, which should prevent any surprises when upgrading. This also targets just FS2 3.0 now, not 2.x.

PR #1917, opened 2020-06-21
Add information about FS2-based library

susuro:
I've recently published a CSV parser based on FS2 and I would be grateful for the possibility to add it to the list of libraries in FS2 ecosystem.


SystemFw:
Thanks!

PR #1935, opened 2020-07-02
Missing cats instances

oskin1:
`cats.Defer` and `cats.Alternative` instances for `Stream`.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4968,14 +4968,14 @@ object Stream extends StreamLowPriority {
   implicit def functionKInstance[F[_]]: F ~> Stream[F, *] =
     FunctionK.lift[F, Stream[F, *]](Stream.eval)
 
-  implicit def monoidKInstance[F[_]]: MonoidK[Stream[F, *]] =
-    new MonoidK[Stream[F, *]] {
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4990,3 +4993,16 @@ private[fs2] trait StreamLowPriority {
         }
     }
 }
+
+private[fs2] trait StreamLowPriority1 {
```
kubukoz, line 4997: I'm moderately sure I tried this already and it'll fail the laws :(

oskin1: @kubukoz, you mean there is no lawful `Alternative` instance for `Stream`?

SystemFw: See here https://github.com/functional-streams-for-scala/fs2/pull/1418

SystemFw: and discussion here https://github.com/functional-streams-for-scala/fs2/issues/1394

oskin1: I see, thanks!

oskin1:
I'm closing this PR, going to create new one only adding `Defer` instance.

PR #1936, opened 2020-07-02
`cats.Defer` instance for `Stream`.

oskin1:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4973,6 +4973,12 @@ object Stream extends StreamLowPriority {
       def empty[A]: Stream[F, A] = Stream.empty
       def combineK[A](x: Stream[F, A], y: Stream[F, A]): Stream[F, A] = x ++ y
     }
+
+  /** `Defer` instance for `Stream` */
+  implicit def deferInstance[F[_]]: Defer[Stream[F, *]] =
+    new Defer[Stream[F, *]] {
+      override def defer[A](fa: => Stream[F, A]): Stream[F, A] = Stream(()) >> fa
```
djspiewak, line 4980: Wouldn't it be a bit faster to do `Stream.empty ++ fa`?

oskin1: Makes sense, thanks!

mpilquist:
Thanks!

mpilquist:
BTW, we could do same for `Pull`

Issue #1643, opened 2019-09-30
[Request] Add NoneTerminatedInspectableQueue


Daenyth:
It looks like the InspectableQueue constructors offer most of the behaviors that the Queue constructors offer, but it doesn't offer `noneBounded` versions - I had a use case recently where this would have been convenient

diesalbla:
@Daenyth In a recent Pull Request #2316, the `InspectableQueue` trait was removed from fs2, in favour of the `Queue` implementations from `cats-effect`. Given that, is this still needed?

Daenyth:
I'm not familiar enough to say. I need to look at the types involved. I have wanted specifically a NoneTerminated queue, and the ability to inspect its size. If the CE queue can offer both features, then this could definitely be closed

kubukoz:
Any reason why we can't make all queues inspectable in cats-effect?

PR #1937, opened 2020-07-02
None terminated inspectable queue

bedux:
Releated to Issue #1643

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/Queue.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Queue.scala
@@ -1,9 +1,9 @@
 package fs2
 package concurrent
 
-import cats.{Applicative, Eq, Functor, Id}
 import cats.effect.{Concurrent, Sync}
 import cats.implicits._
+import cats.{Applicative, Eq, Functor, Id}
 import fs2.internal.{SizedQueue, Token}
 
 /** Provides the ability to enqueue elements to a `Queue`. */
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/concurrent/Queue.scala
```diff
@@ -454,6 +454,37 @@ trait InspectableQueue[F[_], A] extends Queue[F, A] {
   def getSize: F[Int]
 }
 
+/**
+  * Like [[InspectableQueue]], but allows allows signalling of no further enqueues by enqueueing `None`.
```
diesalbla, line 458:
```suggestion
  * Like [[InspectableQueue]], but allows signalling of no further enqueues by enqueueing `None`.
```

core/shared/src/main/scala/fs2/concurrent/Queue.scala
```diff
@@ -454,6 +454,37 @@ trait InspectableQueue[F[_], A] extends Queue[F, A] {
...
+  * Optimizes dequeue to minimum possible boxing.
+  */
+trait NoneTerminatedInspectableQueue[F[_], A] extends NoneTerminatedQueue[F, A] {
+
+  /**
+    * Returns the element which would be dequeued next,
```
diesalbla, line 464:
```suggestion
    * Gets the element that is next to be dequeued,
```

core/shared/src/main/scala/fs2/concurrent/Queue.scala
```diff
@@ -454,6 +454,37 @@ trait InspectableQueue[F[_], A] extends Queue[F, A] {
...
+    * Offsetting enqueues and de-queues may not result in refreshes.
+    *
+    * Finally, note that operations like `dequeue` are optimized to
+    * work on chunks when possible, which will result in faster
+    * decreases in size that one might expect.
+    * More granular updates can be achieved by calling `dequeue1`
+    * repeatedly, but this is less efficient than dequeueing in
+    * batches.
+    */
+  def size: Stream[F, Int]
```
diesalbla, line 482: Should we use a `Signal` for this?  _If I understand correctly_, the difference between a Signal and a Stream is that the latter has a returns a fixed number of elements, whereas the former can be sampled by the consumer downstreams and only get values at those points.

core/shared/src/main/scala/fs2/concurrent/Queue.scala
```diff
@@ -465,6 +496,24 @@ object InspectableQueue {
...
+        maxSize: Int
+    )(implicit F: Concurrent[F]): G[NoneTerminatedInspectableQueue[F, A]] =
+      forStrategyNoneTerminated[F, (Boolean, SizedQueue[A]), A](
+        PubSub.Strategy.closeDrainFirst(Queue.Strategy.boundedFifo(maxSize))
+      ) { case (_, h) => h.headOption } { case (_, h) => h.size }
+
+    /** Created an unbounded queue terminated by enqueueing `None`. All elements before `None`. */
+    def noneTerminated[F[_], A](implicit
+        F: Concurrent[F]
+    ): G[NoneTerminatedInspectableQueue[F, A]] =
```
diesalbla, line 510:
```suggestion
    def noneTerminated[F[_]: Concurrent, A]: G[NoneTerminatedInspectableQueue[F, A]] =
```

No explicit use of the `F` parameter.

core/shared/src/main/scala/fs2/concurrent/Queue.scala
```diff
@@ -572,6 +621,113 @@ object InspectableQueue {
...
+              case Right(_) => -1
+            }
+
+          override def dequeue1: F[Option[A]] =
+            pubSub.get(Right(1)).flatMap {
+              case Left(s) =>
+                Sync[F].raiseError(
+                  new Throwable(
+                    s"Inspectable `dequeue1` requires chunk of size 1 with `A` got Left($s)"
+                  )
```
diesalbla, line 689: Perhaps we should we use specialised subclasses of `Throwable` for the errors we use.

mpilquist:
What's the status on this one?

diesalbla:
Considering that we have removed the implementations of `Queue` from `fs2`, in favour of those of `cats-effect`, and that Fabio Labella recently introduced the `Channel` as an alternative, I think we can close this one was left behind.

PR #1938, opened 2020-07-03
Unify FreeC and Pull

mpilquist:
This builds on the work @diesalbla has been doing and merges `FreeC`/`Algebra` in to `Pull`. As a result, `Pull` is a sealed abstract class with private constructors in the companion and `Stream` is a wrapper on `Pull`. Further, I made `Stream` a regular wrapper of `Pull` instead of an `AnyVal` to assist with future binary compatibility concerns. We didn't see much benefit of that `AnyVal` as a result of `Stream` getting passed around in `Option` -- e.g., via `uncons`.

Note this breaks binary compatibility and hence is targeted for FS2 3. ~We can either create a new branch for FS2 3 development or we can use `main` branch for that, after another 2.x maintenance release. Once we get a branch created, I'll make a pass at removing stuff that's been deprecated for a while.~ I created the `develop` branch for FS2 3 development.


SystemFw:
I think we should use a dedicated branch, cause there might be several other 2.x releases to include new combinators, fix bugs etc

diesalbla:
Nice! The old draft PR from a while ago had got too many conflicts.

I was thinking if, as a first step and to reduce diff distance, we could make a PR on the current `main` branch, to turn `Stream` from a wrapper of `FreeC` into a wrapper of `Pull` (an empty wrapper of an empty wrapper).

mpilquist:
@diesalbla Check out just the diff to Stream.scala - that's basically all that's changed there. `FreeC` -> `Pull`, drop `AnyVal`, implicit conversions -> implicit classes.

PR #1943, opened 2020-07-05
Dotty source compatibility

mpilquist:
This PR makes a bunch of adjustments to the source of coreJVM project so that it cross compiles against 2.12, 2.13, and latest Dotty nightly build. Note Dotty 0.25.0-RC2 has a bug that prevents its use with FS2 (https://github.com/lampepfl/dotty/pull/9275).

In particular, Dotty is much better w.r.t. detecting soundness issues. I fixed as many as possible without casts but had to add some, in particular in `Pull` -- some of those seem to be Dotty limitations but I'm not certain about that. Some of the others are definitely us doing dodgy things.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -477,7 +477,7 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] { self =>
 
   override def equals(a: Any): Boolean =
     a match {
-      case c: Chunk[O] =>
+      case c: Chunk[_] =>
         size == c.size && {
           var i = 0
           var result = true
@@ -497,7 +497,7 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] { self =>
 object Chunk extends CollectorK[Chunk] with ChunkCompanionPlatform {
 
   /** Optional mix-in that provides the class tag of the element type in a chunk. */
-  trait KnownElementType[A] { self: Chunk[A] =>
+  trait KnownElementType[A] extends Chunk[A] {
     def elementClassTag: ClassTag[A]
   }
 
@@ -593,9 +593,9 @@ object Chunk extends CollectorK[Chunk] with ChunkCompanionPlatform {
   /** Creates a chunk from a `scala.collection.Iterable`. */
   def iterable[O](i: collection.Iterable[O]): Chunk[O] =
     platformIterable(i).getOrElse(i match {
-      case a: mutable.ArraySeq[O]          => arraySeq(a)
+      case a: mutable.ArraySeq[o]          => arraySeq[o](a).asInstanceOf[Chunk[O]]
       case v: Vector[O]                    => vector(v)
-      case b: collection.mutable.Buffer[O] => buffer(b)
+      case b: collection.mutable.Buffer[o] => buffer[o](b).asInstanceOf[Chunk[O]]
       case l: List[O] =>
         if (l.isEmpty) empty
         else if (l.tail.isEmpty) singleton(l.head)
@@ -851,14 +851,9 @@ object Chunk extends CollectorK[Chunk] with ChunkCompanionPlatform {
       val b = readOnly(buf)
       (b: JBuffer).position(offset)
       (b: JBuffer).limit(offset + size)
-      if (xs.isInstanceOf[Array[C]]) {
-        get(b, xs.asInstanceOf[Array[C]], start, size)
-        ()
-      } else {
-        val arr = new Array[C](size)
-        get(b, arr, 0, size)
-        arr.copyToArray(xs, start)
-      }
+      val arr = new Array[C](size)
+      get(b, arr, 0, size)
+      arr.copyToArray(xs, start)
     }
 
     protected def splitAtChunk_(n: Int): (A, A) = {
@@ -1756,7 +1751,7 @@ object Chunk extends CollectorK[Chunk] with ChunkCompanionPlatform {
 
     override def equals(that: Any): Boolean =
       that match {
-        case that: Queue[A] => size == that.size && chunks == that.chunks
+        case that: Queue[_] => size == that.size && chunks == that.chunks
         case _              => false
       }
 
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -830,7 +830,6 @@ object Pull extends PullLowPriority {
               }
 
             case alg: AlgEffect[F, r] =>
-              // Safe case but needed for Dotty
               translateAlgEffect(alg)
                 .transformWith(r =>
                   translateStep(view.next(r.asInstanceOf[Result[y]]), isMainLevel)
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -4,7 +4,6 @@ import scala.annotation.tailrec
 import scala.util.control.NonFatal
 
 import cats.{Eval => _, _}
-import cats.arrow.FunctionK
 import cats.effect._
 import cats.implicits._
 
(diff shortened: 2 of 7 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -683,59 +682,6 @@ final class Stream[+F[_], +O] private[fs2] (private val underlying: Pull[F, O, U
...
-    * scala> Stream(1, 2).append(Stream(3, 4)).debug(o => s"a: $o").toList
-    * a: 1
-    * a: 2
-    * a: 3
-    * a: 4
-    * res0: List[Int] = List(1, 2, 3, 4)
-    * }}}
-    */
-  def debug(
-      formatter: O => String = _.toString,
```
mpilquist, line 709: These had to move to invariant extension methods.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1090,13 +1036,13 @@ final class Stream[+F[_], +O] private[fs2] (private val underlying: Pull[F, O, U
     * Not as powerful as `observe` since not all pipes can be represented by `O => F[_]`, but much faster.
     * Alias for `evalMap(o => f(o).as(o))`.
     */
-  def evalTap[F2[x] >: F[x]: Functor](f: O => F2[_]): Stream[F2, O] =
```
mpilquist, line 1093: Without this change, Dotty complains:

```scala
[error] 1040 |    evalMap(o => f(o).as(o))
[error]      |                 ^^^^^^^
[error]      |value as is not a member of F2[?]
[error]      |
[error]      |where:    F2 is a type in method evalTap with bounds >: F and <: [x] =>> Any
```

Rewriting as `evalMap(o => Functor[F2].as(f(o), o))` results in:

```scala
[error] 1040 |    evalMap(o => Functor[F2].as(f(o), o))
[error]      |                                ^^^^
[error]      |Found:    F2[?]
[error]      |Required: F2[Any]
[error]      |
[error]      |where:    F2 is a type in method evalTap with bounds >: F and <: [x] =>> Any
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1496,7 +1442,7 @@ final class Stream[+F[_], +O] private[fs2] (private val underlying: Pull[F, O, U
                     emitNonEmpty(acc) ++ startTimeout.flatMap { newTimeout =>
                       go(Chunk.Queue.empty, newTimeout)
                     }
-                  case Left(t) if t != currentTimeout => go(acc, currentTimeout)
```
mpilquist, line 1499: With this guard, we get non-exhaustive pattern match warnings. Similar for all removed guards.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3867,9 +3719,159 @@ object Stream extends StreamLowPriority {
...
+      * Finalizers on the outer stream are run after all inner streams have been pulled
+      * from the outer stream but not before all inner streams terminate -- hence finalizers on the outer stream will run
+      * AFTER the LAST finalizer on the very last inner stream.
+      *
+      * Finalizers on the returned stream are run after the outer stream has finished
+      * and all open inner streams have finished.
+      *
+      * @param maxOpen    Maximum number of open inner streams at any time. Must be > 0.
+      */
+    def parJoin(
```
mpilquist, line 3761: This had to move to an invariant `NestedStreamOps` due to Dotty complaining about the type signature of old `parJoin`:

```scala
[error] -- [E043] Type Error: /Users/mpilquist/Development/oss/fs2/core/shared/src/main/scala/fs2/Stream.scala:1759:44
[error] 1759 |  )(implicit ev: O <:< Stream[F2, O2], ev2: F[_] <:< F2[_], F2: Concurrent[F2]): Stream[F2, O2] =  ???
[error]      |                                            ^^^^
[error]      |unreducible application of higher-kinded type F to wildcard arguments
[error] -- [E043] Type Error: /Users/mpilquist/Development/oss/fs2/core/shared/src/main/scala/fs2/Stream.scala:1759:53
[error] 1759 |  )(implicit ev: O <:< Stream[F2, O2], ev2: F[_] <:< F2[_], F2: Concurrent[F2]): Stream[F2, O2] =  ???
[error]      |                                                     ^^^^^
[error]      |unreducible application of higher-kinded type F2 to wildcard arguments
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3606,10 +3400,15 @@ object Stream extends StreamLowPriority {
     */
   def resourceWeak[F[_], O](r: Resource[F, O]): Stream[F, O] =
     r match {
-      case r: Resource.Allocate[F, O] =>
-        Stream.bracketCaseWeak(r.resource) { case ((_, release), e) => release(e) }.map(_._1)
-      case r: Resource.Bind[F, x, O] => resourceWeak(r.source).flatMap(o => resourceWeak(r.fs(o)))
-      case r: Resource.Suspend[F, O] => Stream.eval(r.resource).flatMap(resourceWeak)
+      case r: Resource.Allocate[f, o] =>
```
mpilquist, line 3403: This is a house of cards. The old version fails to compile on Dotty due to variance issues. Fixing those in a straightforward way via extractors breaks compilation on 2.12. This is cleanest version I could get compiling on all 3 versions.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4885,7 +4887,9 @@ object Stream extends StreamLowPriority {
     * }}}
     */
   implicit def functionKInstance[F[_]]: F ~> Stream[F, *] =
-    FunctionK.lift[F, Stream[F, *]](Stream.eval)
+    new (F ~> Stream[F, *]) {
```
mpilquist, line 4890: `lift` is a macro which was removed in the Scala 3 version of cats.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -337,7 +338,7 @@ object Pull extends PullLowPriority {
     }
   }
 
-  private abstract class Bind[F[_], O, X, R](val step: Pull[F, O, X]) extends Pull[F, O, R] {
+  private abstract class Bind[+F[_], +O, X, +R](val step: Pull[F, O, X]) extends Pull[F, O, R] {
```
mpilquist, line 341: With no changes to this encoding, Dotty complains about 8 uncheckable runtime type checks due to variance mismatches. I was able to remove all of those by changing variance but more casts were needed. I minimized the number of such casts but there might be room here for more improvements or perhaps lurking bugs in Dotty.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -477,7 +477,7 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] { self =>
 
   override def equals(a: Any): Boolean =
     a match {
-      case c: Chunk[O] =>
+      case c: Chunk[_] =>
```
mpilquist, line 480: This `O` param was unnecessary here - we don't use it and Dotty correctly complains about it being uncheckable at runtime.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -851,14 +851,9 @@ object Chunk extends CollectorK[Chunk] with ChunkCompanionPlatform {
...
       (b: JBuffer).limit(offset + size)
-      if (xs.isInstanceOf[Array[C]]) {
-        get(b, xs.asInstanceOf[Array[C]], start, size)
-        ()
-      } else {
-        val arr = new Array[C](size)
-        get(b, arr, 0, size)
-        arr.copyToArray(xs, start)
-      }
+      val arr = new Array[C](size)
```
mpilquist, line 854: This was just wrong / unsound:

```scala
scala> trait Chunk[+A] { def copyTo[A2 >: A](arr: Array[A2]): Unit }
trait Chunk

scala> class Wrapped[A](under: Array[A]) extends Chunk[A] {
     |   def copyTo[A2 >: A](arr: Array[A2]): Unit = {
     |     println(s"${arr.isInstanceOf[Array[A]]} / ${arr.asInstanceOf[Array[A]]}")
     |   }
     | }
class Wrapped

scala> val w = new Wrapped(new Array[Int](10))
val w: Wrapped[Int] = Wrapped@3db48e6

scala> w.copyTo(new Array[Int](10))
true / [I@f61f06e

scala> w.copyTo(new Array[Any](10))
true / [Ljava.lang.Object;@425a3af5
```

PR #1948, opened 2020-07-09
Add some optimizations to utf8 decoding

johnynek:
This does four things:

1. adds Chunk.Queue.startsWith so we can be a bit more precise when checking for utf8 byte order mark
2. be miserly in internal methods with allocations and use null instead of Option (avoid allocation of Some on a loop).
3. use a Builder to avoid having to return a tuple and to reverse a list in an internal method.
4. avoid fold on an internal loop and just write out the while loop.

### DIFF
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -60,12 +60,12 @@ object text {
       val splitAt = allBytes.size - lastIncompleteBytes(allBytes)
 
       if (splitAt == allBytes.size) {
-        bldr += new String(allBytes.toArray, utf8Charset)
+        bldr += new String(allBytes, utf8Charset)
         Chunk.empty
       } else if (splitAt == 0)
         Chunk.bytes(allBytes)
       else {
-        bldr += new String(allBytes.take(splitAt).toArray, utf8Charset)
+        bldr += new String(allBytes.take(splitAt), utf8Charset)
         Chunk.bytes(allBytes.drop(splitAt))
       }
     }
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -38,17 +38,46 @@ object text {
      * 0 is returned.
      */
     def lastIncompleteBytes(bs: Array[Byte]): Int = {
-      val lastThree = bs.drop(0.max(bs.size - 3)).toArray.reverseIterator
-      lastThree
-        .map(continuationBytes)
-        .zipWithIndex
-        .find {
-          case (c, _) => c >= 0
+      /*
+       * This is logically the same as this
+       * code, but written in a low level way
+       * to avoid any allocations and just do array
+       * access
+       *
+       *
+       *
+        val lastThree = bs.drop(0.max(bs.size - 3)).toArray.reverseIterator
+        lastThree
+          .map(continuationBytes)
+          .zipWithIndex
+          .find {
+            case (c, _) => c >= 0
+          }
+          .map {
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/fs2/text.scala
```diff
@@ -35,74 +38,119 @@ object text {
...
+      var idx = bs.size - 1
+      var counter = 0
+      var res = 0
+      while (minIdx <= idx) {
+        val c = continuationBytes(bs(idx))
+        if (c >= 0) {
+          if (c == counter)
+            res = 0
+          else
+            res = counter + 1
```
non, line 73: I'm pretty sure `res` is already `0` so i think you just want `if (c != counter) res = counter + 1` don't you?

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -35,74 +38,119 @@ object text {
...
+          // exit the loop
+          idx = 0
```
non, line 75: If `bs.size` is `1` then what would happen? It seems like `minIdx` would be set to `0` (`0.max(-2)`) and `idx` would also be set to `0` (`1-1`). In that case, `idx = 0` won't break you out of the loop. I think `idx = -1` is a safer way to do this (although I would just use `return` myself).

Oh I see, you're counting on the decrement operator that occurs later. That makes sense, although IMO using `-1` or `return` here might be a bit more reassuring to the reader. (That said, there's no bug so maybe it's not a big deal.)

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -35,74 +38,119 @@ object text {
...
-      if (splitAt == allBytes.size)
-        (new String(allBytes.toArray, utf8Charset) :: output, Chunk.empty)
-      else if (splitAt == 0)
-        (output, Chunk.bytes(allBytes))
-      else
-        (
-          new String(allBytes.take(splitAt).toArray, utf8Charset) :: output,
-          Chunk.bytes(allBytes.drop(splitAt))
-        )
+      if (splitAt == allBytes.size) {
```
non, line 95: It's a small nit but I'd use `length` instead of `size` when working with arrays:

```scala
scala> val arr = Array(1,2,3)
val arr: Array[Int] = Array(1, 2, 3)

scala> u.reify { arr.size }
val res1: reflect.runtime.universe.Expr[Int] = Expr[Int](Predef.intArrayOps(arr).size)

scala> u.reify { arr.length }
val res2: reflect.runtime.universe.Expr[Int] = Expr[Int](arr.length)
```

johnynek:
```
Benchmark results:

on main:
[info] Benchmark                  (asciiStringSize)   Mode  Cnt       Score       Error  Units
[info] TextBenchmark.asciiDecode                128  thrpt    6  195334.730 ± 18599.880  ops/s
[info] TextBenchmark.asciiDecode               1024  thrpt    6  153746.264 ± 61576.197  ops/s
[info] TextBenchmark.asciiDecode               4096  thrpt    6  118761.218 ±  4695.784  ops/s

on this branch:
[info] Benchmark                  (asciiStringSize)   Mode  Cnt       Score      Error  Units
[info] TextBenchmark.asciiDecode                128  thrpt    6  218236.471 ± 1145.261  ops/s
[info] TextBenchmark.asciiDecode               1024  thrpt    6  192552.031 ± 1689.683  ops/s
[info] TextBenchmark.asciiDecode               4096  thrpt    6  146657.477 ±  589.035  ops/s
```

So this is 11% to 25% faster depending on the size, for this simple benchmark.

PR #1947, opened 2020-07-09
Add some clarity to the documentation on pauseWhen

barryoneill:
I found the existing documentation very confusing - as did someone else https://gitter.im/functional-streams-for-scala/fs2?at=5dd6afa31bf5192e667c4129

PR #1949, opened 2020-07-09
Flip the description of pauseWhen

kubukoz:
Follow-up to #1947.


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2309,7 +2309,7 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
   )(f: (O2, O3) => O4): Stream[F2, O4] =
     this.parZip(that).map(f.tupled)
 
-  /** Pause this stream when `pauseWhenTrue` emits `False`, resuming when `True` is emitted. */
+  /** Pause this stream when `pauseWhenTrue` emits `true`, resuming when `false` is emitted. */
```
kubukoz, line 2312: Perhaps it's also worth mentioning that this stream will be interrupted if the `pauseWhenTrue` stream terminates.

barryoneill:
:facepalm: nice catch

PR #1953, opened 2020-07-12
Convert from ScalaTest to Munit + ScalaCheck

mpilquist:
This positions us better for Dotty support.

The magic is in `AsyncPropertySupport` which implements our own async property checking by sampling a `Gen` and then just traversing the samples with the property function.

We have to keep the dependency on ScalaTest for now for two reasons: (1) sbt-doctest (https://github.com/tkawachi/sbt-doctest/issues/134) (2) TestNG support for reactive streams TCK.

All test files were renamed from `XyzSpec` to `XyzSuite` to match MUnit naming conventions. The old `StreamSpec` was split up in to a bunch of different files. `StreamSuite` contains the basics, `StreamCombinatorsSuite` contains simple combinators that have a 1, 2, or a few test cases. Bigger combinators get their own files, like `StreamParJoinSuite`, `StreamZipSuite`, etc.

BTW, the `parZip` test that uses `TestContext` had to be made JVM only, as ticking the test context didn't seem to be working on Scala.js. Not sure what's different compared to the old tests.


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4374,15 +4376,15 @@ object Stream extends StreamLowPriority {
           s: Stream[Pure, O],
           init: () => B
       )(foldChunk: (B, Chunk[O]) => B, finalize: B => C): C =
-        finalize(Compiler.compile(s.covary[IO].free, init())(foldChunk).unsafeRunSync)
+        finalize(Compiler.compile(s.covary[SyncIO].free, init())(foldChunk).unsafeRunSync)
```
mpilquist, line 4379: These were bugs, which caused exceptions on Scala.js. I think we had some tests that weren't running before, only looked like they were running.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -260,12 +260,14 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
     * }}}
     */
   def buffer(n: Int): Stream[F, O] =
-    this.repeatPull {
-      _.unconsN(n, allowFewer = true).flatMap {
-        case Some((hd, tl)) => Pull.output(hd).as(Some(tl))
-        case None           => Pull.pure(None)
+    if (n <= 0) this
```
mpilquist, line 263: Bug - would hang when n <= 0

johnynek:
this is exciting. Munit is great!

reactive-streams/src/test/scala/fs2/interop/reactivestreams/PublisherToSubscriberSpec.scala
```diff
@@ -17,7 +17,8 @@ final class PublisherToSubscriberSpec extends Fs2Suite {
 
   object TestError extends Exception("BOOM")
 
-  test("should propagate errors downstream") {
+  test("should propagate errors downstream".ignore) {
```
mpilquist, line 20: This suite wasn't running at all before. This test hangs on unsafeRunSync. Ignoring it for now.

core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala
```diff
@@ -9,11 +9,11 @@ import java.nio.file.{Files, Path}
...
-import org.scalatest.funsuite.AnyFunSuite
+import munit.FunSuite
 
 import fs2.concurrent._
 
-class MemoryLeakSpec extends AnyFunSuite {
+class MemoryLeakSpec extends FunSuite {
 
   lazy protected implicit val ioContextShift: ContextShift[IO] =
     IO.contextShift(ExecutionContext.Implicits.global)
```
gvolpe, line 19: Unrelated to this PR (feel free to ignore): this can be `ExecutionContext.global.

core/shared/src/test/scala/fs2/Fs2Suite.scala
```diff
@@ -0,0 +1,87 @@
...
+    with Generators {
+
+  override def scalaCheckTestParameters =
+    super.scalaCheckTestParameters
+      .withMinSuccessfulTests(if (isJVM) 25 else 5)
+      .withWorkers(1)
+
+  override def munitFlakyOK = true
+
+  implicit val realExecutionContext: ExecutionContext =
```
gvolpe, line 23: Does this need to be `implicit`? Regardless, the assigned value can be `ExecutionContext.global` instead.

core/shared/src/test/scala/fs2/AsyncPropertySuite.scala
```diff
@@ -0,0 +1,146 @@
...
+  def forAllAsync[F[_], A](
+      f: A => F[Unit]
+  )(implicit arbA: Arbitrary[A], F: Effect[F], loc: Location): Future[Unit] = {
+    val seed = Seed.random()
+    samples(arbA.arbitrary)
+      .runA(seed)
+      .value
+      .traverse_(a => reportPropertyFailure(f(a), seed, a.toString))
+      .toIO
+      .unsafeToFuture
```
gvolpe, line 50: Curious: can these `forAllAsync` methods return `F[Unit]` or just `IO[Unit]`? Seems possible since we have the `munitValueTransform` for `IO ~> Future` but I might be missing something

core/shared/src/test/scala/fs2/Fs2Suite.scala
mpilquist: Nope, made it non explicit and assigned to `ExecutionContext.global`

core/shared/src/test/scala/fs2/AsyncPropertySuite.scala
mpilquist: Good catch! I changed these to return `F[Unit]` which means the constraint goes from `Effect[F]` to `ApplicativeError[F, Throwable]`. I added a `SyncIO` runner to `Fs2Suite` as well.

PR #1957, opened 2020-07-14
Optimize Chunk.traverse take 2

johnynek:
this is a followup to #1952 

In that, I used a binary tree to implement Chunk.traverse. This gives stack safety, but it doesn't make it faster. I observed later and noted in https://github.com/typelevel/cats/issues/3517 that nothing requires the tree to be binary.

So, here, I use a tree with 128 children, and in all the internal nodes, I use `map2Eval` which allows short-circuiting and laziness for applicative instances that can avoid evaluating one side in some cases.

This results in a significant speed-up over the foldRight based implementation:
```
[info] Benchmark                       (chunkCount)  (chunkSize)   Mode  Cnt        Score       Error  Units
[info] ChunkBenchmark.traverse                    1           16  thrpt    6  1694967.533 ± 28309.177  ops/s
[info] ChunkBenchmark.traverse                    1          256  thrpt    6    45790.706 ±  1118.671  ops/s
[info] ChunkBenchmark.traverse                    1         4096  thrpt    6    38639.014 ± 12449.586  ops/s
[info] ChunkBenchmark.traverse                    1       409600  thrpt    6    16266.069 ±   512.302  ops/s
[info] ChunkBenchmark.traverseViaFold             1           16  thrpt    6   830798.147 ± 95479.003  ops/s
[info] ChunkBenchmark.traverseViaFold             1          256  thrpt    6    40813.175 ±  1310.140  ops/s
```

note, fold failed with stack overflow for 4096 and 409600 so I couldn't compare them.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -627,7 +627,17 @@ object Chunk extends CollectorK[Chunk] with ChunkCompanionPlatform {
 
   /** Creates a chunk backed by a `Chain`. */
   def chain[O](c: Chain[O]): Chunk[O] =
-    seq(c.toList)
+    if (c.isEmpty) empty
+    else {
+      val itr = c.iterator
+      val head = itr.next
+      if (itr.hasNext) {
+        val bldr = collection.mutable.Buffer.newBuilder[O]
+        bldr += head
+        bldr ++= itr
+        buffer(bldr.result)
+      } else singleton(head)
+    }
 
   /**
     * Creates a chunk backed by a mutable buffer. The underlying buffer must not be modified after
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -627,7 +627,17 @@ object Chunk extends CollectorK[Chunk] with ChunkCompanionPlatform {
 
   /** Creates a chunk backed by a `Chain`. */
   def chain[O](c: Chain[O]): Chunk[O] =
-    seq(c.toList)
+    if (c.isEmpty) empty
```
johnynek, line 630: this is the generic `Iterable` implementation and seems strictly better than converting to `List` first, which is what we were doing before.

PR #1959, opened 2020-07-16
Remove ScalaTest dependency and more Dotty source compat

mpilquist:
Rather than waiting for sbt-doctest to implement support for munit, this PR just changes sbt-doctest to use ScalaCheck instead of ScalaTest.

This PR also makes a few additional syntax changes due to recent Dotty changes -- in particular, it seems lambdas without parens are no longer supported. E.g. `map { x: Int =>` is no longer allowed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,8 +46,7 @@ lazy val commonSettingsBase = Seq(
   scalacOptions ++= Seq(
     "-feature",
     "-deprecation",
-    "-language:implicitConversions",
-    "-language:higherKinds",
+    "-language:implicitConversions,higherKinds",
     "-Xfatal-warnings"
   ) ++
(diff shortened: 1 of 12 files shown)


Daenyth:
> E.g. map { x: Int => is no longer allowed.

Seriously? What an awful change :/

mpilquist:
`map { x =>` is fine but if you ascribe a type, it needs to be `map { (x: Int) =>`

djspiewak:
> map { x => is fine but if you ascribe a type, it needs to be map { (x: Int) =>

It's mostly to try to make the parsing less hideously complex. :-)

mpilquist:
Closing in favor of #1960 which is targeted to main branch.

PR #1973, opened 2020-07-28
Add Dotty support for FS2 2.x

mpilquist:
This PR backports Dotty support to FS2 2.x. Most of the changes are mechanical based on the develop branch, though a handful needed different workarounds. In particular:

- Instead of moving `debug` and `debugChunks` to `InvariantOps`, which would be a binary incompatible, we use `@uncheckedVariance`
- `parJoin` and `parJoinUnbounded` get their `F2[_] <:< F[_]` param changed to `F2[Any] <:< F[Any]` (whereas on develop branch, they get moved to `InvariantOps`)
- the base64 encoding/decoding operations were deprecated and replaced with a new `text.base64` object to work around Dotty overloading issues


core/jvm/src/test/scala/fs2/CompressionSuite.scala
```diff
@@ -228,8 +229,8 @@ class CompressionSuite extends Fs2Suite {
   test("deflate and inflate are reusable") {
     val bytesIn: Int = 1024 * 1024
     val chunkSize = 1024
-    val deflater = deflate[IO](bufferSize = chunkSize)
-    val inflater = inflate[IO](bufferSize = chunkSize)
+    val deflater = deflate[IO](DeflateParams(bufferSize = chunkSize))
```
mpilquist, line 232: The other overload is not callable from Dotty -- appears to be a bug with overloads & default parameters, but I haven't minimized it yet.

mpilquist:
I'm going to eagerly merge this one but feel free to provide review comments and I'll happily address.

johnynek:
Why did so many case classes lose final? Are all case classes final in dotty?

mpilquist:
@johnynek case objects are final in dotty -- compiler complains otherwise. Case classes can still be marked final though:

```scala
scala> final case object One
1 |final case object One
  |^^^^^
  |final modifier is redundant for objects
```

Issue #1982, opened 2020-08-09
Scope-shifting issue using peek


sebastianvoss:
I posted this also in Gitter and dicussed in `http4s` channel with @rossabaker but maybe it is easier to track/ follow up as an Github issue.

I have a program which reads data from JDBC (using `doobie`) and streams it via an `http4s` endpoint. When my SQL throws an error I would like to send a non-200 response.

To identify SQL errors, I have to start consuming the stream. Therefore I thought to use `peek` and check if the stream is returning at least one chunk. The problem with this approach is that the `doobie` stream gets finalized and I cannot consume the ResultSet anymore (I get `PSQLException: This ResultSet is closed`).

Here is a sample snippet (I tried to remove dependencies, hopefully it makes it easier to reproduce it):

```scala
object Test extends IOApp {
  override def run(args: List[String]): IO[ExitCode] = {
    val s = for {
      closed <- Stream.eval(Ref.of[IO, Boolean](false))
      stream <- Stream.eval(stream(result(closed)))
      _      <- Stream.eval(closed.get.flatMap(b => IO(println(b))))  // <- prints true
      _      <- stream.evalTap(s => IO(println(s))) // <- in real-world this then fails with ResultSet closed exception
    } yield ()

    s.compile.drain.as(ExitCode.Success)
  }

  private def result(closed: Ref[IO, Boolean]): Stream[IO, String] =
    Stream
      .emits[IO, String]((1 to 10).map(_.toString))
      .onFinalize(closed.set(true))

  private def stream(result: Stream[IO, String]): IO[Stream[IO, String]] =
    result.pull.peek
      .flatMap {
        case Some((_, s)) => Pull.output1(IO(Stream.emit("ok") ++ s))
        case None         => Pull.output1(IO(Stream.emit("error")))
      }
      .stream
      .compile
      .lastOrError
      .flatten
}
```

mpilquist:
The `stream` method as written in this example (intentionally) leaks a reference to the tail of the stream provided by `peek`. As a result of compiling `result`, all resources are freed at the completion of evaluation of the compiled `IO`. Here's an annotated version that attempts to clarify that:

```scala
  private def stream(result: Stream[IO, String]): IO[Stream[IO, String]] =
    result.pull.peek
      .flatMap {
        case Some((_, s)) => Pull.output1(IO(Stream.emit("ok") ++ s)) // <-- s is emitted here
        case None         => Pull.output1(IO(Stream.emit("error")))
[...]

PR #1995, opened 2020-08-17
Stream - parJoin: small rewrites for readability.

diesalbla:
- Extract functions for ending "fork" and queue operations
- Write chain of flatMap into a for comprehensions.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3843,43 +3843,44 @@ object Stream extends StreamLowPriority {
         // "block" and await until the `running` counter drops to zero.
         val awaitWhileRunning: F[Unit] = running.discrete.dropWhile(_ > 0).take(1).compile.drain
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3843,43 +3843,44 @@ object Stream extends StreamLowPriority {
...
-        // runs one inner stream,
-        // each stream is forked.
+        // signals that a running stream, either inner or ourer, finished with or without failure.
+        def endWithResult(result: Either[Throwable, Unit]): F[Unit] =
+          result match {
+            case Right(()) => decrementRunning
+            case Left(err) => stop(Some(err)) >> decrementRunning
+          }
+
+        def leaseOrFail(scope: Scope[F]): F[Scope.Lease[F]] =
```
mpilquist, line 3853: Thoughts on moving this directly to `Scope`? I just ran in to the need for the same thing in #1982.

PR #1996, opened 2020-08-17
Add Pull.extendScopeTo combinator

mpilquist:
A combinator to simplify #1982


ChristopherDavenport:
Yeah, this is necessary for a lot the the prebuffering patterns we want. Thank you.

PR #2000, opened 2020-08-21
Typelevel organisation.

diesalbla:
For dependencies too

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -54,7 +54,7 @@ lazy val contributors = Seq(
 )
 
 lazy val commonSettingsBase = Seq(
-  organization := "co.fs2",
+  organization := "org.typelevel",
   scalacOptions ++= Seq(
     "-feature",
     "-deprecation",


build.sbt
```diff
@@ -54,7 +54,7 @@ lazy val contributors = Seq(
 )
 
 lazy val commonSettingsBase = Seq(
-  organization := "co.fs2",
+  organization := "org.typelevel",
```
kubukoz, line 57: We'll need to add this to @scala-steward, otherwise people's repos will stop being updated. And to be honest I'm not a fan of the idea of changing it, especially since the old name was shorter.

mpilquist: I’m okay with this either way but definitely agree we need Scala Steward to be aware. Is that straightforward? I know poms support relocation but iirc it’s kind of a pain to set up.

djspiewak: Yeah I'm definitely not a fan of changing this. Remember it breaks eviction detection too. Netty did this, and even in a non-Scala, pre-Scala-steward era it still created classpath hell for years.

I don't think the group id is that big of a deal, especially not relative to the havoc it creates to change it.

kubukoz: It's quite straightforward to let Steward know, it's been done e.g. for kind projector. https://github.com/scala-steward-org/scala-steward/blob/master/modules/core/src/main/resources/group-migrations.conf

rossabaker: The package should always change when the group ID does, and that would force an fs2-3.0. Please do not change this in the 2.x stream!

diesalbla:
Dropping for now.

PR #2011, opened 2020-08-26
Handle Outcome.Canceled

mpilquist:


SystemFw (approved):
Let's get this fixed for now, and we can discuss if we're missing something in CE since variations of this type pop up left and right

core/shared/src/main/scala/fs2/internal/InterruptionOutcome.scala
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package fs2.internal
+
+private[fs2] sealed trait InterruptionOutcome
+
+private[fs2] object InterruptionOutcome {
+  case class Completed(token: Token) extends InterruptionOutcome
+  case class Errored(t: Throwable) extends InterruptionOutcome
+  case object Canceled extends InterruptionOutcome
```
djspiewak, line 29: o_O

Where exactly does this come from?

mpilquist: @djspiewak What do you mean?

djspiewak: Actually backing up a bit, I was initially concerned about the bare nature of `Completed`, but I realized that CE itself actually uses this shape internally in a few places.

```scala
type InterruptionOutcome = Outcome[Id, Throwable, Token]
```

Could use `Need` instead if `Id` makes you as nervous as it makes me. `PureConc` has something a bit like this internally. Obviously this is an internal type, so it doesn't matter that much either way, but I just wanted to think about it a bit to make sure there wasn't something important missing in CE3 w.r.t. this type of case.

mpilquist: Cool, using type alias worked out nicely.

Issue #1484, opened 2019-05-27
Handle UTF-8 BOM in utf8Decode


coreteam-co:
I ran into some issues processing a CSV file which contains an UTF-8 BOM (Byte order mark, see https://en.wikipedia.org/wiki/Byte_order_mark). The Unicode Standard permits the BOM in UTF-8 but does not require or recommend its use. Unfortunately Microsoft uses it for example when saving an Excel file in CSV format.

I solved it with:

```
val stream: Stream[F, Byte] = ???
stream
   .through(fs2.text.utf8Decode)
   .map(s => if(s.startsWith("\ufeff")) s.substring(1) else s)
```

Would it make sense to make `fs2.text.utf8Decode` able to handle UTF-8 byte order mark? If so I can look into preparing a pull request.

djspiewak:
Yes, this would most definitely make sense. I've been bitten by similar things as well.

mpilquist:
A PR would be great!

mpilquist:
@coreteam-co Are you working on a PR for this? If not, I can put one together but don't want to step on any work you have in progress. Thanks!

coreteam-co:
@mpilquist Hi Michael - sorry for getting silent but I didn't have to get to it yet. Feel free to jump in.

mpilquist:
Fixed by #1501

PR #2012, opened 2020-08-27
Fix intermittent test failures in TextSuite

### LINKED ISSUE #1950: TextSpec failure on CI build
```scala
 [info]   - 1 byte sequences *** FAILED *** (66 milliseconds)
[info]     GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation. (TextSpec.scala:78)
[info]       Falsified after 4 successful property evaluations.
[info]       Message: List() did not equal List("﻿")
[info]       Location: (TextSpec.scala:79)
[info]       Occurred when passed generated values (
[info]         "﻿"
[info]       )
[info]       Init Seed: 1594344922141
```

mpilquist:
Fixes #1950

### DIFF
--- a/core/shared/src/test/scala/fs2/TextSuite.scala
+++ b/core/shared/src/test/scala/fs2/TextSuite.scala
@@ -22,7 +22,7 @@
 package fs2
 
 import cats.implicits._
-import org.scalacheck.Gen
+import org.scalacheck.{Arbitrary, Gen}
 import org.scalacheck.Prop.forAll
 import scodec.bits._
 import scodec.bits.Bases.Alphabets.Base64Url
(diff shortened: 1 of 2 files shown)


core/shared/src/test/scala/fs2/TextSuite.scala
```diff
@@ -136,13 +139,13 @@ class TextSuite extends Fs2Suite {
     group("handles byte order mark") {
       val bom = Chunk[Byte](0xef.toByte, 0xbb.toByte, 0xbf.toByte)
       property("single chunk") {
-        forAll { (s: String) =>
+        forAll(genStringNoBom) { (s: String) =>
```
djspiewak, line 142: This seems like a very ironic diff.

mpilquist: @djspiewak Took me a minute to catch on.

BTW, I'm not sure FS2 should really strip the byte order mark -- if FS2 just passed it along, these tests would all have passed. Looking at https://github.com/typelevel/fs2/issues/1484, we never really discussed why removing it is right behavior. Here's a quote from wikipedia: "The standard also does not recommend removing a BOM when it is there, so that round-tripping between encodings does not lose information, and so that code that relies on it continues to work." (https://en.wikipedia.org/wiki/Byte_order_mark). Thoughts?

djspiewak: IMO if the conversion is to `String`, it's better to strip the BOM. `String` itself already internally establishes all of the invariants that are communicated via the BOM (codec and endianness), so it's just redundant information which messes up offsets invisibly. However, as long as things are in byte-form, the BOM needs to be preserved because we have no other way of signalling those things.

In other words, I *think* I believe that `utf8decode` is doing the right thing. A `Pipe[F, Byte, String]` should strip the BOM, while a `Pipe[F, Byte, Byte]` (or analogous) should not.

Issue #2010, opened 2020-08-25
Stream.fromIterator doesn't create proper chunks


susuro:
`Stream.fromIterator` fetches values one by one, effectively creating single-value chunks. This has negative impact on performance, especially when working with primitive values.
The situation is even worse for `Stream.fromBlockingIterator`, because context switch for each fetched value increases the performance penalty.

I think this could be improved with following implementation of `PartiallyAppliedFromIterator`:
```scala
private[fs2] final class PartiallyAppliedFromIterator[F[_]](
    private val dummy: Boolean
) extends AnyVal {
  def apply[A](iterator: Iterator[A], chunkSize: Int = 1)(implicit F: Sync[F]): Stream[F, A] = {
    def getNextChunk(i: Iterator[A]): F[Option[(Chunk[A], Iterator[A])]] =
      F.delay {
        for (_ <- 1 to chunkSize if i.hasNext) yield i.next()
      }.map { s =>
        if (s.isEmpty) None else Some((Chunk.seq(s), i))
      }

    Stream.unfoldChunkEval(iterator)(getNextChunk)
  }
}
```
Similarly with `PartiallyAppliedFromBlockingIterator`.

If you find this solution OK I would be happy to prepare and submit a PR.

mpilquist:
Any thoughts on `Stream.fromIterator(iterator).chunkN(n)` instead?

djspiewak:
> Any thoughts on Stream.fromIterator(iterator).chunkN(n) instead?

It's more general, but it would incur the chunk management overhead. It's going to be significantly slower than batching `next` calls within a tight loop. I'm a little skeptical of using `Iterator` in performance-sensitive code to begin with, but if you are, then `chunkN` isn't going to solve the problem.

susuro:
I haven't done proper performance testing, only a few tests within a CSV parsing library, so my claims below my not be fully correct. However, the processing time I've measured for `Stream.fromIterator(iterator).chunkN(chunkSize).flatMap(Stream.chunk)` is about 2x shorter that using simple `Stream.fromIterator(iterator)`, while the tight loop I proposed gives me 10x better result.
In case of `fromBlockingIterator`, `chunkN` doesn't provide any significant improvement while tight loop executes 100x faster.

You're probably right that we should avoid using iterator for performance-sensitive operations. But sometimes, in existing projects, the iterator may be the only interface you have available and even if the code isn't really performance-sensitive, making it more efficient can still be valuable.

[...]

PR #2013, opened 2020-08-28
Create chunks in Stream.fromIterator (#2010)

susuro:
Improves performance of `Stream.fromIterator` and `Stream.fromBlockingIterator` by creating chunks of requested size, as described in #2010.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3426,7 +3426,17 @@ object Stream extends StreamLowPriority {
   private[fs2] final class PartiallyAppliedFromIterator[F[_]](
       private val dummy: Boolean
   ) extends AnyVal {
-    def apply[A](iterator: Iterator[A], chunkSize: Int = 1)(implicit F: Sync[F]): Stream[F, A] = {
+
+    def apply[A](iterator: Iterator[A])(implicit F: Sync[F]): Stream[F, A] = {
+      def getNext(i: Iterator[A]): F[Option[(A, Iterator[A])]] =
+        F.delay(i.hasNext).flatMap { b =>
+          if (b) F.delay(i.next()).map(a => (a, i).some) else F.pure(None)
+        }
+
+      Stream.unfoldEval(iterator)(getNext)
+    }
+
+    def apply[A](iterator: Iterator[A], chunkSize: Int)(implicit F: Sync[F]): Stream[F, A] = {
       def getNextChunk(i: Iterator[A]): F[Option[(Chunk[A], Iterator[A])]] =
         F.delay {
           for (_ <- 1 to chunkSize if i.hasNext) yield i.next()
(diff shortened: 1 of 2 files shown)


mpilquist:
Looks like mima configuration is messed up: `[info] fs2-core: mimaPreviousArtifacts is empty, not analyzing binary compatibility.`

If we fix that so it correctly checks binary compatibility, the build will fail on this change. Can you try making it an overload instead?

mpilquist:
Mima check is failing with:

```scala
[error] fs2-core: Failed binary compatibility check against co.fs2:fs2-core_2.12:2.0.0! Found 2 potential problems (filtered 662)
[error]  * extension method apply$extension(Boolean,scala.concurrent.ExecutionContext,scala.collection.Iterator,cats.effect.Sync,cats.effect.ContextShift)fs2.internal.FreeC in object fs2.Stream#PartiallyAppliedFromBlockingIterator does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream#PartiallyAppliedFromBlockingIterator.apply$extension")
[error]  * extension method apply$extension(Boolean,scala.collection.Iterator,cats.effect.Sync)fs2.internal.FreeC in object fs2.Stream#PartiallyAppliedFromIterator does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream#PartiallyAppliedFromIterator.apply$extension")
[error] stack trace is suppressed; run last mimaReportBinaryIssues for the full output
```

susuro:
I have no clue why this has broken the binary compatibility and how to cope with it. Any idea?

susuro:
I don't see how this can really break the compatibility. I would suggest excluding it from compatibility checks:
``` sbt
ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream#PartiallyAppliedFromIterator.apply$extension"),
ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Stream#PartiallyAppliedFromBlockingIterator.apply$extension")
```

mpilquist:
Can you try compiling some code against 2.4.4 that uses fromIterator and then running that compiled code against the current head?

susuro:
I did two tests on Scala 2.13, calling `fromIterator` directly and through a library. Both, the application and the library were built against 2.4.4 and first run with this version on classpath, then with the version from head. Each execution was correct. The head version was built with `publishLocal` after slight modification of `build.sbt` to get the library for Scala 2.13 (only 0.26 was published originally).
To check if the methodology is correct, I built also a modified head version with default chunk size parameter instead of an overload. This time I got indeed an error `Exception in thread "main" java.lang.NoSuchMethodError: 'fs2.internal.FreeC fs2.Stream$PartiallyAppliedFromIterator$.apply$extension(boolean, scala.collection.Iterator, cats.effect.Sync)'`.
I am not sure if this proves the compatibility enough, though.

mpilquist:
OK cool, I guess it's a false report by mima. Let's add the exclusions.

susuro:
I see you have already added it. Thanks.

PR #2014, opened 2020-08-28
fix compression: buffer usage

svalaskevicius:
1. Fix buffer size limits - to allow higher than 128 *bytes*

2. Use a preallocated buffer so we don't allocate it on every new chunk

### DIFF
--- a/core/jvm/src/main/scala/fs2/compression.scala
+++ b/core/jvm/src/main/scala/fs2/compression.scala
@@ -243,10 +243,12 @@ object compression {
       deflateParams: DeflateParams,
       deflater: Deflater,
       crc32: Option[CRC32]
-  ): Pipe[F, Byte, Byte] = {
-    val deflatedBuffer = new Array[Byte](deflateParams.bufferSizeOrMinimum)
-    _deflate_stream(deflateParams, deflater, crc32, deflatedBuffer)(_).stream
-  }
+  ): Pipe[F, Byte, Byte] =
+    in =>
+      Stream.suspend {
+        val deflatedBuffer = new Array[Byte](deflateParams.bufferSizeOrMinimum)
+        _deflate_stream(deflateParams, deflater, crc32, deflatedBuffer)(in).stream
+      }
 
   private def _deflate_chunk[F[_]](
       deflateParams: DeflateParams,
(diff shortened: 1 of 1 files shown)


svalaskevicius:
An issue I noticed but not fixed here, is that when the chunks are split because of insufficient buffer when deflating, the later chunks are only shipped when there are new elements in the stream, any ideas?

core/jvm/src/main/scala/fs2/compression.scala
```diff
@@ -243,14 +243,17 @@ object compression {
       deflateParams: DeflateParams,
       deflater: Deflater,
       crc32: Option[CRC32]
-  ): Pipe[F, Byte, Byte] =
-    _deflate_stream(deflateParams, deflater, crc32)(_).stream
+  ): Pipe[F, Byte, Byte] = {
+    val deflatedBuffer = new Array[Byte](deflateParams.bufferSizeOrMinimum)
```
mpilquist, line 247: This change makes the returned pipe unsafe for reuse -- consider calling this method once but then using resulting pipe in 2 or more streams.

core/jvm/src/main/scala/fs2/compression.scala
```diff
@@ -1222,10 +1254,17 @@ object compression {
     1024 * 1024 // A limit is good practice. Actual limit will be max(chunk.size, soft limit). 1 MiB feels reasonable for a comment.
 
   private def asChunkBytes(values: Array[Byte]): Chunk[Byte] =
-    asChunkBytes(values, values.length)
-
-  private def asChunkBytes(values: Array[Byte], length: Int): Chunk[Byte] =
-    if (length > 0) Chunk.Bytes(values, 0, length) else Chunk.empty[Byte]
+    asChunkBytes(values, values.length, false)
```
mpilquist, line 1257: Is this always safe? Given the buffer sharing, it's not clear to me at first glance. Safe because only call site is from header processing, which doesn't share a buffer?

core/jvm/src/main/scala/fs2/compression.scala
```diff
@@ -377,31 +394,41 @@ object compression {
       crc32: Option[CRC32]
   )(implicit
       SyncF: Sync[F]
-  ): Pipe[F, Byte, Byte] =
+  ): Pipe[F, Byte, Byte] = {
+    val inflatedBuffer = new Array[Byte](inflateParams.bufferSizeOrMinimum)
```
mpilquist, line 398: Needs same `Stream.suspend` treatment.

mpilquist:
cc @greg-a-atkinson

core/jvm/src/main/scala/fs2/compression.scala
svalaskevicius: Yes, that was the reasoning. I guess could be split, renamed to be more dangerous sounding.., and in that place it would be obvious .. Or prefixed move vs copy

core/jvm/src/main/scala/fs2/compression.scala
svalaskevicius: Good catch, and suggestion, thanks

svalaskevicius:
@mpilquist PR updated :)

greg-a-atkinson:
> 1. Fix buffer size limits - to allow higher than 128 _bytes_
> 2. Use a preallocated buffer so we don't allocate it on every new chunk

@svalaskevicius, @mpilquist 

Thank you for these improvements, it's greatly appreciated. My apologies, I don't know how I missed the CC. The inversion of the min/mix constrain on the buffer size was careless. I hope this implementation is working better now. I look forward to FS3.

Much appreciated

Kind regards,
Greg

PR #2017, opened 2020-08-29
Concurrency without effect suspension

SystemFw:


SystemFw:
One of the memory leak tests has failed. I don't know if I want to address it now because the problem in cats-effect async still persists, and I need to go fix that. WDYT @mpilquist ?

mpilquist:
Let’s mark it flaky for now

SystemFw:
let me mark it as flaky before merge

PR #2035, opened 2020-09-15
Update to latest cats-effect snapshot

mpilquist:
Don't merge this before typelevel/cats-effect#1192 is merged.


mpilquist:
@SystemFw 1 rx streams interop test started failing when the return type of `Deferred#complete` was changed to `F[Boolean]`. Haven't figured out why though.

djspiewak:
> Haven't figured out why though.

Did they previously accidentally depend upon the exception-throwing behavior?

mpilquist:
Not that I’ve been able to determine. I’be already replaced all voids with failures in rx-stream project without success. I’ll extend that to the core project too and see how it goes.

mpilquist:
This is the test that fails and I haven't succeeded in getting it passing by adding `.ifM(F.unit, F.raiseError(..))`. Makes me wonder if there's some other change that is causing a leak.

djspiewak:
> This is the test that fails and I haven't succeeded in getting it passing by adding .ifM(F.unit, F.raiseError(..)). Makes me wonder if there's some other change that is causing a leak.

Possible. Concerning.

mpilquist:
Ha, forgot a link to the failing test: https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/tck/src/main/java/org/reactivestreams/tck/PublisherVerification.java#L997-L1031

mpilquist:
@djspiewak @vasilmkd The memory "leak" is due to the work stealing pool. Replacing it with `ExecuctionContext.global` fixed the test. Any ideas as to why and whether this is a real issue or just a test environment thing?

Workaround here:

```scala
diff --git a/reactive-streams/src/test/scala/fs2/interop/reactivestreams/StreamUnicastPublisherSpec.scala b/reactive-streams/src/test/scala/fs2/interop/reactivestreams/StreamUnicastPublisherSpec.scala
index a0896ba4..5848539a 100644
--- a/reactive-streams/src/test/scala/fs2/interop/reactivestreams/StreamUnicastPublisherSpec.scala
+++ b/reactive-streams/src/test/scala/fs2/interop/reactivestreams/StreamUnicastPublisherSpec.scala
@@ -24,7 +24,7 @@ package interop
 package reactivestreams
 
 import cats.effect._
-import cats.effect.unsafe.implicits.global
+// import cats.effect.unsafe.implicits.global
 import org.reactivestreams._
 import org.reactivestreams.tck.{PublisherVerification, TestEnvironment}
 import org.scalatestplus.testng._
@@ -50,6 +50,14 @@ final class StreamUnicastPublisherSpec
       if (n == java.lang.Long.MAX_VALUE) Stream.range(1, 20).repeat
       else Stream(1).repeat.scan(1)(_ + _).map(i => if (i > n) None else Some(i)).unNoneTerminate
 
+    import cats.effect.unsafe.IORuntime
+
+    implicit val ioRuntime: IORuntime = IORuntime(
+      scala.concurrent.ExecutionContext.Implicits.global,
+      IORuntime.createDefaultBlockingExecutionContext()._1,
+      IORuntime.createDefaultScheduler()._1,
+      () => ()
+    )
     s.covary[IO].toUnicastPublisher
   }

```

vasilmkd:
It's probably a real issue. Do I just run the tests without the latest commit to see it happen? Also, how does the leak manifest itself?

mpilquist:
Yeah:
1) `git checkout d5b4db1c145fd25d100517cdebc5631205004687`
2) `sbt ";project reactiveStreams;testOnly *Unicast*"`

vasilmkd:
@mpilquist I'm on it. Thanks.

vasilmkd:
Found the issue and I fixed it! I'll work some more on the final code though.

SystemFw (approved):

PR #2036, opened 2020-09-15
SSL: Add Application Protocol Selection and Access

ChristopherDavenport:
Adds SSL Application Protocol Support, both accessing the protocol and Setting it in the selector.

### DIFF
--- a/io/src/main/scala/fs2/io/tls/TLSEngine.scala
+++ b/io/src/main/scala/fs2/io/tls/TLSEngine.scala
@@ -39,6 +39,7 @@ import cats.syntax.all._
   */
 private[tls] trait TLSEngine[F[_]] {
   def beginHandshake: F[Unit]
+  def applicationProtocol: F[String]
   def session: F[SSLSession]
   def stopWrap: F[Unit]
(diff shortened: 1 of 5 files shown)


io/src/main/scala/fs2/io/tls/TLSEngine.scala
```diff
@@ -77,6 +77,7 @@ private[tls] object TLSEngine {
 
       def beginHandshake = Sync[F].delay(engine.beginHandshake())
       def session = Sync[F].delay(engine.getSession())
+      def getApplicationProtocol = Sync[F].delay(engine.getApplicationProtocol())
```
mpilquist, line 80: Need to add this to `TLSSocket` too

ChristopherDavenport: Should be there now, changed to be consistent with session.

mpilquist:
The mima errors about `DefaultTLSParameters` can all be suppressed but the ones on `TLSParameters` need to be addressed. I guess by keeping the old `apply` and removing all the default params and maybe making it private[fs2]?

Issue #2018, opened 2020-09-02
SignallingRef#access problems


durban:
The current (v2.4.4) implementation of the `access` method of `SignallingRef` seems to have some problems (as far as I can tell, since v2.4.0, in #1889):

- It calls `equals` in [this line](https://github.com/typelevel/fs2/blob/v2.4.4/core/shared/src/main/scala/fs2/concurrent/Signal.scala#L200). It seems at least unintuitive, that an operation which is basically a compare-and-swap would use value equality.
- Regardless of the previous issue, `access` doesn't seem to work. That is, in a simple case (no contention), it cannot change the value in the ref. (Maybe the `==` should be `!=` instead? It might work that way, although the first issue would remain.)

SystemFw:
yeah that sounds weird. I'll have a look later, thanks for reporting :)

mpilquist:
I have no recollection of changing the implementation of `access` as represented in that PR...

SystemFw:
I think you just restored the old implementation since PubSub had a leak, and given that `access` is pretty much never used, and especially not on Signal, it's likely it has always been broken...

mpilquist:
Ah, thanks, that makes sense.

SystemFw:
Haven't forgotten about this :)

There is some bigger cleanup I want to do on Signal in `develop`, after which we can backport an `access` fix to `main`

mpilquist:
@SystemFw Any more thoughts on this one?

SystemFw:
oh, I think I fixed it in #2041 , but the automatic closing didn't trigger

PR #2041, opened 2020-09-18
Signal

SystemFw:
This PR has become a bit sprawling, I'm hoping in some leniency 😄 . The general thread is `Signal`.

- Fix the hilarious bug in #2018 where `access` would never succeed.
- Fix a more subtle bug in `access`, which failed to update `discrete`.
- Document a general limitation of `access` (there is contention with `discrete`)
- General cleanup of `Signal`, no more tuples etc.
- Strengthen bound on `Token.apply` to make our usage completely airtight (one liner)
- Added some TestContext infra (which proved to be a non strictly necessary yak to shave, but it's one off the todo list)
- Add more effective tests for `pauseWhen`, the existing test tested nothing
- Change behaviour of `pauseWhen` to not evaluate anything when started in paused state
- Significantly less convoluted implementation of the two `pauseWhen` overloads


mpilquist:
Sweet!

PR #2047, opened 2020-09-22
Add helpers for creating streams asynchronously via Queue

bplommer:
Hopefully this reduces the learning curve around applying this pattern and improves readability, especially as the order of dequeueing and queueing is fairly counterintuitive.

Will add scaladocs and tests if this is approved in concept 🙂

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -672,7 +672,7 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
   def debounce[F2[x] >: F[x]](
       d: FiniteDuration
   )(implicit F: Concurrent[F2], timer: Timer[F2]): Stream[F2, O] =
-    Stream.eval(Queue.bounded[F2, Option[O]](1)).flatMap { queue =>
+    Stream.asyncNoneTerminated[F2, O](1) { queue =>
       Stream.eval(Ref.of[F2, Option[O]](None)).flatMap { ref =>
         val enqueueLatest: F2[Unit] =
           ref.modify(s => None -> s).flatMap {
@@ -688,12 +688,8 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
               case Some(_) => F.unit
             }
 
-        val in: Stream[F2, Unit] = chunks.evalMap(onChunk) ++
+        chunks.evalMap(onChunk) ++
           Stream.eval_(enqueueLatest >> queue.enqueue1(None))
-
-        val out: Stream[F2, O] = queue.dequeue.unNoneTerminate
-
-        out.concurrently(in)
       }
     }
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3150,6 +3143,26 @@ object Stream extends StreamLowPriority {
   /** Creates a pure stream that emits the supplied values. To convert to an effectful stream, use `covary`. */
   def apply[F[x] >: Pure[x], O](os: O*): Stream[F, O] = emits(os)
 
+  def async[F[x] >: Pure[x], O](maxWaiting: Int)(f: Enqueue[F, O] => Stream[F, Unit])(implicit
```
Daenyth, line 3146: I think if we want these to reduce the learning curve, we're going to need to add some javadoc to them, and some usage examples somewhere

SystemFw:
I'm slightly unsure about these, although I agree with the basic thinking behind this PR.

A few random thoughts:
- why are we abstracting the write side rather than the read like (like we do in `hold`)? Unfortunately they both seem useful
- I kinda don't want all the variants, feels like going in this direction makes the api explode even though most are never used. I feel like this about Queues in fs2 in general.
- We cannot use the name `async` imo, too prominent in cats-effect for something completely different (although  it does fit in a vacuum).

I need to think about this more though :)

bplommer:
> * why are we abstracting the write side rather than the read like (like we do in `hold`)? Unfortunately they both seem useful

I'm not 100% following, but can't we have both?

> * I kinda don't want all the variants, feels like going in this direction makes the api explode even though most are never used. I feel like this about Queues in fs2 in general.

I agree, the combinatorial explosion in the Queue API isn't pretty. It seems like that could be solved using a (typed) builder pattern, so you'd have something like `Queue[F, O].noneTerminated.bounded(n).enqueueToStream { queue => /* */ }` - what do you think?

> * We cannot use the name `async` imo, too prominent in cats-effect for something completely different (although  it does fit in a vacuum).

Is it really completely different? They don't have the same signature, but they have the same basic shape - they take a function that provides values to a callback (in one case encoded as `Function1`, the other as `Enqueue`), run that function in the background, and output the provided values. I'm not really attached to the naming though 🙂

core/shared/src/main/scala/fs2/Stream.scala
bplommer: Yep, happy to add those if/when the API is agreed

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2083,30 +2079,27 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
...
-                  val enqueue =
-                    queue.enqueue1(Some(value.get)).as {
-                      Stream.eval(f(o).attempt).evalMap(value.complete)
-                    }
-
-                  Concurrent[F2].race(dequeueDone.get, enqueue).map {
-                    case Left(())      => Stream.empty
-                    case Right(stream) => stream
-                  }
+    Stream.eval(Deferred[F2, Unit]).flatMap { dequeueDone =>
```
diesalbla, line 2082: I am sorry to say, this may have conflicts with https://github.com/typelevel/fs2/pull/1998/files, which is in the `develop` branch but which I forgot to backport...

diesalbla:
@bplommer Considering the changes to `Stream` since this PR was opened, and that we have removed the queue implementations from `fs2` (in favour of `cats-effect`) and Fabio Labella introduced the `Channel` in #2326 , this PR seems to have fallen to the side. Closing for now.

PR #2048, opened 2020-09-23
fs2.Stream convenient fromOption method

nikiforo:
Consider this synthetic example:
```
  import fs2.Stream

  def getFraction(n: Int): Option[Float] = if(n == 0) None else Some(1.toFloat / n)

  val stream =
    for {
      i <- fs2.Stream.repeatEval(IO.delay(Random.nextInt))
      fractionOpt = getFraction(i)

      fraction1 = Stream.emits(fractionOpt.toList)
      fraction2 = Stream.fromIterator(fractionOpt.iterator)
      fraction3 = Stream.fromOption(fractionOpt)

      fraction <- fraction1
    } yield fraction
```

I want to transform the result of the stream with use of `getFraction` method. I've found 2 method to accomplish it: emits, fromIterator. But the most accurate method is fromOption, which seems to be absent from Stream companion object. This PR adds fromOption.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3273,6 +3273,26 @@ object Stream extends StreamLowPriority {
     now.flatMap(go)
   }
 
+  private[fs2] final class PartiallyAppliedFromOption[F[_]](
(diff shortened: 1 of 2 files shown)


mpilquist:
Thanks!

PR #2050, opened 2020-09-26
Readability (+ future performance) improvements

bplommer:
- Remove redundancy/duplication
- Use combinators where appropriate
- Reduce nesting in flatMap chain


Daenyth:
I'm curious if the changes to use combinators affects the benchmarks at all - do you see a difference with/without them?

bplommer:
> I'm curious if the changes to use combinators affects the benchmarks at all - do you see a difference with/without them?

I haven't run the whole suite, but the `parJoin` benchmark with 256x concurrency seems to be about 2% slower. Maybe it's worth overriding the combinators in the standard CE `Ref` implementation to remove the extra indirection, especially as `Semaphore` now uses them in CE3?

bplommer:
With https://github.com/typelevel/cats-effect/pull/1242, derived methods on `Ref` Are overridden to operate directly on the underlying `AtomicReference`. So the slight performance hit will become a slight boost.

mpilquist:
@bplommer When I merged this forward to develop branch, I lost the changes to `parJoin`, as the implementation on develop is a bit different. In general, let's target develop branch for readability improvements and keep main branch for performance or correctness issues. Thanks!

PR #2054, opened 2020-09-28
Cleanup Compiler.apply

mpilquist:
This PR moves the builder creation and finalization in to the `to` method, resulting in `Compiler#apply` being a simple chunky fold. This did require exposing a `Monad[G]` though. Tradeoff seems worth it.

### DIFF
--- a/core/shared/src/main/scala/fs2/Compiler.scala
+++ b/core/shared/src/main/scala/fs2/Compiler.scala
@@ -51,14 +51,7 @@ private[fs2] trait CompilerLowPriority2 {
       )(foldChunk: (B, Chunk[O]) => B): Resource[F, B] =
         Resource
           .makeCase(CompileScope.newRoot[F])((scope, ec) => scope.close(ec).rethrow)
-          .flatMap { scope =>
-            def resourceEval[A](fa: F[A]): Resource[F, A] =
-              Resource.suspend(fa.map(a => a.pure[Resource[F, *]]))
-
-            resourceEval {
-              Pull.compile(stream, scope, true, init)(foldChunk)
-            }
-          }
+          .evalMap(scope => Pull.compile(stream, scope, true, init)(foldChunk))
     }
 }
 
--- a/core/shared/src/main/scala/fs2/Compiler.scala
+++ b/core/shared/src/main/scala/fs2/Compiler.scala
@@ -21,7 +21,7 @@
 
 package fs2
 
-import cats.{Applicative, Id}
+import cats.{Applicative, Id, Monad}
 import cats.effect._
 import cats.effect.concurrent.Ref
 import cats.syntax.all._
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Compiler.scala
```diff
@@ -34,31 +34,29 @@ import scala.annotation.implicitNotFound
   "Cannot find an implicit Compiler[F, G]. This typically means you need a Concurrent[F] in scope"
 )
 sealed trait Compiler[F[_], G[_]] {
-  private[fs2] def apply[O, B, C](stream: Pull[F, O, Unit], init: () => B)(
-      fold: (B, Chunk[O]) => B,
-      finalize: B => C
-  ): G[C]
+  private[fs2] val target: Monad[G]
```
mpilquist, line 37: Alternatively, we could change `.to(..)` to take a `Monad[G]` constraint, which would then percolate up to `toList`, `toVector`, and `string`.

core/shared/src/main/scala/fs2/Compiler.scala
```diff
@@ -34,31 +34,29 @@ import scala.annotation.implicitNotFound
...
+      def apply[O, B](
           stream: Pull[F, O, Unit],
-          init: () => B
-      )(foldChunk: (B, Chunk[O]) => B, finalize: B => C): Resource[F, C] =
+          init: B
+      )(foldChunk: (B, Chunk[O]) => B): Resource[F, B] =
         Resource
           .makeCase(CompileScope.newRoot[F])((scope, ec) => scope.close(ec).rethrow)
           .flatMap { scope =>
             def resourceEval[A](fa: F[A]): Resource[F, A] =
```
SystemFw, line 55: Since you're at it, I've fixed this in ce some time ago, we can delete it

SystemFw (approved):

core/shared/src/main/scala/fs2/Compiler.scala
mpilquist: Cool, fixed with an evalMap

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4525,8 +4525,12 @@ object Stream extends StreamLowPriority {
       * res3: scodec.bits.ByteVector = ByteVector(5 bytes, 0x0001020304)
       * }}}
       */
-    def to(collector: Collector[O]): G[collector.Out] =
-      compiler(underlying, () => collector.newBuilder)((acc, c) => { acc += c; acc }, _.result)
+    def to(collector: Collector[O]): G[collector.Out] = {
+      implicit val G: Monad[G] = compiler.target
+      G.unit.map(_ => collector.newBuilder).flatMap { builder =>
```
SystemFw, line 4530: It's ok since it's very localised, but this appears a third time (here, compile token, and somewhere else), it probably benefits from a `suspendSynchronousInfallible` or something, with a `Compiler.Target` bound and the same scaladoc that's in `Token`

mpilquist: BTW, note that we don't have a `Compiler.Target[G]` here -- I originally tried making the new `compiler.target` val return a `Compiler.Target[G]` but we don't have an inductive instance for `Compiler.Target[F] => Compiler.Target[Resource[F, *]]`

SystemFw: ah, I missed this. Well, nevermind then :)

PR #2056, opened 2020-09-29
Prevent double completion of deferred during scope interruption

SystemFw:


Daenyth (approved):

SystemFw:
MemoryLeak spec failed. Don't know if it's to be expected or not.

mpilquist:
Passed on re-run -- I think the `groupWithin` tests are leak prone. I don't *think* they are real leaks, just test environment interactions.

SystemFw:
>  I think the groupWithin tests are leak prone. I don't think they are real leaks, just test environment interactions.

That sucks because my next PR is going to reimplement groupWithin 😬

Issue #2060, opened 2020-10-01
Add groupedWeightedWithin


RobertoUa:
Similar to Akka Streams https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/groupedWeightedWithin.html
Zio seems to have it as well https://github.com/zio/zio/pull/1137
For Monix I've found this (not included in Monix) https://gist.github.com/rcano/7d709ff76ca2416999864a7da0a50d42

Daenyth:
I think this is the signature we want:
```scala
def groupWeightedWithin[F[_], A, B](maxWeight: Long, maxTime: FiniteDuration)(elemWeight: A => Long): Pipe[F, A, Chunk[B]]
```

And it would have a semantic like `groupWithin(max, dur) = groupWeightedWithin(max, dur)(_ => 1)`

SystemFw:
I'm working on a general solution to this problem. Can't promise it's going to work yet, but let's see where it goes.

Daenyth:
Also this probably will be best done *after* @SystemFw's rework of `groupWithin`, to minimize wasted/duplicated effort

rehanone:
Are there any plans to progress this in near future?

RobertoUa:
I've implemented `groupWeighedWithin` based on Timed Pull
https://gist.github.com/RobertoUa/189d6ccd3aa82021ed9f22af145c1bca

One question though. Is it possible to push buffered items downstream on stream interruption/cats-effect cancelation? And make that downstream uncancellable?

PR #2062, opened 2020-10-02
Timed Pull

SystemFw:
Timed Pulls aim to cover a gap in our current api: aggregation with timeouts.

Examples:
- The current implementation of `groupWithin` , which is very complex
- `groupWeighedWithin`, which was asked in #2060 
- In the past I've needed `groupAdjacentWithin` for some Kafka code
- Things like posting a metric if the stream is not "advancing" in time

Since the demise of `AsyncPull`, `uncons` cannot be timed out, and the alternative approach needed to implement this set of combinators is complex: you need queues, a mechanism of resettable timeout, a mechanism to avoid losing messages, etc.

Timed Pull brings a much simpler, `Pull`-like api to the problem, but without requiring any "kernel mode" operations.

This is still _very_ WIP, but I'd like to get some feedback about the api before I go implement the various combinators and relative tests and docs.


core/shared/src/main/scala/fs2/concurrent/TimedPull.scala
```diff
@@ -0,0 +1,289 @@
...
+import cats.syntax.all._
+
+object tp {
+  trait TimedPull[F[_], A] {
+    type Timeout
+    def uncons: Pull[F, INothing, Option[(Either[Timeout, Chunk[A]], TimedPull[F, A])]]
+    def startTimer(t: FiniteDuration): Pull[F, INothing, Unit]
+  }
+  object TimedPull {
+    def go[F[_]: Temporal, A, B](pull: TimedPull[F, A] => Pull[F, B, Unit]): Pipe[F, A, B] = { source =>
```
mpilquist, line 38: Instead of this signature, what do you think of:

```scala
// On ToPull[F[_], O]
def timed[A](pull: TimedPull[F, O] => Pull[F, B, Unit])(implicit t: Temporal[F]): Pull[F, B, Unit]
```

Usage like:
```
s.pull.timed { tp => ... }.stream
```

mpilquist (commented):
Looks pretty cool. One question - how much of the pull creation API would need to be duplicated on `TimedPull` -- e.g., how do you get a `unconsN` and `unconsLimit` that's time aware?

core/shared/src/main/scala/fs2/concurrent/TimedPull.scala
SystemFw: I was thinking of doing both, kinda like `Broadcast` does it. 
I prefer yours for normal usage tbh, my only "issue" with it is that if you go to look at `TimedPull`, there's nothing there showing you how to use it (and you need to refer to `ToPull`)

SystemFw:
> how do you get a unconsN and unconsLimit that's time aware?

That's a great question, one I do not have a great answer to yet. My instinct is to be light on duplication, but perhaps the two you have mentioned should be there.

SystemFw:
> e.g., how do you get a unconsN and unconsLimit that's time aware?

Btw, if `Pull` was left abstract (more like an interface than a sealed trait), there is a scary/intriguing possibility of transforming existing Pulls into timed versions, but it would need to be able to reinterpret `Pull.pure` as well. Not quite sure whether that's a fantastic or terrible idea :)

mpilquist:
We need explicit signaling of the timeout starts though, right? E.g. startTimeout then uncons 2 chunks versus timeout uncons timeout uncons? Or would the transformed pulls just add a timeout to each uncons?

SystemFw:
> Or would the transformed pulls just add a timeout to each uncons?

I mean, it's a pretty far out idea, but I was thinking each `uncons` would handle a potential a timeout and each "return" would start/reset one. You need to know which `Pull` you're reinterpreting though, cause for some "return" is `pure`, and for others is "output".  But overall I'd like to get this api (the manual one) in regardless, given how hard it is to implement the equivalent functionality right now, and then think longer about Pull introspection

core/shared/src/main/scala/fs2/concurrent/TimedPull.scala
```diff
@@ -0,0 +1,289 @@
...
+import fs2.internal.Token
+import cats.effect._
+import cats.syntax.all._
+
+object tp {
+  trait TimedPull[F[_], A] {
+    type Timeout
+    def uncons: Pull[F, INothing, Option[(Either[Timeout, Chunk[A]], TimedPull[F, A])]]
+    def startTimer(t: FiniteDuration): Pull[F, INothing, Unit]
+  }
```
Daenyth, line 36: This all makes sense to me, it's a nice mirror of the Pull api 👍

core/shared/src/main/scala/fs2/concurrent/TimedPull.scala
```diff
@@ -0,0 +1,289 @@
...
+  //  the stream enqueues a chunk
+  //  the timeout enqueues
+  //  go pulls a chunk
+  //  go reaches limit, resets and starts another timeout, then recurs
+  //  go receives stale timeout, needs to be able to discard it
+  //  the state tracking the current timeout needs to be set before the next iteration of go
+
+  // if we express timedUncons and startTimer as `F`, that guarantee is hard to achieve
+  // cause they could be concurrent. We could express both as Pulls, forcing a sync loop.
+  // need to add a note that the `unconsTimed` `Pull`, unlike `uncons`, is not idempotent
```
Daenyth, line 153: Can you elaborate at this? For me it's not obvious what the cause or implication of this is

SystemFw: sure, I need some time to formulate a response, it's bit subtle (which is why I noted it down first thing).
As a start, try and get familiar with the stale timeout check in the old `groupWithin`

SystemFw: ah wait, do you mean the idempotency comment? That's not true anymore, it is idempotent now, since it's expressed in terms of `uncons1`. I thought you meant why using `Pull` in TimedPull rather than `F[Unit]` and `F[Option[Either[Timeout, Chunk[A]]]]`, that's the subtle bit, but it's settled now

Daenyth: I was referring to the idempotency comment, yes

SystemFw: Right, so I knew I wanted to have `Pull` to forbid concurrency in the api (that's the thing I thought you were asking), and at first I was thinking about having an explicit queue and do `uncons = Pull.eval(dequeue1)`, which would mean however that `uncons >> uncons` would get different element, whereas the normal Pull.uncons is idempotent. I thought it didn't matter too much anyway since it's generally used linearly, but I was happier when I found out I could avoid the explicit queue altogether and rely on `merge` only, instead.

core/shared/src/main/scala/fs2/concurrent/TimedPull.scala
SystemFw: The alternative is to put `TimedPull` inside `Stream`, which is how `StepLeg` does it, and have the signature you propose exclusively. I don't mind that.

SystemFw: Yeah I'm going with your suggestion for sure, since it allows reusing things like `repeatPull`

core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
```diff
@@ -706,6 +706,9 @@ class StreamCombinatorsSuite extends Fs2Suite {
   }
 
   group("groupWithin") {
+    // TODO more tests:
+    // with no waiting, it should be equal to chunkN
```
Daenyth, line 710: done

core/shared/src/main/scala/fs2/concurrent/TimedPull.scala
```diff
@@ -0,0 +1,332 @@
...
+  // on then emits 3 elems, the current implementation will wait another
+  // 800 millis before emitting those 3, whereas arguably they should be
+  // emitted immediately
+  //
+  // a solution would be to go in another recursive state, has timed out,
+  // which emits the next chunk as soon as it arrives, and resets the timer then.
+  // the question is what to do if the chunk which arrives after the limit is bigger
+  // than the size: do we emit all of it (after splitting), or do we accumulate the remainder
+  // for the next chunk?
+
```
SystemFw, line 131: I'd like to get some opinions here. I think my preferred semantics is: if a timeout triggers but there are accumulated elements, then the timer is not reset. When the next chunk arrives, up to `n` elements of it are emitted immediately, the timer is reset, and the remainder (if any) is carried in the next iteration of `go`.

SystemFw: The main point of doubt for me is between this

> up to n elements of it are emitted immediately [.. ]and the remainder (if any) is carried in the next iteration of `go`

and emitting the remainder immediately as well

SystemFw: I have settled for the semantics described above

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4537,28 +4537,22 @@ object Stream extends StreamLowPriority {
...
-              if (timeToWait <= 0.nanos)
-                Stream.emit(currentTimeout.id) ++ nextAfter(currentTimeout)
-              else Stream.sleep_[F](timeToWait)
-            }.repeat
+            time
+              .discrete
+              .dropWhile(_.id == initial.id)
+              .switchMap { timeout =>
+                Stream.sleep(timeout.d).as(timeout.id)
+              }
```
SystemFw, line 4555: I have a question on the commit here @mpilquist @djspiewak . I think only data can give a definite answer but I just want to gauge what you think first. Look at the diff for implementing `timeouts` in that commit. 
The problem is implementing resettable timeouts.
The first one only uses one extra fiber, but it only works if  calls to `resetTimer` never try to reset a timeout which hasn't triggered yet with a shorter one.
The second is fully general, but it switches fiber at every timeout reset.

My question is, do you think the first (based on your sensibility) would offer a significant advantage in perf and/or memory utilisation? If so, do you think it's worth trying to introduce it as a "fast-path" for when `resetTimer` is called in the right conditions (and only fallback to `switchMap` if not), or is that too much trouble for little gain?

djspiewak: Interesting question. Pondering…

One thought here is that fibers are extremely cheap. The context shift around `start` doesn't even involve memory barriers by default, since it just plops the new fiber onto the same worker thread, so you can pretty safely spam that function. The allocation itself creates some memory pressure, but the resident size of `IOFiber` is not enormous. It's basically analogous to the magnitude of allocations you need to perform if you interpret `.pull.uncons1` on a small (~16) chunk, but with fewer objects. Specifically, five objects get allocated in addition to the `IOFiber` when you call `start`, two of which are small arrays. The object header of `IOFiber` itself is the largest memory footprint in the whole menagerie.

Tldr, I suspect that vomiting out fibers is going to be fine. It's not something I would want to be doing in a hot path, but I also wouldn't want to be hitting `Stream` chunk boundaries within a hot path either (for the same reason), and it's around the same order of magnitude cost.

SystemFw: Cool. Just note that `switchMap` is a lot more than _just_ creating a Fiber (queues, semaphores etc are involved), but I think the memory pressure concern would be the same, and sounds like it shouldn't be too much of an issue

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4560,6 +4593,39 @@ object Stream extends StreamLowPriority {
...
+  *   }
+  * }}}
+  *
+  * Where `doSomething` and `doSomethingElse` are `Pull` computations
+  * such as `Pull.output`, in addition to `TimedPull.timeout`.
+  *
+  * See below for detailed descriptions of `timeout` and `uncons`, and
+  * look at the [[ToPull.timed]] scaladoc for an example of usage.
+  */
+  trait TimedPull[F[_], O] {
```
mpilquist, line 4623: Thoughts on defining this as `Pull.Timed` instead of `Stream.TimedPull`?

SystemFw: oh interesting. It is closer to `ToPull` in spirit, but I do like `Pull.Timed` and it's probably more discoverable, so if you're definitely :+1: on `Pull.Timed`, I'll move it there :)

SystemFw:
Great, there is a failure in MemoryLeakSpec for groupWithin. Although I have a vague memory that you've seen it recently (but before this change) @mpilquist ?

SystemFw:
oh yeah https://github.com/typelevel/fs2/pull/2056#issuecomment-700817183 , I'll restart

SystemFw:
I ran the `groupWithin 2` memory leak test locally a couple of times and it passes with no problems (manual inspection with `visualvm` agrees)

PR #2067, opened 2020-10-05
Move file operations to a Files[F] tagless interface

mpilquist:
Stuff to address:
- ~`FileHandle` currently can be constructed via `Files[F].open*` or directly via `FileHandle.from*`. Should the `from` constructors be made private or otherwise removed?~
- ~`ReadCursor` and `WriteCursor` have convenience constructors in their companions. Should those be removed or moved to `Files`?~
- ~Is there a better name for `BaseFiles`?~ Renamed to `SyncFiles`
- Docs

### DIFF
--- a/io/src/main/scala/fs2/io/file/Files.scala
+++ b/io/src/main/scala/fs2/io/file/Files.scala
@@ -106,7 +106,7 @@ trait BaseFiles[F[_]] {
   def move(source: Path, target: Path, flags: Seq[CopyOption] = Seq.empty): F[Path]
 
   /** Creates a `FileHandle` for the file at the supplied `Path`. */
-  def openPath(path: Path, flags: Seq[OpenOption]): Resource[F, FileHandle[F]]
+  def open(path: Path, flags: Seq[OpenOption]): Resource[F, FileHandle[F]]
 
   /** Creates a `FileHandle` for the supplied `FileChannel`. */
   def openFileChannel(channel: F[FileChannel]): Resource[F, FileHandle[F]]
@@ -122,6 +122,11 @@ trait BaseFiles[F[_]] {
     * Reads all data from the file at the specified `java.nio.file.Path`.
     */
   def readAll(path: Path, chunkSize: Int): Stream[F, Byte]
+
+  /**
+    * Returns a `ReadCursor` for the specified path. The `READ` option is added to the supplied flags.
+    */
+  def readCursor(path: Path, flags: Seq[OpenOption] = Nil): Resource[F, ReadCursor[F]]
 
   /**
     * Reads a range of data synchronously from the file at the specified `java.nio.file.Path`.
@@ -143,23 +148,11 @@ trait BaseFiles[F[_]] {
   def size(path: Path): F[Long]
 
   /**
-    * Creates a stream containing the path of a temporary file.
-    *
-    * The temporary file is removed when the stream completes.
-    */
-  def tempFileStream(
-      dir: Path,
-      prefix: String = "",
-      suffix: String = ".tmp",
-      attributes: Seq[FileAttribute[_]] = Seq.empty
-  ): Stream[F, Path]
-
-  /**
     * Creates a resource containing the path of a temporary file.
     *
     * The temporary file is removed during the resource release.
     */
-  def tempFileResource(
+  def tempFile(
       dir: Path,
       prefix: String = "",
       suffix: String = ".tmp",
@@ -167,22 +160,11 @@ trait BaseFiles[F[_]] {
   ): Resource[F, Path]
 
   /**
-    * Creates a stream containing the path of a temporary directory.
-    *
-    * The temporary directory is removed when the stream completes.
-    */
-  def tempDirectoryStream(
-      dir: Path,
-      prefix: String = "",
-      attributes: Seq[FileAttribute[_]] = Seq.empty
-  ): Stream[F, Path]
-
-  /**
     * Creates a resource containing the path of a temporary directory.
     *
     * The temporary directory is removed during the resource release.
     */
-  def tempDirectoryResource(
+  def tempDirectory(
       dir: Path,
       prefix: String = "",
       attributes: Seq[FileAttribute[_]] = Seq.empty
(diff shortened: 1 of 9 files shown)


io/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -0,0 +1,510 @@
...
+    *
+    * Note that the result of this method is immediately outdated. If this
+    * method indicates the file exists then there is no guarantee that a
+    * subsequence access will succeed. Care should be taken when using this
+    * method in security sensitive applications.
+    */
+  def exists(path: Path, flags: Seq[LinkOption] = Seq.empty): F[Boolean]
+
+  /**
+    * Moves (or renames) a file from the source to the target path.
```
Daenyth, line 102: Is this guaranteed to be atomic?

io/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -0,0 +1,510 @@
...
+  /** Creates a `FileHandle` for the file at the supplied `Path`. */
+  def openPath(path: Path, flags: Seq[OpenOption]): Resource[F, FileHandle[F]]
+
+  /** Creates a `FileHandle` for the supplied `FileChannel`. */
+  def openFileChannel(channel: F[FileChannel]): Resource[F, FileHandle[F]]
+
+  /**
+    * Get file permissions as set of [[PosixFilePermission]].
+    *
+    * Note: this will only work for POSIX supporting file systems.
```
Daenyth, line 117: If not posix, then does this return empty set, or fail?

io/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -0,0 +1,510 @@
...
+
+  /**
+    * Writes all data to a sequence of files, each limited in size to `limit`.
+    *
+    * The `computePath` operation is used to compute the path of the first file
+    * and every subsequent file. Typically, the next file should be determined
+    * by analyzing the current state of the filesystem -- e.g., by looking at all
+    * files in a directory and generating a unique name.
+    */
+  def writeRotate(
```
Daenyth, line 428: Possible out of scope for this PR, but there was a question about how to ensure that when using `writeRotate` with strings, of how to align chunks, for example so that printing a single line definitely goes into either one file or another, and wouldn't split in the middle - for example for logging. If that's easy, might be good to get it in here

Daenyth (approved):
This looks really nice 👍

io/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -0,0 +1,510 @@
...
+    * Returns the size of a file (in bytes).
+    */
+  def size(path: Path): F[Long]
+
+  /**
+    * Creates a stream containing the path of a temporary file.
+    *
+    * The temporary file is removed when the stream completes.
+    */
+  def tempFileStream(
```
diesalbla, line 150: So, the implementation for this funciton, below, is defined as `Stream.resource` applied to the function below. 

- On one hand, since the `Stream.resource` method does not use the `Sync[F]` instance, this implementation could be added here, on the top-level definition. 
- On the other hand, perhaps there is no need to add a method for what is a fixed combination of a primitive and other method in this trait.

io/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -0,0 +1,510 @@
...
+
+  /**
+    * Writes all data to the file at the specified `java.nio.file.Path`.
+    *
+    * Adds the WRITE flag to any other `OpenOption` flags specified. By default, also adds the CREATE flag.
+    */
+  def writeAll(
+      path: Path,
+      flags: Seq[StandardOpenOption] = List(StandardOpenOption.CREATE)
+  ): Pipe[F, Byte, INothing]
```
diesalbla, line 214: I wonder if we could introduce another definition with variance, of the following form: 
```Scala
def writeAll[ InF[x] >: F[x], OutF[x] <: F[x]] (-----): InF[Byte] => OutF[INothing]
```

io/src/main/scala/fs2/io/file/Files.scala
mpilquist: It is if you pass `StandardCopyOption.ATOMIC_MOVE`

io/src/main/scala/fs2/io/file/Files.scala
mpilquist: `UnsupportedOperationException`

io/src/main/scala/fs2/io/file/Files.scala
mpilquist: Yeah, I remember seeing that conversation. Maybe one option is defining how `writeRotate` handles a chunk that spans the limit. E.g., given a rotation limit of 1024 and file position of 1000, receiving a chunk of 32 bytes currently writes 24 bytes, rotates, and then 8 more to new file. Maybe that could be configurable as to whether chunks get split -- e.g. a hard limit vs a soft limit.

io/src/main/scala/fs2/io/file/Files.scala
mpilquist: Perhaps, though if doing so, I'd rather just avoid aliases completely and return `Stream[F, Byte] => Stream[F, INothing]`, which has the right variance. This would be a different PR though -- one that deprecates `Pipe` completely.

io/src/main/scala/fs2/io/file/Files.scala
mpilquist: Yeah, not sure why we have both. Let's keep just the `Resource` returning versions.

io/src/main/scala/fs2/io/file/Files.scala
Daenyth: Might be worth adding that to the docs - I wouldn't have know how to find that necessarily

io/src/main/scala/fs2/io/file/Files.scala
Daenyth: 👍 for only keeping the `=> Resource` version

io/src/main/scala/fs2/io/file/Files.scala
mpilquist: Yeah - want to open an issue for these types of doc improvements? Note this PR didn't introduce these methods but rather just moved them from `fs2.io.file` to `Files`.

PR #2071, opened 2020-10-08
Add TCP, UDP, and TLS networking capabilities

mpilquist:
I don't feel as good about this one as `Files[F]`, but interested in thoughts.


SystemFw:
I think the basic idea is solid, especially because we have a passthrough instance so the user can choose to just use `Async` anyway (we need docs for that). The only thing is that maybe it's a bit too granular? we could have `Files[F]` and `Network[F]` (or `Sockets[F]`), although it's arbitrary any way.

Why don't you feel as good about this one?

mpilquist:
:+1: for `Network[F]` instead of `tcp.Sockets[F], udp.Sockets[F], tls.TLS[F]`

Seems kind of odd so directly wrap `Async[F]` and have no operations on the capability trait. It's okay though.

SystemFw:
> Seems kind of odd so directly wrap Async[F] and have no operations on the capability trait. It's okay though.

mm ok I see

PR #2072, opened 2020-10-08
Failing test case for resource acquired from another thread pool failing to clean up when stream is cancelled

jatcwang:
If you run the test you'll see "starting acq" and "acq" being printed out but not "released".

Without evalOn, the resource is successfully released. 

This is a minimized example from from https://github.com/typelevel/fs2/pull/2072

### DIFF
--- a/core/shared/src/test/scala/fs2/StreamSuite.scala
+++ b/core/shared/src/test/scala/fs2/StreamSuite.scala
@@ -21,8 +21,9 @@
 
 package fs2
 
-import scala.concurrent.duration._
+import java.util.concurrent.Executors
 
+import scala.concurrent.duration._
 import cats.data.Chain
 import cats.effect.{ExitCase, IO, Resource, SyncIO}
 import cats.effect.concurrent.{Deferred, Ref}
@@ -31,8 +32,9 @@ import org.scalacheck.Gen
 import org.scalacheck.Arbitrary.arbitrary
 import org.scalacheck.Prop.forAll
 import org.scalacheck.effect.PropF.forAllF
-
 import fs2.concurrent.{Queue, SignallingRef}
+
+import scala.concurrent.ExecutionContext
 
 class StreamSuite extends Fs2Suite {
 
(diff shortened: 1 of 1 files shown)


core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -550,6 +552,40 @@ class StreamSuite extends Fs2Suite {
...
+          println("starting acq")
+          Thread.sleep(1000)
+          println("acq")
+        }))(_ => IO(println("released")))
+
+      otherECRes.use { otherEC =>
+        (for {
+          st <- Ref.of[IO, List[String]](List.empty)
+          res <- IO.race(
+            Stream.resource(res(st, otherEC)).compile.toVector,
```
Daenyth, line 578: Does it also repro with this:
```suggestion
            res(st, otherEC).use(_ => IO.unit),
```

core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -550,6 +552,40 @@ class StreamSuite extends Fs2Suite {
...
+            IO.sleep(100.millis)
+          )
+          result <- st.get
+          _ <- IO.sleep(1.seconds)
```
mpilquist, line 582: Does this still fail if this sleep is 2 seconds?

jatcwang: Yes still fails with 5 sec

core/shared/src/test/scala/fs2/StreamSuite.scala
jatcwang: Nope it's fine I get 

```
starting acq
acq
released
```

mpilquist:
Fixes by #2073

jatcwang:
Thanks for the fast fix @mpilquist, howeve it doesn't seem to fix the bug in doobie.

After minimizing, it seem like nested `Stream.resource` causes no release to occur. Here's the test case:

```
test(
      "Resource acquired on another thread pool is released when compiled stream is interrupted"
    ) {
      import java.util.concurrent.Executors
      val otherECRes =
        Resource.make(IO {
          scala.concurrent.ExecutionContext.fromExecutorService(Executors.newCachedThreadPool())
        })(ec =>
          IO {
            ec.shutdown
          }
        )
      def res(idx: Int, ec: ExecutionContext): Resource[IO, Unit] =
        Resource.make(munitContextShift.evalOn(ec)(IO {
          println(s"starting acq ${idx}")
          Thread.sleep(1000)
          println(s"acq ${idx}")
        }))(_ => IO(println(s"released ${idx}")))

      otherECRes.use { otherEC =>
        for {
          _ <- testCancelation(
            Stream
              .resource(res(0, otherEC))
              .flatMap(_ => Stream.resource(res(1, otherEC)))
          )
          _ <- IO.sleep(5.seconds) // give ample time for cleanups to happen
        } yield ()
      }
    }
```

For me it prints

```
starting acq 0
acq 0
```
(no cleanup happens)

Replacing the second Stream.resource with anything else seems to be fine.

mpilquist:
@jatcwang Can you try again with 2.4-cecaf46?

jatcwang:
That seem to have fixed it @mpilquist :)

PR #2073, opened 2020-10-08
Fix resource leak that occurs when stream is canceled while a resource is being acquired

mpilquist:
Includes a minimized test case for the bug reported in #2072.

### DIFF
--- a/core/shared/src/test/scala/fs2/StreamSuite.scala
+++ b/core/shared/src/test/scala/fs2/StreamSuite.scala
@@ -427,13 +427,14 @@ class StreamSuite extends Fs2Suite {
       }
     }
 
-    test("#2072 - stream canceled while resource acquisition is running".only) {
+    test("#2072 - stream canceled while resource acquisition is running") {
       for {
         ref <- Ref.of[IO, Boolean](false)
         _ <- testCancelation {
           // This will be canceled after a second, while the acquire is still running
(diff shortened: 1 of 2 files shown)


core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -427,176 +427,188 @@ class StreamSuite extends Fs2Suite {
...
-        def unfoldTree(seed: Int): Tree[Int] =
-          Tree(seed, Stream(seed + 1).map(unfoldTree))
-
-        assert(unfoldTree(1).flatten.take(10).toList == List.tabulate(10)(_ + 1))
-      }
+        _ <- IO.sleep(200.milliseconds)
+        released <- ref.get
+      } yield assert(released)
     }
+  }
```
mpilquist, line 441: From here to the end is just the result of noticing the stream cancelation test group accidentally included tests unrelated to stream cancelation -- I moved the closing brace up to right point which then left shifted all the accidentally captured tests.

Daenyth (approved):

core/shared/src/test/scala/fs2/StreamSuite.scala
diesalbla: It is all whitespace enough, so Github can hide it.

diesalbla (approved):

PR #2074, opened 2020-10-09
Fix for resource leak that occurs when parent scope is closed during resource acquisition

mpilquist:
This needs a lot more work but committing what I've found so far. When the resource acquisition completed in `CompileScope#acquireResource`, the wrapper `ScopedResource` instance was getting registered on the containing scope, but the containing scope's state was already closed. Not sure why the test I added yesterday passed though.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/CompileScope.scala
+++ b/core/shared/src/main/scala/fs2/internal/CompileScope.scala
@@ -93,11 +93,10 @@ private[fs2] final class CompileScope[F[_]] private (
 
   /**
     * Registers supplied resource in this scope.
-    * This is always invoked before state can be marked as closed.
+    * Returns false and makes no registration if this scope has been closed.
     */
   private def register(resource: ScopedResource[F]): F[Boolean] =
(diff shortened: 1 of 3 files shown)


mpilquist:
Okay, figured it out. The fix from yesterday was only partial. When the resource acquisition completed, the associated `ScopedResource` was registered with the current scope, which was already closed. However, that scope was still the current scope at stream termination and hence was closed again, freeing the resource. Adding a subsequent bracket resulted in changing the current scope, and hence there was no double closure.

In this PR, we correctly prevent any resources from being registered on a closed scope. This causes the original test to fail as well as the new variant. We fix by invoking the finalizer if registration fails.

I'd like to change these internals a bit -- e.g., why does a closed scope even have the ability to reference resources -- but I'll save the more drastic changes for develop branch.

SystemFw:
nice work!

> I'd like to change these internals a bit -- e.g., why does a closed scope even have the ability to reference resources -- but I'll save the more drastic changes for develop branch.

:+1: :+1:

PR #2077, opened 2020-10-11
Add Stream.bracketFull

mpilquist:
This was relatively straightforward, but I don't know how to implement `translate`, which naively needs a `G ~> F` now. Thoughts?

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -819,9 +819,19 @@ object Pull extends PullLowPriority {
       self match {
         // safe to cast, used in translate only
         // if interruption has to be supported concurrent for G has to be passed
-        case a: Acquire[f, r] =>
-          ??? // TODO
-        // Acquire[G, r](p => fK(a.resource(p)), (r, ec) => fK(a.release(r, ec)))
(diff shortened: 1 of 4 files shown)


SystemFw:
> which naively needs a G ~> F now.

is this needed to translate `Poll`?

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -818,8 +819,9 @@ object Pull extends PullLowPriority {
       self match {
         // safe to cast, used in translate only
         // if interruption has to be supported concurrent for G has to be passed
-        case a: Acquire[F, r] =>
-          Acquire[G, r](fK(a.resource), (r, ec) => fK(a.release(r, ec)))
```
mpilquist, line 822: We're given a `Poll[G]`, `Poll[F] => F[R]`, and `F ~> G` here.

mpilquist:
> is this needed to translate Poll?

Yeah, that was my thought, though we'd need to add `imapK` to `Poll`.

mpilquist:
On second thought, adding `Resource.makeMask` should introduce the same issue in CE3.

SystemFw:
I'm wondering about the impact on real world scenarios, especially doobie

Issue #2076, opened 2020-10-11
Race condition in reactivestreams StreamSubscriber


vigoo:
There is a race condition in `StreamSubscriber` between `onNext` and `onComplete` because they run the effects manipulating the underlying state machine via `f.runAsync().unsafeRunSync` (in the `Runner` implicit class). 

This can lead to a case where `OnNext` arrives in the `UpstreamCompletion` state. 

(I found this issue with AWS SDK's [SimpleHttpContentPublisher](https://github.com/aws/aws-sdk-java-v2/blob/cdfb7d3f6a7dafb37be84f6e95e49feb0d0af273/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/http/async/SimpleHttpContentPublisher.java) connected to a http4s request entity stream)

SystemFw:
If you have the time, a repro test case would be awesome.

Completely coincidentally, someone posted your project https://github.com/vigoo/prox in the cats-effect dev channel a couple of days ago, I think it would make a great addition to the fs2 ecosystem page https://fs2.io/ecosystem.html, if you wish to include a link to it there

vigoo:
Thanks, that would be nice. I'll open a PR for adding it, and will also try to add a test (and maybe a fix) for this issue.

vigoo:
@SystemFw the above pull request has a test case reproducing the issue, and also a fix.

SystemFw:
Thanks, I'll have a look soon, once I get some time :)

vigoo:
@SystemFw @mpilquist  do you think it would be possible to release this fix as a 2.4.5?

PR #2078, opened 2020-10-11
Reproducer and fix for race condition in StreamSubscriber

vigoo:
Test case demonstrating https://github.com/typelevel/fs2/issues/2076 and a proposed fix.


SystemFw (approved):

SystemFw:
@mpilquist what should we do with the bincompat failures here? Do you think it's enough to eliminate `Runner` as an argument?

vigoo:
Yeah I wanted to ask for advice regarding those. I only introduced that runner parameter to be able to inject the delays in the reproducer test case. I considered only the `fromPublisher` / `toUnicastPublisher` / `toStream` functions part of the public API and those did not change, but I guess I was wrong about that :)

SystemFw:
>  I only introduced that runner parameter to be able to inject the delays in the reproducer test case.

maybe you can provide a custom instance of `ConcurrentEffect` for that purpose instead?

vigoo:
Good idea, let me try

vigoo:
Binary compatibility restored :)

PR #2082, opened 2020-10-12
Weave translation into compilation

diesalbla:
Closes #2801

The `Stream.translate[G[_]]` method changes the effect type `F[_]` of a stream by applying a natural transformation
to each action in the stream. Translation was implemented as tree construction, by building a new Pull object and its components are built during translation. This was no longer satisfactory, due to ongoing changes in cats-effect.

This Pull Request replaces that approach with one that incorporates and weaves translation into stream compilation.
We introduce a new Pull subclass, `Translate`, which embeds a natural transformation and the stream to be translated.

#### Compile loop 

Translation is now part of the compile loop. As compilation unrolls the stream, each action is interpreted and translated to the "final" target effect. For this, we modify the recursive `go` method to 1) add a type parameter `G[_]`, to mark the  effect the stream is "written in",  from the final target one; (2) a parameter for the accumulated translation , which is initialised to the `FunctionK.id`; (3) we handle a `Translate` class by composing the translation; and (4) we handle stops (from the `Step` class) by translating the `Stop` directly and passing it to the top level.

#### OpenScope and Concurrent

As a consequence of these changes, the way that we handle concurrency for opening scopes changes. Before, the `OpenScope` class had an optional `Concurrent` instance, to indicate if the action of opening the scope could be interrupted or not. Threading the `Concurrent` instances was done through the type-class `TranslateInterrupt`.

Now, the class `OpenScope` no longer incorporates an optional `Concurrent` instance, which makes it a pure (non-F-parametric) subclass of `Pull`. A way of looking at this is that  `OpenScope` is just an _instruction_, but it is the _interpretation_ of the stream into an effect F that really defines and implement that concurrency and interruption. Thus, the typeclass `TranslateInterrupt` has been removed.


SystemFw (approved):
Roughly makes sense to me

PR #2088, opened 2020-10-17
Add Stream.bracketFull

mpilquist:


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -474,6 +480,13 @@ object Pull extends PullLowPriority {
       resource: F[R],
       release: (R, Resource.ExitCase) => F[Unit]
   ) extends AlgEffect[F, R]
+
+  private final case class AcquireCancelable[F[_], R](
+      resource: Poll[F] => F[R],
+      release: (R, Resource.ExitCase) => F[Unit],
+      monadCancel: MonadCancel[F, Throwable]
+  ) extends AlgEffect[F, R]
+
```
diesalbla, line 489: Should we try joining `Acquire` and `AcquireCancelable` into a single class with an optional `MonadCancel`? The constructor for `acquire` could then become 

```Scala
  private[fs2] def acquire[F[_], R](
      resource: F[R],
      release: (R, Resource.ExitCase) => F[Unit]
  ): Pull[F, INothing, R] = AcquireCancelable(_ => resource, release, None)

  private[fs2] def acquireCancelable[F[_], R](
      resource: Poll[F] => F[R],
      release: (R, Resource.ExitCase) => F[Unit]
  )(implicit F: MonadCancel[F, Throwable]): Pull[F, INothing, R] =
    AcquireCancelable(resource, release, Some(F))

  private final case class AcquireCancelable[F[_], R](
      resource: Poll[F] => F[R],
      release: (R, Resource.ExitCase) => F[Unit],
      monadCancel: Option[MonadCancel[F, Throwable]]
  ) extends AlgEffect[F, R]
```

mpilquist: I tried that but it didn't work out. We'd need a `Poll[F]` to call `acquire(p)` but the only way to get a `Poll[F]` is via a `MoandCancel[F, Throwable].uncancelable(p => ...)`.

We could make all resource acquisition require `MonadCancel` instead. I tried that last night but it forced me to define all the `compression` methods as `(implicit F: Sync[F], F2: MonadCancel[F, Throwable])`.

mpilquist: I just pushed an update that uses a single case class for `Acquire` and represents the acquisition as `Either[F[R], (Poll[F] => F[R], MonadCancel[F, Throwable])]`

PR #2090, opened 2020-10-17
RFC: Weave output mapping into compilation

diesalbla:
After weaving and mixing translation into compilation, it seems that a similar thing can be done with mapping the outputs... If this can continue, we can remove all the existing tree-level unrollings (calls to `ViewL`) and weave them into the main compilation loop (which may become large though)

The onlt thing is that this only works for streams (unit). The `Pull.mapOutput` method seems not to be used anywhere... however it is part of the API of Pull... could it be needed somewhere out there...


mpilquist:
I like it. Cleans up the interpreter quite a bit.

diesalbla:
will have to start again...

PR #2091, opened 2020-10-19
RFC: scalafmt configuration: add fewer lines in parameters

diesalbla:
Current ScalaFMT configuration forces a new line between the parameters of a function/method declaration, or the arguments of a call, even when there are too few parameters/arguments and they fit in one line within the text width (100 characters). These configurations avoid that when there are only 2 parameters.

Also, for the implicit parameters, it may be easier to read the implicit keyword before the first parameter, rather than on a previous line at the end.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -12,6 +12,11 @@ rewrite.rules = [
   PreferCurlyFors
 ]
 
+optIn.configStyleArguments = false
+runner.optimizer.forceConfigStyleMinArgCount = 3
+
+newlines.implicitParamListModifierPrefer = before
+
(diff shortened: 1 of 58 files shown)


mpilquist:
No strong opinions on this but in general want to find a way to avoid churn in syntax changes. I had originally hoped the "default" scalafmt syntax would be stable but it seems to change frequently. Ideally, we would come up with a stable configuration that never changes.

Also, if we change syntax, we should PR it to main and then manually merge main to develop and fix conflicts, to make forward merges from main to develop conflict free.

mpilquist:
Closing this for now but feel free to reopen when ready

PR #2092, opened 2020-10-19
Upgrade to latest CE3 release

mpilquist:
One TCP test is failing, perhaps related to changes to `Semaphore`

### DIFF
--- a/io/src/main/scala/fs2/io/tcp/SocketGroup.scala
+++ b/io/src/main/scala/fs2/io/tcp/SocketGroup.scala
@@ -314,9 +314,9 @@ final class SocketGroup(channelGroup: AsynchronousChannelGroup) {
                 case None       => F.async.unit
                 case Some(took) => go(buff, (remains - took).max(0))
               }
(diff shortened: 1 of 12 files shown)


djspiewak:
That definitely seems like a bug in either semaphore or resource. I don't understand how though.

io/src/main/scala/fs2/io/tcp/SocketGroup.scala
```diff
@@ -314,7 +314,7 @@ final class SocketGroup(channelGroup: AsynchronousChannelGroup) {
                 case None       => F.async.unit
                 case Some(took) => go(buff, (remains - took).max(0))
               }
-          writeSemaphore.withPermit {
+          writeSemaphore.permit.use { _ =>
```
mpilquist, line 317: Note: previously, the mutable byte buffer was created eagerly here. As a result, multiple uses of a single write action ended up being no-ops for every use after the first. The concurrent write test in SocketSuite only passed as a result of this bug. When suspending the byte buffer creation correctly, like in this PR, the test failed as a result of the socket write buffers filling.

PR #2097, opened 2020-10-20
Import from cats.effect.kernel where possible

rkrzewski:
While working on upgrading Doobie to Cats Effect 3 https://github.com/tpolecat/doobie/pull/1280 I have been advised that in order to maintain compatibility with all Cats Effect implementations, I should depend on `cats-effect-kernel` and `cats-effect-std` artifacts for compilation and on `cats-effect` only in tests. This leads to a problem for code that also depends on `fs2-core`. I'm getting compilation errors like the following:
```
[error] /home/rafal/workspace/scala/tpolecat/doobie/modules/core/src/main/scala/doobie/util/transactor.scala:167:9: Symbol 'type cats.effect.package.Resource' is missing from the classpath.
[error] This symbol is required by 'value fs2.Stream.r'.
[error] Make sure that type Resource is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
[error] A full rebuild may help if 'Stream.class' was compiled against an incompatible version of cats.effect.package.
[error]         Stream.resource(connect(kernel)).flatMap { conn =>
[error]         ^
```
Fs2 depends on `SyncIO` from `cats-effect` module in the main code and untangling it to the point of depending on `cats-effect-std` might be not feasible, but a simple workaround for the problem I'm facing would be foregoing the aliases in `cats.effect` package and importing symbols from `cats.effect.kernel` and `cats.effect.std` directly.

### DIFF
--- a/core/jvm/src/main/scala/fs2/compression.scala
+++ b/core/jvm/src/main/scala/fs2/compression.scala
@@ -26,7 +26,7 @@ import java.nio.charset.StandardCharsets
 import java.time.Instant
 import java.util.zip._
 
-import cats.effect.Sync
+import cats.effect.kernel.Sync
 
 /** Provides utilities for compressing/decompressing byte streams. */
 object compression {
--- a/core/shared/src/main/scala/fs2/Compiler.scala
+++ b/core/shared/src/main/scala/fs2/Compiler.scala
@@ -22,8 +22,8 @@
 package fs2
 
 import cats.{Id, Monad}
-import cats.effect._
-import cats.effect.kernel.Ref
+import cats.effect.SyncIO
+import cats.effect.kernel.{Concurrent, Poll, Ref, Resource, Sync}
 import cats.syntax.all._
 
 import fs2.internal._
--- a/core/shared/src/main/scala/fs2/Hotswap.scala
+++ b/core/shared/src/main/scala/fs2/Hotswap.scala
@@ -23,9 +23,8 @@ package fs2
 
 import cats.ApplicativeError
 import cats.syntax.all._
-import cats.effect.{Concurrent, Resource}
-import cats.effect.kernel.Ref
-import cats.effect.implicits._
+import cats.effect.kernel.{Concurrent, Ref, Resource}
+import cats.effect.kernel.implicits._
 
 /** Supports treating a linear sequence of resources as a single resource.
   *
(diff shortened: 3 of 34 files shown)


djspiewak:
Oh. This is unpleasant. Are there any evictions here or is this actually same-same versions of cats-effect? If there aren't any evictions and you're still getting this error, it kind of calls into question the whole aliasing idea.

rkrzewski:
No relevant evictions:
```
[info] org.tpolecat:doobie-free_2.13:0.9.2+39-77c669ba+20201021-0129-SNAPSHOT [S]
[info]   +-co.fs2:fs2-core_2.13:3.0.0-M1
[info]   | +-org.scodec:scodec-bits_2.13:1.1.20
[info]   | 
[info]   +-org.scala-lang.modules:scala-collection-compat_2.13:2.1.6
[info]   +-org.typelevel:cats-core_2.13:2.2.0
[info]   | +-org.typelevel:cats-kernel_2.13:2.2.0
[info]   | 
[info]   +-org.typelevel:cats-effect-kernel_2.13:3.0.0-M1
[info]   +-org.typelevel:cats-effect-std_2.13:3.0.0-M1
[info]   +-org.typelevel:cats-free_2.13:2.1.1
[info]     +-org.typelevel:cats-core_2.13:2.1.1 (evicted by: 2.2.0)
[info]     +-org.typelevel:cats-macros_2.13:2.1.1
```
I've excluded `fs2-core` -> `cats-effect` dependency and :boom:

mpilquist:
I'm fine with skipping aliases, though I don't recommend excluding cats-effect-core from the classpath. Like you noted, FS2 has an internal dependency on `SyncIO` for running non-effectful streams.

mpilquist (approved):

djspiewak:
Oh I see. This was caused by the exclusion. Okay that was quite scary.

rkrzewski:
I've added the exclusion to check that I'm not using something that I should't be. I can remove it later.
But this makes me wonder what is the interoperability story for fs2 and Monix? Monix users would run effectful streams using `Task` and non-effectful using `SyncIO` and for that, they need `cats-effect` (core) module on the classpath? Is mixing / matching those kinds of streams possible and is it a potential problem?

mpilquist:
That's totally fine. FS2 has used `SyncIO` (and before that, `IO`) internally to evaluate pure/fallible/id streams for a very long time.

rkrzewski:
I'm wondering if `SyncIO` could be moved to `cats-effect-std` module then. I had quick look at the source and it's relatively small and self-contained. What do you think @djspiewak?

PR #2098, opened 2020-10-21
generalize `Stream.groupWithin`

zakolenko:
I was trying to generalize `Stream.groupWithin` so there are more ways to aggregate:
```
  def groupBy[F2[x] >: F[x], O2 >: O](
      f: Chunk[O] => F2[(Chunk[O2], Option[Chunk[O2]])],
      d: FiniteDuration
  )(implicit timer: Timer[F2], F: Concurrent[F2]): Stream[F2, Chunk[O2]]

where `f` returns two chunks, the first chunk goes to the current group, if there is a second chunk, then it opens a new group
```

At the time of creating this PR, noticed that @SystemFw created https://github.com/typelevel/fs2/pull/2062, so now I'm not sure if these changes are up to date.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1445,6 +1445,36 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
       d: FiniteDuration
   )(implicit timer: Timer[F2], F: Concurrent[F2]): Stream[F2, Chunk[O]] =
     Stream
+      .eval(Ref.of[F2, Int](0))
+      .flatMap { sizeRef =>
+        groupBy(
+          chunk =>
+            sizeRef.modify { size =>
+              val toTake = n - size
+              if (toTake > chunk.size) (size + chunk.size, (chunk, None))
+              else {
+                val (unit, rest) = chunk.splitAt(toTake)
+                (0, (unit, Some(rest)))
+              }
+            },
+          d
+        )
+      }
+
+  /**
+    * Divide this streams into groups of elements received within a time window,
+    * or limited by `f` that splits chunk into two new chunks.
+    * The first goes to the current group, the second, if exists, opens the new group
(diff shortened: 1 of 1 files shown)


SystemFw:
Yeah these are out of date now, sorry :)

Basically `TimedPull` gives a way of running arbitrary aggregations with timeouts

zakolenko:
@SystemFw , got it, thanks :)

What do you think about keeping a `groupBy` as it could be easier to express the aggregation?

SystemFw:
well, I'm not sure it is easier because I haven't fully understood how `groupBy` is supposed to be used. Can you give some examples? E.g. can you express `groupWithin` with it?

zakolenko:
Sure, like [this](https://github.com/typelevel/fs2/pull/2098/files#diff-c74e0aabdb2667a6642b051521123685e2957facf8f23b76c10fddc9cda2727fR1447 )

SystemFw:
mm, it kinda forces you to thread extra state "externally" through a `Ref`, and I think the treatment of that `Ref` is going to get more complex once you integrate some of the changes I've done to the behaviour of `groupWithin` to fix what I think are bugs in the current behaviour.
 In particular, if a timeout triggers but the accumulator is empty, the new `groupWithin` will emit the next chunk immediately if its size is `< n`, and emit the first `n` elements of it otherwise, and only then it resets the timeout, and I'm not entirely sure if you can express that with your shape. 
On the other hand, it also makes some assumptions in the general api which seem to be `groupWithin` specific, i.e. the decision to not emit empty chunks.

All that being said, how about waiting for #2062 to get merged (shouldn't be long now), and then we can see if what you have, with a different name, can integrate with it? (probably worth looking at a few examples and see how it stacks up)

zakolenko:
@SystemFw , I totally agree, let's wait for https://github.com/typelevel/fs2/pull/2062.
Thank you very much for a great explanation, quick reply and your work!

mpilquist:
Closing this one for now but feel free to reopen after integrating with `TimedPull`

PR #2099, opened 2020-10-22
Slightly optimize Monad.pure for Stream

satorg:
Leverage `Stream.emit` in `Monad.pure` implementation instead of `Stream.apply` (which always takes `Seq`).
Conversation it [here](https://gitter.im/functional-streams-for-scala/fs2?at=5f90715aa0a3072b43973c3c)


satorg:
I am uncertain though to which branch it is better to PR. If `main` is a wrong branch, I can rebase this "fix".

kubukoz (approved):
I think this belongs in main, but it should land on develop (fs2 3.x) as well 👍 

Thanks!

PR #2101, opened 2020-10-23
Restrict Pull.ViewL to Streams (Unit result)

diesalbla:
The ViewL operation and trait, which represents loop unrolling,  is only ever used when interpreting or compiling a Stream.
Therefore, we can assume that it onlt works on Pull with Unit as the result type. For this we remove references to it from the Pull class itself. For that:

- Remove the class method `mapOutput` of the Pull class. Instead we write a new method with large match.
- Remove the `asHandler` method of the `Pull` class and expand it in where it was used, at top of compilation.
- Remove nested object ViewL, but make all its subclasses Private to the Pull object.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -54,15 +54,6 @@ sealed abstract class Pull[+F[_], +O, +R] {
   /** Alias for `_.map(_ => o2)`. */
   def as[R2](r2: R2): Pull[F, O, R2] = map(_ => r2)
 
-  private[Pull] def asHandler(e: Throwable): Pull[F, O, R] =
-    ViewL(this) match {
-      case Result.Succeeded(_) => Result.Fail(e)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -444,63 +430,51 @@ object Pull extends PullLowPriority {
...
+        case e: Action[F, O, Unit] => new EvalView[F, O](e)
+        case b: Bind[F, O, y, Unit] =>
           b.step match {
             case r: Result[_] =>
               val ry: Result[y] = r.asInstanceOf[Result[y]]
               mk(b.cont(ry))
             case e: Action[F, O, y2] =>
-              new ViewL.View[F, O, y2, Z](e) {
-                def next(r: Result[y2]): Pull[F, O, Z] = b.cont(r.asInstanceOf[Result[y]])
+              new View[F, O, y2](e) {
```
mpilquist, line 461: Not related to this PR but why do we use an anonymous class here?

PR #2102, opened 2020-10-23
Added isFile and isDirectory to Files

domaspoliakas:
Hello,

Added a couple of small calls to the `Files` capability: `isFile` and `isDirectory`

I opted to make these more granular calls instead of the ability to fetch `BasicFileAttributes` because I wasn't sure if `BasicFileAttributes` is referentially transparent. 

I added `isDirectory` and `isFile` separately (as opposed to implementing one and allowing the other via negation of the implemented one) because according to the [javadoc](https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/BasicFileAttributes.html) a path may also be a symbolic link or "other". I did not how to test the other 2 cases however, so I did not implement appropriate methods for them.

Lastly, according to the contributor docs I should run `fmtCheck; test; doc; mimaReportBinaryIssues; docs/mdoc; microsite/makeMicrosite`; I tried my best, but `docs/mdoc` failed due to not being a recognised task and `microsite/makeMicrosite` failed due to a dependency (org.tpolecat/tut-core) not being available for dotty, which I wasn't sure how to resolve. The other tasks succeeded fine.

Let me know if you need me to make any changes. If you'd prefer to expose `BasicFileAttributes` instead I'd be happy to make such changes.

### DIFF
--- a/io/src/main/scala/fs2/io/file/Files.scala
+++ b/io/src/main/scala/fs2/io/file/Files.scala
@@ -198,11 +198,40 @@ sealed trait SyncFiles[F[_]] {
       append: Boolean
   ): F[WriteCursor[F]]
 
+  /** Tests whether a file is a directory.
+    *
+    * The options sequence may be used to indicate how symbolic links are handled for the case that the file is a symbolic link.
+    * By default, symbolic links are followed and the file attribute of the final target of the link is read.
+    * If the option NOFOLLOW_LINKS is present then symbolic links are not followed.
+    *
+    * Where is it required to distinguish an I/O exception from the case that the
+    * file is not a directory then the file attributes can be read with the
+    * readAttributes method and the file type tested with the BasicFileAttributes.isDirectory() method.
+    *
+    * @param path the path to the file to test
+    * @param options - options indicating how symbolic links are handled
+    * @return true if the file is a directory; false if the file does not exist, is not a directory, or it cannot be determined if the file is a directory or not.
+    */
   def isDirectory(
       path: Path,
       linkOption: Seq[LinkOption] = Nil
   ): F[Boolean]
 
+  /** Tests whether a file is a regular file with opaque content.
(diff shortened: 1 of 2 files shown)


io/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -197,6 +197,17 @@ sealed trait SyncFiles[F[_]] {
       file: FileHandle[F],
       append: Boolean
   ): F[WriteCursor[F]]
+
+  def isDirectory(
+      path: Path,
+      linkOption: Seq[LinkOption]
```
mpilquist, line 203: Add default argument here and on isFile?

io/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -366,6 +377,22 @@ object SyncFiles {
     ): F[WriteCursor[F]] =
       if (append) file.size.map(s => WriteCursor(file, s)) else WriteCursor(file, 0L).pure[F]
 
+    def isDirectory(
+        path: Path,
+        linkOption: Seq[LinkOption] = Nil
```
mpilquist, line 382: Don't need the default here since this is the implementation, which is private.

domaspoliakas:
@mpilquist thanks for the review! I made the changes from your comments now.

For future reference, do you get notified when PRs get updated with new commits, or would one normally drop a message or something?

mpilquist:
I generally get notified but sometimes I get so many notifications that I lose them so never be shy about bumping an issue.

domaspoliakas:
It appears I am a fool and `isDirectory` and `isRegularFile` _are_, in fact, in JDK 8 as well, except I somehow missed them. I changed the implementations to use the appropriate methods instead of using .getAttribute.

I also copied the javadoc for those methods as well to serve as documentation.

PR #2107, opened 2020-10-26
Temp file dirs optional for file system default

domaspoliakas:
Hello.

Made the dir parameter on `tempFile` and `tempDirectory` an `Option[Path]`. Specifying `None` (which is also the default) delegates to using the system's default temp directory. Specifying `Some` works same as it did previously.

### DIFF
--- a/io/src/main/scala/fs2/io/file/Files.scala
+++ b/io/src/main/scala/fs2/io/file/Files.scala
@@ -137,10 +137,14 @@ sealed trait SyncFiles[F[_]] {
     */
   def size(path: Path): F[Long]
 
-  /** Creates a resource containing the path of a temporary file.
+  /** Creates a [[Resource]] which can be used to create a temporary file.
+    *  The file is created during resource allocation, and removed during its release.
     *
-    * The temporary file is removed during the resource release.
     * @param dir the directory which the temporary file will be created in. Pass in None to use the default system temp directory
+    * @param prefix the prefix string to be used in generating the file's name
+    * @param suffix the suffix string to be used in generating the file's name
+    * @param attributes an optional list of file attributes to set atomically when creating the file
+    * @return a resource containing the path of the temporary file
     */
   def tempFile(
       dir: Option[Path] = None,
@@ -149,10 +153,13 @@ sealed trait SyncFiles[F[_]] {
       attributes: Seq[FileAttribute[_]] = Seq.empty
   ): Resource[F, Path]
 
-  /** Creates a resource containing the path of a temporary directory.
+  /** Creates a [[Resource]] which can be used to create a temporary directory.
+    *  The directory is created during resource allocation, and removed during its release.
     *
-    * The temporary directory is removed during the resource release.
     * @param dir the directory which the temporary directory will be created in. Pass in None to use the default system temp directory
+    * @param prefix the prefix string to be used in generating the directory's name
+    * @param attributes an optional list of file attributes to set atomically when creating the directory
+    * @return a resource containing the path of the temporary directory
     */
   def tempDirectory(
       dir: Option[Path] = None,
--- a/io/src/main/scala/fs2/io/file/Files.scala
+++ b/io/src/main/scala/fs2/io/file/Files.scala
@@ -140,9 +140,10 @@ sealed trait SyncFiles[F[_]] {
   /** Creates a resource containing the path of a temporary file.
     *
     * The temporary file is removed during the resource release.
+    * @param dir the directory which the temporary file will be created in. Pass in None to use the default system temp directory
     */
   def tempFile(
-      dir: Path,
+      dir: Option[Path] = None,
       prefix: String = "",
       suffix: String = ".tmp",
       attributes: Seq[FileAttribute[_]] = Seq.empty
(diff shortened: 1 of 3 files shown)


SystemFw:
let's hold on for #2106 pls :)

domaspoliakas:
Sure thing!

SystemFw:
It's going to take me a bit of time to fix one thing, and then you'll have to do some changes to your test suite

domaspoliakas:
Rebased against latest develop, everything is fixed and running now 👍

io/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -151,9 +152,10 @@ sealed trait SyncFiles[F[_]] {
   /** Creates a resource containing the path of a temporary directory.
     *
     * The temporary directory is removed during the resource release.
```
SystemFw, line 154: You didn't write this, but I think it can be improved, I found the wording slightly confusing. How about something like:

> Creates a temporary directory, and returns its path. The directory is removed during the release of the resulting `Resource`

Same applies to the `tempFile` scaladoc above

io/src/test/scala/fs2/io/file/FilesSuite.scala
```diff
@@ -247,19 +247,46 @@ class FilesSuite extends Fs2Suite with BaseFileSuite {
...
+      tempDirectory
+        .use { tempDir =>
+          val files = Files[IO]
+          files
+            .tempFile(Some(tempDir))
+            .use { file =>
+              files.exists(tempDir.resolve(file.getFileName))
+            }
+        }
+        .map(it => assert(it))
```
SystemFw, line 266: `.assertEquals(true)`, for consistency

io/src/test/scala/fs2/io/file/FilesSuite.scala
```diff
@@ -247,19 +247,46 @@ class FilesSuite extends Fs2Suite with BaseFileSuite {
...
+      val files = Files[IO]
+
+      files
+        .tempFile(None)
+        .use { file =>
+          IO(System.getProperty("java.io.tmpdir")).flatMap(dir =>
+            files.exists(Paths.get(dir).resolve(file.getFileName))
+          )
+        }
+        .map(existsInDefault => assert(existsInDefault))
```
SystemFw, line 280: `.assertEquals(true)`

io/src/test/scala/fs2/io/file/FilesSuite.scala
```diff
@@ -274,11 +301,38 @@ class FilesSuite extends Fs2Suite with BaseFileSuite {
...
+      tempDirectory
+        .use { tempDir =>
+          val files = Files[IO]
+          files
+            .tempDirectory(Some(tempDir))
+            .use { directory =>
+              files.exists(tempDir.resolve(directory.getFileName))
+            }
+        }
+        .map(it => assert(it))
```
SystemFw, line 320: `.assertEquals(true)`

io/src/test/scala/fs2/io/file/FilesSuite.scala
```diff
@@ -274,11 +301,38 @@ class FilesSuite extends Fs2Suite with BaseFileSuite {
...
+      val files = Files[IO]
+
+      files
+        .tempDirectory(None)
+        .use { directory =>
+          IO(System.getProperty("java.io.tmpdir")).flatMap(dir =>
+            files.exists(Paths.get(dir).resolve(directory.getFileName))
+          )
+        }
+        .map(existsInDefault => assert(existsInDefault))
```
SystemFw, line 334: `.assertEquals(true)`

SystemFw (changes requested):
Small tidbits

io/src/main/scala/fs2/io/file/Files.scala
domaspoliakas: Reworded + added some bits from nio Files' javadoc.

io/src/test/scala/fs2/io/file/FilesSuite.scala
```diff
@@ -263,7 +263,7 @@ class FilesSuite extends Fs2Suite with BaseFileSuite {
               files.exists(tempDir.resolve(file.getFileName))
             }
         }
-        .map(it => assert(it))
+        .map(existsInTemp => assertEquals(existsInTemp, true))
```
SystemFw, line 266: sorry, what I meant is that you can just call `.assertEquals(true)` instead of doing the `.map` at all. Sorry for the confusion

domaspoliakas: Ah right, that _does_ make more sense than how I interpreted it at first 😅 serves me right for only glancing over munit's docs. 

Will get it amended now.

SystemFw: Thanks. It's the munit cats-effect integration that Michael added very recently which lets you do that

domaspoliakas: That's done now

SystemFw (approved):

domaspoliakas:
I don't know what the normal process for publishing dev branch artifacts is; could I ask for a publish with these changes?

mpilquist:
@domaspoliakas The develop branch publishes hash releases to maven central -- in this case, the version you want is `3.0-cd73a32`

domaspoliakas:
Awesome, thanks 🙇

PR #2108, opened 2020-10-26
Pull - Compile Loop : some renames...

diesalbla:
- `F` (the target effect of compilation) => `Final`
- `B` (the value type of compiled) => `Blob`
- `f` (the fold function) => `foldChunk`
- `R` => `Rap` (see the comment)

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -596,133 +596,137 @@ object Pull extends PullLowPriority {
    * Interpreter uses this to find any parents of this scope that has to be interrupted, and guards the
    * interruption so it won't propagate to scope that shall not be anymore interrupted.
    */
-  private[fs2] def compile[F[_], O, B](
-      stream: Pull[F, O, Unit],
-      initScope: CompileScope[F],
+  private[fs2] def compile[Final[_], O, Blob](
+      stream: Pull[Final, O, Unit],
+      initScope: CompileScope[Final],
       extendLastTopLevelScope: Boolean,
-      init: B
-  )(g: (B, Chunk[O]) => B)(implicit
-      F: MonadError[F, Throwable]
-  ): F[B] = {
+      init: Blob
+  )(foldChunk: (Blob, Chunk[O]) => Blob)(implicit
+      Final: MonadError[Final, Throwable]
+  ): Final[Blob] = {
 
-    sealed trait R[+G[_], +X]
-    case class Done(scope: CompileScope[F]) extends R[Pure, INothing]
-    case class Out[+G[_], +X](head: Chunk[X], scope: CompileScope[F], tail: Pull[G, X, Unit])
-        extends R[G, X]
-    case class Interrupted(scopeId: Token, err: Option[Throwable]) extends R[Pure, INothing]
+    /* A Rap is a Rap-ture in the interpretation of the Pull... which happens when there is an operation
+     */
+    sealed trait Rap[+G[_], +X]
+    case class Done(scope: CompileScope[Final]) extends Rap[Pure, INothing]
+    case class Out[+G[_], +X](head: Chunk[X], scope: CompileScope[Final], tail: Pull[G, X, Unit])
+        extends Rap[G, X]
+    case class Interrupted(scopeId: Token, err: Option[Throwable]) extends Rap[Pure, INothing]
 
-    def go[G[_], X](
-        scope: CompileScope[F],
-        extendedTopLevelScope: Option[CompileScope[F]],
-        translation: G ~> F,
-        stream: Pull[G, X, Unit]
-    ): F[R[G, X]] =
-      viewL(stream) match {
-        case _: Pull.Result.Succeeded[_] =>
-          F.pure(Done(scope))
+    def go[Carrier[_], Elem](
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -596,133 +596,137 @@ object Pull extends PullLowPriority {
...
+                  out.copy(tail = Translate[src, g, Elem](out.tail, tst.fk))
                 case dd @ Done(_)              => dd
                 case inter @ Interrupted(_, _) => inter
               }
 
-            case uU: Step[f, y] =>
-              val u: Step[G, y] = uU.asInstanceOf[Step[G, y]]
-              val stepScopeF: F[CompileScope[F]] = u.scope match {
-                case None          => F.pure(scope)
+            case uU: Step[f, z] =>
```
mpilquist, line 692: Should `f` be `carrier` here? Since there's no relation to `Final`?

mpilquist:
Mixed feelings on this one:
- :+1: on `Final` and `foldChunk`
- :-1: on `Blob` and `Rap`
- neutral on `K` and `carrier`

In general, I'd prefer to err on the side of single letter names unless the longer name provides some semantic clue to the reader. `Blob` and `Rap` don't provide me any additional context but the fact that they have names makes me think I'm missing something.

diesalbla:
> Can we move go and the definition of R and subtypes somewhere external from the body of compile? Any clarity benefit of doing so?

For the `R` ADT, we would then need to add an extra parameter (because the `CompileScope` is parameterised on final, not carrier. For the compile loop, we would need to add a few type and value parameters from main definition, but can be done.

> Instead of Rap, how about Unroll (and rename go to unroll)? I want to say Step (taken) or StepResult (confusable with Step).

Yeah, `Step` and `StepLeg` are already causing confusion... As for `unroll`, we use that word in the comment of `viewL`, which is closer to the idea of manipulating the data structure. What `go` does is folding and running the Pull structure into the `F` effect, until it reaches a point that sends an event that needs to be handled "up the call stack"... 
- Perhaps `Pop`, which indicates both that we are popping an element from the the recursive call stack, and that something popped out.
- Or `Prompt`, since the what the `R` represent is prompting a "controller" with an stopped point.

SystemFw:
Similar thoughts as Michael tbh. For me the main thing with type parameters is being to see which ones are equal to which, and which change, and single letter names do a way better job for that than long meaningful names (to me), especially because the long names end up not being _that_ meaningful anyway. Perhaps we have a top level comment to establish context instead?

diesalbla:
> [...] the main thing with type parameters is being to see which ones are equal to which, and which change, and single letter names do a way better job for that than long meaningful names (to me), especially because the long names end up not being that meaningful anyway. 

Given three or four variables, maybe... after six or seven the confusion raises too much. Within `go`, there are `F`, `O`, `B`, `G`, `X`, `y`, `g`, `z` , `f`, `h` `q`... To add confusion, some are introduced as type parameters of the constructor, and the compiler itself has some problem deciding if they are patterns or they are new declarations or they are constraints...

The criteria followed here is one that also applies to term or variable names: short names for short-scoped variables,  long names for long-scoped variables. The variables `F`, `G`, and `X`, (or `Final`, `Carrier`, and `Elem`) scopes span over 200 lines, so long names are warranted.

> Perhaps we have a top level comment to establish context instead?

That would help, but readable code is a better code comment.

mpilquist:
Can we unify `R` and `Result`? They are almost identical, but `R` carries a `CompileScope` in all 3 constructors whereas `Result` carries a scope id in one constructor only.

diesalbla:
> Can we unify R and Result? They are almost identical, but R carries a CompileScope in all 3 constructors whereas Result carries a scope id in one constructor only.

AAMOF I was trying in a branch to unify both _Interrupted_ classes, since they are identical in their fields and in a few places we just convert one into the other... but I left it after a while. I saw it could be a bit ubly, to have overlapping ADTs, and it has been past practice not to microoptimise unfrequent cases of failure or interruption.

In general, R and Result cannot be unify, because they represent completely different notions. The `Result` is a type of Pull, and therefore Streams, data structure. It can be used externally, pass through, or stored. The `R` classes are an ephemeral data structure, which we used to convey information through phases/stages/levels of the interpretation execution, and whose life-time is bound to that process.

SystemFw:
> One criteria followed here is one that also applies to term or variable names: short names for short-scoped variables, long names for long-scoped variables. The variables F, G, and X, (or Final, Carrier, and Elem) scopes span over 200 lines, so long names are warranted.

do you know what, I actually think that for type params you want the _opposite_ rule, because fundamentally you are doing some different: unification. For long-scoped variables, whose unification you need to keep track of for a long time, short names keep the signatures uncluttered. For short-scoped ones (where you tend to lose track of where you are in the stack of type variables in your head), a name that reminds you what they are can be helpful

SystemFw:
I'd be in favour of renaming `R` though, a single letter for a concrete type (instead of a type param) is confusing. `Rap` doesn't do it for me though :)

SystemFw:
It's also a shame that shadowing will trigger warning, cause I think it can be helpful in things like `innerMapOutput`

diesalbla:
Closed in favour of #2110

PR #2113, opened 2020-10-29
introduce Stream.paginate

yanns:
When consuming resources from an API using pagination, it is usefull to signal at the same time that there are no more elements to read and to emit the last ones read.

For example, when iterating 10 by 10 elements, the last "page" can return 6 elements.
In the case, the `Stream` can emit those 6 elements and stop there, as we know that there are no more elements to read (`6 < 10`)

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3586,6 +3586,54 @@ object Stream extends StreamLowPriority {
     suspend(go(s))
   }
 
+  /** Creates a stream by successively applying `f` until a `None` is returned, emitting
+    * each output `O` and using each output `S` as input to the next invocation of `f`.
+    * Compared to [[unfold]], `f` can still emit last elements at the last iteration.
+    *
+    * @example {{{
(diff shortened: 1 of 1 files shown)


SystemFw:
mm, isn't this the same as `unfoldLoop` and `unfoldLoopEval`?

yanns:
@SystemFw yes I oversaw that. Thx!
Should I add the `unfoldLoopChunkEval` version?

```
  def unfoldLoopChunkEval[F[_], S, O](s: S)(f: S => F[(Chunk[O], Option[S])]): Stream[F, O] =
    Stream.unfoldLoopEval(s)(f).flatMap(Stream.chunk)
```

SystemFw:
> Should I add the unfoldLoopChunkEval version?

I'm inclined to say no, because you can then add another version of any combinator, just by adding `flatMap(Stream.chunk)`. I do wish we had a better website structure to allow people to add small docs/tutorials,   even if it's just "hey, if you need pagination use `unfoldLoopEval`"

Daenyth:
A "cookbook" site would be sweet

PR #2114, opened 2020-10-29
MapOutput: expand last Bind...

diesalbla:
The MapOutput case had an extra Bind wrapper created before iterating. What we can do is to expand the call to `go`
by an inner call and handle cases, much as we do in the goStep method...


mpilquist:
Is this still needed?

PR #2128, opened 2020-11-08
Pull - Replace OpenScope with InScope

diesalbla:
The OpenScope pull constructor, to change the evaluation into a new resource scope, was only used at one place, followed by a `Pull.flatMap` into a continuation, to run an inner stream in the opened scope.

We replace the `OpenScope` class with an `InScope` class, which includes the inner stream. Most of the semantics is now moved to the `goInScope` method, of evaluation.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -439,11 +439,10 @@ object Pull extends PullLowPriority {
   private sealed abstract case class View[+F[_], +O, X](step: Action[F, O, X])
       extends ViewL[F, O]
       with (Result[X] => Pull[F, O, Unit]) {
-    def next(r: Result[X]): Pull[F, O, Unit]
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -436,9 +436,13 @@ object Pull extends PullLowPriority {
     */
   private sealed trait ViewL[+F[_], +O]
 
-  private sealed abstract case class View[+F[_], +O, X](step: Action[F, O, X]) extends ViewL[F, O] {
+  private sealed abstract case class View[+F[_], +O, X](step: Action[F, O, X])
+      extends ViewL[F, O]
+      with (Result[X] => Pull[F, O, Unit]) {
     def next(r: Result[X]): Pull[F, O, Unit]
```
mpilquist, line 442: Can we just delete `next`?

PR #2130, opened 2020-11-09
Upgrade to Scala 3.0.0-M1

mpilquist:


build.sbt
```diff
@@ -18,10 +18,11 @@ ThisBuild / organizationName := "Functional Streams for Scala"
 ThisBuild / homepage := Some(url("https://github.com/typelevel/fs2"))
 ThisBuild / startYear := Some(2013)
 
-ThisBuild / crossScalaVersions := Seq("2.13.3", "2.12.10", "0.27.0-RC1")
+ThisBuild / crossScalaVersions := Seq("2.13.3", "2.12.10", "0.27.0-RC1", "3.0.0-M1")
```
diesalbla, line 21: What is the difference between Dotty 0.27.0-RC1 and Scala 3.0.0-M1?

mpilquist: 3.0.0-M1 is the next release after 0.27. When possible, we're publishing for the last 2 Dotty releases to simplify upgrades for folks downstream.

Issue #2143, opened 2020-11-19
Unable to interrupt stream that has been translated


mpilquist:
This is the root cause of https://github.com/http4s/http4s/issues/3878.

Reproduction for unit test suite:

```scala
  test("translateInterruptible") {
    type Eff[A] = cats.data.EitherT[IO, String, A]
    val Eff = Async[Eff]
    Stream
      .eval(Eff.never)
      .merge(Stream.eval(Eff.delay(1)).delayBy(5.millis).repeat)
      .interruptAfter(10.millis)
      .translateInterruptible(new (Eff ~> IO) {
        def apply[X](eff: Eff[X]) = eff.value.flatMap {
          case Left(t) => IO.raiseError(new RuntimeException(t))
          case Right(x) => IO.pure(x)
        }
      })
      .compile
      .drain
  }
```

There's a hidden `ClassCastException` occurring here, as a result of a soundness issue with `Stream.getScope`. Consider:

```scala
val s = Stream.getScope[OptionT[IO, *]].
  evalMap(_.interrupt(Right(()))).
  translateInterruptible(new (OptionT[IO, *] ~> IO) {
    def apply[X](o: OptionT[IO, X]) = 
      o.value.flatMap {
        case None => IO.raiseError(new NoSuchElementException)
        case Some(x) => IO.pure(x)
      }
    }
  )

s.compiler.lastOrError.unsafeRunSync
java.lang.ClassCastException: class cats.effect.IO$RaiseError cannot be cast to class cats.data.OptionT (cats.effect.IO$RaiseError and cats.data.OptionT are in unnamed module of loader ammonite.runtime.SpecialClassLoader @71def8f8)
```

The problem is that `getScope` returns the current `CompileScope` that's being used during evaluation, but that's a `CompileScope[IO]`, not a `CompileScope[OptionT[IO, *]]`. The call to `scope.interrupt` returns an `IO` which is then casted to an `OptionT[IO, *]`.

mpilquist:
One potential avenue for fixing is (1) deprecating `getScope` (2) for each operation on `Scope`, provide a new operation on `Stream` companion that returns a `Stream[F, X]` (3) implement those new operations via new nodes in stream algebra.

mpilquist:
On second look, this isn't as straightforward as I had hoped, as going down the path I suggested requires implementing `interruptWhen` in terms of `concurrently` whereas `concurrently` is implemented in terms of `interruptWhen`.

[...]

PR #2145, opened 2020-11-21
Fix soundness issue with stream interruption + translation

mpilquist:
Sketch of a fix for #2143, though not sure how to resolve the outstanding issues here.

The main issue is `Stream.getScope`:

```scala
  def getScope[F[x] >: Pure[x]]: Stream[F, Scope[F]] =
    new Stream(Pull.getScope[F].flatMap(Pull.output1(_)))
```

This operation provides a `Scope[F]` which provides various operations that return `F[X]`, like `interrupt`. During stream evaluation, a `CompileScope[G]` is created, where `G` is the target effect type. The problem here is that `getScope` is implemented by returning the current `CompileScope[G]` and when `G != F`, we get class cast exceptions (because `interrupt` returns a `G[Unit]` but caller thinks they have an `F[Unit]`).

The sketch of a solution in this PR introduces a new primitive, `Interrupt`, which invokes the interrupt operation on the current `CompileScope`. However, we have no mechanism to use this operation in the definition of `interruptWhen`, as doing so requires concurrent stream evaluation via `concurrently` or `merge` or `parJoin` but those all use `interruptWhen` internally.


mpilquist:
One option is adding a private `unsafeInterrupt` operation to `CompileScope`, but this would require stream compilation needing an `Async` instance instead of `Concurrent`.

SystemFw:
>  but this would require stream compilation needing an Async instance instead of Concurrent.

that means we will be back to every concurrent stream combinator, and therefore pretty much the whole stack, requiring Async again :(

mpilquist:
I found a better fix, basically `interruptWhen(condition: F[Either[Throwable, Unit]])` becomes a new primitive. I need to clean it up which I'll do later today.

mpilquist:
OK, I pushed a sketch of the better fix. I need to get it compiling on 2.12 and then do something similar for the other operations on Scope, then finally delete getScope entirely.

mpilquist:
Alright, ready for review. I made `getScope` private -- the existing uses are all sound but keeping it public is too error prone.

SystemFw (approved):

PR #2149, opened 2020-11-23
Remove version from README.md

### LINKED ISSUE #1824: Update README to *not* include version numbers
Instead, let's use a badge or something that pulls latest version from maven.org.

asakaev:
Closes #1824 

There is a challenge with links to Javadocs — they carry versions inside. Do we need to generate them dynamically?


asakaev:
@mpilquist

PR #2167, opened 2020-12-05
Introduce FlatMapOutput constructor.

diesalbla:
The implementation of `flatMap` was exposing to the ou

- Introduce a new subtype of `Pull`, which is `flatMapOutput`.
- Interpretation of the `flatMapOutput` is done within the loop-unrolling mechanism. When a `FlatMapOutput` object is found, we direcly create a `View` from it. The `View` issues a `Step` on the stream as its action. 
- For the continuation of those views, we extract a `flatMapOutputCont` function.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -463,56 +463,56 @@ object Pull extends PullLowPriority {
 
   }
 
+  private class FlatMapView[F[_], P, O](fm: FlatMapOutput[F, P, O])
+      extends View[F, O, Option[StepStop[F, P]]](Step[F, P](fm.stream, None)) {
+
+    def apply(r: Result[Option[StepStop[F, P]]]): Pull[F, O, Unit] =
+      r match {
+        case Result.Succeeded(r) =>
+          try flatMapOutputCont[F, P, O](fm.fun)(r)
+          catch { case NonFatal(e) => Result.Fail(e) }
+        case res @ Result.Interrupted(_, _) => res
+        case res @ Result.Fail(_)           => res
+      }
+  }
+
+  private class FlatMapBindView[F[_], P, O](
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -464,16 +464,55 @@ object Pull extends PullLowPriority {
   }
 
   /* unrolled view of Pull `bind` structure * */
-
   private def viewL[F[_], O](stream: Pull[F, O, Unit]): ViewL[F, O] = {
 
+    def flatMapView[Q, P](fm: FlatMapOutput[F, P, Q]): ViewL[F, Q] = {
+      val st = Step[F, P](fm.stream, None)
+      new View[F, Q, Option[StepStop[F, P]]](st) {
```
mpilquist, line 471: Is it possible/worth pulling out a named subtype of `View` that applies a continuation on success and fails on interrupted/fail?

core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -453,6 +454,20 @@ class StreamSuite extends Fs2Suite {
     }
   }
 
+  group("translate") {
```
mpilquist, line 457: Note there's a `StreamTranslateSuite` - maybe move this to that file?

diesalbla: This is not strictly related to the main changes... reverted

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: Done...

diesalbla:
@mpilquist If you approve please hold on from merging... I would like to squash first.

mpilquist (approved):

PR #2171, opened 2020-12-06
MapOutput: small cleanup to handling.

diesalbla:
- Instead of creating Bind at top, we pass down to the `goView` method
  which can take care of it.
- Add variance to the MapOutput class.
- Replace mout case with call to goMapOutput.


mpilquist:
Does this have any functional change or just cleanup?

diesalbla:
> Does this have any functional change or just cleanup?

It is just a cleanup...

mpilquist (approved):

PR #2173, opened 2020-12-06
Pass end-runners through the compile loop.

diesalbla:
The `R` trait, with the subclasses `Done`, `Out`, and `Interrupted`, was only used to carry information from some points of the compilation to other levels with smaller nesting. They were short-lived objects, which only serve a procedural function. 
Instead of passing the message up, the idea is to thread the interpreter

- WE add in the `go` method of the compile loop another parameter, of type `Run`which is the "interpreter" on which this call of `go` is to be paused. As we go deeper (as in `Step` or `Translate`) into the loop, we nest instances of `Run`
- We remove the `R` trait of short-lived objects. We replace the "runner" that was creating them with a runner that generates a callback-like closure, one that would invoke another `Run`. 
- Since at some points of the interpretation we want to fail quickly (with `F.raiseError`), we extend the trait `Run` with a `fail` method.
- For the outer loop, where we want to fold the stream into the result of type `B`, we create an Outer subclass of `Run`, that  directly interprets to an `F[B]`.

There are some places in which we needed to use the `BuildR` instance. In those cases, passing the other interpreter was failing tests.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -650,28 +650,46 @@ object Pull extends PullLowPriority {
 
     type Cont[-Y, +G[_], +X] = Result[Y] => Pull[G, X, Unit]
 
-    trait RunR[G[_], X, End] extends (R[G, X] => End) {
+    trait Run[G[_], X, End] {
+      def done(scope: Scope[F]): End
+      def out(head: Chunk[X], scope: Scope[F], tail: Pull[G, X, Unit]): End
+      def interrupted(scopeId: Unique, err: Option[Throwable]): End
+      def fail(e: Throwable): End
+    }
+
+    trait RunR[G[_], X, End] extends Run[G, X, End] with (R[G, X] => End) {
       def apply(r: R[G, X]): End = r match {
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -869,18 +870,20 @@ object Pull extends PullLowPriority {
...
-        val frgx1 = interruptGuard(scope, view)(tail)
-        goView(frgx1, view)
+        val runner = new BuildR[G, X, End]
+        val tail = for {
+          closedExtendedScope <- maybeCloseExtendedScope
+          newExtendedScope = if (closedExtendedScope) None else extendedTopLevelScope
+          childScope <- scope.open(useInterruption).rethrow
+          bts = boundToScope(childScope.id)
+          fgrx <- go(childScope, newExtendedScope, translation, runner, bts)
+        } yield fgrx
```
mpilquist, line 880: We don't have better-monadic-for on Dotty -- will this final suspended map cause a memory leak?

diesalbla: Perhaps it could. I wrote these changes to make this code more readable, but they are not really necessary.

PR #2174, opened 2020-12-07
Use cats.effect.std.Queue

mpilquist:
~There's one use left, in the implementation of `parJoin`, which uses `Queue.synchronousNoneTerminated`, for which there's no analog in cats-effect. Simply changing that to `Queue.synchronous` resulted in test suite hanging.~


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3652,7 +3660,7 @@ object Stream extends StreamLowPriority {
         // starts with 1 because outer stream is running by default
         running <- SignallingRef(1L)
         // sync queue assures we won't overload heap when resulting stream is not able to catchup with inner streams
-        outputQ <- Queue.synchronousNoneTerminated[F, Chunk[O]]
+        outputQ <- fs2.concurrent.Queue.synchronousNoneTerminated[F, Chunk[O]]
```
mpilquist, line 3663: We'll need an alternative here in order to fully deprecate the fs2 Queue

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3202,6 +3215,60 @@ object Stream extends StreamLowPriority {
...
     new PartiallyAppliedFromIterator(blocking = true)
 
+  /** Returns a stream of elements from the supplied queue.
+    *
+    * The stream terminates upon dequeuing a `None`.
+    *
+    * All elements that are available, up to the specified limit,
+    * are dequeued and emitted as a single chunk.
+    */
+  def fromQueueNoneTerminated[F[_]: Functor, A](
```
mpilquist, line 3225: Assuming cats-effect gets an `Dequeue` type, maybe this should be named `dequeueNoneTerminated` instead of `fromQueueNoneTerminated`.

djspiewak: I think it already has a `Dequeue` type

djspiewak:
I'm curious as to whether the performance of the CE queue in this context is sufficient. Is it on par with what you were seeing from the native fs2 queue?

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: Eek, namespace clash -- I mean a trait that has `take` and `tryTake` but not `offer` or `tryOffer`

mpilquist:
@djspiewak I haven't directly compared the current fs2 Queue with the current cats-effect Queue but I have compared current fs2 Queue with a naive single ref backed Queue and have shown they are equivalent for most use cases.

SystemFw (approved):

PR #2180, opened 2020-12-10
Add IArray support to Chunk

mpilquist:
We'll be able to remove some casts once we compile against M3 https://github.com/lampepfl/dotty/pull/10743

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [3.0.0-M1, 3.0.0-M2, 2.12.10, 2.13.4]
+        scala: [3.0.0-M2, 2.12.10, 2.13.4]
         java: [adopt@1.11]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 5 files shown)


mpilquist:
Tests that are failing are due to an existing FS2 bug on Scala.js:

```scala
    val b = Chunk.newBuilder[Int]
    b += Chunk.singleton(1)
    val c = b.result
    println(c.getClass) // class fs2.Chunk$Bytes
```

The runtime type check in `Chunk.concat` is hitting the `Byte` case b/c a runtime type check on `1` passes for any of the integral primitive types.

mpilquist:
One path forward here...

1) Change `Chunk.concat` to just take a `ClassTag` and instantiate an array of that type. Get rid of all the runtime type tests.
2) Change `Chunk.Queue` to implement `Chunk` instead of requiring a call to `.toChunk`, removing the need to concat all the individual chunks
3) Remove all the specialized primitive subtypes of `Chunk` like `Ints`, `Bytes`, etc. Rename `Chunk.Boxed` to `Chunk.ArrayChunk` or `Chunk.WrappedArray`
4) Remove `Chunk.KnownElementType` as it is no longer needed

(2) is important as otherwise the `ClassTag` constraint bubbles through dozens of methods on `Stream`/`Pull`.

mpilquist:
Closed in favor of #2181

PR #2184, opened 2020-12-16
Move traverse, traverseFilter, and mapFilter to methods on Chunk

mpilquist:


Daenyth (approved):
I assume the purpose here is to make it more autocomplete-friendly?

👍 lgtm. The test failure is in reactive streams so I assume it's a flakey one

PR #2189, opened 2020-12-22
Initial sketch of making Network a full algebra

mpilquist:
@kubukoz @SystemFw This is the start of what it would take to make `Network` a more typical capability trait. Note the fairly significant API changes. Thoughts? /cc @tpolecat


tpolecat:
This seems good at first glance. I ended up doing something like your `TLSContext.Builder` in Skunk and it would be nice to be able to delete it 😄. The API changes don't look like an issue for anything I'm doing.

kubukoz:
I like it quite a lot :)

PR #2194, opened 2020-12-27
 Pull: move FlatMapOutput logic to specialised Run

diesalbla:
- We remove the FlatMap-specific subclasses of `View`, so that now the `viewL` loop handles a FlatMapOut just as a normal `Action`.
- Introduce class FlatMapRunR, a subclass of Run, derived from `StepRunR` class, but taylored to FlatMap operations. In particular, this class performs the inner-pull expansion within the `out` method, which saves the wrapping in `Result`.
- Compiling/interpreting a `FlatMapOut` instruction no longer requires creating a Step, or a Result, or an Option.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -445,7 +445,7 @@ object Pull extends PullLowPriority {
     def apply(r: Result[Unit]): Pull[F, O, Unit] = r
   }
 
-  private final class BindView[+F[_], +O, Y](step: Action[F, O, Y], b: Bind[F, O, Y, Unit])
+  private final class BindView[+F[_], +O, Y](step: Action[F, O, Y], val b: Bind[F, O, Y, Unit])
       extends View[F, O, Y](step) {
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -786,6 +746,68 @@ object Pull extends PullLowPriority {
         }
       }
 
+      def goFlatMapOut[Y](fmout: FlatMapOutput[G, Y, X], view: View[G, X, Unit]): F[End] = {
+        val runr = new BuildR[G, Y, End]
+        val stepRunR = new FlatMapR(view, fmout.fun)
+        F.unit >> // The F.unit is needed because otherwise an stack overflow occurs.
```
mpilquist, line 752: Hm, is this `F.unit >> go` safe?

diesalbla: What part of `safe` are you worried about?  The reason for adding this `F.unit >> go` is that, without it, there is a stack overflow in the evaluation of some tests.

PR #2198, opened 2020-12-29
Upgrade to latest CE3 snapshot where Sync extends MonadCancel

mpilquist:


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -1118,8 +1118,10 @@ private[fs2] class PullMonadErrorInstance[F[_], O] extends MonadError[Pull[F, O,
 
 private[fs2] class PullSyncInstance[F[_], O](implicit F: Sync[F])
     extends PullMonadErrorInstance[F, O]
-    with Sync[Pull[F, O, *]] {
+    with Sync[Pull[F, O, *]]
+    with MonadCancel.Uncancelable[Pull[F, O, *], Throwable] {
   def monotonic: Pull[F, O, FiniteDuration] = Pull.eval(F.monotonic)
   def realTime: Pull[F, O, FiniteDuration] = Pull.eval(F.realTime)
   def suspend[A](hint: Sync.Type)(thunk: => A): Pull[F, O, A] = Pull.eval(F.suspend(hint)(thunk))
+  def forceR[A, B](fa: Pull[F, O, A])(fb: Pull[F, O, B]): Pull[F, O, B] = flatMap(fa)(_ => fb)
```
mpilquist, line 1126: Not sure if this is the right implementation but until there's law checking for CE3, there's no way to check.

SystemFw: You need to `attempt`. `fa.attempt.flatMap(_ => fb)`

mpilquist: Good catch, fixed

mpilquist:
Waiting on https://github.com/typelevel/cats-effect/pull/1563 and https://github.com/typelevel/cats-effect/pull/1556

core/shared/src/main/scala/fs2/CompositeFailure.scala
```diff
@@ -24,9 +24,9 @@ package fs2
 import cats.data.NonEmptyList
 
 /** Represents multiple (>1) exceptions were thrown. */
-final class CompositeFailure(
-    val head: Throwable,
-    val tail: NonEmptyList[Throwable]
+case class CompositeFailure(
```
mpilquist, line 27: cats-effect-testkit provides an `Eq[Throwable]` which uses `.equals`

build.sbt
```diff
@@ -112,12 +112,13 @@ lazy val core = crossProject(JVMPlatform, JSPlatform)
     libraryDependencies ++= Seq(
       "org.typelevel" %%% "cats-core" % "2.3.1",
       "org.typelevel" %%% "cats-laws" % "2.3.1" % Test,
-      "org.typelevel" %%% "cats-effect" % "3.0.0-M5",
-      "org.typelevel" %%% "cats-effect-laws" % "3.0.0-M5" % Test,
-      "org.typelevel" %%% "cats-effect-testkit" % "3.0.0-M5" % Test,
+      "org.typelevel" %%% "cats-effect" % "3.0-65-7c98c86",
```
mpilquist, line 115: We need this version to pick up (1) adjustment to MonadCancel laws (2) fix in TestInstances for an integer overflow

SystemFw (approved):

PR #2200, opened 2020-12-30
Simplify Acquire and weaken MonadCancel constraint

joroKr21:
We don't need `MonadCancelThrow` - we only make use of `uncancelable` and don't actually care about the error type.


mpilquist (approved):
Awesome!

PR #2201, opened 2021-01-01
Add Stream.fromQueue and Stream.fromQueueChunked

RaasAhsan:
I figured these were worth adding, since it's by and far better than blindly running `Stream.repeatEval(queue.take)`. The covariant supertype of `Queue` presents an alternative implementation, but we don't have that for now (https://github.com/typelevel/cats-effect/pull/1553).

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3209,6 +3209,30 @@ object Stream extends StreamLowPriority {
 
   /** Returns a stream of elements from the supplied queue.
     *
+    * All elements that are available, up to the specified limit,
+    * are dequeued and emitted as a single chunk.
+    */
+  def fromQueue[F[_]: Functor, A](queue: Queue[F, A], limit: Int = Int.MaxValue): Stream[F, A] =
+    fromQueueNoneTerminatedChunk_[F, A](
+      queue.take.map(a => Some(Chunk.singleton(a))),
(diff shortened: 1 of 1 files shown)


RaasAhsan:
Just noticed there are corresponding `enqueue` and `enqueueChunks`, maybe we should have them for these ones too. names might need to change though

SystemFw:
So, I want to hold on this for a bit: if we add an fs2 Queue with things like closure (which is super common when using it with Stream), it's better to save these names for that Queue, and give you a stream that closes itself

RaasAhsan:
👍  The original motivation here was to have direct replacements for the deprecated fs2 Queue `enqueue` and `dequeue` operations, which are nonterminating. http4s uses `Queue[F, A]` in a few places and the current migration path is either to change the type signature to include `Option` or write subpar one-liner consume/produce operations

mpilquist:
@SystemFw Seems like your objection is to the already merged `fromQueue` method and its variants. Thoughts on merging this and making a change if a stream specific queue materializes? BTW, I still have doubts we'll have such a thing.

SystemFw:
> Seems like your objection is to the already merged fromQueue method and its variants. Thoughts on merging this and making a change if a stream specific queue materializes?

ah yeah, then sure 

> BTW, I still have doubts we'll have such a thing.

yeah I share them :) it will be cool to have, but between lack of time and no obvious design direction, I'm not so sure

Issue #2197, opened 2020-12-29
Errors are not always catched with .handleErrorWith


nikiforo:
Take a look at https://github.com/nikiforo/fs-handle/pull/1
The example is minimized, but can be minimized even further.

A way to reproduce.
On one tab launch the client `sbt runMain name.nikiforo.Client`
On the other tab launch the server
```
sbt
> runMain name.nikiforo.Server
acquired localhost 20111
got 0 messages
```
Wait a minute
```
<...>
got 538 messages
got 485 messages
exception handling socket Connection reset
got 320 messages
exception handling socket Connection reset
<...>
got 1 messages
DISASTER: Multiple exceptions were thrown (45), first java.io.IOException: Connection reset
```

Despite some errors being catched at https://github.com/nikiforo/fs-handle/pull/1/files#diff-3e571cc6ca077b3bd8dc2bde7f5f8e9787724d1f98e64d56a34e7b263957bfd5R47 others propagate to the encompassing stream, halt it, are catched only at https://github.com/nikiforo/fs-handle/pull/1/files#diff-4a1eb1d4e3714fb61d4adc5f666a0a5c435065c20bfd0433f08697a0a2c96715R27

At this point I want to differentiate whether or not the provided behavior is expected.

@mpilquist

mpilquist:
I looked at this some today and I'm pretty sure the problem is due to `partitionEither`. 
Consider:

```scala
    socket
      .reads(8192)
      .through(pipe)
      .flatMap {
        case OutputResponse(v) => Stream.eval_(socket.write(Chunk.bytes(v)))
        case other => Stream.emit(other)
      }
      .onComplete(Stream.eval_(socket.endOfOutput))
      .handleErrorWith(handler)
```

The `pipe` value has a `partitionEither` deep inside its definition. In essence, `partitionEither` is causing the source stream exception to occur while publishing to the underlying `Broadcast`, which is not directly accessible when using the returned stream. That's why calling `.handleErrorWith` before calling `.through(pipe)` works fine, whereas attaching a `handleErrorWith` to the result of `partitionEither` doesn't.

We should be able to minimize this a bunch by combining a source that raises an exception through a `Broadcast`. Something like `Stream.raiseError[IO](new Err).through(Broadcast(1)).flatten`

mpilquist:
This works okay, so it's not quite as straightforward as I had hoped:

```
scala> implicit class PartitionEither[F[_], O](self: Stream[F, O]) {
[...]

PR #2206, opened 2021-01-08
Reproduction case for #2197

mpilquist:


SystemFw (approved):

RaasAhsan:
Will try to take a look at this later tonight

mpilquist:
@RaasAhsan Awesome, I'm a bit stuck on this one currently. I suspect a race condition in stream interruption (`prefetch` uses `concurrently` which uses `interruptWhen` on both the producer and consume, which uses `F.race`) but haven't worked out why we haven't been able to minimize further. There are two paths to investigate -- why the exception isn't propagating in CE3 and why the exception isn't getting handled by the inner handler.

RaasAhsan:
I think I have a vague understanding of what's going on here, but keep in mind my understanding of the FS2 interpreter shaky at best :) So your minimization uses `prefetch` which transitively call `concurrently` and `interruptWhen`. I'm certain what we're seeing here manifests from race condition between the time the consumer stream's scope is interrupted and the time the consumer stream observes that interruption.

To skip ahead a bit, the effect of the race condition is that, after interruptions are observed, all the remaining bind continuations are just... lost, or something.  Error handlers and scope closures are both encoded as binds, which I think explains why the error propagates to the outer handler in your minimization. I actually minimized it a bit further so no errors are involved (I can push this if you want), and the outcome is that the stream just terminates normally.

I noticed that the race condition happens after an `Output` node is reached and it passes on its `Chunk` to the `endRunner` in scope. I tried to delete this interrupt point (https://github.com/typelevel/fs2/blob/develop/core/shared/src/main/scala/fs2/Pull.scala#L780) and I don't see the outer handler activate anymore. I kind of doubt that's the correct solution, and also that it conclusively fixes the problem, but maybe that gives you a pointer to what's going wrong. It sort of makes sense to me though, since I think both error handlers and scope closures happen later on in `tail`

I think that's about all the info I can give you right now without drowning more time into understanding FS2, but hope it helps :)

SystemFw:
>  drowning more time

drown more time! Also keep in mind that most code in fs2 is still written essentially for a version of IO without interruption. It works in many cases because there are a lot of `start` that decouple interruption, and we did add/fix some things, but I'm not that surprised when these things pop up.

We really need to look at the internals again, but I think both Michael and I are a bit scared to tackle it right now, so any insight you might have (especially coming with fresh eyes) is super welcome :)

mpilquist:
cc @diesalbla

mpilquist:
@RaasAhsan Feel free to push a further minimization.

RaasAhsan:
#2207

RaasAhsan:
I've got a chunk of time available this weekend to dig into this deeper again

diesalbla:
So, I have been experimenting a bit with the reproduction case.

Tracing back the `develop` branch, I have found a different behaviour at  https://github.com/typelevel/fs2/commit/47e6ea681f2bd87951c085200361fc1c38a2479e. On this commit, when one runs the reproduction program, what happens is that there is an uncaught exception... 

```
[info] Inner stream failed: java.lang.RuntimeException
[info] Inner stream failed: java.lang.RuntimeException
[info] Inner stream failed: java.lang.RuntimeException
[info] Inner stream failed: java.lang.RuntimeException
[info] Inner stream failed: java.lang.RuntimeException
[error] java.lang.RuntimeException
[error] 	at fs2.Bug2197$.run(Bug2197.scala:36)
[error] 	at cats.effect.IOApp$Simple.run(IOApp.scala:97)
[error] 	at cats.effect.IOApp$Simple.run$(IOApp.scala:97)
[error] 	at fs2.Bug2197$.run(Bug2197.scala:33)
[error] 	at cats.effect.IOApp.main(IOApp.scala:35)
[error] 	at cats.effect.IOApp.main$(IOApp.scala:27)
[error] 	at fs2.Bug2197$.main(Bug2197.scala:33)
[error] 	at fs2.Bug2197.main(Bug2197.scala)
```

diesalbla:
> I noticed that the race condition happens after an Output node is reached and it passes on its Chunk to the endRunner in scope. I tried to delete this interrupt point (https://github.com/typelevel/fs2/blob/develop/core/shared/src/main/scala/fs2/Pull.scala#L780) and I don't see the outer handler activate anymore. I kind of doubt that's the correct solution, and also that it conclusively fixes the problem, but maybe that gives you a pointer to what's going wrong. It sort of makes sense to me though, since I think both error handlers and scope closures happen later on in tail

So, the line you are referring to is the `out` method of the `FlatMapR` subclass of `Run`. A Stream flatmap operation  `str.flatMap(f)` is represented as a `FlatMapOutput` constructor of `Pull`. This constructor has a _seed_ or _source_ pull, which outputs values of a type `O`, and a function that expands every output from the _source_ stream into a new Pull, in the same effect `F`, but which outputs value of a type `P`.   

```Scala
  private final case class FlatMapOutput[+F[_], O, +P](
      stream: Pull[F, O, Unit],     // let's call this the _source_ string
      fun: O => Pull[F, P, Unit]   /// let us call this function the _expansion_
  ) extends Action[F, P, Unit]
```

Compiling (or interpreting if you like)  a `FlatMapOutput` pull is doing by compiling the source pull. Whenever this compilation of the source pull reaches an `Output`, which would contain a chunk of `O` values, the compilation:  1) constructs the expanded  `Pull` data structure for the expanded pull.  (line 781); 2) binds it to the continuation in the view (Lines 782-790); and 3) continues by compiling that expanded stream 

What this `interruptGuard` does is: _before_ we expand those `O` outputs into a new `Pull[F, P, Unit]`, we check if our current evaluation scope was interrupted.  If it was, it skips ahead to the `outView` (the tail of the stream that goes _after_ the flatmap).

PR #2207, opened 2021-01-09
Further minimization for 2197

RaasAhsan:
Marking as draft because I ripped out parts of `concurrently`


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -550,26 +550,17 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
       def runR: F2[Unit] =
-        that.interruptWhen(interrupt.get.attempt).compile.drain.attempt.flatMap { r =>
-          doneR.complete(r) >> {
-            if (r.isLeft)
-              interrupt
-                .complete(())
-                .void // interrupt only if this failed otherwise give change to `this` to finalize
-            else F.unit
-          }
+        that.compile.drain.attempt.flatMap { r =>
```
RaasAhsan, line 555: I was thinking we could go further and maybe just spawn a fiber that immediately interrupts, didn't get a chance to do that

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -550,26 +550,17 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   )(implicit F: Concurrent[F2]): Stream[F2, O] = {
     val fstream: F2[Stream[F2, O]] = for {
       interrupt <- F.deferred[Unit]
-      doneR <- F.deferred[Either[Throwable, Unit]]
     } yield {
       def runR: F2[Unit] =
-        that.interruptWhen(interrupt.get.attempt).compile.drain.attempt.flatMap { r =>
-          doneR.complete(r) >> {
```
RaasAhsan, line 557: Don't propagate the error

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -550,26 +550,17 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+        that.compile.drain.attempt.flatMap { r =>
+          if (r.isLeft)
+            interrupt
+              .complete(())
+              .void // interrupt only if this failed otherwise give change to `this` to finalize
+          else F.unit
         }
 
-      // stop background process but await for it to finalise with a result
-      val stopBack: F2[Unit] = interrupt.complete(()) >> doneR.get.flatMap(
```
RaasAhsan, line 567: No stream interruption from this to that, since it's not really necessary for the bug to manifest

core/jvm/src/main/scala/fs2/Bug2197.scala
```diff
@@ -0,0 +1,52 @@
...
+  *
+  * Instead, under CE2, the runtime exception escapes to the outer handler occassionally - within
+  * a few seconds to a minute.  Under CE3, the program terminates successfully - which I think is
+  * the result of the exception being raised in an effect finalizer somewhere.
+  */
+object Bug2197 extends IOApp.Simple {
+
+  val s = Stream.eval[IO, Boolean](IO.delay(true)) ++ Stream.raiseError[IO](new RuntimeException)
+
+  // this will terminate successfully
```
RaasAhsan, line 38: after a nondeterministic amount of runs

PR #2212, opened 2021-01-16
Improve performance of TCP socket reads

mpilquist:
- Initializes the socket read buffer to 8192 bytes instead of 0
- Uses `allocateDirect`
- Stores read buffer in a private var instead of a `Ref`

### DIFF
--- a/io/src/main/scala/fs2/io/net/Socket.scala
+++ b/io/src/main/scala/fs2/io/net/Socket.scala
@@ -93,7 +93,7 @@ object Socket {
       writeSemaphore: Semaphore[F]
   )(implicit F: Async[F])
       extends Socket[F] {
-    private var readBuffer: ByteBuffer = ByteBuffer.allocateDirect(8192)
+    private[this] var readBuffer: ByteBuffer = ByteBuffer.allocateDirect(8192)
 
     private def getBufferForRead(size: Int): F[ByteBuffer] = F.delay {
       if (readBuffer.capacity() < size)
--- a/io/src/main/scala/fs2/io/net/Socket.scala
+++ b/io/src/main/scala/fs2/io/net/Socket.scala
@@ -24,6 +24,13 @@ package io
 package net
 
 import com.comcast.ip4s.{IpAddress, SocketAddress}
+import cats.effect.{Async, Resource}
+import cats.effect.std.Semaphore
+import cats.syntax.all._
+
+import java.net.InetSocketAddress
+import java.nio.channels.{AsynchronousSocketChannel, CompletionHandler}
+import java.nio.{Buffer, ByteBuffer}
 
 /** Provides the ability to read/write from a TCP socket in the effect `F`.
   */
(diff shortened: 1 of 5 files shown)


djspiewak:
Almost no difference.

## Before

### Throughput

```
Ramped up to 500 connections.
Total data sent:     16886.5 MiB (17706826192 bytes)
Total data received: 16885.3 MiB (17705547200 bytes)
Bandwidth per channel: 18.886⇅ Mbps (2360.8 kBps)
Aggregate bandwidth: 4721.438↓, 4721.779↑ Mbps
Packet rate estimate: 437386.1↓, 437942.3↑ (6↓, 8↑ TCP MSS/op)
```

### Requests per Second

```
200 clients, running 128 bytes, 30 sec.

Speed: 77217 request/sec, 77213 response/sec
Requests: 2316522
Responses: 2316394
```

## After

### Throughput

```
Ramped up to 500 connections.
Total data sent:     23417.2 MiB (24554712260 bytes)
Total data received: 18057.8 MiB (18934929536 bytes)
Bandwidth per channel: 23.192⇅ Mbps (2899.0 kBps)
Aggregate bandwidth: 5048.715↓, 6547.145↑ Mbps
Packet rate estimate: 451692.5↓, 576245.5↑ (11↓, 20↑ TCP MSS/op)
```

### Requests per Second

```
200 clients, running 128 bytes, 30 sec.

Speed: 78482 request/sec, 78478 response/sec
Requests: 2354488
Responses: 2354342
```

io/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -69,3 +76,139 @@ trait Socket[F[_]] {
...
+            ch.write(
+              buff,
+              (),
+              new CompletionHandler[Integer, Unit] {
+                def completed(result: Integer, attachment: Unit): Unit =
+                  cb(Right(()))
+                def failed(err: Throwable, attachment: Unit): Unit =
+                  cb(Left(err))
+              }
+            )
```
diesalbla, line 181: We cal also use the `CallBackHandler` here: 

```suggestion
            ch.write(buff, (), new CallBackHandler[Integer, Unit](cb))
```

io/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -69,3 +76,139 @@ trait Socket[F[_]] {
...
+        ch.read(
+          buffer,
+          null,
+          new CompletionHandler[Integer, AnyRef] {
+            def completed(result: Integer, attachment: AnyRef): Unit =
+              cb(Right(result))
+            def failed(err: Throwable, attachment: AnyRef): Unit =
+              cb(Left(err))
+          }
+        )
```
diesalbla, line 118: We can extract this `CompletionHandler` out into a class that takes the `cb`. 

```Scala
private class CallBackHandler[Res, Attach](cb: Either[Res, Throwable] => Unit) 
  extends CompletionHandler[Res, Attach] {
    def completed(result: Res, attachment: Attach): Unit =  cb(Right(result))
    def failed(err: Throwable, attachment: Attach): Unit =  cb(Left(err)) 
}
```

With it, we can replace these lines with: 
```suggestion
      F.async_[Int] { cb => 
        ch.read(buffer, null, new CallBackHandler[Integer, AnyRef](cb))
```

diesalbla (approved):

io/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -69,3 +76,140 @@ trait Socket[F[_]] {
...
+    private[this] final val defaultReadSize = 8192
+    private[this] var readBuffer: ByteBuffer = ByteBuffer.allocateDirect(defaultReadSize)
+
+    private def getBufferForRead(size: Int): F[ByteBuffer] = F.delay {
+      if (readBuffer.capacity() < size)
+        readBuffer = ByteBuffer.allocateDirect(size)
+      else
+        (readBuffer: Buffer).limit(size)
+      readBuffer
+    }
```
diesalbla, line 105: This would make clearer that all alterations to the `readBuffer` are under the `readSemaphore`. 
```suggestion
    private def withReadBuffer[A](size: Int)(k: ByteBuffer => F[A]): F[A] = 
      readSemaphore.permit.use { _ => F.delay {
        if (readBuffer.capacity() < size)
          readBuffer = ByteBuffer.allocateDirect(size)
        else
          (readBuffer: Buffer).limit(size)
        k(readBuffer)
      }.flatten
    }
```

io/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -69,3 +76,140 @@ trait Socket[F[_]] {
...
+    def readN(max: Int): F[Option[Chunk[Byte]]] =
+      readSemaphore.permit.use { _ =>
+        getBufferForRead(max).flatMap { buffer =>
+          def go: F[Option[Chunk[Byte]]] =
+            readChunk(buffer).flatMap { readBytes =>
+              if (readBytes < 0 || buffer.position() >= max)
+                releaseBuffer(buffer).map(Some(_))
+              else go
+            }
+          go
```
diesalbla, line 153: Since this recursion can only end up returning a `Some`, could we remove the `Option` wrapper and change the type to `F[Chunk[Byte]]`?

PR #2221, opened 2021-01-22
added second argument to def sliding

nikiforo:


nikiforo:
I'll make the function aware of chunks

nikiforo:
```
2de178db
[info] Benchmark                  (n)   Mode  Cnt    Score   Error  Units
[info] StreamBenchmark.sliding     10  thrpt       104,814          ops/s
[info] StreamBenchmark.sliding    100  thrpt       105,679          ops/s
[info] StreamBenchmark.sliding   1000  thrpt       104,443          ops/s
[info] StreamBenchmark.sliding  10000  thrpt       248,643          ops/s

f9a3896a
[info] Benchmark                  (n)   Mode  Cnt    Score   Error  Units
[info] StreamBenchmark.sliding     10  thrpt       258,618          ops/s
[info] StreamBenchmark.sliding    100  thrpt       265,793          ops/s
[info] StreamBenchmark.sliding   1000  thrpt       254,515          ops/s
[info] StreamBenchmark.sliding  10000  thrpt       576,929          ops/s

d2ca1985
[info] StreamBenchmark.sliding     10  thrpt       334,255          ops/s
[info] StreamBenchmark.sliding    100  thrpt       353,308          ops/s
[info] StreamBenchmark.sliding   1000  thrpt       332,867          ops/s
[info] StreamBenchmark.sliding  10000  thrpt       669,034          ops/s
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2384,27 +2386,79 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
-    this.pull
-      .unconsN(n, true)
-      .flatMap {
-        case None => Pull.done
-        case Some((hd, tl)) =>
-          val window = hd.foldLeft(collection.immutable.Queue.empty[O])(_.enqueue(_))
-          Pull.output1(window) >> go(window, tl)
-      }
-      .stream
+  def sliding(n: Int): Stream[F, collection.immutable.Queue[O]] = sliding(n, 1)
```
mpilquist, line 2389: Hm, should this return `Chunk.Queue` instead of `c.i.Queue`? Performance should be better in general that way...

nikiforo: Yep, I'll change it and measure the perfomance difference.

But, this how it won't be backward compatible.

nikiforo:
```
2de178db
[info] Benchmark                  (n)   Mode  Cnt    Score   Error  Units
[info] StreamBenchmark.sliding     10  thrpt       104,814          ops/s
[info] StreamBenchmark.sliding    100  thrpt       105,679          ops/s
[info] StreamBenchmark.sliding   1000  thrpt       104,443          ops/s
[info] StreamBenchmark.sliding  10000  thrpt       248,643          ops/s

f9a3896a
[info] Benchmark                  (n)   Mode  Cnt    Score   Error  Units
[info] StreamBenchmark.sliding     10  thrpt       258,618          ops/s
[info] StreamBenchmark.sliding    100  thrpt       265,793          ops/s
[info] StreamBenchmark.sliding   1000  thrpt       254,515          ops/s
[info] StreamBenchmark.sliding  10000  thrpt       576,929          ops/s

d2ca1985
[info] StreamBenchmark.sliding     10  thrpt       334,255          ops/s
[info] StreamBenchmark.sliding    100  thrpt       353,308          ops/s
[info] StreamBenchmark.sliding   1000  thrpt       332,867          ops/s
[info] StreamBenchmark.sliding  10000  thrpt       669,034          ops/s

94a7a5ba
[info] StreamBenchmark.sliding     10  thrpt    5   746,362 ± 74,009  ops/s
[info] StreamBenchmark.sliding    100  thrpt    5   758,551 ± 27,865  ops/s
[info] StreamBenchmark.sliding   1000  thrpt    5   749,664 ±  3,378  ops/s
[info] StreamBenchmark.sliding  10000  thrpt    5  1600,152 ± 10,578  ops/s
```

Wow, Chunk.Queue is great :fire:

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -573,7 +573,8 @@ object Chunk extends CollectorK[Chunk] with ChunkCompanionPlatform {
       if (i == 0) value else throw new IndexOutOfBoundsException()
     def copyToArray[O2 >: O](xs: Array[O2], start: Int): Unit = xs(start) = value
     protected def splitAtChunk_(n: Int): (Chunk[O], Chunk[O]) =
-      sys.error("impossible")
+      if (n == 0) (Chunk.empty, this)
```
nikiforo, line 576: I think this is correct

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -573,7 +573,9 @@ object Chunk extends CollectorK[Chunk] with ChunkCompanionPlatform {
...
```
mpilquist, line 576: Hm, why do we need this? `splitAtChunk_` is only called by the `splitAtChunk` method from the super type, which is defined as:

```scala
  def splitAt(n: Int): (Chunk[O], Chunk[O]) =
    if (n <= 0) (Chunk.empty, this)
    else if (n >= size) (this, Chunk.empty)
    else splitAtChunk_(n)
```

So it appears that these new guards will never be true.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2380,31 +2382,83 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+        window: Chunk[O],
+        prev: Chunk[O]
+    ): Pull[F, Chunk[O], Unit] =
+      s.pull.uncons
+        .flatMap {
+          case None =>
+            if (prev.isEmpty) Pull.done
+            else Pull.output1((window ++ prev).take(size))
+          case Some((hd, tl)) =>
+            val buffer = ArrayBuffer.empty[Chunk[O]]
```
mpilquist, line 2436: I'm curious if here and above where `ArrayBuffer` is used if using `Chunk.Queue` would be better. Initialize buffer to an empty chunk.queue and append via `:+`

nikiforo: I'm not sure, whether I’ve understood you correctly, but I’ve changed this and the part above to:
```
case Some((hd, tl)) =>
            var buffer = Chunk.Queue.empty[Chunk[O]]
            var w = window
            var (heads, tails) = (prev ++ hd).splitAt(step)
            while (tails.nonEmpty) {
              val wind = w ++ heads.take(step)
              buffer :+= Chunk.singleton(wind)
              w = wind.drop(step)
              heads = tails.take(step)
              tails = tails.drop(step)
            }

            Pull.output(buffer) >> stepSmallerThanSize(tl, w, heads)

```

I think, that Chunk.Queue’s job is to flatCombine Chunks effectively thus preserving chunk structure as close as possible. Here the task of buffer is different - extract distinct elements from input into a sequence of distinct elements(no flatten action)
Moreover, it has worser performance characteristics:
```
[info] StreamBenchmark.sliding     10  thrpt    5   650,337 ± 48,467  ops/s
[info] StreamBenchmark.sliding    100  thrpt    5   593,517 ± 26,828  ops/s
[info] StreamBenchmark.sliding   1000  thrpt    5   596,785 ± 24,417  ops/s
[info] StreamBenchmark.sliding  10000  thrpt    5  1370,548 ± 50,358  ops/s
```

core/shared/src/main/scala/fs2/Chunk.scala
nikiforo: Returned to the previous implementation

PR #2231, opened 2021-01-29
make readOutputStream use chunkSize consistently

bpholt:
I noticed that regardless of the `chunkSize` passed into `readOutputStream`, the resulting stream has 1024-byte chunks, due to the default buffer size in `PipedInputStream`. Passing the chunk size into the `PipedInputStream` constructor syncs the two up.


bpholt:
Ugh well I thought renaming the branch would just retarget the PR. I'll open a new one, sorry

bpholt:
https://github.com/typelevel/fs2/pull/2234

Issue #2213, opened 2021-01-17
Simplify and improve documentation on Queue


LLCampos:
I'm trying to learn to use `Queue` and struggling even to have a simple hello-wordish example working (add stuff to the queue, take it out and print it).

I find the examples in [here](https://fs2.io/concurrency-primitives.html) and [here](https://fs2.io/guide.html) to be way too complex for a introduction on `Queue`. It would be nice to have some really simple example, accompanied with a detailed example of what's happening.

LLCampos:
Here's my proposal (inspired by this [SO answer](https://stackoverflow.com/a/65762139/5082718)):

```scala
import cats.effect.{ConcurrentEffect, ContextShift, IO, Timer}
import fs2.Stream
import fs2.concurrent.{NoneTerminatedQueue, Queue}

import scala.concurrent.ExecutionContextExecutor

implicit val ec: ExecutionContextExecutor = scala.concurrent.ExecutionContext.global
implicit val cs: ContextShift[IO] = IO.contextShift(ec)
implicit val c: ConcurrentEffect[IO] = IO.ioConcurrentEffect
implicit val t: Timer[IO] = IO.timer(ec)

def enqueueValues(queue: NoneTerminatedQueue[IO, Int]) = {
  // Have to wrap stream's values in Option because of the type of queue we're using
  val stream = Stream(1,2,3).map(Some(_))
  // Enqueues each value in stream
  val enqueuingStream = stream.through(queue.enqueue)
  // Terminate stream
  enqueuingStream ++ Stream.eval(queue.enqueue1(None))
}

val program = for {
  queue <- Queue.noneTerminated[IO, Int] // Queue that terminates when None is enqueued
  _ <- enqueueValues(queue).compile.drain
  result <- queue.dequeue.compile.toList
} yield result

program.unsafeRunSync()
```

mpilquist:
Note that `Queue` has moved to cats-effect (in the `cats.effect.std` package) in the forthcoming CE3 / FS2 3 releases.

LLCampos:
I think we can close this one down. then. ;)

PR #2237, opened 2021-01-30
Simpler examples of concurrency primitives

LLCampos:
~~Didn't add an example for Queue because [it is being moved to cats-effect](https://github.com/typelevel/fs2/issues/2213). :)~~

### DIFF
--- a/site/concurrency-primitives.md
+++ b/site/concurrency-primitives.md
@@ -6,7 +6,35 @@ In the [`fs2.concurrent` package](https://github.com/functional-streams-for-scal
 - `Topic[F, A]`
 - `Signal[F, A]`
 
-These data structures could be very handy in a few cases. See below examples of each of them originally taken from [gvolpe examples](https://github.com/gvolpe/advanced-http4s/tree/master/src/main/scala/com/github/gvolpe/fs2):
+## Simple Examples
+
+### Topic
+
+(based on [Pera Villega](https://perevillega.com/)'s example [here](https://underscore.io/blog/posts/2018/03/20/fs2.html))
+
+Topic implements the publish-subscribe pattern. In the following example, `publisher` and `subscriber` are started concurrently with the publisher continuously publishing the string `"1"` to the topic and the subscriber consuming it until it received four elements, including the initial element, the string `"Topic start"`.
+
+```scala
+import cats.effect.{ConcurrentEffect, ContextShift, IO}
+import fs2.Stream
+import fs2.concurrent.Topic
+
+import scala.concurrent.ExecutionContextExecutor
+
+implicit val ec: ExecutionContextExecutor = scala.concurrent.ExecutionContext.global
+implicit val cs: ContextShift[IO] = IO.contextShift(ec)
+implicit val c: ConcurrentEffect[IO] = IO.ioConcurrentEffect
+
+Topic("Topic start").flatMap { topic =>
+  val publisher = Stream.emit("1").repeat.covary[IO].through(topic.publish)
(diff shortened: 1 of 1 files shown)


site/concurrency-primitives.md
```diff
@@ -6,7 +6,61 @@ In the [`fs2.concurrent` package](https://github.com/functional-streams-for-scal
...
-These data structures could be very handy in a few cases. See below examples of each of them originally taken from [gvolpe examples](https://github.com/gvolpe/advanced-http4s/tree/master/src/main/scala/com/github/gvolpe/fs2):
+## Simple Examples
+
+### Topic
+
+(based on [Pera Villega](https://perevillega.com/)'s example [here](https://underscore.io/blog/posts/2018/03/20/fs2.html))
+
+Topic implements the publish-subscribe pattern. In the following example, `publisher` and `subscriber` are started concurrently with the publisher continuously publishing the string `"1"` to the topic and the subscriber consuming it until it received four elements, including the initial element, the string `"Topic start"`.
+
+```scala
```
mpilquist, line 17: If you change this to `scala mdoc`, the example will get compiled/tested as part of site generation. In this case, it'll show that the example is using the old CE2 / FS2 2.x API and hence needs to be updated for CE3 / FS2 3.x.

site/concurrency-primitives.md
```diff
@@ -6,7 +6,61 @@ In the [`fs2.concurrent` package](https://github.com/functional-streams-for-scal
...
+}.unsafeRunSync()
+```
+
+### Signal
+
+`Signal` can be used as a means of communication between two different streams. In the following example `s1` is a stream that emits the current time every second and is interrupted when the signal has value of `true`. `s2` is a stream that sleeps for four seconds before assigning `true` to the signal. This leads to `s1` being interrupted.
+
+(Note that `SignallingRef` extends `Signal`)
+
+```scala
```
mpilquist, line 41: Same as above, needs updates for CE3

mpilquist (changes requested):
Thanks for working on docs! Added a couple comments about how we're targeting CE3/FS2 3.x now.

We could add docs on `Queue` too, given that FS2 has some methods for working with `cats.effect.std.Queue` -- e.g., `Stream.fromQueueNoneTerminated` being the most common.

diesalbla (approved):
Thanks!

LLCampos:
@mpilquist, adapted things for CE3/FS2 3.x. Should I also change the branch this is being merge into? :thinking:

LLCampos:
Hm... CI workflow failed... is this expected?

site/concurrency-primitives.md
```diff
@@ -14,18 +15,15 @@ In the [`fs2.concurrent` package](https://github.com/functional-streams-for-scal
...
+import cats.effect.unsafe.implicits.global
 import fs2.Stream
 import fs2.concurrent.Topic
 
-import scala.concurrent.ExecutionContextExecutor
-
-implicit val ec: ExecutionContextExecutor = scala.concurrent.ExecutionContext.global
-implicit val cs: ContextShift[IO] = IO.contextShift(ec)
-implicit val c: ConcurrentEffect[IO] = IO.ioConcurrentEffect
+val topic: IO[Topic[IO, String]] = Topic("Topic start")(Concurrent[IO])
```
ritschwumm, line 24: imho typeclass instances should never be passed around manually, except as a workaround where more than one instance exists. especially not in example code that will be widely copied.
if it works, i'd suggesst writing `Topic[IO,String](...)` instead.

PR #2240, opened 2021-01-30
Stream / Pull: bypass some uses of `.pull.uncons`.

diesalbla:
This is an attempt to move some of the code that works by `uncons`ing
a stream, to be closer to methods of Pull.

- Add method `cons` to pull, to prefix pull with a chunk.
- Add method `unconsNonEmptyChunk` in Pull.
- Replace uses of `.pull.uncons`, which includes spurious uses of `Pull.map` (just to do noop conversions), by Pull.uncons.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -578,10 +578,23 @@ object Pull extends PullLowPriority {
       haltOnSignal: F[Either[Throwable, Unit]]
   ): Pull[F, O, Unit] = InterruptWhen(haltOnSignal)
 
+  private[fs2] def cons[F[_], O](c: Chunk[O], p: Pull[F, O, Unit]): Pull[F, O, Unit] =
+    if (c.isEmpty) p else Output(c) >> p
+
   private[fs2] def uncons[F[_], O](
       s: Pull[F, O, Unit]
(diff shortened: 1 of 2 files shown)


mpilquist:
I think `unconsNonEmptyChunk` is redundant with `uncons` -- we used to have something like that but the interpreter was changed such that it's impossible to get an empty chunk from `uncons` (IIRC).

I'm not really keen on `Pull.uncons(s)` -- I don't like the built-in operations having more power than stuff users can write, and I'm not sure that exposing `Pull.uncons(s)` publicly is worth the API change. For this type of change, I think we'd need some benchmarks showing it pays off.

diesalbla:
> I think unconsNonEmptyChunk is redundant with uncons -- we used to have something like that but the interpreter was changed such that it's impossible to get an empty chunk from uncons (IIRC).

Hmmm Should we add an assertion to the `Output` constructor, to ensure that any chunk inserted into it is non-empty?

mpilquist:
Sure, I'm pretty sure that's safe, b/c the only way to construct an `Output` checks for empty:

```scala
  def output[F[x] >: Pure[x], O](os: Chunk[O]): Pull[F, O, Unit] =
    if (os.isEmpty) Pull.done else Output[O](os)
```

diesalbla:
ok... incoming PR...

PR #2241, opened 2021-01-30
Change: all Output operators are non-empty.

diesalbla:
Currently, the constructors of a Stream (or Pull) ensure that every Chunk in a Stream, that is every `Output` node in a pull, must have a non-empty Chunk. Therefore, where we uncons we always stop at a non-empty chunk-stream, so we can omit the "nonEmpty" methods..


diesalbla:
@mpilquist I just noticed that this was targeted to the `main` branch. Is that the 3.0 branch now, rather than the 2.5.x branch? If so, could you delete the `develop` branch, to avoid confusion?

mpilquist:
@diesalbla Yep, fixed. I tried to do it earlier but had to change all the open PRs to point to main.

PR #2248, opened 2021-02-01
Stream - BufferBy - use internal "dumpBuffer" method.

diesalbla:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -321,6 +321,9 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * }}}
     */
   def bufferBy(f: O => Boolean): Stream[F, O] = {
+    def dumpBuffer(bb: List[Chunk[O]]): Pull[F, O, Unit] =
```
mpilquist, line 324: Unrelated to this PR, but curious why we use `List[Chunk[O]]` here instead of `Chunk.Queue[O]`

PR #2251, opened 2021-02-06
Pull 'onCons' function

diesalbla:
Looking through the code of `Stream`, we noticed that the `uncons` method is used to implement a lot of the Stream combinators. The `uncons` is often followed by that of `Pull.flatMap`. Furthermore, the `uncons` implementation itself includes a `map` operation. 

This commit introduces an `onCons` method, that combines the `uncons` and the `flatMap`, to provide a better combinator.
The `onCons` function takes a function on what to do upon the first "cons", which is to say, on the first Output from the stream.


mpilquist:
I have the same objections to this one as the similar PR from a week ago or so. If we keep it private then we get a small perf benefit that users can’t get when writing their own pulls. If we make it public, we break a ton of code and introduce a new API. Further, we had operations like this in 0.10 or 0.9 — receive and receive1.

diesalbla:
I can see the reference to those `receive` methods here... https://github.com/typelevel/fs2/pull/701 looking for the PR at which they are removed to find the rationale...

diesalbla:
Some of the savings have been done in #2256

PR #2252, opened 2021-02-06
New topic

SystemFw:


core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -89,127 +91,52 @@ abstract class Topic[F[_], A] { self =>
...
-                selector ->
-                  pubSub.getStream(Right(selector)).flatMap {
-                    case Right(q) => Stream.emit(q)
-                    case Left(_)  => Stream.empty // impossible
-                  }
+          def publish1(a: A): F[Unit] =
+            state.get.flatMap { case (subs, _) =>
+              subs.foldLeft(F.unit) { case (op, (_, q)) =>
+                op >> q.offer(a)
               }
```
diesalbla, line 112: I may be approaching this from afar, but is this a case of `traverse_`?

core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -22,34 +22,39 @@
...
 
   /** Publishes one `A` to topic.
     *
-    * This waits until `a` is published to all subscribers.
-    * If any of the subscribers is over the `maxQueued` limit, this will wait to complete until that subscriber processes
-    * enough of its elements such that `a` is enqueued.
+    * If any of the subscribers is over the `maxQueued` limit, this
+    * will wait to complete until that subscriber processes enough of
+    * its elements such that `a` is enqueued.
+    *
```
diesalbla, line 54:
```suggestion
    * This `F` operation deos not complete until after the 
    * given element has been enqued on all subscribers. 
    * If enqueuing on any subscriber is semantically blocked 
    * (because that subscriber is on its `maxQueued` limit), 
    * then this operation will also be blocked, waiting on that  
    * subscriber to dequeue an element.
```

One further clarification: does the whole `Topic` ensure that there is at most one call to `publish1` at a given point?

core/shared/src/main/scala/fs2/concurrent/Topic.scala
SystemFw: it is `traverse_`, but there is no `Traverse[LongMap]`

core/shared/src/main/scala/fs2/concurrent/Topic.scala
SystemFw: no, you can have multiple producers on a Topic, although I imagine it to be rarer

diesalbla: In that case, is there a risk that different subscribers may receive messages from different producers on a different order?

SystemFw: Yeah, there is (and there's always been). I think fixing it would imply either enforcing single producer, if you still want full broadcast, or a KeyedTopic to do sharding a la Kafka

mpilquist (approved):

core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -89,127 +91,52 @@ abstract class Topic[F[_], A] { self =>
...
-    */
-  def apply[F[_], A](initial: A)(implicit F: Concurrent[F]): F[Topic[F, A]] = {
-    implicit def eqInstance: Eq[Strategy.State[A]] =
-      Eq.instance[Strategy.State[A]](_.subscribers.keySet == _.subscribers.keySet)
-
-    PubSub(PubSub.Strategy.Inspectable.strategy(Strategy.boundedSubscribers(initial)))
-      .map { pubSub =>
+  /** Constructs a Topic */
+  def apply[F[_], A](implicit F: Concurrent[F]): F[Topic[F, A]] =
+    F.ref(LongMap.empty[Q[F, A]] -> 1L)
```
SystemFw, line 104: @mpilquist I don't think we will have issues here, Long is very big (even though we only use half of them), but I'd welcome a second opinion.

mpilquist: No concerns from me

core/shared/src/main/scala/fs2/concurrent/Topic.scala
SystemFw: @mpilquist wdyt about this? Should we enforce a single producer, or just document it and it's up to the user?

mpilquist:
Document it only, we haven’t seen this cause issues over last 5 years

PR #2256, opened 2021-02-09
Pull: split Step operation in two constructors.

diesalbla:
The `Step` Pull constructor is used to turn a closed pull (a Stream) into a pull that stops at the first `Output` node. It had an optional scopeId parameter, to indicate if we shift to a scope or not.
- The `None` case is used for the `uncons` operation, which maps the return triplet to the tuple and ignores the second element. 
- The `Some` case is used for the `stepLeg`, which maps the triplet to a `StepLeg`.

Since these are different usages with different data and different end return type, it may be better to split the `Step` into two cases `StepCons` and `StepLeg`.

- StepCons is a step without shifting scopes, and it returns just the (option of) tuple of Chunk and tail pull.
- StepLeg is a step that is run in a separate scope, shifting there and back again. Upon reaching an `Output`, it gives back all three elements in a `StepLeg` object.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -515,16 +515,23 @@ object Pull extends PullLowPriority {
       fun: O => Pull[F, P, Unit]
   ) extends Action[F, P, Unit]
 
-  /** Steps through the stream, providing either `uncons` or `stepLeg`.
+  /** Steps through the given inner stream, until the first `Output` is reached.
+    * It returns the possible `uncons`.
     * Yields to head in form of chunk, then id of the scope that was active after step evaluated and tail of the `stream`.
     *
     * @param stream             Stream to step
-    * @param scopeId            If scope has to be changed before this step is evaluated, id of the scope must be supplied
     */
-  private final case class Step[+F[_], X](stream: Pull[F, X, Unit], scope: Option[Unique])
-      extends Action[Pure, INothing, Option[StepStop[F, X]]]
+  private final case class StepCons[+F[_], +O](stream: Pull[F, O, Unit])
+      extends Action[Pure, INothing, Option[(Chunk[O], Pull[F, O, Unit])]]
 
-  private type StepStop[+F[_], +X] = (Chunk[X], Unique, Pull[F, X, Unit])
+  /** Steps through the stream, providing a `stepLeg`.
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -577,15 +582,15 @@ object Pull extends PullLowPriority {
       haltOnSignal: F[Either[Throwable, Unit]]
   ): Pull[F, O, Unit] = InterruptWhen(haltOnSignal)
 
+  private type Cons[+F[_], +O] = (Chunk[O], Pull[F, O, Unit])
```
mpilquist, line 585: Can we remove this? I find type aliases very confusing in general. They introduce names without introducing any type safety. I find that to be a bad tradeoff.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -515,16 +515,23 @@ object Pull extends PullLowPriority {
...
+  /** Steps through the given inner stream, until the first `Output` is reached.
+    * It returns the possible `uncons`.
     * Yields to head in form of chunk, then id of the scope that was active after step evaluated and tail of the `stream`.
     *
     * @param stream             Stream to step
-    * @param scopeId            If scope has to be changed before this step is evaluated, id of the scope must be supplied
     */
-  private final case class Step[+F[_], X](stream: Pull[F, X, Unit], scope: Option[Unique])
-      extends Action[Pure, INothing, Option[StepStop[F, X]]]
+  private final case class StepCons[+F[_], +O](stream: Pull[F, O, Unit])
```
mpilquist, line 524: Can we call this `Uncons`?

mpilquist (changes requested):
Looks good to me. I don't have strong opinions about whether these are a single subtype of Pull or 2 subtypes of Pull.

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: The idea occurred to me... was not sure if that would be too different from existing one.

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: I use type aliae, mostly to have type signatures shorter and avoid deep nesting.
Just as one would use a local `val` to capture an expression without having to write a lot of nested parens.
Compilers may be good at LISP, but humans need line breaks and short sentences. At least I do :)

PR #2262, opened 2021-02-11
Pull: split compilation loop apart.

diesalbla:
We remove from the `Pull` object the long parts dedicated to the compilation of that Pull, to a separate file in `internals`. This is inspired on the design of `cats.effect.IO`, which has the `IORunLoop` separated to the internals.

- The case classes of `Pull` are no longer sealed, but are now restricted to `fs2`.


mpilquist:
I'd prefer to keep this stuff in `Pull`. BTW, cats-effect 3.x is organized in a much different fashion than 2.x. In 3.x, the main interpreter is in `IOFiber.scala`.

PR #2264, opened 2021-02-11
Add Stream#ifEmpty

luis3m:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2584,6 +2584,19 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+   * @example {{{
+   * scala> Stream.empty.unconsNonEmpty(0).toList
+   * res0: List[Int] = List(0)
+   * }}}
+   */
+  def unconsNonEmpty[O2 >: O](o: => O2): Stream[F, O2] =
+    this.pull.uncons1.flatMap {
+      case Some((hd, tl)) => Pull.output1(hd) >> tl.underlying
+      case None           => Pull.output1(o) >> Pull.done
+    }.stream
```
diesalbla, line 2598: So, using `uncons1` involves extracting and partitioning chunks. We can just use `uncons` for this: 
```suggestion
    this.pull.uncons.flatMap {
      case Some((hd, tl)) => Pull.output(hd) >> tl.underlying
      case None           => Pull.output1(o) >> Pull.done
    }.stream
```
Note that `Pull` is guaranteed to only contain non-empty chunks.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2584,6 +2584,19 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
     }
 
+  /** Outputs a singleton pure stream if the stream is empty.
+   *
+   * @example {{{
+   * scala> Stream.empty.unconsNonEmpty(0).toList
+   * res0: List[Int] = List(0)
+   * }}}
+   */
+  def unconsNonEmpty[O2 >: O](o: => O2): Stream[F, O2] =
```
diesalbla, line 2594: So, `unconsNonEmpty` is a bit confusing. Perhaps `ifEmptyThen` would be more clear: 
If `this` is empty, then we add this single item.

It may be worth adding two variants: `ifEmptyThen`, that takes a `Chunk[O2]`, and rename this one as `ifEmptyThen1`.

core/shared/src/main/scala/fs2/Stream.scala
luis3m: @diesalbla good point, thanks. Done.

core/shared/src/main/scala/fs2/Stream.scala
luis3m: @diesalbla I took @mpilquist name suggestions. However, I'm not sure how to call a chunk variant. `ifEmptyChunk`?

diesalbla: So, for many methods we use the simple wor for passing a chunk, and the word ended in `1` to indicate a single element.

PR #2268, opened 2021-02-13
Expand documentation of the Pull

diesalbla:


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
 import Resource.ExitCase
 import Pull._
 
-/** A `p: Pull[F,O,R]` reads values from one or more streams, returns a
-  * result of type `R`, and produces a `Stream[F,O]` when calling `p.stream`.
+/** A pure abstraction representing the intention to perform a sequence of actions
```
mpilquist, line 36: "A process which may evaluate actions in an effect type `F`, emit any number of output values of type `O`, and terminate with a single result of type `R`."

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * in an effect type `F`, which emits any number of values of an output type `O`.
+  * This sequence may end up in several ways:
```
mpilquist, line 38: This feels a bit too detailed for the introduction -- feels like it's internal docs on the interpreter. Suggest something like: 

```
There are various ways in which a pull may terminate abnormally:
 - exceptions that occur during evaluation of a pull result
 - requested interruption of execution
```

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * a. it may not terminate and keep unrolling new items;
+  * b. it may be aborted by a `Throwable` error;
+  * c. be interrupted by a signal, which may include an error,
+  * d. terminate successfully and returning one single carry-over result of type `R`.
   *
-  * Any resources acquired by `p` are freed following the call to `stream`.
+  * Much like the `IO` values fromo the `cats-effect` library,
```
mpilquist, line 44: s/fromo/from/

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * `Pull` values are pure, immutable values. Thus, they preserve referential
+  * transparency and are usable in functional programming.
+  * A `Pull` is a data structure that represents just a description
```
mpilquist, line 47: Strike -- this is redundant with the first sentence

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * of a sequence of potentially side effectful computation.
   *
-  * Laws:
+  * A pull emits the output values not one-at-a-time, but one-chunk at a time.
```
mpilquist, line 50: Suggest: "For efficiency, output values are processed in chunks. A chunk is an immutable sequence with O(1) indexed lookup. A `Pull[F, Byte, R]` operates internally on `Chunk[Byte]` values, which can wrap unboxed byte arrays -- avoiding significant boxing/unboxing costs. The `Pull` API provides mechanisms for working at both the chunk level and the individual element level. Generally, working at the chunk level will result in better performance but at the cost of more complex implementations."

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * There are some special types of Pull, which are reflected on the arguments of its type parameters:
+  *
+  * A '''pure''' pull is one that performs no effectful actions.
+  * The type of a pure pull has the `F` effect type  bound to `Nothing`.
+  * _Note_ that type `Nothing` is polykinded type, so it can also be
+  * applied as an argument to the type parameter `F[_]`.
+  * Nevertheless, for the sake of clarity, in `fs2` we use the type alias
+  * `type Pure[A] = Nothing` to represent a pure pull as `Pull[Pure, O, R]`.
+  *
+  * A '''silent''' pull is one known to emit no outputs. In a silent pull, the
```
mpilquist, line 64: Not sure I want to introduce this silent, broken, and closed terminology -- three new concepts which we don't currently use anywhere else.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -48,20 +123,85 @@ import Pull._
...
+    *
+    * The composed pull will first emit all the outputs emitted by `this` pull,
+    * and if successful will start emitting the outputs from the generated pull.
+    *
+    * This operation does not modify resource scope boundaries. The generated `post` 
+    * pull starts compilation on the same scope on which the compilation of this 
+    * pull (the prefix) ends. The composed pull will end on whatever scope the 
+    * compilation of the `post` pull does.
+    */
+  def flatMap[F2[x] >: F[x], O2 >: O, R2](kont: R => Pull[F2, O2, R2]): Pull[F2, O2, R2] =
```
mpilquist, line 152: 👎 On renaming the function `kont` -- let's stick with `f`.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -48,20 +123,85 @@ import Pull._
...
+    * In both cases, the effectful actions and outputs of the appended pull 
+    * consists of those outputs and actions from the first pull, followed by
+    * those from the `post` pull, in the same order as they would come out of each pull.
+    *
+    * This operation does not add or remove any resource scope boundaries. 
+    * The `post` pull will run on the same scope on which the compilation of 
+    * this pull ends. The composed pull will end on whatever scope the `post` 
+    * pull ended on.
+    */
+  def >>[F2[x] >: F[x], O2 >: O, S](post: => Pull[F2, O2, S]): Pull[F2, O2, S] =
```
mpilquist, line 187: :-1: on the `post` rename here -- should just be `f`

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * in an effect type `F`, which emits any number of values of an output type `O`.
+  * This sequence may end up in several ways:
+  * a. it may not terminate and keep unrolling new items;
+  * b. it may be aborted by a `Throwable` error;
+  * c. be interrupted by a signal, which may include an error,
+  * d. terminate successfully and returning one single carry-over result of type `R`.
   *
-  * Any resources acquired by `p` are freed following the call to `stream`.
+  * Much like the `IO` values fromo the `cats-effect` library,
+  * `Pull` values are pure, immutable values. Thus, they preserve referential
```
mpilquist, line 45: I'd strike the "Thus, they ..." sentence

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  *
+  * === Pulls variations
+  *
+  * There are some special types of Pull, which are reflected on the arguments of its type parameters:
+  *
+  * A '''pure''' pull is one that performs no effectful actions.
+  * The type of a pure pull has the `F` effect type  bound to `Nothing`.
+  * _Note_ that type `Nothing` is polykinded type, so it can also be
+  * applied as an argument to the type parameter `F[_]`.
+  * Nevertheless, for the sake of clarity, in `fs2` we use the type alias
```
mpilquist, line 61: We don't necessarily need to mention it here but `Pure` really exists to workaround Scala 2 type inference limitations. When we eventually drop Scala 2, I'd imagine we'll drop `Pure` too.

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: IMHO we should start adding more informative names for the methods, so as to make the code the first documentation. https://hilton.org.uk/blog/naming-smells

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: Would it still be helpful, to use `Pure` as an alias of "no effects"?

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: The `f` name in particular seems to be misleading, seems it is often used for functions and this parameter isn't.

mpilquist: Good point about this not being a function -- I'd call it `that`. In general, I'm strongly against trying to create evocative names for function arguments.

diesalbla: Can you ellaborate on that preference, or what downsides have you found for "informative" parameter names? 
That is something often recommended by that speaker ([saw a talk at past Scala eXchange](https://skillsmatter.com/skillscasts/11207-how-to-name-things-the-hardest-problem-in-programming)) and other authors such as [Dave Fairley](https://www.youtube.com/watch?v=1KeJc6V4Jjk.)

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: There are new concepts/words. The reason is, a lot of places had the words "that returns unit" or "without effects", so I thought that a quick word to indicate that would be helpful.

It also helps later on the method or constructors below, like "fail" or "pure".

core/shared/src/main/scala/fs2/Pull.scala
mpilquist: Some miscellaneous conventions:
- `f`, `g`, `h` for function/method parameters of type `FunctionX[...]`
- `A`, `B`, `C` for proper type params
- `F`, `G`, `H` for higher kinded type params
- `A2`, `B2` for type params used for variance reasons (e.g. `def foo[A2 >: A](...)`)

I find small functions, especially when using parametric polymorphism, to be easier to read with such identifiers. There's some commentary about this in FPiS. Considering `def map[B](f: A => B): F[B]` versus `def map[B](convert: A => B): F[B]`, I find the second both harder to write, as we need to come up with a meaningful name for the function, and harder to read, as the `convert` name doesn't stand out as much in the function body.

Consistency is very important too. In general, the codebase follows the conventions laid out above (which we should document somewhere!). Adopting a new style incrementally is jarring.

core/shared/src/main/scala/fs2/Pull.scala
mpilquist: You mean when we get to the point where we can drop Scala 2? I'd rather not but I wouldn't want to break everyone's code over the removal of a type alias. That decision is years and years away though, as I imagine we'll be supporting 2.13 for a very long time.

core/shared/src/main/scala/fs2/Pull.scala
mpilquist: Yeah, I would much prefer we just let the types do the talking here. I find these new concepts *harder* to understand than just looking at the types. For example:
- Why is a pull which has the output type = `Nothing` considered silent but a pull which has `O != Nothing` and yet outputs no elements not silent?
- If a pull with `O = Nothing` is silent, are pulls with `O != Nothing` audible? Loud?
- Are broken pipes fixable?
- Are pipes with `R != Unit` open?
I find these analogies more confusing than helpful.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * that fails or is known not to terminate, such as `IO.never` from `cats.effect`.
+  *
+  * A '''closed''' pull is one that, if may terminate successfully with no
+  * carry-over information. The type of a closed pull has the type parameter `R`
+  * bound to `Unit`. A `Stream[F, O]` is just a wrapper  around a closed pull.
+  *
+  * === Evaluation and Compilation ===
+  *
+  * Like functional effect types (e.g. `IO` in Cats Effect),   a pull is just
+  * a passive data structure that describes a sequence of action.
```
mpilquist, line 82: What's a "passive" data structure? Do we need to reiterate that pulls are descriptions of computation here again? We have it in the intro.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * It is neither a running process nor a container to be eventually filled
+  * with the  result of a concurrent process (like `scala.concurrent.Future`).
+  *
+  * To "run" or "interpret" a pull, it has to be compiled.
```
mpilquist, line 86: From an API perspective, pulls don't get compiled, only streams do. This doc feels more like internal documentation as to how the pull interpreter works. Not sure if we need this stuff here or if it'd be better to just talk about converting pulls to streams and then compiling. Stuff like how pulls are implemented via a monadic fold are irrelevant to users IMO.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  *
+  * Compilation is pull-based, rather than push-based (hence the name of the datatype).
+  * It is the compilation process itself, that determines when the compilation
+  * of each single effect can proceed or is hold back. Effects and outputs later
+  * in the pull are not performed or emitted, _unless and until_ the compilation
+  * calls for them.
+  *
+  * === Resource scoping ===
+  *
+  * The side effects performed by a pull may operate on resources. These resources
```
mpilquist, line 106: s/side effects/effects/

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * need to be 1. retained during the execution of the pull (compilation),
+  * 2. shared by several pulls that we may want to unroll and evaluate independently,
+  * and 3. cleaned up once they are no longer needed.
+  * To keep track of resources, pull compilation uses resource scopes: collection
```
mpilquist, line 110: I don't think we should mention compilation here. Scopes are part of the overall API and don't need to be described in terms of how they are used during compilation.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -33,12 +33,87 @@ import fs2.internal._
...
+  * It is the compilation process itself, that determines when the compilation
+  * of each single effect can proceed or is hold back. Effects and outputs later
+  * in the pull are not performed or emitted, _unless and until_ the compilation
+  * calls for them.
+  *
+  * === Resource scoping ===
+  *
+  * The side effects performed by a pull may operate on resources. These resources
+  * need to be 1. retained during the execution of the pull (compilation),
+  * 2. shared by several pulls that we may want to unroll and evaluate independently,
```
mpilquist, line 108: I'd avoid the term "unroll" in all API docs -- that's an implementation detail.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -48,20 +123,85 @@ import Pull._
...
 
-  /** Alias for `_.map(_ => o2)`. */
-  def as[R2](r2: R2): Pull[F, O, R2] = map(_ => r2)
+  /** Applies the result of this pull to `f` and returns the result.
+    *
+    * This method returns a new composed pull. The semantics of this pull
+    *
+    * - If `this` pull completes successfully with a result of type R, 
+    *   the `f` function is applied to the result to build a new Pull, 
+    *   and compilation continues by running that pull. The end return of
```
mpilquist, line 139: Similar to earlier comment -- no need to mention compilation when describing the API elements.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -48,20 +123,85 @@ import Pull._
...
+    *   the `f` function is applied to the result to build a new Pull, 
+    *   and compilation continues by running that pull. The end return of 
+    *   the composed pull is that of the generated pull.
+    * - If this pull fails or is interrupted, the failure of interruption is 
+    *   the returned of the composed pull.
+    *
+    * The composed pull will first emit all the outputs emitted by `this` pull,
+    * and if successful will start emitting the outputs from the generated pull.
+    *
+    * This operation does not modify resource scope boundaries. The generated `post`
```
mpilquist, line 147: Scope stuff here is too detailed and not really relevant to users.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -48,20 +123,85 @@ import Pull._
...
-  def as[R2](r2: R2): Pull[F, O, R2] = map(_ => r2)
+  /** Applies the result of this pull to `f` and returns the result.
+    *
+    * This method returns a new composed pull. The semantics of this pull
+    *
+    * - If `this` pull completes successfully with a result of type R, 
+    *   the `f` function is applied to the result to build a new Pull, 
+    *   and compilation continues by running that pull. The end return of 
+    *   the composed pull is that of the generated pull.
+    * - If this pull fails or is interrupted, the failure of interruption is
```
mpilquist, line 141: Reword "the failure of interruption is the returned of the composed pull"

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -48,20 +123,85 @@ import Pull._
...
+            catch { case NonFatal(e) => Fail(e) }
+          case res @ Interrupted(_, _) => res
+          case res @ Fail(_)           => res
+        }
+    }
 
-  /** Returns a pull with the result wrapped in `Right`, or an error wrapped in `Left` if the pull has failed. */
-  def attempt: Pull[F, O, Either[Throwable, R]] =
-    map(r => Right(r)).handleErrorWith(t => Succeeded(Left(t)))
+  /** Lazily appends the given `post` pull, to be run after `this` pull.
```
mpilquist, line 164: I think this documentation should be really lightweight with a simple reference to the `flatMap` definition. Describe this as an alias for flatMap and describe how it differs from `*>` re: laziness and allowing tail recursive pulls.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -48,20 +123,85 @@ import Pull._
...
+          case r: Interrupted  => r
+          case r: Fail         => r
+        }
+    }
 
   /** Interpret this `Pull` to produce a `Stream`, introducing a scope.
     *
-    * May only be called on pulls which return a `Unit` result type. Use `p.void.stream` to explicitly
-    * ignore the result type of the pull.
+    * The transformation to `Stream` wraps the given `this` pull to be run in
```
mpilquist, line 199: I'd remove this paragraph or otherwise reword to "A scope is introduced to ensure any resources acquired in this pull are released in a timely fashion, even when the resulting stream is not run until completion."

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
   }
 
-  /** Applies the resource of this pull to `f` and returns the result. */
-  def flatMap[F2[x] >: F[x], O2 >: O, R2](f: R => Pull[F2, O2, R2]): Pull[F2, O2, R2] =
-    new Bind[F2, O2, R, R2](this) {
-      def cont(e: Terminal[R]): Pull[F2, O2, R2] =
-        e match {
-          case Succeeded(r) =>
-            try f(r)
+  /** Attaches to `this` pull a handler for errors that may be raised from it.
```
mpilquist, line 225: Reword to something like:

```
Allows recovery from an error that occurs during evaluation of this pull.

If an error occurs, the supplied function is called to determine how to continue processing. Note that the original computation cannot be resumed at the point of an error.
```

Then provide an example using `@example`.

I'd avoid describing continuation pulls and compilation semantics.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -48,20 +123,85 @@ import Pull._
   *
   * `raiseError` is caught by `handleErrorWith`:
   *   - `handleErrorWith(raiseError(e))(f) == f(e)`
+  *
+  * @tparam F[_] the type of functional effects that can be performed by this pull.
+  * @tparam O The type of outputs emitted by this Pull.
+  * @tparam R The type of carry-over result that is returned by this Pull, if it returns successfully.
   */
 sealed abstract class Pull[+F[_], +O, +R] {
```
mpilquist, line 131: The methods on `Pull` were roughly in alphabetical order, with a handful of exception cases. This PR reorders them a bit, though I don't understand the scheme.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
+    */
+  def onComplete[G[x] >: F[x], O2 >: O, R2](post: => Pull[G, O2, R2]): Pull[G, O2, R2] =
+    handleErrorWith(e => post >> Fail(e)) >> post
 
-  /** Lifts this pull to the specified effect type. */
-  def covary[F2[x] >: F[x]]: Pull[F2, O, R] = this
+  /** Short-hand for `(this: Pull[G, P, S])`
+    * useful to override compiler type inference.
+    */
+  def covaryAll[G[x] >: F[x], O2 >: O, R2 >: R]: Pull[G, O2, R2] = this
```
mpilquist, line 265: Keep `F2` instead of changing to `G`, update doc to:

```
Short-hand for `(this: Pull[F2, O2, R2])`. Used to assist with type inference.
```

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
 
-  /** Lifts this pull to the specified effect type, output type, and resource type. */
-  def covaryAll[F2[x] >: F[x], O2 >: O, R2 >: R]: Pull[F2, O2, R2] = this
+  /** Short-hand for `(this: Pull[G, O, R])`,
+    * useful to override compiler type inference.
+    */
+  def covary[G[x] >: F[x]]: Pull[G, O, R] = this
```
mpilquist, line 270: F2

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
+    * useful to override compiler type inference.
+    */
+  def covaryResult[R2 >: R]: Pull[F, O, R2] = this
+
+  /** Returns a pull with the result wrapped in `Right`, or an error wrapped in `Left` if the pull has failed. */
+  def attempt: Pull[F, O, Either[Throwable, R]] =
+    map(r => Right(r)).handleErrorWith(t => Succeeded(Left(t)))
 
-  /** Applies the resource of this pull to `f` and returns the result in a new `Pull`. */
+  /** Transforms the carry-over result type of this pull to `f` and returns the result in a new `Pull`.
```
mpilquist, line 286: Like above, avoid mention of compilation. Also, avoid "carry-over" terminology.

Reword to something like:

```
Converts the result of this pull using the supplied function.
```

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
+    */
   def map[R2](f: R => R2): Pull[F, O, R2] =
     new Bind[F, O, R, R2](this) {
       def cont(r: Terminal[R]) = r.map(f)
     }
 
-  /** Run `p2` after `this`, regardless of errors during `this`, then reraise any errors encountered during `this`. */
-  def onComplete[F2[x] >: F[x], O2 >: O, R2 >: R](p2: => Pull[F2, O2, R2]): Pull[F2, O2, R2] =
-    handleErrorWith(e => p2 >> Fail(e)) >> p2
+  /** Discards the result type of this pull.
```
mpilquist, line 297: s/type//

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
+    *
+    * Alias for `this.map(_ => () )`.
+    *
+    * This is equivalent to `Functor[Pull[F, O, *]].void(this)`
```
mpilquist, line 301: Drop the reference to Functor.void

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
-        e match {
-          case Fail(e) =>
-            try h(e)
-            catch { case NonFatal(e) => Fail(e) }
-          case other => other
-        }
-    }
+  /** Alias for `_.map(_ => o2)`.
+    *
+    * This is equivalent to `Functor[Pull[F, O, *]].as(this, constant)`
```
mpilquist, line 307: Drop the reference to Functor.as

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
+    *
+    * @tparam S     The type of the constant,
+    * @param  value the new result value of the pull
+    */
+  def as[S](value: S): Pull[F, O, S] = map(_ => value)
```
mpilquist, line 312: s/value/s/

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
 }
 
 object Pull extends PullLowPriority {
 
+  private[this] val unit: Terminal[Unit] = Succeeded(())
+
+  /** The `done` pull is a pure, closed, silent pull. It succeeds by emitting a `Unit`. */
+  val done: Pull[Pure, INothing, Unit] = unit
+
+  /** Lifts a constant result value into a pure silent Pull that succeesfully
```
mpilquist, line 323: "constant result value" doesn't say much to me. I'd also drop the reference to applicative. I'd rewrite this to:

```
Creates a pull which evaluates no effects, outputs no elements, and terminates with the supplied value.
```

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
+    * returns the given value. This is the `pure` function of the Applicative
+    * instance for Pull.
+    */
+  def pure[F[_], R](r: R): Pull[F, INothing, R] = Succeeded(r)
+
+  /** Lifts the effectful computation `F[R]` into a silent pull in that effect.
```
mpilquist, line 329: "Creates a pull which evaluates the supplied effect and terminates with the result of that evaluation, emitting no output. If the effect fails with an error, the pull fails with that error."

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
+    * @tparam O the type of the output of the generated chunk.
+    * @tparam F the effect type of the pull. This is given as a convenience for
+    *         the API users, much like the type parameter in the `Option.empty`
+    *         method, to avoid closing on `Nothing`. Note, however, that the
+    *         returned pull is _pure_ and performs no effects of any type.
+    */
+  def output1[F[x] >: Pure[x], O](o: O): Pull[F, O, Unit] = Output(Chunk.singleton(o))
+
+  /** Outputs a chunk of values. */
+  def output[O](os: Chunk[O]): Pull[Pure, O, Unit] =
```
mpilquist, line 351: Why did signature of `output` change here? The original had the `F[x] >: Pure[x]` constraint. Seems odd that we'd support that for `output1` and not `output`.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
+  def eval[F[_], R](fr: F[R]): Pull[F, INothing, R] = Eval[F, R](fr)
+
+  /** Lifts the given output value `O` into a pure, closed, pull that emits the
+    * single output.
+    *
+    * _Note_ using singleton chunks is not memory efficient. If possible,
+    * use the chunk-based `output` method instead.
+    *
+    * @tparam O the type of the output of the generated chunk.
+    * @tparam F the effect type of the pull. This is given as a convenience for
```
mpilquist, line 343: I'd remove this or else we'd need to add it to hundreds of methods throughout the library

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -70,83 +210,147 @@ sealed abstract class Pull[+F[_], +O, +R] {
...
+  def pure[F[_], R](r: R): Pull[F, INothing, R] = Succeeded(r)
+
+  /** Lifts the effectful computation `F[R]` into a silent pull in that effect.
+    *
+    * If the `fr` action fails with an error, the lifted pull will fail with
+    * the thrown error.
+    */
+  def eval[F[_], R](fr: F[R]): Pull[F, INothing, R] = Eval[F, R](fr)
+
+  /** Lifts the given output value `O` into a pure, closed, pull that emits the
```
mpilquist, line 336: `Creates a pull which outputs a single value and terminates with a unit, evaluating no effects."

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: Thouhgt about removing those. IMHO, since higher-kinded type parameters are always a bit of a problem, removing them if they are not needed would improve the usability of the endpoint.

mpilquist: If we can do it without any inference issues, I'd be fine with it, though I think we should be consistent. E.g., if we can remove the param and replace with Pure in both output/output1 and everything complies for 2.12/2.13/3, then I'm on board.

SystemFw: aren't they needed to avoid `covary`?

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: Ok, I have removed these words and added some notes to the type parameter descriptions.

diesalbla (commented):
Addressed some comments...

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: Saying that it is for efficiency suggests that it is a merely operational concern, to be optimised later. 

However, as has been pointed elsewhere, Chunks are also boundaries of laziness for interruptions and terminations, so they are part of the observable "value-level", correctness.

Nevertheless, it is a paragraph to add.

core/shared/src/main/scala/fs2/Pull.scala
diesalbla: The scheme I followed was to group the functor-level methods, `map`, `as`, and `void`, and to bring forward the monadic errors `flatMap` and `>>`, which are the most heavily used ones.

mpilquist: OK, I don't feel strongly about this as the Pull API is pretty small. Maybe we could add some comment fencing like `// === Functor operations ===`, not sure.

core/shared/src/main/scala/fs2/Pull.scala
mpilquist: OK, I don't feel strongly. Add a mention that this is the same as `flatMap(_ => that)` though.

diesalbla:
@mpilquist One note: if you want to merge, when you do, could you use the "Squash & Merge" option?

mpilquist:
@diesalbla I squash merged but note we lost the stream/streamNoScope as implicit class change in the merge. See my comment on #2310 and let me know if you want to restore that implicit class versus using regular methods with evidence params.

diesalbla:
@mpilquist IMHO evidence params are a most complicated notion that extension methods, a la implicit class. Considering that those are to be directly supported in dotty, it may be worth restoring that.

mpilquist:
@diesalbla OK, I just restored it.

PR #2271, opened 2021-02-13
Pull: use carry-over nomenclature 

diesalbla:
Thinking about the right name for the 3rd pull parameter, the one used to chain elements in `Pull.flatMap`. We currently use the `R` letter for "return" or "result" or "resource", but the first two are overloaded. "Resource" is even worst, since that has a rather specific meaning in `cats-effect`, and appears in other places of `fs2` itself to denote other things. ;

A better name may be "Carry": much like the carry-over operand of an addition, which is one value from previous computation passed over to keep computing the rest of the additions, a Pull leaves a carry-over result to be passed to build next oncoming pulls. Using a separate letter (from `O`) allows us to use `D` for the alternative carry-over parameter, rather than `R2`. 

_Note_ this should be regarded as a documentation PR. The first documentation should be the code itself. Clearer, distinctive names for value and type parameters is worth phrases of scaladocs, and more likely to ever be read.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -33,14 +33,14 @@ import fs2.internal._
 import Resource.ExitCase
 import Pull._
 
-/** A `p: Pull[F,O,R]` reads values from one or more streams, returns a
-  * result of type `R`, and produces a `Stream[F,O]` when calling `p.stream`.
+/** A `p: Pull[F,O,C]` reads values from one or more streams, returns a
+  * carry-over value of type `C`, and produces a `Stream[F,O]` when calling `p.stream`.
   *
   * Any resources acquired by `p` are freed following the call to `stream`.
   *
   * Laws:
   *
-  * `Pull` forms a monad in `R` with `pure` and `flatMap`:
+  * `Pull` forms a monad in `C` with `pure` and `flatMap`:
   *   - `pure >=> f == f`
   *   - `f >=> pure == f`
   *   - `(f >=> g) >=> h == f >=> (g >=> h)`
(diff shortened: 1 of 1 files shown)


rossabaker:
No strong opinion either way, but there are still a handful of "result of `r`."

mpilquist:
I'm used to `R` for result so changing this seems pretty jarring without a significant improvement.

SystemFw:
I also prefer `R-for-result`. However, I would like to eliminate the inconsistency in the scaladoc where it's called "resource" in a couple of places, and stick with "result" everywhere

diesalbla:
> I'm used to R-for-result so changing this seems pretty jarring without a significant improvement.

IMHO it may be better if we were to consider, and perhaps ask, people who are _not_ familiar with the code. We are as likely to understand it under any name, but we may be suffering from the curse of knowledge. It is potential new contributors that would benefit from distinctive names for `Result` (the sub-ADT of the `Pull` ADT), and the _result_ of evaluating/compiling a pull.

SystemFw:
> that would benefit from distinctive names for Result (the sub-ADT of the Pull ADT)

I don't think users will ever see subtypes of Pull, just like they don't see the subtypes of `IO` (or even really know that `IO` has subtypes)

diesalbla:
> I don't think users will ever see subtypes of Pull, just like they don't see the subtypes of IO

As mentioned, the renaming is mostly for future contributors benefit. 

Nevertheless, estimate that users are most likely to tell appart the two type parameters of `Pull`, if the name for referring to the second one is more specific than _result_, which is far more overloaded in programming, and thus more confusing.

SystemFw:
> As mentioned, the renaming is mostly for future contributors benefit.

which is a fair point, but then I'd rather rename the `Result` ADT than the type parameter.

> than result, which is far more overloaded in programming, and thus more confusing.

but here it's not overloaded, it means the same thing it means everywhere else: the outcome of an action, that you can operate on (which is consistent with its `Monad` instance).

diesalbla:
> which is a fair point, but then I'd rather rename the Result ADT than the type parameter.

That is a good idea, regardless. https://github.com/typelevel/fs2/pull/2272

PR #2275, opened 2021-02-14
Question - Eager and Strict pull append?

diesalbla:
In `Pull` we have the `>>` method which performs a lazy append: although the receiver `Pull` object is allocated before the call, the `p2` or `post`  pull argument is passed lazily. Lazy evaluation may have some allocation costs. When the receiver / prefix pull is simple enough that, barring interruption, the post is certain to be allocated.

The proposal (not certain how useful it would be) is to complement `Pull.>>` with a `Pull.*>` method, which much like the method in the `cats.Applicative` typeclass, is strict on the second argument (the suffix pull) 

The risk of this, of course, is that by preallocating the right-hand side we may be leaking memory, for instance if the prefix is expanded in an infinite pull. However, wouldn't that risk also exist with thunks? Lazy evaluation, at least in Haskell, is not free; so the one in Scala may not be either.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -105,6 +105,19 @@ sealed abstract class Pull[+F[_], +O, +R] {
         }
     }
 
+  /** Appends the given `post` pull after `this` pull.
+    * This is like `*>` except that the `post` pull is eagerly evaluated.
+    */
+  def *>[F2[x] >: F[x], O2 >: O, R2](p2: Pull[F2, O2, R2]): Pull[F2, O2, R2] =
+    new Bind[F2, O2, R, R2](this) {
+      def cont(r: Terminal[R]): Pull[F2, O2, R2] =
+        r match {
+          case _: Succeeded[_] => p2
+          case r: Interrupted  => r
+          case r: Fail         => r
+        }
+    }
+
   /** Lifts this pull to the specified effect type. */
   def covary[F2[x] >: F[x]]: Pull[F2, O, R] = this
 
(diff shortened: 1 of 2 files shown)


SystemFw:
I'm fairly strongly against `*>` on `Pull`, because it constitutes a large user trap: 99.99% of pulls are recursive, and `>>` is what you want there, or you will StackOverflow (the pull doesn't need to be infinite, it's enough for it to be recursive)

diesalbla:
> I'm fairly strongly against *> on Pull, because it constitutes a large user trap: 99.99% of pulls are recursive, and >> is what you want there.

Perhaps would it be acceptable if it used a less subtle name, such as `strictAppend` ? 

Note that, in most cases where the `Pull` is recursive, that I saw on the `Stream` the `post` soon enough reaches an `uncons` or a `Pull.flatMap`, so they are not **_left_**-recursive.

mpilquist:
Are there any benchmarks showing this makes a big difference? I worry that we will provide a super dangerous api that has a theoretical performance improvement but one that doesn't actually manifest in real programs.

diesalbla:
> Are there any benchmarks showing this makes a big difference? I worry that we will provide a super dangerous api that has a theoretical performance improvement but one that doesn't actually manifest in real programs.

Not really. This is something I wondered while looking through the code, that there are many calls to `>>`  which may be lazy without need, but benchmarks would be needed.

kubukoz:
@SystemFw maybe having `*>` available directly, but as a deprecated method, would be better than not having it at all? That would shadow the `Applicative` syntax and make it harder to do the wrong thing... although it's the right thing in the remaining 1%.

PR #2286, opened 2021-02-20
Stream - Concurrently - Require INothing

diesalbla:
The type signature of the `concurrently` method accepts any type `O2`. Perhaps it would be better to require it to be `INothing`, so that the caller is aware that no outputs can come off it.


mpilquist:
I agree with comments in Gitter -- this is a lot of source breakage for something that in practice hasn't seemed to have caused confusion.

diesalbla:
@mpilquist Making it clear that the background stream is unused is just one reason. Another important reason is to steer callers towards performance. If a stream's outputs are unused, forcing the caller to _drain_ those outputs can lead them to avoid generating those outputs in the first place.

This is the same kind of reason we have for defining methods on chunks, rather than single outputs.

PR #2287, opened 2021-02-20
Stream: Introduce foreach / foreachChunk functions.

diesalbla:
The `evalMap` and `evalTap` operations can be used to perform an effectful action on each element of the stream. However, because Streams are lazy, and we must allow the consumer of a stream to control which effects are performed, the `evalMap` function needs to emit the new outputs in singleton chunks. However, when a stream is drained, the consumer can no longer control where the stream is stopped. If `evalMap` is followed by `drain`, as happens at several places, that creates an unnecessary use of memory. 

To remedy that, this introduces the functions `foreach` and `foreachChunk`, which take an effectful action and perform
it in every element or chunk of the stream, perform them eagerly, but emit nothing and thus allocate no new chunks.


kubukoz:
Why the closing? It seemed pretty cool 😊

diesalbla:
@kubukoz There is a `foreach` method, but which is simply defined. Follow-up on #2288

PR #2288, opened 2021-02-20
Stream: add execMap and execMapChunk functions.

diesalbla:
This introduces the functions `execMap` and `execMapChunk`. The latter takes an effectful action on chunks, performs it, and emit nothing. The former does the action on elements, but running them on all elements of each chunk.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -967,21 +967,21 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   def enqueueUnterminatedChunks[F2[x] >: F[x], O2 >: O](
       queue: Queue[F2, Chunk[O2]]
   ): Stream[F2, Nothing] =
-    this.chunks.foreach(queue.offer)
+    execMapChunk(queue.offer(_))
 
   /** Enqueues the elements of this stream to the supplied queue and enqueues `None` when this stream terminates.
     */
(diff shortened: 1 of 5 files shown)


mpilquist:
Any benchmarks showing these new variants carry their weight? I doubt they will meaningfully increase performance and hence I’d rather stick with a smaller set of composable operations.

diesalbla:
Copying conversation here fromt he Gitter channel...

> [Diego E. Alonso Blas](@diesalbla)  IIUC, once a stream is drained there is no way for the caller to control it or stop it. All the "F"-actions in it are performed until one fails, so not sure if traversing chunks is a problem, even if action fails. I.e. with str.foreach, from each chunk of N items, the end pull would have N nodes of the Pull.eval type, each one being a F[_] based on the. With str.execMap, on the other hand, the end pull would have 1 Pull.eval node, that would be a traverse_ on each chunk. However, even if one of the F fails, either way, be it at stream compilation or at chunk traversing, the semantics would be the same. Unless I am missing something?

> [Michael Pilquist](@mpilquist)  Pretty sure interruptWhen has different behavior between those two 

> [Fabio Labella](@SystemFw)  the other thing though is this line of optimisation is when the eval mapping is terminal but the loss of perf from of unchunking is not primarily due to the fact that you create lots of chunks but from the fact that subsequent operations happen elem by elem after that or in other words, I don't think it matters that much that evalMap(..).drain produces singleton chunks, because you don't operate on them after that

PR #2294, opened 2021-02-21
Stream - Intersperse: keep chunks.

diesalbla:
Currently, the `intersperse` method modifies the Chunk structure of
the receiver stream, by extracting the first element into a separate
chunk. We modify the `intersperse` method to avoid  that.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1653,29 +1653,32 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * scala> Stream(1, 2, 3, 4, 5).intersperse(0).toList
     * res0: List[Int] = List(1, 0, 2, 0, 3, 0, 4, 0, 5)
     * }}}
+    *
+    * This method preserves the Chunking structure of `this` stream.
     */
-  def intersperse[O2 >: O](separator: O2): Stream[F, O2] =
-    this.pull.echo1.flatMap {
-      case None => Pull.done
-      case Some(s) =>
-        s.repeatPull {
-          _.uncons.flatMap {
-            case None => Pull.pure(None)
-            case Some((hd, tl)) =>
-              val interspersed = {
-                val bldr = Vector.newBuilder[O2]
-                bldr.sizeHint(hd.size * 2)
-                hd.foreach { o =>
-                  bldr += separator
-                  bldr += o
-                }
-                Chunk.vector(bldr.result())
-              }
-              Pull.output(interspersed) >> Pull.pure(Some(tl))
-          }
-        }.pull
-          .echo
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1653,29 +1653,32 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+    *
+    * This method preserves the Chunking structure of `this` stream.
+    */
+  def intersperse[O2 >: O](separator: O2): Stream[F, O2] = {
+    def doChunk(hd: Chunk[O], isFirst: Boolean): Chunk[O2] = {
+      val bldr = Vector.newBuilder[O2]
+      bldr.sizeHint(hd.size * 2 + (if (isFirst) 1 else 0))
+      if (isFirst)
+        bldr += (hd.head.get)
+      val ch = if (isFirst) hd.drop(1) else hd
```
mpilquist, line 1665: This `hd.drop(1)` will result in an unnecessary copy of the chunk -- no worse than the original implementation but shame to do this copy when indexing or advancing an iterator would be sufficient.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1653,29 +1653,32 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
-              Pull.output(interspersed) >> Pull.pure(Some(tl))
-          }
-        }.pull
-          .echo
+    *
+    * This method preserves the Chunking structure of `this` stream.
+    */
+  def intersperse[O2 >: O](separator: O2): Stream[F, O2] = {
+    def doChunk(hd: Chunk[O], isFirst: Boolean): Chunk[O2] = {
+      val bldr = Vector.newBuilder[O2]
```
mpilquist, line 1661: Not related to the PR but vector-based chunks are generally slow unless we know the chunks are big enough to overcome the constant factors. I think we're better off with a mutable buffer here like is done inside various chunk combinators.

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: That copy, does it not share the array or underlying data structure? Would it clone the entire array (if array-backed)? 
Yes, an Iterator may be better, but that also has a little deallocation

mpilquist: Yeah, the built in chunk types all use structural sharing. Array backed chunks will share the underlying array and increment the offset / decrement the size.

Note even with structural sharing, `hd.drop(1)` will allocate a new chunk instance, which is the same number of allocations as an iterator allocation. We can avoid both by using integer indexing here (w/ a var and while loop). Though honestly, it won't matter either way w.r.t. performance.

Issue #2178, opened 2020-12-09
Broadcast. High CPU/Memory usage


iRevive:
Using `Stream#broadcastTo` with the high message rate source leads to increased CPU usage.
Meanwhile, Akka shows 2-3x less CPU usage. 

I tried different JVMs: OpenJDK 11, OpenJDK 14, OpenJDK 14 OpenJ9. The CPU usage is more or less the same on every virtual machine. 

## Prerequisites
**JVM:** OpenJDK 64-Bit Server VM (11.0.4+11, mixed mode)
**fs2:** 2.4.6
**akka-streams:** 2.6.6

## Scenario 1. fs2. map-discard. ~15% CPU usage

```scala
import cats.effect.{ExitCode, IO, IOApp}
import fs2._

object CPUTest extends IOApp {
  override def run(args: List[String]): IO[ExitCode] = {
    val source: Stream[Pure, Int] = Stream.range(1, Int.MaxValue)
    val discard: Pipe[IO, Int, Unit] = _.map(_ => ())

    source.map(discard).compile.drain.as(ExitCode.Success)
  }
}
```

![image](https://user-images.githubusercontent.com/6395483/101655200-02914d80-3a4a-11eb-9503-45644b762d7c.png)

## Scenario 2. fs2. Broadcast to 1 pipe. ~28% CPU usage

```scala
import cats.effect.{ExitCode, IO, IOApp}
import fs2._

object CPUTest extends IOApp {
  override def run(args: List[String]): IO[ExitCode] = {
    val source: Stream[Pure, Int] = Stream.range(1, Int.MaxValue)
    val discard: Pipe[IO, Int, Unit] = _.map(_ => ())

    source.broadcastTo(pipe).compile.drain.as(ExitCode.Success)
  }
}
```

![image](https://user-images.githubusercontent.com/6395483/101655545-67e53e80-3a4a-11eb-9eba-0fc3b3fc1de7.png)

## Scenario 3. fs2. Broadcast to 100 pipes. 50-80% CPU usage

```scala
import cats.effect.{ExitCode, IO, IOApp}
import fs2._

object CPUTest extends IOApp {
  override def run(args: List[String]): IO[ExitCode] = {
    val source: Stream[Pure, Int] = Stream.range(1, Int.MaxValue)
    val discard: Pipe[IO, Int, Unit] = _.map(_ => ())
    val pipes: List[Pipe[IO, Int, Unit]] = List.fill(100)(discard)

[...]

PR #2296, opened 2021-02-23
Add Benchmark on the broadcast method.

diesalbla:
Based on #2178

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/BroadcastBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/BroadcastBenchmark.scala
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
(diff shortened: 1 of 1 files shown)


benchmark/src/main/scala/fs2/benchmark/BroadcastBenchmark.scala
```diff
@@ -0,0 +1,70 @@
...
+import cats.effect.{IO}
+import cats.effect.unsafe.implicits.global
+import org.openjdk.jmh.annotations.{Benchmark, Param, Scope, State, Setup, TearDown}
+import java.util.concurrent.Executors
+import scala.concurrent.ExecutionContext
+
+@State(Scope.Thread)
+class BroadcastBenchmark {
+  val jExecutor = Executors.newSingleThreadExecutor
+  val sExecutor = ExecutionContext.fromExecutor(jExecutor)
```
SystemFw, line 34: Can we setup things to use the right thread pool?

nikiforo: This example compiles as is, 
https://github.com/typelevel/fs2/issues/2178#issuecomment-799593738

However, it still uses `import cats.effect.unsafe.implicits.global` and I'm not sure whether it's the right one.

SystemFw: the other thing about this benchmark is that it does measure overhead, but not real performance I think: the stream is unchunked, and there is no per chunk processing, which is the opposite of what you need to do to make fs2 perform well

nikiforo: I agree, but the benchmark doesn't estimate abstract an fs2 perfomance.
It compares baseline(BroadcastBenchmark.through) with some other function(BroadcastBenchmark.broadcastTo1), they both should overhead with chunks exactly the same way. And the thing that benchmark is measuring how bad/good one function(baseline) performs in comparison with other(the benchmarked one).
That's why I'm drawing an attention to ration between `BroadcastBenchmark.broadcastTo1` and `BroadcastBenchmark.through`, which is 25:1. Moreover, previous ration was 20:1, so either `through` became 25% faster or `broadcastTo1` became 20% slower.

Ohh... I might have understood your point. You're talking that previous version of `broadcastTo1` worked poorly with chunks, but this version works with chunks better?

SystemFw: > You're talking that current version of broadcastTo1 works better with chunks, than pervious one did?

I'm not _stating_ that is the case (it might as well gone slower since it now goes through Topic, although in the Channel branch there is yet another implementation). I'm saying that if that was the case, this benchmark wouldn't show it

SystemFw: Oh another minor thing about this specific op is that I would be very surprised if the number of subs goes past maybe 10 (since those are hardcoded pipes with different logic, I suspect 2 is the most common case), but something like this can still be helpful for other things, such as Topic (where a large number of dynamic subscribers is very possible)

PR #2300, opened 2021-02-25
Shared server sockets

### LINKED ISSUE #2289: Semantically breaking change in SocketGroup.serverResource
On 2.5.x, the return type for `serverResource` is:
```scala
Resource[F, (InetSocketAddress, Stream[F, Resource[F, Socket[F]]])]
```
On the latest 3.x milestone, the return type is now:
```scala
Resource[F, (SocketAddress[IpAddress], Stream[F, Socket[F]])]
```

To preface, the change in the signature makes complete sense from a resource safety perspective. IIRC the former definition is actually a little bit misleading; the socket associated with the inner `Resource` has already been allocated before `use` is even called! As a consequence, if `use` is _not_ called, the socket is never closed and gets stuck in limbo, resulting in a file leak.

The larger issue here is that the lifecycles of the server socket and all accepted sockets are now bound to the same stream. However, in practice, it's completely sound for them to have disjoint lifecycles e.g. you can close a server socket while leaving sockets that were accepted open. 

Ember Server exploited the lack of safety in the old definition to implement graceful shutdowns: we can interrupt the server stream and force closure upon it while allowing existing connections to complete their work for some time. `parJoin` semantics are actually too strict because inner streams lease the input stream's scope and prevents its resources from finalizing until all inner streams are closed. So we implemented a weaker version of `parJoin` called [forking](https://github.com/http4s/http4s/blob/main/ember-server/src/main/scala/org/http4s/ember/server/internal/StreamForking.scala) that doesn't lease any scopes, which is unsafe in general but completely fine for this particular use case.

`forking` completely breaks with the new definition of `serverResource` precisely because it doesn't lease the scope. Since the accepted sockets are bound to the server stream, as soon as a connection stream is spawned, the socket is going to finalize. Is there a way to only partially lease some resources from the server stream in a safe manner? In particular, lease the accepted socket but not the server socket.

This is blocking Http4s builds on CE3 ATM, and reverting to `parJoin` would be a pretty major regression in behavior (and performance to a lesser degree), so it would be cool if we could come to a resolution on this soon

RaasAhsan:
Intends to fix #2289. Relevant Gitter conversation located here: https://gitter.im/functional-streams-for-scala/fs2?at=6035dec142f30f75c7c35746

So, the idea here is that the lifecycles of server sockets and client sockets have independent lifecycles, so it should be possible for us to finalize them separately. This was possible to do in 2.5.x because the `SocketGroup` server signature returned a `Stream[F, Resource[F, Socket[F]]]`, in conjunction with the `forking` combinator that was developed in Ember. Unfortunately, that method is leaky (what happens if you don't call `use` on the inner resource?), so the signature changed to `Stream[F, Socket[F]]`, where the sockets are actually bound to the server stream. `parJoin` is necessary here in order to lease the scope and prevent the socket from immediately finalizing, but this sneakily also leases the server socket, preventing that from finalizing until _all_ client sockets have finalized!

One solution to maintaining independent resource lifecycles while preserving safety is to implement a data structure called `Shared` that can facilitate the transfer of ownership of resources, which is captured in this PR. A client socket can be allocated on the server stream into a shared resource and handed off to the client stream which will acquire the resource. After that, the server socket can release the resource before moving onto accepting the next socket (this won't actually happen when using `parJoin` because of leasing).

I think the `forking` function is going to have change a bit; it'll have to operate on a `Stream[F, Stream[F, Shared[F, O]]]` because we need to ensure that the shared socket is acquired _before_ control is returned to the accept stream. If it doesn't, the accepted socket is going to be immediately finalized after return. See: https://github.com/http4s/http4s/blob/main/ember-server/src/main/scala/org/http4s/ember/server/internal/StreamForking.scala#L71 . In comparison to `parJoin`, `forking` replaces the notion of implicit scope leasing with the notion of explicit resource transfer. 

Notes:
1. The current `server` signatures are completely safe when used with `parJoin`, but I think we can even just revert them to their previous versions, and add a separate signature that exposes `Shared`.
2. I'm wondering if we should try to include a (safer) variant of `forking` in fs2 that people can use to exercise the graceful shutdown behavior
3. Need to look at fs2-netty to see how this plays out there

### DIFF
--- a/core/shared/src/main/scala/fs2/Shared.scala
+++ b/core/shared/src/main/scala/fs2/Shared.scala
@@ -22,7 +22,6 @@
 package fs2
 
 import cats.effect._
-import cats.effect.kernel.Resource
 import cats.syntax.all._
 
 sealed trait Shared[F[_], A] {
--- a/core/shared/src/main/scala/fs2/Shared.scala
+++ b/core/shared/src/main/scala/fs2/Shared.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2
+
+import cats.effect._
+import cats.effect.kernel.Resource
+import cats.syntax.all._
+
+sealed trait Shared[F[_], A] {
+  def resource: Resource[F, A]
+}
+
+object Shared {
+  def allocate[F[_], A](
+      resource: Resource[F, A]
+  )(implicit F: Concurrent[F]): Resource[F, (Shared[F, A], A)] = {
+    final case class State(value: A, finalizer: F[Unit], permits: Int) {
+      def addPermit: State = copy(permits = permits + 1)
+      def releasePermit: State = copy(permits = permits - 1)
+    }
+
+    MonadCancel[Resource[F, *]].uncancelable { _ =>
+      for {
+        shared <- Resource.eval(resource.allocated.flatMap { case (a, fin) =>
+          F.ref[Option[State]](Some(State(a, fin, 0))).map { state =>
+            def acquire: F[A] =
+              state.modify {
+                case Some(st) => (Some(st.addPermit), F.pure(st.value))
+                case None =>
+                  (None, F.raiseError[A](new Throwable("finalization has already occurred")))
+              }.flatten
+
+            def release: F[Unit] =
+              state.modify {
+                case Some(st) if st.permits > 1 => (Some(st.releasePermit), F.unit)
+                case Some(st)                   => (None, st.finalizer)
+                case None                       => (None, F.raiseError[Unit](new Throwable("can't finalize")))
+              }.flatten
+
+            new Shared[F, A] {
+              override def resource: Resource[F, A] =
+                Resource.make(acquire)(_ => release)
+            }
+          }
+        })
+        a <- shared.resource
+      } yield (shared, a)
+    }
+  }
+}
--- a/core/shared/src/main/scala/fs2/Shared.scala
+++ b/core/shared/src/main/scala/fs2/Shared.scala
@@ -44,21 +44,15 @@ object Shared {
         shared = new Shared[F, A] {
           def acquire: F[A] =
             state.modify {
-              case Some(st) => 
-                println("acquired")
-                (Some(st.addPermit), F.pure(st.value))
+              case Some(st) => (Some(st.addPermit), F.pure(st.value))
               case None =>
                 (None, F.raiseError[A](new Throwable("finalization has already occurred")))
             }.flatten
 
           def release: F[Unit] =
             state.modify {
-              case Some(st) if st.permits > 1 => 
-                println("released with many")
-                (Some(st.releasePermit), F.unit)
-              case Some(st)                   => 
-                println("released with one")
-                (None, st.finalizer)
+              case Some(st) if st.permits > 1 => (Some(st.releasePermit), F.unit)
+              case Some(st)                   => (None, st.finalizer)
               case None                       => (None, F.raiseError[Unit](new Throwable("can't finalize")))
             }.flatten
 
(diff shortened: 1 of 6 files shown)


io/src/main/scala/fs2/io/net/Network.scala
```diff
@@ -132,13 +132,13 @@ object Network {
           address: Option[Host],
           port: Option[Port],
           options: List[SocketOption]
-      ): Stream[F, Socket[F]] = globalSocketGroup.server(address, port, options)
+      ): Stream[F, Shared[F, Socket[F]]] = globalSocketGroup.server(address, port, options)
```
RaasAhsan, line 135: I think both these signature can go back to the old version, and we just introduce a more general one that exposes `Shared`

io/src/main/scala/fs2/io/net/SocketGroup.scala
```diff
@@ -177,8 +179,17 @@ private[net] object SocketGroup {
             }
 
           Stream.eval(acceptChannel.attempt).flatMap {
```
RaasAhsan, line 181: Note that the leak of a single socket when being interrupted is still present here. I tried to replace this `eval` with a `Stream.resource` backed by an underlying `Resource.makeFull` which polled over the `sch.accept` operation, but it didn't seem like it responded to interruption immediately.

core/shared/src/main/scala/fs2/Shared.scala
```diff
@@ -0,0 +1,66 @@
...
+object Shared {
+  def allocate[F[_], A](
+      resource: Resource[F, A]
+  )(implicit F: Concurrent[F]): Resource[F, (Shared[F, A], A)] = {
+    final case class State(value: A, finalizer: F[Unit], permits: Int) {
+      def addPermit: State = copy(permits = permits + 1)
+      def releasePermit: State = copy(permits = permits - 1)
+    }
+
+    MonadCancel[Resource[F, *]].uncancelable { _ =>
```
RaasAhsan, line 40: I looked at the implementation and it seems to make sense to me

SystemFw: you made the whole resource uncancelable here though

io/src/main/scala/fs2/io/net/SocketGroup.scala
mpilquist: Hm, that should work fine...

core/shared/src/main/scala/fs2/Shared.scala
RaasAhsan: Yeah I think I need to `poll` on the `allocated`

io/src/main/scala/fs2/io/net/SocketGroup.scala
RaasAhsan: I'll update with a snippet sometime today

RaasAhsan: Looks like this was because of what Fabio described in his comment (wasn't polling on the `allocated` resource)

mpilquist:
@RaasAhsan What's that status of this one?

RaasAhsan:
Sorry for dropping the ball on this (I've been a bit busy), but I think it's ready for a deeper review. It does deserve some tests which I'll try to get to ASAP, but I'm curious to hear thoughts on general approach. I wonder if it's worth exploring a more `Stream`-native mechanism for resource ownership transfer in the future.

TimWSpence:
@mpilquist @RaasAhsan I was actually looking at the `StreamForking` code yesterday and it seems like it would be a big win. Would you be interested in me trying to get this over the line and if so, is there agreement on what is required? Thanks! :)

armanbilge:
@TimWSpence I need to take a deeper look at this PR, but you may be interested in the change I just made in https://github.com/typelevel/fs2/pull/3063/commits/9e6a59b8490689a2301d843c9587a5d82e6fbdb5.

In fact I have some crazy idea that it fixed this :P but yeah, let me actually read about the issue first 😂

TimWSpence:
> @TimWSpence I need to take a deeper look at this PR, but you may be interested in the change I just made in [9e6a59b](https://github.com/typelevel/fs2/commit/9e6a59b8490689a2301d843c9587a5d82e6fbdb5).
> 
> In fact I have some crazy idea that it fixed this :P but yeah, let me actually read about the issue first 😂

Oh interesting! Is there a parent issue or PR for some conext on that change?

armanbilge:
Yes, it's part of https://github.com/typelevel/fs2/pull/3063.

TimWSpence:
@armanbilge thinking about it a bit more, a reference-counted resource wrapper like this might be quite useful generally in cats effect. What do you think?

TimWSpence:
@armanbilge I don't think your change does fix this - we still have https://github.com/typelevel/fs2/pull/3063/files#diff-f11c9120aaf385f6214d8d1a82560c137b058a244c84ad7d8b7126e78b2081aeL128 which means that the lifecycle of the connection is tied to that of the outer server stream

armanbilge:
@TimWSpence perhaps I misunderstood the original problem. I'm going on https://github.com/typelevel/fs2/issues/2289:

> The larger issue here is that the lifecycles of the server socket and all accepted sockets are now bound to the same stream. However, in practice, it's completely sound for them to have disjoint lifecycles e.g. you can close a server socket while leaving sockets that were accepted open.

I'm pretty sure the changes I made make those lifecycles disjoint: the server now accepts sockets into `Channel`, which can be consumed as a stream. Furthermore, the connection lifecycles are tied to the consuming stream, not the stream producing accepted sockets into the `Channel`. So it should be fully possible to close the server socket and thus close the `Channel`, without terminating the consuming stream.

TimWSpence:
> @TimWSpence perhaps I misunderstood the original problem. I'm going on #2289:
> 
> > The larger issue here is that the lifecycles of the server socket and all accepted sockets are now bound to the same stream. However, in practice, it's completely sound for them to have disjoint lifecycles e.g. you can close a server socket while leaving sockets that were accepted open.
> 
> I'm pretty sure the changes I made make those lifecycles disjoint: the server now accepts sockets into a `Channel`, which is consumed as a stream. Furthermore, the connection lifecycles are tied to the consuming stream, not the stream producing accepted sockets into the `Channel`. So it should be fully possible to close the server socket and thus close the `Channel`, without terminating the consuming stream.

Oh sorry, yeah I missed that those resources are tied to the `stream` coming from the channel. From some brief experimentation it didn't look like it was working for me though. I'll do some more digging

armanbilge:
Well, maybe that's not really the problem 😂 see subsequent comment in https://github.com/typelevel/fs2/issues/2289#issuecomment-783510663. I'm not sure if FS2 actually needs any changes. I suspect we can do this http4s, maybe with some strategic use of `allocatedCase` to manage the outer resource.

TimWSpence:
> Well, maybe that's not really the problem 😂 see subsequent comment in [#2289 (comment)](https://github.com/typelevel/fs2/issues/2289#issuecomment-783510663). I'm not sure if FS2 actually needs any changes. I suspect we can do this http4s, maybe with some strategic use of `allocatedCase` to manage the outer resource.

🤦 how did I miss that. Nice, I'll poke at http4s a bit and see what I can come up with

PR #2306, opened 2021-02-28
Stream - Rewrite some concurrent methods.

diesalbla:
List of several changes to improve code readability.

- Concurrently: extract inner `watch` method, and rename some vairables.
- Format some scaladocs, to keep on the 80 flexible line.
- InterruptWhen: use a `Stream.force` form, without the  `Stream.eval` operations. Use some variables.
- observeAsync: use the `Stream.force` form, and make some use of recursive pull functions for the sinkers.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -548,26 +548,22 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   )(implicit F: Concurrent[F2]): Stream[F2, O] = {
     val fstream: F2[Stream[F2, O]] = for {
       interrupt <- F.deferred[Unit]
-      doneR <- F.deferred[Either[Throwable, Unit]]
+      backResult <- F.deferred[Either[Throwable, Unit]]
     } yield {
-      def runR: F2[Unit] =
-        that.interruptWhen(interrupt.get.attempt).compile.drain.attempt.flatMap { r =>
-          doneR.complete(r) >> {
-            if (r.isLeft)
-              interrupt
-                .complete(())
-                .void // interrupt only if this failed otherwise give change to `this` to finalize
-            else F.unit
-          }
-        }
+      def watch[A](str: Stream[F2, A]) = str.interruptWhen(interrupt.get.attempt)
+
+      val compileBack: F2[Boolean] = watch(that).compile.drain.attempt.flatMap {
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -642,33 +639,32 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
-            case v @ Some(_) => queue.offer(v)
-            case None        => F.unit
-          }
+  )(implicit F: Temporal[F2]): Stream[F2, O] = Stream.force {
+    for {
+      queue <- Queue.bounded[F2, Option[O]](1)
+      ref <- F.ref[Option[O]](None)
+    } yield {
+      val enqueueLatest: F2[Unit] =
+        ref.getAndSet(None).flatMap(prev => if (prev.isEmpty) F.unit else queue.offer(prev))
```
kubukoz, line 648: Is this... `prev.traverse_(queue.offer)`?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -642,33 +639,32 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+        ref.getAndSet(Some(o)).flatMap {
+          case None    => F.start(F.sleep(d) >> enqueueLatest).void
+          case Some(_) => F.unit
+        }
 
-        val in: Stream[F2, Unit] = chunks.evalMap(onChunk) ++
-          Stream.exec(enqueueLatest >> queue.offer(None))
+      def go(tl: Pull[F2, O, Unit]): Pull[F2, INothing, Unit] =
+        Pull.uncons(tl).flatMap {
+          // Note: hd is non-empty, so hd.last.get is sage
```
kubukoz, line 658:
```suggestion
          // Note: hd is non-empty, so hd.last.get is safe
```

core/shared/src/main/scala/fs2/Stream.scala
diesalbla: Yes.

diesalbla: `flatMap(_.traverse_(queue.offer))`

diesalbla: I had noticed that it is `traverse_` and had looked at it... but then I decanted towards explicitly using `prev` to indicate it is the previous value, if any, that is inserted to queue.

PR #2308, opened 2021-03-01
Pull: turn `.stream` and `.streamNoScope` into extension methods.

diesalbla:
The Pull methods `.scope` and `.streamNoScope` are currently implemented with the type-bounds evidence pattern and the
`asInstanceOf`. It may be better to define them as extension methods in an implicit class which, unlike the Pull class,
call be specialised to pulls where the result type is Unit.

This did not seem to require any special imports anywhere. It may be that, as implicits, they are attached to the `Pull` type everywhere.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -58,29 +58,6 @@ sealed abstract class Pull[+F[_], +O, +R] {
   def attempt: Pull[F, O, Either[Throwable, R]] =
     map(r => Right(r)).handleErrorWith(t => Succeeded(Left(t)))
 
-  /** Interpret this `Pull` to produce a `Stream`, introducing a scope.
-    *
-    * May only be called on pulls which return a `Unit` result type. Use `p.void.stream` to explicitly
-    * ignore the result type of the pull.
-    */
-  def stream(implicit ev: R <:< Unit): Stream[F, O] = {
-    val _ = ev
-    new Stream(Pull.scope(this.asInstanceOf[Pull[F, O, Unit]]))
-  }
-
-  /** Interpret this `Pull` to produce a `Stream` without introducing a scope.
-    *
-    * Only use this if you know a scope is not needed. Scope introduction is generally harmless and the risk
-    * of not introducing a scope is a memory leak in streams that otherwise would execute in constant memory.
-    *
-    * May only be called on pulls which return a `Unit` result type. Use `p.void.stream` to explicitly
-    * ignore the result type of the pull.
-    */
-  def streamNoScope(implicit ev: R <:< Unit): Stream[F, O] = {
-    val _ = ev
-    new Stream(this.asInstanceOf[Pull[F, O, Unit]])
-  }
-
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -147,6 +124,28 @@ sealed abstract class Pull[+F[_], +O, +R] {
 
 object Pull extends PullLowPriority {
 
+  implicit class StreamPullOps[F[_], O](val pull: Pull[F, O, Unit]) extends AnyVal {
```
mpilquist, line 127: Make `val` private or else this extension will allow syntax like `p.pull.pull.pull.pull.pull.pull`

diesalbla: `value class parameter must be a val and not be private[this]`

diesalbla: I can rename to `self`, if that helps...

mpilquist: `private val self: Pull[F, O, Unit]` should be fine... Note the other implicit classes we have with a public `self` val have the same issue -- `foo.self.self.self.self.self`

diesalbla: Ok... the problem is that this would mean allocating (you can no longer make them `AnyVal`).

mpilquist: That's not right...

```scala
io/src/main/scala-2.13/fs2/io/CollectionCompat.scala
27:  implicit class JIteratorOps[A](private val self: java.util.Iterator[A]) extends AnyVal {
30:  implicit class JIterableOps[A](private val self: java.lang.Iterable[A]) extends AnyVal {
33:  implicit class JSetOps[A](private val self: java.util.Set[A]) extends AnyVal {
36:  implicit class ListOps[A](private val self: List[A]) extends AnyVal {
39:  implicit class SetOps[A](private val self: Set[A]) extends AnyVal {
42:  implicit class EnumerationOps[A](private val self: java.util.Enumeration[A]) extends AnyVal {
```

diesalbla: 🤔

diesalbla: Ok, just pushed version with private and anyval. I got mistaken with the `private[this]` not allowed.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -147,6 +124,28 @@ sealed abstract class Pull[+F[_], +O, +R] {
 
 object Pull extends PullLowPriority {
 
+  implicit class StreamPullOps[F[_], O](private val pull: Pull[F, O, Unit]) extends AnyVal {
```
nikiforo, line 127: I looked on other random implicit classes in fs2, some of them have `final`, others don't.

Shouldn't final be added here?

diesalbla: changed.

PR #2309, opened 2021-03-01
Pull - avoid use of asInstanceOf.

diesalbla:
With newer Scala 2.13.5 we can avoid some, and make more use of types.


mpilquist:
I think the 3.0.0-RC1 build has more warnings as a result of some of these changes. I'd prefer to keep casts than trade them for warnings about compiler inserted casts. https://github.com/typelevel/fs2/runs/2000458651

diesalbla:
So, the cause for these warnings is that, at some places, I added the predefined type parameter `Unit` or `X` in the pattern match of the type. Leaving that type parameter to `_` would not allow `Scala` to infer that this type parameter must be `Unit` or `X` in each case.

mpilquist:
Right, but Dotty is warning us that the match is unsafe and will result in class cast exceptions at runtime. I'd rather explicitly cast those with a comment assuming we can't make the match safe / sound.

PR #2310, opened 2021-03-01
Stream replace evidence & asInstanceOf to implicit class

nikiforo:
extensions are placed in companion object, thus no imports should be required


mpilquist:
I don't have strong opinions on evidence params vs implicit classes but I'd like to hear from other folks. Avoiding casts is nice but discoverability is a little worse. I personally tend to lean towards evidence params for simple stuff and implicit classes for more complex stuff.

Daenyth:
IME, fs2 is hard for newcomers to learn. I think the ops classes make things harder for people to learn, so I'm a soft -1

PR #2311, opened 2021-03-01
sliding2

nikiforo:
Imagine we want to count amount of sign inversions in `val list = List(1,-1,-1,1)`
In this scenario I prefer to use `zip(self.tail)` over `sliding(2)`, because it is more convenient to match on tuples than variable sized collections, `list.zip(list.tail).count((prev, next) => prev < 0 xor next < 0)`.
Inasmuch as side effects `zip` on streams has a different meaning.


mpilquist:
There's also `zipWithPrevious` and `zipWithNext` which are pretty similar

PR #2312, opened 2021-03-02
Broadcast

SystemFw:
This PR revisits the broadcasting abstractions in fs2:

-  `.broadcast` is removed. In its raw form, this method is almost impossible to use safely, and depends on subtle scoping details (e.g. `chunkN` works, `foldMap(List(_))`) doesn't, and at the same time it's not flexible enough to encode fully dynamic subscribers like `Topic` is.
- (not in this PR, just for clarity) The improved `Topic`  is recommended for cases where you might have used `.broadcast`. If you managed to get that to work, that is.
- the main method for static broadcasting is `broadcastThrough`, which is safe and easy to use. It has been reimplemented using Topic.
- The various `broadcast*` overloads have even removed
- `Broadcast` has been removed.
- Fixes a bug with topic which causes all subscribers to freeze when one was removed


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -202,62 +202,52 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+          .map { pipe =>
+            Stream
+              .resource(topic.subscribeAwait(1))
+              .flatMap { sub =>
+                // crucial that awaiting on the latch is not passed to
+                // the pipe, so that the pipe cannot interrupt it and alter
+                // the latch count
+                Stream.exec(latch.release >> latch.await) ++
+                  sub.unNoneTerminate.flatMap(Stream.chunk).through(pipe)
+              }
```
diesalbla, line 243: Could we extract this lambda block as a function `def pump(pipe:  Pipe[F2, O, O2])` ?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -202,62 +202,52 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+          }
+          .parJoinUnbounded
+          .concurrently {
+            Stream.eval(latch.await) ++
+              chunks.noneTerminate.through(topic.publish)
```
diesalbla, line 248: Could we extract this as a value `backgroundPublish`?

core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -124,9 +125,21 @@ object Topic {
...
+                    // _After_ we remove the bounded queue for this
+                    // subscriber, we need to drain it to unblock to
+                    // publish loop which might have already enqueued
+                    // something.
+                    def drainQueue: F[Unit] =
+                      subs.get(id).traverse_ { q =>
+                        q.tryTake.flatMap {
+                          case None    => F.unit
+                          case Some(_) => drainQueue
+                        }
```
diesalbla, line 138: Perhaps another traverse? 
```suggestion
                        q.tryTake.flatMap(_.traverse_(_ => drainQueue))
```

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: I'm factored out the `unNoneTerminate...` stuff, but I want to keep the waiting logic all together, so it's easy to understand how to `parJoin`, the `latch` and the `Resource` all work

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: I've factored out the `chunks.noneTerminate` stuff, I want to keep the `await` logic where it is

core/shared/src/main/scala/fs2/concurrent/Topic.scala
SystemFw: Used `iterateUntil`, which avoid redoing the `get` as well

PR #2316, opened 2021-03-05
Remove Queue in favor of cats.effect.std.Queue

mpilquist:


SystemFw (approved):
I was expecting the diff to be bigger :)

diesalbla (approved):

PR #2321, opened 2021-03-08
Fix for interruption skipping error handlers in certain cases

### LINKED ISSUE #2197: Errors are not always catched with .handleErrorWith

mpilquist:
Fixes #2197

Based on the analysis from @RaasAhsan. All tests pass with this change and while it feels pretty adhoc, I think we should fix this in the simplest way possible and then look at revamping stream interruption in general later.

### DIFF
--- a/core/shared/src/test/scala/fs2/StreamConcurrentlySuite.scala
+++ b/core/shared/src/test/scala/fs2/StreamConcurrentlySuite.scala
@@ -149,4 +149,24 @@ class StreamConcurrentlySuite extends Fs2Suite {
     }
   }
 
+  test("bug 2197".only) {
+    val iterations = 1000
+    Stream
+      .eval((IO.deferred[Unit], IO.ref[Int](0)).tupled)
+      .flatMap { case (done, innerErrorCountRef) =>
+        def handled: IO[Unit] =
+          innerErrorCountRef.modify { old => 
+            (old + 1, if (old < iterations) IO.unit else done.complete(()).void)
+          }.flatten
+        Stream(Stream(()) ++ Stream.raiseError[IO](new Err)).repeat.flatMap { fg =>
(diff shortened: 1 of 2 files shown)


core/shared/src/test/scala/fs2/StreamConcurrentlySuite.scala
```diff
@@ -149,4 +149,24 @@ class StreamConcurrentlySuite extends Fs2Suite {
     }
   }
 
+  test("bug 2197".only) {
```
SystemFw, line 152: leftover

SystemFw (commented):
I think we'd have to leave with the ad-hoc nature of this for now, as you say

diesalbla (approved):

PR #2326, opened 2021-03-11
Channel

SystemFw:


core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -0,0 +1,318 @@
...
+ * unblocked, before `stream` terminates and further `send` calls
+ * become no-ops. However, if `close` is called after `stream` has
+ * terminated (e.g because it was interrupted, or had a `.take(n)`),
+ * then blocked producers will stay blocked unless they get explicitly
+ * unblocked, either by a further call to `stream` to drain the
+ * Channel, or by a a `race` with `closed`.
+ *
+ */
+
+trait Channel[F[_], A] {
```
mpilquist, line 75: `sealed`? Just in case we want to add operations later?

mpilquist (approved):

core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -85,71 +88,136 @@ abstract class Topic[F[_], A] { self =>
...
+              val result = if (completedNow) Topic.rightUnit else Topic.closed
+
+              state.get
+                .flatMap { case (subs, _) =>
+                  subs.foldLeft(F.unit) { case (op, (_, chan)) =>
+                    op >> chan.close.void
+                  }
+                }
+                .as(result)
+            }
```
diesalbla, line 213:
```suggestion
          (for {
            completedNow <- signalClosure.complete(())
            stt <- state.get
             _ <- stt._2.traverse_(_.close)
            } yield if (completedNow) Topic.rightUnit else Topic.closed
            ).uncancelable
```

diesalbla: Oh, just realised there is no `traverse_` for `LongMap`. Perhaps we could add a function 

```Scala
  def allChannels(f: Channel[F, A] => F[_]): F[Unit] = 
    state.get.flatMap { case (subs, _) => 
      subs.foldLeft(F.unit){ case (op, (_, chan)) => op >> f(chan) }
    } 
```

core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -85,71 +88,136 @@ abstract class Topic[F[_], A] { self =>
...
+      new Topic[F, A] {
+
+        def publish1(a: A): F[Either[Topic.Closed, Unit]] =
+          signalClosure.tryGet.flatMap {
+            case Some(_) => Topic.closed.pure[F]
+            case None =>
+              state.get.flatMap { case (subs, _) =>
+                subs
+                  .foldLeft(F.unit) { case (op, (_, chan)) =>
+                    op >> chan.send(a).void
```
diesalbla, line 155: Perhaps it would help to extract an `foreachChannel` function...

```Scala
def foreachChannel(a: Channel[F, A] => F[_]): F[Unit] = 
```

mpilquist (approved):

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3825,7 +3833,7 @@ object Stream extends StreamLowPriority {
                 Some(Some(CompositeFailure(err0, err)))
               }
             case _ => Some(rslt)
-          } >> outputQ.offer(None).start.void
+          } >> outputChan.close.start.void // TODO is the starting really necessary here? probably needed because of the queue.offer(None) which could block
```
mpilquist, line 3836: Do we still need this .start?

SystemFw: ah, no, I will remove

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -0,0 +1,240 @@
...
+package fs2
+package concurrent
+
+import cats.effect._
+import cats.effect.implicits._
+import cats.syntax.all._
+
+/** Stream aware, multiple producer, single consumer closeable channel.
+  */
+sealed trait Channel[F[_], A] {
```
diesalbla, line 31: I wonder if the words `FanInChannel` or `Fannel` would better illustrate for the purpose of this class. `Channel` is a bit too overloaded.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -0,0 +1,240 @@
...
+    * executing, blocked producers will eventually become unblocked,
+    * before `stream` terminates and further `send` calls become
+    * no-ops.
+    * However, if `close` is called after `stream` has terminated (e.g
+    * because it was interrupted, or had a `.take(n)`), then blocked
+    * producers will stay blocked unless they get explicitly
+    * unblocked, either by a further call to `stream` to drain the
+    * channel, or by a a `race` with `closed`.
+    */
+  def close: F[Either[Channel.Closed, Unit]]
```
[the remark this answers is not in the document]
SystemFw: It's the other way around yeah, good shout.

`AlreadyClosed` is a bit clearer but a bit verbose. Tbh I was tempted by `Boolean` both here and in `send`, but then stopped, cause I'm not sure what the type of potential `trySend` would look like. @mpilquist any thoughts here?

diesalbla:
It was suggested in the comments, but I am a bit more convinced now: we should probably rename this new class from `Channel` to `Funnel`. This last name [better illustrates](https://upload.wikimedia.org/wikipedia/commons/thumb/1/11/Kitchen_Funnel.jpg/1024px-Kitchen_Funnel.jpg)  the _fan-in_ or "several produces, single consumer" idea. Furthermore, the word `Channel` is a bit too common and [overloaded in CS](https://en.wikipedia.org/wiki/Channel_(programming)).  In the `Go` language, for instance, it means [bidirectional 1:1 communication](https://golangdocs.com/channels-in-golang). In Akka, it was used for prototypes of typed actors. In [`monix`](https://monix.io/api/3.1/monix/catnap/ConcurrentChannel.html), it means any cardinality consumers/producers.

diesalbla (approved):

SystemFw:
I'm going to merge this because I need it, we can discuss the naming later I think.

Personally I like the name `Channel`, it's simple and clear. MPSC semantics for channels are fairly natural too, e.g. the channel in Rust stdlib is MPSC https://doc.rust-lang.org/std/sync/mpsc/index.html.

I'm not too keen on trying to convey the specific  semantics of this type of Channel in the name, because there are other very important aspects to it such as graceful shutdown or rechunking that you need to understand anyway, and ultimately that's why it's heavily documented.

I also prefer to have a nice name than to avoid ambiguity,  I got convinced otherwise with `Deferred`, and I don't really like that name. At the end of day, nice names are already all taken _somewhere_, and the  name `Stream` is much more overloaded/common than Channel

Issue #1389, opened 2019-01-04
groupWithin causes double execution


kaishh:
See the example below:

```scala
package example

import java.util.concurrent.atomic.AtomicLong

import cats.effect._
import cats.syntax.flatMap._
import cats.syntax.functor._
import fs2._

import scala.compat.Platform.ConcurrentModificationException
import scala.concurrent.duration._
import scala.concurrent.blocking

class MiniKafka {
  final val NO_CURRENT_THREAD = -1L
  val currentThread = new AtomicLong(NO_CURRENT_THREAD)

  def acquire() = {
    val threadId = Thread.currentThread().getId
    if (threadId != currentThread.get && !currentThread.compareAndSet(NO_CURRENT_THREAD, threadId))
      throw new ConcurrentModificationException
  }

  def release() = {
    currentThread.set(NO_CURRENT_THREAD)
  }

  def subscribe(): this.type = {
    acquire()
    try this finally release()
  }

  def commit() = {
    acquire()

    try {
    } finally release()
  }

  def close() = {
    acquire()
    try Thread.`yield`()
    finally release()
  }


  def poll(): String = {
    acquire()
    try {
      "abc"
    } finally {
      release()
    }
  }
}

object App extends App {

  def repro[F[_]: ContextShift: Timer](implicit F: Concurrent[F]) = {
    Stream.resource {
      Resource.make(F.delay(new MiniKafka().subscribe()))(c => F.delay(c.close()))
    }.flatMap {
      kafka =>
        Stream.repeatEval {
            F.delay(blocking(kafka.poll())).map(kafka -> _)
        }
    }
      .groupWithin(1000, 1.millis).map(_.toVector)
      .evalMap { x =>
        val kafka = x.unzip._1.head

        F.delay(blocking(kafka.commit())) >>
          F.delay(println(x.unzip._2.toString())).as(kafka)
      }
      .compile.drain
  }

  new IOApp {
    override def run(args: List[String]): IO[ExitCode] =
      repro[cats.effect.IO].as(ExitCode.Success)
  }.main(Array())
}
```

Run it and leave it for some time, up to ~10 minutes, with current `fs2-1.0.2` after some time I get

```
Vector(abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc, abc)
[...]

PR #2327, opened 2021-03-14
Faster groupWithin implementation

DGolubets:
I've tried making `groupWithin` as fast as I could. 
I would appreciate a thorough review and any feedback, cos it's my first time writing fs2 stream operator and this particular one is quite complex.

The reason I want it to be faster is because I see `groupWithin` as one of the most important streaming operators. 
It allows me to choose an acceptable latency for a stream whilst still keeping high throughput by batching.
E.g. if I have night spikes of data but very few to none records during a day, I don't won't a couple unlucky records stuck in buffer, and `groupWithin` helps me with it.

I did simple benchmarks on my machine and I see this:

- original `groupWithin` limits any stream to **~34k** per second
- this implementation goes **~190k** per second

This is **x6** improvement.
Maybe some edge cases are not handled. But I think it's worth looking at cos the gain is huge.

Edit: 
the benchmarks I did were with 2.5.x.
I tried v3 and `groupWithin` implementation there is even slower.


diesalbla:
@DGolubets Good evening, Dmitry! Thanks for the PR, we are looking through it. 🙂 

Could you add to the PR the benchmarks you used to get your results? It would help us confirm the results, and FS2 would also benefit from more benchmark coverage.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1413,55 +1412,108 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
-                    val (toEmit, rest) = resize(newAcc, Pull.done)
-                    toEmit >> resetTimerAndGo(rest)
+                  supply.acquire.flatMap { _ =>
+                    supply.available.flatMap { m =>
+                      val k = m.min(n - 1)
+                      supply.tryAcquireN(k).map {
+                        case true  => k + 1
+                        case false => 1
+                      }
+                    }
```
diesalbla, line 1448: Perhaps a for comprehension here?
```suggestion
                  for {
                    _ <- supply.acquire
                    m <- supply.available
                    k  = m.min(n - 1)
                    b <- supply.tryAcquireN(k)
                  } yield if (b) then k + 1 else 1
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1413,55 +1412,108 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+            def endSupply(result: Either[Throwable, Unit]): F2[Unit] =
+              buffer.update(_.copy(endOfSupply = Some(result))) *> supply.releaseN(Int.MaxValue)
+
+            def endDemand(result: Either[Throwable, Unit]): F2[Unit] =
+              buffer.update(_.copy(endOfDemand = Some(result))) *> demand.releaseN(Int.MaxValue)
+
+            val enqueueAsync = F.start {
+              this
+                .evalMap(enqueue)
+                .takeWhile(_ == true)
```
diesalbla, line 1488:
```suggestion
                .forall(_)
```
Because `forall` is a fold, it discards elements (chunks of booleans) sooner.

SystemFw:
> Could you add to the PR the benchmarks you used to get your results? It would help us confirm the results, and FS2 would also benefit from more benchmark coverage.

If not too much of a problem, could this be added as a separate PR? I think it has independent value (and for example I want to see if Queue vs Channel makes a difference to the current implementation)

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1413,55 +1412,108 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+                buffer
+                  .modify { buf =>
+                    (buf.copy(buf.data :+ t), buf)
+                  }
+                  .flatMap { buf =>
+                    supply.release.map { _ =>
+                      buf.endOfDemand.isEmpty
+                    }
+                  }
+              }
```
diesalbla, line 1433: Perhaps a `for` comprehension would help? 
```suggestion
              for { 
                _ <- demand.acquire
                buf <- buffer.modify(buf => (buf.copy(buf.data :+ t), buf))
                _ <- supply.release
              } yield buf.endOfDemand.isEmpty
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1413,55 +1412,108 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
-            val (unit, rest) = c.splitAt(n)
-            resize(rest, s >> Pull.output1(unit))
-          }
+    fs2.Stream.eval(Semaphore[F2](n.toLong)).flatMap { demand =>
+      fs2.Stream.eval(Semaphore[F2](0L)).flatMap { supply =>
+        fs2.Stream
+          .eval(
+            Ref[F2].of(JunctionBuffer[O](Vector.empty[O], endOfSupply = None, endOfDemand = None))
+          )
+          .flatMap { buffer =>
```
diesalbla, line 1421: We sometimes use a `Stream.force` as a way of reducing code nesting.

```suggestion
    fs2.Stream.force { 
      for {
        demand <- Semaphore[F2](n.toLong)
        supply   <- Semaphore[F2](0L)
        buffer    <- Ref[F2].of(JunctionBuffer[O](Vector.empty[O], endOfSupply = None, endOfDemand = None))
      } yield { 
         /* here follows the stream definition*/
      }
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1413,55 +1412,108 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+                  .eval(dequeueN(n))
+                  .repeat
+                  .collectWhile { case Some(data) => Chunk.vector(data) }
+                  .onFinalizeCase {
+                    case ExitCase.Succeeded =>
+                      endDemand(Right(())).flatMap(_ => upstream.cancel)
+                    case ExitCase.Errored(e) =>
+                      endDemand(Left(e)).flatMap(_ => upstream.cancel)
+                    case ExitCase.Canceled =>
+                      endDemand(Right(())).flatMap(_ => upstream.cancel)
```
diesalbla, line 1511: Perhaps we can factor out the `Either[Throwable, Unit]` out? 
```Scala
  val ending: Either[Throwable, Unit] = exitCase match {
    /**/
  }
  endDemand(ending) *> upstream.cancel
```

core/shared/src/main/scala/fs2/JunctionBuffer.scala
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2
```
diesalbla, line 22: Perhaps `fs2.internal` would be a more suitable location.

diesalbla (commented):
A few small code comments.

core/shared/src/main/scala/fs2/JunctionBuffer.scala
SystemFw: I think this belongs inside `groupWithin` to be honest

SystemFw:
Quick warning, it's going to take me a long time to review this PR due to the complexity of the combinator :)

SystemFw:
By the way, I think you have hit a deadlock in two of the CI runs. I'm going to try restarting it to see if it's an accident but unfortunately I don't think so, because it's happened on two separate runs (2.12, 3.0RC1(

DGolubets:
Thanks for the feedback guys! I'll try to address it tomorrow.

> By the way, I think you have hit a deadlock in two of the CI runs.
I noticed a problem with one test under JS target. 
I haven't figured out why yet.

diesalbla:
Coincidentally, there is an open issue regarding `groupWithin` https://github.com/typelevel/fs2/issues/1389, but I do not know  if that got solved in some PR that were merged since then #1711 https://github.com/typelevel/fs2/pull/2062.

DGolubets:
It seems that under JS target the cancellation of `race` "loser" branch does not happen sometimes.
I cannot reproduce it in a trivial test however.. But if I put debug statements in my test I can see that happening every second run.

So the workaround is to always cancel what I acquire in race (essentially it's like "isNavailable" if there was such a method):
```
supply.acquireN(n) *> supply.releaseN(n)
```
But I need to make sure `releaseN` is always called in case acquireN succeeds. Is there an operator for that?

core/shared/src/main/scala/fs2/Stream.scala
Daenyth: ❓ Does the default desugaring give worse peformance? I'm thinking of the way that it passes down tuples rather than closing over an outer scoped val. I don't know how much, if any impact that makes once the JIT does its stuff

nikiforo: @Daenyth, I'm trying to find `better-monadic-for` string in fs2 sources. However, there is none. Do you know the reason why it hasn't been added?

Daenyth: AFAIK it's a general sentiment that it isn't worth the benefit vs deviating from the standard compile. Also it won't work for scala3

DGolubets: There is no noticeable performance impact in my test. So I changed the code based on the suggestions, since it looks prettier indeed.

core/shared/src/main/scala/fs2/Stream.scala
DGolubets: :+1:

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1413,55 +1413,106 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+              case ExitCase.Succeeded  => endSupply(Right(()))
+              case ExitCase.Errored(e) => endSupply(Left(e))
+              case ExitCase.Canceled   => endSupply(Right(()))
+            }
+            .compile
+            .drain
+        }
+
+        fs2.Stream
+          .eval(enqueueAsync)
```
SystemFw, line 1499: if this gets cancelled, you won't interrupt the producer stream

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1413,55 +1413,106 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+          F.guaranteeCase(s.acquireN(n.toLong)) {
+            case Succeeded(_) => s.releaseN(n.toLong)
+            case _            => F.unit
+          }
+
+        def acquireSupplyUpToNWithin(n: Long, timeout: FiniteDuration): F2[Long] =
+          // in JS cancellation doesn't always seem to run, so race conditions should restore state on their own
+          F.race(
+            F.sleep(timeout),
+            waitN(supply)
```
SystemFw, line 1441: I'm having some trouble understanding how this behaves with respect to
```
However, if no elements at all have been accumulated when
    * `timeout` expires, empty chunks are *not* emitted, and `timeout`
    * is not reset.
    * Instead, the next chunk to arrive is emitted immediately (since
    * the stream is still in a timed out state), and only then is
    * `timeout` reset. 
    ```

File: core/shared/src/main/scala/fs2/JunctionBuffer.scala (hunk 6 above)
Comment (DGolubets) [reply core/shared/src/main/scala/fs2/JunctionBuffer.scala]: I moved it to `fs2.internal`. If I recall correctly, placing the class within another will close over it. So it's better to keep it alone to avoid any potential memory leak.

Comment (SystemFw) [reply core/shared/src/main/scala/fs2/JunctionBuffer.scala]: Having a class inside a method does not cause the class to be created there, it gets hoisted out by the compiler

File: core/shared/src/main/scala/fs2/Stream.scala (hunk 7 above)
Comment (DGolubets) [reply core/shared/src/main/scala/fs2/Stream.scala:1499]: Won't that work: `endDemand(ending) *> upstream.cancel`?

Comment (SystemFw) [reply core/shared/src/main/scala/fs2/Stream.scala:1499]: if the stream gets canceled during `eval`, you won't reach that point. The `concurrently` combinator covers this general use case for you. (there is a general meta point here about perf vs correctness, I'm finding it fairly hard to fully reason about the correctness of this implementation, even though I believe it's important enough to justify high performance)

Comment (SystemFw) [reply core/shared/src/main/scala/fs2/Stream.scala:1499]: perhaps you can move the `onFinalize` one level up though

File: core/shared/src/main/scala/fs2/Stream.scala (hunk 8 above)
Comment (DGolubets) [reply core/shared/src/main/scala/fs2/Stream.scala:1441]: If there are no elements `waitN` will lose the race and we go into the left branch. 
There I do `supply.acquire` to make sure at least 1 element is available. 
That's how it should never emit empty.

Comment (SystemFw) [reply core/shared/src/main/scala/fs2/Stream.scala:1441]: yeah but what about not resetting the timeout in that case and making sure that whatever chunk arrives after that gets immediately emitted?

Comment (SystemFw) [reply core/shared/src/main/scala/fs2/Stream.scala:1441]: ah actually I see, good

File: core/shared/src/main/scala/fs2/Stream.scala [hunk 9]
```diff
@@ -1413,55 +1413,106 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+          acquireSupplyUpToNWithin(n.toLong, timeout).flatMap { n =>
+            buffer
+              .modify { buf =>
+                if (buf.data.size >= n) {
+                  val (head, tail) = buf.data.splitAt(n.toInt)
+                  (buf.copy(tail), buf.copy(head))
+                } else {
+                  (buf.copy(Vector.empty), buf)
+                }
+              }
```
diesalbla, line 1464: Could we extract this as a method of `JunctionBuffer`?

```Scala
def splitAt(n: Long): (JunctionBuffer[O], JunctionBuffer[O]) = 
  if (n >= data.size) {
    val (head, tail) = data.splitAt(n.toInt)
    (copy(tail), copy(head))
  } else 
    (copy(Vector.empty), this)
``` 

So that this becomes  `buffer.modify(_.splitAt(n))`

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1413,55 +1413,106 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+            _ <- supply.release
+          } yield buf.endOfDemand.isEmpty
+
+        def waitN(s: Semaphore[F2]) =
+          F.guaranteeCase(s.acquireN(n.toLong)) {
+            case Succeeded(_) => s.releaseN(n.toLong)
+            case _            => F.unit
+          }
+
+        def acquireSupplyUpToNWithin(n: Long, timeout: FiniteDuration): F2[Long] =
```
diesalbla, line 1437: Since this `timeout` is the same as the parameter of the `main` function, we can remove the parameter here.
```suggestion
        def acquireSupplyUpToNWithin(n: Long): F2[Long] =
```

core/shared/src/main/scala/fs2/Stream.scala
DGolubets: Done

core/shared/src/main/scala/fs2/Stream.scala
DGolubets: Done

core/shared/src/main/scala/fs2/Stream.scala
DGolubets: `bracketCase` should do the trick I think

core/shared/src/main/scala/fs2/JunctionBuffer.scala
DGolubets: Good to know, done

mpilquist:
This ready for merge @diesalbla @SystemFw ?

SystemFw:
I think it can be merged, although the logic is rather tricky, so if you wanted to review it @mpilquist , I think a further set of eyes could be useful. If you don't have time, I think we can merge, on the balance.

diesalbla:
The existing test coverage for `groupWithin` seems to be wide enough.

diesalbla (approved):

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1412,56 +1412,115 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+                demand.releaseN(buf.data.size.toLong).flatMap { _ =>
+                  buf.endOfSupply match {
+                    case Some(Left(error)) =>
+                      F.raiseError(error)
+                    case Some(Right(_)) if buf.data.isEmpty =>
+                      F.pure(None)
+                    case _ =>
+                      F.pure(Some(buf.data))
                   }
+                }
```
diesalbla, line 1483: Perhaps a for comprehension here as well? 
```suggestion
          for {
            n <- acquireSupplyUpToNWithin(n.toLong)
            buf <- buffer.modify(_.splitAt(n.toInt))
            _ <- demand.releaseN(buf.data.size.toLong)
            beos = buf.endOfSupply
            _ <- beos.traverse(F.fromEither)  // Raise errors from Some(Left(_)) case 
          } yield beos match {
            case Some(Right(_)) if buf.data.isEmpty => None
            case _ => Some(buf.data)
          }
```

@DGolubets Not a blocker, merge if you like it as is.

DGolubets: @diesalbla I don't think I can: I only have an option to close or comment.

PR #2328, opened 2021-03-15
Add groupWithin benchmark

DGolubets:
FYI it runs OOM on `main`.
On `v2.5.3` it's alright.

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/GroupWithinBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/GroupWithinBenchmark.scala
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2.benchmark
+
+import cats.effect.IO
+import cats.effect.unsafe.implicits.global
+import fs2.Stream
+import org.openjdk.jmh.annotations.{Benchmark, OperationsPerInvocation, Scope, State}
+
+import scala.concurrent.duration._
+
+object GroupWithinBenchmark {
+  final val N = 1000000
+  final val bufferSize = 256
+  final val bufferWindow = 4.seconds
+}
+
+@State(Scope.Thread)
+class GroupWithinBenchmark {
+
+  import GroupWithinBenchmark._
+
+  @Benchmark
+  @OperationsPerInvocation(N)
+  def groupWithin(): Unit =
+    Stream
+      .range(0, N)
+      .covary[IO]
+      .groupWithin(bufferSize, bufferWindow)
+      .compile
+      .drain
+      .unsafeRunSync
+}
--- a/benchmark/src/main/scala/fs2/benchmark/GroupWithinBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/GroupWithinBenchmark.scala
@@ -19,34 +19,44 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-package fs2.benchmark
+package fs2
+package benchmark
 
 import cats.effect.IO
-import cats.effect.unsafe.implicits.global
-import fs2.Stream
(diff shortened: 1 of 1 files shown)


SystemFw:
> FYI it runs OOM 

great, so we might have a memory leak with it again

diesalbla (approved):

diesalbla:
Bisecting back, we found out that the leak was introduced in https://github.com/typelevel/fs2/commit/1ea8e426c5a5a3c7c24a0ea2e250c8007b6fa433. 

Undoing some of those changes seems to have fixed that memory leak. Could you cherry-pick and confirm?

DGolubets:
@diesalbla sorry I'm too late. I've tried with the last change you've committed to `main` and there is no leak anymore indeed.

SystemFw:
Out of curiosity, what does the benchmark say for the 3.0 version of `groupWithin`? Again, I'm ok using this implementation for `groupWithin` specifically if it's much faster (since `groupWithin` helps you chunking, which then helps you with the performance of every other combinator), but perhaps we also improve the performance of `TimedPull` for things users write which aren't `groupWithin`

DGolubets:
This benchmark shows ~16k on `main`, compared to ~82k on `2.5.x` and ~700k on my branch.

It would be nice to have a generalized approach for building concurrent steps in easy manner, which is the idea behind `TimedPull` I suppose. But it looks something is not right with it at the moment, performance wise. I only had a quick glance at it and couldn't figure out the inner workings.

nikiforo:
> Out of curiosity, what does the benchmark say for the 3.0 version of `groupWithin`? Again, I'm ok using this implementation for `groupWithin` specifically if it's much faster (since `groupWithin` helps you chunking, which then helps you with the performance of every other combinator), but perhaps we also improve the performance of `TimedPull` for things users write which aren't `groupWithin`
Do you think that current groupWithin implementation is limited by TimedPull performance?

SystemFw:
> Do you think that current groupWithin implementation is limited by TimedPull performance?

I actually don't know

diesalbla:
This morning I have taken a look at this benchmark. I have modified it in the following ways: 
- Making the size of the input stream range (all singleton chunks) a parameter, using sizes 100, 10_000, and 100_000. 
- Making the size of the window into another parameter,  with values 16, 256, 4096 
- Reducing that `bufferWindow` time to 1 millisecond. This is mostly for practical purposes, to let the examples finish quick enough. Although 1ms is small for persons, it is still large enough for a computer (even 100 µs could be enough).

```Scala
package fs2.
package benchmark

import cats.effect.IO
import org.openjdk.jmh.annotations.{Benchmark, Scope, State, Param}

@State(Scope.Thread)
class GroupWithinBenchmark {

/*  For Branch 2.5.x 
  import cats.effect.{Timer, ContextShift}
  lazy protected implicit val ioTimer: Timer[IO] = IO.timer(ExecutionContext.global)

  lazy protected implicit val ioContextShift: ContextShift[IO] =
    IO.contextShift(ExecutionContext.global)
*/

  // for branch 3.x

  val bufferWindow = 1.millis

  @Param(Array("100", "10000", "100000"))
  var rangeLength: Int = _

  @Param(Array("16", "256", "4096"))
  var bufferSize: Int = _

  @Benchmark
  def groupWithin(): Unit =
    Stream
      .range(0, rangeLength)
      .covary[IO]
      .groupWithin(bufferSize, bufferWindow)
      .compile
      .drain
      .unsafeRunSync()
}
```

The code also marks the sections that were for branch 2.5.x (with `cats-effect`) and for branch 3..x. I have compared three versions with this benchmark. 

- The HEAD of current `main` branch, which includes the `Channel` PR. 
- The HEAD of current `2.5.x` branch
- The new implementation of `groupWithin` suggested by @DGolubets in #2327, on top of `main` (rebased).

These three versions I run on a Macbook Pro laptop, with 2.4 GHz Quad-Core Intel Core i5 CPU, using the following JDK: 
```
openjdk version "11.0.10" 2021-01-19
OpenJDK Runtime Environment GraalVM CE 21.0.0.2 (build 11.0.10+8-jvmci-21.0-b06)
OpenJDK 64-Bit Server VM GraalVM CE 21.0.0.2 (build 11.0.10+8-jvmci-21.0-b06, mixed mode, sharing)
```

To run these benchmarks, I used the following command on an open SBT console: 
```
jmh:run -i 10 -wi 3 -f1 -t2 -gc true -jvmArgs "-XX:MaxRecursiveInlineLevel=3 -XX:MaxInlineSize=50" .*GroupWithin.*
```
The `-gc true` in particular is to force a GC between executions, given that this is a kind of "garbage" or resource that each program would leave behind.

These three executions took me between 20 to 30 minutes each, and had the following results:



### On `main` current

```
  (bufferSize)  (rangeLength)  Cnt    Score    Error  Units
            16            100   10  464.805 ±  7.930  ops/s
            16          10000   10    4.652 ±  0.536  ops/s
            16         100000   10    0.436 ±  0.002  ops/s
           256            100   10  482.793 ±  2.501  ops/s
           256          10000   10    4.492 ±  0.075  ops/s
           256         100000   10    0.407 ±  0.020  ops/s
          4096            100   10  447.112 ± 39.356  ops/s
          4096          10000   10    5.057 ±  0.021  ops/s
          4096         100000   10    0.443 ±  0.037  ops/s
```

### On 2.5.x series.

```
 (bufferSize)  (rangeLength)  Cnt    Score    Error  Units
            16            100   10  893.203 ±  8.037  ops/s
            16          10000   10   10.043 ±  0.062  ops/s
            16         100000   10    0.966 ±  0.019  ops/s
           256            100   10  970.414 ±  9.908  ops/s
           256          10000   10   10.126 ±  0.421  ops/s
           256         100000   10    1.037 ±  0.022  ops/s
          4096            100   10  906.796 ± 76.104  ops/s
          4096          10000   10    9.740 ±  0.784  ops/s
          4096         100000   10    1.021 ±  0.003  ops/s
```

### Using new implementation of groupWithin

```
 (bufferSize)  (rangeLength)  Cnt      Score      Error  Units
            16            100   10   7793.901 ±   22.708  ops/s
            16          10000   10    101.716 ±    0.427  ops/s
            16         100000   10      9.890 ±    0.267  ops/s
           256            100   10  11500.278 ±  705.942  ops/s
           256          10000   10    167.005 ±    1.355  ops/s
           256         100000   10     17.122 ±    0.218  ops/s
          4096            100   10  11133.196 ± 1115.649  ops/s
          4096          10000   10    170.064 ±    0.543  ops/s
          4096         100000   10     16.141 ±    2.400  ops/s
```

SystemFw:
@diesalbla Want to push the improvements to the benchmark?

diesalbla:
@DGolubets I have pushed some changes to the benchmark.

diesalbla:
@mpilquist @SystemFw Are there any other dimensions we should parameterise these benchmarks? The size of the input chunks, perhaps? Otherwise, this should be ready to merge.

PR #2333, opened 2021-03-20
Don't lose precision in fixedRate

SystemFw:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3145,30 +3145,29 @@ object Stream extends StreamLowPriority {
...
-          else Stream.emit(()).repeatN(cnt)
-        out ++ fixedRate_(periodMillis, dampen, next)
-      } else {
-        val toSleep = next - now
-        Stream.sleep_(toSleep.millis) ++ Stream.emit(()) ++ fixedRate_(periodMillis, dampen, next)
+    if (period.toNanos == 0) Stream(()).repeat
+    else
+      Stream.eval(Temporal[F].monotonic).flatMap { now =>
+        val next = t + period
+        if (next <= now) {
```
diesalbla, line 3159: Could we factor out the recursive call to `fixedRate_`? 
```Scala
val step = if (next <= now) { /***/

step ++ fixedRate_(period, next, dampen)
```

diesalbla (approved):

PR #2340, opened 2021-03-27
Remove balance (and PubSub)

SystemFw:
`balance` sees little use, and it's the only remaining usage of PubSub, which we want to remove.
This PR removes it, to unblock 3.0.
A new specialised implementation of this combinator can be readded later.


diesalbla (approved):

matil019:
I stumbled on this while migrating to fs2-3. It turned out that `Stream#parEvalMapUnordered` was sufficient for my use case. Just my two cents for other developers facing the same problem as me.

PR #2341, opened 2021-03-28
Remove some uses of Sync constraint, fixes #2338

### LINKED ISSUE #2338: Make .showLinesStdOut and friends use Console instead of Sync?
CE3 introduces a `Console[F]` algebra - should we use that instead of Sync in `showLines` and `showLinesStdOut`? What about `.debug`?

mpilquist:
This PR removes `lines`, `showLines`, and `showLinesStdOut` and replaces them with `printlns`.

`Stream.random` was also updated to use `cats.effect.std.Random`.

The only uses of `Sync` constraints left are:
1) Creating a stream from an interator
2) Compression package

Thoughts on updating compression package to use a capability trait?

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3306,7 +3306,7 @@ object Stream extends StreamLowPriority {
     * }}}
     */
   def range[F[x] >: Pure[x], O: Numeric](start: O, stopExclusive: O): Stream[F, O] =
-    range(start, stopExclusive, Numeric[O].one)
+    range(start, stopExclusive, implicitly[Numeric[O]].one)
 
   /** Lazily produce the sequence `[start, start + step, start + 2 * step, ..., stopExclusive)`.
     * If you want to produce the sequence in one chunk, instead of lazily, use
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3307,11 +3298,8 @@ object Stream extends StreamLowPriority {
 
   /** Creates a random stream of integers using a random seed.
     */
-  def random[F[_]](implicit F: Sync[F]): Stream[F, Int] =
-    Stream.eval(F.delay(new scala.util.Random())).flatMap { r =>
-      def go: Stream[F, Int] = Stream.emit(r.nextInt()) ++ go
-      go
-    }
+  def random[F[_]](implicit F: Random[F]): Stream[F, Int] =
+    Stream.repeatEval(F.nextInt)
```
diesalbla, line 3302: I wonder if this is a worthy combinator, rather than something best left to users of `fs2` to run with the combinators.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2057,6 +2056,13 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       }
     }
 
+  /** Prints each element of this stream to standard out, converting each element to a `String` via `Show`. */
+  def printlns[F2[x] >: F[x], O2 >: O](implicit
+      F: Console[F2],
+      showO: Show[O2] = Show.fromToString[O2]
+  ): Stream[F2, INothing] =
+    covaryAll[F2, O2].map(_.show).foreach(F.println)
```
diesalbla, line 2064: Is this method worth keeping, or is this something easilly accessible by the existing combinators? 

If we keep it, perhaps we should save the middle chunks: 
```suggestion
    foreach( (o: O2) => F.println(o.show))
```

diesalbla (approved):

PR #2344, opened 2021-03-29
FS2-Core: use lower dependencies

diesalbla:
Since cats has split into modules, and we do not always use IO, we can instead depend on the Kernel (for typeclasses) and std (for the semaphores and refs)


kubukoz:
What do you plan to do about SyncIO?

diesalbla:
> What do you plan to do about SyncIO? 

Askin in Gitter... but perhaps either remove those `Compiler` instances, or look for a typeclass constraint for it.

diesalbla:
> I think we've discussed this before, and established that it's a non goal to not depend on core. @mpilquist  might remember more

From the Gitter channel, by @SystemFw.

PR #2348, opened 2021-03-30
Upgrade 2.5.x to Scala 3 RC2

mpilquist:


kubukoz:
hmm no jobs here?

Should we add `series/2.5.x` to the branch list?

mpilquist:
Not sure why, it's setup that way...

PR #2354, opened 2021-04-01
Make repeated maps stack-safe using AndThen

TimWSpence:


core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -445,6 +445,13 @@ class StreamSuite extends Fs2Suite {
 
       assertEquals(unfoldTree(1).flatten.take(10).toList, List.tabulate(10)(_ + 1))
     }
+
+    test("regression #2353 - stack safety of map") {
+      def loop(str: Stream[Pure, Int], i: Int): Stream[Pure, Int] =
```
kubukoz, line 450: let's add `@tailrec`

mpilquist:
Awesome! We'll need mima exclusions but otherwise this looks good to go.

TimWSpence:
@mpilquist are those exclusions ok? I'm actually wondering if we might be able to get away with keeping the field type as `Function1` anyway and just constructing `AndThen` in `mapOutput`. It's a subtype so should be ok

mpilquist:
Yep they look good. Thanks!

PR #2358, opened 2021-04-04
Add a note to handle client socket stream errors.

ScalaWilliam:
My TLS server died because somebody connected to it via plaintext; this was because of no error handling at the client socket level, and @mpilquist suggested I make a PR to warn users of this.

Also I also [wrote up a short retro](https://www.scalawilliam.com/letsencrypt-fs2/) about my experience 😀


mpilquist:
Thanks!

PR #2361, opened 2021-04-05
BufferBy - Rewrite for association

diesalbla:
For Pull unrolling and evaluation, it is better to prefer right associativity, and to keep a simpler `Pull` at left hand side of
top `>>`. To do this, we change `dumpBuffer` to take the "tail" as a lazy parameter. Also, we can use a mutable  `ListBuffer` to do FIFO order while at the same time avoiding the need to do a `reverse`.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -38,7 +38,7 @@ import fs2.compat._
 import fs2.concurrent._
 import fs2.internal._
 
-import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable.{ArrayBuffer, ListBuffer}
 
 /** A stream producing output of type `O` and which may evaluate `F` effects.
   *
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -312,28 +312,31 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
     */
   def bufferBy(f: O => Boolean): Stream[F, O] = {
-    def dumpBuffer(bb: List[Chunk[O]]): Pull[F, O, Unit] =
-      bb.reverse.foldLeft(Pull.done: Pull[F, O, Unit])((acc, c) => acc >> Pull.output(c))
+    @inline
+    def dumpBuffer(bb: ListBuffer[Chunk[O]], rest: => Pull[F, O, Unit]): Pull[F, O, Unit] =
+      if (bb.isEmpty) rest else Pull.output(bb.head) >> dumpBuffer(bb.tail, rest)
 
-    def go(buffer: List[Chunk[O]], last: Boolean, s: Stream[F, O]): Pull[F, O, Unit] =
+    def go(buffer: ListBuffer[Chunk[O]], last: Boolean, s: Stream[F, O]): Pull[F, O, Unit] =
```
nikiforo, line 319: AFAIK, half a year ago Chunk was redesigned to achieve effective splitAt & merge. Shouldn't we try to write this method with Chunk & Chunk.Queue, but without Buffers & Vectors?

mpilquist: @diesalbla Any thoughts on updating this to use Chunk or Chunk.Queue? I vaguely remember trying once before and not liking the result. I'm curious about the use of ListBuffer though.

PR #2367, opened 2021-04-19
Use a sliced chunk instead of slicing an array

vasilmkd:
Just a small detail.

### DIFF
--- a/io/src/main/scala/fs2/io/io.scala
+++ b/io/src/main/scala/fs2/io/io.scala
@@ -74,7 +74,7 @@ package object io {
     F.blocking(is.read(buf)).map { numBytes =>
       if (numBytes < 0) None
       else if (numBytes == 0) Some(Chunk.empty)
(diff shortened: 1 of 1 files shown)


diesalbla (approved):

io/src/main/scala/fs2/io/io.scala
```diff
@@ -74,7 +74,7 @@ package object io {
     F.blocking(is.read(buf)).map { numBytes =>
       if (numBytes < 0) None
       else if (numBytes == 0) Some(Chunk.empty)
-      else if (numBytes < buf.size) Some(Chunk.array(buf.slice(0, numBytes)))
```
mpilquist, line 77: I think this may have been leftover from when the `buf` was shared across multiple calls, making a copy necessary.

vasilmkd: But is the change correct?

mpilquist: Yep, b/c the only use of this private method is one which allocates a fresh buffer on each invocation

PR #2368, opened 2021-04-19
loopEither - a more generic method of loop

nikiforo:
I don't have a strong opinion whether this function should exist in FS2. Thus, I submit this to hear different views/opinions.

Personally I wrote it for error handling, but I think that my case can be implemented with https://github.com/cb372/cats-retry.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -414,11 +414,17 @@ object Pull extends PullLowPriority {
   /** Repeatedly uses the output of the pull as input for the next step of the
     * pull. Halts when a step terminates with `None` or `Pull.raiseError`.
     */
-  def loop[F[_], O, R](f: R => Pull[F, O, Option[R]]): R => Pull[F, O, Unit] = {
-    val ef = (r: R) => f(r).map(_.fold(().asRight[R])(_.asLeft))
-    loopEither(ef)
-  }
+  def loop[F[_], O, R](f: R => Pull[F, O, Option[R]]): R => Pull[F, O, Unit] =
+    (r: R) =>
+      f(r).flatMap {
+        case None    => Pull.done
+        case Some(s) => loop(f)(s)
+      }
 
+  /** Intantiates with a state. Repeatedly uses the left value of the result of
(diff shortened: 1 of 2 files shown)


core/shared/src/test/scala/fs2/PullSuite.scala
```diff
@@ -50,4 +50,39 @@ class PullSuite extends Fs2Suite {
       }
     }
   }
+
+  property("loop") {
+    forAll { (list: List[Int]) =>
+      def customOutput(s: List[Int]): Pull[Fallible, Int, Option[List[Int]]] =
```
nikiforo, line 56: `Pure` is enough here

core/shared/src/test/scala/fs2/PullSuite.scala
```diff
@@ -50,4 +50,39 @@ class PullSuite extends Fs2Suite {
...
+
+      val result = Pull.loop(customOutput)(list).stream.compile.toList
+
+      assertEquals(result, Right(list))
+    }
+  }
+
+  property("loopEither") {
+    forAll { (list: List[Int]) =>
+      def customOutputReturnsString(s: List[Int]): Pull[Fallible, Int, Either[List[Int], String]] =
```
nikiforo, line 70: `Pure` is enough here

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -414,11 +414,16 @@ object Pull extends PullLowPriority {
...
     */
-  def loop[F[_], O, R](f: R => Pull[F, O, Option[R]]): R => Pull[F, O, Unit] =
-    (r: R) =>
-      f(r).flatMap {
-        case None    => Pull.done
-        case Some(s) => loop(f)(s)
+  def loop[F[_], O, R](f: R => Pull[F, O, Option[R]]): R => Pull[F, O, Unit] = {
+    val ef = (r: R) => f(r).map(_.fold(().asRight[R])(_.asLeft))
+    loopEither(ef)
+  }
```
diesalbla, line 420: I would prefer to keep `loop` as defined on its own.

mpilquist:
Looks good to me - let's keep original definition of `loop` like suggested by @diesalbla and add ScalaDoc.

PR #2369, opened 2021-04-19
Added `meteredStartImmediately` 

### LINKED ISSUE #1966: A variant of debounce/metered that begins immediately
I found it surprising that metered/debounce doesn't immediately "start". Think of debouncing a button - You probably want to react immediately to the event instead of waiting for the whole debounce period to finish. Similarly for things like polling, you want to start immediately.

I can achieve this with a `emit` merged with `fixedRate`, but I feel like this is a common enough pattern for it to be an alternative method? e.g. `meteredStartImmediate`

Or at least documented. WDYT?

```
import cats.effect._
import fs2._

object FixmeFs2Main extends IOApp {
    (Stream.eval(IO { println("hello") }) ++
      Stream
        .emit(())
        .mergeHaltR(Stream.fixedRate(1.seconds))
        .evalTap(_ => IO { println("HI") })).compile.drain
      .map(_ => ExitCode.Success)
}
```
^ @SystemFw FYI. I think this works.

hogiyogi597:
Closes #1966 

Added `meteredStartImmediately` after seeing issue #1966 and running into something similar where I was expecting the metered stream to start immediately and at every `rate` afterwards.


core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
```diff
@@ -1018,6 +1017,30 @@ class StreamCombinatorsSuite extends Fs2Suite {
...
+  test("metered should not start immediately") {
+    Stream
+      .emit[IO, Int](1)
+      .repeatN(10)
+      .metered(1.second)
+      .interruptAfter(500.milliseconds)
+      .compile
+      .toList
+      .map(results => assert(results.isEmpty))
+      .unsafeRunAndForget()
```
mpilquist, line 1029: No need to run these here -- the test framework will do so:

```suggestion
```

core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
```diff
@@ -1018,6 +1017,30 @@ class StreamCombinatorsSuite extends Fs2Suite {
...
+  test("meteredStartImmediately should start immediately") {
+    Stream
+      .emit[IO, Int](1)
+      .repeatN(10)
+      .meteredStartImmediately(1.second)
+      .interruptAfter(500.milliseconds)
+      .compile
+      .toList
+      .map(results => assert(results.size == 1))
+      .unsafeRunAndForget()
```
mpilquist, line 1041:
```suggestion
```

mpilquist (approved):

diesalbla (approved):

PR #2376, opened 2021-04-28
Unix Socket Implementation

ChristopherDavenport:
I think its adapted to the newer fs2 interface. Not sure if what I did is correct at all for the local and remote address.


unixsocket/src/main/scala/fs2/io/unixsocket/UnixSocket.scala
```diff
@@ -0,0 +1,226 @@
...
+          val sock = serverChannel.socket()
+          sock.bind(address)
+          serverChannel
+        }
+      )
+
+    def cleanup(sch: UnixServerSocketChannel): F[Unit] =
+      F.blocking {
+        if (sch.isOpen) sch.close()
+        if (sch.isRegistered()) println("Server Still Registered")
```
ChristopherDavenport, line 68: Not sure how you want to manage if this occurs. It shouldn't, but it has.

Also could delete when we are done if this is not the case.

mpilquist:
I pushed an update but there's still some work to do:
- Tests
- Deciding on localAddress/remoteAddress behavior -- I'm leaning towards `F.raiseError(new UnsupportedOperationException)`
- Checking if cleanup really works -- when using this at the REPL, I noticed I could only use a given path once as a server socket, even across JVM restarts. Makes me wonder if some file is sticking around.

mpilquist:
Alright pushed some cleanup -- @ChristopherDavenport let me know what you think.

PR #2378, opened 2021-04-29
Add a test for watching multiple files - passes on OS X

mpilquist:


djspiewak:
Should probably consider adding macOS and windows builds to the matrix.

PR #2382, opened 2021-05-01
Fix #2375 - bug where multiple watches on paths which map to same wat…

### LINKED ISSUE #2375: [fs2-io] Watcher only remembers last file it was asked to watch
fs2 3.0.2

The way I understood how `Watcher` can be used is:

As long as the watcher resource is open, calling `w.watch(path)` (where **path** is a **file**) on it should **add** files to the list of watches. And `w.events()` would be a stream of events coming from both files.

What I think is happening instead, is only the **last** path added via `w.watch(path)` receives notifications.

Here's a reproduction as a self-contained ammonite script (alternatively, here it is in a Scastie: https://scastie.scala-lang.org/KiUg7NdURt6J7ULU8R3CUw) :

```scala
import $ivy.`co.fs2::fs2-core:3.0.2`
import $ivy.`co.fs2::fs2-io:3.0.2`

import cats.effect.kernel.Resource
import fs2.io.file.Files
import cats.effect._
import cats.syntax.all._
import fs2.io.Watcher
import java.nio.file.Paths

@main def test() = {

  val monster =
    (
      Files[IO].tempDirectory(),
      Watcher.default[IO],
      Resource.eval(IO.ref(List.empty[Watcher.Event]))
    ).parTupled
      .flatMap { case (temp, w, ref) =>
        val file1 = Paths.get(temp.toAbsolutePath().toString, "file1")
        val file2 = Paths.get(temp.toAbsolutePath().toString, "file2")

        val pump = w
          .events()
          .evalMap(ev => ref.update(ev +: _))
          .compile
          .drain
          .background
          .void

        pump.evalMap { _ =>
          w.watch(file2) *>
            w.watch(file1) *>
            IO.sleep(100.millis) *> // just in case
            IO.blocking(os.write(os.Path(file1), "hello file1")) *>
            IO.blocking(os.write(os.Path(file2), "hello file2")) *>
            IO.sleep(100.millis) *> // just in case
            ref.get.flatMap(IO.println)
        }

      }

  monster.use_.unsafeRunSync()
}
```

It outputs 

```
List(Modified(/tmp/17267169841305025399/file1,1), Created(/tmp/17267169841305025399/file1,1))
```

whereas I would expect events from both paths.

If you switch the order of `w.watch` calls, you'll see notifications for the other file.

If you replace both calls with one `w.watch(temp)` (containing folder), you'll see notifications for both files

mpilquist:
…ch key result in loss of events for all but last registered path

### DIFF
--- a/io/src/main/scala/fs2/io/file/Watcher.scala
+++ b/io/src/main/scala/fs2/io/file/Watcher.scala
@@ -29,7 +29,6 @@ import cats.syntax.all._
 import java.nio.file._
 import java.nio.file.attribute.BasicFileAttributes
 import java.util.concurrent.TimeUnit
-import fs2.concurrent.SignallingRef
 
 import CollectionCompat._
 
@@ -163,9 +162,11 @@ object Watcher {
     })
 
   private case class Registration[F[_]](
+      path: Path,
+      singleFile: Boolean,
+      key: WatchKey,
       types: Seq[EventType],
       modifiers: Seq[WatchEvent.Modifier],
-      eventPredicate: Event => Boolean,
       recurse: Boolean,
       suppressCreated: Boolean,
       cleanup: F[Unit]
@@ -175,25 +176,28 @@ object Watcher {
   def fromWatchService[F[_]](
       ws: WatchService
   )(implicit F: Async[F]): F[Watcher[F]] =
-    SignallingRef
-      .of[F, Map[WatchKey, Registration[F]]](Map.empty)
+    Ref
+      .of[F, List[Registration[F]]](Nil)
       .map(new DefaultWatcher(ws, _))
 
   private class DefaultWatcher[F[_]](
       ws: WatchService,
-      registrations: SignallingRef[F, Map[WatchKey, Registration[F]]]
+      registrations: Ref[F, List[Registration[F]]]
   )(implicit
       F: Async[F]
   ) extends Watcher[F] {
     private def isDir(p: Path): F[Boolean] =
       F.blocking(Files.isDirectory(p))
 
-    private def track(key: WatchKey, r: Registration[F]): F[F[Unit]] =
+    private def track(r: Registration[F]): F[F[Unit]] =
       registrations
-        .update(_.updated(key, r))
+        .update(rs => r :: rs)
         .as {
-          F.blocking(key.cancel) >> registrations.modify { s =>
-            (s - key) -> s.get(key).map(_.cleanup).getOrElse(F.unit)
+          registrations.modify { s =>
+            val filtered = s.filterNot(_ == r)
+            val cleanup = s.find(_ == r).map(_.cleanup).getOrElse(F.unit)
+            val cancelKey = if (filtered.forall(_.key != r.key)) F.blocking(r.key.cancel) else F.unit
+            filtered -> (cancelKey *> cleanup)
           }.flatten
         }
 
@@ -234,11 +238,10 @@ object Watcher {
         dirs
       }
       dirs.flatMap(
-        _.traverse(
-          registerUntracked(_, supplementedTypes, modifiers).flatMap(key =>
+        _.traverse(path =>
+          registerUntracked(path, supplementedTypes, modifiers).flatMap(key =>
             track(
-              key,
-              Registration(supplementedTypes, modifiers, _ => true, true, suppressCreated, F.unit)
+              Registration(path, false, key, supplementedTypes, modifiers, true, suppressCreated, F.unit)
             )
           )
         ).map(_.sequence.void)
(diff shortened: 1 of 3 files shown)


ChristopherDavenport (commented):
Another really great change!

io/src/main/scala/fs2/io/file/Watcher.scala
```diff
@@ -288,45 +302,50 @@ object Watcher {
...
+          val regs = registrations.filter(_.key == key)
+          val filteredEvents = events.filter { e =>
+            regs.foldLeft(false) { (acc, reg) =>
+              acc || {
+                val singleFileMatch = reg.singleFile && Event.pathOf(e) == Some(reg.path)
+                val dirMatch =
+                  !reg.singleFile && !(e.isInstanceOf[Event.Created] && reg.suppressCreated)
+                singleFileMatch || dirMatch
+              }
+            }
```
vasilmkd, line 316: This looks like `exists` to me. Would that be clearer?

diesalbla: It also saves some allocations.

io/src/main/scala/fs2/io/file/Watcher.scala
```diff
@@ -175,25 +176,29 @@ object Watcher {
...
+    private def track(r: Registration[F]): F[F[Unit]] =
       registrations
-        .update(_.updated(key, r))
+        .update(rs => r :: rs)
         .as {
-          F.blocking(key.cancel) >> registrations.modify { s =>
-            (s - key) -> s.get(key).map(_.cleanup).getOrElse(F.unit)
+          registrations.modify { s =>
+            val filtered = s.filterNot(_ == r)
+            val cleanup = s.find(_ == r).map(_.cleanup).getOrElse(F.unit)
```
diesalbla, line 198:
```suggestion
            val cleanup = s.collectFirst { case x if x == r  => x.cleanup }.sequence
```

vasilmkd: This is great.

io/src/main/scala/fs2/io/file/Watcher.scala
```diff
@@ -288,45 +302,50 @@ object Watcher {
...
-                        .getOrElse(m)
-                    )
-                    .void
+                  val updateRegistration: F[Unit] = this.registrations.update { m =>
+                    val idx = m.indexWhere(_.key == key) // TODO indexOf(_)
+                    if (idx >= 0) {
+                      val existing = m(idx)
+                      m.updated(idx, existing.copy(cleanup = existing.cleanup >> cancelAll))
+                    } else m
+                  }.void
```
vasilmkd, line 348: `cancelAll` can be moved inside the `if` statement to avoid an allocation when it is not necessary to evaluate that branch.

vasilmkd, line 348: This should help avoid traversing the list three times.

```scala
val updateRegistration: F[Unit] = this.registrations.update { m =>
                    m.collect {
                      case existing if existing.key == key =>
                        val cancelAll: F[Unit] = cancels.sequence.void
                        existing.copy(cleanup = existing.cleanup >> cancelAll)
                    }
                  }.void
```

io/src/main/scala/fs2/io/file/Watcher.scala
```diff
@@ -288,45 +302,50 @@ object Watcher {
...
+                    watch(
+                      p,
+                      Seq(EventType.Created),
+                      regs.headOption.map(_.modifiers).getOrElse(Nil)
+                    ).flatMap { cancel =>
+                      val events: F[List[Event]] = F.blocking {
+                        var dirs: List[Path] = Nil
+                        Files.list(p).forEach(d => dirs = d :: dirs)
+                        dirs.map(Event.Created(_, 1))
+                      }
```
vasilmkd, line 333:
```suggestion
                      val events: F[List[Event]] = F.blocking {
                        var evs: List[Event.Created] = Nil
                        Files.list(p).forEach(d => evs = Event.Created(d, 1) :: evs)
                        evs
                      }
```

io/src/main/scala/fs2/io/file/Watcher.scala
```diff
@@ -288,45 +302,48 @@ object Watcher {
...
-                    )
-                    .void
+                  val updateRegistration: F[Unit] = this.registrations.update { m =>
+                    val idx = m.indexWhere(_.key == key)
+                    if (idx >= 0) {
+                      val existing = m(idx)
+                      val cancelAll: F[Unit] = cancels.sequence.void
+                      m.updated(idx, existing.copy(cleanup = existing.cleanup >> cancelAll))
+                    } else m
+                  }.void
```
vasilmkd, line 346:
```suggestion
                  val updateRegistration: F[Unit] = this.registrations.update { m =>
                    m.collect {
                      case existing if existing.key == key =>
                        val cancelAll: F[Unit] = cancels.sequence.void
                        existing.copy(cleanup = existing.cleanup >> cancelAll)
                    }
                  }.void
```

mpilquist: Don't we need the fall through case there too? As is, we'd filter out all the entries for which the key doesn't match, right?

vasilmkd: Oh wow, you're completely right, I'm so sorry for the confusion.

PR #2383, opened 2021-05-03
Reimplement PipedInput/OutputStream

### LINKED ISSUE #2037: readOutputStream binds the OutputStream to the writing thread's lifetime
Consider this stream (FS2 2.4.4):
```scala
readOutputStream(blocker, 1024) { os =>
  blocker.delay[IO, Unit] {
    val t = new Thread(
      () => os.write(123)
    )
    t.start
    t.join
    Thread.sleep(1000)
  }
}
```
When run (OpenJDK 11), it produces this exception:
```
java.io.IOException: Write end dead
	at java.base/java.io.PipedInputStream.read(PipedInputStream.java:310)
	at java.base/java.io.PipedInputStream.read(PipedInputStream.java:377)
	at java.base/java.io.InputStream.read(InputStream.java:205)
	at fs2.io.package$.$anonfun$readBytesFromInputStream$1(io.scala:77)
...
```
This is caused by an absolutely ridiculous `PipedInputStream` gotcha - it throws an error if the last `Thread` to write to the `PipedOutputStream` has completed without closing the stream! This is a 20 year old [wontfix](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-4028322). :facepalm: 

Tying the `OutputStream` to a particular thread is extremely confusing, even more so in the Cats/FS2 context. We should fix this - or at the very least add a warning to the documentation and to the scaladoc.

vasilmkd:
Resolves #2037. I included the original reproduction as a test.

This was not pleasant to write.

### DIFF
--- a/io/src/main/scala/fs2/io/internal/InputOutputBuffer.scala
+++ b/io/src/main/scala/fs2/io/internal/InputOutputBuffer.scala
@@ -24,6 +24,23 @@ package fs2.io.internal
 import java.io.{InputStream, OutputStream}
 import java.util.concurrent.Semaphore
 
+/** Thread safe circular byte buffer which connects a [[java.io.OutputStream]]
+  * to a [[java.io.OutputStream]] in a memory efficient manner, without copying
+  * bytes unnecessarily.
+  *
+  * @note As per the interfaces of the [[java.io]] classes, all of the
+  * operations are blocking in nature and extra care should be taken when using
+  * the exposed input/output streams. Thread safety is ensured by
+  * synchronizing on individual objects of this class.
+  *
+  * This is, in spirit, a clean room reimplementation of the
+  * [[java.io.PipedInputStream]] and [[java.io.PipedOutputStream]] pair of
+  * classes which can be used to achieve similar functionality, without the
+  * thread bookkeeping which is confusing in a multi threaded environment like
+  * the effect systems in which this code runs.
+  *
+  * @param capacity the capacity of the allocated circular buffer
+  */
 private[io] final class InputOutputBuffer(private[this] val capacity: Int) { self =>
 
   private[this] val buffer: Array[Byte] = new Array(capacity)
--- a/io/src/main/scala/fs2/io/internal/InputOutputBuffer.scala
+++ b/io/src/main/scala/fs2/io/internal/InputOutputBuffer.scala
@@ -105,6 +105,10 @@ private[io] final class InputOutputBuffer(private[this] val capacity: Int) { sel
       readerPermit.release()
       writerPermit.release()
     }
+
+    override def available(): Int = self.synchronized {
+      if (closed) 0 else tail - head
+    }
   }
 
   val outputStream: OutputStream = new OutputStream {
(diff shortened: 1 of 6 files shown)


vasilmkd:
The failing tests are already solved by #2373.

io/src/main/scala/fs2/io/internal/InputOutputBuffer.scala
```diff
@@ -0,0 +1,277 @@
...
+ */
+
+package fs2.io.internal
+
+import java.io.{InputStream, OutputStream}
+import java.util.concurrent.Semaphore
+
+/** Thread safe circular byte buffer which connects a [[java.io.OutputStream]]
+  * to a [[java.io.OutputStream]] in a memory efficient manner, without copying
+  * bytes unnecessarily.
```
diesalbla, line 29: Would one of this actually be an _input_ stream?

vasilmkd: You're right, good catch! 😆
Fixed.

io/src/main/scala/fs2/io/internal/InputOutputBuffer.scala
```diff
@@ -0,0 +1,277 @@
...
+  *
+  * This is, in spirit, a clean room reimplementation of the
+  * [[java.io.PipedInputStream]] and [[java.io.PipedOutputStream]] pair of
+  * classes which can be used to achieve similar functionality, without the
+  * thread bookkeeping which is confusing in a multi threaded environment like
+  * the effect systems in which this code runs.
+  *
+  * @param capacity the capacity of the allocated circular buffer
+  */
+private[io] final class InputOutputBuffer(private[this] val capacity: Int) { self =>
```
diesalbla, line 44: One general question: given that this entire class only depends on Java libraries, could it be submitted for the Scala standard library? Sys or process.

io/src/main/scala/fs2/io/internal/InputOutputBuffer.scala
```diff
@@ -0,0 +1,277 @@
...
+
+  private[this] val buffer: Array[Byte] = new Array(capacity)
+
+  private[this] var head: Int = 0
+  private[this] var tail: Int = 0
+
+  private[this] var closed: Boolean = false
+
+  private[this] val readerPermit: Semaphore = new Semaphore(1)
+  private[this] val writerPermit: Semaphore = new Semaphore(1)
```
mpilquist, line 54: Could you add a summary of how these permits are used? Are their values every > 1?

vasilmkd: I believe they can be, they are not completely foolproof, they are just here to prevent threads from looping endlessly when there is nothing to be read from the buffer. This is fine, because there is one more monitor, the `self.synchronized` block, which actually guards the critical section. Does this make any sense?

mpilquist: Okay cool. My only concern is that some strange access pattern could result in `writerPermit` getting very large, which then causes lots of looping when waiting to write.

vasilmkd: Ok, let's not rush this, I will reiterate on the code.

vasilmkd: I implemented a custom synchronizer that allows at most 1 permit at any time.

vasilmkd:
Wow the `AbstractQueuedSynchronizer` API is something else...

io/src/main/scala/fs2/io/internal/InputOutputBuffer.scala
mpilquist: Looks great!

PR #2385, opened 2021-05-04
Compile fs2 on all supported Scala versions without warnings

### LINKED ISSUE #1887: [Internal] - Deal with fatal warnings
Due to the compiler flags `-Xfatal-warnings` and `-Ywarn-unused:*`, we end up having this kind of code:

```scala
  val _ = ev
  new Stream(Algebra.scope(free.asInstanceOf[FreeC[F, O, Unit]]))
}
```

The `val _ = ev` is only necessary because of such flags. In Scala 2.13.2, we can silence the warning using the `@nowarn` annotation, for example. We could also think about disabling some flags.

This particular issue with implicits is probably better solved by removing the `-Ywarn-unused:implicits` compiler flag (enabled by default in `sbt-tpolecat`) since it gives a lot of false-positives.

vasilmkd:
Resolves #1887. And then some.

I'm hoping none of these changes are controversial. It's just some code reordering to help the Scala 2.13 pattern matcher in `Pull.scala` and using `scala.annotation.nowarn` to get rid of the `dead code` warnings around the `INothing` type in tests, as well as silencing of `unused implicit parameters` warnings.

I also trusted the Scala 3 compiler to figure out whether some type tests cannot be checked at runtime (although this is consistent across Scala 2.13 and 3).

Blocked on #2373.

### DIFF
--- a/core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala
+++ b/core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala
@@ -1,5 +1,6 @@
 package fs2
 
+import scala.annotation.nowarn
 import scala.concurrent.duration._
 
 import java.lang.management.ManagementFactory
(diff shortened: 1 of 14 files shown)


mpilquist:
This is awesome!

PR #2386, opened 2021-05-05
Fix bug in fixedRate which fails to dampen ticks when multiple period…

mpilquist:
…s have passed since last execution

Spotted originally by @gvolpe


gvolpe (approved):
Awesome, thanks!

ChristopherDavenport (commented):
This is a nice improvement!

SystemFw (approved):

Issue #2394, opened 2021-05-13
Memory leaks


svalaskevicius:
Hi, we've been hit by a memory leak in fs2:

```scala
lazy val catsVersion = "2.6.0"
lazy val catsEffectVersion = "3.1.0"
lazy val fs2Version = "3.0.2"
```

Simplest example to reproduce:

```scala
package example

import cats.effect.IOApp
import cats.effect.{ExitCode, IO}
import fs2.Stream
import fs2.Pull
import scala.concurrent.duration._

object Hello extends IOApp {

  override def run(args: List[String]): IO[ExitCode] =
    Stream
      .awakeEvery[IO](2.milli)
      .map(identity _)
      .foreach(_ => IO.unit)
      .compile
      .drain
      .as(ExitCode.Success)

}
```

The fs2 pull/bind etc objects keep being accumulated in memory and never freed, also, they seem to form a cycle:

![2021-05-13-122215_2472x1560](https://user-images.githubusercontent.com/893660/118119524-64355980-b3e6-11eb-8456-4e392665bf22.png)

Also, a colleague mentioned, that Stream.repeat method also leaks, and while I haven't looked at that, it may be related.

Needless to say, this is a rather serious issue and is a blocker for us.

diesalbla:
@svalaskevicius Good afternoon, Sarunas.

We would like to help you fix this issue as soon as possible, although of course we may need some help. For some convenience, could you please open a Pull Request adding a test, or an executable program with that leak? That would help us work on it.

Also, on which version of `fs2` have you found this problem? If you could try to reproduce it in previous versions of `fs2`, that would help us to bisect the problem.

svalaskevicius:
Hi! Both the fs2 version and the program are in the description, as snippets of text. That is the full program that can replicate it. I can ofc upload the whole project as well - will do as a pr, as you've suggested

nikiforo:
@diesalbla 
fs2 has `MemoryLeakSpec`, however, it's located at `core/jvm/src/it/scala/fs2` and isn't executed when `coreJVM/ testOnly fs2.MemoryLeakSpec`. After I moved the spec to `core/jvm/src/test/scala/fs2` I was able to execute the spec with `coreJVM/ testOnly fs2.MemoryLeakSpec`. So, should MemoryLeakSpec be moved to `src/test`?

diesalbla:
> So, should MemoryLeakSpec be moved to src/test?

[...]

PR #2395, opened 2021-05-13
add memory leak example

svalaskevicius:
example of a leak in fs2 3.0.2 - for issue #2394 

visual vm with `Perform GC` clicked many times:

2021-05-13-131108_2748x1060


diesalbla:
@svalaskevicius Could you try running the same program in some previous versions of `fs2`, such as [2.5.0](https://github.com/typelevel/fs2/releases/tag/v2.5.0) or [2.4.0](https://github.com/typelevel/fs2/releases/tag/v2.4.0), or any previous version? We would need to find out in which tagged version the leak first occurred.

svalaskevicius:
Hi, I'm pretty confident it only started happening after the upgrade to the 3.x series

diesalbla:
> Hi, I'm pretty confident it only started happening after the upgrade to the 3.x series

Ok, thanks. That should make it easier to locate the fault. It may be one recent change to the intepreter. I will look into this soon. In the meantime, could you help us please to detect the commit to `main` that introduced this bug? We can chat about this through the Discorc channel, or the Gitter channel, as you prefer.

PR #2399, opened 2021-05-14
Fix 2394 - Memory Leak.

### LINKED ISSUE #2394: Memory leaks

diesalbla:
Resolves #2394. 
We fix a memory leak that was introduced in a previous Pull Request https://github.com/typelevel/fs2/pull/2196
Note that some of the changes in that PR had already been reverted to fix another memory leak.

@svalaskevicius With this changes, I was able to run the `Hello` program from #2395 for more than 20 seconds. Could you take a look to see if it fixes the problem you found?

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -930,9 +930,13 @@ object Pull extends PullLowPriority {
       }
 
       def goFlatMapOut[Y](fmout: FlatMapOutput[G, Y, X], view: View[G, X, Unit]): F[End] = {
+        val runr = new BuildR[G, Y, End]
         val stepRunR = new FlatMapR(view, fmout.fun)
+
         // The F.unit is needed because otherwise an stack overflow occurs.
-        F.unit >> go(scope, extendedTopLevelScope, translation, stepRunR, fmout.stream)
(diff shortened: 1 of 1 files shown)


svalaskevicius:
will check, thanks!

svalaskevicius:
Just to confirm - the project seems running happily for a while now, thanks!

PR #2406, opened 2021-05-16
Add partial apply factory for SignallingRef

### LINKED ISSUE #2396: Add partial apply factory for SignallingRef
`apply` and `of` require both `F[_]` and `A` parameters, which often has inference issues requiring the user to annotate. Instead these methods could use the partial apply "trick" to infer at least the `A` type.

```scala
def apply[F[_], A](initial: A)(implicit arg0: Concurrent[F]): F[SignallingRef[F, A]]
```

Other data types might have similar issues, but this is the one I've noticed so far.

IVaSmi:
Closes #2396.

The primary intention of this PR is to add partial apply factory for SignallingRef.

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/Signal.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Signal.scala
@@ -95,6 +95,30 @@ abstract class SignallingRef[F[_], A] extends Ref[F, A] with Signal[F, A]
 
 object SignallingRef {
 
+  private[fs2] final class PartiallyApplied[F[_]](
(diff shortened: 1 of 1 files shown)


vasilmkd:
Neat little trick with the dummy boolean value. I think you need to run `sbt scalafmtAll` to pass the CI. 😄

IVaSmi:
> Neat little trick with the dummy boolean value. I think you need to run `sbt scalafmtAll` to pass the CI. 😄

Hurried... I fixed that.  😅

kubukoz (approved):

PR #2412, opened 2021-05-20
Fix memory leak in Channel

mpilquist:
Fixes both "hung merge" and "broadcastThrough identity". The `unlessA` breaks monadic tail recursion.

For posterity, here's how I found this one:
- bisect pointed to ef57231 as the first commit for which "hung merge" test failed
- that commit changed `Stream.merge` but none of the changes there changed the "shape" of streams besides swapping in `Channel` for `Queue`
- by poking around in the .hprof written out by `MemoryLeakSpec`, I noticed the `BindBind` chains contained lots of references to `Pull.eval(...).flatMap`, so I searched `Channel.scala` for `Pull.eval` -- there were only 2 matches, one of which had the `unlessA`

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/Channel.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Channel.scala
@@ -209,7 +209,8 @@ object Channel {
                     (
                       State(values, size, waiting.some, producers, closed),
                       F.pure(
-                        (Pull.eval(waiting.get) >> consumeLoop).unlessA(closed)
+                        if (closed) Pull.done
+                        else (Pull.eval(waiting.get) >> consumeLoop)
                       )
                     )
                   }


pchiusano (approved):
Nice. This sounds like it might have been tricky to track down. :)

catostrophe:
@mpilquist I personally know quite a few people who have stepped on the same rake because of the non-laziness of the argument in the syntax extensions `.whenA`/`.unlessA`, despite the fact that the argument in the functions in `Applicative` is call-by-name. Maybe we should finally add the correct syntax extension to `cats`?

SystemFw:
Oh, nice Mike! (and apologies)

prova:
Yes, amazing, the tests pass now. We'll check the bigger codebase soon.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -209,7 +209,8 @@ object Channel {
                     (
                       State(values, size, waiting.some, producers, closed),
                       F.pure(
-                        (Pull.eval(waiting.get) >> consumeLoop).unlessA(closed)
```
johnynek, line 212: if you have a `Defer[Pull]` which I guess you should be able to implement, then you could also write this as:

`Defer[Pull].defer(Pull.eval(waiting.get) >> consumeLoop).unlessA(closed)`

although explicitly putting the branch here to get the result you want is probably clearer (although I would probably comment it so no one refactors it later and regresses the bug).

SystemFw: I hope it gets fixed upstream in cats as well

PR #2415, opened 2021-05-21
Update scalafix to the latest version, fix some rules and run the tests in the CI

vasilmkd:
Please review carefully. I believe the `sbt-scalafix` that was being updated by `scala-steward` was not actually doing anything. This PR also enables running the scalafix tests in the CI, similar to Cats Effect's CI.


vasilmkd:
It seems like this is hitting some incompatibilities with JDK 16. I need to investigate further.

kubukoz (approved):

build.sbt
```diff
@@ -19,7 +19,9 @@ ThisBuild / organizationName := "Functional Streams for Scala"
 ThisBuild / homepage := Some(url("https://github.com/typelevel/fs2"))
 ThisBuild / startYear := Some(2013)
 
-ThisBuild / crossScalaVersions := Seq("3.0.0", "2.12.13", "2.13.5")
+val NewScala = "2.13.5"
+
+ThisBuild / crossScalaVersions := Seq("3.0.0", "2.12.13", NewScala)
```
bjaglin, line 24: I am not sure if your scalafix rules are meant to run against 3.0.0 sources, but if that's the case (and eventually anyway), you should consider tweaking the build to test that scenario.

See https://github.com/liancheng/scalafix-organize-imports/pull/179 for an example (which manually applies the [updated scalafix.g8 template](https://github.com/scalacenter/scalafix.g8/pull/19/commits/71a1160f5d0853a6c796d7e47208b3384bec0687)).

vasilmkd: They aren't we're testing them against `2.13`, but you're right that they are available for `3.0` as well. Not sure if that's a strictly negative thing.

bjaglin: @vasilmkd :sure, willl open a separate PR

vasilmkd: Awesome, thank you.

PR #2421, opened 2021-05-27
fix socket option  for tcp server

### LINKED ISSUE #2414: Cannot set tcp keep_alive to server
Currently scoket options are set to the `AsynchronousSeverSocketChannel` but it should  be applied to the accepted `AsynchronousSocketChannel`,  `AsynchronousSeverSocketChannel` cannot set keep_alive, actually

jilen:
Fixes #2414

### DIFF
--- a/io/src/main/scala/fs2/io/net/SocketGroup.scala
+++ b/io/src/main/scala/fs2/io/net/SocketGroup.scala
@@ -144,11 +144,8 @@ private[net] object SocketGroup {
       val setup: F[AsynchronousServerSocketChannel] =
         address.traverse(_.resolve[F]).flatMap { addr =>
           Async[F].delay {
-            val ch = AsynchronousChannelProvider.provider
-              .openAsynchronousServerSocketChannel(channelGroup)
-            options.foreach { opt =>
-              ch.setOption[opt.Value](opt.key, opt.value)
-            }
+            val ch =
+              AsynchronousChannelProvider.provider.openAsynchronousServerSocketChannel(channelGroup)
             ch.bind(
               new InetSocketAddress(
                 addr.map(_.toInetAddress).orNull,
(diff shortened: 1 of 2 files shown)


io/src/main/scala/fs2/io/net/SocketGroup.scala
```diff
@@ -97,9 +97,11 @@ private[net] object SocketGroup {
...
           val ch =
             AsynchronousChannelProvider.provider.openAsynchronousSocketChannel(channelGroup)
-          options.foreach(opt => ch.setOption(opt.key, opt.value))
-          ch
-        }
+          Either
+            .catchNonFatal(options.foreach(opt => ch.setOption(opt.key, opt.value)))
+            .liftTo[F]
+            .as(ch)
+        }.flatten
```
vasilmkd, line 104: You could probably use `Async[F].defer` for this whole block.

io/src/main/scala/fs2/io/net/SocketGroup.scala
```diff
@@ -94,11 +94,13 @@ private[net] object SocketGroup {
...
     ): Resource[F, Socket[F]] = {
       def setup: F[AsynchronousSocketChannel] =
-        Async[F].delay {
+        Async[F].defer {
           val ch =
             AsynchronousChannelProvider.provider.openAsynchronousSocketChannel(channelGroup)
-          options.foreach(opt => ch.setOption(opt.key, opt.value))
-          ch
+          Either
+            .catchNonFatal(options.foreach(opt => ch.setOption(opt.key, opt.value)))
```
mpilquist, line 101: Hm, why is this `catchNonFatal` necessary? The old code would still result in any exception getting thrown from `ch.setOption` failing the overall effect.

jilen (commented):
Confirmed delay is safe for exception handling

mpilquist (approved):

PR #2445, opened 2021-06-19
Remove use of BuildR

diesalbla:
Retries the change in #2196 , which were reverted in #2394.

This "simplification", was tried before and rolled back because  it caused some Memory Leaks. Inspecting a heapdump, we saw that the leaks were caused by the compilation creating many objects of type `Bind`, `BindBind`, or `DelegateBind`, in which the  `step` (the pull on the left of >>=) was just a terminal.  To avoid this problem, we add a "smart constructor" method `bindView`, used when creating a Pull by binding a Pull with a View. This constructor, instead of always creating a new object, applies the "continuation" function directly. This is admisible during compilation.

_Edit:_ one consequence of directly calling the `cont` method, to avoid accumulating the `DelegateBind` objects, is that the `tailrecM` method of the Monad instance for Stream was no longer stacksafe. We could cause a stack overflow due to the long chain of recursive calls to the method `cont` going up the chain of `BindBind` objects. To avoid this, we extract a loop (tail-recursive method) to perform this simplification.   

~One problem, however, is that this seems to cause one test on the MonadError laws for Stream to fail, the stack-safety of `tailRecM`. However, when we run the code of the stack-safety test, using a different runtime, it seems to pass.~

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -657,14 +657,28 @@ object Pull extends PullLowPriority {
   // This class is not created by combinators in public Pull API, only during compilation
   private class BindBind[F[_], O, X, Y](
       step: Pull[F, O, X],
-      bb: Bind[F, O, X, Y],
-      del: Bind[F, O, Y, Unit]
+      val bb: Bind[F, O, X, Y],
+      val del: Bind[F, O, Y, Unit]
   ) extends Bind[F, O, X, Unit](step) {
-    def cont(zr: Terminal[X]): Pull[F, O, Unit] = bb.cont(zr) match {
-      case t: Terminal[_] =>
-        try del.cont(t)
-        catch { case NonFatal(e) => Fail(e) }
-      case x => new DelegateBind(x, del)
+    def cont(tx: Terminal[X]): Pull[F, O, Unit] =
+      try bindBindAux(this, tx)
+      catch { case NonFatal(e) => Fail(e) }
+  }
+
+  @tailrec
+  private def bindBindAux[F[_], O, X, Y](
+      bibi: BindBind[F, O, X, Y],
+      tx: Terminal[X]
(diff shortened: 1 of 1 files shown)


core/shared/src/test/scala/fs2/StreamLawsSuite.scala
```diff
@@ -36,6 +36,20 @@ class StreamLawsSuite extends Fs2Suite with TestInstances {
...
+    val n = 50000
+    def advance(i: Int): Stream[IO, Either[Int, Int]] =
+      Stream.emit(if (i < n) Left(i + 1) else Right(i))
+
+    val res = Stream
+      .monadErrorInstance[IO]
+      .tailRecM[Int, Int](0)(advance)
+      .compile
+      .toList
+      .unsafeRunSync()(ioRuntime)
```
diesalbla, line 49: This is where something strange is going on... When we use this `ioRuntime`, defined [here](https://github.com/typelevel/munit-cats-effect/blob/main/ce3/shared/src/main/scala/munit/CatsEffectSuite.scala#L30), this test work. However, if we use the `materializeRuntime` defined [here](https://github.com/typelevel/cats-effect/blob/series/3.x/testkit/shared/src/main/scala/cats/effect/testkit/TestInstances.scala#L260-L261), it hangs. 

However, it is not clear why stack-safety of `tailRecM` would be affected by this runtime.., 🤔

mpilquist: I don't see any calls to `tickAll()` which seems related to tests hanging -- not sure why that hasn't been a problem until these changes though. It's also not related to stack safety...

diesalbla: Ok, the tests were right. There was a stack overflow in there (lack of stack safety) in `tailRecM`, with the changes we had done. I have reverted changes to the tests.

mpilquist (approved):

PR #2446, opened 2021-06-22
backport PipedStreamBuffer to 2.5.x

bpholt:
This backports #2383 to 2.5.x, to address #2037 for those who can't upgrade to 3.x yet.

PR #2450, opened 2021-06-24
Mask interruption as IOException

vasilmkd:
[Here](https://github.com/openjdk/jdk/blob/7c31903dd3f2f27de1c352294558a4c1bd6c51e7/src/java.base/share/classes/java/io/PipedInputStream.java#L274) is how the official `PipedInput/OutputStream` OpenJDK implementation cheats interruption by masking it as an `IOException`, which in Scala becomes a `NonFatal` exception, which in turn plays nicely with CE2's `Blocker`. With CE3, due to different masking of interruptions when using `IO.blocking`, this was not an immediate issue, but it is important to be consistent.

### DIFF
--- a/io/src/main/scala/fs2/io/internal/PipedStreamBuffer.scala
+++ b/io/src/main/scala/fs2/io/internal/PipedStreamBuffer.scala
@@ -21,7 +21,7 @@
 
 package fs2.io.internal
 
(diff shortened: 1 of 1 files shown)


vasilmkd:
Fixes issues encountered after #2446 on main.

PR #2454, opened 2021-06-29
Syntax Sugar: Scope: use for comprehension.

diesalbla:
Given that last operation was a `map`, this should not cause problems...

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Scope.scala
+++ b/core/shared/src/main/scala/fs2/internal/Scope.scala
@@ -23,7 +23,7 @@ package fs2.internal
 
 import scala.annotation.tailrec
 
-import cats.{Id, Traverse, TraverseFilter}
+import cats.{Id, Traverse}
 import cats.data.Chain
 import cats.effect.kernel.{Fiber, Outcome, Poll, Ref, Resource, Unique}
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/internal/Scope.scala
```diff
@@ -478,6 +474,8 @@ private[fs2] final class Scope[F[_]] private (
 
 private[fs2] object Scope {
 
+  private[Scope] val closedScopeException = new RuntimeException("Scope closed at time of lease")
```
mpilquist, line 477: Is this safe to reuse? Seems like each time it's thrown it will get a new stack trace which then could cause frames to be retained.

diesalbla: I am not sure about that. Will keep inline, for good measure.

mpilquist: One issue is that `fillInStackTrace` will get called each time this is thrown. But since there was a single instance stored on the object, the last recorded stack trace hangs around forever. Hence if using singleton exceptions, `fillInStackTrace` should be overridden to be a no-op (e.g. by mixing in `NoStackTrace`).

PR #2462, opened 2021-07-03
Avoid allocations in TLSEngine when logging is disabled

mpilquist:
This gives us syntax like:
```scala
tlsContext.client(socket).use(...)
tlsContext.clientBuilder(socket).withLogging(...).build.use(...)
```


diesalbla (commented):
👍🏼

PR #2465, opened 2021-07-04
Relax queue-related signatures to use `QueueSink` / `QueueSource`

armanbilge:
I thought this would be a helpful change, but I am surprised/confused that it doesn't pass MiMa. I was under the impression that it would be okay because `Queue` extends `QueueSource` and `QueueSink` but I guess I don't understand bincompat 🙃  

```
[error] fs2-core: Failed binary compatibility check against co.fs2:fs2-core_2.13:3.0.3! Found 4 potential problems (filtered 21)
[error]  * static method fromQueueNoneTerminatedChunk(cats.effect.std.Queue,Int)fs2.Stream in class fs2.Stream's type is different in current version, where it is (cats.effect.std.QueueSource,Int)fs2.Stream instead of (cats.effect.std.Queue,Int)fs2.Stream
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("fs2.Stream.fromQueueNoneTerminatedChunk")
[error]  * static method fromQueueNoneTerminated(cats.effect.std.Queue,Int,cats.Functor)fs2.Stream in class fs2.Stream's type is different in current version, where it is (cats.effect.std.QueueSource,Int,cats.Functor)fs2.Stream instead of (cats.effect.std.Queue,Int,cats.Functor)fs2.Stream
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("fs2.Stream.fromQueueNoneTerminated")
[error]  * static method fromQueueUnterminatedChunk(cats.effect.std.Queue,Int,cats.Functor)fs2.Stream in class fs2.Stream's type is different in current version, where it is (cats.effect.std.QueueSource,Int,cats.Functor)fs2.Stream instead of (cats.effect.std.Queue,Int,cats.Functor)fs2.Stream
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("fs2.Stream.fromQueueUnterminatedChunk")
[error]  * static method fromQueueUnterminated(cats.effect.std.Queue,Int,cats.Functor)fs2.Stream in class fs2.Stream's type is different in current version, where it is (cats.effect.std.QueueSource,Int,cats.Functor)fs2.Stream instead of (cats.effect.std.Queue,Int,cats.Functor)fs2.Stream
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("fs2.Stream.fromQueueUnterminated")
```

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -30,7 +30,7 @@ import cats.data.Ior
 import cats.effect.{Concurrent, SyncIO}
 import cats.effect.kernel._
 import cats.effect.kernel.implicits._
-import cats.effect.std.{Console, Queue, Semaphore}
+import cats.effect.std.{Console, Queue, QueueSink, QueueSource, Semaphore}
 import cats.effect.Resource.ExitCase
 import cats.syntax.all._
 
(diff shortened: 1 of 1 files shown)


mpilquist:
Yep, this change is indeed binary incompatible and would result in `NoSuchMethodError`s at runtime.

armanbilge:
Oh ... this is irrecoverable? Even if so, at least part of this PR was binary-compatible, is that not of interest?

mpilquist:
Feel free to reopen if you can get bincompat working but I don't see a way to do so.

Issue #2437, opened 2021-06-08
Stream.fromQueue* could take a QueueSource instead of a Queue


arosien:
I believe the signatures for `Stream.fromQueueUnterminated`, etc., could be narrowed to take `QueueSource` instead of `Queue`, because they only rely on the `take` and `tryTake` methods. Then if users want to translate a `Queue` to a `Stream` they don't have to require knowledge of the "write" methods of the `Queue`, only the "read" methods.

kubukoz:
Agreed. I don't know if this will be a binary compatible change, though - if not, we can probably work around by making the existing functions package-private and adding new ones.

djspiewak:
Should be binary compatible since it's a contravariant parameter adjustment.

armanbilge:
> Should be binary compatible since it's a contravariant parameter adjustment

I thought this too! But, MiMa didn't agree, I'd be really really interested to hear an explanation for why.

mpilquist:
@djspiewak These are static methods, not class methods.

armanbilge:
`fromQueue` is static, but the `enqueue` methods I also modified in my PR are class methods.

PR #2466, opened 2021-07-05
Relax queue-related signatures to use `QueueSink` / `QueueSource`

armanbilge:
Supersedes https://github.com/typelevel/fs2/pull/2465

(GH wouldn't let me re-open the old PR)

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3234,7 +3234,12 @@ object Stream extends StreamLowPriority {
       limit
     )
 
-  private[fs2] def fromQueueUnterminated[F[_]: Functor, A](
+  /** Returns a stream of elements from the supplied queue.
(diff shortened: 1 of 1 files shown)


mpilquist (approved):

vasilmkd (approved):

arosien:
Fixes #2437 I believe.

diesalbla (approved):

kubukoz:
I'm just curious... I thought type erasure would consider these identical:

https://github.com/typelevel/fs2/blob/f855bf8d3fc515e1a6c8d8e112106573a8bc0a79/core/shared/src/main/scala/fs2/Stream.scala#L3227-L3245

(also, shouldn't the second one be `private[fs2]`?)

armanbilge:
@kubukoz I would have liked to, but marking it as `private[fs2]` still disappears the static forwarders :(

PR #2467, opened 2021-07-07
Pull.outputOption

nikiforo:
```scala
import cats.effect.IOApp
import cats.effect.{ExitCode, IO}
import cats.syntax.all._

object PullOptionApp extends IOApp {

  def run(args: List[String]): IO[ExitCode] = {
    val single = Pull.outputOption(Some(1))
    val empty = Pull.outputOption(None)

    val msg = s"single: ${single.stream.compile.toList}\nempty: ${empty.stream.compile.toList}"

    IO(println(msg)).as(ExitCode.Success)
  }
}
```

```
[info] running fs2.PullOptionApp 
single: List(1)
empty: List()
```

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -350,6 +350,13 @@ object Pull extends PullLowPriority {
     */
   def output1[F[x] >: Pure[x], O](o: O): Pull[F, O, Unit] = Output(Chunk.singleton(o))
 
+  /** Lifts the given optional value `O` into a pull that performs no
+    * effects, emits the content of that option, and always
+    * terminates successfully with a unit result.
+    */
+  def outputOption[F[x] >: Pure[x], O](opt: Option[O]): Pull[Pure, O, Unit] =
+    output(Chunk.fromOption(opt))
+
   /** Creates a pull that emits the elements of the given chunk.
     * The new pull performs no effects and terminates successfully with a unit result.
     */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -350,6 +350,13 @@ object Pull extends PullLowPriority {
     */
   def output1[F[x] >: Pure[x], O](o: O): Pull[F, O, Unit] = Output(Chunk.singleton(o))
 
+  /** Lifts the given optional value `O` into a pull that performs no
+    * effects, emits the content of that option, and always
+    * terminates successfully with a unit result.
+    */
+  def outputOption[F[x] >: Pure[x], O](opt: Option[O]): Pull[Pure, O, Unit] =
+    output(Chunk.fromOption(opt))
```
diesalbla, line 358:
```suggestion
    opt.fold(done)(output1)
```

diesalbla (approved):

mpilquist:
Naming wise, would `output1Option` or `outputOption1` be better?

core/shared/src/main/scala/fs2/Pull.scala
nikiforo: I've done it with `map` and `getOrElse` because fold failed compilation.

However, I haven't removed Chunk.fromOption, should I?

nikiforo:
> Naming wise, would `output1Option` or `outputOption1` be better?

Indeed, I like your options more. Additionally, `output1Option` looks like it contains `10`, so I ended up with `outputOption1`

PR #2468, opened 2021-07-07
Rework parJoin to work with transformers

### LINKED ISSUE #1945: Stream.parJoin freezes using an EitherT as effect context
In my project we use an `EitherT` as effect context instead of directly using an `IO` (it allows better error handling without the need of using a `MonadError` or just throwing up exceptions). We have seen that, in case we would want to parallelize a stream of streams using this type as context, in case an effect evaluates as a left that function gets stuck at `Stream#parJoin` function.

I have written some tests [here](https://github.com/aldtid/fs2-parjoin-error-with-eithert) that can reproduce the error while running streams in parallel and how running the same function sequentially finishes the execution as expected. The interesting test is this one:

https://github.com/aldtid/fs2-parjoin-error-with-eithert/blob/2adc2ea42973405c783a2acd3ba3faa2307ca0ba/src/test/scala/com/example/JoinTest.scala#L49-L62

which is the one that freezes, the rest returns the expected results.

vasilmkd:
Resolves #1945.

The reasoning behind this change is the following.

It is ok to join the _inner_ forked fibers because they are `uncancelable` and cannot result in a deadlock. It is also ok to join the _outer_ forked fiber because we completely control it.

The `.fold(F2.unit)(_ >> _.join)` does indeed create a potentially big `flatMap` chain, but IMO, this is warranted due to the inherent complexity of this combinator, which already had significant overhead in terms of allocations.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2070,32 +2070,14 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
     val outer = this.asInstanceOf[Stream[F2, Stream[F2, O2]]]
 
     val fstream: F2[Stream[F2, O2]] = for {
-      done <- SignallingRef(None: Option[Option[Throwable]])
+      done <- Deferred[F2, Either[Throwable, Unit]]
       available <- Semaphore(maxOpen.toLong)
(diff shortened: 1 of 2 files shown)


vasilmkd:
This does make the memory leak test for this operator flaky. It does pass when running it locally, and it sometimes passes in the CI too. 😖

Issue #2471, opened 2021-07-08
Add text.hex.{encode, decode}


mpilquist:
Model these after the methods in `text.base64` -- something like:

```scala
object text:
  object hex:
    def decode[F[_]: RaiseThrowable]: Pipe[F, String, Byte] = ...
    def decodeWithAlphabet(alphabet: Bases.HexAlphabet)[F[_]: RaiseThrowable]: Pipe[F, String, Byte] = ...

    def encode[F[_]]: Pipe[F, Byte, String] = ...
    def encodeWithAlphabet[F[_]](alphabet: Bases.HexAlphabet): Pipe[F, Byte, String] = ...
```

Example usage:

```scala
def sha256sum(path: Path): IO[Unit] =
  Files[IO].readAll(path, 4096)
    .through(hash.sha256)
    .through(text.hex.encode)
    .through(text.utf8Encode) // maybe we should alias this one as text.utf8.encode
    .through(Files[IO].writeAll(path.resolve(".sha256"))
    .compile
    .drain
```

Without this combinator, the above would be:

```scala
def sha256sum(path: Path): IO[Unit] =
  Files[IO].readAll(path, 4096)
    .through(hash.sha256)
    .chunks
    .map(_.toByteVector.toHex)
    .through(text.utf8Encode) // maybe we should alias this one as text.utf8.encode
    .through(Files[IO].writeAll(path.resolve(".sha256"))
    .compile
    .drain
```

mpilquist:
Initial sketch of this is on topic/hex branch

PR #2477, opened 2021-07-10
Add text.hex.{decode, encode}

mpilquist:


mpilquist:
Fixes #2471

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -21,153 +21,174 @@
...
 /** Provides utilities for working with streams of text (e.g., encoding byte streams to strings). */
 object text {
-  private val utf8Charset = Charset.forName("UTF-8")
-  private[this] val utf8BomSeq: Seq[Byte] = Array(0xef.toByte, 0xbb.toByte, 0xbf.toByte).toSeq
 
-  /** Converts UTF-8 encoded byte stream to a stream of `String`. */
-  def utf8Decode[F[_]]: Pipe[F, Byte, String] =
-    _.chunks.through(utf8DecodeC)
+  object utf8 {
+    private val utf8Charset = Charset.forName("UTF-8")
```
rossabaker, line 34: Doesn't really matter, but the JDK already provides this as a constant: `StandardCharsets.UTF_8`.

mpilquist: Showing my age - I avoided that when we wrote this stuff initially (in scodec) to avoid the dependency on Java 7. :)

PR #2478, opened 2021-07-11
Create `fs2.io.file` APIs for Node.js

armanbilge:
Picking up from #2453.

This PR introduces a new `PosixFiles` typeclass, currently implemented only for Node.js. It is closely modeled on Node.js's [`fs` module](https://nodejs.org/api/fs.html) but otherwise shouldn't be specific to Node.js at all. I hope future work will implement it for JVM.

For the sake of offering some platform-agnostic API, I also introduced a `ReadFiles` typeclass which both `Files` and `PosixFiles` extend. File-reading seems both safe and straightforward enough to do in a platform-agnostic manner. I'd love to see a similar `WriteFiles` typeclass as well, but it would have to be carefully implemented to get right the semantics of open/create/append/truncate etc.

Both new typeclasses are `sealed` to give us some room to iterate on these APIs.

For testing, I copy-pastad the JVM file suite and re-worked for JS. So, essentially anything we can do on JVM we can do on JS, just with different semantics.

I also slipped in the Unix sockets implementation now that File APIs are available.

The only thing missing is a `Watcher` implementation. We can definitely do this on Node.js as well, but at a first glance seems fairly complicated to implement and I don't want the good stuff here blocked by it. So, I'll finish it off in one more PR.

Thanks in advance for feedback and reviews.


io/js/src/main/scala/fs2/io/Stream.scala
```diff
@@ -0,0 +1,36 @@
+/*
```
mpilquist, line 1: Can you rename this file to something else? It will conflict with folks using editor quick open features for Stream.scala

mpilquist (approved):

io/js/src/main/scala/fs2/io/Stream.scala
armanbilge: Good point, done. Thanks!

PR #2483, opened 2021-07-12
Make ScalablyTyped quiet

armanbilge:


build.sbt
```diff
@@ -10,6 +10,7 @@ addCommandAlias("testJVM", ";rootJVM/test")
 addCommandAlias("testJS", "rootJS/test")
 
 Global / onChangedBuildSource := ReloadOnSourceChanges
+Global / stQuiet := true
```
mpilquist, line 13: Should this be `Global` or `ThisBuild` or a local project setting in io?

armanbilge: Not sure, I can play with it. I always get confused by SBT scoping. I grabbed it straight from here.
https://scalablytyped.org/docs/conversion-options#customize-the-build

PR #2491, opened 2021-07-15
Catch errors during Node.js socket initialization

armanbilge:
Adds a listener for errors thrown during client socket connection and server binding. Also adds test coverage.

### DIFF
--- a/io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala
+++ b/io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala
@@ -160,10 +160,10 @@ class SocketSuite extends Fs2Suite with SocketSuitePlatform {
     }
 
     test("errors - should be captured in the effect") {
-      (for {
+      (for { // Connection refused
         bindAddress <- Network[IO].serverResource(Some(ip"127.0.0.1")).use(s => IO.pure(s._1))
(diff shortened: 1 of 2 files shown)


io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala
```diff
@@ -158,6 +159,16 @@ class SocketSuite extends Fs2Suite with SocketSuitePlatform {
...
     }
 
+    test("errors - should be captured in the effect") {
+      (for {
+        bindAddress <- Network[IO].serverResource(Some(ip"127.0.0.1")).use(s => IO.pure(s._1))
+        _ <- Network[IO].client(bindAddress).use(_ => IO.unit)
+      } yield ()).attempt.map(r => assert(r.isLeft)) >> (for {
+        bindAddress <- Network[IO].serverResource(Some(ip"127.0.0.1")).map(_._1)
+        _ <- Network[IO].serverResource(Some(bindAddress.host), Some(bindAddress.port))
+      } yield ()).attempt.use(r => IO(assert(r.isLeft)))
```
ChristopherDavenport, line 169: May want to document what errors this is checking for.

armanbilge: Good idea, done.

PR #2495, opened 2021-07-20
Small fixes for Node.js Readable/Writable converters

armanbilge:


io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -126,7 +130,8 @@ private[fs2] trait ioplatform {
                 )
               }
             } >> go(tail)
-          case None => Pull.eval(F.delay(writable.end())) >> Pull.done
+          case None =>
+            Pull.eval(F.async_[Unit](cb => writable.end(() => cb(Right(()))))) >> Pull.done
```
mpilquist, line 134: The `Pull.done` looks redundant here

io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -45,7 +45,11 @@ private[fs2] trait ioplatform {
     Stream
       .resource(for {
         readable <- Resource.makeCase(readable.map(_.asInstanceOf[streamMod.Readable])) {
-          case (readable, Resource.ExitCase.Succeeded) => F.delay(readable.destroy())
+          case (readable, Resource.ExitCase.Succeeded) =>
+            if (!readable.readableEnded)
```
mpilquist, line 49: I don't know node apis at all so excuse this question, but why is the destroy call not needed when the natural EOF is reached?

io/js/src/main/scala/fs2/io/ioplatform.scala
armanbilge: I was wondering that too! Still new to fs2, thanks :)

io/js/src/main/scala/fs2/io/ioplatform.scala
armanbilge: Ugh I don't understand the Node APIs either and the docs aren't great. In Node many things (like Sockets) are "duplex" streams i.e. implement both Readable and Writable. So calling destroy when we are done reading would actually tear down the whole thing, preventing us from writing.

Writing it out, tbh I'm still not sure this is the best approach, I'll take any ideas?

armanbilge: Basically, the behavior we are encoding here is that if you don't consume the entire `Readable` we will destroy it in a short-circuiting/error-like manner. So maybe we should better encode this semantics into `fs` ... if you try to close a `Readable` resource before it's `ended` we throw an error?

mpilquist: Closing before consuming whole file input should be fine and shouldn’t raise an error (at least if we want to remain consistent with JVM side)

armanbilge: I've been thinking about this too today and I agree. 👍

PR #2496, opened 2021-07-21
Improve Node.js `Readable`/`Writable` converters, introduce `Duplex` converter

armanbilge:
Idiomatic APIs as inspired by their [JVM cousins](https://github.com/typelevel/fs2/blob/main/io/jvm/src/main/scala/fs2/io/ioplatform.scala).

Edit: this PR is becoming a catch-all for other fixes.

Edit 2: okay, I ended up doing some major reworking in anticipation of new socket implementations.

### DIFF
--- a/io/js/src/main/scala/fs2/io/ioplatform.scala
+++ b/io/js/src/main/scala/fs2/io/ioplatform.scala
@@ -41,12 +41,14 @@ import scala.scalajs.js.|
 
 private[fs2] trait ioplatform {
 
-  def readReadable[F[_]](readable: F[Readable])(implicit F: Async[F]): Stream[F, Byte] =
+  def readReadable[F[_]](readable: F[Readable], destroyIfNotEnded: Boolean = true)(implicit
+      F: Async[F]
+  ): Stream[F, Byte] =
     Stream
       .resource(for {
         readable <- Resource.makeCase(readable.map(_.asInstanceOf[streamMod.Readable])) {
           case (readable, Resource.ExitCase.Succeeded) =>
-            if (!readable.readableEnded)
+            if (!readable.readableEnded & destroyIfNotEnded)
               F.delay(readable.destroy())
             else
               F.unit
@@ -117,7 +119,8 @@ private[fs2] trait ioplatform {
     } yield readable.asInstanceOf[Readable]
 
   def writeWritable[F[_]](
-      writable: F[Writable]
+      writable: F[Writable],
+      endAfterUse: Boolean = true
   )(implicit F: Async[F]): Pipe[F, Byte, INothing] =
     in =>
       Stream.eval(writable.map(_.asInstanceOf[streamMod.Writable])).flatMap { writable =>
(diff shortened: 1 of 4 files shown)


io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -134,7 +137,12 @@ private[fs2] trait ioplatform {
         }.drain
       }
 
-  def mkWritable[F[_]](implicit F: Async[F]): Resource[F, (Writable, Stream[F, Byte])] =
+  def mkWritable[F[_]: Async](f: Writable => F[Unit]): Stream[F, Byte] =
```
armanbilge, line 140: Not sure about the name of this method, thoughts?

mpilquist: `readWritable`? Not my favorite but matches the convention used on JVM side

armanbilge: Not a fan myself, but I do like the consistency. Also renamed `from{Readable,Writable}` to `readReadable` and `writeWritable`.

io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -41,12 +41,14 @@ import scala.scalajs.js.|
...
+  ): Stream[F, Byte] =
     Stream
       .resource(for {
         readable <- Resource.makeCase(readable.map(_.asInstanceOf[streamMod.Readable])) {
           case (readable, Resource.ExitCase.Succeeded) =>
-            if (!readable.readableEnded)
+            if (!readable.readableEnded & destroyIfNotEnded)
               F.delay(readable.destroy())
             else
               F.unit
```
armanbilge, line 54: Still not thrilled with this but I think it is the best we can do. The underlying issue is that Node.js has no way to close/"end" a `Readable` besides the nuclear option `destroy`. This same thing came up in https://github.com/typelevel/fs2/pull/2453#discussion_r666153173 when we couldn't implement the `endOfInput` method for sockets.

io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -131,77 +119,109 @@ private[fs2] trait ioplatform {
...
         go(in).stream.handleErrorWith { ex =>
           Stream.eval(F.delay(writable.destroy(js.Error(ex.getMessage))))
         }.drain
       }
 
-  def mkWritable[F[_]](implicit F: Async[F]): Resource[F, (Writable, Stream[F, Byte])] =
+  def readWritable[F[_]: Async](f: Writable => F[Unit]): Stream[F, Byte] =
+    Stream.empty.through(toDuplexAndRead(f))
+
+  def toDuplexAndRead[F[_]: Async](f: Duplex => F[Unit]): Pipe[F, Byte, Byte] =
```
armanbilge, line 136: An even funkier name! Thoughts? 😅

armanbilge:
@mpilquist apologies for all the churn. In these latest commits I introduced the `Duplex` (i.e., readable and writable) facade along with a converter. To avoid duplicating code, I reimplemented the readable/writable converters in terms of the duplex converter although this does add quite a bit of overhead. It is very unfortunate that Node.js doesn't offer a way to create a `Duplex` from a `Readable` and a `Writable`.

`Duplex` streams are pretty important in Node.js (e.g., their sockets implement `Duplex`) so the idea is to use these changes to cleanup the `Socket` implementation in a forthcoming PR.

Thanks!

armanbilge:
Closing in favor of #2504 because it's easier to debug these converters alongside the socket implementations.

PR #2500, opened 2021-07-22
Minor fixes in Pull docs

valencik:
This PR fixes a handful of things in the `Pull` docs.

In this I noticed that only `as` has the `@tparam` and `@param` annotations, I could add these to the other methods like `map` and `flatMap` if that's desired. I'm not sure what the best practice is here.

The docs for fs2 are really great and helpful, thank you so much!


SystemFw (approved):

mpilquist:
Re: `@param` and `@tparam`, I don't have strong opinions besides one -- that we don't put a policy in place that results in lots of low quality documentation boilerplate. E.g., if we say "every method should have `@param` docs for every parameter" and then start writing docs like "`@param f The function to apply`", we're doing it wrong :)

PR #2504, opened 2021-07-25
Overhaul Node.js socket implementations

armanbilge:
Builds on #2496 to dramatically simplify the Node.js (TLS) socket implementations. (Feel free to close #2496 and review the changes all together here.)


armanbilge:
At this point I'm no longer sure if anything got simpler. I think it is more robust though.

PR #2511, opened 2021-07-30
`file2` wip

armanbilge:


io/shared/src/main/scala/fs2/io/file2/PathApi.scala
```diff
@@ -23,7 +23,7 @@ package fs2
 package io
 package file2
 
-private[file2] trait PathApi {
+private[file2] trait PathApi extends Any {
```
mpilquist, line 26: Why universal trait?

armanbilge: So that JS Path can be an `AnyVal` for `String`. But now I realize can't do this if we also want to store the filesystem in the path.

armanbilge:
I was thinking I can use this branch for my wip and anything you like you can cherry-pick into your branch?

mpilquist:
sure

PR #2517, opened 2021-07-31
Flesh out Path APIs

armanbilge:


io/shared/src/main/scala/fs2/io/file/PathApi.scala
```diff
@@ -23,13 +23,46 @@ package fs2
...
+
+  def resolve(path: Path): Path
+  
   def normalize: Path
+
+  def isAbsolute: Boolean
+
+  def absolute: Path
+  
+  def names: Seq[Path]
```
mpilquist, line 47: s/Seq/List/ so we have type class instances available

armanbilge: I thought they got added for list?

armanbilge: https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/instances/seq.scala

mpilquist: Okay but still should be List - Seq offers no advantage here

PR #2530, opened 2021-08-04
Fix null pointer exception when Node.js stream is destroyed

armanbilge:


io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -203,7 +203,10 @@ private[fs2] trait ioplatform {
               .setDestroy { (duplex, err, cb) =>
                 dispatcher.unsafeRunAndForget {
                   error
-                    .complete(js.JavaScriptException(err))
+                    .complete(
+                      Option(err)
+                        .fold[Exception](new StreamDestroyedException)(js.JavaScriptException(_))
+                    )
```
armanbilge, line 209: `err` may be null, causing a null pointer exception when trying to render the `JavaScriptException`

PR #2547, opened 2021-08-07
Catch a couple more javascript exceptions

armanbilge:


io/js/src/main/scala/fs2/io/net/tls/SSLException.scala
```diff
@@ -33,7 +33,9 @@ private class JavaScriptSSLException(cause: js.JavaScriptException)
 object SSLException {
   private[io] def unapply(cause: js.JavaScriptException): Option[SSLException] = cause match {
     case js.JavaScriptException(error: js.Error)
-        if error.message.contains("TLS") || error.message.contains("SSL") =>
+        if error.message.contains("TLS") || error.message.contains(
+          "SSL"
+        ) || error.message.toLowerCase.contains("certificate") =>
```
armanbilge, line 38: Too catch-all?

PR #2552, opened 2021-08-10
Rework `parJoin` to work with transformers

### LINKED ISSUE #1945: Stream.parJoin freezes using an EitherT as effect context

vasilmkd:
Resolves #1945.

The memory leak tests seem happy, I ran the CI many times yesterday and this morning.

I will follow this up with a PR that fixes this issue on `main`. It should be even simpler given that Cats Effect 3 has support for the `Outcome` in `guaranteeCase`, which we can use to sequence short-circuiting monad transformers.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2069,158 +2069,162 @@ final class Stream[+F[_], +O] private[fs2] (private val free: FreeC[F, O, Unit])
     val _ = (ev, ev2)
     val outer = this.asInstanceOf[Stream[F2, Stream[F2, O2]]]
 
-    val fstream: F2[Stream[F2, O2]] = for {
-      done <- SignallingRef(None: Option[Option[Throwable]])
-      available <- Semaphore(maxOpen.toLong)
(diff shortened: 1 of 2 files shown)


mpilquist:
Was there any reason not to merge this one? We've already merged the equivalent PR on main, right?

vasilmkd:
No reason really, the tests pass and the leak tests are fine too.

PR #2554, opened 2021-08-12
Tweak Node.js server to be more idiomatic

armanbilge:
Just some minor tweaking to make the code a bit nicer.

More importantly, I think this suffers from same/similar problems as #2300, namely: how do you gracefully shutdown/handle errors? The critical thing seems to be that any sockets remaining in the queue should be destroyed.

Perhaps I'm missing something, but it seems to me that the problem lies with the `server` method. With the `serverResource` method, I can conceptualize the "listening" part of the server being tied to the resource, so that even after the resource is closed the stream of sockets is still usable. But maybe this is leaky?

cc @RaasAhsan

### DIFF
--- a/io/js/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
@@ -33,6 +33,7 @@ import com.comcast.ip4s.Host
 import com.comcast.ip4s.IpAddress
 import com.comcast.ip4s.Port
 import com.comcast.ip4s.SocketAddress
+import fs2.internal.jsdeps.node.eventsMod
 import fs2.internal.jsdeps.node.netMod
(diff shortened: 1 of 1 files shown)


mpilquist:
@armanbilge Does this PR change the behavior to match the JVM version?

armanbilge:
Hmm, a bit fuzzy since I made this. It makes two changes:
1. The use of `Async` instead of a `Deferred` to handle errors at server creation. This should just be code cleanup.
2. Use a `None`-terminated `Queue` of `Sockets` (the previous was unterminated). A `None` is added to the queue when the server resource is closed.

What is still an open problem is how to guarantee that all the sockets are destroyed. Right now, if they are unconsumed by the `Stream`, they will sit open in the queue indefinitely.

> Does this PR change the behavior to match the JVM version?

Not sure, what exactly does the JVM side do?

PR #2561, opened 2021-08-15
Scala 2.13: avoid compiler warnings.

diesalbla:
In the Scala 2.13.6 compiler, due to some compiler bug, when matching against a `ViewL` (which is either a terminal or an action), if the case for the action goes after the cases for terminals, the compiler thinks it is unreachable and issues a warning. To sidestep this problem, we reorder cases to put terminals at the bottom.


vasilmkd (approved):
I've done this before myself and I can also confirm it works.

PR #2577, opened 2021-08-27
Special case for `parEvalMap` with max concurrency of 1

bastewart:
This avoids the penalty of using the combinator if there is no
concurrency. Analogous to #2557 which did the same for `parJoin`.


bastewart:
@Daenyth sorry, not sure what to do here! Do I wait for another maintainer to give a second opinion and they'll potentially merge it?

mpilquist:
@bastewart Thanks!

PR #2588, opened 2021-09-03
Initial import of time series from scodec-protocols

mpilquist:


core/shared/src/main/scala/fs2/Scan.scala
```diff
@@ -0,0 +1,242 @@
...
+    )
+
+  /** Returns a scan that inputs/outputs pairs of elements, with `I` and `O` in the first element of the pair. */
+  def first[A]: Scan[S, (I, A), (O, A)] =
+    lens(_._1, (t, o) => (o, t._2))
+
+  /** Returns a scan that inputs/outputs pairs of elements, with `I` and `O` in the second element of the pair. */
+  def second[A]: Scan[S, (A, I), (A, O)] =
+    lens(_._2, (t, o) => (t._1, o))
+
```
diesalbla, line 141: The use of `first`, `second`, `left`, and `right`, make me wonder if we could define a `cats.arrow.Arrow` instance for this datatype, and use the nomenclature from those typeclasses.

core/shared/src/main/scala/fs2/Scan.scala
```diff
@@ -0,0 +1,242 @@
...
+    * the left pass through directly.
+    */
+  def right[A]: Scan[S, Either[A, I], Either[A, O]] =
+    semilens(_.fold(a => Left(Left(a)), i => Right(i)), (_, o) => Right(o))
+
+  /** Combines this scan with the supplied scan such that elements on the left
+    * are fed through this scan while elements on the right are fed through the
+    * suppplied scan. The outputs are joined together.
+    */
+  def or[S2, I2, O2 >: O](that: Scan[S2, I2, O2]): Scan[(S, S2), Either[I, I2], O2] =
```
diesalbla, line 180: This function looks similar to the `choice` method from the `cats.arrow.Choice` typeclass, except that whereas the typeclass assumes the same `F`, here the state type would be part of that type.

core/shared/src/main/scala/fs2/Scan.scala
```diff
@@ -0,0 +1,242 @@
...
+          case Right(i2) =>
+            val (s2p, o2s) = that.transform(s2, i2)
+            ((s, s2p), o2s)
+        }
+      },
+      { case (s, s2) => onComplete(s) ++ that.onComplete(s2) }
+    )
+
+  /** Like [[or]] but the output elements are kept separate. */
+  def either[S2, I2, O2](t: Scan[S2, I2, O2]): Scan[(S, S2), Either[I, I2], Either[O, O2]] =
```
diesalbla, line 196: Simmilarly, except for the change of state types, this seems to correspond to the `choose` function (or `+++` operator) from the `cats.arrow.ArrowChoice` typeclass. Perhaps we can rename it to that?

core/shared/src/main/scala/fs2/timeseries/TimeSeries.scala
```diff
@@ -0,0 +1,142 @@
...
+        nextTick: FiniteDuration,
+        s: Stream[Pure, TimeStamped[A]]
+    ): Pull[Pure, TimeSeriesValue[A], Unit] = {
+      def tickTime(x: Int) = nextTick + (x * tickPeriod)
+      s.pull.uncons.flatMap {
+        case Some((hd, tl)) =>
+          hd.indexWhere(_.time >= nextTick) match {
+            case None =>
+              if (hd.isEmpty) Pull.pure(())
+              else Pull.output(hd.map(_.toTimeSeriesValue)) >> go(nextTick, tl)
```
diesalbla, line 74: The first case should not happen, given that a pull only contains non-empty chunks.
```suggestion
              Pull.output(hd.map(_.toTimeSeriesValue)) >> go(nextTick, tl)
```

core/shared/src/main/scala/fs2/timeseries/TimeSeries.scala
```diff
@@ -0,0 +1,142 @@
...
+            case Some(idx) =>
+              val (prefix, suffix) = hd.splitAt(idx)
+              val out =
+                if (prefix.isEmpty) Pull.pure(()) else Pull.output(prefix.map(_.toTimeSeriesValue))
```
diesalbla, line 78: The `Pull.output` constructor should default to `Pull.done` if the chunk is empty. 
```suggestion
              val out = Pull.output(prefix.map(_.toTimeSeriesValue))
```

core/shared/src/main/scala/fs2/Scan.scala
mpilquist: I added a `Strong` instance but `Arrow`, `Compose`, `Category`, and `Choice` instances are not possible due to the initial state that's kept in a `Scan`.

core/shared/src/main/scala/fs2/Scan.scala
mpilquist: Renamed to `choice`

core/shared/src/main/scala/fs2/Scan.scala
mpilquist: Renamed to `choose`

PR #2596, opened 2021-09-07
Add a convenient shorthand for predicating effects on a boolean signal.

mn98:
Purely for convenience, this allows for writing
  `signal.predicate(f)`
instead of
  `signal.get.flatMap(f.whenA)`

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/Signal.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Signal.scala
@@ -22,10 +22,10 @@
 package fs2
 package concurrent
 
-import cats.{Applicative, FlatMap, Functor, Invariant}
 import cats.data.OptionT
 import cats.effect.kernel.{Concurrent, Deferred, Ref}
 import cats.syntax.all._
+import cats.{Applicative, Functor, Invariant, Monad}
 
 import scala.collection.immutable.LongMap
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -73,10 +74,16 @@ object Signal extends SignalInstances {
...
-    )(implicit F: Concurrent[F]): Stream[F, A] =
+
+    /** Interrupts the supplied `Stream` when this `Signal` is `true`.
+      */
+    def interrupt[A](s: Stream[F, A])(implicit F: Concurrent[F]): Stream[F, A] =
       s.interruptWhen(self)
+
+    /** Predicate the supplied effect `f` on this `Signal` being `true`.
+      */
+    def predicate(f: F[_])(implicit F: Applicative[F] with FlatMap[F]): F[Unit] =
```
mpilquist, line 85:
```suggestion
    def predicate(f: F[_]: Monad): F[Unit] =
```

mn98: Hmm, that form doesn't compile for me, but this form does:
`def predicate(f: F[_])(implicit F: Monad[F]): F[Unit]`
Is that ok?

mpilquist: Whoops, yep

mn98: Great; committed.

mn98:
I would add that although this is a minor convenience feature it's probably still worth merging. I get the impression much of the `*Ops` sugar will eventually move out to a separate module/project, which would then include this little gem. I probably wouldn't have raised this PR if `BooleanSignalOps` didn't already exist.

Issue #2598, opened 2021-09-08
Memory leak using merged stream in parJoinUnbounded


oleg-py:
fs2 3.1.1 / cats-effect 3.2.5 / Scala 2.13.5

---

Minimized reproduction:

```scala
import cats.effect.{IO, IOApp}
import cats.implicits._
import scala.concurrent.duration._

object MergeParJoinLeak extends IOApp.Simple {
  case class Box(i: Int)
  val src = fs2.Stream(1).delayBy[IO](10.millis).repeat.scanMonoid.map(Box(_))
  val run = fs2.Stream(src merge fs2.Stream()).parJoinUnbounded
    .compile.drain
}
```
This app continuously leaks `Box`es and a lot of fs2 Stream internal objects:
![изображение](https://user-images.githubusercontent.com/4143670/132464909-3823fb72-c5b7-4a30-8d6b-db4b0fde984b.png)
Apparently, not a single `Box` is released. Most stream bits seem to be nested, with the selected pattern repeating:
![изображение](https://user-images.githubusercontent.com/4143670/132466482-547bcdfa-7690-428f-a6a5-9631a664d3af.png)

---

This doesn't happen if you remove _either_ merge or parJoinUnbounded. It also doesn't leak if you replace parJoinUnbounded with merge (so that's a viable workaround)

mpilquist:
Thanks! I added a further minimization as a test here: https://github.com/typelevel/fs2/pull/2599/files

diesalbla:
@oleg-py The leak reported, with some minimisations, is covered by the new tests added https://github.com/typelevel/fs2/pull/2602, which also seems to fix the issue.

mpilquist:
Fixed in 3.1.2

PR #2599, opened 2021-09-08
Add reproduction case of leak with merge+parJoinUnbounded

mpilquist:
Reproduction case for #2598.

Fails on 3.1.1 and 3.0.0 but passes on 2.5.9.

### DIFF
--- a/core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala
+++ b/core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala
@@ -268,7 +268,11 @@ class MemoryLeakSpec extends FunSuite {
     unfold.map(x => x)
   }
 
-  leakTest("merge + parJoinUnbounded".only) {
+  leakTest("merge + parJoinUnbounded") {
     Stream(Stream.constant(1).covary[IO].merge(Stream())).parJoinUnbounded
+  }
+
+  leakTest("flatMap(flatMap) then uncons".only) {
+    Stream.constant(1).flatMap(s => Stream(s, s).flatMap(s => Stream(s))).drain
   }
 }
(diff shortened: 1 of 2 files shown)


mpilquist:
Some observations:

- None of these leak:
```scala
Stream(Stream.constant(1).covary[IO]).parJoinUnbounded // NO LEAK
Stream(Stream.constant(1).covary[IO], Stream.constant(1).covary[IO]).parJoinUnbounded // NO LEAK
Stream(Stream(Stream.constant(1).covary[IO]).parJoinUnbounded).parJoinUnbounded // NO LEAK
Stream(Stream.eval(IO(Stream.constant(1))).flatten).parJoinUnbounded // NO LEAK
Stream(Stream(Stream.constant(1).covary[IO], Stream()).parJoinUnbounded).parJoinUnbounded // NO LEAK
```

This does leak:

```scala
Stream(Stream.constant(1).covary[IO].map(Stream(_).onFinalize(IO.unit)).flatten).parJoinUnbounded
```
Whereas removing the finalizer does NOT leak:

```scala
Stream(Stream.constant(1).covary[IO].map(Stream(_)).flatten).parJoinUnbounded // NO LEAK
```

diesalbla:
The following line, which does not involve the parallelism of `parJoinUnbounded`, also causes leak: 
```Scala
    str.chunks.foreach(_ => IO.unit).interruptWhen(IO.never: IO[Either[Throwable, Unit]])
```

mpilquist:
We've been able to further minimize to:

```scala
Stream.constant(1).flatMap(s => Stream(1, 1).flatMap(_ => Stream(s))).chunks
```

diesalbla:
> We've been able to further minimize to:
> 
> ```scala
> Stream.constant(1).flatMap(s => Stream(1, 1).flatMap(_ => Stream(s))).chunks
> ```

This specific case is added and fixed in #2602 . Still looking for the main one, though.

diesalbla:
Closed by #2602

PR #2602, opened 2021-09-09
Issue-2598: Fix memory leaks and minimisations.

diesalbla:
From https://github.com/typelevel/fs2/issues/2598, we add a leak test for pure streams, one with flatMap and uncons, which does not involve parallelism.

We then make the changes in `pull` interpreter that would fix that specific leak test case.

### DIFF
--- a/core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala
+++ b/core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala
@@ -271,4 +271,12 @@ class MemoryLeakSpec extends FunSuite {
   leakTest("flatMap(flatMap) then uncons") {
     Stream.constant(1).flatMap(s => Stream(s, s).flatMap(s => Stream(s))).drain
   }
+
+  leakTest("onFinalize + flatten + drain") {
+    Stream.constant(1).covary[IO].map(Stream(_).onFinalize(IO.unit)).flatten.drain
+  }
+
+  leakTest("merge + parJoinUnbounded") {
+    Stream(Stream.constant(1).covary[IO].merge(Stream())).parJoinUnbounded
+  }
 }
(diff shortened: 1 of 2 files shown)


mpilquist:
Memory tests pass but looks like we're getting a stack overflow exception on map: https://github.com/typelevel/fs2/blob/5d4436a05cba3ee95337f1d06b5dd6064f087b15/core/shared/src/test/scala/fs2/StreamSuite.scala#L455-L461

```
  + map - regression #1335 - stack safety of map 0.004s
java.lang.StackOverflowError
  | => cat fs2.Pull$.fs2$Pull$$viewL$1(Pull.scala:839)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1161)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
	at fs2.Pull$.fs2$Pull$$go$1(Pull.scala:1190)
```

diesalbla:
@mpilquist That stack overflow is similar to the one that appears with `FlatMap`, because the call to `go` has a left-recursive call to `go`. Adding a `F.unit >>` in front of it seems to delay that recursive call.

PR #2606, opened 2021-09-11
Create `fs2.compression.checksum`

armanbilge:
Creates a new `fs.compression.Checksum[F]` trait which so far provides the `CRC32` algorithm. This is used for example in http4s compression middleware.

The JVM implementation uses `java.util.zip.CRC32`. The JS implementation is kindly provided by means of [Kafka](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/utils/Crc32.java). It's worth noting that they reimplemented CRC32 in Java for performance reasons, and since we've already ported the code here we may as well consider using it on the JVM as well.


mpilquist:
Note this is built in to scodec-bits: http://scodec.org/api/scodec-bits/1.1.0/scodec/bits/crc$.html

armanbilge:
Oops! Lets see where that gets us.

armanbilge:
Does the scodec-bits implementation support streaming? If so, it's not obvious to me how it can be used here.

armanbilge:
Ahh, it's a linear function. So actually, we don't need any of this, do we?

armanbilge:
Reopening, now that I've done my homework. CRC32's _can_ be combined due to their linearity, but implementing this is fairly involved and would not be as performant. See this in [zlib](https://github.com/madler/zlib/blob/master/crc32.c#L372).

Interestingly, the CRC32 implementation I've ported here was mentioned in https://github.com/scodec/scodec-bits/commit/d69a4251e882a7cdbcd9ef7b9fd6cbe7e404a1ec#r18797902.

Thoughts?

mpilquist:
Couple thoughts:
- It would be nice to allow the codec-bits crc code to be used in a streaming fashion (via a builder pattern -- create, add bit vectors, and then finalize when done)
- The `fs2.hash` object takes a different stylistic approach, supporting arbitrary effect types (not requiring Sync)

armanbilge:
Makes sense, thanks. Fix (1) in scodec-bits, everything becomes easier here and we should get (2) for free :) Now if I can muster up the courage to poke at that code 😅

armanbilge:
Updated to use the new builder from https://github.com/scodec/scodec-bits/pull/316. I followed the style of `fs2.hash` which made things much nicer all around.

PR #2612, opened 2021-09-17
Make Compiler.Target extend MonadCancelThrow

bplommer:
This is intended to facilitate using `Target` as a context bound when you need `MonadCancelThrow` but don't specifically need either `Sync` or `Concurrent`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -145,7 +145,13 @@ ThisBuild / mimaBinaryIssueFilters ++= Seq(
   ProblemFilters.exclude[IncompatibleMethTypeProblem]("fs2.Pull.mapOutput"),
   ProblemFilters.exclude[NewMixinForwarderProblem]("fs2.compression.Compression.gzip*"),
   ProblemFilters.exclude[NewMixinForwarderProblem]("fs2.compression.Compression.gunzip*"),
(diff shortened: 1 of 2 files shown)


kubukoz:
Personal take, I wish Target didn't extend either of these traits at all... but I guess this avoids the conflict since you can no longer require `Target` and a typeclass with only partial overlap?

bplommer:
> Personal take, I wish Target didn't extend either of these traits at all... 

Why's that?

> but I guess this avoids the conflict since you can no longer require `Target` and a typeclass with only partial overlap?

Yeah - as things stand, if you need both `Target` and `MonadCancelThrow` you either have ambiguous instances of `MonadThrow` or you pick one of `Sync` or `Concurrent` - then if you have dependencies that made both choices you end up using `Async` to satisfy both, when one or the other should have sufficed.

kubukoz:
> Why's that?

Conflicts with other context bounds when you want to use extension methods for a common subset (e.g. `.map`), thus requiring PRs like this one :D

I guess there's a reason why I use `Compiler[F, F]` instead

bplommer:
Linking https://github.com/http4s/http4s/pull/5206 in case anyone wants to weigh in on why the whole idea of using `Target` as a context bound is a bad idea.

mpilquist:
I pushed an additional commit to remove some of the intermediate classes.

PR #2614, opened 2021-09-17
Pull - fix Translate bug.

diesalbla:
Looking at the translate method, I noticed that, strangely enough,  it was not using the `view` or continuation. This means that any stream after a Translate would be dropped. This was not detected because, in the `StreamTranslateSuite`, none of the unit/property tests was appending something after the translation. Along the way, we also add a leak test that uses translate, to cover that a little bit.


mpilquist:
Nice find!

PR #2615, opened 2021-09-17
Use AtomicReference instead of Ref

### LINKED ISSUE #2611: Deadlock when using ReactiveStreams `toStream` interop with ZIO
FS2: 2.5.9
Cats Effect: 2.5.3
ZIO: 1.0.11

When the number of concurrent calls to `toStream[Task]` is more than or equal to ZIO runtime pool size we get a dead lock with all threads waiting for `IO#unsafeRunSync` to finish in `StreamSubscriber#onSubscribe` or other `on*` methods.

Repro: https://scastie.scala-lang.org/aPU1vQVWS9qRRcquUV6wcw

```scala
import fs2.interop.reactivestreams._
import zio.interop.catz._
import zio.{Task, UIO}

object ObservableFs2Test extends App {
  implicit val runtime = zio.Runtime.default
  val corePoolSize = java.lang.Runtime.getRuntime.availableProcessors() * 2 // Same as in default
  @volatile
  var i = 0
  (0 to corePoolSize).foreach { _ =>
    runtime.unsafeRunAsync_(
      fs2.Stream.empty
        .covary[Task]
        .toUnicastPublisher
        .toStream[Task]
        .compile
        .drain *>
        UIO({ i = i + 1 })
    )
  }
  while (i < corePoolSize) {}
}
```

Excerpt from VisualVM

image

notxcain:
Resolves #2611

### DIFF
--- a/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
+++ b/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
@@ -135,13 +135,14 @@ object StreamSubscriber {
             val err = new Error(s"received subscription in invalid state [$o]")
             o -> { () =>
               s.cancel()
-              throw err
+              reportFailure(err)
             }
         }
         case OnNext(a) => {
           case WaitingOnUpstream(s, r) => Idle(s) -> (() => r(a.some.asRight))
           case DownstreamCancellation  => DownstreamCancellation -> (() => ())
-          case o                       => o -> (() => throw new Error(s"received record [$a] in invalid state [$o]"))
+          case o =>
+            o -> (() => reportFailure(new Error(s"received record [$a] in invalid state [$o]")))
         }
         case OnComplete => {
           case WaitingOnUpstream(_, r) => UpstreamCompletion -> (() => r(None.asRight))
--- a/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
+++ b/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
@@ -23,11 +23,10 @@ package fs2
 package interop
 package reactivestreams
 
-import cats.ApplicativeError
-import cats.effect.{Async, Concurrent}
-import cats.implicits._
-import fs2.Stream
-import org.reactivestreams.{Subscriber, Subscription}
+import cats._
+import cats.effect._
+import cats.syntax.all._
+import org.reactivestreams._
 
 import java.util.concurrent.atomic.AtomicReference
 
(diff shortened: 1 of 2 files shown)


diesalbla:
@notxcain Does this also need to be applied on the 3.x / main branch? The bug report only mentions the 2.x series.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
```diff
@@ -72,23 +74,23 @@ final class StreamSubscriber[F[_]: ConcurrentEffect, A](val sub: StreamSubscribe
 }
 
 object StreamSubscriber {
-  def apply[F[_]: ConcurrentEffect, A]: F[StreamSubscriber[F, A]] =
+  def apply[F[_]: Async, A]: F[StreamSubscriber[F, A]] =
```
rossabaker, line 77: This will break binary compatibility.  Leaving a private overload with `ConcurrentEffect` will still complain about a missing static method for Java interop, I think.  I think one of three choices is necessary:

1. Create a new method for the `Async`, which the old one could call.
2. Break Java interop, which is probably fine, but reactivestreams is a place I'd most expect people to do weird Java shit.
3. Keep the constraint that's now stronger than needed.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
```diff
@@ -72,23 +74,23 @@ final class StreamSubscriber[F[_]: ConcurrentEffect, A](val sub: StreamSubscribe
...
-    def onNext(a: A): F[Unit]
+    def onNext(a: A): Unit
 
     /** receives error from upstream */
-    def onError(t: Throwable): F[Unit]
+    def onError(t: Throwable): Unit
 
     /** called when upstream has finished sending records */
-    def onComplete: F[Unit]
+    def onComplete(): Unit
```
rossabaker, line 93: These are also binary incompatible, but you can probably get away with exemptions here since it's package-private.

rossabaker (commented):
I have not contemplated the safety, but some drive-by MiMa concerns:

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
notxcain: I like the first one. And maybe add `@deprecated` to the old method?

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
notxcain: Yeah, I think we're fine here.

notxcain:
@diesalbla Yes, 3.x is also [affected](https://github.com/typelevel/fs2/issues/2611#issuecomment-922441084)

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
notxcain: On the other hand there's also `fs2.interop.reactivestreams` package object with a couple of methods having `ConcurrentEffect` constraint. So either 2 or 3.

notxcain: Decided to keep the old constraint.

rossabaker:
I'm not really a core contributor, but I'll review in the next couple days if none of the regulars step up.

notxcain:
We've deployed this patch to production yesterday. Works great so far.

SystemFw (approved):
I'd say let's go ahead and merge, but this code is getting a bit unstable in terms of bug count. Probably we need to do a more through review soon

mpilquist:
@notxcain Thanks! Are you going to prepare a 3.x PR as well?

notxcain:
@mpilquist sure!

Issue #2600, opened 2021-09-08
Some pipes occasionally get ahead when broadcasting through multiple pipes. 


mn98:
Some pipes occasionally get ahead by 1 element when broadcasting an un-chunked stream through multiple pipes using `broadcastThrough`.

A reproduction of the issue, using Scala 2.13.6 and fs2 3.1.1, can be found in [this project](https://github.com/mn98/fs2-snippets/blob/master/src/main/scala/snippets/BroadcastingRace.scala) or on [scastie](https://scastie.scala-lang.org/mn98/0QJ4r5N3Rt2w3sScvr7v3A/15).

SystemFw:
As explained in the chat, this is due to the fact that publishing to a Topic (which backs `broadcastThrough`) is not atomic, but rather it enqueues the same elements to a bunch of bounded channel, so it's true that `publish` will block when one of the channels is full (which is why one or more subs can't go ahead by more than 1), but it might have published to some of the subscribers in the meantime (which is why they can go ahead by 1).

Making the publishing atomic should be possible, but it isn't trivial, especially without avoiding too much contention (like in the PubSub topic, which had performance issues)

diesalbla:
@SystemFw In that case, we may want to correct the current documentation of the `broadcastThrough` method. 

> The pipes are all run concurrently with each other, but note
> that elements are pulled from the source as chunks, and the next
> chunk is pulled only when all pipes are done with processing the
> current chunk, which prevents faster pipes from getting too far ahead.
 
The phrase "only when all pipes are done", to me would indicate "not until after all pipes are done processing the current chunk". However, that does not match the operation as described in your comment.

mn98:
For me, as an end user, there's an argument for two flavours of `broadcastThrough` to exist, e.g. a `raceThrough` for cases where performance is key and a `coherentlyThrough` where control trumps performance and it's necessary to process one chunk/element at a time. I can think of use cases for both semantics.

At present the documentation describes the `coherentlyThrough` flavour. In the short term, either documenting the bug or describing the current implementation accurately would be useful.

However, longer term, I would like to see the current implementation - which is somewhere in between and is probably not the semantics anyone would reasonably want - bifurcate into the two forms described above.

SystemFw:
What do you mean by raceThrough? If you something that relaxes backpressure, you can do it by using prefetch on the pipes, you don't need a different version of broadcastThrough

mn98:
[...]

PR #2621, opened 2021-09-19
Extend `broadcastThrough` documentation

diesalbla:
Opened to address some of the inaccuracies discussed in #2600 

- Formulate the coordination condition as: "no pipe can pull and start processing any source chunk until after all pipes have pulled and started processing the previous chunk", which is a bit more accurate than "are done"

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -201,22 +201,35 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   ): Stream[F2, Either[Throwable, O]] =
     attempt ++ delays.flatMap(delay => Stream.sleep_(delay) ++ attempt)
 
-  /** Broadcasts every value of the stream through the pipes provided
-    * as arguments.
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -201,22 +201,36 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
-    * that elements are pulled from the source as chunks, and the next
-    * chunk is pulled only when all pipes are done with processing the
-    * current chunk, which prevents faster pipes from getting too far ahead.
-    *
-    * In other words, this behaviour slows down processing of incoming
-    * chunks by faster pipes until the slower ones have caught up. If
-    * this is not desired, consider using the `prefetch` and
-    * `prefetchN` combinators on the slow pipes.
+  /** Feeds the values from this stream (source) to all the given pipes,
+    * which process them in parallel. and coordinates its progress.
```
mpilquist, line 205:
```suggestion
    * which process them in parallel, and coordinates their progress.
```

PR #2622, opened 2021-09-19
Implement Files#realPath

armanbilge:
Came up while working on https://github.com/http4s/http4s/pull/5226.


io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -270,6 +270,8 @@ sealed trait Files[F[_]] extends FilesPlatform[F] {
     */
   def readRange(path: Path, chunkSize: Int, start: Long, end: Long): Stream[F, Byte]
 
+  def realPath(path: Path): F[Path]
```
mpilquist, line 273: ScalaDoc

PR #2624, opened 2021-09-20
Fix UnixSockets classpath detection logic

### LINKED ISSUE #2623: UnixSockets throws a NoClassDefFoundError when used on JDK15 or prior
When using the Scala 3 artifact of fs2-io along with JDK15 or prior, any use of `UnixSockets` results in a `NoClassDefFoundError`: 

```scala
@ fs2.io.net.unixsocket.UnixSockets[IO]
java.lang.NoClassDefFoundError: java/net/UnixDomainSocketAddress
  fs2.io.net.unixsocket.UnixSocketsCompanionPlatform.forAsync(UnixSocketsPlatform.scala:36)
  fs2.io.net.unixsocket.UnixSocketsCompanionPlatform.forAsync$(UnixSocketsPlatform.scala:34)
  fs2.io.net.unixsocket.UnixSockets$.forAsync(UnixSockets.scala:50)
  ammonite.$sess.cmd2$.<clinit>(cmd2.sc:846)
java.lang.ClassNotFoundException: java.net.UnixDomainSocketAddress
  java.net.URLClassLoader.findClass(URLClassLoader.java:435)
  ammonite.runtime.SpecialClassLoader.findClass(ClassLoaders.scala:241)
  java.lang.ClassLoader.loadClass(ClassLoader.java:589)
  java.lang.ClassLoader.loadClass(ClassLoader.java:522)
  fs2.io.net.unixsocket.UnixSocketsCompanionPlatform.forAsync(UnixSocketsPlatform.scala:36)
  fs2.io.net.unixsocket.UnixSocketsCompanionPlatform.forAsync$(UnixSocketsPlatform.scala:34)
  fs2.io.net.unixsocket.UnixSockets$.forAsync(UnixSockets.scala:50)
  ammonite.$sess.cmd2$.<clinit>(cmd2.sc:846)
```

The `UnixSockets.forAsync` method is designed to dynamically test whether it's executing on JDK16+. If so, it uses the JDK backend and otherwise it falls back to the JNR backend. When running on Scala 3 however, the `ClassNotFoundException` is thrown when loading the `JdkUnixSockets` class.

The implementation of `JdkUnixSockets` is:

```scala
object JdkUnixSockets {

  def supported: Boolean = StandardProtocolFamily.values.size > 2

  implicit def forAsync[F[_]](implicit F: Async[F]): UnixSockets[F] =
    new UnixSockets.AsyncUnixSockets[F] {
      protected def openChannel(address: UnixSocketAddress) = F.delay {
        val ch = SocketChannel.open(StandardProtocolFamily.UNIX)
        ch.connect(UnixDomainSocketAddress.of(address.path))
        ch
      }

      protected def openServerChannel(address: UnixSocketAddress) = F.blocking {
        val serverChannel = ServerSocketChannel.open(StandardProtocolFamily.UNIX)
        serverChannel.configureBlocking(false)
        serverChannel.bind(UnixDomainSocketAddress.of(address.path))
        (F.blocking(serverChannel.accept()), F.blocking(serverChannel.close()))
      }
    }
}
```

Scala 3 emits much different byte code for this than Scala 2. In particular, the anonymous functions defined inside the anonymous class are emitted by Scala 3 as static methods on the `JdkUnixSockets` object. Scala 2 emitted those anonymous functions as static methods on the anonymous class instead.

This change in strategy results in a static method on `JdkUnixSockets` for each appearance of `F.delay`. And those static methods take `UnixSocketAddress` arguments, resulting in the linkage problem.

To fix, we can simply move the anonymous class to a regular named class. We should do this for both `JdkUnixSockets` and `JnrUnixSockets`.

mpilquist:
Fixes #2623

### DIFF
--- a/io/jvm/src/main/scala/fs2/io/net/unixsocket/JdkUnixSockets.scala
+++ b/io/jvm/src/main/scala/fs2/io/net/unixsocket/JdkUnixSockets.scala
@@ -29,19 +29,22 @@ object JdkUnixSockets {
 
   def supported: Boolean = StandardProtocolFamily.values.size > 2
 
-  implicit def forAsync[F[_]](implicit F: Async[F]): UnixSockets[F] =
-    new UnixSockets.AsyncUnixSockets[F] {
-      protected def openChannel(address: UnixSocketAddress) = F.delay {
-        val ch = SocketChannel.open(StandardProtocolFamily.UNIX)
-        ch.connect(UnixDomainSocketAddress.of(address.path))
-        ch
-      }
+  implicit def forAsync[F[_]: Async]: UnixSockets[F] =
+    new JdkUnixSocketsImpl[F]
+}
 
-      protected def openServerChannel(address: UnixSocketAddress) = F.blocking {
-        val serverChannel = ServerSocketChannel.open(StandardProtocolFamily.UNIX)
-        serverChannel.configureBlocking(false)
-        serverChannel.bind(UnixDomainSocketAddress.of(address.path))
-        (F.blocking(serverChannel.accept()), F.blocking(serverChannel.close()))
-      }
-    }
+private[unixsocket] class JdkUnixSocketsImpl[F[_]](implicit F: Async[F])
+    extends UnixSockets.AsyncUnixSockets[F] {
+  protected def openChannel(address: UnixSocketAddress) = F.delay {
+    val ch = SocketChannel.open(StandardProtocolFamily.UNIX)
+    ch.connect(UnixDomainSocketAddress.of(address.path))
+    ch
+  }
+
+  protected def openServerChannel(address: UnixSocketAddress) = F.blocking {
+    val serverChannel = ServerSocketChannel.open(StandardProtocolFamily.UNIX)
+    serverChannel.configureBlocking(false)
+    serverChannel.bind(UnixDomainSocketAddress.of(address.path))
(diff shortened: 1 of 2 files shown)


armanbilge (approved):
Do we care enough to add any sort of regression tests?

mpilquist:
I don't know how we could do that. Good news is that http4s build will validate this though.

ChristopherDavenport (approved):

armanbilge:
Can't we do CI for fs2 on older JDKs? It's only UnixSockets that needs JDK 16 right, and this is specifically what we'd want to test?

vasilmkd:
It complicates the build with two JDKs, one for building, one for running.

armanbilge:
Yeah, now that its being tested downstream probably more trouble than it's worth.

mpilquist:
Yeah, not sure what that build would even look like. If we tried just running the same test suite on a different JDK, then SBT will recompile everything. We'd have to somehow publish artifacts somewhere and then run a completely different SBT project using those artifacts.

armanbilge:
> We'd have to somehow publish artifacts somewhere and then run a completely different SBT project using those artifacts.

Aren't the artifacts already pushed into GHA cache?

mpilquist:
Yes but SBT will notice the JDK changed and invalidate them / rebuild them.

diesalbla (approved):

PR #2629, opened 2021-09-22
Fix typos

TimWSpence:


TimWSpence:
Apologies if this is just being pedantic. Just noticed these as I was browsing and thought I'd fix them

Issue #2627, opened 2021-09-21
fs2 vs fs3 - chunks-to-list conversion poor performance


asjadoun:
fs2 vs fs3 chunks to list conversion performance

**fs2 version and code:** 
implementation group: 'co.fs2', name: 'fs2-core_2.13', version: '2.5.9'

```
object ChunkToListConversionFs2App extends IOApp {
  private def time[R](action: String, thunk: => R): R = {
    val start = System.currentTimeMillis()
    val result = thunk
    val end = System.currentTimeMillis()
    println(s"$action elapsed time : ${end - start} ms")
    result
  }
  override def run(args: List[String]): IO[ExitCode] = {
    implicit val ec: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
    implicit val timer: Timer[IO] = IO.timer(ExecutionContext.global)
    val stream = Stream
      .repeatEval(IO(System.currentTimeMillis()))
      .chunkN(10000)
      .evalMap(c => IO(time(s"""Using fs2 2.5.9, Chunk of ${c.size} elements conversion 'toList'""", c.toList)))
      .take(3)
      .compile
      .drain
      .unsafeRunSync()

    IO(ExitCode.Success)
  }
}
```

**Output:**
Using fs2 2.5.9, Chunk of 10000 elements conversion 'toList' elapsed time : 5 ms
Using fs2 2.5.9, Chunk of 10000 elements conversion 'toList' elapsed time : 1 ms
Using fs2 2.5.9, Chunk of 10000 elements conversion 'toList' elapsed time : 2 ms

**fs3 version and code:**
implementation group: 'co.fs2', name: 'fs2-core_2.13', version: '3.1.2'

```
object ChunkToListConversionFs3App extends IOApp {
  private def time[R](action: String, thunk: => R): R = {
    val start = System.currentTimeMillis()
    val result = thunk
    val end = System.currentTimeMillis()
    println(s"$action elapsed time : ${end - start} ms")
    result
  }
  override def run(args: List[String]): IO[ExitCode] = {
    implicit val runtime: IORuntime = IORuntime.global

    val stream: Unit = Stream
      .repeatEval(IO(System.currentTimeMillis()))
      .chunkN(10000)
      .evalMap(c => IO(time(s"""Using fs2 3.1.2, Chunk of ${c.size} elements conversion 'toList'""", c.toList)))
      .take(3)
      .compile
      .drain
      .unsafeRunSync()

    IO(ExitCode.Success)
[...]

PR #2630, opened 2021-09-22
Remove unnecessary indexing in Chunk combinators

### LINKED ISSUE #2603: performance degradation when using chunkN
We faced performance degradation when we started to use fs2 3.x in our app.
[async-profiler](https://github.com/jvm-profiling-tools/async-profiler) helped us to locate the problem:
```
--- 4219943188 ns (53.96%), 421 samples
  [ 0] scala.collection.immutable.List.reverse
  [ 1] scala.collection.immutable.Queue.tail
  [ 2] fs2.Chunk$Queue.go$1
  [ 3] fs2.Chunk$Queue.apply
  [ 4] fs2.Chunk.map
```

Code sample:
---
fs2 3.1.1 / cats-effect 3.1.1 / Scala 2.13.6
---
```scala
  val s1 = Stream.range(0, 10000).flatMap(x => Stream.emits(List.fill(10)(x)))
  val s2 = s1.chunkN(100000)

  s2.map(chunk => chunk.map(_ + 1)).compile.drain          // took 19500 ms
  s2.map(chunk => chunk.compact.map(_ + 1)).compile.drain  // took 250 ms
```

s1 is a stream of chunks all having size 10. While s2 is a stream of the single chunk, but this chunk is constructed (see `Chunk.Queue`) of many little chunks. And if you don't use `.compact` before doing `.map` you get O(L^2) time complexity where L is the number of small chunks used to construct a big chunk.

Despite the fs2.Chunk docs provide a caution about using Chunk in this way, it is constructed implicitly (inside `.chunkN`) in the provided example. And an indexed lookup (which is O(L)) is used inside `.map`.

fs2 2.x used `toChunk` in `unconsN` in order to make Chunk.Queue accessible by index for O(1). Why is it not used now?

mpilquist:
Fixes #2603 and #2627. Also addresses the issue described here: https://github.com/circe/circe-fs2/issues/273 (cc @samspills)

The `Chunk.Queue` type provides the ability to efficiently create a single chunk out of any number of individual chunks. In 2.x, `Chunk.Queue` was not itself a subtype of `Chunk`. After various individual chunks had been accumulated in to a `Chunk.Queue`, the queue had to be converted to a single chunk via `toChunk`. This operation returned a single array backed chunk. Unfortunately, this relied on reflection and various tricks to build the array backed chunk without access to a `ClassTag` at the call site and ultimately was unsound. We fixed this in https://github.com/typelevel/fs2/pull/2181.

One of the main changes in #2181 was making `Chunk.Queue` a subtype of `Chunk`. Doing so meant that `Chunk` could no longer guarantee O(1) lookup by index -- rather, it was now O(number of chunks in chunk queue). When a chunk queue consists only of singleton chunks, that becomes O(n). We added a `Chunk#compact` operation which built a chunk backed by an array, potentially primitive array.

As a result of #2181, the various combinators that used to call `Chunk.Queue#toChunk` no longer had to, resulting in a significant performance improvement as they no longer had to copy all the accumulated elements to a new array backed chunk. This tradeoff normally pays off, though not in the case where the O(n) lookup ends up dominating. The performance regressions reported in #2603 and #2627 are examples of such cases.

So why not add a call to `.compact` in the implementation of operations like `chunkN`? In short, we can't. We don't have a `ClassTag[O]` in those cases so we can't create primitive arrays. This gets us right back to the initial issue #2181 was addressing -- unsoundness as a result of too clever reflection.

This PR takes a different approach. We maintain the worst case O(n) lookup and we leave `chunkN`/`unconsN` as-is. Instead, we reimplement all the combinators on `Chunk` to avoid indexed lookup. If a user wants O(1) indexed lookup, they must call `compact`.

This PR also adds a new `compactUntagged` operation which stores the elements in an `Array[Any]` (boxed).

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -127,8 +127,11 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
     }
 
   /** Left-folds the elements of this chunk. */
-  def foldLeft[A](init: A)(f: (A, O) => A): A =
-    iterator.foldLeft(init)(f)
+  def foldLeft[A](init: A)(f: (A, O) => A): A = {
+    var res = init
+    foreach(o => res = f(res, o))
+    res
+  }
 
   /** Returns true if the predicate passes for all elements. */
   def forall(p: O => Boolean): Boolean =
@@ -201,9 +204,8 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
 
   /** Maps the supplied function over each element and returns a chunk of just the defined results. */
   def mapFilter[O2](f: O => Option[O2]): Chunk[O2] = {
-    val sz = size
     val b = collection.mutable.Buffer.newBuilder[O2]
-    b.sizeHint(sz)
+    b.sizeHint(size)
     foreach { o =>
       val o2 = f(o)
       if (o2.isDefined) b += o2.get
@@ -275,6 +277,7 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
     arr
   }
 
+  /** Converts this chunk to a `Chunk.ArraySlice`. */
   def toArraySlice[O2 >: O](implicit ct: ClassTag[O2]): Chunk.ArraySlice[O2] =
     this match {
       case as: Chunk.ArraySlice[_] if ct.wrap.runtimeClass eq as.getClass =>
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -583,33 +524,7 @@ object Chunk
     if (v.isEmpty) empty
     else if (v.size == 1) // Use size instead of tail.isEmpty as vectors know their size
       singleton(v.head)
-    else new VectorChunk(v)
-
-  private final class VectorChunk[O](v: Vector[O]) extends Chunk[O] {
```
mpilquist, line 588: Identical to `IndexedSeqChunk`

SystemFw:
So if I understand correctly, by using `foreach` in the combinators instead of `apply` directly,  we can override `foreach` in `Chunk.Queue` and keep the combinators `O(N)` rather than `O(N^2)` right?

mpilquist:
> So if I understand correctly, by using foreach in the combinators instead of apply directly, we can override foreach in Chunk.Queue to just traverse linearly once and keep the combinators O(N) rather than O(N^2) right?

Yep, that's right -- likewise for `iterator`

SystemFw (approved):
I think this is a good idea for our internal code. We might still want to explore a better apply for client code, but that's an orthogonal concern, so I'd go ahead with this

diesalbla (approved):

PR #2631, opened 2021-09-23
Make indexed traversal of `Chunk.Queue` amortized logarithmic time

mpilquist:
Per suggestion from @diesalbla

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -972,22 +972,16 @@ object Chunk
       else if (i == size - 1) chunks.last.last.get
       else {
         val (lengths, chunks) = accumulatedLengths
-        var lo = 0
-        var hi = lengths.length - 1
-        var mid = (hi - lo) / 2
-        var accLen = 0
-        var chunk: Chunk[O] = null
-        while (lo <= hi) {
-          val delta = hi - lo
-          mid = lo + (delta / 2)
-          accLen = lengths(mid)
-          chunk = chunks(mid)
-          if (i >= accLen) lo = if (delta <= 1) mid + 1 else mid
-          else if (i >= accLen - chunk.size) lo = hi + 1
-          else hi = mid
+        val j = java.util.Arrays.binarySearch(lengths, i)
+        if (j >= 0) {
+          // The requested index is exactly equal to an accumulated length so the head of the next chunk is the value to return
+          chunks(j + 1)(0)
+        } else {
+          // The requested index is not an exact match but located in the chunk after the returned insertion point 
+          val k = -j + 1
+          val accLenBefore = if (k == 0) 0 else lengths(k - 1)
+          chunks(k)(i - accLenBefore)
         }
-        val accLenBefore = accLen - chunk.size
-        chunk(i - accLenBefore)
       }
     }
 
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -978,7 +978,7 @@ object Chunk
           chunks(j + 1)(0)
         } else {
           // The requested index is not an exact match but located in the chunk after the returned insertion point
-          val k = -j + 1
+          val k = -(j + 1)
           val accLenBefore = if (k == 0) 0 else lengths(k - 1)
           chunks(k)(i - accLenBefore)
         }
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -922,16 +922,18 @@ object Chunk
     */
   final class Queue[+O] private (val chunks: SQueue[Chunk[O]], val size: Int) extends Chunk[O] {
 
-    private[this] lazy val accumulatedLengths: Array[(Int, Chunk[O])] = {
-      val arr = new Array[(Int, Chunk[O])](chunks.size)
+    private[this] lazy val accumulatedLengths: (Array[Int], Array[Chunk[O]]) = {
+      val lens = new Array[Int](chunks.size)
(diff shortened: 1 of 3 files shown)


SystemFw:
Do we know if this fixes or mitigates the original issue without the other change? I still think both should go in, but I'm curious

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -922,6 +922,18 @@ object Chunk
...
+      var accLen = 0
+      var i = 0
+      chunks.foreach { c =>
+        accLen += c.size
+        arr(i) = (accLen, c)
+        i += 1
+      }
+      arr
+    }
+
```
diesalbla, line 936: So, there could be some memory svings in returning a pair of arrays, rather than an array of pairs. First, that would avoid allocating the tuple objects. Second, it would allow boxing, and thus dereferencing, the integers.

```suggestion
    private[this] lazy val accumulatedLengths: (Array[Int], Array[Chunk[O]]) = {
      val lens = new Array[Int](chunks.size)
      val arr   = new Array[Chunk[O]](chunks.size)
      var accLen = 0
      var i = 0
      chunks.foreach { c =>
        accLen += c.size
        lens(i) = accLen
        arr(i) = c
        i += 1
      }
      (lens, arr)
    }

```

mpilquist:
Yep, on a chunk with 1 million elements, the new foreach based `toList` is 2-3x faster than the original index based `toList` but with the PR applied. I gave up waiting for the original version of toList without this PR to finish executing.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -945,19 +957,37 @@ object Chunk
...
+        while (lo <= hi) {
+          val delta = hi - lo
+          mid = lo + (delta / 2)
+          val entry = table(mid)
+          val accLen = entry._1
+          chunk = entry._2
+          if (i >= accLen) lo = if (delta <= 1) mid + 1 else mid
+          else if (i >= accLen - chunk.size) lo = hi + 1
+          else hi = mid
+        }
```
diesalbla, line 986: In line with the above, once the array of accumulated lengths is separate, perhaps we can use a Java utility for binary search. https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#binarySearch(int[],%20int,%20int,%20int)

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: Sounds good, will make the change

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: Note we're not searching for a specific key so I don't think we can reuse `binarySearch`.

diesalbla: Right... the point of the binary search is to give you the last position, `j`, in the array of accumulated lengths, which is smaller than the intended element position `i`. That would be where the chunk containing element `i` resides. 

So, for chunks of lengths `5, 3, 6, 12, 4`, your accumulated lenghts would be `5, 8, 15, 26, 30`, so looking for element 7 should yield you to the second chunk, and looking for element 29 should take you to the last one. The Java utility would remove the loop, but the result would have to be adjusted...

mpilquist: Understood but I still don't see how that's possible to implement with `Arrays.binarySearch`. Can you give it a try?

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -945,19 +959,36 @@ object Chunk
...
+          val delta = hi - lo
+          mid = lo + (delta / 2)
+          accLen = lengths(mid)
+          chunk = chunks(mid)
+          if (i >= accLen) lo = if (delta <= 1) mid + 1 else mid
+          else if (i >= accLen - chunk.size) lo = hi + 1
+          else hi = mid
+        }
+        val accLenBefore = accLen - chunk.size
+        chunk(i - accLenBefore)
```
diesalbla, line 990:
```suggestion
        // 1. What is the chunk that would contain the element? 
        var binS = Arrays.binarySearch(lengths, i)
        if (binS >= 0) {
          /* Index `i` is contained in lenghts array, which means one sum of 
          * chunks lengths is in array, so `i` is head of next chunk 
            Note that, since `i < size`, binS must be less than `lengths.length - 1` */
          chunk(binS + 1).head 
        } else {
          /* binS is (- insertion point - 1), so the index `i` is at the chunk at insertion point */ 
          val j = - binS + 1 
          val accLenBefore = if (j == 0) 0 else lengths(j-1)
          val chunk = chunks(j)
          chunk(i - accLenBefore)
        }
```
I think this should work, but given that this is array indexitis, I would not send it forth without a few tests first. 

Note that this is taking into account that we start our `lengths` with the length of the first chunk, and it ends with the size of the whole queue.

mpilquist: Oh cool, I didn't realize the Java built-in binary search returned the negative insertion point. That'll definitely work.

What's your concern regarding arrays? A Scala array *is* a Java array.

SystemFw:
>  the new foreach based toList is 2-3x faster than the original index based toList but with the PR applied.

Do I read this correctly that this PR without the `foreach` is waaay faster than current "buggy" code, but 2-3x slower than using `foreach` ? 
In any case I think it's good to merge this once ready, it will make client code better.

mpilquist:
> Do I read this correctly that this PR without the foreach is waaay faster than current "buggy" code, but 2-3x slower than using foreach ?

Yep! Turns out `n^2` is much faster than `n log n` when n = 1000000. :)

PR #2635, opened 2021-09-25
Scalafmt - Align on for comprehension arrows

diesalbla:
There may be some bike-shedding here, so to avoid much ado about next to nothing, just attach a 👍🏼  for yay or 👎🏼 for nay... 🙂


rossabaker:
It looks a little nicer but it increases the size of diffs.  For me, it's not worth the added tax on reviews, maintenance branches, and history browsing.

PR #2644, opened 2021-09-28
fix socket leak for unresolvable domains

stephennancekivell:
This PR fixes a socket / file descriptor leak when opening sockets to domains that dont resolve. eg Network[IO].client(SocketAddress(Host.fromString("nodomain.fs2.io").get, port"80"))

Fix for https://github.com/typelevel/fs2/issues/2643

Please review. The main change is just to `close` with `Resource.make` in `def client(..` rather than later in `Socket.forAsync`. Im not sure why it was in `Socket.forAsync`, the other usages I found in `fs2.io.net` seemed to be already closing. Im not sure if that instance could have been used elsewhere, it shouldnt because its private.

PS. Thanks for the excellent library 😊

### DIFF
--- a/io/jvm/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
+++ b/io/jvm/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
@@ -32,10 +32,8 @@ import java.nio.channels.{
 }
 import java.nio.channels.AsynchronousChannelGroup
 import java.nio.channels.spi.AsynchronousChannelProvider
-
 import cats.syntax.all._
 import cats.effect.kernel.{Async, Resource}
-
 import com.comcast.ip4s.{Host, IpAddress, Port, SocketAddress}
 
 private[net] trait SocketGroupCompanionPlatform { self: SocketGroup.type =>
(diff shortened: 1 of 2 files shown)


io/jvm/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
```diff
@@ -49,13 +47,13 @@ private[net] trait SocketGroupCompanionPlatform { self: SocketGroup.type =>
...
+        Resource.make(Async[F].delay {
           val ch =
             AsynchronousChannelProvider.provider.openAsynchronousSocketChannel(channelGroup)
           options.foreach(opt => ch.setOption(opt.key, opt.value))
           ch
-        }
+        })(ch => Async[F].delay(if (ch.isOpen) ch.close else ()))
 
       def connect(ch: AsynchronousSocketChannel): F[AsynchronousSocketChannel] =
         to.resolve[F].flatMap { ip =>
```
stephennancekivell, line 59: It's this `.resolve` that can fail.

mpilquist:
Nice find and fix! Thanks!

Issue #2371, opened 2021-04-20
`parEvalMap` + `Sync` => interrupt called for Scope that cannot be interrupted


Taig:
```scala
import cats.effect.{Concurrent, IO, IOApp, Sync}
import fs2.Stream

object Main extends IOApp.Simple {
  def source[F[_]](implicit F: Concurrent[F]): Stream[F, Int] =
    Stream.emits(1 to 100).covary[F].parEvalMap(maxConcurrent = 2)(x => F.pure(x))

  def runF[F[_]: Sync](source: Stream[F, Int]): F[Int] = source.compile.lastOrError

  override def run: IO[Unit] = runF[IO](source[IO]).void
}
```

```
java.lang.IllegalStateException: Scope#interrupt called for Scope that cannot be interrupted
	at fs2.internal.Scope.interruptWhen(Scope.scala:400)
	at fs2.Pull$.$anonfun$compile$12(Pull.scala:1013)
	at fs2.internal.Scope.$anonfun$acquireResource$2(Scope.scala:189)
	at cats.effect.IOFiber.runLoop(IOFiber.scala:493)
	at cats.effect.IOFiber.execR(IOFiber.scala:1077)
	at cats.effect.IOFiber.run(IOFiber.scala:137)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:372)
```

This is an error I encountered while upgrading to fs2 3. Note that when replacing the `Sync` constraint with `Concurrent` (and thereby probably also the inferred compiler instance), the code works as expected. Also `parEvalMapUnordered` does not trigger the error.

Edit: I tried to come up with a naive replacement in the meantime that just enqueues a bunch of `Deferred`s but was surprised to see the exact same stacktrace while using the usual `fs2.Stream` combinators such as `concurrently` or `merge`, so the issue is probably more fundamental than `parEvalMap`.

mpilquist:
Yeah, this happens any time a stream containing `interruptWhen` is compiled with the "synchronous compiler" provided here: https://github.com/typelevel/fs2/blob/bb69f51505d7871af318da56be5281f25fa6b5af/core/shared/src/main/scala/fs2/Compiler.scala#L169-L173

With a full `Concurrent[F]`, you get the fully featured compiler here: https://github.com/typelevel/fs2/blob/bb69f51505d7871af318da56be5281f25fa6b5af/core/shared/src/main/scala/fs2/Compiler.scala#L180-L188

[...]

PR #2650, opened 2021-09-29
Remove implicit Compiler.Target instance for Sync effects

mpilquist:


core/shared/src/main/scala/fs2/Compiler.scala
```diff
@@ -177,7 +177,10 @@ object Compiler extends CompilerLowPriority {
       private[fs2] def interruptContext(root: Unique.Token): Option[F[InterruptContext[F]]] = None
     }
 
-    implicit def forSync[F[_]: Sync]: Target[F] =
+    def forSync[F[_]: Sync]: Target[F] =
+      new SyncTarget
+
+    implicit def forSyncIO: Target[SyncIO] =
```
rossabaker, line 183: Is this because one can't translate `Concurrent` to `SyncIO` without shame?

rossabaker (commented):
I'll compile Doobie against this and inspect the fallout.

rossabaker:
This breaks doobie in four places:

1. `yolo`: where things are permitted to fail at runtime
2. `lostreaming.createLOFromStream`: this is a known landmine, and now fs2 is warning Doobie developers.  It's not obvious how to warn Doobie users.
3. `FragmentOps#copyIn`: TIL this is a landmine, but I don't know what Doobie can do about it
4. Examples. All of those examples are safe, because they weren't translated from a concurrent stream.  So this becomes a chore for users, unless Doobie makes its `Compiler[ConnectionIO, ConnectionIO]` implicit.

Cursed use of Doobie in 2 and 3 requires calling `WeakAsync.liftK`.  It is used twice in `Transactor`, safely.  If the ConnectionIO compiler were implicit and we and stopped exposing that natural transformation, I think Doobie would be safe again.  I'm not sure whether there are legitimate application uses of `WeakAsync.liftK`.

This change would also raise flags for the misuse in the original report in #2371, but it would be worth adding Scaladoc why the compiler was demoted.

rossabaker:
Can we write a `PureConc ~> SyncIO` without voiding the warranty?  I think the dirtiest thing would be wrapping `Outcome.Canceled` in a raised error, which would make it lossy?

I'm trying to brainstorm other translations from something `Concurrent` to something only `Sync` outside Doobie, which had to invent an intermediate type class to do it.

rossabaker:
This change looks approximately like [this](https://github.com/rossabaker/doobie/tree/explicit-compiler) on Doobie.  Every stream compilation in `ConnectionIO` needs to be opted into at each compile site, or with an implicit "waiver" in just the right scope.  The two operations that compile a stream of unknown origin must also accept a compiler to surface the risk.  If we provide the implicit instance in doobie, then we undo all the safety of this PR.

mpilquist:
Really doesn't seem worth it. I lean towards closing this and keeping current behavior (with the improved error message you already contributed).

rossabaker:
We hit two versions of this early in our migration.  This is great for the one that bit @zmccoy, where `Sync` was the least powerful constraint that compiled.  But it's not helpful in the Doobie, which would be better served by hiding `WeakAsync.liftK` from userspace.

Proceeding with this, _and_ making the Doobie `Compiler` implicit, _and_ hiding Doobie's dubious `~>` would improve safety in both APIs.  But, wow, that's a lot of hacks.

mpilquist:
To elaborate a bit on my comment in the Doobie issue.

In 2.5.x, stream compilation required a `Sync[F]` and we captured a `Concurrent[F]` from high level combinators like `merge` and `parJoin` in a primitive operation responsible for scope creation. During evaluation, upon encountering an interruptible scope, the capture `Concurrent[F]` instance was used to race effect evaluation against an interruption signal. This worked fine except for the case of `translate`, since the captured `Concurrent[F]` isn't useful in implementing an interruption scope in `G`. So we had a `translateInterruptible` operation captured a `Concurrent[G]` if available.

In 3.x, stream compilation requires a `Concurrent[F]` or `Sync[F]` -- this is provided by the internal `Compiler.Target[F]` interface. If a `Sync` instance is used to instantiate a `Target`, then interruption is disabled in the resulting stream. This allows `translate` to be implemented without any constraints.

One thing we could investigate: could we make `interruptWhen` a no-op when compiling with only a `Sync`?

rossabaker:
I was wondering about a no-op during my jog.  I think that's an angle worth pursuing.

Taking a modified example from #2371.  This triggers an error immediately, even though the stream is trivial and the halt signal never fires:

```scala
import cats.effect.{Concurrent, IO, IOApp, Sync}
import fs2.Stream

object Main extends IOApp.Simple {
  def source[F[_]](implicit F: Concurrent[F]): Stream[F, Int] =
    Stream(0).interruptWhen(F.never[Either[Throwable, Unit]])

  def runF[F[_]: Sync](source: Stream[F, Int]): F[Int] =
    source.compile.lastOrError

  override def run: IO[Unit] = runF[IO](source[IO]).void
}
```

Similarly, this fails immediately, when ideally we'd like it to return an empty stream:

```scala
object Main2 extends IOApp.Simple {
  def source[F[_]](implicit F: Concurrent[F]): Stream[F, Int] =
    Stream.never.interruptWhen(F.pure(Right(()): Either[Throwable, Unit]))

  def runF[F[_]: Sync](source: Stream[F, Int]): F[Int] =
    source.compile.lastOrError

  override def run: IO[Unit] = runF[IO](source[IO]).void
}
```

Replacing the Scope error with:

```scala
      case None =>
        F.pure(new Fiber[F, Throwable, Unit] {
          def cancel: F[Unit] = F.unit
          def join: F[Outcome[F, Throwable, Unit]] =
            F.pure(Outcome.Succeeded(F.unit))
        })
```

we get `Main` completing (big win!) and `Main2` never halting (maybe worse?).  Seems like a net win, but needs deeper thought.

rossabaker:
So the status quo fails fast at the expense of some unnecessary errors, and a no-op succeeds where possible at the expense of some non-terminations?

Some higher level interrupts, like `concurrently`, would be interesting.

mpilquist:
I like the no-op solution.

mpilquist:
Closing in favor of #2657

PR #2655, opened 2021-09-30
Improve IO exceptions on Node.js

armanbilge:
This improves the consistency of IO exceptions on Node.js with their JVM counterparts by providing identical messages. This came up in https://github.com/http4s/http4s/pull/5290.

### DIFF
--- a/io/js/src/main/scala/fs2/io/net/NetException.scala
+++ b/io/js/src/main/scala/fs2/io/net/NetException.scala
@@ -33,14 +33,14 @@ import scala.util.matching.Regex
 
 class SocketException(message: String = null, cause: Throwable = null)
     extends IOException(message, cause)
-private class JavaScriptSocketException(cause: js.JavaScriptException)
-    extends SocketException("Connection reset", cause)
+private class JavaScriptSocketException(message: String, cause: js.JavaScriptException)
+    extends SocketException(message, cause)
     with NoStackTrace
 object SocketException {
   private[io] def unapply(cause: js.JavaScriptException): Option[SocketException] =
     cause.exception match {
       case error: js.Error if error.message.contains("ECONNRESET") =>
-        Some(new JavaScriptSocketException(cause))
+        Some(new JavaScriptSocketException("Connection reset by peer", cause))
       case _ => BindException.unapply(cause).orElse(ConnectException.unapply(cause))
     }
 }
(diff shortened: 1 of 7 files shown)


io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala
```diff
@@ -162,13 +163,26 @@ class SocketSuite extends Fs2Suite with SocketSuitePlatform {
...
+          .void
+          .recover { case ex: BindException =>
+            assertEquals(ex.getMessage, "Address already in use")
+          }
+      } yield ()).use_ >> (for {
+        _ <- Network[IO].client(SocketAddress.fromString("not.example.com:80").get).use_.recover {
+          case ex: UnknownHostException =>
+            assertEquals(
+              ex.getMessage,
+              "not.example.com: nodename nor servname provided, or not known"
```
armanbilge, line 182: CI is failing because it returns `not.example.com: Name or service not known` instead. Seems to be different across OSes ... any ideas? OS-specific code?

armanbilge: I guess we can pass the test with either message. So, the question is which message should the Node.js implementation use?

armanbilge:
Even after these improvements, none of these errors provide messages that would match what https://github.com/http4s/http4s/pull/5290 is looking for:
* "Connection reset by peer" 
* "Broken pipe"

It'll take some additional hunting to figure out what JS errors these are supposed to be.

armanbilge:
No pressure, but would be awesome to get a release with this PR before the imminent http4s release train.
1. Gets JS compatible with https://github.com/http4s/http4s/pull/5290
2. Let's us replace `org.http4s.ProtocolException` with `fs2.io.net.ProtocolException` before the former gets branded into the binary record.

io/js/src/main/scala/fs2/io/net/NetException.scala
```diff
@@ -107,4 +108,8 @@ object UnknownHostException {
       case _ => None
     }
   private[this] val pattern = raw"(?:ENOTFOUND|EAI_AGAIN)(?: (\S+))?".r
+  private[net] val message = osMod.`type`() match {
```
mpilquist, line 114: What about Windows?

armanbilge: If someone can run on Windows and tell me what JVM does I'm happy to add it!

PR #2657, opened 2021-09-30
Make interruptWhen a no-op with the Sync compiler

rossabaker:
Addresses #2371.  Solution first discussed in #2650.

Before, compiling a concurrent stream with the Sync compiler would fail quickly and pessimistically.  Now we proceed with no-op interruption.  We can now successfully process more streams with this compiler at the expense of non-termination of some.

This is helpful when constraining to `Sync` when `Async` would have brought a more powerful compiler, or translating a Concurrent stream to a Synchronous effect like Doobie's `ConnectionIO`.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Scope.scala
+++ b/core/shared/src/main/scala/fs2/internal/Scope.scala
@@ -396,11 +396,12 @@ private[fs2] final class Scope[F[_]] private (
   def interruptWhen(haltWhen: F[Either[Throwable, Unit]]): F[Fiber[F, Throwable, Unit]] =
     interruptible match {
       case None =>
-        F.raiseError(
-          new IllegalStateException(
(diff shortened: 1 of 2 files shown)


rossabaker:
I'm trying to understand the subtle ways `Sync` and `Concurrent` compilers might diverge.  I'm surprised this test fails: I thought if `interruptWhen` isn't processed, the left would not be interrupted and still emit the `1`.  Instead, we get `Nil`.  I'm not disappointed that it behaves more like `Concurrent`, but I'm confused.

```scala
    Deferred[IO, Unit].flatMap { rightHalt =>
      val left = Stream.eval(rightHalt.get).drain ++ Stream.sleep[IO](100.millis) ++ Stream.emit(1)
      val right = Stream.eval(rightHalt.complete(())).drain
      compileWithSync(left.mergeHaltBoth(right)).toList
    }.assertEquals(List(1))
```

mpilquist:
@rossabaker That one works b/c `merge` internally compiles and runs both streams. Hence, the compilation of the internal left/right is using the `Concurrent` instance picked up from `merge`.

PR #2659, opened 2021-10-01
Build fs2 with JDK 8

vasilmkd:
This PR shows how to use `java.lang.MethodHandle` to essentially do reflection, just without the bad APIs. If initialized (lazily), these MethodHandles will try to find the appropriate methods that are available only on JDK16+ (mainly for Unix sockets) and link to them. This linking only happens once, so any further calls will be directly linked and are subject to optimizations by the JIT (unlike reflection).

This allows the project to be built using JDK 8, and be sure that JDK 8 platform specification is respected when linking against some problematic APIs like byte buffers (which is not exactly the same as building with a newer JDK and targeting an old one). How highly this feature is desired, is another debate.

Usage of new APIs works and it is tested in the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         scala: [3.0.2, 2.12.15, 2.13.6]
-        java: [adoptium@17]
+        java: [adoptium@8, adoptium@17]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch (full)
@@ -84,7 +84,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         scala: [2.13.6]
-        java: [adoptium@17]
+        java: [adoptium@8]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch (full)
@@ -152,7 +152,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         scala: [2.13.6]
-        java: [adoptium@17]
+        java: [adoptium@8, adoptium@17]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Download target directories (3.0.2)
--- a/build.sbt
+++ b/build.sbt
@@ -25,7 +25,7 @@ val NewScala = "2.13.6"
 ThisBuild / crossScalaVersions := Seq("3.0.2", "2.12.15", NewScala)
 
 ThisBuild / githubWorkflowEnv += ("JABBA_INDEX" -> "https://github.com/typelevel/jdk-index/raw/main/index.json")
-ThisBuild / githubWorkflowJavaVersions := Seq("adoptium@17")
+ThisBuild / githubWorkflowJavaVersions := Seq("adoptium@8", "adoptium@17")
 
 ThisBuild / spiewakCiReleaseSnapshots := true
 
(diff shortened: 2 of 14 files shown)


vasilmkd:
`WatcherSuite`
1. `supports watching a file - for modifications`
2. `supports watching a directory - static recursive watching`
are two unit tests that seem to deadlock when run on JDK 8.

vasilmkd:
No, it is an issue with `munit-cats-effect-3` `1.0.6`, specifically https://github.com/typelevel/munit-cats-effect/pull/124 and I shouldn't have merged this PR https://github.com/typelevel/fs2/pull/2656.

mpilquist:
This is a lot of complexity. What do we get in return?

vasilmkd:
Nothing really, other than being able to build with JDK 8. I wanted to see how hard it would be. Feel free to close the PR.

mpilquist:
OK gotcha. Definitely interesting as a proof of concept and happy to keep PR open, just very much afraid of build complexity.

vasilmkd:
Or I will close it, I still want to experiment with a few things before that.

vasilmkd:
I updated the description, but overall, there is not much going on for this PR. The diff is large because of Java's verbosity and ceremony around try/catch and checked exceptions. Some code may be factored out, but we then get bitten by Java's package visibility problems, and the duplication starts to seem necessary. This will forever be the case, until Scala decides to support annotations for creating guaranteed `static final` fields. These are necessary for GraalVM native image to be able to resolve them at compile time.

vasilmkd:
> `WatcherSuite`
> 
> 1. `supports watching a file - for modifications`
> 2. `supports watching a directory - static recursive watching`
>    are two unit tests that seem to deadlock when run on JDK 8.

@mpilquist Regardless of the outcome of this PR, I'd be interested to hear your thoughts on this, it seems that these 2 tests just hang on JDK 8, at least they do on my local machine. Is there anything particular about these 2 tests and can you maybe try to reproduce the issue? Thanks.

vasilmkd:
The `Modified` events seem to not trigger every time on MacOS with JDK 8, seems like a platform issue.

vasilmkd:
I'm still experimenting, it might be possible to define everything in Scala and cut down on the boilerplate.

vasilmkd:
Scala 3 does not have enough support for this feature however. Scala 2 works just fine.

https://github.com/lampepfl/dotty/issues/11332

rossabaker:
I broke Java 8 compatibility in a PR yesterday, only caught by @mpilquist's watchful eye.  An http4s PR is now hung only on Java 8 starting with the fs2-3.1.4 upgrade.  CI on Java 8 would be highly beneficial!

vasilmkd:
But releasing fs2 compiled using JDK8 would circumvent all of these problems, wouldn't it?

mpilquist:
Yes, that's correct. At the cost of the complexity in this PR though.

vasilmkd:
This isn't ideal, I agree, I don't want to merge it either. We will be able to write all of this in Scala, but dotty is missing some features for the support of these methods. I am trying to port the Scala 2 support to dotty, but it's not going great at the moment.

mpilquist:
Agreed. Even if we could write it in Scala, I'd still lean against using varhandles to this degree. If it was only 1 or 2 methods we had to call, then sure, but it's a lot more than that. :)

PR #2661, opened 2021-10-03
Upgrade scalacheck-effect to 1.0.3

mpilquist:

PR #2663, opened 2021-10-04
Re-working `readReadable` API

armanbilge:
So far this fixes the issue reported in https://github.com/typelevel/fs2/pull/2661, but I think the API for converting a Node.js [`Readable` stream](https://nodejs.org/api/stream.html#stream_readable_streams) is still unsafe. I have another idea I want to try.

### DIFF
--- a/io/js/src/main/scala/fs2/io/ioplatform.scala
+++ b/io/js/src/main/scala/fs2/io/ioplatform.scala
@@ -41,6 +41,7 @@ import scala.scalajs.js.|
 
 private[fs2] trait ioplatform {
 
+  @deprecated("Use readReadableResource for safer variant", "3.1.4")
   def readReadable[F[_]](
       readable: F[Readable],
       destroyIfNotEnded: Boolean = true,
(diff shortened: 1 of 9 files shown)


io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -41,58 +42,83 @@ import scala.scalajs.js.|
...
+    (for {
+      dispatcher <- Dispatcher[F]
+      queue <- Queue.synchronous[F, Option[Unit]].toResource
+      error <- F.deferred[Throwable].toResource
+      // Implementation Note: why suspend in `SyncIO` and then `unsafeRunSync()` inside `F.delay`?
+      // In many cases creating a `Readable` starts async side-effects (e.g. negotiating TLS handshake or opening a file handle).
+      // Furthermore, these side-effects will invoke the listeners we register to the `Readable`.
+      // Therefore, it is critical that the listeners are registered to the `Readable` _before_ these async side-effects occur:
+      // in other words, before we next yield (cede) to the event loop. Because an arbitrary effect `F` (particularly `IO`) may cede at any time,
+      // our only recourse is to suspend the entire creation/listener registration process within a single atomic `delay`.
```
armanbilge, line 72: ☝️

armanbilge:
I'd really appreciate some critical eyes on this. These methods for translating between Node.js and fs2 Streams are at the heart of all the implementations (Sockets, TLS, files). So it's really important to get this right.

armanbilge:
I `publishedLocal` and tested against ember-{client/server}. No regressions. I'd like to test skunk too, but it's easier to use the CI there so I'll wait for a hash snapshot when this merges.

PR #2666, opened 2021-10-05
Publisher.toStream can demand for more than one element

yanns:
Context: https://discord.com/channels/632277896739946517/632310980449402880/894663878133317722
Asking for more than one element allows a much better performance.
For example, when subscribing to a mongo publisher, we can use the opened cursor in much less queries.


reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
```diff
@@ -69,15 +69,19 @@ final class StreamSubscriber[F[_], A](
...
 }
 
 object StreamSubscriber {
 
+  def apply[F[_]: Async, A](bufferSize: Long): F[StreamSubscriber[F, A]] =
+    fsm[F, A](bufferSize).map(new StreamSubscriber(_))
+
   def apply[F[_]: Async, A]: F[StreamSubscriber[F, A]] =
-    fsm[F, A].map(new StreamSubscriber(_))
+    apply(bufferSize = 10L)
```
yanns, line 84: This changes the previous behavior asking elements one by one.
It's also a magic number coming from a quick thinking about the trade-offs memory vs performance. 
Should we handle that differently? Using `1L` to keep the previous behavior? Using some configuration?

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
```diff
@@ -69,15 +69,19 @@ final class StreamSubscriber[F[_], A](
     sub.onError(t)
   }
 
-  def stream(subscribe: F[Unit]): Stream[F, A] = sub.stream(subscribe)
+  def streamChunk(subscribe: F[Unit]): Stream[F, Chunk[A]] = sub.stream(subscribe)
+  def stream(subscribe: F[Unit]): Stream[F, A] = streamChunk(subscribe).unchunks
```
mpilquist, line 73: Echoing @diesalbla's comment here -- I think we should offer just the `stream(...): Stream[F, A]` method and let folks manually call `.chunks` if they want to operate on chunks explicitly.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
```diff
@@ -69,15 +69,19 @@ final class StreamSubscriber[F[_], A](
...
 
   private def nonNull[B](b: B): Unit = if (b == null) throw new NullPointerException()
 }
 
 object StreamSubscriber {
 
+  def apply[F[_]: Async, A](bufferSize: Long): F[StreamSubscriber[F, A]] =
```
mpilquist, line 80: The `bufferSize` should be an `Int` here as `Chunk` is `Int` based.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
mpilquist: What's the behavior if requesting 10 but there's only 5 available?

This does feel a bit too magical IMO -- I lean towards setting it to 1 or an arbitrary power of 2 :)

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
```diff
@@ -102,19 +106,23 @@ object StreamSubscriber {
...
 
     /** producer for downstream */
-    def dequeue1: F[Either[Throwable, Option[A]]]
+    def dequeue1: F[Either[Throwable, Option[Chunk[A]]]]
 
     /** downstream stream */
-    def stream(subscribe: F[Unit])(implicit ev: ApplicativeError[F, Throwable]): Stream[F, A] =
+    def stream(
+        subscribe: F[Unit]
+    )(implicit ev: ApplicativeError[F, Throwable]): Stream[F, Chunk[A]] =
```
mpilquist, line 114: Change return here to `Stream[F, A]`

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
```diff
@@ -126,9 +134,10 @@ object StreamSubscriber {
 
     sealed trait State
     case object Uninitialized extends State
-    case class Idle(sub: Subscription) extends State
+    case class Idle(sub: Subscription, buffer: Vector[A]) extends State
```
mpilquist, line 137: Change `Vector[A]` to `Chunk[A]` and use `++` to build up the buffer.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
```diff
@@ -126,9 +134,10 @@ object StreamSubscriber {
...
     case class RequestBeforeSubscription(req: Out => Unit) extends State
-    case class WaitingOnUpstream(sub: Subscription, elementRequest: Out => Unit) extends State
+    case class WaitingOnUpstream(sub: Subscription, buffer: Vector[A], elementRequest: Out => Unit)
```
mpilquist, line 139: Same here, change buffer to `Chunk[A]`

mpilquist (changes requested):
Looks great -- some minor comments

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
yanns: It only 5 are available, then we build a chunk of 5 elements, and it's being send in the `OnComplete` step: https://github.com/typelevel/fs2/pull/2666/files#diff-493b2533a887032dfbf8ea09dca7201413e124b5f45732ad6a877b8d7f69ba2cR178-R180

By setting it to `1` by default can lead to bad performances by default. I'd suggest to take an arbitrary number, not too high to avoid buffering too many elements in memory. `16`?

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
yanns: https://github.com/typelevel/fs2/pull/2666/commits/f112b4ccb8da9ac7f2c1dae413376cc6bd150ecc

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
yanns: https://github.com/typelevel/fs2/pull/2666/commits/769b2fe568a2627b454fbbc7b3f6bcfb2c17968a

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
yanns: https://github.com/typelevel/fs2/pull/2666/commits/769b2fe568a2627b454fbbc7b3f6bcfb2c17968a

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
yanns: https://github.com/typelevel/fs2/pull/2666/commits/d99c684026159103000af81b482f4c6de25fc6c3

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
yanns: https://github.com/typelevel/fs2/pull/2666/commits/d99c684026159103000af81b482f4c6de25fc6c3

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
yanns: https://github.com/typelevel/fs2/pull/2666/commits/5de1aa4e0bbcb109a875900a0b9b6ee55d424b4b

notxcain (approved):
🚀

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
Daenyth: I would strongly recommend *not* changing existing behavior in source-compatible ways

mpilquist: Yep agreed, let's stick with 1 here as that's what every has used for years.

diesalbla:
@yanns That solution would be mostly source-compatible. However, we are also keen on binary compatibility, which is to say, the contents of the compiled bytecode across versions. The current version: 
```Scala
    // current API
    def toStream[F[_]: Async]: Stream[F, A] =
      fromPublisher(publisher)
```
In the bytecode, we have a method `toStream` with one parameter, of type `cats.effect.Async`. Instead, a Scala method with a default parameter: 
```Scala
    def toStream[F[_]: Async](bufferSize: Int = StreamSubscriber.DefaultBufferSize): Stream[F, A] =
      fromPublisher(publisher, bufferSize)
```
is compiled in the bytecode to a single method with two parameters, not to two methods with or without the parameter. Defaults values are inserted, by the compiler, in the "caller" bytecode, as arguments for the two-parameter method. However, the existing method would then be missing in the new bytecode version. As a consequence, "callers" to that method, when linked to the new bytecode, would fail.

Thus, I would recommend keeping the existing method, but add to it a deprecation notice.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala
mpilquist: 1 versus 10 versus 16 could make a performance difference in certain types of streams but not others, where folks would need 100 or 1000 or more. Hence, I think it's best if we make no assumptions at all. So let's go with deprecating the old API and forcing the buffer size decision on to users.

yanns: https://github.com/typelevel/fs2/pull/2666/commits/4ad2bbb6ac4ef7cb44d64f960cacbf3ff70627b0
to keep the existing behavior + deprecations

reactive-streams/src/main/scala/fs2/interop/reactivestreams/package.scala
```diff
@@ -57,6 +67,11 @@ package object reactivestreams {
   implicit final class PublisherOps[A](val publisher: Publisher[A]) extends AnyVal {
 
     /** Creates a lazy stream from an `org.reactivestreams.Publisher` */
+    def toStream[F[_]: Async](bufferSize: Int): Stream[F, A] =
```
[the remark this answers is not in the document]
notxcain: Perfect.

yanns:
We now have a breaking change:
```
[error] fs2-reactive-streams: Failed binary compatibility check against co.fs2:fs2-reactive-streams_2.12:3.1.4! Found 1 potential problems
[error]  * extension method toStream$extension(org.reactivestreams.Publisher,cats.effect.kernel.Async)fs2.Stream in object fs2.interop.reactivestreams.package#PublisherOps does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.interop.reactivestreams.package#PublisherOps.toStream$extension")
```

mpilquist:
I confirmed this change is *not* binary compatible.

Compiling this program against 3.1.4 and then running against this PR results in the following linkage error:
```scala
package test

import fs2._, fs2.interop.reactivestreams._
import cats.effect._

object App extends IOApp.Simple {
  def run = {
    val upstream: Stream[IO, Int] = Stream(1, 2, 3).covary[IO]
    val publisher: Resource[IO, StreamUnicastPublisher[IO, Int]] = upstream.toUnicastPublisher
    val downstream: Stream[IO, Int] = Stream.resource(publisher).flatMap(_.toStream[IO])
    downstream.foreach(IO.println).compile.drain
  }
}
```

```
➜  ./target/universal/stage/bin/fs2test
java.lang.NoSuchMethodError: 'fs2.Stream fs2.interop.reactivestreams.package$PublisherOps$.toStream$extension(org.reactivestreams.Publisher, cats.effect.kernel.Async)'
	at test.App$.$anonfun$run$1(app.scala:10)
	at fs2.Stream.$anonfun$flatMap$1(Stream.scala:1152)
	at fs2.Pull$FlatMapR$1.go$2(Pull.scala:1000)
	at fs2.Pull$FlatMapR$1.unconsed(Pull.scala:1008)
	at fs2.Pull$FlatMapR$1.out(Pull.scala:1017)
	at fs2.Pull$ViewRunner$1.outLoop$1(Pull.scala:928)
	at fs2.Pull$ViewRunner$1.out(Pull.scala:930)
	at fs2.Pull$.$anonfun$compile$24(Pull.scala:1175)
	at fs2.Pull$.$anonfun$compile$2(Pull.scala:903)
	at cats.effect.IOFiber.next$2(IOFiber.scala:361)
	at cats.effect.IOFiber.runLoop(IOFiber.scala:372)
	at cats.effect.IOFiber.execR(IOFiber.scala:1169)
	at cats.effect.IOFiber.run(IOFiber.scala:128)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:410)
Exception in thread "main" java.lang.NoSuchMethodError: 'fs2.Stream fs2.interop.reactivestreams.package$PublisherOps$.toStream$extension(org.reactivestreams.Publisher, cats.effect.kernel.Async)'
	at test.App$.$anonfun$run$1(app.scala:10)
	at fs2.Stream.$anonfun$flatMap$1(Stream.scala:1152)
	at fs2.Pull$FlatMapR$1.go$2(Pull.scala:1000)
	at fs2.Pull$FlatMapR$1.unconsed(Pull.scala:1008)
	at fs2.Pull$FlatMapR$1.out(Pull.scala:1017)
	at fs2.Pull$ViewRunner$1.outLoop$1(Pull.scala:928)
	at fs2.Pull$ViewRunner$1.out(Pull.scala:930)
	at fs2.Pull$.$anonfun$compile$24(Pull.scala:1175)
	at fs2.Pull$.$anonfun$compile$2(Pull.scala:903)
	at cats.effect.IOFiber.next$2(IOFiber.scala:361)
	at cats.effect.IOFiber.runLoop(IOFiber.scala:372)
	at cats.effect.IOFiber.execR(IOFiber.scala:1169)
	at cats.effect.IOFiber.run(IOFiber.scala:128)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:410)
```

AL333Z:
Definitely not a binary compatibility expert here, but the diffs in the bytecode https://editor.mergely.com/Pq7KTAXv/ are showing a new `bufferSize` arg (around line 30 and following).

yanns:
Can someone trigger the workflow?

yanns:
I've made a little talk about this PR: https://yanns.github.io/blog/2021/12/21/fs2-mongo-reactivestreams/

PR #2667, opened 2021-10-05
Flush CharsetDecoder to produce final chunk

### LINKED ISSUE #2664: `decodeWithCharset` doesn't flush
In fs2-3.1.3, this test fails:

```scala
    // Found by scalacheck in old http4s implementation
    test("handle overflow") {
      if (isJVM) { // Charset not supported by Scala.js
        val cs = Charset.forName("x-ISCII91")
        val s = Stream(-36.toByte)
          .covary[Fallible]
          .through(decodeWithCharset(cs))
          .compile
          .string
        assertEquals(s, Right("ी"))
      }
    }
```

~~The output buffer is sized to one character based on chunk length and average size queried from the decoder, but the byte expands to two characters.  We need to check the `CoderResult` for an overflow.~~ [edit: misdiagnosis]

This is one of the rare character sets that needs to be flushed after the final input.  You can find more [here](https://github.com/openjdk/jdk17u/search?p=2&q=implFlush).

rossabaker:
Only necessary in a handful of non-UTF-8 charsets.  Setting the last input flag on the final `decode` isn't sufficient: the decoder state must be flushed in a separate call.

Fixes #2664.

### DIFF
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -21,6 +21,7 @@
 
 package fs2
 
+import cats.ApplicativeThrow
 import java.nio.{Buffer, ByteBuffer, CharBuffer}
 import java.nio.charset.{
   CharacterCodingException,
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/text.scala
```diff
@@ -250,9 +251,32 @@ object text {
...
 
+    def flush(
+        decoder: CharsetDecoder,
+        out: CharBuffer
+    ): Pull[F, String, Unit] = {
+      out.clear()
+      decoder.flush(out) match {
+        case res if res.isUnderflow =>
+          if (out.position() > 0) {
+            out.flip()
```
mpilquist, line 265: I think we need a `(out: Buffer).flip()` here to maintain JDK8 support.

PR #2681, opened 2021-10-17
Remove unused dependency to scalajs-dom

armanbilge:
I was pretty shocked to find fs2-io the top user of scalajs-dom for browsers...

https://mvnrepository.com/artifact/org.scala-js/scalajs-dom/usages

### DIFF
--- a/io/js/src/main/scala/fs2/io/internal/ByteChunkOps.scala
+++ b/io/js/src/main/scala/fs2/io/internal/ByteChunkOps.scala
@@ -23,6 +23,7 @@ package fs2.io.internal
 
 import fs2.Chunk
 import fs2.internal.jsdeps.node.bufferMod
+import fs2.internal.jsdeps.std
 
 import scala.scalajs.js.typedarray.{ArrayBuffer, TypedArrayBuffer}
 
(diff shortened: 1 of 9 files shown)


armanbilge:
Darn, it's never easy.

armanbilge:
So annoyingly it seems that ScalablyTyped has convoluted the `stUseScalaJsDom` to refer to not just scalajs-dom but also all the types in the SJS stand lib (e.g., `js.Error`, `Uint8Array`, etc.). So these all had to be switched to its own generated facades for those.

armanbilge:
> Are we shading the SJS standard library now?

Not really. There's just another set of facades (under our internal namespace) for the same underlying Javascript types. At the Scala level these appear to be unrelated classes, even though they are modelling the same Javascript types.

See for example this line which converts between the SJS std lib facade for `Uint8Array` to the ST-auto-generated facade for `Uint8Array`:
https://github.com/typelevel/fs2/pull/2681/files#diff-98450717b1764f56ec34f86f4c240563625aa001a68b64f062ec827c8ffe342eR36

> Does the resulting artifact remain interoperable with folks using the SJS standard library?

Yes. In fact, _we_'re still using the SJS standard library ourselves. This change affects only how we interact with our internal, private Node.js facade.

PR #2711, opened 2021-11-16
Update scalafmt to 3.1.1

fthomas:
This is #2708 but without the changes in the scalafix directory. It also adds `project.excludeFilters = [ "scalafix/*" ]` to `.scalafmt.conf` so that future Scalafmt updates do not modify files there again. I've tested locally that the exclude filter works.


mpilquist:
Thanks! I hadn't gotten a chance yet to give this a try. Looks great!

PR #2716, opened 2021-11-21
rethrow

nikiforo:
I expect this property to hold:
```
Stream[F, Either[Throwable, O]].evalMap(_.pure.rethrow).mask <-> .rethrow.mask
```

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2378,11 +2378,20 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       rt: RaiseThrowable[F2]
   ): Stream[F2, O2] =
     this.asInstanceOf[Stream[F, Either[Throwable, O2]]].chunks.flatMap { c =>
-      val firstError = c.collectFirst { case Left(err) => err }
-      firstError match {
-        case None    => Stream.chunk(c.collect { case Right(i) => i })
-        case Some(h) => Stream.raiseError[F2](h)
-      }
+      val size = c.size
+      val buffer = new ArrayBuffer[O2](size)
+      var i = 0
+      var exOpt: Option[Throwable] = None
+      while (i < size)
+        c(i) match {
+          case Left(ex) =>
+            exOpt = ex.some
+            i = size
+          case Right(o) =>
+            buffer += o
+            i += 1
+        }
+      Stream.emits(buffer) ++ exOpt.map(Stream.raiseError[F2]).getOrElse(Stream.empty)
     }
 
   /** Left fold which outputs all intermediate results.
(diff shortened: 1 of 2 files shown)


mpilquist:
I'm not sure about this one. Seems strange for `rethrow` to reorder each input chunk in to all the successes then all the failures.

nikiforo:
> I'm not sure about this one. Seems strange for `rethrow` to reorder each input chunk in to all the successes then all the failures.

`rethrow` is a `MonadError` aware operation. Thus, I think that the method should be consistent with `MonadError`'s `rethrow`, So, I think that property
```
Stream[F, Either[Throwable, O]].evalMap(_.pure.rethrow).mask <-> .rethrow.mask
```
should hold. It implies that all elements **before** the **first** error should be emitted and only then the **first** error should be raised. Implementation is written with this in mind - we gather all the elements that precede an error, emit them and later if an error is found the error is raised.

mpilquist:
Hm, maybe I misunderstood this implementation. Consider:

```scala
val c = Chunk(Right(1), Right(2), Left(new Exception), Right(3))
val out = Stream.chunk(c).covary[SyncIO].rethrow.mask.compile.toList.unsafeRunSync()
```

Currently `out` returns `List()`, which seems problematic. If I understand this PR though, it would return `List(1, 2, 3)`, which also seems problematic. I'd expect `List(1, 2)`.

nikiforo:
I think that line might have confused you...
https://github.com/typelevel/fs2/pull/2716/files#diff-c74e0aabdb2667a6642b051521123685e2957facf8f23b76c10fddc9cda2727fR2389

I used it as an optimization not to check two variable in the loop conditions. And it's definetely a premature one..

nikiforo:
> Hm, maybe I misunderstood this implementation. Consider:
> 
> ```scala
> val c = Chunk(Right(1), Right(2), Left(new Exception), Right(3))
> val out = Stream.chunk(c).covary[SyncIO].rethrow.mask.compile.toList.unsafeRunSync()
> ```
> 
> Currently `out` returns `List()`, which seems problematic. If I understand this PR though, it would return `List(1, 2, 3)`, which also seems problematic. I'd expect `List(1, 2)`.

It would return List(1,2)

mpilquist:
Ha, yes, I missed that. Thanks!

mpilquist:
Looks like the doctest needs updating:
```
failing seed for Stream.scala:2381: rethrow.example at line 2386: Stream(Right(1), Right(2), Left(new RuntimeException), Right ... is weKPF3ZCzFBDnzPDtdn0d_pSTeEQvqNmBoItEbVyawD=
1945
failing seed for Stream.scala.Stream.scala:2381: rethrow.example at line 2386: Stream(Right(1), Right(2), Left(new RuntimeException), Right ... is weKPF3ZCzFBDnzPDtdn0d_pSTeEQvqNmBoItEbVyawD=
1946
[info] ! Stream.scala.Stream.scala:2381: rethrow.example at line 2386: Stream(Right(1), Right(2), Left(new RuntimeException), Right ...: Falsified after 0 passed tests.
1947
[info] > Labels of failing property: 
1948
[info] 'List(1, 2, -1)' is not equal to 'List(-1)'
```

nikiforo:
I looked at comments for `rethrow` as well. However, they are correct. I also had an urge to change elements in the test. Yet, List(1, 2, -1) is still the most understandable option from my POV.

PR #2720, opened 2021-11-24
Improve memory efficiency of Chunk.filter and replace internal use of mutable.Buffer with mutable.ArrayBuilder

mpilquist:
This PR:
- replaces uses of mutable.Buffer with mutable.ArrayBuilder
- allocates primitive arrays when possible (e.g. filter)

```
# Map

## new Array(size)
[info] ChunkBenchmark.map           16  thrpt    5  20039615.265 ± 492279.237  ops/s
[info] ChunkBenchmark.map          256  thrpt    5   1278011.968 ±  42407.622  ops/s
[info] ChunkBenchmark.map         4096  thrpt    5     46243.164 ±    295.401  ops/s

## ArrayBuilder
[info] ChunkBenchmark.map           16  thrpt    5  43344785.011 ± 665760.099  ops/s
[info] ChunkBenchmark.map          256  thrpt    5   3551895.914 ±  12759.777  ops/s
[info] ChunkBenchmark.map         4096  thrpt    5     69298.960 ±   4435.243  ops/s

# Filter

## ArrayBuilder
[info] ChunkBenchmark.filter           16  thrpt    5  30132296.897 ± 378366.725  ops/s
[info] ChunkBenchmark.filter          256  thrpt    5   2326540.853 ±   7729.015  ops/s
[info] ChunkBenchmark.filter         4096  thrpt    5     66920.922 ±   1146.317  ops/s

## mutable.Buffer
[info] ChunkBenchmark.filter           16  thrpt    5  15843176.346 ± 109023.863  ops/s
[info] ChunkBenchmark.filter          256  thrpt    5   1031608.241 ±   3938.483  ops/s
[info] ChunkBenchmark.filter         4096  thrpt    5     55229.486 ±   4883.909  ops/s
```

PR #2722, opened 2021-11-26
Create a Chunk from an IterableOnce

peterneyens:
The implementation is basically the one from `iterable` and `chain`.

I replaced the implementations in `iterable` and `chain` with a call to `iterableOnce`. This does add an `Iterator#isEmpty` call.


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -596,18 +596,24 @@ object Chunk
...
     })
 
+  /** Creates a chunk from a `scala.collection.IterableOnce`. */
+  def iterableOnce[O](i: collection.IterableOnce[O]): Chunk[O] = {
+    val itr = i.iterator
+    if (itr.isEmpty) empty
+    else {
+      val head = itr.next()
+      if (itr.hasNext) {
+        val bldr = collection.mutable.Buffer.newBuilder[O]
```
nikiforo, line 609: According to this
https://github.com/typelevel/fs2/pull/2720/files#diff-e8fdf3ccd3d2479432355b3abf34a16cd548373afeddeaf231b8d8a1147b1d48R115
ArrayBuilder shows x2 performance compared to mutable.Buffer, shouldn't we change a type of builder used here?

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -596,18 +596,24 @@ object Chunk
       case ix: GIndexedSeq[O] => indexedSeq(ix)
       case _ =>
         if (i.isEmpty) empty
```
nikiforo, line 598: emptiness gets checked in `iterableOnce`, can't we remove this check from here and `L626`?

core/shared/src/main/scala/fs2/Chunk.scala
peterneyens: In those cases we have the (expected) size, we don't in this case.

diesalbla (approved):

peterneyens:
I already forgot `IterableOnce` wasn't a thing in 2.12.

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: ArrayBuilder should still beat mutable buffer though, even without a size hint.

core/shared/src/main/scala/fs2/Chunk.scala
nikiforo: Now the implementation has changed, this comment is outdated

core/shared/src/main/scala/fs2/Chunk.scala
peterneyens: Using `ArrayBuilder` requires us to add a `ClassTag` constraint, I don't think that is something that we want?

That would mean that we couldn't use it in `iterable` and `chain`.

mpilquist: Merge the main branch in to this PR and then use `makeClassTag[Any]`. We'll have to cast the resulting chunk back to a `Chunk[A]` just like we do in methods like `filter`.

mpilquist:
Thanks!

PR #2727, opened 2021-11-29
ioJS: handle NoSuchFileException in isDirectory, isFile and isSymbolicLink

yurique:
In nodejs, calling `stat` throws an exception if the path does not exist, thus making `isDirectory`, `isFile` and `isSymbolicLink` fail as well. In jvm, on the other hand, the underlying calls do not throw and return `false` when the path does not exist.

This is an attempt to make the behaviour of these functions more aligned across platforms.

---

~There's a couple of TODOs in the tests, not sure how important these are:~
* ~not sure how to easily create a symlink for tests~
* ~not sure how to create a "hidden" file for tests~

### DIFF
--- a/io/shared/src/test/scala/fs2/io/file/FilesSuite.scala
+++ b/io/shared/src/test/scala/fs2/io/file/FilesSuite.scala
@@ -614,18 +614,60 @@ class FilesSuite extends Fs2Suite with BaseFileSuite {
   }
 
   group("isHidden") {
-    // TODO test the true-case
 
-    test("returns false if the path is for a readable file") {
+    test("returns false if the path is for a non-hidden file") {
       tempFile
         .use(Files[IO].isHidden(_))
         .assertEquals(false)
     }
 
-    test("returns false if the path is for a directory") {
+    test("returns false if the path is for a non-hidden directory") {
       tempDirectory
         .use(Files[IO].isHidden(_))
         .assertEquals(false)
+    }
+
+    test("returns true if the path is for a hidden file") {
+      val hiddenFile = for {
+        dir <- tempDirectory
+        hiddenFilePath = dir.resolve(".should-be-hidden")
+        _ <- Resource.make(Files[IO].createFile(hiddenFilePath)) { _ =>
+          Files[IO].deleteIfExists(hiddenFilePath).void
+        }
+      } yield hiddenFilePath
+      hiddenFile
+        .use(Files[IO].isHidden(_))
+        .assertEquals(true)
+    }
+
+    test("returns false if the path is for a non-hidden file in a hidden directory") {
+      val fileInHiddenDir = for {
+        dir <- tempDirectory
+        hiddenDirPath = dir.resolve(".should-be-hidden")
+        _ <- Resource.make(Files[IO].createDirectory(hiddenDirPath)) { _ =>
+          Files[IO].deleteIfExists(hiddenDirPath).void
+        }
+        filePath = hiddenDirPath.resolve("not-hidden")
+        _ <- Resource.make(Files[IO].createFile(filePath)) { _ =>
+          Files[IO].deleteIfExists(filePath).void
+        }
+      } yield filePath
+      fileInHiddenDir
+        .use(Files[IO].isHidden(_))
+        .assertEquals(false)
+    }
+
+    test("returns true if the path is for a hidden directory") {
+      val hiddenDir = for {
+        dir <- tempDirectory
+        hiddenDirPath = dir.resolve(".should-be-hidden")
+        _ <- Resource.make(Files[IO].createDirectory(hiddenDirPath)) { _ =>
+          Files[IO].deleteIfExists(hiddenDirPath).void
+        }
+      } yield hiddenDirPath
+      hiddenDir
+        .use(Files[IO].isHidden(_))
+        .assertEquals(true)
     }
 
     test("returns false if the path does not exist") {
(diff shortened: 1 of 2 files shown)


armanbilge:
Thanks for the PR and improving the test coverage!

On the topic of tests, I think you can use this method to create a symlink (actually, I don't think that method is tested right now!):
https://github.com/typelevel/fs2/blob/109a71f2a935c3fc476d48dfdfdf00b1ee1c8c38/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala#L85-L89

Note sure how to easily create hidden files cross-OS (i.e., unix and Windows).

yurique:
Added a true-case test for `isSymbolicLink`, which was pretty straight forward.

`isHidden` was more interesting.

First, the regex that was used (`raw"/(^|\/)\.[^\/\.]/g".r`) was no longer working at all. I believe it's due to the changes in Scala.js 1.7.0: https://www.scala-js.org/news/2021/08/04/announcing-scalajs-1.7.0/

> This release fixes a number of bugs. In particular, regular expressions, available through java.util.regex.Pattern or Scala’s Regex and .r method, now behave in the same way as on the JVM. This change has compatibility implications, which we discuss below.

Confirmed this in `ioJS/console` (same in `ioJVM/console`):

The previous regex:
```
scala> val HiddenPattern = raw"/(^|\/)\.[^\/\.]/g".r
val HiddenPattern: scala.util.matching.Regex = /(^|\/)\.[^\/\.]/g

scala> def test(s: String) = HiddenPattern.findFirstIn(s).isDefined
def test(s: String): Boolean

scala> test("/var/folders/13/gsqmg365255bnrlnmn7m6pfr0000gn/T/GlfN5L/.should-be-hidden/not-hidden")
val res0: Boolean = false
```

Removing the `g` modifier:
```
scala> val HiddenPattern = raw"/(^|\/)\.[^\/\.]/".r
val HiddenPattern: scala.util.matching.Regex = /(^|\/)\.[^\/\.]/

scala> def test(s: String) = HiddenPattern.findFirstIn(s).isDefined
def test(s: String): Boolean

scala> test("/var/folders/13/gsqmg365255bnrlnmn7m6pfr0000gn/T/GlfN5L/.should-be-hidden/not-hidden")
val res1: Boolean = false
```

Finally, without `/`s it worked:
```
scala> val HiddenPattern = raw"(^|\/)\.[^\/\.]".r
val HiddenPattern: scala.util.matching.Regex = (^|\/)\.[^\/\.]

scala> def test(s: String) = HiddenPattern.findFirstIn(s).isDefined
def test(s: String): Boolean

scala> test("/var/folders/13/gsqmg365255bnrlnmn7m6pfr0000gn/T/GlfN5L/.should-be-hidden/not-hidden")
val res2: Boolean = true

scala> test("/var/folders/13/gsqmg365255bnrlnmn7m6pfr0000gn/T/GlfN5L/should-be-hidden/not-hidden")
val res3: Boolean = false
```

Made a test for it which checked for `true` "if the path is for a non-hidden file in a hidden directory", and it passed. But it failed in the JVM: as far as I understand, in JVM, only the file itself is checked to be hidden, so if it's not hidden itself, it doesn't matter if it is located somewhere in a hidden directory. So that test is now checking for `false`.

Having that, I removed the regex completely, leaving a simple implementation instead:
```scala
    override def isHidden(path: Path): F[Boolean] = F.pure {
      val fileName = path.fileName.toString
      fileName.length >= 2 && fileName(0) == '.' && fileName(1) != '.'
    }
```

I'm uncomfortable with doing `head`-like things, but doing something like `fileName.length >= 2 && fileName.headOption.contains('.') && !fileName.drop(1).headOption.contains('.')` feels like an overkill (and a bunch of unnecessary allocations).

yurique:
As for testing the `.isHidden` in Windows... I don't even know, I'm hurting from even trying to think about that :) (and we don't have Windows CI)

(also this https://stackoverflow.com/questions/53791740/why-does-files-ishiddenpath-return-false-for-directories-on-windows)

io/shared/src/test/scala/fs2/io/file/FilesSuite.scala
```diff
@@ -555,5 +562,149 @@ class FilesSuite extends Fs2Suite with BaseFileSuite {
...
+
+    test("returns false if the path does not exist") {
+      tempDirectory
+        .map(_.resolve("non-existent-file"))
+        .use(Files[IO].isWritable(_))
+        .assertEquals(false)
+    }
+  }
+
+  group("isHidden") {
```
armanbilge, line 616: Even though there's no windows CI, this will be annoying for anyone developing on Windows. Not sure how big a deal that is.

armanbilge (approved):
I'm really thrilled about these fixes, and overall the improved test coverage for the `Files` API. 🤞 this will fix some other issues downstream as well :) thanks!

io/shared/src/test/scala/fs2/io/file/FilesSuite.scala
yurique: Yeah 🤔 
If someone's working on fs2 in Windows and runs the tests.. that will be annoying (I kind of blanked out such a possibility :/ )
I wonder if we can do some platform checks in the test code and, at least, always pass `isHidden` true-case if it's windows?

yurique: > Eh, it's merged. Let's see if anyone complains :P
 
=)

> The problem is we need cross-platform (JVM/JS) checks for checking the platform (Unix/Windows). Possible, but annoying!

Should be non-crazy enough to implement, right?

Is there a good place for it? (like `Files[F]` is for the file stuff)

armanbilge: Oh, as a public API? Probably not in fs2. I was thinking as an internal/test utility.

Issue #2503, opened 2021-07-23
Fs2-IO Java Resources API


ChristopherDavenport:
Right now there are Files and Network, but another commonly used component is java resources. 

Might be cool to have that integrated so that folks can just reference resources via strings or as streams, etc. I asked in discord and it seemed like there may be interest. So before taking a stab at it I figured it'd make sense to open an issue to track it.

Similar to how StaticFile in http4s does so perhaps? https://github.com/http4s/http4s/blob/7ef1ef1930e5a4dcfb8b73043c3368e1acd8c34a/core/src/main/scala/org/http4s/StaticFile.scala#L34

Thoughts appreciated.

Daenyth:
I have some basic stuff around this, but only rudimentary
https://gist.github.com/Daenyth/67575575b5c1acc1d6ea100aae05b3a9#file-testresource-scala-L10

rossabaker:
Impure.  `URLClassLoader` references the file system and has mutable state, and a custom `ClassLoader` can do even more unholy things.

armanbilge:
Here you go: https://github.com/typelevel/fs2/pull/2733

PR #2733, opened 2021-11-30
Implement `io.readResource`

armanbilge:
Addresses #2503.


io/jvm/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -137,4 +137,15 @@ private[fs2] trait ioplatform {
   def stdinUtf8[F[_]: Sync](bufSize: Int): Stream[F, String] =
     stdin(bufSize).through(text.utf8.decode)
 
+  /** Stream of bytes read asynchronously from the specified classloader resource. */
+  def readResource[F[_]](
+      name: String,
+      chunkSize: Int,
+      classLoader: ClassLoader = getClass().getClassLoader()
+  )(implicit F: Sync[F]): Stream[F, Byte] =
+    Stream.eval(F.delay(Option(classLoader.getResourceAsStream(name)))).flatMap {
```
armanbilge, line 146: Maybe this should be blocking 🤔

mpilquist:
It might be nice to do this in a way which can use `Files` if the resource is a `jar:file` URL. We don't currently allow it because we have no way to construct a `java.nio.file.Path` via a URI.

BTW, we should also note that the resource path is absolute, not relative to a class. E.g., `classOf[Foo].getResource("foo.properties")` vs `classOf[Foo].getClassLoader.getResource("a/b/c/foo.properties")`

armanbilge:
> It might be nice to do this in a way which can use Files if the resource is a jar:file URL.

How would that work exactly? Do you mean use `Files` to implement this method? Or do you mean add a method to the `Files` API which works with a `URL` instead of `Path`?

For the former, I'm not sure if that's possible, since the file is embedded in the jar right? Seems like [`JavaURLConnection`](https://docs.oracle.com/javase/8/docs/api/java/net/JarURLConnection.html) would be how to interact with those.

mpilquist:
Yeah okay, I forgot that there's no default `jar:` file system handler. I wanted a way to do this without going through `InputStream` but I guess that's not possible.

armanbilge:
Ohh, no I think I got you. Maybe this is possible, using [`FileSystems.newFileSystem`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystems.html)? Would this actually make a difference?

mpilquist:
Right, we'd have to call `newFileSystem` with a path to the JAR and then create a `java.nio.file.Path` from that file system for the resource in the JAR. And we'd likely still need a fallback to current implementation. Probably best to stick with what you've got -- the original fs2 `InputStream` integration seemed to have high overhead but I'm probably worried about an implementation that no longer exists.

io/jvm/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -137,4 +138,29 @@ private[fs2] trait ioplatform {
   def stdinUtf8[F[_]: Sync](bufSize: Int): Stream[F, String] =
     stdin(bufSize).through(text.utf8.decode)
 
+  /** Stream of bytes read asynchronously from the specified resource relative to the class `C`.
+    * @see [[readClassLoaderResource]] for a resource relative to a classloader.
+    */
+  def readClassResource[F[_], C](
+      name: String,
+      chunkSize: Int
+  )(implicit F: Sync[F], ct: ClassTag[C]): Stream[F, Byte] =
```
armanbilge, line 147: H/t @Daenyth for the classtag trick.

io/jvm/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -137,4 +138,29 @@ private[fs2] trait ioplatform {
...
+      case Some(resource) => io.readInputStream(resource.pure, chunkSize)
+      case None           => Stream.raiseError(new IOException(s"Resource $name not found"))
+    }
+
+  /** Stream of bytes read asynchronously from the specified classloader resource.
+    * @see [[readClassResource]] for a resource relative to a class.
+    */
+  def readClassLoaderResource[F[_]](
+      name: String,
+      chunkSize: Int,
```
mpilquist, line 158: Default chunk size here and above to 64K like in Files?

armanbilge:
```suggestion
      chunkSize: Int = 64 * 1024,
```

io/jvm/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -137,4 +138,29 @@ private[fs2] trait ioplatform {
   def stdinUtf8[F[_]: Sync](bufSize: Int): Stream[F, String] =
     stdin(bufSize).through(text.utf8.decode)
 
+  /** Stream of bytes read asynchronously from the specified resource relative to the class `C`.
+    * @see [[readClassLoaderResource]] for a resource relative to a classloader.
+    */
+  def readClassResource[F[_], C](
+      name: String,
+      chunkSize: Int
```
armanbilge, line 146:
```suggestion
      chunkSize: Int = 64 * 1024
```

Issue #2709, opened 2021-11-14
[fs2.io.net] `read` after a cancelled `read` causes `java.nio.channels.ReadPendingException`


kory33:
## Issue description
Given a TCP client `client: Socket[cats.effect.IO]`, a `client.read` immediately after a cancelled `client.read` (sometimes) causes `java.nio.channels.ReadPendingException`.

## Reproduction
On Windows 10, OpenJDK 17, prepare the following files:

<details>
  <summary>build.sbt</summary>

```sbt
scalaVersion := "3.1.0"
libraryDependencies ++= Seq(
  "co.fs2" %% "fs2-io" % "3.2.2"
)
```

</details>

<details>
  <summary>src/main/scala/Test.scala (binds port 31629, change portNumber if needed)</summary>

```Scala
object Test {
  @main def main: Unit = {
    import cats.Monad
    import cats.effect.{IO, Resource, Spawn}
    import com.comcast.ip4s.SocketAddress
    import fs2.io.net.Network

    import cats.implicits.given
    import cats.effect.unsafe.implicits.global

    val portNumber = 31629
    val serverAddress = SocketAddress.fromString(s"localhost:$portNumber").get
    val clientResource: Resource[IO, fs2.io.net.Socket[IO]] =
      for {
        // we prepare a tcp server which returns 0 for every read from a client
        _ <- Spawn[IO].background {
          Network[IO]
            .server(address = Some(serverAddress.host), port = Some(serverAddress.port))
            .evalMap(socket =>
              Monad[IO].whileM_(socket.isOpen)(socket.write(fs2.Chunk[Byte](0)))
            )
            .compile
            .drain
        }
        socket <- Network[IO].client(serverAddress)
      } yield socket

    val program: IO[Unit] =
      clientResource.use { client =>
        Monad[IO].foreverM {
          client.read(1024).start.flatMap { fiber =>
            fiber.cancel >> client.read(1024) >> fiber.join
          }
        }
      }.void

    program.unsafeRunSync()
  }
}
```

</details>

And `runMain main` in `sbt` console.

<details>
  <summary>example log</summary>

```
[IJ]runMain main
[info] running main
[error] (run-main-1) java.nio.channels.ReadPendingException
[error] java.nio.channels.ReadPendingException
[error]         at java.base/sun.nio.ch.AsynchronousSocketChannelImpl.read(AsynchronousSocketChannelImpl.java:251)
[...]

PR #2734, opened 2021-12-01
Change Socket#read to disallow a read after a timeout

mpilquist:
Fixes https://github.com/typelevel/fs2/issues/2709

We explored four solutions here:
1) Making reads uncancellable
2) Allow read to be canceled and then any subsequent reads hang
3) Improving the error message when reading from a socket that's had a canceled read
4) Making reads after a canceled read behave like a closed socket

## Making reads uncancellable

The JVM's async channel read does not support cancelation. Hence, the simplest approach is wrapping the channel read with an `F.uncancellable`. We modified `readChunk` and wrapped the `F.async_` call with `F.uncancellable(_ => F.async_(...))`. Unfortunately, this resulted in stream hangs any time a read was pending at the time of shutdown. For example:

```scala
val server = Network[IO].server(...)
doStuff.concurrently(server)
```

When `doStuff` completes, the `server` stream is interrupted. However, if there are any accepted sockets which are waiting on a `read`, shutdown hangs until those reads complete -- either due to receiving data or due to the socket getting closed somehow.

Hence, we abandoned this solution.

## Allow read to be canceled and then any subsequent reads hang

The next approach allowed cancelation of read but made any subsequent reads hang (i.e. never complete). This also intentionally leaks the channel read upon cancelation - at most one of these pending reads per socket though, so no memory / resource concerns. This was implemented in ea596c2 by modifying the `readSemaphore` management such that the semaphore was not released in the event of cancelation. Hence any subsequent reads would wait forever for the `readSemaphore`. This worked but seemed confusing from an API perspective.

## Improving the error message when reading from a socket that's had a canceled read

We tried simply improving the error message in the case where a read is performed after a canceled read. The downside of this approach is that the leaked read could eventually succeed and then the next user land read would subsequently succeed.

## Making reads after a canceled read behave like a closed socket

We settled on making a canceled read shutdown the socket for future channel reads. Any user land calls to `read` or `readN` return the same result as if the socket was manually closed for reading (via `endOfInput`).

Note: the Node.js implementation doesn't suffer from this same limitation.

### DIFF
--- a/io/jvm/src/main/scala/fs2/io/net/SocketPlatform.scala
+++ b/io/jvm/src/main/scala/fs2/io/net/SocketPlatform.scala
@@ -24,7 +24,7 @@ package io
 package net
 
 import com.comcast.ip4s.{IpAddress, SocketAddress}
-import cats.effect.{Async, Poll, Resource}
+import cats.effect.{Async, Resource}
 import cats.effect.std.Semaphore
 import cats.syntax.all._
 
@@ -49,16 +49,8 @@ private[net] trait SocketCompanionPlatform {
     private[this] final val defaultReadSize = 8192
     private[this] var readBuffer: ByteBuffer = ByteBuffer.allocateDirect(defaultReadSize)
 
-    private def withReadPermit[A](f: F[A]): F[A] =
-      Resource
-        .makeCaseFull((poll: Poll[F]) => poll(readSemaphore.acquire)) {
-          case (_, Resource.ExitCase.Canceled) => F.unit
-          case (_, _)                          => readSemaphore.release
-        }
-        .use(_ => f)
-
     private def withReadBuffer[A](size: Int)(f: ByteBuffer => F[A]): F[A] =
-      withReadPermit {
+      readSemaphore.permit.use { _ =>
         F.delay {
           if (readBuffer.capacity() < size)
             readBuffer = ByteBuffer.allocateDirect(size)
@@ -66,6 +58,17 @@ private[net] trait SocketCompanionPlatform {
             (readBuffer: Buffer).limit(size)
           f(readBuffer)
         }.flatten
+          .adaptError { case _: java.nio.channels.ReadPendingException =>
+            new IllegalStateException(
+              """|Illegal socket read.
+                                         |
+                                         |A previous read on this socket was cancelled. This is typically due to using
+                                         |timeout or another operation which cancels the fiber performing the read.
+                                         |Once a socket read has been cancelled, the underlying socket is left in an
+                                         |indeterminate state and no further reads may be performed. Consider opening
+                                         |a new socket or otherwise avoiding the cancellation.""".stripMargin
+            )
+          }
       }
 
     /** Performs a single channel read operation in to the supplied buffer. */
--- a/io/jvm/src/main/scala/fs2/io/net/SocketPlatform.scala
+++ b/io/jvm/src/main/scala/fs2/io/net/SocketPlatform.scala
@@ -54,7 +54,7 @@ private[net] trait SocketCompanionPlatform {
       F.uncancelable { p =>
         p(readSemaphore.acquire) *>
           p(fa)
-            .onCancel(F.delay(this.readBuffer = null))
+            .onCancel(F.delay(this.readBuffer = null) *> endOfInput)
             .guarantee(readSemaphore.release)
       }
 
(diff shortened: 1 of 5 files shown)


io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala
```diff
@@ -216,5 +217,47 @@ class SocketSuite extends Fs2Suite with SocketSuitePlatform {
...
+          val prg =
+            client.write(msg) *>
+              client.readN(msg.size) *>
+              client.readN(msg.size).timeout(100.millis).recover { case _: TimeoutException =>
+                Chunk.empty
+              } *>
+              client.write(msg) *>
+              client
+                .readN(msg.size)
+                .timeout(1000.millis)
```
armanbilge, line 241: Why is another timeout needed here? For Node.js?

mpilquist: Oh that can come out now. That was originally needed when we were hanging the 2nd+ request.

armanbilge: No, Node.js succeeds below.

armanbilge:
```suggestion
```

armanbilge (commented):
So actually I think there is a subtlety to this. If between the "cancelled" read and the subsequent read the underlying callback completes, then we won't actually observe this exception, right? Does that matter? Should we consider setting a flag on cancellation?

mpilquist:
Yeah that's a good catch. I suppose a private var flag would be sufficient for that, under the readSemaphore.

armanbilge:
Right, it can be published by the semaphore without additional synchronization 👍

io/jvm/src/main/scala/fs2/io/net/SocketPlatform.scala
```diff
@@ -179,11 +177,17 @@ private[net] trait SocketCompanionPlatform {
...
-  private final class IntCallbackHandler[A](cb: Either[Throwable, Int] => Unit)
+  private final class ChannelBytesHandler[A](cb: Either[Throwable, Int] => Unit)
       extends CompletionHandler[Integer, AnyRef] {
     def completed(result: Integer, attachment: AnyRef): Unit =
       cb(Right(result))
     def failed(err: Throwable, attachment: AnyRef): Unit =
-      cb(Left(err))
+      err match {
+        case _: IOException =>
+          // Connection may be reset while reading/writing; if so, indicate failure via negative bytes instead of propagating exception
```
armanbilge, line 187: Should we try and check for a specific message on the `IOException`?

io/jvm/src/main/scala/fs2/io/net/SocketPlatform.scala
```diff
@@ -112,23 +112,47 @@ private[net] trait SocketCompanionPlatform {
...
+          new CompletionHandler[Integer, AnyRef] {
+            def completed(bytesRead: Integer, attachment: AnyRef) =
+              cb(Right(bytesRead))
+            def failed(err: Throwable, attachment: AnyRef) =
+              if (inputOpen) cb(Left(err))
+          }
         )
+        F.delay(Some(F.delay {
+          inputOpen = false
+          ch.shutdownInput()
```
armanbilge, line 133: Sorry, I got confused why we need these open flags. By the time we make it to this block where we do `ch.shutdownInput()` the `cb` function is already a no-op right?

mpilquist: Fixed...

io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala
```diff
@@ -216,5 +217,51 @@ class SocketSuite extends Fs2Suite with SocketSuitePlatform {
...
+              client.readN(msg.size).timeout(100.millis).recover { case _: TimeoutException =>
+                Chunk.empty
+              } *>
+              client.write(msg) *>
+              client
+                .readN(msg.size)
+                .flatMap { c =>
+                  if (isJVM) {
+                    assertEquals(c.size, 0)
+                    // Read again now that the pending read is no longer pending
```
armanbilge, line 244: I think this comment is no longer relevant since the channel is dead.

armanbilge (approved):
Phew 😓

io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala
mpilquist: Still accurate -- it's testing that a subsequent `readN` returns 0 size chunk instead of actually performing a read or otherwise returning previously read data.

armanbilge: Ah, gotcha.

PR #2735, opened 2021-12-01
Change filterWithPrevious example list in Scaladoc

valencik:
This PR tweaks the documentation example for `filterWithPrevious` to hopefully better highlight that the `previous` value is the last *emitted* value.

The prior example would have returned a similar answer even if the behaviour was to compare one element with its previous element in the original input Stream.

Co-authored-by: @samspills


mpilquist:
Thanks!

PR #2740, opened 2021-12-04
fs2-protocols - pcapng

nikiforo:


mpilquist:
Looking good!

Some high level comments:
- Is there a way to abstract over `pcap` and `pcapng` if you just want to decode captures?
- Note use of Shapeless will mean those files are Scala 2 only and will need reimplementations for Scala 3 using tuples, match types, etc.

nikiforo:
> * Is there a way to abstract over `pcap` and `pcapng` if you just want to decode captures?

I think there is. However, I think this functionality should be not in the `main` but in the `test` folder. Both `pcap` & `pcapng` are in the same subproject and share `test` folder. 
> * Note use of Shapeless will mean those files are Scala 2 only and will need reimplementations for Scala 3 using tuples, match types, etc.

What are the options assuming that I want to define common block structure?  I think I can put HList-based implementation to `scala-2` folder, tuple-based to `scala-3`?

nikiforo:
> * Note use of Shapeless will mean those files are Scala 2 only and will need reimplementations for Scala 3 using tuples, match types, etc.

I think I should look at `scodec`. It should have been ported to scala-3, thus it already abstracts over shapeless 2/3.

mpilquist:
Yeah, scodec uses HLists on Scala 2 and tuples on Scala 3. There's a little bit of trickery to get the same source cross compiling on both: https://github.com/scodec/scodec/blob/7724ae7efed0ea9e32063b56cf81bd3ffeb381f0/shared/src/main/scala/scodec/compat.scala

mpilquist:
@nikiforo Any way I can help get this one over the finish line?

PR #2741, opened 2021-12-04
Implement `TLSSocket#applicationProtocol` on JS

armanbilge:


ChristopherDavenport (commented):
:+1:  This looks good to me and gets js into parity. Might want to note that `null` from before will now end up as `NoSuchElement` exceptions somewhere. I already moved to handle this, but its possible someone else was coding against this and should know for an upgrade.

PR #2756, opened 2021-12-17
Files walk order

SystemFw:
Changes the behaviour of `walk` to emit files in the order it visits them, which in turn enables streaming copy since it lets you create folders before copying the files they contain, which is a prerequisite for `copy`.
The behaviour of `walk` in main is arguably a regression since it behaves differently from `NIO.walk`, which it used to wrap, so I'm targetting `main` with a straight change, rather than a new method or a flag.

### DIFF
--- a/io/shared/src/main/scala/fs2/io/file/Files.scala
+++ b/io/shared/src/main/scala/fs2/io/file/Files.scala
@@ -455,30 +455,30 @@ object Files extends FilesCompanionPlatform {
                   .recoverWith { case _ =>
                     Stream.empty
                   }
-                  else if (attr.isSymbolicLink && followLinks)
-                    Stream.eval(getBasicFileAttributes(start, followLinks = true)).flatMap { attr =>
-                      val fileKey = attr.fileKey
-                      val isCycle = Traverse[List].existsM(ancestry) {
-                        case Right(ancestorKey) => F.pure(fileKey.contains(ancestorKey))
-                        case Left(ancestorPath) => isSameFile(start, ancestorPath)
-                      }
+              else if (attr.isSymbolicLink && followLinks)
+                Stream.eval(getBasicFileAttributes(start, followLinks = true)).flatMap { attr =>
+                  val fileKey = attr.fileKey
+                  val isCycle = Traverse[List].existsM(ancestry) {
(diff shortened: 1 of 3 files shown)


SystemFw:
It shows as formatted on my machine...

io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -443,43 +443,44 @@ object Files extends FilesCompanionPlatform {
...
+        Stream.emit(start) ++ {
+          if (maxDepth == 0) Stream.empty
+          else
+            Stream.eval(getBasicFileAttributes(start, followLinks = false)).flatMap { attr =>
+              if (attr.isDirectory)
+                list(start)
+                  .flatMap { path =>
+                    go(path, maxDepth - 1, attr.fileKey.toRight(start) :: ancestry)
+                  }
+                  .recoverWith { case _ =>
```
nikiforo, line 455: `mask`?

io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -443,43 +443,44 @@ object Files extends FilesCompanionPlatform {
...
-                   else
-                     Stream.raiseError(new FileSystemLoopException(start.toString))
-                 }
-
-               }
-             else
-               Stream.empty) ++ Stream.emit(start)
-          }
+        Stream.emit(start) ++ {
+          if (maxDepth == 0) Stream.empty
```
nikiforo, line 447: hasn't the behavior changed here?

io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -443,43 +443,44 @@ object Files extends FilesCompanionPlatform {
...
+                    case Left(ancestorPath) => isSameFile(start, ancestorPath)
+                  }
+
+                  Stream.eval(isCycle).flatMap { isCycle =>
+                    if (!isCycle)
+                      list(start)
+                        .flatMap { path =>
+                          go(path, maxDepth - 1, attr.fileKey.toRight(start) :: ancestry)
+                        }
+                        .recoverWith { case _ =>
```
nikiforo, line 472: `mask`?

PR #2751, opened 2021-12-10
Channel perf

nikiforo:
`benchmark/ jmh:run -i 10 -wi 6 -f1 -t4 -gc true -jvmArgs "-XX:MaxRecursiveInlineLevel=3 -XX:MaxInlineSize=50 -Dcats.effect.tracing.mode=none" fs2.benchmark.ChannelBenchmark`

`channel-perf`
```
[info] Benchmark                              (size)   Mode  Cnt      Score     Error  Units
[info] ChannelBenchmark.sendPull                  64  thrpt   10  70315,735 ± 990,801  ops/s
[info] ChannelBenchmark.sendPull                1024  thrpt   10   8489,787 ± 220,705  ops/s
[info] ChannelBenchmark.sendPull               16384  thrpt   10    586,689 ±  14,841  ops/s
[info] ChannelBenchmark.sendPullPar8              64  thrpt   10  48603,532 ± 755,156  ops/s
[info] ChannelBenchmark.sendPullPar8            1024  thrpt   10  10497,909 ± 444,052  ops/s
[info] ChannelBenchmark.sendPullPar8           16384  thrpt   10    650,486 ±  36,143  ops/s
[info] ChannelBenchmark.sendPullParUnlimited      64  thrpt   10  46336,865 ± 640,137  ops/s
[info] ChannelBenchmark.sendPullParUnlimited    1024  thrpt   10   3421,420 ± 268,378  ops/s
[info] ChannelBenchmark.sendPullParUnlimited   16384  thrpt   10    162,638 ±   9,527  ops/s
```

`main`
```
[info] Benchmark                              (size)   Mode  Cnt      Score      Error  Units
[info] ChannelBenchmark.sendPull                  64  thrpt   10  65333,103 ± 4694,045  ops/s
[info] ChannelBenchmark.sendPull                1024  thrpt   10   7288,803 ±  309,856  ops/s
[info] ChannelBenchmark.sendPull               16384  thrpt   10    487,553 ±   28,040  ops/s
[info] ChannelBenchmark.sendPullPar8              64  thrpt   10  45527,449 ± 2153,177  ops/s
[info] ChannelBenchmark.sendPullPar8            1024  thrpt   10   8268,452 ±  320,778  ops/s
[info] ChannelBenchmark.sendPullPar8           16384  thrpt   10    490,139 ±   25,620  ops/s
[info] ChannelBenchmark.sendPullParUnlimited      64  thrpt   10  36154,018 ± 2300,873  ops/s
[...]

PR #2758, opened 2021-12-17
`text.lines` enhancements

stephenjudkins:
* add support for input with '\r'-only newlines, like from macos classic. yes, content like this still exists
* add support to throw an error when an accumulated line is over a certain size; good to prevent bad/malicious inputs from causing OOMs

### DIFF
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -317,16 +317,22 @@ object text {
     utf8.encodeC
 
 
-  def linesFor[F[_]: RaiseThrowable](sizeLimit: Option[Int] = None, crsOnly: Boolean = false):Pipe[F, String, String] =
+  /**
+    * Transforms a stream of `String` such that each emitted `String` is a line from the input
+    * @param maxLineLength maximum size to accumulate a line to; throw an error if a line is larger
+    * @param crsOnly separate lines that are delimited only by '\r'
+    * @tparam F
+    */
+  def linesFor[F[_]: RaiseThrowable](maxLineLength: Option[Int] = None, crsOnly: Boolean = false):Pipe[F, String, String] =
     linesImpl[F](
-      maxLineSize = sizeLimit.map((_, implicitly[RaiseThrowable[F]])),
+      maxLineLength = maxLineLength.map((_, implicitly[RaiseThrowable[F]])),
       crsOnly = crsOnly
     )
 
+  /** Transforms a stream of `String` such that each emitted `String` is a line from the input. */
   def lines[F[_]]:Pipe[F, String, String] = linesImpl[F]()
 
-  /** Transforms a stream of `String` such that each emitted `String` is a line from the input. */
-  private def linesImpl[F[_]](maxLineSize: Option[(Int, RaiseThrowable[F])] = None, crsOnly: Boolean = false): Pipe[F, String, String] = {
+  private def linesImpl[F[_]](maxLineLength: Option[(Int, RaiseThrowable[F])] = None, crsOnly: Boolean = false): Pipe[F, String, String] = {
     def fillBuffers(
         stringBuilder: StringBuilder,
         linesBuffer: ArrayBuffer[String],
@@ -374,7 +380,7 @@ object text {
             fillBuffers(stringBuilder, linesBuffer, string)
           }
 
-          maxLineSize match {
+          maxLineLength match {
             case Some((max, raiseThrowable)) if stringBuilder.length() > max =>
               Pull.raiseError[F](new IllegalStateException(
                 s"Max line size is $max but ${stringBuilder.length()} chars have been accumulated"
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -387,7 +387,7 @@ object text {
           maxLineLength match {
             case Some((max, raiseThrowable)) if stringBuilder.length > max =>
               Pull.raiseError[F](
-                new LineTooLongException(stringBuilder.length(), max)
+                new LineTooLongException(stringBuilder.length, max)
               )(raiseThrowable)
             case _ =>
               Pull.output(Chunk.indexedSeq(linesBuffer)) >> go(stream, stringBuilder, first = false)
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -387,9 +387,7 @@ object text {
           maxLineLength match {
             case Some((max, raiseThrowable)) if stringBuilder.length > max =>
               Pull.raiseError[F](
-                new IllegalStateException(
-                  s"Max line size is $max but ${stringBuilder.length} chars have been accumulated"
-                )
+                new LineTooLongException(stringBuilder.length(), max)
               )(raiseThrowable)
             case _ =>
               Pull.output(Chunk.buffer(linesBuffer)) >> go(stream, stringBuilder, first = false)
@@ -398,6 +396,11 @@ object text {
 
     s => Stream.suspend(go(s, new StringBuilder(), first = true).stream)
   }
+
+  class LineTooLongException(val length: Int, val max: Int)
+      extends RuntimeException(
+        s"Max line size is $max but $length chars have been accumulated"
+      )
 
   /** Functions for working with base 64. */
   object base64 {
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -385,10 +385,10 @@ object text {
           }
 
           maxLineLength match {
-            case Some((max, raiseThrowable)) if stringBuilder.length() > max =>
+            case Some((max, raiseThrowable)) if stringBuilder.length > max =>
               Pull.raiseError[F](
                 new IllegalStateException(
-                  s"Max line size is $max but ${stringBuilder.length()} chars have been accumulated"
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/text.scala
```diff
@@ -360,7 +383,17 @@ object text {
           chunk.foreach { string =>
             fillBuffers(stringBuilder, linesBuffer, string)
           }
-          Pull.output(Chunk.buffer(linesBuffer)) >> go(stream, stringBuilder, first = false)
+
+          maxLineLength match {
+            case Some((max, raiseThrowable)) if stringBuilder.length > max =>
+              Pull.raiseError[F](
+                new IllegalStateException(
```
stephenjudkins, line 390: Can anyone recommend any other type of exception here instead of IllegalStateException?

rossabaker: I don't think it's an `IllegalStateException`.  We didn't call it at an inopportune time: we called it with an input that didn't match the configuration.  I think `IllegalArgumentException` is the closest, but I might just give it its own type of `RuntimeException`.

stephenjudkins: Yeah that seems right to me. I've created a new `LineTooLongException`

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -360,12 +383,25 @@ object text {
           chunk.foreach { string =>
             fillBuffers(stringBuilder, linesBuffer, string)
           }
-          Pull.output(Chunk.buffer(linesBuffer)) >> go(stream, stringBuilder, first = false)
+
+          maxLineLength match {
```
nikiforo, line 387: There is `LinesBenchmark`, can you, please, run it against `main` version and this one?

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -316,14 +316,34 @@ object text {
...
+      crsOnly = crsOnly
+    )
+
   /** Transforms a stream of `String` such that each emitted `String` is a line from the input. */
-  def lines[F[_]]: Pipe[F, String, String] = {
+  def lines[F[_]]: Pipe[F, String, String] = linesImpl[F]()
+
+  private def linesImpl[F[_]](
+      maxLineLength: Option[(Int, RaiseThrowable[F])] = None,
+      crsOnly: Boolean = false
```
nikiforo, line 338: I see, that `\r` can also be a line delimiter.
So, should it be a regression? i.e. should the default behavior change from treating `\n` & `\r\n` as a line delimiter to treating `\r`, `\r\n`, `\n` as a line delimiter? I would expect `lines` to handle this kind of ambiguity not with a flag, but always. In that case we should check that `\r\n` doesn’t produce two lines.

stephenjudkins: My assumption is that other users wouldn't want the existing behavior to change. If I were the only consumer of this library I'd just have it always treat bare `\r`s as line separators.

I don't know about the consensus-forming process is here. But I'd be happy to go with the flow and do whatever other stakeholders agree on here.

stephenjudkins:
FYI: Found a couple corner cases that fail, added texts, working on fix now

stephenjudkins:
So I've thought about this a bit more and actually think having the `lines` method split on `\r`s in all cases is actually the right path forward.

Arguments for:
* `\r` is, in fact, a valid newline character and there is real data out there that uses it for whatever reason
* Having text with bare `\r`s and also `\r\n` or `\n` and only splitting on the latter two is a pretty obscure corner case and anyone depending on having lines with `\r` in them is relying on some undocumented behavior
* For `\n` and `\r\n`-delimited text we can add this support with very little performance impact, since we're rarely going to hit any of the new conditions
* `linesFor` or `linesFancy` are weird names, but I can't think of better ones. It's annoying that `linesFor[F]` requires `RaiseThrowable[F]` even when we're not using `maxLineLength`, only `onlyCrs`. Adding multiple methods also gets weird, since in our case we want both functionalities at the same time, and I don't think any type system circus tricks for these methods would be welcome here. `onlyCrs` is also a weird and confusing parameter name, but again, I can't think of a much better one. If the other method were simply `linesLimit[F: RaiseThrowable]` things really would seem much cleaner.

Arguments against:
* macOS/BSD `head -n` and other utilities doesn't respect bare `\r` newlines, so it's obscure
* Don't change existing behavior

I'm prepared to be swayed either way. Regardless, the two pieces of functionality are very useful for us in real-world contexts (especially the max line length!) so I'd like to see them included. Let me know your thoughts

mpilquist:
@stephenjudkins Thanks! I agree with changing the existing default behavior to cover \r as well as \n and \r\n. I like `linesLimit` or `linesLimited`.

stephenjudkins:
Latest benchmarks: 

On `main`:
```
sbt:root> benchmark/Jmh/run fs2.benchmark.LinesBenchmark -i 1 -wi 1 -f 1 -t 1
...
[info] Benchmark                      (asciiLineSize)  (chunkSize)   Mode  Cnt       Score   Error  Units
[info] LinesBenchmark.linesBenchmark                0            4  thrpt       129367.374          ops/s
[info] LinesBenchmark.linesBenchmark                0           16  thrpt       127115.654          ops/s
[info] LinesBenchmark.linesBenchmark                0           64  thrpt       121482.553          ops/s
[info] LinesBenchmark.linesBenchmark                1            4  thrpt        27572.723          ops/s
[info] LinesBenchmark.linesBenchmark                1           16  thrpt        61276.567          ops/s
[info] LinesBenchmark.linesBenchmark                1           64  thrpt        68777.173          ops/s
[info] LinesBenchmark.linesBenchmark               10            4  thrpt         8109.786          ops/s
[info] LinesBenchmark.linesBenchmark               10           16  thrpt        39059.947          ops/s
[info] LinesBenchmark.linesBenchmark               10           64  thrpt        56341.091          ops/s
[info] LinesBenchmark.linesBenchmark              100            4  thrpt         1232.612          ops/s
[info] LinesBenchmark.linesBenchmark              100           16  thrpt         7520.345          ops/s
[info] LinesBenchmark.linesBenchmark              100           64  thrpt        12763.142          ops/s
```

On `lines-enhancements`:
```
sbt:root> benchmark/Jmh/run fs2.benchmark.LinesBenchmark -i 1 -wi 1 -f 1 -t 1
...
[info] Benchmark                      (asciiLineSize)  (chunkSize)   Mode  Cnt       Score   Error  Units
[info] LinesBenchmark.linesBenchmark                0            4  thrpt       129716.313          ops/s
[info] LinesBenchmark.linesBenchmark                0           16  thrpt       133247.824          ops/s
[info] LinesBenchmark.linesBenchmark                0           64  thrpt       132027.374          ops/s
[info] LinesBenchmark.linesBenchmark                1            4  thrpt        29923.197          ops/s
[info] LinesBenchmark.linesBenchmark                1           16  thrpt        49368.691          ops/s
[info] LinesBenchmark.linesBenchmark                1           64  thrpt        77137.007          ops/s
[info] LinesBenchmark.linesBenchmark               10            4  thrpt         8076.095          ops/s
[info] LinesBenchmark.linesBenchmark               10           16  thrpt        38300.109          ops/s
[info] LinesBenchmark.linesBenchmark               10           64  thrpt        54354.817          ops/s
[info] LinesBenchmark.linesBenchmark              100            4  thrpt         1212.297          ops/s
[info] LinesBenchmark.linesBenchmark              100           16  thrpt         7655.271          ops/s
[info] LinesBenchmark.linesBenchmark              100           64  thrpt        12687.502          ops/s
```

nikiforo:
@stephenjudkins 

I'm not sure that EOF is handled correctly. Take a look at 
```scala
def check3 = {
    val bytes1 = "a\r".getBytes()
    val bytes2 = "a\n".getBytes()
    val lines1 = Stream.emits(bytes1).through(text.utf8.decode).through(text.lines).compile.toList
    val lines2 = Stream.emits(bytes2).through(text.utf8.decode).through(text.lines).compile.toList
    println(lines1.length)
    println(lines2.length)
  }
```

it prints 
```
1
2
```

nikiforo:
A small addition to the command:
> sbt:root> benchmark/Jmh/run fs2.benchmark.LinesBenchmark -i 1 -wi 1 -f 1 -t 1

The more warmups and iterations, the more accurate results are. For instance, I had to increase them from 6 and 10 to 10 and 20 [here](https://github.com/typelevel/fs2/pull/2751#issuecomment-994037977) to reduce errors. Also, 25 iterations were made [here](https://github.com/typelevel/cats-effect/issues/2634#issuecomment-988800286). When you have only 1 iteration, the error isn't printed in the output.

stephenjudkins:
@nikiforo well, that's unfortunate. fix here https://github.com/typelevel/fs2/pull/2764

stephenjudkins:
> A small addition to the command:
> 
> > sbt:root> benchmark/Jmh/run fs2.benchmark.LinesBenchmark -i 1 -wi 1 -f 1 -t 1
> 
> The more warmups and iterations, the more accurate results are. For instance, I had to increase them from 6 and 10 to 10 and 20 [here](https://github.com/typelevel/fs2/pull/2751#issuecomment-994037977) to reduce errors. Also, 25 iterations were made [here](https://github.com/typelevel/cats-effect/issues/2634#issuecomment-988800286). When you have only 1 iteration, the error isn't printed in the output.

Would love to see some more documentation about benchmarking tradeoffs here. I don't know enough about the specific context to be helpful

nikiforo:
> Arguments for:
```scala
"abc\rdef".lines().collect(java.util.stream.Collectors.toList()).asScala // == `Buffer(abc, def)`
```

PR #2759, opened 2021-12-19
fsIO - realPath:  re-raise ENOENT javascript exceptions as NoSuchFileException

yurique:
calls to `realPath` in nodejs throw an `ENOENT` exception if the path doesn't exist. In order to align the behaviour of `realPath` with what we have in JVM, those errors are re-raised as a `NoSuchFileException`.

### DIFF
--- a/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
@@ -321,7 +321,10 @@ private[fs2] trait FilesCompanionPlatform {
       readStream(path, chunkSize, Flags.Read)(_.setStart(start.toDouble).setEnd((end - 1).toDouble))
 
     def realPath(path: Path): F[Path] =
-      F.fromPromise(F.delay(fsPromisesMod.realpath(path.toString))).map(Path(_))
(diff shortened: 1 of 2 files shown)


io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
```diff
@@ -321,7 +321,10 @@ private[fs2] trait FilesCompanionPlatform {
       readStream(path, chunkSize, Flags.Read)(_.setStart(start.toDouble).setEnd((end - 1).toDouble))
 
     def realPath(path: Path): F[Path] =
-      F.fromPromise(F.delay(fsPromisesMod.realpath(path.toString))).map(Path(_))
+      F.fromPromise(F.delay(fsPromisesMod.realpath(path.toString))).map(Path(_)).recoverWith {
+        case NoSuchFileException(e) =>
+          F.raiseError(e)
+      }
```
armanbilge, line 327: I _think_ this is equivalent to:
```scala
.adaptError { case NoSuchFileException(ex) => ex }
```

yurique: Aha! Nice! I've completely forgotten about this one :)

yurique: Adapted it :)

armanbilge (approved):
Thanks!

PR #2763, opened 2021-12-22
Fix #2762 - eval after zip may cause interruption

### LINKED ISSUE #2717: Unexpected behaviors around the use of Pull
I'm not even sure how to describe the problem. Please feel free to change the title to something more appropriate.

```scala
  "org.typelevel" %% "cats-effect" % "3.2.9",
  "co.fs2" %% "fs2-core" % "3.2.2",
```

```scala
import cats.effect.{IO, IOApp}
import fs2.Pull.StreamPullOps
import fs2.Stream

object DebugFs2Issue extends IOApp.Simple {

  val run: IO[Unit] = for {
    _ <- problematicStream
      .flatMap(ev => Stream.eval(IO.delay { println(s"Printing: $ev") }))
      .map(s => s) // <- THIS MAP
      .compile
      .drain
  } yield ()

  def problematicStream[B, O]: Stream[IO, (String, Int)] = {
    val stream = Stream
      .emit(1)
      .map(Left(_)) ++ Stream.emit(Right("emitted")).covary[IO]
    stream.pull.uncons1.flatMap { case Some((Left(_), s)) =>
      s.collect { case Right(b) => b }.through(failurePipe).pull.echo
    }.stream
  }

  def failurePipe: Stream[IO, String] => Stream[IO, (String, Int)] =
    _.zipWith(Stream.emit(2))((_, _))

  def successPipe: Stream[IO, String] => Stream[IO, (String, Int)] =
    _.map(v => (v, 2))
}
```

1) If you run things as above, nothing will be printed.
2) From 1), if you remove the line marked as `<- THIS MAP`, "Printing: (emitted,2)" is printed. This is unexpected.
3) From 1), if you use `successPipe` instead of `failurePipe`, "Printing: (emitted,2)" is also printed. This is also unexpected, since I would predict that `_.zipWith(Stream.emit(2))((_, _))` is equivalent to `_.map(v => (v, 2))`.
4) From 1), if you replace `.emit(1).map(Left(_))` with `.emit(Left(1))`, "Printing: (emitted,2)" is printed. I would expect those two commands to be equivalent.

### LINKED ISSUE #2762: Unexpected behaviors around the use of Pull (part II)
```scala
  "org.typelevel" %% "cats-effect" % "3.2.9",
  "co.fs2" %% "fs2-core" % "3.2.3",
```

Similar to https://github.com/typelevel/fs2/issues/2717.

```scala
  def run: IO[Unit] = {
    val stream = Stream.emit(1).map(Left(_)) ++ Stream(Right("a"))
    val streamModified: Stream[IO, (String, Int)] = stream.pull.uncons1
      .flatMap { case Some((Left(_), s)) =>
        s.collect { case Right(b) => b }.through(pipeFailure).pull.echo
      }
      .void
      .stream
    streamModified
      .evalTap(_ => IO.unit)
      .debug()
      .compile
      .drain
  }

  private def pipeFailure: Stream[IO, String] => Stream[IO, (String, Int)] =
    _.zipWith(Stream.emit(0).repeat)((_, _))

  private def pipeSuccess: Stream[IO, String] => Stream[IO, (String, Int)] =
    _.map(v => (v, 0))
```

I would expect `pipeFailure` and `pipeSuccess` to be equivalent, but they don't seem to be in this scenario?

1a) If I use `pipeFailure`, nothing is logged.
1b) If I use `pipeSuccess`, `(a,0)` is logged

I also would expect `Stream.emit(1).map(Left(_))` to be equivalent to `Stream.emit(Left(1))` but, using `pipeFailure`:

2a) if I have `Stream.emit(1).map(Left(_))`, nothing is logged.
2b) If I have `Stream.emit(Left(1))`, `(a,0)` is logged.

mpilquist:
In the fix for #2717, we discovered that sometimes after zipping, a closed scope is used for evaluation. We fixed #2717 by ensuring that manual interruption checks occur on the nearest open scope in the case the current scope has been closed already. We needed a similar check for effect evaluation, which checks interruption inside `interruptibleEval`.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Scope.scala
+++ b/core/shared/src/main/scala/fs2/internal/Scope.scala
@@ -442,7 +442,7 @@ private[fs2] final class Scope[F[_]] private (
     * that caused the interruption is returned so that it can be handled.
     */
   private[fs2] def interruptibleEval[A](f: F[A]): F[Either[InterruptionOutcome, A]] =
-    interruptible match {
+    openScope.map(_.interruptible).flatMap {
       case None =>
         f.attempt.map(_.leftMap(t => Outcome.Errored(t)))
       case Some(iCtx) =>
--- a/core/shared/src/test/scala/fs2/StreamZipSuite.scala
+++ b/core/shared/src/test/scala/fs2/StreamZipSuite.scala
@@ -420,10 +420,24 @@ class StreamZipSuite extends Fs2Suite {
     test("#2717 - unexpected behavior of Pull") {
       val stream = Stream(1).as(1).scope ++ Stream(2)
       val zippedPull =
-        stream.pull.uncons1.flatMap { case Some((_, s)) =>
-          s.zipWith(Stream(3))((_, _)).pull.echo
+        stream.pull.uncons1.flatMap {
+          case Some((_, s)) =>
+            s.zipWith(Stream(3))((_, _)).pull.echo
+          case None => Pull.done
         }
       val actual = zippedPull.stream.map(identity).covary[IO].compile.toList
+      actual.assertEquals(List((2, 3)))
+    }
+
+    test("#2762 - unexpected behavior of Pull") {
+      val stream = Stream(1).as(1) ++ Stream(2)
+      val zippedPull =
+        stream.pull.uncons1.flatMap {
+          case Some((_, s)) =>
+            s.zipWith(Stream(3))((_, _)).pull.echo
+          case None => Pull.done
+        }
+      val actual = zippedPull.stream.evalTap(_ => IO.unit).map(identity).compile.toList
       actual.assertEquals(List((2, 3)))
     }
   }


diesalbla (approved):

diesalbla:
@mpilquist So, I have noticed that this Pull Request has broken the memory leak tests. In particular, the test `broadcastThrough identity` and `observe + zip` seems to have broken after the change.

mpilquist:
I see these failures on HEAD - 2173855b148af506300a46ef2be16d23ddb7d811:

```
==> X fs2.MemoryLeakSpec.parJoin
==> X fs2.MemoryLeakSpec.broadcastThrough identity
```

Moving back a few commits to 61b593c4032efc7ba8fe72149708ba8aea180b68:

```
==> X fs2.MemoryLeakSpec.parJoin
==> X fs2.MemoryLeakSpec.hung merge
==> X fs2.MemoryLeakSpec.broadcastThrough identity
```

I'll keep moving back and update this comment with more findings, but I'm guessing the tests are failing erroneously and need tuning of their limits for new JVMs.

armanbilge:
Hmm, seems plausible to me that these tests may need to be adjusted, due to the increased memory pressure caused by fiber tracing. Run with tracing disabled?

mpilquist:
Yeah, I'm trying to disable tracing on the integration tests now

Issue #2757, opened 2021-12-17
 `observe` stream may be cancelled for successful output streams


zainab-ali:
This was raised following [a discussion in discord](https://discord.com/channels/632277896739946517/632310980449402880/920359666960171018).

The completion behaviour of `observe` has changed to be less deterministic.

```scala
outputStream = inputStream.observe(observerPipe)
```

`observe` constructs a stream from the `inputStream` and `observerPipe` (let's term this the observer stream) and runs this concurrently to the `outputStream`.

Previously, the observer stream would succeed if the output stream succeeded.  In recent fs2 versions, slow observer streams are cancelled when the output stream completes.

## Reproduction

The following code snippet shows an example of this:

```scala
val observer: Pipe[IO, Unit, Nothing] = _.chunkN(2)
  .evalMap(_ => IO.sleep(1.second))
  .onFinalizeCase(c => IO(println(s"Observer has finalized with case $c")))
  .drain

val main: Stream[IO, Unit] = Stream
  .eval(IO.unit)
  .observe(observer)
  .onFinalizeCase(c => IO(println(s"Main has finalized with case $c")))
}
```

This would previously print

```
Observer has finalized with case Completed
Main has finalized with case Completed
```

but now prints

```
Observer has finalized with case Canceled
Main has finalized with case Succeeded
```

You can experiment with this in Scastie using [v2.5.10](https://scastie.scala-lang.org/zainab-ali/n2IzsSJYTgmwkMjVaWt40w) and [v3.2.3](https://scastie.scala-lang.org/zainab-ali/CdXGdfRJSsKeW3SnhIK6NA) respectively.

## Cause

This was introduced in fs2 v3.0.0 in a refactoring of none-terminated queues. 

In [v2.5.10](https://github.com/typelevel/fs2/blob/v2.5.10/core/shared/src/main/scala/fs2/Stream.scala#L3746) the output stream was backed by a none-terminated queue. This was only terminated after the observer stream had completed, meaning that the observer stream would never be cancelled by the concurrent logic.

In [the current implementation](https://github.com/typelevel/fs2/blob/main/core/shared/src/main/scala/fs2/Stream.scala#L3828) we terminate the output stream once elements have been *pulled* by the observer stream, but not processed. The observer stream can be cancelled while processing the last elements.

I suspect this was an inadvertent change introduced [in the following refactor](https://github.com/typelevel/fs2/commit/ab2eaf387ec92ac149bdfaac2f54ca6d19b6b420).

[...]

PR #2765, opened 2021-12-23
Minimal fix of observe termination logic

zainab-ali:
This is a minimal fix for #2757.

While this works, there are a few things that I'm unsure of.

 - I'm not sure why we need both the `sinkChan` and `outChan` channels. A single channel seems to suffice.
 - I'm not sure why we're using `++ Stream.exec(outChan.close.void)` to close the channel instead of `onFinalize`. Is this because we only want the channel to close on success?

   I've created an alternative implementation in [a separate branch](https://github.com/typelevel/fs2/compare/main...zainab-ali:observe-reworked) that uses a single channel and finalizers, but this is a bit of a rehaul. Let me know if you'd prefer that PRed instead.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3824,8 +3824,7 @@ object Stream extends StreamLowPriority {
         val sinkOut: Stream[F, O] = {
           def go(s: Stream[F, Chunk[O]]): Pull[F, O, Unit] =
             s.pull.uncons1.flatMap {
-              case None =>
-                Pull.eval(outChan.close.void)
+              case None => Pull.done
               case Some((ch, rest)) =>
                 Pull.output(ch) >> Pull.eval(outChan.send(ch)) >> go(rest)
             }
--- a/core/shared/src/test/scala/fs2/StreamObserveSuite.scala
+++ b/core/shared/src/test/scala/fs2/StreamObserveSuite.scala
@@ -26,6 +26,7 @@ import scala.concurrent.duration._
 
 import cats.effect.IO
 import cats.effect.kernel.Ref
+import cats.effect.kernel.Resource
 import cats.syntax.all._
 import org.scalacheck.effect.PropF.forAllF
 
@@ -150,6 +151,23 @@ class StreamObserveSuite extends Fs2Suite {
           .compile
           .toList
           .assertEquals(List(1, 2))
+      }
+
+      test("3 - do not halt the observing sink when upstream terminates") {
+        Stream
+          .eval(IO(1))
+          .observe(
+            _.chunkN(2)
+              .foreach(_ => IO.sleep(100.millis))
+              // Have to do some work here, so that we give time for the underlying stream to try pull more
+              .onFinalizeCase {
+                case Resource.ExitCase.Canceled =>
+                  IO(fail("Expected exit case of Succeeded, but got Canceled"))
+                case _ => IO.unit
+              }
+          )
+          .compile
+          .drain
       }
     }
   }


mpilquist:
The alternative branch looks good to me. Any thoughts @SystemFw or @diesalbla?

diesalbla (approved):

SystemFw:
> I'm not sure why we're using ++ Stream.exec(outChan.close.void) to close the channel instead of onFinalize

so, this is weird. It's always been `++` there,  I remember trying to change it to `onFinalize` and seeing some test failures

SystemFw:
I'm cool with the alternative branch, the two channels have been translated directly from two queues, which I suspect are remnants of when we had `diamond`

diesalbla:
@zainab-ali Regarding the choices, IMHO it would be better to get the minimal fix for the bug merged, and then submit the reimplementation on a separate Pull Request. 

@SystemFw I am a bit confused with the alternative branch. 
```Scala
        val runner = sinkOut.through(pipe).onFinalize(outChan.close.void)

        def outStream =
          outChan.stream
            .flatMap { chunk =>
              Stream.chunk(chunk).onFinalize(guard.release)
            }

        outStream.concurrently(runner)
```

So, my intuition is that `observe` is like the Unix `tee`. Thus, if we have `val observed = source.observe(pipe)`, the `observed` stream should emit same elements as the `source` stream. Unless the observer `pipe` raises any error, its presence should not cause any difference in emitted outputs between `observed` and `source`. However, the `outChan`, in the code above, is the conduit through which the output chunks of the `observed` stream come through. Thus, if the pipe were defined just as `_ => Stream.empty` (i.e. finish immediately), wouldn't its finalisation also cut the outputs from the observed stream?

Another doubt, however, is about removing one of the two channels. In the alternative implementation, we only have a channel for the `observed` output, which allows the observer `pipe` to run several chunks ahead of the output, but not too far ahead. Wouldn't we also want to keep the output of `observed` not too far ahead of the progress of the observer pipe?

(Note: this is based on quick code-reading so testing may prove me wrong)

SystemFw:
Okay maybe let's merge the minimal version to avoid holding a release and we can revisit the discussion on the alternative version

zainab-ali:
Thanks @diesalbla ! I'd also prefer it merged now to fix the bug.

> So, one intuition is that observe is like the Unix tee. Thus, if we have val observed = source.observe(pipe), the observed stream should emit same elements as the source stream. Unless the observer pipe raises any error, its presence should not cause any difference in emitted outputs between observed and source. However, the outChan, in the code above, is the conduit through which the output chunks of the observed stream come through. Thus, if the pipe were defined just as _ => Stream.empty (i.e. finish immediately), wouldn't its finalisation also cut the outputs from the observed stream?

I don't have a good intutition on how `observe` should behave in this case. However, there's [an existing test that asserts that the `observer` stream is cut short](https://github.com/typelevel/fs2/blob/main/core/shared/src/test/scala/fs2/StreamObserveSuite.scala#L80).

> Nothing, however, prevents the observed outputs to be pulled far ahead before the observer pipe can catch up.

I think that's what the [semaphore](https://github.com/typelevel/fs2/blob/main/core/shared/src/main/scala/fs2/Stream.scala#L3809) does. This was changed from a bounded queue to a semaphore and an unbounded queue at some point in the past.

Perhaps this could be discussed in a separate issue / PR? If the implementation is reworked, it would be worth adding more tests for edge cases and possibly better docs.

diesalbla:
> I don't have a good intuition on how observe should behave in this case. However, there's an existing test that asserts that the observer stream is cut short.

@zainab-ali  Yeah, I was wrong. The documentation of the `observe` method states that is should halt when the observer halts, so that was not a problem. Sorry for the distraction.

PR #2767, opened 2021-12-24
Disable tracing in memory leak tests

mpilquist:
@diesalbla spotted memory leak tests were failing here: https://github.com/typelevel/fs2/pull/2763#issuecomment-1000571690

A bisect showed it was due to the CE3 upgrade. This PR disables tracing which fixes the leak tests.


armanbilge (approved):
Also @vasilmkd's optimizations in https://github.com/typelevel/cats-effect/pull/2673 have hopefully lessened the memory pressure, not sure.

diesalbla (approved):

PR #2768, opened 2021-12-24
Add `Channel#trySend`

armanbilge:
Opening to get feedback on the idea, which is to provide a "view" to a `Channel` as a `QueueSink`. This allows APIs to use the `QueueSink` interface and for a `Channel` to be plugged in as an implementation detail.

To adapt the `Channel` interface for `QueueSink` I added:
1. an implementation of `trySend` (analog to `tryOffer`), which is a non-blocking version of `send`.
2. a `Channel.ClosedException` which is thrown when encountering a `Channel.Closed`.

Thanks for your consideration.


core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -104,10 +116,16 @@ sealed trait Channel[F[_], A] {
 
   /** Semantically blocks until the channel gets closed. */
   def closed: F[Unit]
+
+  /** Provides an interface to the channel as a [[QueueSink]].
+    * Offering elements to a closed queue will raise a [[Channel.ClosedException]].
+    */
+  def asQueueSink: QueueSink[F, A]
```
armanbilge, line 123: Alternatively, `Channel` itself could extend `QueueSink`?

armanbilge:
Some feedback from @SystemFw
> I'm 50/50 about it
> Channel has a few semantic aspects that I think you ought to keep in mind when using it: graceful closure, single consumer, chunking by default (most cases of a queue consumed by a stream fall into this), so I'm not too sure about hiding those

I think those are fair points. WDYT about reducing the scope of this PR to just the `trySend` method? Then a `QueueSInk` adapter can be a user-land thing, if at all.

armanbilge:
Ok, I came to appreciate Fabio's point, and I changed my mind about the `QueueSink` thing. I'd still like the `trySend` op actually, so I kept that.

PR #2769, opened 2021-12-24
Fix build warnings

mpilquist:


io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -305,7 +305,7 @@ private[fs2] trait ioplatform {
   /** Stream of bytes read asynchronously from standard input.
     * Takes a dummy `Int` parameter for source-compatibility with JVM.
     */
-  @nowarn("cat=unused")
+  @nowarn // ("cat=unused") - unsupported category on scala 3
```
armanbilge, line 308:
```suggestion
  @nowarn("msg=never used")
```
h/t @bplommer when we setup Scala 3 fatal warnings for http4s

mpilquist:
The test failures are a bit worrisome... I assume the cause is the change to `ChunkGenerators`, replacing a buffer instance with an array instance, but why would that be slower?

mpilquist:
Oh hm, I bet it's actually this change:
```scala
ThisBuild / Test / javaOptions ++= Seq(
   "-Dscala.concurrent.context.minThreads=8",
   "-Dscala.concurrent.context.numThreads=8",
   "-Dscala.concurrent.context.maxThreads=8"
 )
```

mpilquist:
OK this should be good to go now

PR #2771, opened 2021-12-25
cross-platforming compression

yurique:


io/js/src/main/scala/fs2/io/compressionplatform.scala
```diff
@@ -65,22 +68,106 @@ private[fs2] trait compressionplatform {
...
+              F.async_[Unit] { cb =>
+                inflate.asInstanceOf[zlibMod.Zlib].close(() => cb(Right(())))
+              }
+            )
+        } yield (inflated, rest)
+      }
+
+      private def weird(
+          bytesWrittenDeferred: Deferred[F, Int],
+          bufferLimitSoft: Int
```
yurique, line 112: this `bufferLimitSoft` needs some thinking

io/js/src/main/scala/fs2/io/compressionplatform.scala
```diff
@@ -65,22 +68,106 @@ private[fs2] trait compressionplatform {
...
+                      case false => done.set(true)
+                    }
+                  }
+                case Some((chunk, rest)) =>
+                  Pull.eval(keptChunk.get).flatMap { keep =>
+                    Pull.eval(writingToTail.get).flatMap {
+                      case false =>
+                        val newKeep =
+                          if (keep.size + chunk.size > bufferLimitSoft / 2) {
+                            keep.take(keep.size / 2) ++ chunk
```
yurique, line 150: this is a weird thing to do here, haven't put a lot of thought into it

yurique: also should be `.takeRight`

yurique:
One question would be: can we think of something to organize all this stuff better? Right now we have (probably) too many "*platforms" and co.

io/js/src/main/scala/fs2/io/compressionplatform.scala
```diff
@@ -59,25 +62,98 @@ private[io] trait compressionplatform {
...
+
+                      Stream
+                        .resource {
+                          F.background {
+                            (bytesWritten.get, bytesPiped.get, lastChunk.get).mapN {
+                              (bytesWritten, bytesPiped, lastChunk) =>
+                                val bytesAvailable = bytesPiped - bytesWritten
+                                val headTrailerBytes = lastChunk.takeRight(bytesAvailable)
+                                val bytesToPull = trailerSize - headTrailerBytes.size
+                                if (bytesToPull > 0) {
```
yurique, line 116: I *think* I got the math right, here. But I'm not sure if this case happens during the tests.

armanbilge:
When you have a moment, would you mind running these JVM benchmarks, both before/after your changes?

https://github.com/typelevel/fs2/blob/main/benchmark/src/main/scala/fs2/benchmark/CompressionBenchmark.scala

I always forget how to run these things, these are the parameters Cats Effect uses:
```
jmh:run -i 10 -wi 10 -f 2 -t 1 cats.effect.benchmarks.DeepBindBenchmark
```

core/shared/src/main/scala/fs2/compression/CountPipe.scala
```diff
@@ -0,0 +1,49 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2.compression
+
+import cats.effect.{Ref, Sync}
+import fs2.{Chunk, Pipe, Pull, Stream}
+
+object CountPipe {
```
armanbilge, line 27:
```suggestion
private object CountPipe {
```

core/shared/src/main/scala/fs2/compression/GunzipResult.scala
```diff
@@ -0,0 +1,159 @@
...
+  *       // Access properties here.
+  *       gunzipResult.content
+  *     }
+  *
+  * @param content Uncompressed content stream.
+  * @param modificationTime Modification time of compressed file.
+  * @param fileName File name.
+  * @param comment File comment.
+  */
+class GunzipResult[F[_]](
```
armanbilge, line 43: I think it's okay if this is not a `case class` anymore. Seems mostly a container for the results.

```suggestion
final class GunzipResult[F[_]](
```

core/shared/src/main/scala/fs2/compression/GunzipResult.scala
```diff
@@ -0,0 +1,159 @@
...
+    val modificationTimeEpoch: Option[FiniteDuration],
```
armanbilge, line 44: I think I've heard it called "epoch time".
```suggestion
    val modificationEpochTime: Option[FiniteDuration],
```

core/shared/src/main/scala/fs2/compression/GunzipResult.scala
```diff
@@ -0,0 +1,159 @@
...
+    comment
+  )
+
+  @deprecated("GunzipResult is no longer a case class.", "x.x")
+  def productArity: Int = 4
+
+  @deprecated("GunzipResult is no longer a case class.", "x.x")
+  def productElement(n: Int): Any = n match {
+    case 0 => content
+    case 1 => modificationTimeEpoch
```
armanbilge, line 90:
```suggestion
    case 1 => modificationTime
```

core/shared/src/main/scala/fs2/compression/GunzipResult.scala
```diff
@@ -0,0 +1,159 @@
...
+      content: Stream[F, Byte],
+      modificationTimeEpoch: Option[FiniteDuration] = None,
+      fileName: Option[String] = None,
+      comment: Option[String] = None
+  ): GunzipResult[F] = new GunzipResult(
+    modificationTimeEpoch,
+    fileName,
+    comment,
+    content
+  )
```
armanbilge, line 144: I may be missing something, but I'm actually not sure if we need a public constructor for this class. Seems like it would only be created inside of fs2.

core/shared/src/main/scala/fs2/compression/ZipException.scala
```diff
@@ -19,28 +19,6 @@
...
-  * @param fileName File name.
-  * @param comment File comment.
-  */
-case class GunzipResult[F[_]](
-    content: Stream[F, Byte],
-    modificationTime: Option[Instant] = None,
-    fileName: Option[String] = None,
-    comment: Option[String] = None
-)
+class ZipException(message: String) extends Throwable(message)
```
armanbilge, line 24:
```suggestion
class ZipException(message: String) extends IOException(message)
```

Is this for compatibility with JS? What we've been doing is shim this on the JS side as `fs2.compression.ZipException` and provide type alias on the JVM side.

Here are some examples:
https://github.com/typelevel/fs2/blob/421c2424566caad1cc545a54c6657dc7bd37b00e/io/js/src/main/scala/fs2/io/IOException.scala#L54

https://github.com/typelevel/fs2/blob/421c2424566caad1cc545a54c6657dc7bd37b00e/io/jvm/src/main/scala/fs2/io/ioplatform.scala#L38

core/shared/src/main/scala/fs2/compression/internal/UnconsUntil.scala
```diff
@@ -0,0 +1,101 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2.compression.internal
+
+import fs2.{Chunk, INothing, Pull, Stream}
+import scodec.bits.BitVector
+import scodec.bits.crc.CrcBuilder
+
+private[compression] object UnconsUntil {
```
armanbilge, line 28: I _think_. Just a nit anyway :)
```suggestion
private object UnconsUntil {
```

io/js/src/main/scala/fs2/io/NodeCompression.scala
```diff
@@ -0,0 +1,24 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2.io
+
+object NodeCompression extends compressionplatform
```
armanbilge, line 24: I don't think we need this one. `fs2.io.compression` already extends `compressionplatform`.

io/js/src/test/scala/fs2/io/NodeJSCompressionSuite.scala
```diff
@@ -22,10 +22,17 @@
...
 import fs2.io.compression._
 import fs2.io.internal.ByteChunkOps._
+import org.scalacheck.effect.PropF.forAllF
+import scodec.bits.{ByteVector, crc}
+
+import java.nio.charset.StandardCharsets
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.FiniteDuration
 
 class NodeJSCompressionSuite extends CompressionSuite {
```
armanbilge, line 37: Oh, also if you've added any additional tests here would be cool move them into `CompressionSuite` to run them on JVM as well :)

core/shared/src/main/scala/fs2/compression/Gzip.scala
```diff
@@ -0,0 +1,631 @@
...
+    val MACINTOSH: Byte = 7
+    val Z_SYSTEM: Byte = 8
+    val CP_M: Byte = 9
+    val TOPS_20: Byte = 10
+    val NTFS_FILESYSTEM: Byte = 11
+    val QDOS: Byte = 12
+    val ACORN_RISCOS: Byte = 13
+    val UNKNOWN: Byte = 255.toByte
+
+    val THIS: Byte = System.getProperty("os.name") match {
```
armanbilge, line 594: We'll need to do something about this for JS :) Node has https://nodejs.org/api/process.html#processplatform.

core/shared/src/main/scala/fs2/compression/Gzip.scala
```diff
@@ -0,0 +1,631 @@
...
+                  F.unit
+                }
+              }
+            }
+          }
+        }
+      }
+      .flatMap(_ => Stream.empty)
+
+  private val gzipHeaderBytes = 10
```
armanbilge, line 543: I think if we make these ones `final` they can be inlined on JS. See https://github.com/scala-js/scala-js-macrotask-executor/pull/22

core/shared/src/main/scala/fs2/compression/CrcPipe.scala
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package fs2.compression
+
+import cats.effect.Ref
+import fs2.{Chunk, Pipe, Pull, Stream}
+import scodec.bits.BitVector
+import scodec.bits.crc.{CrcBuilder, crc32Builder}
+
+object CrcPipe {
```
armanbilge, line 29:
```suggestion
private object CrcPipe {
```

armanbilge (commented):
Just a first pass but this is looking very nice and clever :)

yurique:
A quick evaluation run of benchmarks (doesn't look good :( ).

`main`
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score   Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    2    42.049          ops/s
[info] CompressionBenchmark.deflate              false  thrpt    2   237.676          ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    2  2188.377          ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    2   677.429          ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    2    42.023          ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    2   230.814          ops/s
[info] CompressionBenchmark.inflate               true  thrpt    2  2392.483          ops/s
[info] CompressionBenchmark.inflate              false  thrpt    2   700.718          ops/s
```

this PR:
```
[info] CompressionBenchmark.deflate               true  thrpt   10   42.118 ± 0.141  ops/s
[info] CompressionBenchmark.deflate              false  thrpt   10  238.129 ± 0.500  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt   10  184.502 ± 0.948  ops/s // this one is super slow
[info] CompressionBenchmark.gunzip               false  thrpt   10  233.255 ± 0.820  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt   10   38.405 ± 0.040  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt   10  154.224 ± 0.172  ops/s
[info] CompressionBenchmark.inflate               true  thrpt   10  344.379 ± 3.621  ops/s // this one is fixed now
[info] CompressionBenchmark.inflate              false  thrpt   10  536.547 ± 1.661  ops/s
```

I updated the implementation of of inflate so now its performance is on-par with the old one. But: scodec crc is slow, very slow: `gunzip` is down to `328.634 ops/s` (vs `2188.377 ops/s`). I commented out all crc updates and got `1823.647 ops/s` (something is still slower, but not 10x slower).

yurique:
with `java.util.zip.CRC32`:
```
[info] CompressionBenchmark.gunzip               true  thrpt    5  1360.466 ± 8.239  ops/s
[info] CompressionBenchmark.gunzip              false  thrpt    5   565.759 ± 1.852  ops/s
```

armanbilge:
Thanks for doing those benchmarks. Yikes, that is very slow :(

I guess we have to stick with `java.util.zip.CRC32`. It may also be worth benchmarking with the pure Scala mutable CRC32 implementation, since the comments for it suggested that it is even faster than the `java.util.zip` version.

> A pure-java implementation of the CRC32 checksum that uses the same polynomial as the built-in native CRC32.
> This is to avoid the JNI overhead for certain uses of Checksumming where many small pieces of data are checksummed in succession.
> The current version is ~10x to 1.8x as fast as Sun's native java.util.zip.CRC32 in Java 1.6

If you want, try making that change directly in `main` (before your other changes) so we can see exactly how much difference switching the CRC32 implementation makes. If there's a good case to include it on the Java side purely for performance reasons, then we can share it with JS.

yurique:
After some trial and error and re-arranging stuff (and replacing the crc with a mutable one) I got this:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score    Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    5    42.101 ±  0.131  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    5   237.715 ±  2.536  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    5  1970.658 ± 10.947  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    5   652.282 ±  3.838  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    5    41.559 ±  0.172  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    5   228.952 ±  1.007  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    5  1825.193 ± 10.954  ops/s
[info] CompressionBenchmark.inflate              false  thrpt    5   637.698 ±  3.198  ops/s
```

yurique:
I tried replacing the CRC32 from java.util with a Scala implementation fgor jvm (in this PR, will try main a bit later):
```
[info] Benchmark                    (withRandomBytes)   Mode  Cnt    Score     Error  Units
[info] CompressionBenchmark.gunzip               true  thrpt    3  927.994 ±   8.462  ops/s
[info] CompressionBenchmark.gunzip              false  thrpt    3  458.095 ± 350.139  ops/s
```

(vs this, with java.util
```
[info] Benchmark                    (withRandomBytes)   Mode  Cnt     Score    Error  Units
[info] CompressionBenchmark.gunzip               true  thrpt    3  1949.081 ± 27.320  ops/s
[info] CompressionBenchmark.gunzip              false  thrpt    3   651.453 ±  2.669  ops/s
```
)

armanbilge:
Hm, seems to perform much worse then right? In that case I wouldn't bother with main then. I guess the conditions described in the comment don't apply here (the chunks must be sufficiently large to made the overhead worth). Also possible that the JDK has improved since 1.6 😅

yurique:
> Hm, seems to perform much worse then right? In that case I wouldn't bother with main then. I guess the conditions described in the comment don't apply here (the chunks must be sufficiently large to made the overhead worth). Also possible that the JDK has improved since 1.6 😅

Yeah :)

In any case, the benchmarks after the latest commits are not that bad anymore. `inflate` is still ~25% slower for some reason, but it (probably) should be fixable. Also I'm not sure why that is the case (I mean, I don't really understand the performance tradeoffs here: one of the big perf tweaks I did in the latest commits was removing things like `pull( resl.cons(partOfChunk) )` and dealing with the chunks fully within one `uncons`, and then doing `pull(rest)` – and I have no idea why it made such a huge difference).

armanbilge:
Yes, hopefully should be fixable. I'll try and take an in-depth look later and see if I have any ideas.

> I mean, I don't really understand the performance tradeoffs here

Would you mind pointing me to these changes specifically? Off the top-of-my-head my guesses are a combination of avoiding allocations and/or implementation details of the `Pull` interpreter.

yurique:
> Would you mind pointing me to these changes specifically?

For example this commit: https://github.com/typelevel/fs2/pull/2771/commits/dd41468e678be0da175c07666dfb9bf13ed21e79

There used to be a single `pull` function, that would try to feed the chunk into the inflater, and if not the whole chunk was consumed it would recurse like `pull(rest.cons(remaining))`.

I changed it into a pull that would take a chunk and try to feed it completely into the inflater (using the new `inflateChunk` helper – calling `inflater.setInput` and `inflater.inflate` multiple times if needed), and only then recursing with `pull(rest)` (or stopping if the inflater is finished).

This commit changed the benchmark from this:
```
[info] CompressionBenchmark.inflate               true  thrpt   10  344.379 ± 3.621  ops/s // this one is fixed now
[info] CompressionBenchmark.inflate              false  thrpt   10  536.547 ± 1.661  ops/s
```
to roughly this:
```
[info] CompressionBenchmark.inflate               true  thrpt    5  1825.193 ± 10.954  ops/s
[info] CompressionBenchmark.inflate              false  thrpt    5   637.698 ±  3.198  ops/s
```

yurique:
Although there is also some optimizations around this:
```scala
                          chunk.takeRight(
                            inflater.getRemaining
                          )
```
which I removed and replaced with 
```scala
                bytesChunk: Chunk.ArraySlice[Byte],
                offset: Int 
```

core/jvm/src/test/scala/fs2/JvmCompressionSuite.scala
```diff
@@ -33,6 +33,9 @@ import scala.collection.mutable
 import scodec.bits.crc
 import scodec.bits.ByteVector
 
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.FiniteDuration
+
 class JvmCompressionSuite extends CompressionSuite {
```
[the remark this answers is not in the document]
yurique: All tests are shared now.

io/js/src/test/scala/fs2/io/NodeJSCompressionSuite.scala
yurique: Tests are now shared :) (including that couple of new `unix gzip -> gunzip`)

core/shared/src/main/scala/fs2/compression/GunzipResult.scala
yurique: `ioJS/ testOnly fs2.io.NodeJSCompressionSuite`:

```
[error] Referring to non-existent class java.time.Instant
[error]   called from private fs2.compression.GunzipResult.$anonfun$modificationTime$1(scala.concurrent.duration.FiniteDuration)java.time.Instant
```

🤔

armanbilge: Crap. Also, I bet we have this same problem in http4s with JVM `Path`, except because of the munit bug our test suite doesn't catch it.

One work-around is to put this into platform-specific trait. But also probably not worth the trouble, I am not sure how/why this method would be used.

core/shared/src/main/scala/fs2/compression/Compression.scala
```diff
@@ -32,6 +34,128 @@ sealed trait Compression[F[_]] extends CompressionPlatform[F] {
...
+    *                   the chunks in the input stream will result in
+    *                   performance degradation of roughly 50-75%. Default
+    *                   size is 32 KB.
+    * @param deflateLevel     level the compression level (0-9)
+    * @param deflateStrategy  strategy compression strategy -- see `java.util.zip.Deflater` for details
+    * @param modificationTime optional file modification time (finite duration since unix epoch start)
+    * @param fileName         optional file name
+    * @param comment          optional file comment
+    */
+  def gzip2(
```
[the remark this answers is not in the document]
yurique: Done

core/shared/src/main/scala/fs2/compression/CountPipe.scala
yurique: Done (`private[fs2]` – it's referenced from `fs2.io`).

core/shared/src/main/scala/fs2/compression/GunzipResult.scala
yurique: Done.

core/shared/src/main/scala/fs2/compression/GunzipResult.scala
yurique: Done.

core/shared/src/main/scala/fs2/compression/GunzipResult.scala
yurique: Made it private.

core/shared/src/main/scala/fs2/compression/ZipException.scala
yurique: Aliased `ZipException` to `juz.ZipException` in jvm, new class (extends `IOException`) in js

core/shared/src/main/scala/fs2/compression/internal/UnconsUntil.scala
yurique: Done

io/js/src/main/scala/fs2/io/NodeCompression.scala
yurique: Yeah, that was a mistake =)

core/shared/src/main/scala/fs2/compression/Gzip.scala
yurique: Done.

core/shared/src/main/scala/fs2/compression/Gzip.scala
yurique: Made all vals final.

core/shared/src/main/scala/fs2/compression/CrcPipe.scala
yurique: Done (`private[fs2]` – it's referenced from `fs2.io`).

core/shared/src/main/scala/fs2/compression/internal/UnconsUntil.scala
yurique: Reverted back to `private[compression]` - wouldn't compile with scala 3 otherwise `¯\_(ツ)_/¯`.

yurique:
I think I've addressed the smaller feedback. Please let me know if I missed something. 
Had to make some utility classes `private[fs2]` (vs `private[compression]`) – those are referenced from `fs2.io`. We could move things around in `ioJS`, if it's worth it.

core/shared/src/main/scala/fs2/compression/internal/UnconsUntil.scala
```diff
@@ -0,0 +1,101 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2
+package compression
+package internal
+
+private[compression] object UnconsUntil {
+
+  /** Like Stream.unconsN, but returns a chunk of elements that do not satisfy the predicate, splitting chunk as necessary.
```
yurique, line 28: * [ ] update the doc

core/shared/src/main/scala/fs2/compression/GunzipResult.scala
yurique: Ah yeah, I still remember that one 🤔

yurique: I was thinking which one the `productElement` should return, and then forgot to think about it more.

yurique:
```
[error] fs2-core: Failed binary compatibility check against co.fs2:fs2-core_2.12:3.0.1! Found 2 potential problems (filtered 47)
[error]  * abstract method gzip(scala.Option,scala.Option,scala.Option,fs2.compression.DeflateParams,scala.Predef#DummyImplicit)scala.Function1 in interface fs2.compression.Compression is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("fs2.compression.Compression.gzip")
[error]  * class fs2.compression.GunzipResult is declared final in current version
[error]    filter with: ProblemFilters.exclude[FinalClassProblem]("fs2.compression.GunzipResult")
```

Not even nearly as bad as I was expecting 😅

core/jvm/src/main/scala/fs2/compression/CompressionPlatform.scala
```diff
@@ -262,747 +234,141 @@ private[compression] trait CompressionCompanionPlatform {
...
-      private val gzipInputSizeBytes = 4
-      private val gzipTrailerBytes = gzipInputCrcBytes + gzipInputSizeBytes
-
-      private val zeroByte: Byte = 0
-
-      private val fileNameBytesSoftLimit =
-        1024 // A limit is good practice. Actual limit will be max(chunk.size, soft limit). Typical maximum file size is 255 characters.
-      private val fileCommentBytesSoftLimit =
-        1024 * 1024 // A limit is good practice. Actual limit will be max(chunk.size, soft limit). 1 MiB feels reasonable for a comment.
+//              .onFinalize {
```
yurique, line 350: * [ ] figure this out or remove the commented out piece

yurique: https://discord.com/channels/632277896739946517/632310980449402880/927241378415640716

armanbilge:
Re MiMa:
1. I think the first one is okay, b/c `Compression` is a sealed trait.
2. The second one, I guess we don't have to make it final :) sorry, that may have been my suggestion 😆

yurique:
> * I think the first one is okay, b/c `Compression` is a sealed trait.

So if someone builds against a newer fs2 (and using this new method) but ends up with an older fs2 in the classpath – things will break. But there's no way around it, right? :)

Should I just add this one to the exclusions?

armanbilge:
> So if someone builds against a newer fs2 (using this new method) but ends up with an older fs2 in the classpath – things will break.

That's right, except we are usually less concerned about that case (forwards compatibility). The important one is backwards compatibility: if someone builds against e.g. fs2 3.0.0 and ends up with e.g. fs2 3.2.x in the classpath, we don't want their code to break. This happens all the time, if somebody releases a library depending on fs2 3.0.0, and never publishes it again, and you use that library in your application with the latest fs2.

Btw, I did have some crazy ideas how we can do this without introducing a new method by taking advantage of type erasure. e.g. we had
```scala
def foo(bar: Option[OldBar]) = ???
```
and now we want:
```scala
def foo(bar: Option[NewBar]) = ???
```

Because of type erasure, those signatures are the same actually. So we could maybe do something like this:
```scala
def foo(bar: Option[NewBar]) = internalFoo(bar)

private def internalFoo(bar: Option[Any]) = {
  bar.map {
    case oldBar: OldBar => ???
    case newBar: NewBar = ???
  }
  ???
}
```

This would be source-breaking for anyone using `oldBar` but otherwise I think should be bin-compat. Kind of crazy though 😆

yurique:
> Kind of crazy though 

 Maybe crazy it is, but I like it! 😆 

Though it won't help with our JS issue – `internalFoo` will be referring the old (JVM-only) class (like `File`) and linking will fail, right?

yurique:
```
[error] Failed tests:
[error] 	fs2.interop.reactivestreams.SubscriberStabilitySpec
```

armanbilge:
> Though it won't help with our JS issue – internalFoo will be referring the old (JVM-only) class (like File) and linking will fail, right?

Yes, although we could address that by having a platform-specific internal method e.g.
````
def anyBarToNewBar(bar: Option[Any])
```

The JVM checks cast for `OldBar` and `NewBar`, the JS one can "safely" cast straight to `NewBar`.

yurique:
> Yes, although we could address that by having a platform-specific internal method e.g.

Indeed!

core/jvm/src/main/scala/fs2/compression/CompressionPlatform.scala
yurique: As @mpilquist explained here: https://github.com/typelevel/fs2/issues/2785#issuecomment-1008046506 – we can't rely on `Stream.bracket` here. Changed it to use `Pull.bracketCase`.

yurique:
@armanbilge: The `inflateAndTrailer` always tracks the crc in the current code, even if it's discarded afterwards, so I tried making it optional (expectation: `inflate` benchmarks should improve). Results:

Always compute crc:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score    Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    41.480 ±   5.830  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   234.435 ±  27.687  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  1976.238 ± 118.449  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    3   632.852 ± 650.984  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    41.369 ±   2.747  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   223.206 ±  59.276  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  1888.293 ±  47.462  ops/s <-- difference is here as expected
[info] CompressionBenchmark.inflate              false  thrpt    3   646.041 ±  36.917  ops/s
```

Don't compute crc for `inflate`:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score    Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    41.703 ±  2.884  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   236.067 ± 35.279  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  2010.695 ± 53.009  ops/s <- this one is still ~10% slower than in main
[info] CompressionBenchmark.gunzip               false  thrpt    3   662.955 ±  6.763  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    41.202 ±  1.482  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   226.494 ± 28.391  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  2438.722 ± 59.915  ops/s <-- difference is here as expected
[info] CompressionBenchmark.inflate              false  thrpt    3   706.811 ±  4.072  ops/s
```

`main` (copied from earlier):
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score   Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    2    42.049          ops/s
[info] CompressionBenchmark.deflate              false  thrpt    2   237.676          ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    2  2188.377          ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    2   677.429          ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    2    42.023          ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    2   230.814          ops/s
[info] CompressionBenchmark.inflate               true  thrpt    2  2392.483          ops/s <-- main happened to be slower in that run :)
[info] CompressionBenchmark.inflate              false  thrpt    2   700.718          ops/s
```

armanbilge:
Very nice fix! Those numbers look so much better 😁 so IIUC the only difference now is that 10%  in `CompressionBenchmark.gunzip`?

I wonder if that is due to the additional delegation introduced by `CrcBuilder`, maybe you could try making it an `AnyVal` around the `CRC32`? I _think_ this would change the dynamic method dispatch to static method calls, not sure though.

Actually: can we just use the `CRC32` class directly? On the JS side, we can either keep the mutable `CRC32` shim, or we can shim `CRC32` by implementing its internals with scodec.

armanbilge:
@mpilquist what are your feelings about having this mutable CRC32 in fs2?
https://github.com/typelevel/fs2/blob/7b521773836a0f45175a278dfb646b8e1df18df8/core/js/src/main/scala/fs2/compression/internal/CRC32.scala

It's only for JS, and only for performance (the scodec implementation works fine, albeit slowly). I'm always a bit loathe to use custom code on JS purely for performance reasons, since it's obviously not the performance-oriented platform. @yurique will have to remind us, but I think the difference was considerable in this case.

yurique:
> but I think the difference was considerable in this case

It was huge.

(these are the old benchmarks from above, with the only difference being scodec vs juz.CRC32)
```
// scodec
[info] CompressionBenchmark.gunzip                true  thrpt   10  184.502 ± 0.948  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt   10  233.255 ± 0.820  ops/s

// java util 
[info] CompressionBenchmark.gunzip               true  thrpt    5  1360.466 ± 8.239  ops/s
[info] CompressionBenchmark.gunzip              false  thrpt    5   565.759 ± 1.852  ops/s
```

yurique:
> I wonder if that is due to the additional delegation introduced by CrcBuilder

I'd be really surprised if it was the case, but who knows :) (I'll try removing the delegation).
But the gunzip has changed in this PR (the way the stream is processed is slightly different, although the most changes are in the `inflate`, which is now as fast as it was) – I'll investigate.

yurique:
main:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score     Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    42.538 ±   2.228  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   239.607 ±   2.352  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  2256.685 ±  50.507  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    3   691.749 ±  24.496  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    42.626 ±   0.205  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   240.484 ±  32.535  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  2489.147 ± 119.165  ops/s
[info] CompressionBenchmark.inflate              false  thrpt    3   714.551 ±   9.180  ops/s
```

this PR:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score     Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    42.188 ±   2.239  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   239.414 ±  20.532  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  2059.542 ±  43.628  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    3   665.145 ±   5.446  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    41.829 ±   2.228  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   230.061 ±  26.736  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  2479.163 ± 150.095  ops/s
[info] CompressionBenchmark.inflate              false  thrpt    3   711.984 ±   7.221  ops/s
```

this PR with `type CRC32 = java.util.zip.CRC32` and no wrapper:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score     Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    42.617 ±   0.649  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   240.572 ±  15.511  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  2036.067 ± 135.649  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    3   669.611 ±   6.662  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    42.223 ±   1.626  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   230.407 ±  36.554  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  2513.640 ±  93.324  ops/s
[info] CompressionBenchmark.inflate              false  thrpt    3   709.845 ±   6.639  ops/s
```

yurique:
@armanbilge: What seems to have caused the performance degradation of `gunzip` is the usage of `Ref`s in the `inflateAndTrailer`.

I replaced the `Ref`s with this:
```scala
private[compression] class UnsafeMutableContainer[V] {
  @volatile private[this] var _value: V = _
  def get: V = _value
  def set(v: V): Unit =
    _value = v
}
```

and got this result:
```
[info] Benchmark                    (withRandomBytes)   Mode  Cnt     Score    Error  Units
[info] CompressionBenchmark.gunzip               true  thrpt    3  2400.993 ± 42.913  ops/s
[info] CompressionBenchmark.gunzip              false  thrpt    3   706.313 ±  7.577  ops/s
```

(vs `main`:
```
[info] CompressionBenchmark.gunzip                true  thrpt    3  2256.685 ±  50.507  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    3   691.749 ±  24.496  ops/s
```
)

yurique:
but this broke the `gzip and gunzip are reusable` test case (may be fixable, not sure yet)

yurique:
fixed

yurique:
and after replacing one `Pull.eval(F.delay(...))` with `Pull.pure(...)`:
```
[info] Benchmark                    (withRandomBytes)   Mode  Cnt     Score    Error  Units
[info] CompressionBenchmark.gunzip               true  thrpt    3  2171.996 ± 24.345  ops/s
[info] CompressionBenchmark.gunzip              false  thrpt    3   679.135 ± 19.623  ops/s
```

armanbilge:
Huh, I wasn't expecting that :) Thanks for tracking it down.

Those `Ref` are set only once right, when the `inflate` is completed right? Replacing a `delay` with `pure` seems suspect :) but maybe it points us in the direction that the `Pull` interpreter itself is the bottleneck? I think that the atomic set and volatile set here are pretty much equivalent 🤔

yurique:
> Those `Ref` are set only once right, when the `inflate` is completed right?

Correct. 



> Replacing a `delay` with `pure` seems suspect

That was something I did *after* I removed the Refs and fixed a bug: perf went from `2119.328 ± 64.257 ops/s` to `2171.996 ± 24.345  ops/s` – not huge :)

As I understand – it's those additional `Stream.eval(...).flatMap`s. I removed one (when I got rid of Refs) and the perf went up `2036.067` ->  `2400.993`. Then I had to add one (in a different place) to fix the failing test: perf went down `2400.993` -> `2119.328`.

armanbilge:
Is there a way to reduce the number of `Pull` "stages" in the implementation, without relying on custom constructs like your `UnsafeMutableContainer`? Instead of removing the `Ref`s I wonder if it's possible to combine them into a single `Ref`, and whether that would help at all.

yurique:
> Instead of removing the `Ref`s I wonder if it's possible to combine them into a single `Ref`, and whether that would help at all.

So the `Ref`s are created like this (there's one or two other places with refs): https://github.com/typelevel/fs2/pull/2771/files#diff-8c2b0bb4111ef230a00138813c001efa1fa187ee4c6f43cf514e0074b570e084R246-R253
(those are read similarly: https://github.com/typelevel/fs2/pull/2771/files#diff-4bed624a3e9e0e35ae957f81c77635ef4b34547e2431ec3923bec420a34ddda2R499-R501)

So while there are multiple `Ref`s, there's just one `Stream.eval`. My understanding is that it's this `Stream.eval(...).flatMap` that is hitting the performance, not the `Ref`s per se.

But I'll give it a try (combining the `Ref`s) – hoping I'm wrong about it :).

Another thing about the `Ref`s: even though it feels safer to use them (instead of weird mutable stuff) – in this case it doesn't really matter. In any case we need to make sure the `Ref`s are not read before they are set (kind of what `Deferred` would give us), and we do – that's handled by the `stream1 ++ stream2`: `Ref`s are set in `stream1` and read in `stream2`.

yurique:
> Is there a way to reduce the number of Pull "stages" in the implementation, without relying on custom constructs like your UnsafeMutableContainer?

I couldn't find a way to simplify things more. And I don't think even using more `Ref`s / `UnsafeMutableContainer`s would allow to do so.

The `Pull`s here are like this:
* read the length `N` from the stream and skip `N` bytes (updating the crc32 while skipping)
* read the bytes until there's a `0` (updating the crc32), but only take at most approximately M bytes (soft limits for file name / comment length)
* go through the stream and feed it into the inflater (updating the crc32 with the inflater output), stop when the inflater says it's done, get `trailerSize` bytes from the [ bytes not processed by the inflater ++ rest of the stream ]

Some of those are simpler (but complicated by the need to update the crc32 and/or count the bytes), others are tricky on their own.

I'd be happy to receive suggestions, though :)

yurique:
The results with a combined `Ref`:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score     Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    42.204 ±   8.658  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   240.016 ±  32.985  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  2089.465 ±  87.978  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    3   669.763 ±  12.048  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    42.233 ±   1.149  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   230.807 ±  29.052  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  2471.074 ± 417.530  ops/s
[info] CompressionBenchmark.inflate              false  thrpt    3   715.019 ±  21.269  ops/s
```

Before:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score     Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    42.188 ±   2.239  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   239.414 ±  20.532  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  2059.542 ±  43.628  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    3   665.145 ±   5.446  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    41.829 ±   2.228  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   230.061 ±  26.736  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  2479.163 ± 150.095  ops/s
[info] CompressionBenchmark.inflate              false  thrpt    3   711.984 ±   7.221  ops/s
```

yurique:
combining the `Ref`s complicates stuff in js (those refs are set in different places/callbacks there).

armanbilge:
@yurique thanks for all the updates. Actually that last pair of numbers you posted in https://github.com/typelevel/fs2/pull/2771#issuecomment-1015251279 look quite good! Basically the same within the margin of error.

At this point I'm not sure what constitutes weird anymore :) for example, I wonder if these `def`s could somehow be nested so that they can share a `var` instead of using `UnsafeMutableContainer`—but I'm not sure if that's more or less weird.

I'll try and take a deeper look sometime this week. Thanks again, for the massive effort this turned out to be...

yurique:
> I wonder if these `def`s could somehow be nested so that they can share a `var` instead of using `UnsafeMutableContainer`

That would be... tricky :)

The `Ref`s (or `UnsafeMutableContainer`s) are created and set in platform-specific `inflate` implementations, but  consumed in the shared `Gzip.gunzip`.

The latest numbers are not too bad, I agree :) (but combining `Ref`s didn't give us the perf improvement I was silently hoping for).

yurique:
I'm still concerned about all the "weird" things we have to have with `GunzipResult`.

Should we instead think about introducing a new set of APIs: `Gzip` (next to `Compression`), have the existing jvm gzip functions forward to it, make a new `GunzipResult` (in a new package) and deprecate the old one? 🤔

armanbilge:
> I'm still concerned about all the "weird" things we have to have with `GunzipResult`.

I'm less worried about that. In this case it would be okay to implement that class twice, once on JS and once on JVM and avoid the weirdness.

If you still have appetite to try a different approach, here's an idea: you could try interacting directly with the Node.js APIs, rather than converting them to `Stream`. Then, maybe more of the current JVM code can be reused by building a common interface on top of the JDK/Node.js APIs. This may still be tricky because Node.js is async. WDYT?

yurique:
> you could try interacting directly with the Node.js APIs

> This may still be tricky because Node.js is async

Interesting 🤔 

Also, there are sync alternatives to the async functions in node. I can't remember if I considered this initially (and if so, what it was that made me not do it this was) – I'll take a peek :)

armanbilge:
I'd advise against the sync methods, since they are effectively blocking the main (and only) thread in the Node runtime.

yurique:
Oh, now I also remember – the sync functions operate on whole inputs (as well as the async ones). What we use is the "inflater" and "deflater" which are `Duplex`es, and we have this whole `suspendReadableAndRead` machinery to deal with those (which involves subscribing to events etc) – something tells me we can't really do it better than that :) (or utilize these things in a way we utilizy jvm's inflater and deflater, which are sync and designed to take data in in chunks).

armanbilge:
> or utilize these things the way we utilize jvm's inflater and deflater, which are sync and designed to take data in in chunks

The `Duplex` works basically the same way, you can write chunks, and it emits chunks as events. Only it happens async instead of sync, but it's all `F` in the end 😉 Assuming one chunk comes out per chunk that goes in, it should be easy I would think. Otherwise, it gets a lot harder and you'd probably end up re-implemetning `Stream` 😆

yurique:
Yeah, it can probably be more or less straightforward :)

Those pulls that do (approx) `inflater.inflate(nextChunk); Pull.output(inflaterOutput)` would have to do (approx) `Pull.eval(inflateChunk(nextChunk).flatMap(_.get)).flatMap( output => Pull.output(output))`.

The `inflateChunk(nextChunk)` would probably have to return something like a `Deferred`: but we don't have `Async` in the jvm interface, and also this could affect the performance quite a bit.

yurique:
Otherwise, having only this, lower-level, part have platform-specific implementations and everything else – shared would be nice indeed!

armanbilge:
I don't think it has to return `Deferred`. I'm hoping the difference should be essentially `F.delay( ... do something sync ... )` for JVM vs `F.async_(cb => ... do something async ...)` for JS, which both give you `F`.

yurique:
> `F.async_(cb => ... do something async ...)` for JS

Indeed! :)

Well, I'll give it a try as soon as I can. This should work. Curious to see how it affects the benchmark.

yurique:
@armanbilge 

I tried implementing inflate in terms of this:
```scala
trait ChunkInflater[F[_]] {
  /** @param bytesChunk bytes to inflate
    * @return (inflatedChunk, remainingBytes, finished)
    */
  def inflateChunk(
      bytesChunk: Chunk[Byte]
  ): Pull[F, INothing, (Chunk[Byte], Int, Boolean)]
}
```

jvm implementation:
```scala
new ChunkInflater[F] {
        def inflateChunk(
            bytesChunk: Chunk.ArraySlice[Byte],
            offset: Int
        ): Pull[F, INothing, (Chunk[Byte], Int, Boolean)] = {
          inflater.setInput(
            bytesChunk.values,
            bytesChunk.offset + offset,
            bytesChunk.length - offset
          )
          val inflatedBytes = inflater.inflate(inflatedBuffer)
          Pull.pure(
            (
              copyAsChunkBytes(inflatedBuffer, inflatedBytes),
              inflater.getRemaining,
              inflater.finished()
            )
          )
        }
      }
```

(changes: https://github.com/yurique/fs2/pull/1/files)

Have the tests passing for jvm.

Inflate implementation is the same, except using `ChunkInflater` instead of directly interfacing with `juz.Inflater`.

Benchmarks look weird, though (ran it a few times, it's consistent):

```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score    Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    4    41.692 ±  0.356  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    4   235.074 ±  1.212  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    4  1557.763 ± 19.680  ops/s <----
[info] CompressionBenchmark.gunzip               false  thrpt    4   603.812 ±  2.886  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    4    41.193 ±  0.687  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    4   226.022 ±  3.354  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    4  2446.234 ± 15.744  ops/s <----
[info] CompressionBenchmark.inflate              false  thrpt    4   700.276 ±  4.743  ops/s
```


before:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score    Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    41.596 ±  0.976  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   234.454 ± 13.996  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  2023.898 ± 87.731  ops/s <----
[info] CompressionBenchmark.gunzip               false  thrpt    3   655.208 ±  5.633  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    41.279 ±  1.060  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   225.414 ± 28.269  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  2450.515 ± 53.559  ops/s <----
[info] CompressionBenchmark.inflate              false  thrpt    3   700.596 ± 14.874  ops/s
```

I can't explain the difference in the `gunzip` benchmarks given near-identical results for `inflate` 🤔  (gunzip implementation is the same). Can you spot something I'm missing?

---

Kind of half way there with the JS implementation (though it turns out even trickier than I expected), but unless we can fix this performance issue, not sure if that's the way to go.

---

As a side note: in the inflate implementation, I have this:
```
              bytesChunk: Chunk.ArraySlice[Byte],
              offset: Int,
```

This is used to send the whole chunk into the inflater (inflater doesn't always accept the whole input because the output might overflow the output buffer, so this slice is being re-used, and the `offset` within that slice is increased as needed).

So I thought if I should maybe just use the `Chunk` and `.takeRight(...)` instead.

(changes: https://github.com/yurique/fs2/pull/2/files)

Got a rather surprising result:

with `.dropRight`
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt    Score    Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3   41.563 ±  1.225  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3  234.662 ± 25.051  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  374.088 ±  8.450  ops/s  <----
[info] CompressionBenchmark.gunzip               false  thrpt    3  601.324 ±  5.705  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3   41.181 ±  0.693  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3  223.710 ± 29.040  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  422.174 ± 16.779  ops/s  <----
[info] CompressionBenchmark.inflate              false  thrpt    3  697.039 ±  4.659  ops/s
```

armanbilge:
@yurique so if I'm reading those benchmarks correctly, `gunzip` performance decreased but `inflate` stayed the same. IIRC the difference is that `gunzip` calculates the CRC for verification but `inflate` does not (apologies if I remember incorrectly). So could it be related to that somehow?

yurique:
@armanbilge yes, that's right – `gunzip` "requests" the calculation of the crc and the trailer. 

I was thinking "those parts haven't changed at all either" but your comment reminded me of something,  I looked through it again. And there *is* a subtle difference: 

```scala
if (track) crcBuilder.update(inflatedChunk)
```

So now `inflatedChunk` is a `Chunk`, and `crcBuilder.update` calls the `.toArraySlice` under the hood. Before, there was no `.toArraySlice` involved as the inflater output buffer was available as `Array[Byte]`. This must be it.

I changed it so that the `ChunkInflater` returns its buffer + length without wrapping in a chunk, so there's no extra `.toArraySlice` when calling `crcBuilder.update`.

And the benchmark is almost back to where it was:
```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score     Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    41.041 ±   4.382  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   231.742 ±  18.002  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  1971.619 ±  67.197  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    3   646.719 ±   6.831  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    40.631 ±   2.349  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   221.801 ±  35.506  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  2408.610 ± 173.101  ops/s
[info] CompressionBenchmark.inflate              false  thrpt    3   690.958 ±  10.401  ops/s
```

But I'm really surprised: the `Chunk` is created with `Chunk.array(target, 0, length)` – I'd assumed `.toArraySlice` would be nearly a no-op in such a case. I guess it's not :) 
```scala
  def toArraySlice[O2 >: O](implicit ct: ClassTag[O2]): Chunk.ArraySlice[O2] =
    this match {
      case as: Chunk.ArraySlice[_] if ct.wrap.runtimeClass eq as.getClass =>
        as.asInstanceOf[Chunk.ArraySlice[O2]]
      case _ => Chunk.ArraySlice(toArray, 0, size)
    }
```

---

But this is good! I'll try finishing the JS implementation.

armanbilge:
Nice! 🎉 And very interesting about `toArraySlice` 🤔

yurique:
@armanbilge so I managed to make a JS implementation – it was hell, but it works =)

The benchmark results are the same as before the JS implementstion (as they should be – I added a couple of things to the shared inflate implementation, but those do not kick in in JVM).

```
[info] Benchmark                     (withRandomBytes)   Mode  Cnt     Score     Error  Units
[info] CompressionBenchmark.deflate               true  thrpt    3    41.846 ±   0.349  ops/s
[info] CompressionBenchmark.deflate              false  thrpt    3   234.625 ±  14.218  ops/s
[info] CompressionBenchmark.gunzip                true  thrpt    3  2015.259 ± 119.846  ops/s
[info] CompressionBenchmark.gunzip               false  thrpt    3   655.609 ±  10.494  ops/s
[info] CompressionBenchmark.gzip                  true  thrpt    3    41.318 ±   1.437  ops/s
[info] CompressionBenchmark.gzip                 false  thrpt    3   225.468 ±  14.121  ops/s
[info] CompressionBenchmark.inflate               true  thrpt    3  2413.895 ± 354.697  ops/s
[info] CompressionBenchmark.inflate              false  thrpt    3   697.494 ±   5.701  ops/s
```

---

As the result – the inflate and gzip implementation is shared between JS and jvm, apart from the platform specific interface to the inflater.

Changes are here: https://github.com/yurique/fs2/pull/1/files

jvm chunk inflater: https://github.com/yurique/fs2/pull/1/files#diff-8c2b0bb4111ef230a00138813c001efa1fa187ee4c6f43cf514e0074b570e084R147-R181

js chunk inflater: https://github.com/yurique/fs2/pull/1/files#diff-1b982c5a5c6bd8cdbb47ca7c2114327cb202075ee95749b8024d7078b6fa7490R73-R207

shared code: https://github.com/yurique/fs2/pull/1/files#diff-e7f835b1438d510176b8ad23b915f2c27e152bc2768847265da54e9ce98acb50R33 

its mostly the same as the jvm implementation used to be before this change, but with an additional pull "stage" – `drain`, which is only invoked in JS; it reads the output without writing to the inflater (when there's nothing more to write) until the read `end` event is emitted.

armanbilge:
Haven't had a chance to look yet but just wanted to 🎉 that's amazing!

armanbilge:
I'm really sorry, has this PR just been waiting for my review? 😬 😅 I'll try and resolve the merge conflicts.

armanbilge:
Oh I see, there are more changes in a PR against this PR :)
- https://github.com/yurique/fs2/pull/1

yurique:
I'm trying to recollect what was going on here. 

I think the PR agains this PR was an "experiment" which turned out to be a successful one – more shared code between JS and JVM, and without a performance hit. But I can't remember all the details off the top of my head now :'(

armanbilge:
I remember :) you were able to implement a common `Pull`-based inflater with both the JDK and Node.js APIs which (finally!) wasn't a performance regression on the JVM 🎉 Then everything else could be expressed in terms of that.

I _think_ that's the hardest part of this PR 😅 so thanks for your _big_ effort on it!

IIRC some remaining things are dealing with bincompatibly changing the API to use `FiniteDuration` instead `java.time` and also platforming the OS-detection stuff. These are annoying, but doable.

The other thing that I remember was frustrating was the poor performance of the CRC calculation on JS, particularly when using the scodec-bits implementation. This PR inlines a mutable implementation which works better (I can't remember how much better). I wish we didn't have to take that on but not sure if there's another way since Node.js doesn't expose native methods for calculating the CRC.

armanbilge:
So I took another look at how http4s is using the `Compression` trait especially because I was wondering how the Node.js native HTTP client supports compression without all this functionality e.g. see:
https://nodejs.org/api/zlib.html#compressing-http-requests-and-responses

It seems none of the http4s gzip middlewares are using the additional headers such as file name or modification time, it is just directly interfacing with the underlying byte stream:

https://github.com/http4s/http4s/blob/b3220ab944042a6729ecdcbbd2b8a034250e98f5/client/jvm/src/main/scala/org/http4s/client/middleware/GZip.scala#L65

https://github.com/http4s/http4s/blob/b3220ab944042a6729ecdcbbd2b8a034250e98f5/server/jvm/src/main/scala/org/http4s/server/middleware/GZip.scala#L78

IIUC it _should_ be possible to support these http4s use-cases just by directly shelling out to the Node.js APIs like in the example there. Not to discount your huge efforts on this PR :( but maybe the easy win here for http4s is to add new cross-platform APIs to `Compression` for gzipping without the optional headers. I think this will also be more performant, since calculating CRCs in raw JS was a bottleneck.

Meanwhile, I see there is an (unfortunately inactive) feature request for Node.js itself to add native support for these optional headers. If that ever materializes, we would be able to implement these APIs much more easily and with much better performance.

- https://github.com/nodejs/node/issues/39545

What do you think?

PR #2784, opened 2022-01-03
Stream: extend docs for `attempts`, `last`, `map`, `emits`.

diesalbla:
For `map` and `attempts`, add some documentation concerning chunk structure
and the behaviour in the case of error or interruption.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -174,7 +174,14 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     */
   def as[O2](o2: O2): Stream[F, O2] = map(_ => o2)
 
-  /** Returns a stream of `O` values wrapped in `Right` until the first error, which is emitted wrapped in `Left`.
+  /** Returns a stream of `O` values wrapped in `Right` until an error is raised.
+    * If an error is raised, it is emitted wrapped in `Left`.
+    *
+    * This combinator preserves chunk structure. Until an error is raised, the
+    * resulting stream emits the same elements as this stream, in the same order
+    * and chunk structure, but with the elements wrapped in a `Right`.
+    * If an error is raised, the resulting stream ends by emitting the error
+    * wrapped in a `Left`, in a singleton chunk.
     *
     * @example {{{
     * scala> import cats.effect.SyncIO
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1722,7 +1729,11 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     }.stream
   }
 
-  /** Returns the last element of this stream, if non-empty.
+  /** Returns the last element of this stream, if non-empty and finite.
+    *
+    * Note that this is a fold-right combinator. Pulling from the resulting stream
```
mpilquist, line 1734: Recommend removing "fold-right combinator" as that's not a well known description.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1747,11 +1758,33 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+    *
+    * **Errors** If an error is raised when pulling from the source (`this`)
+    * stream, then the mapped stream rethrows that error. If the function `f`
+    * throws an Exception `e` when applied to an element `o`, then the entire
+    * chunk containing that element is discarded, and the stream fails with
+    * that exception `e`.
+    *
+    * **Interruption**: if an interruption occurs, then the mapped stream is
+    * cut between chunks: the last output chunk is still the image of a whole
+    * source chunk. Once an interrupted is detected, the stream stops pulling
```
mpilquist, line 1774: s/an interrupted/interruption/

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1747,11 +1758,33 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+    * from the source chunk. However, an interruption can be detected after
```
mpilquist, line 1775: s/an interruption/interruption/

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1747,11 +1758,33 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+    * a chunk is pulled from the source but before it is mapped and emitted.
+    * When this happens, that chunk is dropped and lost.
+    *
+    * This function follows the Functor laws.
```
mpilquist, line 1779: I'd remove this note about functor laws - unless otherwise stated, all instances should be assumed lawful.

PR #2789, opened 2022-01-09
Add `userHome`, `currentWorkingDirectory` to `Files` API

armanbilge:
Not entirely sure if `Files` is the best place for these.

Maybe after https://github.com/typelevel/cats-effect/pull/2650 lands we can add a test that compares to env.


mpilquist (approved):

ChristopherDavenport:
Really looking forward to this, with this plus some env/os-detect actual applications can be written in pure style super straight-forward. Where we can customize where we run, where we save data, etc.

diesalbla (approved):

rossabaker:
`java.io.tmpdir` is another meaningful one to the JVM, but I don't know what the Scala.js equivalent would be.  Don't block this on that if it's not obvious.

armanbilge:
We already have facilities for creating temporary files and temporary directories. Is it helpful to expose the tmp dir directly like this?

rossabaker:
It's useful for providing the default, but that's already tucked away behind an Option.  Distraction withdrawn.

PR #2794, opened 2022-01-19
Remove reference to scalaz-stream from readme

bplommer:
It's probably time.


diesalbla (approved):

mpilquist:
Maybe we could add a history section referencing the origins of FS2 in chapter 15 of FPiS and its first few years as scalaz-stream?

PR #2798, opened 2022-01-23
Remove unneeded uses of `covary`

bplommer:


armanbilge:
JS failure is a flake, sorry. See https://github.com/typelevel/sbt-typelevel/issues/81. I'm working on a fix.

PR #2808, opened 2022-02-01
Drop-Right: reimplement to keep chunk structure

diesalbla:
The current implementation of drop-right is constantly splitting and appending chunks. This means that the "chunk structure", the partitioning of output items in the chunks, is distorted by a method that should not need do that.

This commit changes the implementation. Instead of Chunk-append and split, we use a Queue of chunks and a total size counter. Chunks from the source are added to this queue until total size is enough to cover the number of dropped elements. Afterwards, whenever a chunk can be dropped while keeping that covered, the chunk is emitted as output.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -24,6 +24,7 @@ package fs2
 import scala.annotation.{nowarn, tailrec}
 import scala.concurrent.TimeoutException
 import scala.concurrent.duration._
+import scala.collection.mutable.ArrayDeque
 import cats.{Eval => _, _}
 import cats.data.Ior
 import cats.effect.{Concurrent, SyncIO}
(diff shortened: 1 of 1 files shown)


mpilquist:
Hm, the current version should basically perform the same since the accumulated chunk is a `Chunk.Queue` instance, which does similar total size tracking, right?

diesalbla:
> Hm, the current version should basically perform the same since the accumulated chunk is a Chunk.Queue instance, which does similar total size tracking, right?

So, using a `Chunk.Queue` may mean that there are no underlying array copies or duplications, only a change of indexes on prefixes. However, that is still less optimal than emitting the same whole chunks just as they come.

mpilquist:
OK, I see. Can we implement this without the mutable ArrayDeque?

nikiforo:
I haven't looked at the problem thoroughly, so consider my comment as probably wrong.
Consider we have 3 5-sized chunks. We do `stream.dropRight(11)`. Will the new implementation output first 4 elements as soon as the third chunk arrives? From my POV strategy of "emitting the same whole chunks just as they come" doesn't look possible.

diesalbla:
> Consider we have 5-sized chunks stream. We do stream.dropRight(11). Will the new implementation output first 4 elements as soon as the third chunk arrives? From my POV strategy of "emitting the same whole chunks just as they come" doesn't look possible.

For example sake, each chunk is five times same character and go 'a' to 'z'. The new implementation would first pull the 'a', 'b', and 'c' chunks into the queue, giving a total size of 15. After pulling 'd', the total size of the queue without its head chunk (the 'a' chunk) is enough to cover the dropped ones, so it releases the 'a' chunk. Similarly, after pulling 'e' it would release 'b', etc. Once it has pulled 'z' and reached the end of the stream, the queue should have the chunks 'w', 'x', 'y', and 'z'. At that point it would emit the 'w' chunk whole, and the 4-length prefix of 'x', dropping the last 11 characters.

nikiforo:
>  After pulling 'd', the total size of the queue without its head chunk (the 'a' chunk) is enough to cover the dropped ones, so it releases the 'a' chunk.

So, if every chunk creation takes 1 min, `main`'s implementation will emit first four `a`s after 3 minutes while `pr` version will emit five `a`s after 4 minutes?

PR #2815, opened 2022-02-05
Add Signal.waitUntil

SystemFw:


core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -48,6 +48,69 @@ trait Signal[F[_], A] {
...
+    * - Scenarios where conditions don't change instantly, such as
+    *   periodic timed processes updating the `Signal`/`SignallingRef`.
+    * - Scenarios where conditions might change instantly, but the `p`
+    *   predicate is monotonic, i.e. if it tests true for an event, it
+    *   will test true for the following events as well.
+    *   Examples include waiting for a unique ID stored in a `Signal`
+    *   to change, or waiting for the value of the `Signal` of an
+    *   ordered `Stream[IO, Int]` to be greater than a certain number.
+    */
+  def waitUntil(p: A => Boolean): F[Unit]
```
armanbilge, line 113: Since `Signal` isn't sealed isn't adding this method a breaking change? :(

SystemFw: Yeah...that sucks

SystemFw: I'll add it to SignalOps and that should fix it I think

armanbilge: Does it need to go on `SignalOps`? You could just put that default implementation and the `Concurrent` constraint directly here.

SystemFw: I could yeah, I guess I just went with how `map` is implemented. @mpilquist opinions?

core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -71,6 +71,65 @@ object Signal extends SignalInstances {
...
+      * - Scenarios where conditions don't change instantly, such as
+      *   periodic timed processes updating the `Signal`/`SignallingRef`.
+      * - Scenarios where conditions might change instantly, but the `p`
+      *   predicate is monotonic, i.e. if it tests true for an event, it
+      *   will test true for the following events as well.
+      *   Examples include waiting for a unique ID stored in a `Signal`
+      *   to change, or waiting for the value of the `Signal` of an
+      *   ordered `Stream[IO, Int]` to be greater than a certain number.
+      */
+    def waitUntil(p: A => Boolean)(implicit F: Concurrent[F]): F[Unit] =
```
diesalbla, line 131: Some Signals can change very frequently. If it does, there could be no limit to the number of values that `self.discrete` would emit. If so, could the `.compile.drain` result in a fiber that could keep running and exhaust CPU?

If so, perhaps it could help if this method took a delay parameter, a FiniteDuration, so as to limit the sampling frequency on this signal.

core/shared/src/main/scala/fs2/concurrent/Signal.scala
mpilquist: No strong ones

core/shared/src/main/scala/fs2/concurrent/Signal.scala
SystemFw: mm I think I disagree, the same concern would apply to everything that uses a Signal (e.g. `pauseWhen`) , forcing this noncompositional sampling parameter everywhere, and potentially slowing reaction times for the vastly more common case where this isn't a concern. I think if you're in the corner case where the frequency changes so frequently and so consistently that your CPU gets exhausted even with `discrete`, you will need to solve that at the source (e.g. by creating another `Signal` that `debounce`s the previous one). I think in practice though, this would only be a problem when using `continuous` , not `discrete`

core/shared/src/main/scala/fs2/concurrent/Signal.scala
SystemFw: I found a clear justification for having it as an instance method: scaladoc friendliness. I think `map` has a similar problem, but then you'd have the question of whether removing `SignalOps` entirely constitutes source breakage or not, so I'll punt on `map` for now

armanbilge: We can keep `SignalOps` deprecated, just remove the `implicit` modifier. So there's no source-breakage but the `map` on `Signal` will clearly take precedence.

SystemFw: I'll leave that for a potential future PR

diesalbla (approved):

Issue #2825, opened 2022-02-16
parEvalMapUnordered and friends can deadlock on exceptions in fs2 3.x, but not in fs2 2.x


wjoel:
The following program gets stuck (deadlocks, never terminates) on fs2 3.2.4, but immediately finishes with an exception on fs2 2.5.10:

```scala
import cats.effect.{IO, IOApp}
import fs2.Stream

object Main extends IOApp.Simple {
  def explodeIf1(int: Int): IO[Unit] = {
    if (int == 1) {
      throw new RuntimeException("kaboom")
      // If using IO.raiseError instead, the program terminates also for fs2 3.x.
      // IO.raiseError(new RuntimeException("kaboom!"))
    } else {
      IO.unit
    }
  }

  override def run: IO[Unit] = {
    println("Hello, world!")

    val nums = 1.to(1).iterator

    Stream
      .fromIterator[IO](nums, chunkSize = 10)
      .parEvalMapUnordered[IO, Unit](2) { i =>
      explodeIf1(i)
    }.compile.drain
  }
}
```
The program also finishes (with an exception) on fs2 3.2.4 if we change `maxConcurrent` to 1, or use `IO.raiseError` instead of `throw`.

I don't think this is expected behavior, and that it should work the same as in fs2 2.x. Although we should ideally not throw, exceptions can not always be avoided.

wjoel:
Fiber trace:
```
cats.effect.IOFiber@23e95b WAITING
 
cats.effect.IOFiber@210ab13f WAITING
 ├ get @ fs2.Stream.$anonfun$parEvalMapAction$2(Stream.scala:2142)
 ├ complete @ fs2.Stream.$anonfun$parEvalMapAction$2(Stream.scala:2142)
 ├ complete @ fs2.Stream.$anonfun$parEvalMapAction$2(Stream.scala:2142)
 ├ complete @ fs2.Stream.$anonfun$parEvalMapAction$2(Stream.scala:2142)
 ├ *> @ fs2.Stream.$anonfun$parEvalMapAction$2(Stream.scala:2115)
 ├ flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:162)
 ├ handleErrorWith @ fs2.Compiler$Target.handleErrorWith(Compiler.scala:160)
 ├ modify @ fs2.internal.Scope.close(Scope.scala:262)
 ├ flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:162)
[...]

PR #2826, opened 2022-02-17
Catch thrown errors in `parEvalMap`, `parEvalMapUnordered`

armanbilge:
Fixes https://github.com/typelevel/fs2/issues/2825.

### DIFF
--- a/core/shared/src/test/scala/fs2/ParEvalMapSuite.scala
+++ b/core/shared/src/test/scala/fs2/ParEvalMapSuite.scala
@@ -268,4 +268,26 @@ class ParEvalMapSuite extends Fs2Suite {
       (1 to iterations).toList.as(action).sequence_
     }
   }
+
+  group("issue-2825, Stream shouldn't hang after exceptions thrown in") {
+    test("parEvalMapUnordered") {
+      Stream
+        .eval(IO.unit)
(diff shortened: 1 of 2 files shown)


nikiforo (approved):

core/shared/src/test/scala/fs2/ParEvalMapSuite.scala
```diff
@@ -268,4 +268,26 @@ class ParEvalMapSuite extends Fs2Suite {
...
+
+  group("issue-2825, Stream shouldn't hang after exceptions thrown in") {
+    test("parEvalMapUnordered") {
+      Stream
+        .eval(IO.unit)
+        .parEvalMapUnordered(Int.MaxValue)(_ => throw new RuntimeException)
+        .compile
+        .drain
+        .attempt
+        .void
```
nikiforo, line 280: maybe some `.timeout(2.seconds)`?

armanbilge: Maybe. Munit already puts a 30 second timeout on each test, and I just followed the pattern of the test above which is checking a very similar issue 😆

armanbilge: Oh, my bad, you are right it puts a timeout. 👍

PR #2827, opened 2022-02-20
Fix Collector.supportsFactory

BalmungSan:
AFAIK this is the best practice and allows types like `BitSet` to be properly used.

### DIFF
--- a/core/shared/src/main/scala-2.13/fs2/CollectorPlatform.scala
+++ b/core/shared/src/main/scala-2.13/fs2/CollectorPlatform.scala
@@ -26,9 +26,11 @@ import scala.collection.{Factory, IterableFactory, MapFactory}
 import scala.reflect.ClassTag
 
 private[fs2] trait CollectorPlatform { self: Collector.type =>
-  implicit def supportsFactory[A, C[_], B](
-      f: Factory[A, C[B]]
-  ): Collector.Aux[A, C[B]] = make(Builder.fromFactory(f))
+  implicit def supportsCollectionFactory[A, C](f: Factory[A, C]): Collector.Aux[A, C] =
+    make(Builder.fromCollectionFactory(f))
+
+  def supportsFactory[A, C[_], B](f: Factory[A, C[B]]): Collector.Aux[A, C[B]] =
+    make(Builder.fromFactory(f))
 
   implicit def supportsIterableFactory[A, C[_]](f: IterableFactory[C]): Collector.Aux[A, C[A]] =
     make(Builder.fromIterableFactory(f))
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala-2.13/fs2/CollectorPlatform.scala
```diff
@@ -26,9 +26,11 @@ import scala.collection.{Factory, IterableFactory, MapFactory}
 import scala.reflect.ClassTag
 
 private[fs2] trait CollectorPlatform { self: Collector.type =>
-  implicit def supportsFactory[A, C[_], B](
-      f: Factory[A, C[B]]
-  ): Collector.Aux[A, C[B]] = make(Builder.fromFactory(f))
+  implicit def supportsCollectionFactory[A, C](f: Factory[A, C]): Collector.Aux[A, C] =
```
BalmungSan, line 29: If anyone has a better name I am all ears.

core/shared/src/main/scala-2.13/fs2/CollectorPlatform.scala
```diff
@@ -50,6 +52,9 @@ private[fs2] trait CollectorPlatform { self: Collector.type =>
   }
 
   private[fs2] trait BuilderPlatform { self: Collector.Builder.type =>
+    def fromCollectionFactory[A, C](f: Factory[A, C]): Builder[A, C] =
+      fromBuilder(f.newBuilder)
+
     def fromFactory[A, C[_], B](f: Factory[A, C[B]]): Builder[A, C[B]] =
```
BalmungSan, line 58: Should we deprecate this one?

core/shared/src/main/scala/fs2/Collector.scala
```diff
@@ -83,12 +83,12 @@ object Collector extends CollectorPlatform {
     def array[A: ClassTag]: Builder[A, Array[A]] =
       Chunk.newBuilder.mapResult(_.toArray)
 
-    protected def fromBuilder[A, C[_], B](
-        builder: collection.mutable.Builder[A, C[B]]
-    ): Builder[A, C[B]] =
-      new Builder[A, C[B]] {
+    protected def fromBuilder[A, C](
```
BalmungSan, line 86: I hope this change makes sense and doesn't break binary compatibility.
Although, the code compiled without it so we revert it if you prefer.

BalmungSan (commented):
I originally thought that with this change we could also _"remove"_ the other `Factory` related implicit, but it seems no. It turns out that things like `List` are not a `Factory[A, List[A]]` but rather they provide an implicit conversion to such type but the compiler doesn't apply two implicit conversions at the same time.

PR #2836, opened 2022-03-04
FS2 Suites: add an extension class for stream-specific assertion

diesalbla:
- An assertion to check if a stream emits the same outputs as   a given list, or as another stream (by compiling to List).
- An assertion to check if a stream compilation raises an error.


core/shared/src/test/scala/fs2/Fs2Suite.scala
```diff
@@ -58,3 +60,42 @@ abstract class Fs2Suite
     registered.foreach(t => munitTestsBuffer += t.withName(s"$name - ${t.name}"))
   }
 }
+
+trait StreamAssertions { self: CatsEffectSuite =>
```
mpilquist, line 64: Instead of a mixin trait, can we just define these in Fs2Suite directly?

core/shared/src/test/scala/fs2/Fs2Suite.scala
```diff
@@ -58,3 +60,42 @@ abstract class Fs2Suite
...
+  implicit class StreamAssertionsOf[A](val str: Stream[IO, A]) {
+
+    def assertEmpty(): IO[Unit] =
+      str.compile.toList.assertEquals(Nil)
+
+    def emitsOutputs(expectedOutputs: List[A]): IO[Unit] =
```
mpilquist, line 70: Style wise, can we start all these with `assert` -- e.g. `assertSameOutputAs(...)`. That way it remains clear where the assertions are being performed.

diesalbla: Renamed methods to use assert word. Also, added a method that takes a varargs.

diesalbla: Removed that method, it did not compile in Scala 3.

PR #2837, opened 2022-03-06
Pull - Translate compilation - Fix error case.

diesalbla:
If there is an error, it should be handled with the continuation of the `Translate` node, like most nodes.


mpilquist:
Can we get a unit test to cover this one?

diesalbla:
> Can we get a unit test to cover this one?

Was waiting for #2836 to be merged before writing one.

diesalbla:
Added a unit test to check interaction between `translate`, and raising error, and handling errors. I have checked that this unit test fails without the changes to the compile method, but  succeeds with them.

PR #2838, opened 2022-03-06
Publish on Java 8 to ensure maximal compatibility

sideeffffect:


armanbilge:
Thanks for the PR! Unfortunately it is not possible easily because fs2 uses some JDK 16 APIs in some places.

You may find the approach proposed here interesting:
- https://github.com/typelevel/fs2/pull/2659#issue-1013684517

sideeffffect:
I see. Does that mean that the fs2 project as a whole has dropped Java 8 compatibility? Or is > 8 required in a very small number of places while the vast majority is 8-compatible?

armanbilge:
> Or is > 8 required in a very small number of places while the vast majority is 8-compatible?

Everything should be Java 8 compatible. Many downstream projects such as http4s, skunk, etc. successfully use fs2 in Java 8 CI without issues.

 JDK 16+ is used to substitute an alternative implementation here:

https://github.com/typelevel/fs2/blob/5ca9dbf460e0add23525f3e7e4ebd02a0fe1956a/io/jvm/src/test/scala/fs2/io/net/unixsocket/UnixSocketsSuitePlatform.scala#L27-L30

sideeffffect:
Good to hear :+1: Thanks for the explanation

PR #2839, opened 2022-03-07
Fix slowdown in throttle due to improper use of stepLeg

mpilquist:
See this Scastie, provided on Discord: https://scastie.scala-lang.org/mn98/oNhMT2wZR0O42S6lN0lUMQ/8

https://discord.com/channels/632277896739946517/632310980449402880/950382134110081084

When the tick resolution argument of `throttle` is 100 millis, the output occurs at the correct speed. When the tick resolution is changed to 10 millis, the outputs get progressively slower.

### DIFF
--- a/core/shared/src/main/scala/fs2/timeseries/TimeStamped.scala
+++ b/core/shared/src/main/scala/fs2/timeseries/TimeStamped.scala
@@ -222,10 +222,9 @@ object TimeStamped {
             val tl = ticks.setHead(ticks.head.drop(1))
             val newUpto = upto + ((1000 / ticksPerSecond) * throttlingFactor).toLong.millis
             val (toOutput, stillPending) = takeUpto(pending, newUpto)
-            if (stillPending.isEmpty) {
-              Pull.output(toOutput) >> read(newUpto)(src, tl)
-            } else {
-              Pull.output(toOutput) >> awaitTick(newUpto, stillPending)(src, tl)
+            Pull.output(toOutput) >> {
+              if (stillPending.isEmpty) read(newUpto)(src, tl)
+              else awaitTick(newUpto, stillPending)(src, tl)
             }
           }
       }
--- a/core/shared/src/main/scala/fs2/timeseries/TimeStamped.scala
+++ b/core/shared/src/main/scala/fs2/timeseries/TimeStamped.scala
@@ -185,7 +185,10 @@ object TimeStamped {
     def doThrottle: Pipe2[F, TimeStamped[A], Unit, TimeStamped[A]] = {
 
       type PullFromSourceOrTicks =
-        (Stream.StepLeg[F, TimeStamped[A]], Stream.StepLeg[F, Unit]) => Pull[F, TimeStamped[A], Unit]
+        (
+            Stream.StepLeg[F, TimeStamped[A]],
+            Stream.StepLeg[F, Unit]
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/timeseries/TimeStamped.scala
```diff
@@ -196,32 +199,34 @@ object TimeStamped {
...
+            }
+          } else {
+            val tl = ticks.setHead(ticks.head.drop(1))
+            val newUpto = upto + ((1000 / ticksPerSecond) * throttlingFactor).toLong.millis
+            val (toOutput, stillPending) = takeUpto(pending, newUpto)
+            if (stillPending.isEmpty) {
+              Pull.output(toOutput) >> read(newUpto)(src, tl)
+            } else {
+              Pull.output(toOutput) >> awaitTick(newUpto, stillPending)(src, tl)
+            }
```
diesalbla, line 229:
```suggestion
            Pull.output(toOutput) >> {
              if (stillPending.isEmpty) read(newUpto)(src, tl)
              else awaitTick(newUpto, stillPending)(src, tl)
            }
```

core/shared/src/main/scala/fs2/timeseries/TimeStamped.scala
```diff
@@ -196,32 +199,34 @@ object TimeStamped {
...
-          case Some(leg) =>
-            val chunk = leg.head
-            if (chunk.isEmpty) read(upto)(leg.stream, ticks)
-            else {
-              val (toOutput, pending) = takeUpto(chunk, upto)
-              if (pending.isEmpty) Pull.output(toOutput) >> read(upto)(leg.stream, ticks)
-              else Pull.output(toOutput) >> awaitTick(upto, pending)(leg.stream, ticks)
-            }
-          case None => Pull.done
+        if (src.head.isEmpty) {
```
diesalbla, line 202: Can this chunk `head` be empty?

mpilquist: Yeah, e.g. when first chunk in source stream has a single element.

diesalbla (approved):

diesalbla:
@mpilquist It may help to add in the commit message, amended, a note about what was causing the slowdown in the old code that is not causing it in the new one. Was it an allocation and retention of not strictly necessary objects? Was this allocation growing over time running the throttle?

Do we have any way of measuring or testing for slow-downs, just as we have for memory leaks? For instance, if the cause of the slowdown was a growing number of objects being allocated and retained , would running a `throttle` in a small heap highlight it?

mpilquist:
I'm not sure actually. The problem was caused by converting the `StepLeg` to a stream on each step, which the docs explicitly say not to do: https://github.com/typelevel/fs2/blob/3f40b8c0d99e261e373ee5234340f76741d3cf3d/core/shared/src/main/scala/fs2/Stream.scala#L4837-L4843

I didn't work through exactly what was accumulating though.

We don't currently have a way to measure for this type of slowdown.

mn98:
> 

Here's my original snippet that I shared with Michael and I first discussed on discord: https://scastie.scala-lang.org/mn98/oNhMT2wZR0O42S6lN0lUMQ/11

diesalbla:
I have tried locally the given snippet. On the `main` branch, it only emits elements up to 14.  With the new branch, it emits up to 20. So this is good.

PR #2845, opened 2022-03-10
Add methods to timeout on pulls

bplommer:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2663,6 +2663,55 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+    * scala> import cats.effect.unsafe.implicits.global
+    * scala> import scala.concurrent.duration._
+    * scala> val s = Stream(1) ++ Stream.sleep_[IO](100.millis) ++ Stream(2).repeat.meteredStartImmediately[IO](200.millis)
+    * scala> s.timeoutOnPullTo(150.millis, Stream(3)).compile.toVector.unsafeRunSync()
+    * res0: Vector[Int] = Vector(1, 2, 3)
+    * }}}
+    */
+  def timeoutOnPullTo[F2[x] >: F[x]: Temporal, O2 >: O](
+      timeout: FiniteDuration,
+      onTimeout: Stream[F2, O2]
```
ChristopherDavenport, line 2689: Should this be lazy?

bplommer: I guess so yeah.

bplommer:
```suggestion
      onTimeout: => Stream[F2, O2]
```

diesalbla (approved):

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4551,6 +4600,31 @@ object Stream extends StreamLowPriority {
...
+      * scala> val s = (Stream("elem") ++ Stream.sleep_[IO](600.millis)).repeat.take(3)
+      * scala> s.pull.timeoutWith(450.millis)(Pull.output1("late!") >> _).stream.compile.toVector.unsafeRunSync()
+      * res0: Vector[String] = Vector(elem, late!, elem, late!, elem)
+      * }}}
+      */
+    def timeoutWith[O2 >: O](t: FiniteDuration)(f: Pull[F, O2, Unit] => Pull[F, O2, Unit])(implicit
+        F: Temporal[F]
+    ): Pull[F, O2, Unit] =
+      timed { timedPull =>
+        def go(timedPull: Pull.Timed[F, O]): Pull[F, O2, Unit] =
```
nikiforo, line 4619: Both `timed { timedPull =>` and `go(timedPull` have the same name. I would prefer to avoid name collision here.

bplommer: That's intentional to prevent the outer `timedPull` from being referenced inside `go`.

diesalbla:
@bplommer Could you rebase on top of current `main` branch? Just so it is up to date before merging.

bplommer:
@diesalbla I've rebased this on main, sorry for the long delay!

PR #2846, opened 2022-03-10
Set `tlJdkRelease := 8` for all modules but `io`

armanbilge:
Although @vasilmkd's release flag idea doesn't work for the io module, there's no reason we shouldn't apply it for all the other modules.

### DIFF
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -209,7 +209,7 @@ object text {
         }
 
         val isLast = r.isEmpty
-        (lastOutBuffer: Buffer).clear()
+        lastOutBuffer.clear()
 
         val outBufferSize = (decoder.averageCharsPerByte() * toDecode.size).toInt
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/text.scala
```diff
@@ -209,7 +209,7 @@ object text {
         }
 
         val isLast = r.isEmpty
-        (lastOutBuffer: Buffer).clear()
+        lastOutBuffer.clear()
```
armanbilge, line 212: This is more about making a point :P we can revert to play it safe.

vasilmkd (approved):

armanbilge:
In fact, this suggests the solution is to isolate the JDK 16 stuff into its own tiny module, which fs2-io can depend on. Then, we can compile everything but that with this release flag.

mpilquist:
Okay with this PR but not okay with factoring out a new module. Too complex.

core/shared/src/main/scala/fs2/text.scala
mpilquist: Yeah let's revert these and the `flip` ones below. When we need to target 9+ I don't want to have to fix these all over again.

PR #2848, opened 2022-03-12
Introduce aliases for parameterized `Pull` types

bplommer:
Introduces type aliases and a new extension method to better (I hope) convey an intuition for how the differently parameterized `Pull` types fit together. It's a big diff but most of it is just changing code to use the new aliases. The core of the PR is in `Pull.Scala` and is as follows:

```scala
  /** A [[Pull]] instance that has has no unprocessed resource and can be converted to
    * a [[Stream[F, O]]] using [[StreamPullOps.stream]].
    */
  type ToStream[+F[_], +O] = Pull[F, O, Unit]

  /** A [[Pull]] instance with no output. Because its output type is [[INothing]] it can
    * be transformed via `flatMap` into Pulls with, and ultimately produce a Stream of, any
    * output type.
    */
  type From[+F[_], +R] = Pull[F, INothing, R]

  /** A [[Pull]] that is transforming a [[Stream[F, O]]] into a new [[Stream[F, O]]] but
    * may still have further elements from the source stream to process. Can be converted
    * to [[Pull.ToStream]] in three ways:
    *
    * - process the input stream further via `flatMap`
    * - echo remaining elements using `echo`
    * - drop remaining elements using `void`
    */
  type TransformStream[+F[_], +O] = Pull[F, O, Option[Stream[F, O]]]

  implicit final class StreamTransformPullOps[F[_], O](
      private val self: Pull.TransformStream[F, O]
  ) {
    /* Echoes the resource of this [[Pull]] to the output
     */
    def echo: Pull.ToStream[F, O] =
      self.flatMap(_.fold(Pull.done: Pull.ToStream[F, O])(_.pull.echo))
  }
```

### DIFF
--- a/core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala
+++ b/core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala
@@ -156,7 +156,7 @@ class MemoryLeakSpec extends FunSuite {
 
   leakTest("repeatEval") {
     def id[F[_], A]: Pipe[F, A, A] = {
-      def go(s: Stream[F, A]): Pull[F, A, Unit] =
+      def go(s: Stream[F, A]): Pull.From[F, A] =
         s.pull.uncons1.flatMap {
           case Some((h, t)) => Pull.output1(h) >> go(t); case None => Pull.done
         }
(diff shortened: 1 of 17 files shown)


diesalbla:
@bplommer There have been some proposals to add type aliases for some cases of `Pull` type arguments. Until now the practice has been to avoid them. 

Having said that, one can suggest other metaphors that could be more illustrative. As an example, one can suggest the following ones:

- For pulls where the result type is `Unit`, it may be easier to called it `Closed`, as in closed-ended pull, as opposed to an _open-ended_ pull, one that returns a non-unit result with which a follow-up operation is needed.
- For the case where the output type is `Nothing`, the `From` is a bit confusing since that pull may be composed of several operations, so it would not be a "source" in itself. Perhaps a better name for a stream without output is `Dry`. It short-of makes sense for a stream to be dry after draining it.

PR #2851, opened 2022-03-17
Add `evalFlatten` methods for stream of effects

bplommer:


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4117,6 +4114,25 @@ object Stream extends StreamLowPriority {
...
+    /** Evaluates up to `maxConcurrent` inner effects concurrently, emitting
+      * the results in order.
+      */
+    def parEvalFlatten(
+        maxConcurrent: Int
+    )(implicit F: Concurrent[F]) = self.parEvalMap(maxConcurrent)(identity)
+
+    /** Evaluates all inner effects concurrently, emitting the results in order.
+      */
+    def parEvalFlattenUnbounded(implicit F: Concurrent[F]) = self.parEvalMapUnbounded(identity)
```
diesalbla, line 4133: So, just to clarify something: compiling this stream would immediately start a background action that would start launching _all_ the actions in this stream. Without a bound in the internal queue or the semaphore inside `parEvalMapAction`, and with `this` stream being compiled in the background, nothing allows the user to control, based on the outputs of the resulting stream, if any actions from the source are delayed... Is that a desirable mode of operation? 

(BTW, this is missing the return type)

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4117,6 +4114,25 @@ object Stream extends StreamLowPriority {
...
+    /** Sequences the inner effects into the stream. */
+    def evalFlatten: Stream[F, O] =
+      self.evalMap(identity)
+
+    /** Evaluates up to `maxConcurrent` inner effects concurrently, emitting
+      * the results in order.
+      */
+    def parEvalFlatten(
+        maxConcurrent: Int
+    )(implicit F: Concurrent[F]) = self.parEvalMap(maxConcurrent)(identity)
```
diesalbla, line 4129: Missing return type.

core/shared/src/main/scala/fs2/Stream.scala
nikiforo: I think this method is as unsafe as other `*Unbounded` methods. All concurrent methods: `parJoin`, `parEvalMap` and other - are effectively push based and the only thing that guards users from launching too many computations is `semaphore`. In case of `*Unbounded` semaphore isn't working, so they all should be unsafe in that regard.

However, I see you make a distinction between `parJoinUnbounded` and `parEvalMapUnbounded` in how they block on `Channel`. I don't think this functionality is intended as a protection and should be relied on.

Personally, when I use `par*Unbounded` methods I **assume** that the stream won't be parallel enough to overwhelm the consumer and choose not to restrict the level of parallelism. Otherwise, I would have to widen interfaces(functions or constructors) with the configuration of parallelism. Because when you really want to restrict parallelism, you should make that restriction configurable.

diesalbla: > Personally, when I use par*Unbounded methods I assume that the stream won't be parallel enough to overwhelm the consumer and choose not to bother restricting the level of parallelism

There are two mixed concerns here: parallelism, and buffering vs backpressuring. 

In `parJoinUnbounded` you take an outer stream of sources of data, and it launches all sources to pull, and push into the queue, in parallel. That outer stream may not end and keep incorporating new sources; so `parJoinUnbounded` does not limit parallelism. However, what it does restrict, by means of that output queue, is how many items are pulled _from each source_ into memory. 

The problem with this new combinator is that it would prefetch and load all data from its source into local memory, without any feedback from the consumer to stop it. Thus, a short slowdown in the consumer would cause this combinator to accrue a lot of data, and thus crash the program. A basic reliability guideline is to avoid infinite buffers.

> However, I see you make a distinction between parJoinUnbounded and parEvalMapUnbounded in how they block on Channel. I don't think this functionality is intended as a protection and should be relied on.

Given that back-pressuring and laziness is an essential part of fs2 streams, building that check and balance into the pipeline seems to me a crucial part of that combinator.

As an aside, there are FS2 combinators that do fetch a lot of data, like `prefetchAll`, but those are some legacies. Also, some choices of parameters in other combinators can cause trouble, but those cannot be avoided.

bplommer: Sorry for the radio silence on this!

Yes, the use case I have in mind for this is when backpressure is provided upstream - the case I have in mind (which occurs for example in [fs2-kafka](https://github.com/fd4s/fs2-kafka/blob/f503479396150f694e6e6bafc41e15c07777a7d9/modules/core/src/main/scala/fs2/kafka/KafkaProducer.scala#L33-L59)) is where you have an operation that returns `F[F[Result]]`, where the outer `F` does a backpressured enqueue operation and then returns the inner `F` which is a handle for waiting on the result - so there is no need for backpressure on the inner `F`.

This should definitely have documentation and an example though, so I'm going to remove `parEvalFlattenUnbounded` from this PR and maybe put it in a separate one later.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4117,6 +4114,21 @@ object Stream extends StreamLowPriority {
       parJoin(Int.MaxValue)
   }
 
+  /** Provides syntax for a stream of `F` effects. */
+  implicit class StreamFOps[F[_], O](private val self: Stream[F, F[O]]) {
```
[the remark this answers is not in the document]
kubukoz: good to know @Daenyth, I'd never think it could blow the memory. Would a suspend help resolve that? (at least to reduce the overhead of the non-trivial ones)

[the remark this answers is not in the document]
Daenyth: Probably not because the issue is the memory used by having large `Chunk[IO[A]]`

It's also possibly my codebase just was doing something really silly and that's why it cost so much memory for that structure

PR #2856, opened 2022-03-26
Reimplemented `Channel` in terms of `Queue`

### LINKED ISSUE #2852: Reimplement `Channel` in terms of `Queue`
It's very unclear that `Channel` is actually meaningfully faster than an implementation based directly on Cats Effect's `Queue`. More importantly, since `Queue` is now being optimized substantially, `Channel` should probably lean on that primitive rather than its own data structures so as to inherit the performance gains. This will be particularly dramatic when typelevel/cats-effect#2890 is completed (which will be during the 3.4.x series, and possibly 3.4.0).

A simple strawman implementation of `Channel` (which is probably at least *close* to functional) can be found in this gist: https://gist.github.com/djspiewak/847532614a166af9f811e972d6f2c36f

djspiewak:
~~Do not merge. Depends on a snapshot of Cats Effect 3.4 (well, technically could be implemented without the snapshot, but it's better to wait for `tryTakeN`)~~. The snapshot in question includes the (still unmerged) improvements to `Queue` on any `Async[F]` (typelevel/cats-effect#2885 and typelevel/cats-effect#2914), but critically does not include an implementation of typelevel/cats-effect#2890, which is unquestionably the main bottleneck here:

Before:

```
[info] Benchmark                              (size)   Mode  Cnt     Score     Error  Units
[info] ChannelBenchmark.sendPull                  64  thrpt   10  7348.611 ± 207.064  ops/s
[info] ChannelBenchmark.sendPull                1024  thrpt   10  3850.531 ±  10.141  ops/s
[info] ChannelBenchmark.sendPull               16384  thrpt   10   372.989 ±   1.049  ops/s
[info] ChannelBenchmark.sendPullPar8              64  thrpt   10  9122.642 ±  49.265  ops/s
[info] ChannelBenchmark.sendPullPar8            1024  thrpt   10  3558.006 ±   5.951  ops/s
[info] ChannelBenchmark.sendPullPar8           16384  thrpt   10   254.749 ±   0.650  ops/s
[info] ChannelBenchmark.sendPullParUnlimited      64  thrpt   10  9367.322 ±  25.509  ops/s
[info] ChannelBenchmark.sendPullParUnlimited    1024  thrpt   10   897.318 ±  17.716  ops/s
[info] ChannelBenchmark.sendPullParUnlimited   16384  thrpt   10    39.973 ±   0.367  ops/s
```

After:

```
[info] Benchmark                              (size)   Mode  Cnt     Score    Error  Units
[info] ChannelBenchmark.sendPull                  64  thrpt   10  5383.059 ± 42.282  ops/s
[info] ChannelBenchmark.sendPull                1024  thrpt   10  2564.141 ±  5.962  ops/s
[info] ChannelBenchmark.sendPull               16384  thrpt   10   254.210 ±  0.407  ops/s
[info] ChannelBenchmark.sendPullPar8              64  thrpt   10  4840.209 ± 21.118  ops/s
[info] ChannelBenchmark.sendPullPar8            1024  thrpt   10  2427.766 ±  2.502  ops/s
[info] ChannelBenchmark.sendPullPar8           16384  thrpt   10   179.873 ±  0.524  ops/s
[info] ChannelBenchmark.sendPullParUnlimited      64  thrpt   10  7256.407 ± 29.061  ops/s
[info] ChannelBenchmark.sendPullParUnlimited    1024  thrpt   10  1334.681 ±  6.618  ops/s
[info] ChannelBenchmark.sendPullParUnlimited   16384  thrpt   10    77.232 ±  1.446  ops/s
```

To me, the fact that this is even *close* to the hand-rolled version is pretty cool, and a slightly less naive `stream` would probably put it almost on-par even without the optimized `tryTakeN`. Once we get the latter though this should be head and shoulders above the prior implementation.

Closes #2852


core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -105,149 +105,59 @@ sealed trait Channel[F[_], A] {
...
-            F.deferred[Unit].flatMap { waiting =>
-              state
-                .modify { state =>
-                  if (shouldEmit(state)) (empty(state.closed), state)
-                  else (state.copy(waiting = waiting.some), state)
+        def stream: Stream[F, A] = {
+          val takeN: F[Chunk[A]] =
+            q.tryTakeN(None).flatMap {
+              case None =>
+                val fallback = Spawn[F].race(q.take, closedR.get).map {
```
SystemFw, line 145: I don't think this is correct, I think you can lose elements here on a race:
producer `A` successfully enqueues, producer B closes, and now the two notifications for "is closed" and "there are new elements" are racing, if "is closed arrives first", elements are lost

djspiewak: Ah! Yes this is indeed an issue. Will ponder…

djspiewak: Thought about it some more. This is the adjacent critical regions problem in a different disguise, and the solution is the same: the first `flatMap` is uninterruptible.

`take` on all `Queue` implementations must have the invariant that, if it is canceled, it either preserves the contents of the queue *or* fully produces the `A` it took. Thus, we should be able to `flatMap` with an `uncancelable` region which tosses the `A` somewhere that we can get it. Or, more likely, we use something a bit more powerful than `race`.

djspiewak: I think I fixed it. Interestingly, this appears to be the very first (valid) use of `racePair` I've ever seen.

SystemFw: > his appears to be the very first (valid) use of racePair I've ever seen.

https://github.com/SystemFw/upperbound/blob/466aacdabd3297aa86221f7adae49627f58fdf7d/modules/core/src/main/scala/internal/Task.scala#L44 ;)

I'll ponder on the implementation later

djspiewak: Aha! Your linked use-case is almost identical to this one.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -105,149 +105,72 @@ sealed trait Channel[F[_], A] {
...
+            q.tryTakeN(None).flatMap {
+              case None =>
+                val fallback = MonadCancel[F].uncancelable { poll =>
+                  poll(Spawn[F].racePair(q.take, closedR.get)).flatMap {
+                    case Left((oca, fiber)) =>
+                      oca.embedNever.flatMap(a => fiber.cancel.as(Chunk.singleton(a)))
+
+                    case Right((fiber, ocb)) =>
+                      ocb.embedNever.flatMap { _ =>
+                        (fiber.cancel *> fiber.join).flatMap { oca =>
```
SystemFw, line 152: I think this works if `fiber.cancel` arrives after `take` has successfully produced the elements, in which case it will be a no-op, and then the subsequent `join` gets you the elements. However imagine this:
`cancel` hits after the elements have arrived, but before `take` can complete. Due to the take invariant, the elements aren't lost and go back to the queue, but this code is still buggy: we will emit an empty chunk from here, and terminate the stream.

 We can try and do an additional `tryTake` in that case, but I haven't thought if there are further corner cases there. I think the take away here is that separating the state makes things very very complex to reason about, so before including this I'd want to see extremely convincing performance results :)

djspiewak: > Due to the take invariant, the elements aren't lost and go back to the queue, but this code is still buggy: we will emit an empty chunk from here, and terminate the stream.

Ah yes, I actually thought about this but didn't implement the fix. We basically need a robust bit of signalling than the size of the `Chunk`. When we hit this case, we need to basically retry the `tryTake`.

> so before including this I'd want to see extremely convincing performance results :)

Agreed, though there's also value in reducing the maintenance burden and getting the benefit of *future* improvements in `Queue`.

> Btw, I'm not a hundred percent sure that invariant actually holds in the current version of queue, there's a race between deferred.complete and cancel a deferred.get, and get being cancelled cannot communicate to complete to put the elements back...

That's a bug then and we should fix it. We *really* need that invariant in general, and not just in this use of the queue.

SystemFw: Yeah I'd rather merge if possible, it's just that this implementation strategy seems both unavoidable and risky, as its based essentially on recovering from interruption, and that's how I originally started with groupWithin and then had to change


As per the take guarantee, I'm not sure it's achievable without a locking implementation of queue, think about the two ops on Deferred racing _after_ they both have relinquished control of the queue state. Maybe it can be done with a more complex state machine in the Ref to encode the extra communication, but then how easily can that be ported to the low level implementation of queue?

djspiewak: I *think* that the invariant should be achievable if we correctly manage our `uncancelable` regions. I'm not sure if we're doing that right now though (in fact, I'm certain that in some cases, we're *not* doing that in the `Async` queue). Need to stare at it a bit more. It feels like `Queue` itself would be fundamentally leaky if we can't achieve that invariant, though.

djspiewak: Digging in a little deeper still, I don't think that `async` *itself* preserves this property. In fact, there's a specific place in the state machine where it catches the race condition between cancelation and completion... and resolves it in favor of cancelation. I need to think about this more, actually, because I honestly wonder if we're in a place where it's impossible to avoid data loss in all cases.

djspiewak: Spamming this single thread... I think we can do this but it has to be encoded more like what the `Async` queue and the old `Channel` implementation does. Basically, the `Deferred` *must* contain a `Unit` because otherwise it simply isn't guaranteed to not lose data. Alternatively, we can probably fix `Deferred` itself.

SystemFw: The Deferred of Unit is where my thoughts were going as well before I had to step afk

SystemFw: >  If the taker.get is canceled after it has been fulfilled

This is not the problem though. The enqueue method will spin on a Ref whose state contains an unfulfilled Deferred, select it for delivery, and commit on the Ref. At this point, _Queue is no longer at play_. You now have a race between `deferred.complete` and cancelling a `deferred.get`, and it's perfectly possible that cancelling will win, i.e. it `get` gets cancelled _before_ being fulfilled, at which point `complete` does fulfill, but the element has been out of the queue for while, and can be put back in A) because `complete` doesn't know that `get` has been canceled, and `B` because even if it knew the order of elements in the queue might have changed (i..e another element might have been delivered) so you violate FIFO: you'd need to lock the Queue until you know the element has been successfully delivered, that's what I mean by wanting at least once processing.

An alternative would be to alter `take` instead to only allow cancellation when the relative Deferred is still in the state. It gets tricky cause you can't just return `Deferred.get`, you have to start dancing with extra fibers so that when the cancelation signal hits you first check that the Deferred is indeed still in the queue and hasn't been removed by enqueue about to complete it. But even then I think there is a sliver where after you've checked, found it the queue, then cancel, enqueue will complete, resulting in data loss again

djspiewak: Need to reread and digest more, but I'm thinking of something a little simpler to get at this invariant. In particular, being very careful about where cancelation becomes visible. Basically, from the moment the element is committed to no longer being in the queue all the way to the end of `take`, cancelation must not be observed. If you preserve this property, then it's possible to have the "all or nothing" semantic in a useful way due to adjacency.

This seems actually doable. Hard but doable.

SystemFw: That's what the second paragraph is exploring, but the issue is the element is no longer in the queue as part of a `Ref.modify`, where you can't execute effects, and so you cannot have an effect transactionally with the change in state, whereas you need to mark the Deferred as now uncancelable transactionally with the removal of an element in the queue

djspiewak: I think that's only true though if we _don't_ restructure the Concurrent Queue to use `Deferred[Unit]`. The fact that there are multiple paths for values is, I think, the root of this issue.

SystemFw: > I think that's only true though if we don't restructure the Concurrent Queue to use Deferred[Unit]. The fact that there are multiple paths for values is, I think, the root of this issue.

I agree, but `Deferred[Unit]` brings its own challenges, that I alluded to in:

>  he Deferred of Unit is where my thoughts were going as well before I had to step afk, so you have a consistent place for all the state, but it's not trivial to do in the multiple consumers case,you have to mark elements as "takeable" for a given consumer to preserve FIFO.

In particular consider the multiple consumers case, and the preservation of FIFO: consumer 1 takes on empty queue, deferred gets enqueued, consumer 2 takes on empty queue, deferred gets enqueued. At this point, two elements are enqueued immediately after one another, which means the two Deferred are completed, and the consumers awake, at which point they _race_ on the Ref to take the element, which means that they can now break FIFO. You can think about communicating to each consumer the position of  the element of the queue they have to take, but this can be broken because you can communicate position 1 to consumer 1, position 2 to consumer 2, then consumer 1 gets cancelled, at which point consumer 2 gets position 2, whereas it should have switched to taking position 1. The reason the pattern works so well in the current Channel is because there is a single consumer, so "there are changes" is enough signal for it to work, but it gets very tricky in Queue

djspiewak: IMO whenever you have multiple consumers racing for elements, FIFO goes out the window just a bit. It also becomes a lot less relevant because of the racing, so the lack of strong ordering across consumers is effectively unobservable.

SystemFw: This is not the case here though, look at the example again, with Deferred of Unit two consumers who subscribed at wildly different times (say three seconds apart) end up racing when *production* is fast because after they wake up they have to spin on the Ref again, which would in a very observable bug

djspiewak: I get what you mean; I was being imprecise about what I was referencing. Any number of suspended consumers, regardless of when they subscribe, race *at publication time* on data. Notably, this isn't a fairness issue, since suspended consumers will still be awakened in FIFO order, but it does mean that the multiplexing might be in a different order.

You're basically describing a situation where data **A** and **B** came into the queue *from two publishers* almost simultaneously, but with **A** still arriving first. Meanwhile, you have consumers **1** and **2** who subscribed in that order, and consumer **1** might get **B** instead of **A**. I really don't see why this would cause a serious or even noticeable bug, particularly since other mpmc implementations have analogous issues (e.g. the jctools queue has this same problem). From a parallelism perspective, there really isn't a strong ordering between **A** and **B** since they are arriving on separate publishers so closely interspersed, and thus which consumer gets which one is somewhat nondeterministic.

I would be more concerned if **A** and **B** could arrive on the *same* publisher and the ordering could be mixed up on the consumers, but I don't believe that is possible.

SystemFw: > I would be more concerned if A and B could arrive on the same publisher and the ordering could be mixed up on the consumers, but I don't believe that is possible.

That's exactly what I'm saying will happen with `Deferred[F, Unit]` :)

With `Deferred[F, Unit]` , a publisher wakes a consumer up, then the consumer has to spin again on the Ref to get the value. Crucially, `deferred.complete` terminates well before this happens: `complete` terminates after calling the callback, at which point `get` will awake, and recur back into a `modify`on `Ref`.

Things have this general shape:

```scala
enqueue = transactionally { 
    put elem in elem queue
    take first Deferred[Unit] from listeners queue
} then complete Deferred[Unit]

get = transactionally {
  full => take elem
  empty => enqueue Deferred[Unit]
  } then wait on Deferred[Unit], and recur on get
```

So let's see what happens: Consumer 1 takes on the empty queue, and enqueues a deferred, Consumer 2 (even after some time) takes on the empty queue, and enqueues a deferred, and after some time a single Producer produces two elements, let's break that down:
1) first transaction on enqueue, Consumer 1's deferred gets dequeued
2) Consumer 1 deferred gets completed, Producer can keep going
... At this point Consumer 1's recurring on the Ref is racing with  Producer enqueing second element, imagine Producer wins...
3) Producer's second transaction on enqueue, Consumer 2's deferred gets dequeued
4) Consumer's 2 deferred gets completed,
... now Consumer's 2 recurs on the Ref

At this point both Consumer 1 and Consumer 2 are racing to `modify` on the Ref, and if Consumer 2 wins, FIFO will be broken, even though Consumer 1 and Consumer 2 called dequeue at sufficiently distant times, and even though there is a single producer operating strictly synchronously, which to me sounds exactly like a race condition bug in the implementation.

What you seem to want is `deferred.complete` effectively blocking until the element it's signalling is successfully dequeued, which once again points to a blocking implementation.

SystemFw: Also note that the problem I'm describing does not happen in the current design, because `Deferred[F, A]` means that you transactionally decide which element goes where, so FIFO is respected: by the time the first produce transaction finishes and the second produce transaction starts, the element is already going to the correct destination. There is an unavoidable (I think) tradeoff here: the `Deferred[F, A]` does FIFO correctly because the elements is considered delivered in `modify`, and by the same token if things get cancelled later, the element is "lost" (because it's already considered delivered, there is no extra acking). `Deferred[F, Unit]`  is more lenient with cancellation because the element isn't considered delivered after produce's `modify`, but that's also the reason why it breaks FIFO. I think the current queue implementation has the right tradeoff, and what's unreasonable is expecting to be able to race a `take` and retry without losing elements, by wishing more transactionality on it than possible, which again is trying to cover the fact that one lost transactionality by splitting the concurrent state in two (the `closed` signal, and the `queue`)

djspiewak: Ah I see what you're getting at.

Okay, real talk: how much would we really care if that were the case? Because not being able to implement something like Channel on top of Queue feels much worse. Losing data without any opportunity to do it "right" definitely feels worse.

SystemFw: > Okay, real talk: how much would we really care if that were the case? 

I mean, we're talking about breaking a primary characteristic of what defines a queue, in a case with no contention. I bet if someone reported that right now, you'd treat it as a bug

> Because not being able to implement something like Channel on top of Queue feels much worse.

Well, I can agree that not having STM without its performance cost is not as nice as having it (because that's what we're talking about here), but it's also the sad reality of things. We could try a hybrid implementation strategy that uses both a closed flag and wraps elems in `Option` and see if that makes things behave more reasonably without imposing a perf cost, but that's hard to say because we don't have have convincing perf numbers yet, although perf numbers with relaxed guarantees are also hard to discuss 

> Losing data without any opportunity to do it "right" definitely feels worse.

This is the bit where, regardless of the decision for Channel, I think the expectation is unrealistic. I see it as the same as the operation processing the element of `take` failing immediately: i.e. `Queue` has at most once semantics, and really the implementation strategy of `take-interrupt-tryTake` wants at least once

djspiewak:
Working on this PR uncovered a bug in Cats Effect 3.4.0-RC1 (fixed in typelevel/cats-effect#3180), so this PR is based on a snapshot. If it's accepted, we'll do an RC2.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +117,93 @@ sealed trait Channel[F[_], A] {
...
-          var values = tail
-          while (i >= 0) {
-            if (values.isEmpty) values = init
-            arr(i) = values.head
-            values = values.tail
-            i -= 1
-          }
-          Chunk.array(arr).asInstanceOf[Chunk[A]]
+          // you can do this more efficiently, just proves a point
+          Stream.eval(takeN).repeat.takeWhile(!_.isEmpty).flatMap(Stream.chunk(_))
```
armanbilge, line 203: lol?

armanbilge:
```suggestion
          // you can do this more efficiently, just proves a point
          Stream.eval(takeN).repeat.takeWhile(!_.isEmpty).unchunks
```

armanbilge:
```
Error: Exception in thread "io-compute-4" java.lang.OutOfMemoryError: Requested array size exceeds VM limit
	at cats.effect.std.Queue$UnsafeBounded.<init>(Queue.scala:770)
	at cats.effect.std.Queue$BoundedAsyncQueue.<init>(Queue.scala:412)
	at cats.effect.std.Queue$.boundedForAsync$$anonfun$1(Queue.scala:96)
	at cats.effect.std.Queue$$$Lambda$1222/0x0000000800f5d398.apply(Unknown Source)
	at cats.effect.IOFiber.runLoop(IOFiber.scala:401)
	at cats.effect.IOFiber.execR(IOFiber.scala:1309)
	at cats.effect.IOFiber.run(IOFiber.scala:118)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:580)
```

djspiewak:
Wow. How many elements were you trying to stick into that queue??

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +117,93 @@ sealed trait Channel[F[_], A] {
...
       }
     }
-  }
-
-  // allocate once
-  @inline private final def closed[A]: Either[Closed, A] = _closed
-  private[this] final val _closed: Either[Closed, Nothing] = Left(Closed)
-  private final val rightUnit: Either[Closed, Unit] = Right(())
-  private final val rightTrue: Either[Closed, Boolean] = Right(true)
-  private final val rightFalse: Either[Closed, Boolean] = Right(false)
```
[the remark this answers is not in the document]
djspiewak: Can bring them back. I think I chopped them out a while ago.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: Oh nice. Is that new?

armanbilge:
So it seems like anyone who was using `bounded` with a large number and expecting it to be allocated lazily is in for a bad time 🙃

core/shared/src/main/scala/fs2/concurrent/Channel.scala
armanbilge: It was one of my first PRs to FS2 :)
- https://github.com/typelevel/fs2/pull/2506

djspiewak:
> So it seems like anyone who was using bounded with a large number and expecting it to be allocated lazily is in for a bad time 🙃

Yes. They shouldn't do that. :-P I mean, we can special case large bounds, but genuinely this just isn't a case which should be supported. It only worked before by coincidence.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: I've honestly wanted that convenience function for so long. I spend my whole life doing it.

djspiewak:
This is being caused by typelevel/cats-effect #3187

djspiewak:
Getting better!

Not quite better across the board yet, but doing really well. Because of how things are optimized, I'm pretty sure that the new implementation gets better the faster the producers run relative to the consumer, though the numbers don't seem to intuitively bear that out. Still fiddling with things, and the upstream `Queue` has plenty of room for improvement (especially `unbounded`).

## Before

```
[info] Benchmark                              (size)   Mode  Cnt     Score     Error  Units
[info] ChannelBenchmark.sendPull                  64  thrpt   10  7735.063 ±  33.841  ops/s
[info] ChannelBenchmark.sendPull                1024  thrpt   10  3760.871 ±   9.716  ops/s
[info] ChannelBenchmark.sendPull               16384  thrpt   10   359.674 ±   1.309  ops/s
[info] ChannelBenchmark.sendPullPar8              64  thrpt   10  8533.059 ± 182.194  ops/s
[info] ChannelBenchmark.sendPullPar8            1024  thrpt   10  3464.678 ±   8.370  ops/s
[info] ChannelBenchmark.sendPullPar8           16384  thrpt   10   247.827 ±   2.399  ops/s
[info] ChannelBenchmark.sendPullParUnlimited      64  thrpt   10  8921.664 ±  27.248  ops/s
[info] ChannelBenchmark.sendPullParUnlimited    1024  thrpt   10  1002.150 ±  26.415  ops/s
[info] ChannelBenchmark.sendPullParUnlimited   16384  thrpt   10    39.949 ±   0.363  ops/s
```

## After

```
[info] Benchmark                              (size)   Mode  Cnt     Score     Error  Units
[info] ChannelBenchmark.sendPull                  64  thrpt   10  8092.902 ± 112.613  ops/s
[info] ChannelBenchmark.sendPull                1024  thrpt   10  1934.434 ±  13.293  ops/s
[info] ChannelBenchmark.sendPull               16384  thrpt   10   144.031 ±   2.505  ops/s
[info] ChannelBenchmark.sendPullPar8              64  thrpt   10  7045.267 ±  37.947  ops/s
[info] ChannelBenchmark.sendPullPar8            1024  thrpt   10  2515.080 ±  13.203  ops/s
[info] ChannelBenchmark.sendPullPar8           16384  thrpt   10   225.750 ±   1.203  ops/s
[info] ChannelBenchmark.sendPullParUnlimited      64  thrpt   10  7241.213 ± 106.143  ops/s
[info] ChannelBenchmark.sendPullParUnlimited    1024  thrpt   10  1468.275 ±  20.139  ops/s
[info] ChannelBenchmark.sendPullParUnlimited   16384  thrpt   10    76.752 ±  10.437  ops/s
```

djspiewak:
…and a little better:

```
[info] Benchmark                              (size)   Mode  Cnt     Score     Error  Units
[info] ChannelBenchmark.sendPull                  64  thrpt   10  8847.457 ±  77.061  ops/s
[info] ChannelBenchmark.sendPull                1024  thrpt   10  1870.886 ±  16.247  ops/s
[info] ChannelBenchmark.sendPull               16384  thrpt   10   132.900 ±   0.244  ops/s
[info] ChannelBenchmark.sendPullPar8              64  thrpt   10  8171.220 ± 166.752  ops/s
[info] ChannelBenchmark.sendPullPar8            1024  thrpt   10  2673.881 ±  26.653  ops/s
[info] ChannelBenchmark.sendPullPar8           16384  thrpt   10   218.006 ±   1.377  ops/s
[info] ChannelBenchmark.sendPullParUnlimited      64  thrpt   10  8604.675 ± 247.214  ops/s
[info] ChannelBenchmark.sendPullParUnlimited    1024  thrpt   10  1415.921 ±  18.295  ops/s
[info] ChannelBenchmark.sendPullParUnlimited   16384  thrpt   10    79.396 ±   2.864  ops/s
```

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
+  // used as a marker to wake up q.take when the channel is closed
+  private[this] val Sentinel = new AnyRef
+
+  private[this] val LeftClosed: Either[Channel.Closed, Unit] = Left(Channel.Closed)
+  private[this] val RightUnit: Either[Channel.Closed, Unit] = Right(())
+
+  // technically this should be A | Sentinel.type
+  // the queue will consist of exclusively As until we shut down, when there will be one Sentinel
+  private[this] def impl[F[_]: Concurrent, A](q: Queue[F, AnyRef]): F[Channel[F, A]] =
+    (Concurrent[F].ref(0), Concurrent[F].ref(false), Concurrent[F].deferred[Unit]).mapN {
```
SystemFw, line 144: Splitting the state across non synchronised units makes this a lot harder to reason about. Obviously, there is some unavoidable splitting between the Queue and the rest, but the tradeoffs for splitting the two Refs are less obvious (the Deferred is trivial, I'm ignoring it).

Now, the correctness of the handling of the two Refs seems fine, although I don't have full confidence yet, since `closed` is monotonic and acts as a linearisation point, but if `closed` and `latches` were unified things would be easier.

An argument for splitting the two Refs is that we can then exploit the upcoming Ref specialisation for Integer and Boolean, although ideally we'd get some quantitative confirmation that there are worthy gains to be had there, especially because with two Refs we often double the accesses

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
+              .drain
+
+          // setting the flag means we won't accept any more sends
+          val close: F[Either[Channel.Closed, Unit]] =
+            closedR.getAndSet(true).flatMap { b =>
+              if (b) {
+                LeftClosedF
+              } else {
+                leasesR.get.flatMap { leases =>
+                  if (leases <= 0)
```
SystemFw, line 165: When can this be negative?

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
-                (s, Channel.closed[Unit].pure[F])
-
-              case State(values, size, waiting, producers, closed @ false) =>
-                (
-                  State(values, size, None, producers, true),
-                  notifyStream(waiting).as(rightUnit) <* signalClosure
-                )
+          val isClosed: F[Boolean] = closedR.get
+
+          // there are four states to worry about: open, closing, draining, quiesced
```
SystemFw, line 175: State machine makes sense to me. It's a bit implicit in the various `F[Boolean]`, but reifying it a bit more depends on the decision on splitting Refs

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
+          // in the second state, we have outstanding blocked sends
+          // in the third state we have data in the queue but no sends
+          // in the fourth state we are completely drained and can shut down the stream
+          private[this] val isQuiesced: F[Boolean] =
+            isClosed.flatMap { b =>
+              if (b) {
+                leasesR.get.flatMap { leases =>
+                  if (leases <= 0)
```
SystemFw, line 183: When can this be negative?

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
+                    q.size.map(_ <= 0)
```
SystemFw, line 184: When can this be negative?

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
-                        tailValues += value
-                        unblock = unblock <* producer.complete(())
+          def send(a: A): F[Either[Channel.Closed, Unit]] = {
+            // we track the outstanding blocked offers so we can distinguish closing from draining
+            // the very last blocked send, when closed, is responsible for triggering the sentinel
+            def permit[E](fe: F[E]): F[E] =
+              MonadCancel[F].uncancelable { poll =>
+                (leasesR.update(_ + 1) *> poll(fe)).guarantee {
+                  leasesR.updateAndGet(_ - 1).flatMap { leases =>
+                    if (leases <= 0) {
```
SystemFw, line 200: When can this be negative?

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
-                  State(values, size, None, producers, true),
-                  notifyStream(waiting).as(rightUnit) <* signalClosure
-                )
+          val isClosed: F[Boolean] = closedR.get
+
+          // there are four states to worry about: open, closing, draining, quiesced
+          // in the second state, we have outstanding blocked sends
+          // in the third state we have data in the queue but no sends
+          // in the fourth state we are completely drained and can shut down the stream
+          private[this] val isQuiesced: F[Boolean] =
```
SystemFw, line 179: Should we make this `uncancelable`?

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
+                  }
                 }
-                .uncancelable
-            }
-          }.flatten
+              }
 
-        def notifyStream(waitForChanges: Option[Deferred[F, Unit]]) =
-          waitForChanges.traverse(_.complete(()))
+            isClosed.ifM(
```
SystemFw, line 216: This whole function I feel is pretty complex, _and_ has a ton of access to `closed`, just because we split the Refs

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
-        def signalClosure = closedGate.complete(())
+          val stream: Stream[F, A] = {
+            val takeN: F[Chunk[A]] =
+              q.tryTakeN(None).flatMap {
+                case Nil =>
+                  // if we land here, it either means we're consuming faster than producing
+                  // or it means we're actually closed and we need to shut down
+                  // this is the unhappy path either way
+
+                  val fallback = q.take.flatMap { a =>
```
SystemFw, line 238: Imagine this scenario: queue empty for a while, then a bunch of producers enqueue at once. The old channel has a chance of  chunking up all the new updates up, this code doesn't. It will probably get picked up in the next iteration, but it doesn't seem hard to fix either

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
+                    else
+                      fallback
+                  }
+
+                case as =>
+                  // this is the happy path: we were able to take a chunk
+                  // meaning we're producing as fast or faster than we're consuming
+
+                  isClosed.flatMap { b =>
+                    val back = if (b) {
```
SystemFw, line 260: the detection and stripping of the Sentinel seems a bit obscure, what's the rationale? Don't we have the guarantee that the Sentinel, if present, is the last element of this Chunk?

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -116,174 +118,179 @@ sealed trait Channel[F[_], A] {
...
+                    } else {
+                      Chunk.seq(as).pure[F]
+                    }
+
+                    back.asInstanceOf[F[Chunk[A]]]
+                  }
+              }
 
-        @inline private def shouldEmit(s: State) = s.values.nonEmpty || s.producers.nonEmpty
+            // we will emit non-empty chunks until we see the empty chunk sentinel
```
SystemFw, line 278: Same as above, the use of empty chunks plus sentinel reads a bit convoluted to me

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: Never. I just habitually defensively code my bounds checks. I'm happy to shift this to a `==` if you like

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: Also the fact that `Queue` is separate. We can't really reify the state machine, but we can probably make the documentation more clear, and also perhaps by pulling things out into better-named functions to indicate the state transitions and tests.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: Still never

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: Same dealio as the others

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: Never

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: Why would it need to be? Sometimes this is *used* in `uncancelble` contexts, but I don't think this itself needs to be `uncancelable` since it's just reading state.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: The double-check is actually fundamental. Ultimately this stems from the fact that we need atomicity between `leases` and `q`, which are split by definition.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: So in this implementation, the following will happen in `stream` deterministically:

1. A `singleton` chunk will be emitted containing the first-ish produced element
2. A `seq` chunk will be emitted containing all of the remainder

In the original implementation, it was possible to get two or even just one `seq` chunk, but only if the consumer *repeatedly lost the race*. In other words, this is a pretty fundamental tradeoff between latency and throughput. The new implementation actually has better latency because it will deterministically give you an element *as soon* as it is available, whereas the old implementation could do that but also had a chance to just spin for a little while and then give you a bigger chunk.

Ultimately, singletons are the base mode of operation for `Channel`. When producers and consumers are at the same flow rate, every chunk will be `singleton`. This is good because it means everyone is keeping up! `seq` chunks only happen when the consumer is slower than the aggregate producer flow rate, resulting in trading off latency for throughput in an attempt to catch up. This can be viewed as a degraded case. Valuable in an elastic system, but still degraded.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: > Don't we have the guarantee that the Sentinel, if present, is the last element of this Chunk?

We do, but if the `Chunk` is a `seq` then its contents will be a `List`, so we can't look at the last element efficiently. Though, thinking about it, we can probably pull together a recursive function that is strictly faster than `filter` since we *only* have to look at the last element in this case…

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: Differently purposes. `Sentinel` is a signal to a blocked `take`. `Chunk.empty` is a signal to the typed `stream`. By stripping `Sentinel` in the taker, we avoid the need to continuously check for it in `stream`. We *do* have to continuously check for `Chunk.empty` but that's a lot more efficient since we don't need to scan the chunks.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
djspiewak: I guess I don't see that much of a difference here, since at a minimum the `Queue` is always going to be separated and represents an important component of the state machine (via both `Sentinel` and `size`). I actually had an earlier version of this where I collapsed these two values together and it didn't make any real difference in readability (in part because of the continual packing/unpacking of the compound state), but if you disagree I'm happy to adjust it and we can take a look.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
SystemFw: Mm, it doesn't strike me as fundamental at all because I don't think it's about atomicity between `leases` and `q` (which you can't get anyway):
The first check is there because you don't want to increase the number of leases if the queue is already closed (to ensure that closing logic is done by `close`), you then increase the number of leases, but have to check again to make sure the queue hasn't been closed in the meantime, so you can avoid offering an element in that case. So the issue is the lack of atomicity between `leases` and `closed`, which is my whole point: if they were in one Ref, there couldn't be a case where you increase the leases but the queue has been closed, and you wouldn't need the double check.

SystemFw: Kinda off the top of my head, but this does seem simpler to me:
```scala
def send(a: A): F[Either[Channel.Closed, Unit]] = {
  val doSend = q.offer(a.asInstanceOf[AnyRef]).as(RightUnit)
  val cleanup = state.modify { case State(leases, closed) =>
    (
      State(leases - 1, closed),
      q.offer(Sentinel).start.void.whenA(leases == 1 && closed)
    )
  }.flatten


  F.uncancelable { poll =>
    state.modify { case State(leases, closed)  =>
      if (closed) state -> F.unit
      else State(leases + 1, closed) -> poll(doSend).guarantee(cleanup)
    }.flatten
  }
}
```

core/shared/src/main/scala/fs2/concurrent/Channel.scala
SystemFw: Well, have a look at the effect on `send` https://github.com/typelevel/fs2/pull/2856#discussion_r1002765343

core/shared/src/main/scala/fs2/concurrent/Channel.scala
SystemFw: >  but it doesn't seem hard to fix either
I take this back because I realised there isn't a `takeAvailable` on Queue. That being said, not really agree with

>  resulting in trading off latency for throughput in an attempt to catch up

because it's not like the old implementation adds linger time in order to catch up, it uses the latency it's already wasted to potentially gain better throughput

> e new implementation actually has better latency because it will deterministically give you an element as soon as it is available

Isn't that the same as the old channel, i.e. a `Deferred[Unit]` triggers, except strictly worse because it "discards" additional elements that might have been added in the mean time?

That being said, it's probably not worth changing since there's no `takeAvailable`

SystemFw: > but sometimes the consumer would lose the race since the whole state was synchronized, meaning that it would spin again trying to grab its chunk

Ah I see what you mean now

djspiewak:
@SystemFw I think this is ready for another look. Your `Ref` consolidation advice is very good, and probably actually improved the performance of `send` (I'm checking now). I also corrected the issue with `closed` being fulfilled after quiescence rather than upon closure. Additionally, I fixed a bug in CE's `Queue.synchronous` which was the origin of the broadcast hang.

Depends on a hash snapshot for the time being, hence why this is marked as draft.

djspiewak:
Updated with benchmark results. The simple cases got faster with the consolidated `Ref`, but the more complex cases got *much* slower, and any case involving contention also got noticeably slower:

## `main`

```
[info] Benchmark                              (size)   Mode  Cnt     Score     Error  Units
[info] ChannelBenchmark.sendPull                  64  thrpt   10  6893.574 ± 125.608  ops/s
[info] ChannelBenchmark.sendPull                1024  thrpt   10  3535.811 ±  39.588  ops/s
[info] ChannelBenchmark.sendPull               16384  thrpt   10   361.574 ±   1.307  ops/s
[info] ChannelBenchmark.sendPullPar8              64  thrpt   10  8176.797 ± 118.350  ops/s
[info] ChannelBenchmark.sendPullPar8            1024  thrpt   10  3407.982 ±  22.908  ops/s
[info] ChannelBenchmark.sendPullPar8           16384  thrpt   10   250.676 ±   1.205  ops/s
[info] ChannelBenchmark.sendPullParUnlimited      64  thrpt   10  8622.935 ±  48.615  ops/s
[info] ChannelBenchmark.sendPullParUnlimited    1024  thrpt   10   925.061 ±  27.972  ops/s
[info] ChannelBenchmark.sendPullParUnlimited   16384  thrpt   10    39.134 ±   0.659  ops/s
```

## This PR

```
[info] Benchmark                              (size)   Mode  Cnt     Score     Error  Units
[info] ChannelBenchmark.sendPull                  64  thrpt   10  9343.856 ±  42.086  ops/s
[info] ChannelBenchmark.sendPull                1024  thrpt   10  2161.634 ±  12.066  ops/s
[info] ChannelBenchmark.sendPull               16384  thrpt   10    22.743 ±   0.856  ops/s
[info] ChannelBenchmark.sendPullPar8              64  thrpt   10  7830.841 ±  41.508  ops/s
[info] ChannelBenchmark.sendPullPar8            1024  thrpt   10  2175.548 ±  56.683  ops/s
[info] ChannelBenchmark.sendPullPar8           16384  thrpt   10   116.707 ±   2.704  ops/s
[info] ChannelBenchmark.sendPullParUnlimited      64  thrpt   10  8125.314 ± 164.499  ops/s
[info] ChannelBenchmark.sendPullParUnlimited    1024  thrpt   10  1375.947 ±  33.510  ops/s
[info] ChannelBenchmark.sendPullParUnlimited   16384  thrpt   10    54.626 ±   1.014  ops/s
```

djspiewak:
Fiddled around some more with the performance cases. I took a flier at porting one of the most dramatic performance deltas from the Cats Effect `Queue` benchmarks over to `Channel`. It was very similar to the `sendPar` benchmarks in terms of results, which suggests a couple conclusions:

- A lot of CE `Queue`'s performance benefits may be deriving from true multi-consumer cases. This can be verified by expanding the CE benchmark set a bit
- `Ref` as a single synchronization point may be imposing some fundamental overhead. This overhead is shared with the old `Channel` implementation. Another way of parsing this is to conclude that `Channel` is bottlenecked by its state, rather than by its queue.

I suspect this is still a worthwhile change for a few reasons (e.g. the GC tends to do a lot better with contiguous array structures rather than linked-list equivalents, and this is not well measured in microbenchmarks), and I have some ideas for how to improve it a bit further, but I think we can probably foreclose seeing the kinds of huge performance leaps in `Channel` that we see in the lower level `Queue`.

SystemFw:
> I suspect this is still a worthwhile change for a few reasons (e.g. the GC tends to do a lot better with contiguous array structures rather than linked-list equivalents, and this is not well measured in microbenchmarks), and I have some ideas for how to improve it a bit further, but I think we can probably foreclose seeing the kinds of huge performance leaps in Channel that we see in the lower level Queue.

mm, I don't want to be extremely gatekeeper-y, but this is the type of worst case scenario I feared: changing something that works, complicating the implementation (double refs, complex sentinel logic, `Any`) for unclear perf benefits.

Btw, what do you mean by linked list here? If it the fact that the current impl is based on List? That was benchmarked against a couple of alternatives, but it doesn't seem fundamental one way or another

nikiforo:
That's the benchmark results of List vs Vector: https://github.com/typelevel/fs2/pull/2751#issuecomment-994037977

In `List` version the insert operation during unsuccessful CAS discards only one ListNode, whereas in `Vector` version several nodes(each with a size of 64-element array) might be discarded. That intuition was behind the change from `Vector` to `List` in `Channel`.

djspiewak:
I did my final set of experiments! Tldr, I'm closing this PR.

So my last set of experiments looked at GC pressure (using `-prof gc`), under the hypothesis that the benchmarks aren't capturing the heap differences between a queue based on a mutable circular buffer projected into an array vs a large set of immutable `List` allocations. In theory, the new `Queue`-based implementation should have lower heap pressure, which would imply that its real-world performance could be dramatically better than the old `Ref`-based implementation, even if the microbenchmarks don't show a significant difference.

The results were the exact opposite. The implementation on this PR imposes about double the heap pressure, GC count, GC time, and survivor count. I genuinely have no intuition for *why*, but at this point I think we can safely conclude that the performance of this implementation is somewhere between "worse" and "much worse" than the performance of the old version. Full output attached.

- [old implementation](https://github.com/typelevel/fs2/files/9993398/benchmark.txt)
- [this PR](https://github.com/typelevel/fs2/files/9993397/benchmark-2.txt)

As an aside, I have separately verified that the performance improvements in CE's `Queue` are not just restricted to MPMC cases, and MPSC is similarly much faster than its old implementation. I'm having a hard time figuring out how to rectify this information with the investigation on this PR.

---

All in all, I think we can conclude a few things:

- `Ref`/`Deferred` is a startingly and unintuitively fast general method for implementing stuff
- `Channel`'s performance is bounded mostly by the `Ref#modify` loop, not by the state management in `List`
- Generational GC benefits a lot from the fact that the state objects here are very short lived. I think this is where this PR really fell behind: most of the heavy lifting is handled by data structures which are long-lived
- The tests in `main` are woefully insufficient.

I'll open a new PR bringing over the tests from this PR. :-)

PR #2860, opened 2022-03-28
Improve description of InvariantOps

bplommer:


diesalbla:
Thanks for the Pull Request. The new explanation is clearer and more explicit.

One doubt, though, is whether an implicit class, which is just a way to go around the limitations of the Scala language, is something one should document in the Scaladocs. It does not represent an entity of the data model, it is just a conduit to get some operations defined on the stream. Would perhaps a mere code comment (single star) suffice?

diesalbla (approved):
Approving since, as it is, is improvement.

PR #2877, opened 2022-04-05
Add unidocs artifact

armanbilge:
As previously mentioned, this adds a published unidocs artifact `fs2-docs`, like http4s and other projects are adopting.

https://www.javadoc.io/doc/org.http4s/http4s-docs_2.13/0.23.11/org/http4s/index.html

Still a bit annoying that you have to explicitly list the projects. Haven't figured out how to automate that yet. Feel free to pass if this over-complicates the build and we can consider again when it's easier.

After the undidocs are published in the next release, I'll PR an update to the site with the new API docs URL.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -170,6 +170,7 @@ lazy val root = tlCrossRootProject
     scodec,
     protocols,
     reactiveStreams,
+    unidocs,
     benchmark
   )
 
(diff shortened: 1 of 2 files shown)


armanbilge:
Nevermind, this is not ready for prime time. Maybe after the next scodec-bits release.

mpilquist:
Looks like we just need to disable fatal warnings on scaladoc?

armanbilge:
I thought we don't do shenanigans here :) the real "fix" is for scodec-bits to release with its API URL properly encoded in the POM. sbt-typelevel sets this automatically, which is why its next release should fix it.

build.sbt
```diff
@@ -318,6 +319,20 @@ lazy val reactiveStreams = project
   )
   .dependsOn(coreJVM % "compile->compile;test->test")
 
+lazy val unidocs = project
+  .in(file("unidocs"))
+  .enablePlugins(TypelevelUnidocPlugin)
+  .settings(
+    name := "fs2-docs",
```
armanbilge, line 326:
```suggestion
    name := "fs2-docs",
    tlFatalWarnings := false,
```

PR #2884, opened 2022-04-11
adds Stream.resourceK and Stream.resourceWeakK methods

jbwheatley:
Found this was a useful combinator when I was writing some library code and thought it would be sensible to add it here :)
Creates a `Resource[F, *] ~> Stream[F, *]`


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3614,6 +3614,12 @@ object Stream extends StreamLowPriority {
   def resource[F[_], O](r: Resource[F, O])(implicit F: MonadCancel[F, _]): Stream[F, O] =
     resourceWeak(r).scope
 
+  /** Same as [[resource]], but expressed as a FunctionK. */
+  def resourceK[F[_]](implicit F: MonadCancel[F, _]): Resource[F, *] ~> Stream[F, *] =
```
kubukoz, line 3618: **suggestion (non-blocking):** Should there be a `resourceWeakK` too?

jbwheatley: don't see why not

kubukoz (approved):

PR #2887, opened 2022-04-13
Small simplifications to use of types in Pull

bplommer:


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -854,19 +854,19 @@ object Pull extends PullLowPriority {
   ): F[B] = {
     var contP: ContP[Nothing, Nought, Any, Unit] = null
 
-    def getCont[Y, G[_], X]: Cont[Y, G, X] = contP.asInstanceOf[Cont[Y, G, X]]
+    def getCont(): Cont[Any, Nothing, Nothing] = contP.asInstanceOf[Cont[Any, Nothing, Nothing]]
```
bplommer, line 857: Added parens here to alert the unwary that it isn't RT

mpilquist: I don't love the use of parens here but don't feel super strongly about it either. I'm fine with merging as is if @diesalbla approves.

mpilquist (approved):

diesalbla (approved):

PR #2888, opened 2022-04-13
WIP - evaluating pure prefix of effectful streams

bplommer:
Followup to [discussion on http4s discord](https://discord.com/channels/632277896739946517/839254871646404638/963397757815648346) - the idea is to make a best effort at getting pure values from the head of a stream, then return them as a chunk together with a new Stream of the unevaluated tail. Currently just returns an `Option[ExitCase]` instead of a tail - and that isn't implemented correctly, as it will return `Some(Success)` instead of `None` if the evaluation limit is reached.

cc @diesalbla

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -845,13 +845,13 @@ object Pull extends PullLowPriority {
    * needs to find the recovery point where stream evaluation continues.
    */
   def compileChunk[F[_], O](
-      stream: Pull[F, O, Unit]
+      stream: Pull[Nought, O, Unit]
   )(implicit
       F: MonadError[F, Throwable]
   ): F[Chunk[O]] = {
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -836,6 +836,176 @@ object Pull extends PullLowPriority {
...
   private[this] type Nought[A] = Any
 
+  /* Left-folds the output of a stream in to a single value of type `B`.
+   *
+   * === Interruption ===
+   *
+   * Interruption of the stream is implemented cooperatively between `Pull` and `Scope`.
+   * Unlike interruption of an `F[_]: MonadCancelThrow` type (e.g. `IO`), stream interruption
+   * needs to find the recovery point where stream evaluation continues.
+   */
```
bplommer, line 846:
```suggestion
```

PR #2895, opened 2022-04-30
Fix `Stream#parJoin` for short-circuiting monad transformers

yakivy:
PR fixes hang of the following stream:
```
case object TestException extends Throwable with NoStackTrace
type F[A] = EitherT[IO, Throwable, A]

Stream.range(1, 64).covary[F]
  .map(i => Stream.eval[F, Unit](
    if (i == 7) EitherT.leftT(TestException)
    else EitherT.right(IO.unit)
  ))
  .parJoin(4)
  .map(_ => Stream.eval[F, Unit](EitherT.right(IO.unit)))
  .parJoin(4)
  .compile
  .drain
```

**Problem description:**
For the given stream `lease.cancel` raises monad transformer error, so the following `available.release >> decrementRunning` chain is not invoked and therefore output stream is not closed. But this is only a part of the problem, even if we close the stream, it will swallow monad error and return successful result.

**Fix description:**
To fix the first part of the problem, I replaced `>>` monad composition with `forceR`, that is from `MonadCancel` so should work fine for short-circuiting monad transformers. To fix the second part, I raised `lease.cancel` failures into effect, separated success cancel with `guaranteeCase` and composed cancel result with the stream outcome `oc <* Outcome.succeeded(fu)`.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4049,19 +4049,18 @@ object Stream extends StreamLowPriority {
                       .compile
                       .drain
                       .guaranteeCase(oc =>
-                        lease.cancel.rethrow
-                          .guaranteeCase {
-                            case Outcome.Succeeded(fu) =>
-                              onOutcome(oc <* Outcome.succeeded(fu), Either.unit)
+                        lease.cancel.rethrow.guaranteeCase {
+                          case Outcome.Succeeded(fu) =>
+                            onOutcome(oc <* Outcome.succeeded(fu), Either.unit)
 
-                            case Outcome.Errored(e) =>
-                              onOutcome(oc, Either.left(e))
+                          case Outcome.Errored(e) =>
+                            onOutcome(oc, Either.left(e))
 
-                            case _ =>
-                              F.unit
-                          }
-                          .forceR(available.release >> decrementRunning)
+                          case _ =>
+                            F.unit
+                        }
                       )
+                      .forceR(available.release >> decrementRunning)
                   }.void
                 }
             }
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4049,15 +4049,22 @@ object Stream extends StreamLowPriority {
                       .compile
                       .drain
                       .guaranteeCase(oc =>
-                        lease.cancel
-                          .flatMap(onOutcome(oc, _)) >> available.release >> decrementRunning
+                        lease.cancel.rethrow
+                          .guaranteeCase {
+                            case Outcome.Succeeded(fu) =>
+                              onOutcome(oc <* Outcome.succeeded(fu), Either.unit)
+
+                            case Outcome.Errored(e) =>
+                              onOutcome(oc, Either.right(e))
+
+                            case _ =>
+                              F.unit
+                          }
+                          .forceR(available.release >> decrementRunning)
                       )
-                      .handleError(_ => ())
                   }.void
                 }
             }
-
-          val RightUnit = Right(())
 
           def runOuter: F[Unit] =
             F.uncancelable { _ =>
@@ -4071,7 +4078,7 @@ object Stream extends StreamLowPriority {
                 .interruptWhen(done.map(_.nonEmpty))
                 .compile
                 .drain
-                .guaranteeCase(onOutcome(_, RightUnit) >> decrementRunning)
+                .guaranteeCase(onOutcome(_, Either.unit) >> decrementRunning)
                 .handleError(_ => ())
             }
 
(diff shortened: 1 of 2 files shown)


mpilquist (approved):

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4049,16 +4049,23 @@ object Stream extends StreamLowPriority {
...
+                            case Outcome.Succeeded(fu) =>
+                              onOutcome(oc <* Outcome.succeeded(fu), Either.unit)
+
+                            case Outcome.Errored(e) =>
+                              onOutcome(oc, Either.left(e))
+
+                            case _ =>
+                              F.unit
+                          }
+                          .forceR(available.release >> decrementRunning)
```
SystemFw, line 4063: I think this is correct but slightly hard to reason about. Can it not be moved inside the `guaranteeCase`?

core/shared/src/test/scala/fs2/StreamParJoinSuite.scala
```diff
@@ -210,6 +210,32 @@ class StreamParJoinSuite extends Fs2Suite {
...
+        )
+        .parJoin(4)
+        .map(_ => Stream.eval[F, Unit](EitherT.right(IO.unit)))
+        .parJoin(4)
+        .compile
+        .drain
+        .value
+        .flatMap { actual =>
+          IO(assertEquals(actual, Left(TestException)))
+        }
```
SystemFw, line 236: I think you can just:
```scala
.drain.value.assertEquals(Left(TestException))
```

core/shared/src/main/scala/fs2/Stream.scala
yakivy: Agree, moved out

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4098,16 +4093,15 @@ object Stream extends StreamLowPriority {
...
                 // to wait until all fibers have been joined, so we need to do it manually
                 // by waiting on the counter
-                running.waitUntil(_ == 0) >>
-                signalResult(fiber)
+                running.waitUntil(_ == 0)
             }
-            .flatMap { _ =>
-              output.stream.flatMap(Stream.chunk(_).covary[F])
+            .flatMap { fiber =>
+              output.stream.flatMap(Stream.chunk(_).covary[F]) ++ Stream.exec(signalResult(fiber))
```
diesalbla, line 4104: Tiny detail: we may not need `.covary[F]` here.

diesalbla (commented):
fine to me, but will let @SystemFw give it the final 👍🏼 and merge.

diesalbla (approved):

core/shared/src/main/scala/fs2/Stream.scala
SystemFw: I actually meant something else, can it not be moved inside the guaranteeCase so that `forceR` is not needed at all?

yakivy: So same order as now, but instead of `forceR` a `flatMap`? No, because stream can be compiled with monad error. If this is still not what you mean, please provide code snippet.

SystemFw: What I mean is that now the `release` is no longer part of a `guaranteeCase` block, which _should_ be correct in this specific case since cancelation shouldn't happen on that fiber, but it's hard to reason about, and also the `onOutcome` behaviour wrt interruption seems to have changed, because now if `oc` is `Canceled` `onOutcome` is no longer called on it.

Both these changes seem orthogonal to the actual bug fix, are they there on purpose?

SystemFw: So, I played with this a bit, and this works:
```scala
.guaranteeCase { oc =>                                       
  lease.cancel.rethrow.guaranteeCase { leaseOutcome =>       
    val propagateOutcome = leaseOutcome match {              
      case Outcome.Succeeded(fu) =>                          
        onOutcome(oc <* Outcome.succeeded(fu), Either.unit)  
      case _ =>                                              
        F.unit                                               
    }                                                        
                                                             
    propagateOutcome >> available.release >> decrementRunning
  }.handleError(_ => ())                                     
}.handleError(_ => ())                                       
```

The suggestion I meant to offer was only to move `available.release >> decrementRunning` inside the `guaranteeCase`, but as it turns out you actually don't need the `Errored` case when propagating results. 
I find that behaviour pretty confusing, since `.flatMap` short circuits but `guaranteeCase` has a `Succeeded` result (with a failed EitherT inside).
As an aside, the two `handleError(_ => ())` aren't actually needed but not having them might result in stacktraces getting logged

SystemFw: In any case, it's actually easier to think about that `forceR` inside the `guaranteeCase`, I'll add a commit and merge

SystemFw (approved):

PR #2898, opened 2022-05-13
Upgrade sync compiler when runtime constraint is `Async`

armanbilge:
This adopts the idea from https://github.com/typelevel/cats-effect/issues/2771, https://github.com/typelevel/cats-effect/issues/2772 to upgrade a sync compiler when at runtime the constraint is determined to be `Async`. Thoughts?


mpilquist:
One caution: behavior of compiled stream may be different as a result of upgrading to `Concurrent`. This is different than the `Queue` case where the only difference was improved performance.

armanbilge:
Right, this test goes from non-terminating to terminating I believe. 

https://github.com/typelevel/fs2/blob/cf6560b14b5caaaea109f49ee181a51d16395d71/core/shared/src/test/scala/fs2/StreamInterruptSuite.scala#L370-L374

PR #2909, opened 2022-05-28
Extract Unit Chunks, Pull, Stream. 

diesalbla:
- Add a `unit` chunk, which is just a singleton that contains a unit. 
- Add an `outUnit` Pull, which is just an `Output` around that chunk. Use it in a few places within Pull. 
- Add an `unit` Stream, which merely emits one unit value and ends.
-


rossabaker:
Could also use this in the Applicative instance, I think?

diesalbla:
> Could also use this in the Applicative instance, I think?

Added.

PR #2911, opened 2022-05-29
evalTap: no need for second parameter

diesalbla:
Can just use wildcard.


bplommer:
A cast to `F[Any]` makes this compile in Scala 3 but there's still a compiler warning about unreducible existential types so maybe this change is best avoided.

PR #2912, opened 2022-05-31
Run `SSLEngine` delegated tasks in parallel

armanbilge:
So, not sure about this change at all. Just something I noticed, maybe it's been discussed before?

The Javadoc says:
> Multiple delegated tasks can be run in parallel.

https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#getDelegatedTask--

So ideally running these tasks in parallel would improve throughput.

However the docs also suggest that some of these tasks may not be blocking per se, but actually compute-bound.
> SSLEngine implementations may require the results of tasks that may take an extended period of time to complete, or may even block. For example, a TrustManager may need to connect to a remote certificate validation service, or a KeyManager might need to prompt a user to determine which certificate to use as part of client authentication. Additionally, creating cryptographic signatures and verifying them can be slow, seemingly blocking.

If about one task per "batch" is compute bound and the rest are blocking, I think this change could make sense. However, if there are multiple compute bound tasks, my recent understanding is that this could actually be worse since they'd be granted unbounded parallelism.

Thoughts?

### DIFF
--- a/io/jvm/src/main/scala/fs2/io/net/tls/SSLEngineTaskRunner.scala
+++ b/io/jvm/src/main/scala/fs2/io/net/tls/SSLEngineTaskRunner.scala
@@ -24,9 +24,11 @@ package io
 package net
 package tls
 
-import javax.net.ssl.SSLEngine
+import cats.effect.Async
+import cats.effect.implicits._
+import cats.syntax.all._
 
-import cats.effect.Sync
+import javax.net.ssl.SSLEngine
 
 private[tls] trait SSLEngineTaskRunner[F[_]] {
   def runDelegatedTasks: F[Unit]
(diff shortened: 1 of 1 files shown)


mpilquist:
Hasn't been discussed before. I'm curious if this changes performance -- I suspect it won't but it would be good to know.

armanbilge:
So I know even less about benchmarking :) but I gave it a try here:

https://github.com/typelevel/fs2/blob/54551e7d0e8ecfb366cb6344380ac26733aa75d8/benchmark/src/main/scala/fs2/benchmark/TLSBenchmark.scala

Still a bit bugged, but if it's not entirely stupid I can try running it with/without the changes in this PR.

armanbilge:
Looking at netty sources, they don't seem to be running these delegated tasks in parallel.
https://github.com/netty/netty/blob/871dd2e5730a477738a5a9b9af19be35244137d0/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L1522

armanbilge:
Blaze also runs the tasks sequentially.
https://github.com/http4s/blaze/blob/fa6f48f385d8185870c4b92613beb97681572e35/core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala#L395

armanbilge:
I added a counter/println to see how many delegated tasks are run in a row. Well at least on `Temurin-17.0.3+7` it seems only one-at-a-time. So I think that's enough points against this change 🙂

Issue #2777, opened 2021-12-27
Replace the auto-generated Node.js facade with a hand-maintained facade


armanbilge:
Discussed with @mpilquist today in Discord. The auto-generated Scalably Typed facade has served well, especially during the development of fs2.io.js. However, now that things are stable, the advantages of auto-generation are possibly out-weighed by its downsides:
1. 700+ auto-generated files, of which we use only a small fraction.
2. A Scala.js bundler dependency, which creates 8000 lines of lockfiles and downstream headaches.
    - https://github.com/scala-steward-org/scala-steward/issues/2252
3. Maintenance burden:
    - https://github.com/typelevel/fs2/issues/2489
    - https://github.com/typelevel/fs2/pull/2681
    - https://github.com/typelevel/fs2/pull/2775
    - https://github.com/typelevel/fs2/pull/2857
4. Various unsafe casts, due to deficiencies of the underlying Typescript definitions.
5. Not really important, but lack of flexibility to support features like in-memory filesystems etc.

By crafting a minimal, hand-written facade, we can address all of the above and overall reduce the impact of the fs2.io.js cross.

If someone has an appetite to work on this, I'll be thrilled :) otherwise, on my yak-list.

PR #2918, opened 2022-06-05
Switch to a hand-maintained Node.js facade

armanbilge:
Closes https://github.com/typelevel/fs2/issues/2777. This replaces the 700+ files auto-generated by Scalably Typed with 10 files focused on the Node.js APIs we actually use. This lets us remove the ScalablyTyped and Bundler plugins and lockfiles.

One of the hazards of a hand-maintained facade is that we now rely entirely on runtime tests to know if any of this works, since a facade could be completely wrong and is not statically verified against anything. So there could be some regressions due to this change for any APIs lacking coverage. We can slowly improve this.


armanbilge:
I published cc479f4700c716a095eec555b2ac63d7e47122b7 as `3.2.7-100-cc479f4-SNAPSHOT` and ran CI for http4s and skunk. I think we're good.

(Aside: even better, I think the size of the generated JS artifact (which we monitor in http4s) decreased by 10%, but I can't understand why 🤔 ).

PR #2922, opened 2022-06-09
Respect FIFO for bounded Channel

armanbilge:
H/t @davesmith00000 for reporting and @Daenyth for helping to diagnose it.

The issue was first identified in terms of a synchronous channel: while the channel is blocked awaiting a consumer producers are queued, but then dequeued in reverse-order when there is downstream consumption. This causes the elements to become reversed, effectively FILO.


Daenyth (approved):

Daenyth:
Quick and clean fix 👍

Issue #2931, opened 2022-06-24
`raiseError` causes `writeOutputStream` to hang


armanbilge:
H/t to @bpholt for the original minimization in https://github.com/Dwolla/fs2-hang-repro

```scala
//> using scala "3.1.3"
//> using lib "co.fs2::fs2-io::3.2.8"

package repro

import cats.effect._
import cats.syntax.all._
import fs2._
import fs2.io.{readOutputStream, writeOutputStream}

object Bug extends IOApp.Simple {
   
  override def run: IO[Unit] = {
    val s = Stream
      .emit[IO, Byte](0)
      .repeat
      .take(3)
      .through { in =>
        readOutputStream(1) { out =>
          in
            .through(writeOutputStream(IO.pure(out)))
            .compile
            .drain
        }
      }

    // remove the raiseError and it completes normally
    (s *> Stream.raiseError[IO](new RuntimeException("boom"))).compile.drain
  }

}
```

armanbilge:
```
cats.effect.IOFiber@1e65161f WAITING
 
cats.effect.IOFiber@705cc7b3 WAITING
 ├ main$ @ repro.Bug$.main(bug.scala:11)
 ├ void @ fs2.Stream.$anonfun$3(Stream.scala:567)
 ├ main$ @ repro.Bug$.main(bug.scala:11)
 ├ race @ fs2.internal.InterruptContext.eval$$anonfun$1(InterruptContext.scala:91)
 ╰ map @ fs2.internal.InterruptContext.eval$$anonfun$1(InterruptContext.scala:94)
 
cats.effect.IOFiber@6a21c738 WAITING
 
cats.effect.IOFiber@5cc2f18d WAITING
 ├ get @ fs2.Stream.watch$1(Stream.scala:561)
 ├ get @ fs2.Stream.watch$1(Stream.scala:561)
 ├ flatMap @ fs2.Stream.$anonfun$4(Stream.scala:572)
 ├ >>$extension @ fs2.Pull$.fs2$Pull$$$_$go$1(Pull.scala:1209)
 ├ run$ @ repro.Bug$.run(bug.scala:11)
 ├ complete @ fs2.Stream.$anonfun$2$$anonfun$1(Stream.scala:572)
 ├ get @ fs2.Stream.watch$1(Stream.scala:561)
 ├ >>$extension @ fs2.Pull$.fs2$Pull$$$_$go$1(Pull.scala:1209)
[...]

PR #2932, opened 2022-06-25
Use `interruptible` in `writeOutputStream`

armanbilge:
Fixes https://github.com/typelevel/fs2/issues/2931.

As I understand it, interruptibility [doesn't come for free](https://github.com/typelevel/cats-effect/blob/7cb291426cadb359c786b1d318a46bced55d4ba3/core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala#L27-L146). How do we feel about placing it in the hot path here? Maybe it should be configurable?


djspiewak:
Fwiw, in this type of thing, I suspect the overhead of `interruptible` is unnoticeable.

diesalbla (approved):

Issue #2939, opened 2022-07-08
Possible binary compability problem on fs2-3.2.9 on Scala 2.12


hamnis:
I was upgrading http4s-netty and found this while testing:

java.lang.NoSuchMethodError: 'fs2.Stream fs2.Stream$PartiallyAppliedFromIterator$.apply$extension(boolean, scala.collection.Iterator, int, cats.effect.kernel.Sync)'
    at org.http4s.jdkhttpclient.JdkHttpClient$.$anonfun$apply$17(JdkHttpClient.scala:229)

This works on fs2 3.2.8

armanbilge:
Woops, thanks, I know what happened. But why didn't MiMa catch it :(

PR #2933, opened 2022-06-25
Make `Stream.fromIterator` sync type configurable

armanbilge:
Fixes https://github.com/typelevel/fs2/issues/2903, which I found new sympathy for while working on https://github.com/typelevel/fs2/issues/2931.

This allows a third parameter to be passed to `Stream.fromIterator` indicating whether to use `Delay`, `Blocking`, `Interruptible`, or `InterruptibleMany`.

PR #2940, opened 2022-07-08
Fix bincompat of `PartiallyAppliedFromIterator` on Scala 2.12

armanbilge:
Fixes https://github.com/typelevel/fs2/issues/2939.

This was my fault due to changes in https://github.com/typelevel/fs2/pull/2933.

Inline comments below.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3296,7 +3296,7 @@ object Stream extends StreamLowPriority {
   def fromEither[F[_]]: PartiallyAppliedFromEither[F] =
     new PartiallyAppliedFromEither(dummy = true)
 
-  private[fs2] final class PartiallyAppliedFromIterator[F[_]](
+  final class PartiallyAppliedFromIterator[F[_]](
       private val blocking: Boolean
   ) extends AnyVal {
     def apply[A](iterator: Iterator[A], chunkSize: Int)(implicit F: Sync[F]): Stream[F, A] =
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3296,7 +3296,7 @@ object Stream extends StreamLowPriority {
   def fromEither[F[_]]: PartiallyAppliedFromEither[F] =
     new PartiallyAppliedFromEither(dummy = true)
 
-  private[fs2] final class PartiallyAppliedFromIterator[F[_]](
+  final class PartiallyAppliedFromIterator[F[_]](
```
armanbilge, line 3299: Because this class was package-private, so we received no bincompat support from MiMa. Indeed, the fact that this works at all seems like a manifestation of https://github.com/scala/bug/issues/6794.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3318,7 +3318,18 @@ object Stream extends StreamLowPriority {
...
+  object PartiallyAppliedFromIterator {
+    @deprecated("bincompat", "3.2.10")
+    private[fs2] def `apply$extension`[F[_], A](
+        dummy: Boolean,
+        iterator: Iterator[A],
+        chunkSize: Int,
+        F: Sync[F]
+    ): Stream[F, A] =
+      new PartiallyAppliedFromIterator(dummy).apply(iterator, chunkSize)(F)
+  }
```
armanbilge, line 3330: On Scala 2.12, adding an overload of a method defined on an `AnyVal` causes the the binary name of the original method to change, breaking bincompat. Fortunately, we can restore a method with the same signature manually.

armanbilge:
Bla, this became a bit messy 😕

PR #2941, opened 2022-07-10
Optimize `Chunk#toByteVector` when backed by a `ByteBuffer`

armanbilge:
When a `Chunk` is backed by `ByteBuffer` it is not necessary to copy the bytes to convert it to a `ByteVector`. This PR implements the conversion by creating a `ByteVector` "view" that directly wraps the chunk's underlying `ByteBuffer`.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -365,6 +365,7 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
   def toByteVector[B >: O](implicit ev: B =:= Byte): ByteVector =
     this match {
       case c: Chunk.ByteVectorChunk => c.toByteVector
+      case c: Chunk.ByteBuffer => ByteVector.view(c.buf.asReadOnlyBuffer())
       case other =>
         val slice = other.asInstanceOf[Chunk[Byte]].toArraySlice
         ByteVector.view(slice.values, slice.offset, slice.length)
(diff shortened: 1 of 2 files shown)


diesalbla:
@armanbilge Can you give a quick summary of what would be the effect of this change?

diesalbla (approved):

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -365,6 +365,12 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
   def toByteVector[B >: O](implicit ev: B =:= Byte): ByteVector =
     this match {
       case c: Chunk.ByteVectorChunk => c.toByteVector
+      case c: Chunk.ByteBuffer      =>
+        // if c is a view, position/limit of underlying buffer may have changed
+        val bb = c.buf.asReadOnlyBuffer()
+        bb.position(c.offset)
```
mpilquist, line 371: Do these need to take in to account the original position and limit as well? E.g. `bb.position(bb.position + c.offset)`

armanbilge: I don't think so. `bb.position()` is not constant and `bb.position(i)` is not relative as far as I can make out.

mpilquist: Okay got it, when we create a `Chunk.ByteBuffer`, we copy the position & limit in to offset & length.

PR #2945, opened 2022-07-14
Update build-site.md guide

b-vennes:
The `sbt "microsite / mdoc"` command lands the compiled microsite in the `mdoc/target/mdoc` directory instead of `target/website`.


armanbilge (approved):
Thanks for fixing this!

PR #2953, opened 2022-07-27
Node.js Gzip

armanbilge:
Closes https://github.com/typelevel/fs2/pull/2771, with many apologies to @yurique for their heroic effort on that 🙏 

In https://github.com/typelevel/fs2/pull/2771 we attempted to cross-build the existing Gzip implementation on the JVM. Instead, this PR delegates directly to the Node.js Gzip implementation. This is better for performance and is much simpler to implement. However, we are unable to support the optional headers for filename, comment, and timestamp. Fortunately http4s does not require these in its Gzip middlewares.

There is an unfortunate amount of duplication of tests and scaladocs in this PR.

### DIFF
--- a/core/js/src/main/scala/fs2/compression/CompressionPlatform.scala
+++ b/core/js/src/main/scala/fs2/compression/CompressionPlatform.scala
@@ -22,6 +22,91 @@
 package fs2
 package compression
 
-private[compression] trait CompressionPlatform[F[_]]
+private[compression] trait CompressionPlatform[F[_]] {
+
+  /** Returns a pipe that incrementally compresses input into the GZIP format
+    * as defined by RFC 1952 at https://www.ietf.org/rfc/rfc1952.txt. Output is
+    * compatible with the GNU utils `gunzip` utility, as well as really anything
+    * else that understands GZIP. Note, however, that the GZIP format is not
+    * "stable" in the sense that all compressors will produce identical output
+    * given identical input. Part of the header seeding is arbitrary and chosen by
+    * the compression implementation. For this reason, the exact bytes produced
+    * by this pipe will differ in insignificant ways from the exact bytes produced
+    * by a tool like the GNU utils `gzip`.
+    *
+    * GZIP wraps a deflate stream with file attributes and stream integrity validation.
+    * Therefore, GZIP is a good choice for compressing finite, complete, readily-available,
+    * continuous or file streams. A simpler deflate stream may be better suited to
+    * real-time, intermittent, fragmented, interactive or discontinuous streams where
+    * network protocols typically provide stream integrity validation.
+    *
+    * @param bufferSize The buffer size which will be used to page data
+    *                   into chunks. This will be the chunk size of the
+    *                   output stream. You should set it to be equal to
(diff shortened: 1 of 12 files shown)


core/js/src/main/scala/fs2/compression/CompressionPlatform.scala
```diff
@@ -22,6 +22,91 @@
...
+    * @param fileName         optional file name
+    * @param comment          optional file comment
+    */
+  def gzip(
+      bufferSize: Int = 1024 * 32,
+      deflateLevel: Option[Int] = None,
+      deflateStrategy: Option[Int] = None,
+      modificationTime: Option[Nothing] = None,
+      fileName: Option[Nothing] = None,
+      comment: Option[Nothing] = None
```
armanbilge, line 63: Using `Option[Nothing]` keeps us source-compatible with JVM while only accepting `None`. Furthermore, thanks to type erasure it leaves the door open to add support for these in the future if Node.js ever gets them.
- https://github.com/nodejs/node/issues/39545

core/shared/src/test/scala/fs2/CompressionSuite.scala
```diff
@@ -82,6 +82,36 @@ abstract class CompressionSuite(implicit compression: Compression[IO]) extends F
     )
   )
 
+  test("inflate please wrap") {
```
armanbilge, line 85: Added these tests because actually the existing Node.js inflate/deflate implementation was broken 😅

io/js/src/main/scala/fs2/io/compressionplatform.scala
```diff
@@ -26,16 +26,67 @@ import cats.effect.kernel.Async
...
+
+      def gzip(
+          fileName: Option[Nothing],
+          modificationTime: Option[Nothing],
+          comment: Option[Nothing],
+          deflateParams: DeflateParams
+      ): Pipe[F, Byte, Byte] = deflateImpl(deflateParams) { options =>
+        require(!deflateParams.fhCrcEnabled, "FHCRC is not supported on Node.js")
+        deflateParams.header match {
+          case ZLibParams.Header.GZIP => facade.zlib.createGzip(options)
```
armanbilge, line 65: Delegating to Node.js is as easy is this.

io/js/src/test/scala/fs2/io/NodeJSCompressionSuite.scala
```diff
@@ -40,21 +45,191 @@ class NodeJSCompressionSuite extends CompressionSuite {
...
     val out =
       if (nowrap)
-        facade.zlib.gunzipSync(in, options)
+        facade.zlib.inflateRawSync(in, options)
       else
         facade.zlib.inflateSync(in, options)
     Chunk.uint8Array(out).toArray
   }
 
+  test("gzip |> gunzip ~= id") {
```
armanbilge, line 65: Frustratingly, these tests are cloned from the JVM suite except modified to remove the optional headers unsupported on Node.js. The duplication is unfortunate but I didn't see a clean way to share them.

PR #2955, opened 2022-07-28
Add readUtf8 convenience method

keirlawson:


io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -419,6 +423,9 @@ object Files extends FilesCompanionPlatform {
         cursor.seek(start).readUntil(chunkSize, end).void.stream
       }
 
+    def readUtf8(path: Path): F[String] =
+      readAll(path).through(text.utf8.decode).compile.lastOrError
```
armanbilge, line 427: Is this right? I think it should be 
```suggestion
    def readUtf8(path: Path): F[String] =
      readAll(path).through(text.utf8.decode).compile.string
```

mpilquist: Signature wise, I think we should return a `Stream[F, String]` and let the user decide whether to accumulate, etc. I suggest these signatures:

```scala
def readUtf8(path: Path): Stream[F, String] =
  readAll(path).through(text.utf8.decode)

def readUtf8Lines(path: Path): Stream[F, String] =
  readUtf8(path).through(text.lines)
```

keirlawson: I feel like the accumulation case is likely very common (lots of time you just want to read in a file in a oner), could we also have an convenience method for it? No sure what you would call it though...

mpilquist: How about this?

```scala
def readUtf8AsSingleString(path : Path): F[String] =
  readUtf8(path).compile.string
```

armanbilge: (Please ignore me but) ehhh. It's super easy to tack on `.compile.string` and that's an ugly method name. I understand `.through(text.utf8.decode)` is annoying since it requires an import.

keirlawson: Perfect :) the name is a bit on the wordy side but I don't exactly have a better suggestion so that works for me :)

keirlawson: I guess the reason I'm keen to avoid the `.compile.string` is lots of the time I just want to read a file, and I don't care that it is streaming under the hood, its not hard obviously, but feels like the code shouldn't need to be concerned with streaming at all. This often comes up when teaching as well, ie we build toy apps that do things with files but I don't necessarily want to complicate things with streaming concerns.  I do agree that the name is clumsy though.

mpilquist: I lean towards not including the `F[String]` variant but certainly sympathetic to concerns raised by @keirlawson. I do think the simple method names (e.g. `readAll`, `readUtf8`, `readUtf8Lines`) should all be stream based. I'm okay with a handful of conveniences that tack on stream compilation but they shouldn't be able to be mistaken for the "default" methods. Something you're intentionally reaching for.

keirlawson: Ok, have updated in line with this discussion

PR #2957, opened 2022-07-28
Return client `TLSSocket` only after handshake is complete on Node.js

armanbilge:
This is what I meant to fix before I got distracted by https://github.com/typelevel/fs2/security/advisories/GHSA-2cpx-6pqp-wf35.

The problem came up in the ember h2 client, which needs to know the negotiated application protocol. On the JVM side, writing a `Chunk.empty` to the socket kickstarts the handshaking and blocks (semantically, hopefully) until it completes. So the negotiated application protocol is available in the next operation.

```scala
_ <- Resource.eval(tlsSocket.write(Chunk.empty))
protocol <- Resource.eval(H2TLS.protocol(tlsSocket))
```

https://github.com/http4s/http4s/blob/7fcbd6d58b923a5cae7c66100845c6dd1a857498/ember-core/shared/src/main/scala/org/http4s/ember/core/h2/H2Client.scala#L142-L143

Unfortunately this trick does not work for JS. So the `write` was completing before the handshake was negotiated, and attempting to check the `applicationProtocol` in the next step was coming up empty and the connection would fallback to http/1.1.

In Node.js, a `'secureConnect'` event is emitted when handshaking completes and the negotiated application protocol  is available. This PR modifies the client-mode `TLSSocket` upgrade to await this event before returning the socket.

As it turns out, the changes in https://github.com/typelevel/fs2/security/advisories/GHSA-2cpx-6pqp-wf35 actually do the equivalent thing for a server-mode `TLSSocket` by waiting for the `'secure'` event to indicate handshake completion.

While writing a test for this, I also added a test for session access, which is also controlled by an emitted `'session'` event on Node.js.

Tests also needed some minor refactoring so that the client and server connections are established in parallel.

### DIFF
--- a/io/js/src/main/scala/fs2/io/net/tls/TLSSocketPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/net/tls/TLSSocketPlatform.scala
@@ -84,6 +84,7 @@ private[tls] trait TLSSocketCompanionPlatform { self: TLSSocket.type =>
         readable
       ).race(errorDef.get.flatMap(F.raiseError[SuspendedStream[F, Byte]]).toResource)
         .map(_.merge)
+      _ <- errorDef.tryGet.map(_.toLeft(())).rethrow.toResource
     } yield new AsyncTLSSocket(
       tlsSock,
       readStream,
--- a/io/js/src/main/scala/fs2/io/net/tls/TLSContextPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/net/tls/TLSContextPlatform.scala
@@ -60,7 +60,7 @@ private[tls] trait TLSContextCompanionPlatform { self: TLSContext.type =>
           ): Resource[F, TLSSocket[F]] = Dispatcher[F]
             .flatMap { dispatcher =>
               if (clientMode) {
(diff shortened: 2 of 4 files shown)


ChristopherDavenport:
What happens on handshake failure?

io/js/src/main/scala/fs2/io/net/tls/TLSSocketPlatform.scala
```diff
@@ -84,6 +84,7 @@ private[tls] trait TLSSocketCompanionPlatform { self: TLSSocket.type =>
         readable
       ).race(errorDef.get.flatMap(F.raiseError[SuspendedStream[F, Byte]]).toResource)
         .map(_.merge)
+      _ <- errorDef.tryGet.map(_.toLeft(())).rethrow.toResource
```
armanbilge, line 87: @ChristopherDavenport the error is raised here (it is captured in an `'error'` event listener above).

ChristopherDavenport: Ok, this makes sense and looks good to me then.

ChristopherDavenport (approved):

PR #2962, opened 2022-08-11
Weird interaction between cancelation and interruptWhen

TimWSpence:
Fix  hanging when combining `Stream.eval(F.canceled)` and `interruptWhen`.

I believe that the `.as(canceledError)` is safe as our `fa.attempt` was canceled which means that cancelation is unmasked in this scope and hence `F.canceled.as(canceledError) <-> F.canceled`. This matches the behaviour in `Scope#interruptibleEval` when there is no interrupt context and it binds `fa.attempt` directly.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/InterruptContext.scala
+++ b/core/shared/src/main/scala/fs2/internal/InterruptContext.scala
@@ -92,7 +92,7 @@ final private[fs2] case class InterruptContext[F[_]](
           case Right(oc) =>
             oc.embed(F.canceled.as(canceledError)).map(_.leftMap(Outcome.Errored(_)))
           case Left(oc) =>
-            oc.embed(F.pure(Outcome.Canceled[Id, Throwable, Unique.Token]())).map(Left(_))
+            oc.embedNever.map(Left(_))
         }
     }
 }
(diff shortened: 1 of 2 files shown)


armanbilge (approved):
Nice find+fix!

core/shared/src/main/scala/fs2/internal/InterruptContext.scala
```diff
@@ -88,15 +88,19 @@ final private[fs2] case class InterruptContext[F[_]](
     ref.get.flatMap {
       case Some(outcome) => F.pure(Left(outcome))
       case None =>
-        F.race(deferred.get, fa.attempt).map {
-          case Right(result) => result.leftMap(Outcome.Errored(_))
-          case Left(other)   => Left(other)
+        F.raceOutcome(deferred.get, fa.attempt).flatMap {
+          case Right(oc) =>
+            oc.embed(F.canceled.as(canceledError)).map(_.leftMap(Outcome.Errored(_)))
```
armanbilge, line 93: Just a nit, maybe we can write this with `oc.fold(...)` to save the two separate ops?

TimWSpence: Oh yep, good call 👍 Will fix tomorrow

PR #2967, opened 2022-08-15
Align the semantics of spawn with the ones of concurrently

Baccata:
Context : https://discord.com/channels/632277896739946517/632310980449402880/1007228488555708476

This changes the implementation of `spawn` in order for the resulting stream to fail when the background stream fails, aligning it with the semantics of `concurrently`. This makes spawn behaviouraly equivalent to `Stream(()).concurrently(self)`, whilst still exposing a handle to the background fiber, allowing for its manual cancelation. 

If changing the existing behaviour of "spawn" is deemed risky, could another name be suggested ? (if the addition is considered valuable by the maintainers, that is)

### DIFF
--- a/core/shared/src/test/scala/fs2/StreamSpawnSuite.scala
+++ b/core/shared/src/test/scala/fs2/StreamSpawnSuite.scala
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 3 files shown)


core/shared/src/test/scala/fs2/StreamSpawnSuite.scala
```diff
@@ -0,0 +1,106 @@
...
+
+package fs2
+
+import scala.concurrent.duration._
+
+import cats.effect.IO
+import cats.effect.std.Semaphore
+import org.scalacheck.effect.PropF.forAllF
+
+class StreamSpawnSuite extends Fs2Suite {
```
Baccata, line 30: I copied a few tests verbatim from `StreamConcurrentlySuite`, and added one to assert the manual cancelation. Let me know if I should make some tweaks / port more tests, etc.

diesalbla:
> This makes spawn behaviouraly equivalent to Stream(()).concurrently(self), 

So, the semantics of `concurrently` is for the background process (`self`) be interrupted as soon as the foreground process (`Stream(())`) ends; thus, this would just be equivalent to the downstream consumer to interrupt `self` upon pulling from the stream. Is that the intention?

Baccata:
Sort of. I suppose the usefulness is better explained with monadic composition than sequential aggregation, as it allows to cover similar usecases to "io.background" whilst ensuring that background errors make it to the foreground, which is impossible (I think) to achieve with just Resource

SystemFw (approved):

Issue #2966, opened 2022-08-14
Stream.resource/Stream.bracket release action is not being completed when stream is interrupted


allter:
When lifting Resource to Stream using Stream.resource, or when using Stream.bracket, then "acquire" action is treated correctly, but "release" action is not.

The expected behaviour is that when interrupting stream execution, the Resource/bracket release actions are executed without interruption before advancing to other stream finalizers.

The actual behavior is that other stream finalizers run immediately after interruption, and if program exits right after this, then the release action is not completed.

Everything works OK if the stream is wrapped into an outer stream using parJoin (and this is a workaround)

fs2: 3.2.12
cats-effect: 3.3.14

Example code ( [scastie](https://scastie.scala-lang.org/8WCJmUBERF2Rzy0aQCUf1w) )
```scala
import cats.effect.{ExitCode, IO, IOApp, Resource, Async, Sync, Temporal}
import cats.effect.syntax.all._
import cats.syntax.all._
import fs2.Stream

import scala.concurrent.duration.DurationInt

object BracketTest extends IOApp {
  case class RunF[F[_]: Async]() {
    def say(msg: String) = Sync[F].delay(println(msg))

    val sleeper = (
      Temporal[F].sleep(1.seconds) *>
        say("slept")
    ).replicateA(4).void

    val sleeperResource = Resource
      .make(sleeper)(_ =>
        say("releasing sleeper resource...") *>
          sleeper *>
          say("released sleeper resource")
      )

    val stream =
      Stream.resource(sleeperResource).evalTap(_ => say("using resource")).void
        .onFinalizeCase(ec =>
          say(s"Finalized with ec=$ec")
        )

    val runF =
      stream.compile.drain // Resource is not released OK
      //Stream(stream).parJoinUnbounded.compile.drain // Resource is released OK
  }

  override def run(args: List[String]): IO[ExitCode] = for {
    _ <- RunF[IO]().runF.timeout(6.seconds)
  } yield ExitCode.Success
}
```

armanbilge:
Here's a further minification.

```scala
//> using scala "3.1.3"
//> using lib "co.fs2::fs2-core::3.2.12"

import cats.effect._
import fs2._

import scala.concurrent.duration._

object BracketTest extends IOApp.Simple {
  def run = Stream
    .bracket(IO.unit)(_ =>
      IO.println("releasing") *>
        IO.sleep(2.seconds) *>
        IO.println("released")
    )
    .compile
    .drain
[...]

PR #2968, opened 2022-08-15
Bracket finalizer should be uncancelable

armanbilge:
Possibly fixes https://github.com/typelevel/fs2/issues/2966?

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Scope.scala
+++ b/core/shared/src/main/scala/fs2/internal/Scope.scala
@@ -189,7 +189,7 @@ private[fs2] final class Scope[F[_]] private (
           acquire(poll).redeemWith(
             t => F.pure(Left(t)),
             r => {
-              val finalizer = (ec: Resource.ExitCase) => release(r, ec)
+              val finalizer = (ec: Resource.ExitCase) => F.uncancelable(_ => release(r, ec))
               resource.acquired(finalizer).flatMap { result =>
                 if (result.exists(identity)) {
                   register(resource).flatMap {
--- a/core/shared/src/test/scala/fs2/BracketSuite.scala
+++ b/core/shared/src/test/scala/fs2/BracketSuite.scala
@@ -357,4 +357,16 @@ class BracketSuite extends Fs2Suite {
       .compile
       .drain
   }
+
+  test("#2966 release is uncancelable") {
+    IO.ref(false)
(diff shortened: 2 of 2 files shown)


mpilquist:
Can we try how this fix works with both this:

```scala
Stream.bracket(IO.unit)(_ => IO.sleep(2.seconds)).compile.drain.timeout(1.seconds)
```

...as well as the original bug in #2966?

armanbilge:
This one appears to be fixed:
```scala
//> using scala "3.1.3"
//> using lib "co.fs2::fs2-core::3.2.12-15-ca20578-SNAPSHOT"
//> using repository "https://s01.oss.sonatype.org/content/repositories/snapshots/"

import cats.effect._
import fs2._

object BracketTest extends IOApp.Simple {
  def run = Stream
    .bracket(IO.unit)(_ =>
      IO.println("releasing") *>
        IO.sleep(2.seconds) *>
        IO.println("released")
    )
    .compile
    .drain
    .timeout(1.seconds)
}
```

---

As does this one:
```scala
//> using scala "3.1.3"
//> using lib "co.fs2::fs2-core::3.2.12-15-ca20578-SNAPSHOT"
//> using repository "https://s01.oss.sonatype.org/content/repositories/snapshots/"

import cats.effect.{ExitCode, IO, IOApp, Resource, Async, Sync, Temporal}
import cats.effect.syntax.all._
import cats.syntax.all._
import fs2.Stream

object BracketTest extends IOApp {
  case class RunF[F[_]: Async]() {
    def say(msg: String) = Sync[F].delay(println(msg))

    val sleeper = (
      Temporal[F].sleep(1.seconds) *>
        say("slept")
    ).replicateA(4).void

    val sleeperResource = Resource
      .make(sleeper)(_ =>
        say("releasing sleeper resource...") *>
          sleeper *>
          say("released sleeper resource")
      )

    val stream =
      Stream.resource(sleeperResource).evalTap(_ => say("using resource")).void
        .onFinalizeCase(ec =>
          say(s"Finalized with ec=$ec")
        )

    val runF =
      stream.compile.drain // Resource is not released OK
      //Stream(stream).parJoinUnbounded.compile.drain // Resource is released OK
  }

  override def run(args: List[String]): IO[ExitCode] = for {
    _ <- RunF[IO]().runF.timeout(6.seconds)
  } yield ExitCode.Success
}
```

Issue #2969, opened 2022-08-27
The instances allocated in the `Stream#flatmap` are referenced in the `Pull#BindBind` which can't be GCed before the whole Stream ends


floating-cat:
fs2 version: 3.0.3 and 3.2.12
Full reproduction code: https://github.com/floating-cat/fs2-netty/tree/memory_leak
Related code line: https://github.com/floating-cat/fs2-netty/blob/memory_leak/core/src/main/scala/fs2/netty/SocketHandler.scala#L59-L64

Steps:
1. Clone the whole project above.
2. Run `src/main/scala/fs2/netty/benchmarks/echo/Fs2Netty.scala` with host and port arguments for listening.
3. Run `benchmarks/rps.sh` to benchmark it. (this script needs [rust_echo_bench](https://github.com/haraldh/rust_echo_bench) command, and you may need to use the `cargo run --release -- --help` command to compile it)

Result:
The program causes a memory leak and If I use `-Xmx200m -XX:MaxDirectMemorySize=200m`, the program can out of memory.

The related code is like the below:

```scala
lazy val reads: Stream[F, Byte] =
    Stream.force {
      Functor[F].ifF(isOpen)(
        // note the `++ reads` code here
        // I found a lots of `toChunk(b)` instances referenced in the `Pull#BindBind` which can't be GC
        // before this whole Stream ends and leaks memory a lot.
        fetch.flatMap(b => if (b == null) Stream.empty else Stream.chunk(toChunk(b)) ++ reads)
        Stream.empty
      )
    }

val fetch: Stream[F, ByteBuf] = ???
// wraps the `buf` in the `Chunk.ByteBuffer`
def toChunk(buf: ByteBuf): Chunk[Byte] = ???
```

When the Stream is very long and I found all `toChunk(b)` instance is referenced in the `Pull#BindBind` which can't be GC before this whole Stream ends and leaks memory a lot.

![image](https://user-images.githubusercontent.com/6748077/187018982-79744b83-fd4c-4882-9363-8121df911281.png)
[memory snapshot here](https://github.com/typelevel/fs2/files/9437477/Fs2Netty_39958_27.08.2022_15.27.38.hprof.zip)

If I use the below code instead it doesn't have this issue.
It seems the fs2's Steam#flatmap can all capture all instances in it so causing this issue.

One way with different semantics:
// moved the `reads` outside `flatmap`
[...]

PR #2970, opened 2022-08-27
Fix Memory Leak bug 2969.

diesalbla:
https://github.com/typelevel/fs2/issues/2969 uncovered some memory leaks that happen with the use of FlatMap alone. Using the middle buildR seems to fix it, just as it did for `Uncons` and `StepRun`.

_Pending work_: to add a memory leak test that reproduces the leak found in that issue.


mpilquist:
@diesalbla Why close?

diesalbla:
Because it did not fully solve the issue, and I do not know if it was improving.

mpilquist:
Ah okay!

PR #2971, opened 2022-08-30
Prevent socket leaks due to post-open exceptions

armanbilge:
Currently when opening an socket, other exception-throwing operations were attempted within the same `delay` block. If these operations threw an exception, the socket acquire would be deemed failed, and the resource finalizer which closes the socket would not run, thereby leaking the socket.

The acquire phase of the resource must end as soon as the socket is opened; this ensures that it will be closed even if there are subsequent errors.

H/t the native cross-build for discovering this one 😁

### DIFF
--- a/io/jvm/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
+++ b/io/jvm/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
@@ -48,12 +48,13 @@ private[net] trait SocketGroupCompanionPlatform { self: SocketGroup.type =>
         options: List[SocketOption]
     ): Resource[F, Socket[F]] = {
       def setup: Resource[F, AsynchronousSocketChannel] =
-        Resource.make(Async[F].delay {
-          val ch =
-            AsynchronousChannelProvider.provider.openAsynchronousSocketChannel(channelGroup)
-          options.foreach(opt => ch.setOption(opt.key, opt.value))
-          ch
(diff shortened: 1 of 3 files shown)


armanbilge:
Forgot to mention: I couldn't think of how to write a regression test for this one. Since it requires an error to occur when connecting the client / starting the server, we don't get a handle to the `Socket` in these cases and thus cannot verify if it is indeed closed.

io/jvm/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
```diff
@@ -80,24 +81,27 @@ private[net] trait SocketGroupCompanionPlatform { self: SocketGroup.type =>
...
 
-      val setup: F[AsynchronousServerSocketChannel] =
-        address.traverse(_.resolve[F]).flatMap { addr =>
-          Async[F].delay {
-            val ch =
-              AsynchronousChannelProvider.provider.openAsynchronousServerSocketChannel(channelGroup)
-            ch.bind(
-              new InetSocketAddress(
-                addr.map(_.toInetAddress).orNull,
-                port.map(_.value).getOrElse(0)
```
armanbilge, line 91: We exercise a `BindException` in our test suite, which was causing the open socket to leak.
https://github.com/typelevel/fs2/blob/0e573f2ca899886af57aaea9edd7a1072b4eb00a/io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala#L171-L177

Issue #2319, opened 2021-03-08
Add support for Scala Native


sluramod:
Cats (https://github.com/typelevel/cats) recently added support (https://github.com/typelevel/cats/issues/1549) for cross-compiling to scala native.

Scala Native support in fs2 would be be very helpful.

diesalbla:
This would of course depend on `cats-effect` support for Scala Native, as well. https://github.com/typelevel/cats-effect/issues/1302

armanbilge:
I started some preliminary investigation in https://github.com/armanbilge/fs2/commit/20a1dbb0159105c9e37106aa1e1b52bd74102863 based on https://github.com/typelevel/cats-effect/pull/3057.

Crossing fs2-core is straightforward. 
- porting `java.util.zip.InflaterOutputStream` to Scala Native would unblock compression

fs2-io is a lot more work. I focused only on the `fs2.io.file` package for now.
- `java.nio.file.Path#register` is missing
- so is `java.util.stream.Stream#forEach`

Without those, cross-compiling `fs2.io.file` will require complex splitting/platforming to remove `Watcher` (which is otherwise very nearly supported). We should port the missing APIs to Scala Native instead.

Furthermore, when running the fs2-io suite on Native I encountered the following failures.
```
X fs2.io.file.FilesSuite.permissions - innappropriate access should raise AccessDeniedException
X fs2.io.file.FilesSuite.delete - should fail on a non empty directory
X fs2.io.file.PathSuite.Monoid[Path]: monoid.right identity
```

The underlying bugs should all be fixed in Scala Native.

armanbilge:
PR is up at https://github.com/typelevel/fs2/pull/2975.

PR #2975, opened 2022-09-10
Scala Native

armanbilge:
Closes https://github.com/typelevel/fs2/issues/2319.

Supports:
- everything in `fs2-core`, except `hash`
- everything in scodec and protocols
- blocking (unfortunately) File I/O, since we don't have a blocking threadpool to run it on
- non-blocking TCP sockets, if a suitable implementation of `AsynchronousSocketChannel` is available at link-time, such as [epollcat](https://github.com/armanbilge/epollcat/)
- TLS for TCP sockets, via [s2n-tls](https://github.com/aws/s2n-tls)

Most of the changes in this PR are adding support for TLS. Everything else is code shuffling.

---

The TCP implementation is unopinionated: it shares the `AsynchronousSocketChannel` implementation with the JVM. `AsynchronousSocketChannel` is not implemented in Scala Native (and cannot be until it supports multithreading). This leaves the door open for 3rd parties to implement these APIs. Any implementation compliant with JDK should work.

We run our tests against the implementation provided in [epollcat](https://github.com/armanbilge/epollcat/) which supports Linux and macOS. A hypothetical implementation of these APIs based on [libuv](https://github.com/libuv/libuv/) and/or [scala-native-loop](https://github.com/scala-native/scala-native-loop/) could run on Windows.

---

On the other hand, the TLS implementation is very opinionated. It relies on [s2n-tls](https://github.com/aws/s2n-tls) and introduces configuration classes specifically based on its APIs and features. Note that s2n does not support Windows (https://github.com/aws/s2n-tls/issues/497). s2n is also minimalist by-design:
> s2n-tls avoids implementing rarely used options and extensions, as well as features with a history of triggering protocol-level vulnerabilities. For example there is no support for session renegotiation or DTLS.

In exchange, it offers a modern, simple API that was straightforward to integrate with. I think the trade-off is worth it.

The alternative, unopinionated approach would have been to use the JDK SSL APIs (essentially sharing the JVM code). These are not provided by Scala Native core, so it would be up to third parties to implement them. I explored this idea in [sn-ssl](https://github.com/armanbilge/sn-ssl/), but it proves to be a significant amount of work with a greater surface area for vulnerability.

---

Once CI is green, I will publish a hash version to Maven Central and continue onwards to http4s ember and skunk.


io/jvm/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -111,56 +105,4 @@ private[fs2] trait ioplatform {
     }
   }
 
-  //
-  // STDIN/STDOUT Helpers
```
armanbilge, line 115: This code is relocated to a trait shared across JVM/Native

io/jvm-native/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
```diff
@@ -51,7 +50,7 @@ private[net] trait SocketGroupCompanionPlatform { self: SocketGroup.type =>
         Resource
           .make(
             Async[F].delay(
-              AsynchronousChannelProvider.provider.openAsynchronousSocketChannel(channelGroup)
+              AsynchronousSocketChannel.open(channelGroup)
```
armanbilge, line 53: These are equivalent.

io/native/src/main/scala/fs2/io/net/Network.scala
```diff
@@ -0,0 +1,36 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2
+package io
+package net
+
+import fs2.io.net.tls.TLSContext
+
+sealed trait Network[F[_]] extends NetworkPlatform[F] with SocketGroup[F] {
```
armanbilge, line 28: `Network` does not implement `DatagramSocketGroup` at this time on Native, so it was easier to redefine it than attempt to share code.

io/native/src/main/scala/fs2/io/net/tls/S2nConfig.scala
```diff
@@ -0,0 +1,191 @@
...
+
+    def withCertChainAndKeysToStore(certKeyPairs: List[CertChainAndKey]): Builder
+    def withPemsToTrustStore(pems: List[String]): Builder
+    def withWipedTrustStore: Builder
+    def withSendBufferSize(size: Int): Builder
+    def withVerifyHostCallback(cb: String => SyncIO[Boolean]): Builder
+    def withDisabledX509Verification: Builder
+    def withMaxCertChainDepth(maxDepth: Short): Builder
+    def withDHParams(dhparams: String): Builder
+    def withCipherPreferences(version: String): Builder
```
armanbilge, line 62: s2n lets you specify your cipher preferences by selecting one of their pre-defined security policies.
https://github.com/aws/s2n-tls/blob/main/docs/USAGE-GUIDE.md#security-policies

Note that the default policy (aka "default") does not use TLS1.3; you have to opt-in with "default_tls13".

io/native/src/main/scala/fs2/io/net/tls/S2nConfig.scala
```diff
@@ -0,0 +1,191 @@
...
+    def withMaxCertChainDepth(maxDepth: Short): Builder = copy(maxCertChainDepth = Some(maxDepth))
+
+    def withDHParams(dhparams: String): Builder = copy(dhParams = Some(dhparams))
+
+    def withCipherPreferences(version: String): Builder = copy(cipherPreferences = Some(version))
+
+  }
+
+  // we manually schedule expensive private-key operations so we can add fairness boundaries
+  private def asyncPkeyCallback[F[_]](conn: Ptr[s2n_connection], op: Ptr[s2n_async_pkey_op]): CInt =
```
armanbilge, line 165: This is pretty cool, and something we cannot do on JVM :)

io/native/src/main/scala/fs2/io/net/tls/TLSSocketPlatform.scala
```diff
@@ -0,0 +1,102 @@
...
+private[tls] trait TLSSocketCompanionPlatform { self: TLSSocket.type =>
+
+  private[tls] def apply[F[_]: Async](
+      socket: Socket[F],
+      connection: S2nConnection[F]
+  ): Resource[F, TLSSocket[F]] =
+    Resource.eval(mk(socket, connection)) <*
+      Resource.makeFull[F, Unit](poll => poll(connection.handshake))(_ =>
+        connection.shutdown.attempt.void
+      )
```
armanbilge, line 45: The `.attempt` here is a bit unfortunate. It seems on the JVM `endOfOutput` may be called on the socket before closing the TLS connection, which makes graceful shutdown impossible. I assume this is incorrect use of APIs, but for sake of cross-compilability we suppress the error.

io/native/src/main/scala/fs2/io/net/tls/s2n.scala
```diff
@@ -0,0 +1,220 @@
...
+
+package fs2.io.net.tls
+
+import scala.scalanative.unsafe._
+import scala.annotation.nowarn
+
+@extern
+@link("s2n")
+@nowarn
+private[tls] object s2n {
```
armanbilge, line 30: This is the native binding, no implementation here.

io/native/src/main/scala/fs2/io/net/unixsocket/UnixSocketsPlatform.scala
```diff
@@ -0,0 +1,24 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2.io.net.unixsocket
+
+private[unixsocket] trait UnixSocketsCompanionPlatform
```
armanbilge, line 24: We don't implement Unix Sockets on Native yet, but this allows ember to cross-compile and opens the door for downstream implementations of the trait.

io/native/src/test/scala/fs2/io/net/tls/TLSSocketSuite.scala
```diff
@@ -0,0 +1,288 @@
...
+              clientSocket.reads.take(msg.size.toLong)
+          }
+          client.concurrently(echoServer)
+        }
+        .compile
+        .to(Chunk)
+        .intercept[SSLException]
+    }
+
+    test("mTLS client verification") {
```
armanbilge, line 175: So I don't repeat my mistake from https://github.com/typelevel/fs2/security/advisories/GHSA-2cpx-6pqp-wf35.

io/native/src/test/scala/fs2/io/net/tls/TLSSocketSuite.scala
```diff
@@ -0,0 +1,288 @@
...
+          }
+
+          client.concurrently(echoServer)
+        }
+        .compile
+        .to(Chunk)
+        .assertEquals(msg)
+    }
+
+    test("blinding") {
```
armanbilge, line 249: s2n implements "blinding".
> Blinding is a mitigation against timing side-channels which in some cases can leak information about encrypted data. By default s2n-tls will cause a thread to sleep between 10 and 30 seconds whenever tampering is detected.

To avoid blocking a thread, s2n also offers a "self-serivce" blinding option.

This test asserts two things:
1. Our "self-service" blinding does in fact work.
2. Instead of blocking a thread, it delegates to `IO.sleep`.

Unfortunately that means this test takes ~20 seconds to complete.

armanbilge:
I've published `3.2.14-75-7902cbf` to Maven Central.

If anyone with an M1 wants to run the fs2-io test suite locally I'd be much obliged :) you will need to `brew install s2n`.

mpilquist:
I tried on an M1 Mac and ran in to some test errors in TLSSockeSuite like this:
```scala
==> X fs2.io.net.tls.TLSSocketSuite.TLSSocket - google - default - client writes before reading 0.07s java.io.IOException: getaddrinfo: 8
```

armanbilge:
@mpilquist thanks for that, probably due to broken IPv6 in epollcat. Hopefully the latest commit should fix that (or at the very least, give a more verbose error :)

AesaKamar:
Just wanted to say you are a hero, @armanbilge!

Thanks for all your work!

PR #2979, opened 2022-09-16
Docs for utf8 decoding

TimWSpence:
Clarify a potentially confusing aspect of the decoding API

### DIFF
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -41,7 +41,12 @@ object text {
     private val utf8Charset = Charset.forName("UTF-8")
     private[this] val utf8BomSeq: Seq[Byte] = Array(0xef.toByte, 0xbb.toByte, 0xbf.toByte).toSeq
 
-    /** Converts UTF-8 encoded byte stream to a stream of `String`. */
+    /** Converts UTF-8 encoded byte stream to a stream of `String`.
+      *
+      * Note that the output stream is ''not'' a singleton stream but rather a stream
+      * of strings where each string is the result of UTF8 decoding a chunk of the
+      * underlying byte stream.
+      */
     def decode[F[_]]: Pipe[F, Byte, String] =
       _.chunks.through(decodeC)
 


TimWSpence:
CC @CremboC

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -41,7 +41,12 @@ object text {
     private val utf8Charset = Charset.forName("UTF-8")
     private[this] val utf8BomSeq: Seq[Byte] = Array(0xef.toByte, 0xbb.toByte, 0xbf.toByte).toSeq
 
-    /** Converts UTF-8 encoded byte stream to a stream of `String`. */
+    /** Converts UTF-8 encoded byte stream to a stream of `String`.
+      *
+      * Note that the output stream is ''not'' a singleton stream but rather a stream
+      * of strings where each string is the result of UTF8 decoding a chunk of the
+      * underlying byte stream.
+      */
```
armanbilge, line 49: Oh right, I can see how this was confusing 😅 thanks!

If we _really_ get nitty, IIUC it's not technically not one-stringt-per-`Chunk`, since some multi-byte characters could be split across `Chunk`s. Not sure if there's a good way to say that though, or if it really matters.

armanbilge (approved):

core/shared/src/main/scala/fs2/text.scala
TimWSpence: Oh haha ouch! I suspect attempting to explain that would cause more confusion rather than less but I'm very happy if someone can suggest a simple explanation for it!

diesalbla: It is even more complicated, since a chunk from the input could just be the middle bytes of a multi-byte character, so not a single character would be built off that chunk.

> The result is a stream of strings. Every chunk in the result contains exactly one string, and each string carries all characters that could be fully decoded from the input.

But I think it may be easier to just say 

> For the most part, each string in the output is the result of decoding a chunk of bytes from the input; however, this may not be accurate when the bytes of a multi-byte character are split amongst one or more input chunks.

PR #2981, opened 2022-09-17
Stream: improve evalMap / foreach / traverse

diesalbla:
- Singleton Chunk: override `traverse`, `traverse_`.
- Chunk: add improven tail-recursive implementation for the `traverse_` method. Avoid cats.Eval.
- Pull: add a special "unconsFlatMap" function, that represents a mapping of each chunk to a pull.
- Stream: use the flatMapCons to reimplement `chunks`, `evalMap`, `evalMapChunk`, `foreach`, and `intersperse`.
- Add a `foreachChunk` method.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -668,14 +668,12 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
           case Some(_) => F.unit
         }
 
-      def go(tl: Pull[F2, O, Unit]): Pull[F2, Nothing, Unit] =
-        Pull.uncons(tl).flatMap {
-          // Note: hd is non-empty, so hd.last.get is safe
-          case Some((hd, tl)) => Pull.eval(sendItem(hd.last.get)) >> go(tl)
-          case None           => Pull.eval(sendLatest >> chan.close.void)
-        }
+      def sendPull(hd: Chunk[O]): Pull[F2, Nothing, Unit] =
+        Pull.eval(sendItem(hd.last.get))
 
-      val debouncedSend: Stream[F2, Nothing] = new Stream(go(this.underlying))
+      val debouncedSend: Stream[F2, Nothing] =
+        (Pull.unconsFlatMap[F2, F2, O, Nothing](this.underlying)(sendPull) >>
+          Pull.eval(sendLatest >> chan.close.void)).streamNoScope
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1283,8 +1287,20 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * res0: Unit = ()
     * }}}
     */
-  def foreach[F2[x] >: F[x]](f: O => F2[Unit]): Stream[F2, Nothing] =
-    flatMap(o => Stream.exec(f(o)))
+  def foreach[F2[x] >: F[x]](f: O => F2[Unit]): Stream[F2, Nothing] = {
+    def goChunk(ch: Chunk[O], ix: Int, s: Int): Pull[F2, Nothing, Unit] =
+      if (ix >= s) Pull.done else Pull.eval(f(ch(ix))) >> goChunk(ch, ix + 1, s)
```
diesalbla, line 1292: Looking at the Pull compile code, we do not make a call to `interruptGuard` when interpreting a `Pull.Eval` node, so it may be there is no difference between running this or traversing the chunk (as done below).

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -368,6 +368,20 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
...
 
+  def traverse_[F[_]](f: O => F[Unit])(implicit F: Applicative[F]): F[Unit] =
+    if (isEmpty) F.unit
+    else
+      F match {
+        case m: FlatMap[F] =>
+          val s = size
+          def go(i: Int): F[Unit] =
+            // this is assuming our F is tail-recursive. Is that fair assumption?
+            if (i < s) m.flatMap(f(apply(i)))(_ => go(i + 1)) else F.unit
```
armanbilge, line 379: I don't understand this comment. Do you mean that `F` is stacksafe? In that case why not pattern match for `StackSafeMonad` ?

mpilquist:
Are we ready to merge this one? We still have the commented out code that breaks tests...

diesalbla:
> Are we ready to merge this one? We still have the commented out code that breaks tests...

No, I am marking as a draft.

diesalbla:
The failure in the tests regarding interruption seem to indicate some problem

diesalbla:
After #3015 and similar recent PRs, this may no longer be needed.

PR #2994, opened 2022-09-24
Aggressively short-circuit `processByteOrderMark` to avoid unnecessary `utf8.decode` buffering

armanbilge:
Fixes the buffering demonstrated in this snippet:
```scala
//> using lib "co.fs2::fs2-core::3.3.0"

import fs2._

@main def main = Stream[Pure, Byte](49, 50, 51, 52).chunkLimit(1).unchunks
  .through(text.utf8.decode).debugChunks().compile.drain
```

```
Chunk(1, 2, 3)
Chunk(4)
```

h/t @mn98 for the report.

### DIFF
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -171,8 +171,9 @@ object text {
                 if (newBuffer.startsWith(utf8BomSeq)) newBuffer.drop(3)
                 else newBuffer
               doPull(Chunk.empty, Stream.emits(rem.chunks) ++ tl)
-            } else
+            } else if (newBuffer.startsWith(utf8BomSeq.take(newBuffer.size)))
               processByteOrderMark(newBuffer, tl)
+            else doPull(Chunk.empty, Stream.emits(newBuffer.chunks) ++ tl)
           case None =>
             if (buffer ne null)
               doPull(Chunk.empty, Stream.emits(buffer.chunks))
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/text.scala
```diff
@@ -171,8 +171,9 @@ object text {
                 if (newBuffer.startsWith(utf8BomSeq)) newBuffer.drop(3)
                 else newBuffer
               doPull(Chunk.empty, Stream.emits(rem.chunks) ++ tl)
-            } else
+            } else if (newBuffer.startsWith(utf8BomSeq.take(newBuffer.size)))
```
diesalbla, line 174: Depending on the definition of `take`, could this create a new object of type seq with any call to this method?

Perhaps it would help to do the following: 

- Extract the array as a private variable. 
- Write a method to compare a number of characters of a given array with that array. 

```Scala
private[this] val utf8BomArr: Array[Byte] = Array(0xef.toByte, 0xbb.toByte, 0xbf.toByte)

private[this] def zipsEquals(buffer: Chunk[Byte]): Boolean = 
  /* compare that either buffer is prefix of utf8BomArr or the other way around*/
```

```Scala
```

armanbilge: Yes, I agree 😕 I didn't want to get bogged down in micro-optimizations but if we really care about performance what we want to do here is directly compare to `utf8BomArr(0)`, `utf8BomArr(1)`, `utf8BomArr(2)` without doing any `.startsWith` or `take`.

Counter-point: anyone with a stream that reads byte-by-byte is already drowned in significant overhead (chunk allocations, `F`-interpretation, `Pull`-interpretation). The true fast path is when the `Chunk` actually has a non-trivial size, in which case this condition will never be hit.

So all in all I just don't think it matters.

diesalbla (approved):

core/shared/src/main/scala/fs2/text.scala
diesalbla: Well, I guess that may very well be the case.

PR #2995, opened 2022-09-25
Chunk - Introduce a Constant Chunk subclass.

diesalbla:
When creating a constant stream, out of constant chunks, we do not need to allocate a list of 256 nodes with the same value. instead we can just write a chunk as a constant function.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -550,6 +550,31 @@ object Chunk
     override def map[O2](f: O => O2): Chunk[O2] = singleton(f(value))
   }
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -546,6 +546,32 @@ object Chunk
     override def map[O2](f: O => O2): Chunk[O2] = singleton(f(value))
   }
 
+  def constant[A](value: A, size: Int): Chunk[A] = size match {
```
mpilquist, line 549: Add case for size < 0?

diesalbla: Modified code to cover that case, too.

PR #3000, opened 2022-09-29
Update to Cats Effect 3.2.0-RC1

armanbilge:
This PR replaces deprecated `Dispatcher[IO]` uses with the appropriate `sequential` or `parallel` replacements.

Inline comments forthcoming.


io/jvm/src/main/scala/fs2/io/JavaInputOutputStream.scala
```diff
@@ -191,7 +191,7 @@ private[io] object JavaInputOutputStream {
      * - DownStream signal -  keeps any remainders from last `read` and signals
      *                        that downstream has been terminated that in turn kills upstream
      */
-    Dispatcher[F].flatMap { dispatcher =>
+    Dispatcher.sequential[F].flatMap { dispatcher =>
```
armanbilge, line 194: The `dispatcher` is used only for reading and closing. That seems sequential to me.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscription.scala
```diff
@@ -76,18 +77,17 @@ private[reactivestreams] final class StreamSubscription[F[_], A](
...
-  // request()`, then the request _must_ be a no op. For this reason, we
-  // need to make sure that `cancel()` does not return until the
-  // `cancelled` signal has been set.
+  // request()`, then the request _must_ be a no op. Fortunately,
+  // ordering is guaranteed by a sequential d
   // See https://github.com/zainab-ali/fs2-reactive-streams/issues/29
   // and https://github.com/zainab-ali/fs2-reactive-streams/issues/46
   def cancel(): Unit =
-    dispatcher.unsafeRunSync(cancelled.set(true))
+    requestDispatcher.unsafeRunAndForget(cancelled.set(true))
```
armanbilge, line 90: So this is interesting. Previously `unsafeRunSync` was necessary to achieve this behavior. Now the shared sequential `Dispatcher` guarantees the required ordering out-of-the box.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscription.scala
```diff
@@ -41,7 +41,8 @@ private[reactivestreams] final class StreamSubscription[F[_], A](
     cancelled: SignallingRef[F, Boolean],
     sub: Subscriber[A],
     stream: Stream[F, A],
-    dispatcher: Dispatcher[F]
+    startDispatcher: Dispatcher[F],
+    requestDispatcher: Dispatcher[F]
```
armanbilge, line 45: We need two `Dispatcher`s for `StreamSubscription`, but they can both be sequential.

The first is used in `unsafeStart()` and actually runs the subscription. So it's blocked on that.

The second is used for handling requests and canceling.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamUnicastPublisher.scala
```diff
@@ -56,9 +57,15 @@ final class StreamUnicastPublisher[F[_]: Async, A](
...
       s: Stream[F, A],
       dispatcher: Dispatcher[F]
   ): StreamUnicastPublisher[F, A] =
-    new StreamUnicastPublisher(s, dispatcher)
+    new StreamUnicastPublisher(s, dispatcher, dispatcher)
+
+  def apply[F[_]: Async, A](
+      s: Stream[F, A]
+  ): Resource[F, StreamUnicastPublisher[F, A]] =
+    (Dispatcher.sequential[F], Dispatcher.sequential[F]).mapN(new StreamUnicastPublisher(s, _, _))
```
armanbilge, line 70: Since we can now have local `Dispatcher`s with special sequential semantics, it seems like we should deprecate the pattern of asking for (presumably more global) dispatchers in favor of building our own dispatchers.

Furthermore, since we now rely on the sequential `Dispatcher` to achieve the correct semantics, exposing it as an argument is not good design.

reactive-streams/src/test/scala/fs2/interop/reactivestreams/UnsafeTestNGSuite.scala
```diff
@@ -1,44 +0,0 @@
...
-  private var shutdownDispatcher: IO[Unit] = _
-
-  private val mkDispatcher = Dispatcher[IO].allocated
-  private val t = mkDispatcher.unsafeRunSync()
-  dispatcher = t._1
-  shutdownDispatcher = t._2
-
-  @AfterClass
-  def afterAll() = shutdownDispatcher.unsafeRunSync()
-}
```
armanbilge, line 44: This machinery was no longer necessary since `Dispatcher` is no longer accepted as an argument.

io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -199,9 +199,11 @@ private[fs2] trait ioplatform {
       in: Stream[F, Byte]
   )(implicit F: Async[F]): Resource[F, (Duplex, Stream[F, Byte])] =
     for {
-      dispatcher <- Dispatcher[F]
+      readDispatcher <- Dispatcher.sequential[F]
+      writeDispatcher <- Dispatcher.sequential[F]
+      errorDispatcher <- Dispatcher.sequential[F]
```
armanbilge, line 204: read/write are sequential and error is one-off, but they need to be independent, so they need their own dispatchers.

build.sbt
```diff
@@ -176,6 +176,9 @@ ThisBuild / mimaBinaryIssueFilters ++= Seq(
   ),
   ProblemFilters.exclude[ReversedMissingMethodProblem](
     "fs2.io.net.tls.TLSContext#Builder.insecureResource"
+  ),
+  ProblemFilters.exclude[DirectMissingMethodProblem]( // something funky in Scala 3.2.0 ...
+    "fs2.io.net.SocketGroupCompanionPlatform#AsyncSocketGroup.this"
   )
```
armanbilge, line 182: Super weird, can reproduce just by updating to 3.2.0 on `main`. No idea ...

armanbilge: I minimized it in https://github.com/armanbilge/sandbox/commit/d9453ac24ec96253b61431f0b30b529e353037b5. IDK what change in dotty caused it, but seems safe: seems like the private constructor was generating public bytecode prior to 3.2.0, but is now properly private. This is flagging up MiMa.

It doesn't affect anything public.

armanbilge:
Lol, I mis-titled this and the commit as 3.2.0-RC1 instead of 3.4.0-RC1. oh well.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamUnicastPublisher.scala
mtomko: I'm sorry to come in to this so late in the game after this has merged, but we use this in a Play application to stream CSV files from an fs2 `Stream`; the Play API requires a class that's hard to construct except by using a `StreamUnicastPublisher` but it doesn't give us the opportunity to call close. In previous discussions on discord, I was advised to create a `Dispatcher` for the application and pass it in; we can guarantee the provided `Dispatcher` gets closed at the end.

I see (superficially) the point of your comment here and the rationale behind deprecating the API that takes a provided `Dispatcher`, but I don't see how we can migrate to the non-deprecated API without leaking `Dispatcher` instances. I will try to find you on discord sometime to talk about it a little more.

armanbilge: @mtomko no problem, thanks for the comment.

> but I don't see how we can migrate to the non-deprecated API without leaking `Dispatcher` instances.

I need to think about it more, but so long as you are careful to cancel the subscription, I don't think there will be any leak here. The `Resource` surrounding a `Dispatcher` is intended to make sure that any tasks it started are canceled when the resource closes. If a `Dispatcher` has no running tasks and it goes out of scope, I am fairly certain that the GC will completely clean it up.

Also @BalmungSan is currently working on some usability improvements that might help your usecase )

mtomko: I'm trying to trace through the code, but it means digging deeply into akka streams and Play. The code _we've_ written just creates the source and passes it to Play's `Ok` to construct a result:

```scala
  def streamToClientAsCsvFile(
      enumerator: Stream[IO, Row],
      dispatcher: Dispatcher[IO],
      filename: String
  ): Result = {
    val publisher = StreamUnicastPublisher(enumerator.through(rowToCsv), dispatcher)
    val source = Source.fromPublisher(publisher)

    Ok.chunked(source)
      .withHeaders(CONTENT_DISPOSITION -> s"attachment; filename = $filename")
      .as("text/csv")
  }
```

I would like to hope that Play is smart enough to recognize when the connection is closed and close the `Source`.

mtomko: Ps, thanks for your response, it's really appreciated!

BalmungSan: @mtomko #3107 is the PR @armanbilge mentioned. However, I am not exactly sure how it helps in your use case.

armanbilge: > However, I am not exactly sure how it helps in your use case.

@BalmungSan @mtomko the new `stream.subscribe(...)` syntax that Luis is adding in https://github.com/typelevel/fs2/pull/3107 will enable you to bring back the global `Dispatcher`.

PR #3008, opened 2022-10-04
Use CharsetEncoder for encoding texts

### LINKED ISSUE #3006: fs2.text.encode adds too many BOMs for UTF-16
Reproducer: https://scastie.scala-lang.org/kPRHfJXRQCyB4Y9GYgziVQ

```scala
Stream("1", "2", "3").through(encode(StandardCharsets.UTF_16)).compile.to(List)
// gives: List(-2, -1, 0, 49, -2, -1, 0, 50, -2, -1, 0, 51)
```

The `-2, -1` slices are BOMs added as `StandardCharsets.UTF_16` emits them when encoding. Only the first one is valid, the ones at the string boundaries should not be emitted.

ybasket:
Avoids problems like multiple BOMs when encoding with a charset that adds them (UTF-16).

Opening as a draft as there's a problem with backwards compatibility, see inline comments.

Closes #3006

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/TextBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/TextBenchmark.scala
@@ -36,7 +36,7 @@ class TextBenchmark {
   var asciiBytes: Array[Byte] = _
   var strings: Array[String] = _
 
-  @Param(Array("utf-8", "iso-2022-kr"))
+  @Param(Array("utf-8", "utf-16", "iso-2022-kr"))
   var charsetName: String = _
   var charset: Charset = _
 
--- a/benchmark/src/main/scala/fs2/benchmark/TextBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/TextBenchmark.scala
@@ -46,7 +46,9 @@ class TextBenchmark {
     asciiBytes = (0 until asciiStringSize).map { _ =>
       (rng.nextInt(126) + 1).toByte
     }.toArray
-    strings = new String(asciiBytes).grouped(16).toArray // chunk it up to be more real-world
+    strings = new String(asciiBytes)
+      .grouped(asciiStringSize / 16) // chunk it up to be more real-world
+      .toArray
     charset = Charset.forName(charsetName)
   }
 
--- a/benchmark/src/main/scala/fs2/benchmark/TextBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/TextBenchmark.scala
@@ -26,18 +26,28 @@ import cats.effect.IO
 import cats.effect.unsafe.implicits.global
 import org.openjdk.jmh.annotations.{Benchmark, Param, Scope, Setup, State}
 
+import java.nio.charset.Charset
+
 @State(Scope.Thread)
 class TextBenchmark {
-  @Param(Array("128", "1024", "4096"))
+  @Param(Array("128", "1024", "4096", "16384", "131072"))
   var asciiStringSize: Int = _
 
   var asciiBytes: Array[Byte] = _
+  var strings: Array[String] = _
+
+  @Param(Array("utf-8", "iso-2022-kr"))
+  var charsetName: String = _
+  var charset: Charset = _
+
   @Setup
   def setup(): Unit = {
     val rng = new java.util.Random(7919)
     asciiBytes = (0 until asciiStringSize).map { _ =>
       (rng.nextInt(126) + 1).toByte
     }.toArray
+    strings = new String(asciiBytes).grouped(16).toArray // chunk it up to be more real-world
+    charset = Charset.forName(charsetName)
   }
 
   @Benchmark
@@ -49,4 +59,12 @@ class TextBenchmark {
       .last
       .unsafeRunSync()
       .get
+
+  @Benchmark
+  def asciiEncode(): Array[Byte] =
+    Stream
+      .emits(strings)
+      .through(text.encode(charset))
+      .compile
+      .to(Array)
 }
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -27,12 +27,15 @@ import java.nio.charset.{
   CharacterCodingException,
   Charset,
   CharsetDecoder,
+  CharsetEncoder,
   MalformedInputException,
   StandardCharsets,
   UnmappableCharacterException
 }
 import scala.collection.mutable.{ArrayBuffer, Builder}
 import scodec.bits.{Bases, ByteVector}
+
+import scala.annotation.tailrec
 
 /** Provides utilities for working with streams of text (e.g., encoding byte streams to strings). */
 object text {
@@ -305,12 +308,100 @@ object text {
     utf8.decodeC
 
   /** Encodes a stream of `String` in to a stream of bytes using the given charset. */
-  def encode[F[_]](charset: Charset): Pipe[F, String, Byte] =
-    _.mapChunks(c => c.flatMap(s => Chunk.array(s.getBytes(charset))))
+  def encode[F[_]](charset: Charset): Pipe[F, String, Byte] = {
+    def encodeC(
+        encoder: CharsetEncoder,
+        acc: Chunk[Char],
+        s: Stream[F, Chunk[Char]],
+        lastOutBuffer: ByteBuffer
+    ): Pull[F, Byte, Unit] =
+      s.pull.uncons1.flatMap { r =>
+        val toEncode = r match {
+          case Some((c, _)) => acc ++ c
+          case None         => acc
+        }
+
+        val isLast = r.isEmpty
+        (lastOutBuffer: Buffer).clear()
+
+        val outBufferSize = (encoder.maxBytesPerChar() * toEncode.size).ceil.toInt
+
+        val out =
+          if (outBufferSize > lastOutBuffer.remaining())
+            ByteBuffer.allocate(outBufferSize)
+          else lastOutBuffer
+
+        val inBuffer = toEncode.toCharBuffer
+        val result = encoder.encode(inBuffer, out, isLast)
+        (out: Buffer).flip()
+
+        val nextAcc =
+          if (inBuffer.remaining() > 0) Chunk.charBuffer(inBuffer.slice()) else Chunk.empty
+
+        val rest = r match {
+          case Some((_, tail)) => tail
+          case None            => Stream.empty
+        }
+
+        if (result.isError) {
+          ApplicativeThrow[Pull[F, Byte, *]].raiseError {
+            if (result.isMalformed) new MalformedInputException(result.length())
+            else if (result.isUnmappable) new UnmappableCharacterException(result.length())
+            else new CharacterCodingException()
+          }
+        } else if (out.remaining() > 0) {
+          Pull.output(Chunk.byteBuffer(out)) >> encodeC(encoder, nextAcc, rest, out)
+        } else if (!isLast) {
+          encodeC(encoder, nextAcc, rest, out)
+        } else if (nextAcc.nonEmpty) {
+          encodeC(
+            encoder,
+            nextAcc,
+            rest,
+            ByteBuffer.allocate(
+              outBufferSize + (encoder.maxBytesPerChar() * nextAcc.size).toInt
+            )
+          )
+        } else flush(encoder, lastOutBuffer)
+      }
+
+    @tailrec
+    def flush(
+        encoder: CharsetEncoder,
+        out: ByteBuffer
+    ): Pull[F, Byte, Unit] = {
+      (out: Buffer).clear()
+      encoder.flush(out) match {
+        case res if res.isUnderflow =>
+          if (out.position() > 0) {
+            (out: Buffer).flip()
+            Pull.output(Chunk.byteBuffer(out)) >> Pull.done
+          } else
+            Pull.done
+        case res if res.isOverflow =>
+          val newSize = (out.capacity + encoder.maxBytesPerChar() * 2).toInt
+          val bigger = ByteBuffer.allocate(newSize)
+          flush(encoder, bigger)
+        case res =>
+          ApplicativeThrow[Pull[F, Byte, *]].catchNonFatal(res.throwException())
+      }
+    }
+
+    { s =>
+      Stream.suspend(Stream.emit(charset.newEncoder())).flatMap { encoder =>
+        encodeC(
+          encoder,
+          Chunk.empty,
+          s.map(s => Chunk.charBuffer(CharBuffer.wrap(s))),
+          ByteBuffer.allocate(0)
+        ).stream
+      }
+    }
+  }
 
   /** Encodes a stream of `String` in to a stream of `Chunk[Byte]` using the given charset. */
   def encodeC[F[_]](charset: Charset): Pipe[F, String, Chunk[Byte]] =
-    _.mapChunks(_.map(s => Chunk.array(s.getBytes(charset))))
+    _.through(encode[F](charset)).chunks // TODO: Why?
 
   /** Encodes a stream of `String` in to a stream of bytes using the UTF-8 charset. */
   @deprecated("Use text.utf8.encode", "3.1.0")
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -331,7 +331,7 @@ object text {
         val out = ByteBuffer.allocate(outBufferSize)
 
         val inBuffer = toEncode.toCharBuffer
-        val result = encoder.encode(inBuffer, out, isLast)
+        encoder.encode(inBuffer, out, isLast)
         (out: Buffer).flip()
 
         val nextAcc =
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -41,9 +41,25 @@ import scala.annotation.tailrec
 /** Provides utilities for working with streams of text (e.g., encoding byte streams to strings). */
 object text {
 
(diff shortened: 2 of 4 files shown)


core/shared/src/main/scala/fs2/text.scala
```diff
@@ -305,12 +308,100 @@ object text {
...
+        val nextAcc =
+          if (inBuffer.remaining() > 0) Chunk.charBuffer(inBuffer.slice()) else Chunk.empty
+
+        val rest = r match {
+          case Some((_, tail)) => tail
+          case None            => Stream.empty
+        }
+
+        if (result.isError) {
+          ApplicativeThrow[Pull[F, Byte, *]].raiseError {
```
ybasket, line 347: This will throw an exception. The current behaviour seems to rather replace characters with some placeholder, but on decoding there's a `RaiseThrowable` constraint. So maybe this should actually be named `encodeWithCharset` and the current implementations deprecated. Or I need to dig into mimicking the old behaviour. Opinions?

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -305,12 +308,100 @@ object text {
...
+          ByteBuffer.allocate(0)
+        ).stream
+      }
+    }
+  }
 
   /** Encodes a stream of `String` in to a stream of `Chunk[Byte]` using the given charset. */
   def encodeC[F[_]](charset: Charset): Pipe[F, String, Chunk[Byte]] =
-    _.mapChunks(_.map(s => Chunk.array(s.getBytes(charset))))
+    _.through(encode[F](charset)).chunks // TODO: Why?
```
ybasket, line 404: This implementation obviously works, but I wondered whether there were any performance considerations? Why does this method exist, are there use cases for that type or is it optimization?

core/shared/src/main/scala/fs2/text.scala
armanbilge: Sorry, can you explain more about this? The docs for `CharsetEncoder` say that

> How an encoding error is handled depends upon the action requested for that type of error, which is described by an instance of the [CodingErrorAction](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CodingErrorAction.html) class. The possible error actions are to [ignore](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CodingErrorAction.html#IGNORE) the erroneous input, [report](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CodingErrorAction.html#REPORT) the error to the invoker via the returned [CoderResult](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CoderResult.html) object, or [replace](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CodingErrorAction.html#REPLACE) the erroneous input with the current value of the replacement byte array. The replacement is initially set to the encoder's default replacement, which often (but not always) has the initial value { (byte)'?' }; its value may be changed via the [replaceWith](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CharsetEncoder.html#replaceWith-byte:A-) method.

https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CharsetEncoder.html

ybasket: The default with `CharsetEncoder` is to raise an error, different from what `String.getBytes` does. See this little Scastie: https://scastie.scala-lang.org/2pOZGh6PQBiJ8TQyd5bYkQ

It seems totally possible to retain the old behaviour and just replace any non-mappable character, but it is somewhat inconsistent with `decode` which deals with errors by making them visible in `F`. So this PR would be a good opportunity to revisit this and potentially align both ways (which isn't straight-forward due to bin/src compat).

ybasket: Just pushed b010a950 that restores the replacing behaviour.

mpilquist:
@armanbilge Any ideas on the native test failures in `FilesSuite`? https://github.com/typelevel/fs2/actions/runs/3183721692/jobs/5229431645#step:11:35886

armanbilge:
Darn, missed that. I guess these changes have somehow hit a bug in Native Javalib 🤔

armanbilge:
- https://github.com/scala-native/scala-native/pull/2884

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -306,11 +310,88 @@ object text {
...
+        s: Stream[F, Chunk[Char]]
+    ): Pull[F, Chunk[Byte], Unit] =
+      s.pull.uncons1.flatMap { r =>
+        val toEncode = r match {
+          case Some((c, _)) => acc ++ c
+          case None         => acc
+        }
+
+        val isLast = r.isEmpty
+        val outBufferSize = (encoder.maxBytesPerChar() * toEncode.size).ceil.toInt
```
ybasket, line 329:
```suggestion
        val outBufferSize = math.max(encoder.maxBytesPerChar(), encoder.averageBytesPerChar() * toEncode.size).toInt
```

would be more memory-efficient, especially with UTF-8 (average: 1.1, max: 4). BUT: It breaks the UTF-8 roundtrip test as it can affect chunking if not all chars can directly be written to the output. I'm not sure whether chunking is actually something that needs to be preserved strictly or whether that tests ensures this more by accident?

benchmark/src/main/scala/fs2/benchmark/TextBenchmark.scala
```diff
@@ -49,4 +59,12 @@ class TextBenchmark {
       .last
       .unsafeRunSync()
       .get
+
+  @Benchmark
+  def asciiEncode(): Array[Byte] =
```
ybasket, line 64: Results from running `benchmark/jmh:run -i 3 -wi 2 -f1 -t1 fs2.benchmark.TextBenchmark` on a M1 Macbook with Temurin 17.

Current PR code:

```
[info] Benchmark                  (asciiStringSize)  (charsetName)   Mode  Cnt       Score       Error  Units
[info] TextBenchmark.asciiEncode                128          utf-8  thrpt    3  122702,219 ± 13773,670  ops/s
[info] TextBenchmark.asciiEncode                128    iso-2022-kr  thrpt    3  119576,150 ±  2410,238  ops/s
[info] TextBenchmark.asciiEncode               1024          utf-8  thrpt    3   23868,525 ±   233,647  ops/s
[info] TextBenchmark.asciiEncode               1024    iso-2022-kr  thrpt    3   23077,995 ±   150,249  ops/s
[info] TextBenchmark.asciiEncode               4096          utf-8  thrpt    3    6537,079 ±   148,060  ops/s
[info] TextBenchmark.asciiEncode               4096    iso-2022-kr  thrpt    3    6092,286 ±   320,573  ops/s
[info] TextBenchmark.asciiEncode              16384          utf-8  thrpt    3    1648,121 ±    23,967  ops/s
[info] TextBenchmark.asciiEncode              16384    iso-2022-kr  thrpt    3    1542,366 ±    10,599  ops/s
[info] TextBenchmark.asciiEncode             131072          utf-8  thrpt    3     190,055 ±    10,821  ops/s
[info] TextBenchmark.asciiEncode             131072    iso-2022-kr  thrpt    3     186,668 ±    25,596  ops/s
```

Slight memory optimization from [comment](https://github.com/typelevel/fs2/pull/3008/files#r990769246):

```
[info] Benchmark                  (asciiStringSize)  (charsetName)   Mode  Cnt       Score      Error  Units
[info] TextBenchmark.asciiEncode                128          utf-8  thrpt    3  124311,351 ± 4393,854  ops/s
[info] TextBenchmark.asciiEncode                128    iso-2022-kr  thrpt    3  123602,921 ± 8403,831  ops/s
[info] TextBenchmark.asciiEncode               1024          utf-8  thrpt    3   23800,856 ± 5362,019  ops/s
[info] TextBenchmark.asciiEncode               1024    iso-2022-kr  thrpt    3   23402,259 ± 1177,082  ops/s
[info] TextBenchmark.asciiEncode               4096          utf-8  thrpt    3    6412,433 ±   42,169  ops/s
[info] TextBenchmark.asciiEncode               4096    iso-2022-kr  thrpt    3    6258,803 ±  384,576  ops/s
[info] TextBenchmark.asciiEncode              16384          utf-8  thrpt    3    1616,414 ±   50,169  ops/s
[info] TextBenchmark.asciiEncode              16384    iso-2022-kr  thrpt    3    1539,644 ±   53,911  ops/s
[info] TextBenchmark.asciiEncode             131072          utf-8  thrpt    3     183,273 ±    8,099  ops/s
[info] TextBenchmark.asciiEncode             131072    iso-2022-kr  thrpt    3     194,158 ±    9,477  ops/s
```

3.3.0 implementation:

```
[info] Benchmark                  (asciiStringSize)  (charsetName)   Mode  Cnt       Score       Error  Units
[info] TextBenchmark.asciiEncode                128          utf-8  thrpt    3  459940,502 ± 62479,321  ops/s
[info] TextBenchmark.asciiEncode                128    iso-2022-kr  thrpt    3  313636,796 ± 84643,728  ops/s
[info] TextBenchmark.asciiEncode               1024          utf-8  thrpt    3  167493,377 ± 16709,846  ops/s
[info] TextBenchmark.asciiEncode               1024    iso-2022-kr  thrpt    3   74485,196 ±  2612,422  ops/s
[info] TextBenchmark.asciiEncode               4096          utf-8  thrpt    3   52266,272 ±   714,696  ops/s
[info] TextBenchmark.asciiEncode               4096    iso-2022-kr  thrpt    3   21595,360 ±   198,851  ops/s
[info] TextBenchmark.asciiEncode              16384          utf-8  thrpt    3   15428,697 ±   240,052  ops/s
[info] TextBenchmark.asciiEncode              16384    iso-2022-kr  thrpt    3    5746,304 ±   126,641  ops/s
[info] TextBenchmark.asciiEncode             131072          utf-8  thrpt    3    1842,642 ±    24,347  ops/s
[info] TextBenchmark.asciiEncode             131072    iso-2022-kr  thrpt    3     631,334 ±    33,201  ops/s
```

So 3x-10x slower on UTF-8, iso-2022-kr is 3x-4x. Not cool 😞 

`getBytes` for UTF-8 seems impossible to reach as the JVM has a lot of performance hacks, boiling down to `System.arraycopy` for ASCII-compatible strings and alike. But I was hoping for being a lot closer on larger inputs and exotic charsets. Haven't done any profiling (yet), could imagine there's an unfortunate use of NIO buffers that leads to copying hidden somewhere. Would appreciate any hint by people that have more experience!

If the performance stays in that ballpark, then we could differentiate by charset and fast path to the old implementation if it doesn't affect correctness (either by picking charsets known to have no BOM or by picking those known to have one – the Java stdlib doesn't seem to provide this information anywhere).

core/shared/src/main/scala/fs2/text.scala
armanbilge: > I'm not sure whether chunking is actually something that needs to be preserved strictly or whether that tests ensures this more by accident?

I don't think so, see this note.

https://github.com/typelevel/fs2/blob/51c976855db1afa73c3ff6f6f919aa131076d8c7/core/shared/src/main/scala/fs2/Stream.scala#L117-L137

benchmark/src/main/scala/fs2/benchmark/TextBenchmark.scala
ybasket: I just pushed a new implementation that implements three strategies chosen by the charset in use, based on a discussion with @armanbilge:
* fastpath UTF-8 and charsets without BOM through the old implementation using `getBytes`
* run some known charsets with BOM through `getBytes` and slice the BOMs away (not yet as fast as I hoped for, may need some optimisation)
* fallback to the original implementation of this PR with `CharsetEncoder` as this is the only method known to be correct

Benchmark results (some have high error bars, my machine might have done something during that time):

PR as it is now with the three strategies:

```
[info] Benchmark                  (asciiStringSize)  (charsetName)   Mode  Cnt       Score       Error  Units
[info] TextBenchmark.asciiEncode                128          utf-8  thrpt    3  241078,756 ±  4214,301  ops/s
[info] TextBenchmark.asciiEncode                128         utf-16  thrpt    3   79392,707 ±  3475,150  ops/s
[info] TextBenchmark.asciiEncode                128    iso-2022-kr  thrpt    3   78863,440 ±  6695,626  ops/s
[info] TextBenchmark.asciiEncode               1024          utf-8  thrpt    3  243723,717 ± 17416,529  ops/s
[info] TextBenchmark.asciiEncode               1024         utf-16  thrpt    3   76299,562 ±  2787,154  ops/s
[info] TextBenchmark.asciiEncode               1024    iso-2022-kr  thrpt    3   68631,477 ± 58520,349  ops/s
[info] TextBenchmark.asciiEncode               4096          utf-8  thrpt    3  220528,726 ±  8569,189  ops/s
[info] TextBenchmark.asciiEncode               4096         utf-16  thrpt    3   69097,673 ±  3413,480  ops/s
[info] TextBenchmark.asciiEncode               4096    iso-2022-kr  thrpt    3   45814,458 ±  2688,061  ops/s
[info] TextBenchmark.asciiEncode              16384          utf-8  thrpt    3  177451,798 ± 34625,941  ops/s
[info] TextBenchmark.asciiEncode              16384         utf-16  thrpt    3   39671,226 ±  6648,031  ops/s
[info] TextBenchmark.asciiEncode              16384    iso-2022-kr  thrpt    3   19581,045 ±  1254,195  ops/s
[info] TextBenchmark.asciiEncode             131072          utf-8  thrpt    3   49748,453 ±  3362,374  ops/s
[info] TextBenchmark.asciiEncode             131072         utf-16  thrpt    3    3248,993 ±   122,217  ops/s
[info] TextBenchmark.asciiEncode             131072    iso-2022-kr  thrpt    3    3443,736 ±    22,870  ops/s
```

3.3.0 implementation:

```
[info] Benchmark                  (asciiStringSize)  (charsetName)   Mode  Cnt       Score        Error  Units
[info] TextBenchmark.asciiEncode                128          utf-8  thrpt    3  383735,294 ±  27890,377  ops/s
[info] TextBenchmark.asciiEncode                128         utf-16  thrpt    3  276145,684 ±   9260,096  ops/s
[info] TextBenchmark.asciiEncode                128    iso-2022-kr  thrpt    3  223451,698 ±  61534,850  ops/s
[info] TextBenchmark.asciiEncode               1024          utf-8  thrpt    3  371536,523 ± 117132,997  ops/s
[info] TextBenchmark.asciiEncode               1024         utf-16  thrpt    3  258765,503 ±  50539,873  ops/s
[info] TextBenchmark.asciiEncode               1024    iso-2022-kr  thrpt    3  183692,517 ±   3334,342  ops/s
[info] TextBenchmark.asciiEncode               4096          utf-8  thrpt    3  347988,866 ±  37393,800  ops/s
[info] TextBenchmark.asciiEncode               4096         utf-16  thrpt    3  165533,415 ±   3172,584  ops/s
[info] TextBenchmark.asciiEncode               4096    iso-2022-kr  thrpt    3  108587,378 ±  17709,535  ops/s
[info] TextBenchmark.asciiEncode              16384          utf-8  thrpt    3  242168,751 ±  39448,685  ops/s
[info] TextBenchmark.asciiEncode              16384         utf-16  thrpt    3   64262,596 ±   3758,674  ops/s
[info] TextBenchmark.asciiEncode              16384    iso-2022-kr  thrpt    3   38383,447 ±   2458,706  ops/s
[info] TextBenchmark.asciiEncode             131072          utf-8  thrpt    3   54766,444 ±   4596,102  ops/s
[info] TextBenchmark.asciiEncode             131072         utf-16  thrpt    3    3754,077 ±    194,735  ops/s
[info] TextBenchmark.asciiEncode             131072    iso-2022-kr  thrpt    3    5880,160 ±     65,146  ops/s
```

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -27,19 +27,39 @@ import java.nio.charset.{
...
 import scodec.bits.{Bases, ByteVector}
 
+import scala.annotation.tailrec
+
 /** Provides utilities for working with streams of text (e.g., encoding byte streams to strings). */
 object text {
 
+  /** Byte order mark (BOM) values for different Unicode charsets.
+    */
+  object bom {
```
ybasket, line 46: As they can be useful downstream (for example, I needed them in http4s-fs2-data) I expose them here, but happy to convert to something private if people think this isn't the right place.

core/shared/src/main/scala/fs2/text.scala
```diff
@@ -306,11 +326,111 @@ object text {
...
+
+    def cutBOMs(
+        s: Stream[F, String],
+        bom: Seq[Byte],
+        doDrop: Boolean
+    ): Pull[F, Chunk[Byte], Unit] =
+      s.pull.uncons1.flatMap {
+        case Some((hd, tail)) =>
+          val bytes = Chunk.array(hd.getBytes(charset))
+          val dropped = if (doDrop && bytes.startsWith(bom)) bytes.drop(bom.length) else bytes
```
armanbilge, line 401: `.startsWith` is not particularly optimized, esp. since we are comparing seqs with boxed bytes. No idea if it adds up to something meaningful.

core/shared/src/main/scala/fs2/text.scala
armanbilge: The problem is that `Seq[Byte]` is a really inefficient representation. I wonder if we should use a `ByteVector` instead.

ybasket:
I now optimised as much as I can and would say the performance is fine now, see https://github.com/typelevel/fs2/pull/3008#discussion_r992148309. It also passes all the added tests.

Just one thing left to resolve: As @armanbilge fixed a bug in Scala Native that would be hit when using the new implementation with a charset that is not UTF-8 or UTF-16, but has a BOM or variable length, this PR seems not safe to merge until a new Scala Native release. Unless we introduced platform-specific code to work around it somehow.

armanbilge:
We will get a Scala Native release this week. Anyway, FS2 cannot make a stable release while it's stuck on CE release candidates.

PR #3015, opened 2022-10-16
EvalMap - Quick Optimisation.

diesalbla:
In the implementation of `evalMap`, we expand the `Stream.eval` and the `Stream.flatMap` methods to directly work on Pull. This saves creating one ephemeral `Stream` object (it is a non-AnyVal wrapper) for each element in the output stream.

Maybe this kind of optimisation is already done by the JVM itself, but it is not too long to do on the code.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -974,8 +974,10 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * which has performance implications. For maximum performance, `evalMapChunk`
     * is available, however, with caveats.
     */
-  def evalMap[F2[x] >: F[x], O2](f: O => F2[O2]): Stream[F2, O2] =
-    flatMap(o => Stream.eval(f(o)))
+  def evalMap[F2[x] >: F[x], O2](f: O => F2[O2]): Stream[F2, O2] = {
+    def evalOut(o: O): Pull[F2, O2, Unit] = Pull.eval(f(o)).flatMap(Pull.output1)
+    Pull.flatMapOutput[F, F2, O, O2](underlying, evalOut).streamNoScope
+  }
 
   /** Like `evalMap`, but operates on chunks for performance. This means this operator
     * is not lazy on every single element, rather on the chunks.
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -974,8 +974,10 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * which has performance implications. For maximum performance, `evalMapChunk`
     * is available, however, with caveats.
     */
-  def evalMap[F2[x] >: F[x], O2](f: O => F2[O2]): Stream[F2, O2] =
-    flatMap(o => Stream.eval(f(o)))
+  def evalMap[F2[x] >: F[x], O2](f: O => F2[O2]): Stream[F2, O2] = {
+    def evalOut(o: O): Pull[F2, O2, Unit] = Pull.eval(f(o)).flatMap(Pull.output1)
+    Pull.flatMapOutput[F, F2, O, O2](underlying, evalOut).streamNoScope
```
mpilquist, line 979: Not as part of this PR, but we should probably make `flatMapOutput` a (private?) method on `Pull`.

diesalbla: The problem is that it only applies for a pull where the result type is `Unit`.

armanbilge: You can require an evidence `R =:= Unit` to use the method.

diesalbla: > You can require an evidence R =:= Unit to use the method.

I avoid evidence parameters, since both extension methods in Scala 3 or even implicit classes in Scala 2 are, IMHO, cleaner. We can try that.

armanbilge: Personal taste :) extension methods are much harder on discoverability, I would only reach for them when it is actually impossible to add the method to the type (i.e. you do not control it). I think it is easy to discover a method defined on the type, and the evidence parameter makes it clear what its constraints are.

diesalbla: I guess the reason is: extension methods have the benefit of giving up-front the type pattern (e.g. the binding of R to Unit) on which they apply, so it almost read as "if your pull has a return type unit, then define `foo`", rather than "we have method `foo` over the list of type parameters, value parameters, and implicit parameters, provided that...". My personal taste or sense of style is to put conditions first, and in this case the 

Likewise, the extension class (already exists) looks to me as "for all Pulls such that their result type is Unit, let there be defined such methods as..."

PR #3016, opened 2022-10-16
Pull: turn flatMapOutput into an extension method.

diesalbla:
We may like method calls, even extension methods, better than functions. Plus, we need fewer type arguments in the calls.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -321,6 +321,16 @@ object Pull extends PullLowPriority {
       * Use `p.void.stream` to explicitly ignore the result of a pull.
       */
     def streamNoScope: Stream[F, O] = new Stream(self)
+
+    private[fs2] def flatMapOutput[F2[x] >: F[x], O2](
+        f: O => Pull[F2, O2, Unit]
+    ): Pull[F2, O2, Unit] =
+      self match {
+        case a: AlgEffect[F, Unit] => a
+        case r: Terminal[_]        => r
+        case _                     => FlatMapOutput(self, f)
+      }
+
   }
 
   private[this] val unit: Terminal[Unit] = Succeeded(())
@@ -1029,7 +1039,7 @@ object Pull extends PullLowPriority {
           else {
             def go(idx: Int): Pull[G, X, Unit] =
               if (idx == chunk.size)
-                flatMapOutput[G, G, Y, X](tail, fun)
+                tail.flatMapOutput(fun)
               else {
                 try {
                   var j = idx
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -962,6 +962,8 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       queue: Queue[F2, Option[Chunk[O2]]]
   ): Stream[F2, Nothing] = enqueueNoneTerminatedChunks(queue: QueueSink[F2, Option[Chunk[O2]]])
 
+  import Pull.StreamPullOps
```
armanbilge, line 965: Why is the import necessary, is it not in implicit scope automatically?

diesalbla: Failed to compile in Scala 3.

armanbilge (approved):

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -1288,16 +1297,6 @@ object Pull extends PullLowPriority {
...
 
-  private[fs2] def flatMapOutput[F[_], F2[x] >: F[x], O, O2](
-      p: Pull[F, O, Unit],
-      f: O => Pull[F2, O2, Unit]
-  ): Pull[F2, O2, Unit] =
-    p match {
-      case a: AlgEffect[F, Unit] => a
-      case r: Terminal[_]        => r
-      case _                     => FlatMapOutput(p, f)
-    }
```
armanbilge, line 1299: Instead of adding a MiMa filter, can we just make this deprecated and delegate it to the new implementation?

diesalbla: Hmmmm if it is a private[fs2] method, how could a binary incompatibility happen?

armanbilge: It's because of https://github.com/lampepfl/dotty/issues/10842.

It would be safe to add a filter, but every filter we add risks accidentally masking a real issue. So my personal preference is to avoid filters even for private stuff if there's an easy fix.

diesalbla: Not ideal, but so be it. Added back.

core/shared/src/main/scala/fs2/Pull.scala
```diff
@@ -1288,15 +1297,12 @@ object Pull extends PullLowPriority {
     go(initScope, None, initFk, new OuterRun(init), stream)
   }
 
+  @deprecated("use the extension method", "3.3.1")
```
armanbilge, line 1300: The `tlBaseVersion` is already 3.4.0 (b/c of the Scala 3.2 bump)

```suggestion
  @deprecated("use the extension method", "3.4.0")
```

PR #3021, opened 2022-10-17
Add 'Stream.ToPull.unconsMin', remove 'Stream.ToPull.unconsLimit' ass…

seigert:
I've added `Stream.ToPull.unconsMin` and changed logic of both `Stream.ToPull.unconsN` and `Stream.chunkMin` to use it.

Main motivation: if you working with some 'frames' of size N, sometimes you also can process chunks of M * N size in a batch, right now it is possible only via (afaik) `.chunkMin(N).scan(Chunk.empty)(..).unchunk`. New operation allows this directly on pull.

Also I've removed an assertion of `n > 0` on `unconsLimit` as all other similar methods (`unconsN, chunkMin, chunkM`) to do not fail and return empty output instead.

---

On a side note, usage of both `repeatPull` and `unconsFlatMap` in `chunk*` operators and other methods is confusing at best. Especially as 'unconsFlatMap' to not use the operations defined in `ToPull`. 

For example, current implementation of `unconsLimit` will fail on negative `n` while `chunkLimit` will produce infinite stream of empty chunks.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -434,31 +434,13 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * res0: List[Chunk[Int]] = List(Chunk(1, 2, 3, 4), Chunk(5, 6, 7))
     * }}}
     */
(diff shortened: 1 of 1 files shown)


mpilquist:
Looks great!

PR #3024, opened 2022-10-22
Specialize `parEvalMapUnbounded`

armanbilge:
Uses a `Channel.unbounded` for `parEvalMapUnbounded` instead of a `Channel.bounded(Int.MaxValue)`.

Also adds `parEvalMapUnorderedUnbounded`.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2090,12 +2090,10 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       maxConcurrent: Int
   )(f: O => F2[O2])(implicit F: Concurrent[F2]): Stream[F2, O2] = {
 
-    val init = F.unit
-
     def send(ch: Channel[F2, F2[Either[Throwable, O2]]], release: F2[Unit]) =
       (el: Either[Throwable, O2]) => release <* ch.send(el.pure[F2])
 
-    parEvalMapAction(maxConcurrent, f)((ch, release) => init.as(send(ch, release)))
+    parEvalMapAction(maxConcurrent, f)((ch, release) => send(ch, release).pure[F2])
   }
 
   /** Like parEvalMapUnordered but with unbounded concurrency.
@@ -2104,12 +2102,10 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       f: O => F2[O2]
   )(implicit F: Concurrent[F2]): Stream[F2, O2] = {
 
-    val init = F.unit
-
     def send(ch: Channel[F2, F2[Either[Throwable, O2]]], release: F2[Unit]) =
       (el: Either[Throwable, O2]) => release <* ch.send(el.pure[F2])
 
-    parEvalMapUnboundedAction(f)((ch, release) => init.as(send(ch, release)))
+    parEvalMapUnboundedAction(f)((ch, release) => send(ch, release).pure[F2])
   }
 
   private def parEvalMapAction[F2[x] >: F[x], O2, T](
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2062,8 +2062,18 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     */
   def parEvalMapUnbounded[F2[x] >: F[x], O2](f: O => F2[O2])(implicit
       F: Concurrent[F2]
-  ): Stream[F2, O2] =
-    parEvalMap[F2, O2](Int.MaxValue)(f)
+  ): Stream[F2, O2] = {
+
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2080,14 +2090,28 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+  def parEvalMapUnorderedUnbounded[F2[x] >: F[x], O2](
+      f: O => F2[O2]
+  )(implicit F: Concurrent[F2]): Stream[F2, O2] = {
+
+    val init = F.unit
+
+    def send(ch: Channel[F2, F2[Either[Throwable, O2]]], release: F2[Unit]) =
+      (el: Either[Throwable, O2]) => release <* ch.send(el.pure[F2])
+
+    parEvalMapUnboundedAction(f)((ch, release) => init.as(send(ch, release)))
```
diesalbla, line 2112: Is there any reason to use `init.as(----)` instead of `F.pure(send(ch, release))`?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2062,8 +2062,18 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
-    parEvalMap[F2, O2](Int.MaxValue)(f)
+  ): Stream[F2, O2] = {
+
+    def init(ch: Channel[F2, F2[Either[Throwable, O2]]], release: F2[Unit]) =
+      Deferred[F2, Either[Throwable, O2]].flatTap { value =>
+        ch.send(release *> value.get)
+      }
+
+    def send(v: Deferred[F2, Either[Throwable, O2]]) =
+      (el: Either[Throwable, O2]) => v.complete(el).void
```
diesalbla, line 2073: Rather than a function that builds a `lambda`, could we make this two parameter lists?
```suggestion
    def send(v: Deferred[F2, Either[Throwable, O2]])(el: Either[Throwable, O2]): F[Unit] =
      v.complete(el).void
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2080,14 +2090,28 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       maxConcurrent: Int
   )(f: O => F2[O2])(implicit F: Concurrent[F2]): Stream[F2, O2] = {
 
-    val init = ().pure[F2]
+    val init = F.unit
 
     def send(ch: Channel[F2, F2[Either[Throwable, O2]]], release: F2[Unit]) =
       (el: Either[Throwable, O2]) => release <* ch.send(el.pure[F2])
 
     parEvalMapAction(maxConcurrent, f)((ch, release) => init.as(send(ch, release)))
```
diesalbla, line 2098: Could we just reduce this to `F.pure(send(ch, release))` and drop the `init` val?

core/shared/src/main/scala/fs2/Stream.scala
armanbilge: I don't know, this is just copy-paste of existing code. My best guess is some stack-safety trick.

core/shared/src/main/scala/fs2/Stream.scala
armanbilge: I think this change will cause an unnecessary indirection, since ultimately we need a lambda.

```scala
def send(v: Deferred[F2, Either[Throwable, O2]])(el: Either[Throwable, O2]): F[Unit]
```

will desugar to
```scala
def send(v: Deferred[F2, Either[Throwable, O2]], el: Either[Throwable, O2]): F[Unit]
```
So then we still need to allocate a lambda that then delegates to `send`.

Whereas the current version allocates a lambda that directly invokes `v.complete(el).void`, no indirection.

core/shared/src/main/scala/fs2/Stream.scala
armanbilge: Same comment as above, maybe it's a stack-safety issue?

nikiforo: It wasn't intended to add stack-safety. Instead, it's written this way to resemble the logic in parEvalMap

nikiforo: So yeah, if it raises question of why it is written this way, I have no argument against `F.pure(send(ch, release))`

armanbilge: Thanks @nikiforo, much appreciated!

satorg: Yeah, I am also quite curious about that approach. Does not seem to be stack-safer than just `F.pure(...)`.

armanbilge: It was probably a stupid thing of me to say in this case :) but tricks with `F.unit` for stack-safety are not uncommon in FS2.

PR #3028, opened 2022-10-27
Add accumulating variant of `Stream.debounce`

m50d:
I don't know if there's enough demand/need for this, but I had a use case where I needed something like `debounce` but that keeps all the elements of the stream, and it seemed generic enough to be worth sharing.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -663,6 +663,50 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       chan.stream.concurrently(debouncedSend)
     }
   }
+ 
+  /**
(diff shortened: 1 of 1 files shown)


m50d:
Appreciate there were a number of failures that were my fault, but I can't understand the latest one (native linking failure) - what's the issue here?

armanbilge:
Just a flake :)

mpilquist:
Any thoughts on `NonEmptyChain` versus `Chunk` here?

m50d:
It wasn't a particularly thought-through choice, just the type I reached for. For my specific use case a type that expresses that it's always non-empty is helpful, but happy to switch it to `Chunk` if you think that's more useful in the general case.

mpilquist:
OK cool, I think we should switch to `Chunk` then. There are lots of APIs that would benefit from a `NonEmptyChunk` indication, though we tried that for a while and it didn't end up working out that well.

PR #3031, opened 2022-10-28
Add `SignallingMapRef`

armanbilge:
Based on https://github.com/typelevel/fs2/pull/3030.

A signalling extension of Cats Effect `MapRef`.


armanbilge:
I ended up not using the lens stuff from https://github.com/typelevel/fs2/pull/3030, since it's wildly inefficient: an update at any key in the map would send updates for all keys.

I have replaced it with a new implementation that maintains listeners per-key.

PR #3032, opened 2022-10-30
Relax constraint for `Signal.constant`

armanbilge:


SystemFw (commented):
A bit unsure if this is of any value. The `Concurrent` constraint is everywhere in fs2, and I don't see the difference between Spawn and Concurrent to be that meaningful for users. I can see how Temporal vs Concurrent might help, but not really this one. Open to changing my mind ofc

armanbilge:
FWIW I completely agree, I doubt this would make a difference for anyone. Certainly not to me.

Was this `Concurrent` by mistake, or intentional? Or maybe it's intentional now? e.g.. if we don't merge this should I open an issue so we remember to fix this in FS2 4 or something 😂

diesalbla:
@armanbilge If you fix PR, happy to merge it if you still think it of use.

PR #3039, opened 2022-11-02
Add `Stream#{hold1,hold1Resource}`

armanbilge:
existing:
- `hold` takes an initial value for the signal
- `holdOption` uses `None` as the initial value for the signal

this PR:
- `hold1` blocks until the stream emits its first value, then emits a signal with that initial value

Name (and API) subject to bikeshed of course.


diesalbla (approved):

armanbilge:
I think this operator is the dual of `Signal#discrete`, as in it lets your recover the original signal from the `discrete` stream.

PR #3063, opened 2022-11-22
Fix `async_` usages for new uncancelable semantic

armanbilge:
Canaries https://github.com/typelevel/cats-effect/pull/3205.

### DIFF
--- a/io/jvm-native/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
@@ -32,8 +32,10 @@ import java.nio.channels.{
 }
 import java.nio.channels.AsynchronousChannelGroup
 import cats.syntax.all._
+import cats.effect.syntax.all._
 import cats.effect.kernel.{Async, Resource}
 import com.comcast.ip4s.{Host, IpAddress, Port, SocketAddress}
+import fs2.concurrent.Channel
 
 private[net] trait SocketGroupCompanionPlatform { self: SocketGroup.type =>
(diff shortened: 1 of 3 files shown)


io/jvm-native/src/main/scala/fs2/io/net/SocketGroupPlatform.scala
```diff
@@ -105,16 +109,20 @@ private[net] trait SocketGroupCompanionPlatform { self: SocketGroup.type =>
...
+                    new CompletionHandler[AsynchronousSocketChannel, Void] {
+                      def completed(ch: AsynchronousSocketChannel, attachment: Void): Unit =
+                        cb(Right(ch))
+                      def failed(rsn: Throwable, attachment: Void): Unit =
+                        cb(Left(rsn))
+                    }
+                  )
                 }
-              )
+                .as(Some(Async[F].delay(sch.close())))
```
armanbilge, line 125: There's potentially some weird life-cycle stuff going on here: AFAIK the only way to cancel an on-going `accept` is to close the server socket. As the signature is roughly `Resource[F, Stream[F, Socket]]`, where the server socket is tied to the resource, in theory you might expect to be able to cancel the inner stream and then start it again.

One way to fix this is to introduce a `Channel.synchronous` as an intermediary: that way, the stream can be canceled gracefully, without forcing to kill the server socket.

Thoughts?

djspiewak:
I can take care of the faraday cage here in a bit.

djspiewak:
Actually I can't because I don't have publish rights to `co.fs2`. Someone else has to.

armanbilge:
I'm actually doing it right now :)

PR #3064, opened 2022-11-22
Docs/microsite: Added/cleaned imports in the Single Publisher / Mutiple Subscriber example

codesurf42:
INothing obsolete since fs2 3.3
Leaving `import scala.language.higherKinds` as needed for Scala versions lower than 2.13.1


mpilquist:
Thanks!

PR #3066, opened 2022-11-25
Add Files.createLink for hard links

skennedy:
A bit of a rare use case compared to symlinks - but it was straightforward enough to implement on jvm and js


armanbilge:
Thanks for the PR!

Damn, looks like it's broken in Scala Native. I guess we can check for an `IOException` for now and file an upstream issue.
```
==> X fs2.io.file.FilesSuite.createLink - fails with NoSuchFileException if the target doesn't exist 0.02s munit.FailException: intercept failed, exception 'java.io.IOException' is not a subtype of 'java.nio.file.NoSuchFileException
```

skennedy:
Thanks for looking into that :rocket: 

I loosened the check for now.

armanbilge (approved):

PR #3070, opened 2022-11-30
Cross `hash` for JS and Native

armanbilge:
- On JS, Node.js only. This is the sole Node.js-only thing in core, but meh. Web Crypto doesn't support streaming hashing yet. Further reading:
  - https://github.com/w3c/webcrypto/issues/73
  - https://github.com/wintercg/proposal-webcrypto-streams

- On Native, requires OpenSSL.

### DIFF
--- a/core/native/src/main/scala/fs2/hash.scala
+++ b/core/native/src/main/scala/fs2/hash.scala
@@ -1,0 +1,130 @@
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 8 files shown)


core/shared/src/test/scala/fs2/HashSuite.scala
```diff
@@ -45,30 +42,30 @@ class HashSuite extends Fs2Suite {
...
   }
 
   group("digests") {
-    test("md2")(forAll((s: String) => checkDigest(md2, "MD2", s)))
-    test("md5")(forAll((s: String) => checkDigest(md5, "MD5", s)))
-    test("sha1")(forAll((s: String) => checkDigest(sha1, "SHA-1", s)))
-    test("sha256")(forAll((s: String) => checkDigest(sha256, "SHA-256", s)))
-    test("sha384")(forAll((s: String) => checkDigest(sha384, "SHA-384", s)))
-    test("sha512")(forAll((s: String) => checkDigest(sha512, "SHA-512", s)))
+    if (isJVM) test("md2")(forAllF((s: String) => checkDigest(md2, "MD2", s)))
```
armanbilge, line 49: So, crypto is weird, since ultimately it depends on availability of algorithms at runtime. Seems that default OpenSSL builds (also used by Node.js) don't include MD2. But an alternative build could.

core/native/src/main/scala/fs2/hash.scala
```diff
@@ -0,0 +1,130 @@
...
+    digest(EVP_get_digestbyname(c"SHA384"))
+
+  /** Computes a SHA-512 digest. */
+  def sha512[F[_]: Sync]: Pipe[F, Byte, Byte] =
+    digest(EVP_get_digestbyname(c"SHA512"))
+
+  /** Computes the digest of the source stream, emitting the digest as a chunk
+    * after completion of the source stream.
+    */
+  private[this] def digest[F[_]](digest: => Ptr[EVP_MD])(implicit F: Sync[F]): Pipe[F, Byte, Byte] =
```
armanbilge, line 58: So I caved and added a `Sync` constraint here, unlike the JVM/JS implementations. Since we are allocating and `free`ing native memory I didn't see a way around it.

This might end up being annoying in practice, especially for cross-builds. Maybe a capability typeclass `Hash[F]` would be a better design.

PR #3073, opened 2022-12-01
Simplify creating `Path` in custom `FileSystem`

### LINKED ISSUE #3072: Allow for reading from resources in `Files[F[_]]`
Hi,
In one of the projects I'm working on we've encountered a problem with using `Files` to read from jar resources. We used `Files` to read our configuration files with the hope that they can either be external or internal to the jar. 
The idea was to get an arbitrary `Path` and then read the file not getting into details where it comes from. Then we realized it's not as easy as it seems, as reading from jar using `Files` ends up with `java.nio.file.NoSuchFileException`. We still wanted to hide the logic of figuring out where to look for the file and made a workaround.

We'd like to propose a target solution though. It seems `Files` could use jar resources as a fallback, when it can't find the file in the current way. I think this will make working with (resource) files way more easier. We can come up with a draft implementation but let me know you thoughts first, how do you like the idea?

This has been somewhat mentioned [here](https://github.com/typelevel/fs2/issues/2503) and [here (in the discussion)](https://github.com/typelevel/fs2/pull/2733)

matkob:
Resolves #3072

### DIFF
--- a/io/jvm-native/src/main/scala/fs2/io/file/Path.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/file/Path.scala
@@ -23,7 +23,7 @@ package fs2
 package io
 package file
 
-import java.nio.file.{FileSystems, Path => JPath, Paths}
+import java.nio.file.{FileSystem, FileSystems, Path => JPath, Paths}
 
 /** Path to a file or directory on a file system.
   *
@@ -88,4 +88,5 @@ object Path extends PathCompanionApi {
   private def apply(path: JPath): Path = new Path(path)
   def apply(path: String): Path = fromNioPath(FileSystems.getDefault.getPath(path))
   def fromNioPath(path: JPath): Path = Path(path)
+  def fromFsPath(fs: FileSystem, path: String): Path = fromNioPath(fs.getPath(path))
 }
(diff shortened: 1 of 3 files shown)


mpilquist:
Thanks!

matkob:
You're welcome!

io/jvm-native/src/main/scala/fs2/io/file/Path.scala
```diff
@@ -88,4 +88,5 @@ object Path extends PathCompanionApi {
   private def apply(path: JPath): Path = new Path(path)
   def apply(path: String): Path = fromNioPath(FileSystems.getDefault.getPath(path))
   def fromNioPath(path: JPath): Path = Path(path)
+  def fromFsPath(fs: FileSystem, path: String): Path = fromNioPath(fs.getPath(path))
```
armanbilge, line 91: `getPath` is a var-args, should this API match?
https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPath-java.lang.String-java.lang.String...-

matkob: I guess I chose consistency with other `Path` constructors over API match here. In the end user can use other `Path` functions to get the same result.

armanbilge: Good point! :)

PR #3080, opened 2022-12-14
Fix last-update tracking in `SignallingMapRef`

armanbilge:
This refactoring of `SignallingMapRef` fixes a bug with the `lastUpdate` counter that caused "spurious" updates to be emitted at keys that had not actually been updated.

This could be observed as a sort of race condition, when a listener at one key was unable to resubscribe before there was an update at another key. For example:
1. a listener on key `k1` is notified of an update. it records the `lastUpdate`
2. key `k2` is updated, and `lastUpdate` is incremented
3. the listener on `k1` attempts to resubscribe and sees that `lastUpdate` does not match its records. So it assumes there was an update (there wasn't, at least not at `k1`) and emits a duplicate value. Woops.

To fix this, we need to have a separate `lastUpdate` counter for each key. But that will quickly become a memory leak. 

Specifically, we are concerned with the case where a key has no value in the map and has no listeners. In that case, it should have no memory footprint, which means we cannot keep a `lastUpdate` counter for that key.

The proposed workaround is to ~~fudge~~ relax the semantics for these "deleted" keys.
- For a vanilla `SignallingRef` setting the same value twice-in-a-row will emit two updates, thanks to the `lastUpdate` counter.
- For a `SignallingMapRef`, if a key has no value (i.e. its value is `None`), its `lastUpdate` is fixed at `-1`. This means that deleting a key twice-in-a-row will only issue one update. It also means that if the last value a listener observed at a key was `None`, and by the time it resubscribes the current value is also `None`, it will not receive an update, even if there were several non-`None` updates in the window before it was able to resubscribe. Small price to pay.

  Of course, if a key _does_ have a value, semantics are exactly like `SignallingRef`.

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/Signal.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Signal.scala
@@ -403,38 +403,50 @@ object SignallingMapRef {
       initial: Map[K, V] = Map.empty[K, V]
   )(implicit F: Concurrent[F]): F[SignallingMapRef[F, K, Option[V]]] = {
     case class State(
-        value: Map[K, V],
-        keyState: Map[K, KeyState]
+        lastUpdate: Long,
+        keys: Map[K, KeyState]
     )
 
-    // fix lastUpdate at -1 when the key is not present in the value map
-    // so we don't have to keep an update counter when the key has neither a value nor any listeners
-    case class KeyState(lastUpdate: Long, listeners: LongMap[Deferred[F, (Option[V], Long)]])
+    type Listener = Deferred[F, (Option[V], Long)]
 
-    F.ref(State(initial, initial.flatMap(_ => Nil)))
+    // fix lastUpdate at -1 when the value is not present
+    // so we don't have to keep an update counter when the key has neither a value nor any listeners
+    case class KeyState(
+        value: Option[V],
+        lastUpdate: Long,
+        listeners: LongMap[Listener]
+    )
+
+    F.ref(State(0L, initial.map { case (k, v) => k -> KeyState(Some(v), 0L, LongMap.empty) }))
       .product(F.ref(1L))
       .map { case (state, ids) =>
         def newId = ids.getAndUpdate(_ + 1)
 
         def updateAndNotify[U](state: State, k: K, f: Option[V] => (Option[V], U))
             : (State, F[U]) = {
-          val (newValue, result) = f(state.value.get(k))
 
-          val newMap = newValue.fold(state.value - k)(v => state.value + (k -> v))
+          val keyState = state.keys.get(k)
 
-          val keyState = state.keyState.get(k)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -390,31 +390,65 @@ trait SignallingMapRef[F[_], K, V] extends MapRef[F, K, V] {
...
-              listener.complete(newValue -> lastUpdate)
+
+          val keyState = state.keys.get(k)
+
+          val (newValue, result) = f(keyState.flatMap(_.value))
+
+          val lastUpdate = {
+            val lu = state.lastUpdate + 1
+            // skip -1 b/c of its special semantic
+            if (lu == -1L) 0L else lu
```
filipwiech, line 437: Sorry if it's a stupid question, but should we increment `lu` (`state.lastUpdate + 1`) before checking for `-1`? 🤔 
Also, can we ever get `lastUpdate == -1` in the overarching `State` (not in the `KeyState` for each key)? 🙂

armanbilge: > but should we increment `lu` (`state.lastUpdate + 1`) before checking for `-1`?

Yes, if we increment and it's `-1` then we want to skip. If the current value is `-1` then we've already failed 😉 



> Also, can we ever get `lastUpdate == -1` in the overarching `State` (not in the `KeyState` for each key)?

Every update causes `state.lastUpdate` to increment. Additionally, `state.keys(k).lastUpdate` takes the same value when `k` is updated.

filipwiech:
Would it be possible to add unit test(s) for the memory leaks in case of keys with no value and no listeners? Some scenario which creates and removes keys from the `SignallingMapRef` a lot of times (ideally covering the logic in both `updateAndNotify` and `cleanup`) and would OOM if the implementation wouldn't handle it? 🙂

core/shared/src/main/scala/fs2/concurrent/Signal.scala
filipwiech: Thanks, I think I understand it now. 👍

core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -449,19 +488,23 @@ object SignallingMapRef {
...
+                  state.keys.get(k).fold(state) { case KeyState(value, lastUpdate, listeners) =>
+                    val newListeners = listeners - id
+                    val newKeys =
+                      if (value.isEmpty && newListeners.isEmpty)
+                        state.keys - k // prevent memory leak
+                      else
+                        state.keys.updated(k, KeyState(value, lastUpdate, newListeners))
+                    state.copy(keys = newKeys)
+                  }
                 }
```
diesalbla, line 501: Could we extract this lambda-block into a method of the `State` class? It only consists of non-effectful F code.

diesalbla (approved):

PR #3082, opened 2022-12-15
Fix native deps in build

armanbilge:
I added an optional openssl dep to `coreNative` for hashing in https://github.com/typelevel/fs2/pull/3070. Although the dependency is not needed if hashing is not used, the scodec and protocols modules' tests depend on the test scope of core, which does use hash, and thus openssl becomes a transitive dependency. So it must be configured for them as well.

h/t @TimWSpence


TimWSpence:
Thanks for fixing! :)

PR #3087, opened 2022-12-23
Use CE polling system for non-blocking I/O on Native

armanbilge:
This PR adopts the new polling system APIs from https://github.com/typelevel/cats-effect/pull/3314 to implement non-blocking TCP/unix sockets and stdin/stdout/stderr for Scala Native. This means that epollcat is no longer required: everything now works out of the box with `IOApp` and the default `IORuntime`.

Much of the code was cribbed from epollcat and fs2-io_uring (with thanks to @leetibbert). Some of it can be upstreamed to ip4s or even completely removed when we upgrade to Scala Native 0.5.

This PR is tangled up with the new uncancelable semantics in CE 3.5 so it will need https://github.com/typelevel/fs2/pull/3063 to merge first. It also needs a Scala Native release with https://github.com/scala-native/scala-native/pull/3015 which will also unblock `DatagramSocket`.


armanbilge:
Tests are passing locally on both Linux and macOS (will need to add this to CI going forward). Published `3.5-b0f71fe-SNAPSHOT`.

https://s01.oss.sonatype.org/content/repositories/snapshots/co/fs2/fs2-io_native0.4_3/3.5-b0f71fe-SNAPSHOT/

armanbilge:
TODO need to port this fix.
- https://github.com/armanbilge/epollcat/pull/113

armanbilge:

Issue #3071, opened 2022-11-30
`TopicSuite` is flaking frequently


armanbilge:
```
==> X fs2.concurrent.TopicSuite.unregister subscribers under concurrent load  0.035s munit.ComparisonFailException: /home/runner/work/fs2/fs2/core/shared/src/test/scala/fs2/concurrent/TopicSuite.scala:113
112:        .map { result =>
113:          assertEquals(result.toMap.size, subs / 2)
114:          assertEquals(result.toMap, expected)
values are not the same
=> Obtained
6
=> Diff (- obtained, + expected)
-6
+5
```

armanbilge:
Does anyone know if this test is about _concurrent_ load, or _parallel_ load? If it only needs concurrency, and not true parallelism, it looks like a good candidate for `executeEmbed(...)` to avoid the flakiness.

PR #3088, opened 2022-12-24
Add singletonOrError

ybasket:
Add `CompileOps.singletonOrError` to compile singleton streams while preserving the intent and raising errors if the stream is not a singleton (which indicates a bug – advantage over `lastOrError`).

Inspired by a Discord discussion: https://discord.com/channels/632277896739946517/632310980449402880/1046091945564459070

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4791,6 +4791,36 @@ object Stream extends StreamLowPriority {
     def lastOrError(implicit G: MonadError[G, Throwable]): G[O] =
       last.flatMap(_.fold(G.raiseError(new NoSuchElementException): G[O])(G.pure))
 
+    /** Compiles this stream in to a value of the target effect type `G`,
+      * raising a `NoSuchElementException` if the stream emitted no values, `IllegalStateException`
+      * if the stream emits more than one value and returning the only value emitted otherwise.
+      * Note that the stream execution will not short-circuit once it emits more than one element, it will still
+      * be evaluated to its end before raising the exception.
+      *
+      * When this method has returned, the stream has not begun execution -- this method simply
+      * compiles the stream down to the target effect type.
+      *
+      * @example {{{
+      * scala> import cats.effect.SyncIO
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4791,6 +4791,36 @@ object Stream extends StreamLowPriority {
...
+      * scala> import cats.effect.SyncIO
+      * scala> Stream(1).covary[SyncIO].compile.singletonOrError.unsafeRunSync()
+      * res0: Int = 1
+      * scala> Stream.empty.covaryAll[SyncIO, Int].compile.singletonOrError.attempt.unsafeRunSync()
+      * res1: Either[Throwable, Int] = Left(java.util.NoSuchElementException)
+      * scala> Stream.range(0,10).covary[SyncIO].compile.singletonOrError.attempt.unsafeRunSync()
+      * res2: Either[Throwable, Int] = Left(java.lang.IllegalStateException: Expected singleton stream)
+      * }}}
+      */
+    def singletonOrError(implicit G: MonadError[G, Throwable]): G[O] =
```
ybasket, line 4813: Name is open for discussion. Doobie uses `unique` on queries, but I felt `singleton` being the more used term for streams.

ybasket:
The CI failed for `TopicSuite`, so unrelated to this PR. Maybe someone can restart the job so that we see CI running through?

armanbilge:
> The CI failed for `TopicSuite`

Probably this 😕 
- https://github.com/typelevel/fs2/issues/3071

core/shared/src/main/scala/fs2/Stream.scala
mpilquist: Thoughts on `.compile.only`?

ybasket: My first version was actually named `onlyOrError`. I would keep the `OrError` for consistency with `lastOrError` (and because there's a reasonable implementation of `only` returning `Option`), but other than that, happy to rename…

mpilquist: OK, I prefer `onlyOnError` then :)

PR #3090, opened 2022-12-25
Increase timeout for Native compression suite

armanbilge:
Saw this timeout a couple times. Let's see if this helps, or maybe it's a legit issue 🤔


diesalbla (approved):

diesalbla:
Let us see.

PR #3091, opened 2022-12-26
Implement I/O with CE polling system on JVM

armanbilge:
Tries out:
- https://github.com/typelevel/cats-effect/pull/3332

So far I've done TCP sockets, still need to do unix and datagram sockets.

The `SocketSuite` passes for me locally, but not the `TLSSocketSuite` which revealed a bug using the polling system from blocking threads. After that's fixed I'd like to publish a snapshot for downstream test in Ember etc.

As with https://github.com/typelevel/fs2/pull/3087, this PR is tangled up with https://github.com/typelevel/fs2/pull/3063, so that will have to merge first.


armanbilge:
Published `3.5-ac5c657`.

armanbilge:
So after the update, TCP suite is passing, but TLS TCP suite is hanging.

A notable difference is that the TLS implementation uses `blocking(...)` and indeed the thread dump includes:

```
"io-compute-blocker-9" #32 daemon prio=5 os_prio=0 cpu=115.87ms elapsed=10.09s tid=0x00007f5f604e1680 nid=0x1
4d7 waiting on condition  [0x00007f5f3bffe000]
   java.lang.Thread.State: TIMED_WAITING (parking)
        at jdk.internal.misc.Unsafe.park(java.base@17.0.6/Native Method)
        - parking to wait for  <0x0000000725e008d0> (a java.util.concurrent.LinkedTransferQueue)
        at java.util.concurrent.locks.LockSupport.parkNanos(java.base@17.0.6/LockSupport.java:252)
        at java.util.concurrent.LinkedTransferQueue.awaitMatch(java.base@17.0.6/LinkedTransferQueue.java:676)
        at java.util.concurrent.LinkedTransferQueue.xfer(java.base@17.0.6/LinkedTransferQueue.java:616)
        at java.util.concurrent.LinkedTransferQueue.poll(java.base@17.0.6/LinkedTransferQueue.java:1294)
        at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:449)
```

armanbilge:
Ok, so actually I can't reproduce the hanging on my macOS machine. Which makes me think it's something to do with epoll, and not kqueue? 🤔

armanbilge:
@durban your CE fix definitely helped things a lot, thanks. Unfortunately somehow one of the jobs still completely flaked on all the polling system stuff and then ended up hanging 🤔 at least this time it's not specific to TLS / blocking ...

https://github.com/typelevel/fs2/actions/runs/4816825621/jobs/8576820432?pr=3091#step:12:2466

durban:
I think I can sometimes (rarely) reproduce it locally. The WSTP is in an invalid state, although in a different way than before. I'll try to get details later.

armanbilge:

PR #3092, opened 2022-12-27
Update doc for Stream exists method

### LINKED ISSUE #3034: Fix ScalaDoc of Stream.exists
```scala

  /** Emits `true` as soon as a matching element is received, else `false` if no input matches.
    * '''Pure''': this operation maps to `List.exists`
    *
    * @example {{{
    * scala> Stream.range(0,10).exists(_ == 4).toList
    * res0: List[Boolean] = List(true)
    * scala> Stream.range(0,10).exists(_ == 10).toList
    * res1: List[Boolean] = List(false)
    * }}}
    * @return Either a singleton stream, or a `never` stream.
    *  - If `this` is a finite stream, the result is a singleton stream, with after yielding one single value.
    *    If `this` is empty, that value is the `mempty` of the instance of `Monoid`.
    *  - If `this` is a non-terminating stream, and no matter if it yields any value, then the result is
    *    equivalent to the `Stream.never`: it never terminates nor yields any value.
    */
  def exists(p: O => Boolean): Stream[F, Boolean] =
    this.pull.forall(!p(_)).flatMap(r => Pull.output1(!r)).stream
```

In particular, the `@returns` doc is incorrect.

b-vennes:
Updates the docs for the Stream `exists` method to correctly describe the current implementation.

fixes #3034

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1059,10 +1059,12 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * res1: List[Boolean] = List(false)
     * }}}
     * @return Either a singleton stream, or a `never` stream.
-    *  - If `this` is a finite stream, the result is a singleton stream, with after yielding one single value.
-    *    If `this` is empty, that value is the `mempty` of the instance of `Monoid`.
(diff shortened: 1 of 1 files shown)


diesalbla (approved):

diesalbla:
@b-vennes Thanks.

PR #3093, opened 2022-12-27
Run `fs2.io.stdin` on dedicated single-threaded pool

armanbilge:
See updated description in https://github.com/typelevel/fs2/pull/3093#issuecomment-1366256942.

---

This "fixes" the FS2 manifestation of https://github.com/typelevel/cats-effect/issues/3077 where canceling the `stdin` stream hangs, because you cannot interrupt a blocking read from stdin.

"Fixes" because this is a cantfix: we just leak the blocked thread. The docs are updated to recommend you only compile the `stdin` stream at most once in a program, to mitigate this.

Couldn't think how to test this without setting up lots of infrastructure like we do in Cats Effect to test `IOApp`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import com.typesafe.tools.mima.core._
 
 Global / onChangedBuildSource := ReloadOnSourceChanges
 
-ThisBuild / tlBaseVersion := "3.4"
+ThisBuild / tlBaseVersion := "3.5"
 
 ThisBuild / organization := "co.fs2"
 ThisBuild / organizationName := "Functional Streams for Scala"
--- a/build.sbt
+++ b/build.sbt
@@ -174,7 +174,10 @@ ThisBuild / mimaBinaryIssueFilters ++= Seq(
   ),
   ProblemFilters.exclude[DirectMissingMethodProblem](
     "fs2.io.net.tls.S2nConnection#RecvCallbackContext.readBuffer"
-  )
+  ),
+  ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.io.package.readBytesFromInputStream"),
+  ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.io.package.readInputStreamGeneric"),
+  ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.io.package.<clinit>")
 )
 
 lazy val root = tlCrossRootProject
--- a/io/jvm-native/src/main/scala/fs2/io/iojvmnative.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/iojvmnative.scala
@@ -23,9 +23,11 @@ package fs2
 package io
 
 import cats._
+import cats.effect.kernel.Async
 import cats.effect.kernel.Sync
 import cats.effect.kernel.implicits._
 import cats.syntax.all._
+import fs2.concurrent.Channel
 
 import java.nio.charset.Charset
 import java.nio.charset.StandardCharsets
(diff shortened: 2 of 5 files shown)


build.sbt
```diff
@@ -2,7 +2,7 @@ import com.typesafe.tools.mima.core._
 
 Global / onChangedBuildSource := ReloadOnSourceChanges
 
-ThisBuild / tlBaseVersion := "3.4"
+ThisBuild / tlBaseVersion := "3.5"
```
armanbilge, line 5: This change is source-breaking: `stdin` now requires an `Async` constraint. So we may want to sit on this PR until we are ready for the minor bump.

io/jvm-native/src/main/scala/fs2/io/iojvmnative.scala
```diff
@@ -38,9 +40,31 @@ private[fs2] trait iojvmnative {
...
+        readInputStream(F.blocking(System.in), bufSize, false).chunks
+          .through(ch.sendAll)
+          .compile
+          .drain
+          .start
+      } { fiber =>
+        // cancelation may hang so we leak :(
+        fiber.cancel.start.void
+      } >> ch.stream.unchunks
+    }
```
diesalbla, line 63:
```suggestion
  def stdin[F[_]](bufSize: Int)(implicit F: Async[F]): Stream[F, Byte] = {
    val raw = readInputStream(F.blocking(System.in), bufSize, false)  
    Stream.eval(Channel.synchronous[F, Chunk[Byte]]).flatMap { ch =>
      Stream.bracket {
        ch.sendAll(raw.chunks).compile.drain.start
      } { fiber =>
        // cancelation may hang so we leak :(
        fiber.cancel.start.void
      } >> ch.stream.unchunks
    }
  }
```

diesalbla (approved):

diesalbla:
> This "fixes" the FS2 manifestation of https://github.com/typelevel/cats-effect/issues/3077 where canceling the stdin stream hangs, because you cannot interrupt a blocking read from stdin. "Fixes" because this is a cantfix: we just leak the blocked thread. The docs are updated to recommend you only compile the stdin stream at most once in a program, to mitigate this.

@armanbilge Could you confirm if the following is true? "The root cause of the problem is outside FS2 scope. This PR mitigates the problem and is an improvement over current situation."

armanbilge:
> The root cause of the problem is outside FS2 scope.

Yes, it's a problem in the JVM, which also happens to be affect Cats Effect `Console#readLine`.

> This PR mitigates the problem and is an improvement over current situation.

Well, hard to say. @keynmol makes a good point that this PR exposes a bit of a footgun, since misuse can unboundedly leak threads. That same sort of misuse today would cause your program to hang. In fact even correct use can cause your program to hang, which this PR fixes.

I don't think we should move forward with what I currently have here. Here's an alternative:

- We create a dedicated, single-thread-executor for stdin. So instead of blocking using the usual blocking pool (and spawning new threads), we always block on that same thread. Abuse could still leak to a fiber leak, but maybe not actually 🤔 

  Since there is only one stdin it's fine to have only one thread. It's a bit naughty with global threadpools, but `Network` is precedent for this and stdin is a naughty global anyway.

io/jvm/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -26,12 +26,49 @@ import cats.effect.kernel.{Async, Outcome, Resource, Sync}
...
+          val task: Runnable = () => cb(Right(is.read(buf)))
+          stdinExecutor.submit(task)
+        }.map { fut =>
+          Some(F.delay {
+            fut.cancel(false)
+            ()
+          })
+        }
+      }
+    }
```
armanbilge, line 58: Ok, I think this works :) we run all reads from stdin on a custom, single-threaded executor. That way at most one runs at a time, so we never leak threads. We also can't leak fibers, since all tasks (except the one that's running) can always be canceled and removed.

armanbilge:
@keynmol I published `3.5-2d970e8-SNAPSHOT` if you don't mind giving it a try 🙏

mpilquist (approved):

PR #3098, opened 2023-01-05
Replace "ToPull" operations with "Pull.uncons"

diesalbla:
- Pull - Add an uncons extension method for closed Pulls (unit)
- Stream - Avoid "newStream" use streamNoScope
- Avoid the ".pull" method, use the new ".uncons" method on Pull-Unit

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -70,6 +70,7 @@ ThisBuild / mimaBinaryIssueFilters ++= Seq(
   ProblemFilters.exclude[DirectAbstractMethodProblem]("fs2.Pull#CloseScope.*"),
   ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Pull#BindBind.*"),
   ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Pull#CloseScope.*"),
+  ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.Pull.uncons"),
   ProblemFilters.exclude[MissingClassProblem]("fs2.Pull$CloseScope$"),
   ProblemFilters.exclude[MissingClassProblem]("fs2.Pull$EvalView"),
   ProblemFilters.exclude[MissingClassProblem]("fs2.Pull$View"),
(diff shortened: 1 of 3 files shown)


mpilquist:
Thoughts:
- Moving `Pull.uncons` to `Pull#uncons` looks good, as does the optimization of the implementation.
- I'm neutral on changing `new Stream(p)` to `p.streamNoScope`.
- I don't like rewriting the internal recursive combinators to use `Pull` over `Stream`. The implementations use private APIs that users don't have access to, making it harder to learn from examples.

mpilquist:
I don't know if that's a good idea. Seems like a lot of churn API-wise (all the recursive pulls that would need to change to use  a pull instead of stream argument). What's the benefit we get for that churn?

diesalbla:
> I don't know if that's a good idea. Seems like a lot of churn API-wise (all the recursive pulls that would need to change to use a pull instead of stream argument). What's the benefit we get for that churn?

Perhaps reducing the need for a second type wrapper, `ToPull`, in addition to the default `Pull`, but that may not be worth too many changes. Nevertheless, I will split this PR in twine later today.

diesalbla:
> I'm a slight -1 on using streamNoScope over new Stream: it forces all readers to having to think about what a scope is, which arguably isn't necessary until much later on in their journey

Perhaps we should rename them as `stream` and `scopedStream`, giving the more complex notion the longer name.

mpilquist:
We need the default behavior of a pull-to-stream conversion to introduce a scope, or otherwise risk resource leaks.On Jan 6, 2023, at 5:42 PM, Diego E. Alonso Blas ***@***.***> wrote:﻿

I'm a slight -1 on using streamNoScope over new Stream: it forces all readers to having to think about what a scope is, which arguably isn't necessary until much later on in their journey

Perhaps we should rename them as stream and scopedStream, giving the more complex notion the longer name.

armanbilge:
I thought that introducing a scope is safer, and should be the default? At least that's the feeling I get from the scaladocs. In fact I feel like `streamNoScope` is the more complex notion, since I have to know when it's safe to not introduce a scope.

https://github.com/typelevel/fs2/blob/53465e531253dd14e1a0ab00b859b89ddc02f2fa/core/shared/src/main/scala/fs2/Pull.scala#L315-L318

PR #3102, opened 2023-01-07
Add interop with Java Flow types

BalmungSan:
Adds support to converting `Streams` back and forth with **Java 9+** `Flow` types.

Roadmap

- [x] Copy and paste the `reactive-streams` implementation inside `core` using `Flow` types instead.
- [x] Clean up the new versions.
- [x] Define an appropriate package for the new classes.
- [x] Add tests for the new classes.
- [x] Deprecate the `reactive-streams` module


core/jvm/src/main/scala/fs2/rx/package.scala
```diff
@@ -0,0 +1,74 @@
...
+  * scala> val publisher: Resource[IO, Publisher[Int]] = toPublisher(upstream)
+  * scala> val downstream: Stream[IO, Int] = Stream.resource(publisher).flatMap(p => fromPublisher(p, bufferSize = 16))
+  * scala>
+  * scala> downstream.compile.toVector.unsafeRunSync()
+  * res0: Vector[Int] = Vector(1, 2, 3)
+  * }}}
+  *
+  * @see [[http://www.reactive-streams.org/]]
+  */
+package object rx {
```
armanbilge, line 44: Following the rest of FS2, the package should be `interop.flow` maybe?

https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html

armanbilge:
Looks like reactive streams has also published a testkit for the Java Flow counterparts.
https://www.reactive-streams.org/reactive-streams-tck-flow-1.0.2-javadoc/org/reactivestreams/tck/flow/package-summary.html

core/jvm/src/main/scala/fs2/rx/package.scala
BalmungSan: I was about to ask what should be the right package name.
`interop.flow` sounds good.

build.sbt
```diff
@@ -217,7 +217,7 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
       "org.typelevel" %%% "munit-cats-effect" % "2.0.0-M3" % Test,
       "org.typelevel" %%% "discipline-munit" % "2.0.0-M3" % Test
     ),
-    tlJdkRelease := Some(8),
+    tlJdkRelease := Some(9),
```
BalmungSan, line 220: Is this the right way to ensure the class compiles?

armanbilge: Actually this will generate bytecode that is not compatible with Java 8 ... actually we should set it to `None` 🙃

BalmungSan:
> Deprecate the reactive-streams module?

So, do we want to deprecate those or not?
And if so, what should be the message?

armanbilge:
> So, do we want to deprecate those or not?

Unfortunately I don't think we should "hard" deprecate it, because it leaves Java 8 users out in the cold.

Instead I propose a "soft" deprecation (like we did for Cats `Future` instances) that only appears in Scaladocs, not to the compiler.

The message can say that they will be removed in the next major version of FS2, and that the converters in this new package are the replacement.

armanbilge:
Published as `3.4.0-59-7543784-SNAPSHOT` for downstream testing.

Jasper-M:
> The message can say that they will be removed in the next major version of FS2, and that the converters in this new package are the replacement.

How would one interop with an API that is not updated to Java Flow in that scenario?

BalmungSan:
@Jasper-M the same one interop with a `Flow` API in this moment, using the `FlowAdapters` provided by `reactive-streams`: https://www.reactive-streams.org/reactive-streams-flow-adapters-1.0.2-javadoc/org/reactivestreams/FlowAdapters.html
_(just that it would be backwards in this case)_.

The rationale to prefer this approach is:
1. In theory, all APIs that expose `reactive-streams` should move to `Flow` types once they support **Java 9**.

> This means that there will be a migratory period, while libraries move to adopt the new types in the JDK, however this period is expected to be short - due to the full semantic equivalence of the libraries, as well as the Reactive Streams <-> Flow adapter library as well as a TCK compatible directly with the JDK Flow types.
> -- https://www.reactive-streams.org/

2. Right now, if you need to interop only with `Flow` types, you still need to include the `reactive-streams` library in your classpath. Whereas, with this new approach you won't be adding a new dependency.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,118 @@
...
+    * @param bufferSize setup the number of elements asked each time from the [[Publisher]].
+    *                   A high number can be useful if the publisher is triggering from IO,
+    *                   like requesting elements from a database.
+    *                   The publisher can use this `bufferSize` to query elements in batch.
+    *                   A high number will also lead to more elements in memory.
+    */
+  def fromPublisher[F[_], A](
+      publisher: Publisher[A],
+      bufferSize: Int
+  )(implicit
```
armanbilge, line 63: Luis and I discussed this extensively offline. Unfortunately this current API is not usable as-is in the http4s-jdk-http-client, because it needs to run an effect in its `Publisher`. The workaround would be to `unsafeRunSync()` said effect.

However, noting that a `Publisher[A]` is just a `@FunctionalInterface` for `Subscription[A] => Unit` which is a less ergonomic version of `Subscription[A] => F[Unit]`, I think we should change the signature to something like this:

```scala
def fromPublisher[F[_], A](bufferSize: Int)(
  publisher: Subscription[A] => F[Unit]
)(implicit F: Async[F]): Stream[F, A]
```

This also has a very nice parallel to existing APIs in FS2.
https://github.com/typelevel/fs2/blob/19e691f567470389c29e56c7f1ebcb4964f73d16/io/jvm/src/main/scala/fs2/io/ioplatform.scala#L67-L71

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,118 @@
...
+
+    /** Creates a [[Stream]] from an [[Publisher]].
+      *
+      * @param bufferSize setup the number of elements asked each time from the [[Publisher]].
+      *                   A high number can be useful is the publisher is triggering from IO,
+      *                   like requesting elements from a database.
+      *                   The publisher can use this `bufferSize` to query elements in batch.
+      *                   A high number will also lead to more elements in memory.
+      */
+    def toStream[F[_]](bufferSize: Int)(implicit F: Async[F]): Stream[F, A] =
```
armanbilge, line 84: I am a bit skeptical if we should expose this, for the same reasons we don't expose a `toIO` syntax for `Future`. Specifically, the signature does not ask for a raw `Future`, but a suspended one:
```scala
object IO {
  def fromFuture[A](fut: IO[Future[A]]): IO[A]
}
```

BalmungSan: Well, AFAIK a `Publisher` is just a description, you can create one without needing to run any side effects.

core/jvm/src/main/scala/fs2/interop/flow/StreamUnicastPublisher.scala
```diff
@@ -0,0 +1,103 @@
...
+      new StreamUnicastPublisher(
+        stream,
+        startDispatcher,
+        requestDispatcher,
+        completedSignal,
+        usedCheck
+      ) -> completedSignal
+    }
+
+  private object StreamUnicastPublisherAlreadyUsedException
```
BalmungSan, line 98: Maybe this exception should be moved to the `package` and made public?

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,157 @@
...
+    */
+  def subscribeStream[F[_], A](
+      stream: Stream[F, A],
+      subscriber: Subscriber[A]
+  )(implicit
+      F: Async[F]
+  ): F[Unit] =
+    flow.toPublisher(stream).use { case (publisher, completedSignal) =>
+      F.delay(publisher.subscribe(subscriber)) >>
+        F.race(completedSignal.get, F.never).void
```
armanbilge, line 128: Why `race` with `never` ?

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,253 @@
...
+  )(implicit
+      F: Async[F]
+  ): Stream[F, A] =
+    Stream
+      .eval(StreamSubscriber[F, A](bufferSize))
+      .flatMap { subscriber =>
+        subscriber.stream(
+          F.uncancelable(_ => subscribe(subscriber))
+        )
+      }
```
BalmungSan, line 102: Should we expose another version that gives access to `Poll` to users?

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,253 @@
...
+      onSubscribe: (Publisher[A] => F[Unit]) => F[Unit]
+  )(implicit
+      F: Async[F]
+  ): Stream[F, A] =
+    fromSubscriber[F, A](bufferSize) { subscriber =>
+      val subscribe: Publisher[A] => F[Unit] =
+        publisher => F.delay(publisher.subscribe(subscriber))
+
+      onSubscribe(subscribe)
+    }
```
BalmungSan, line 149: Should we expose another version that gives access to `Poll` to users?

armanbilge: Tbh I'm confused about this API.
- `(Publisher[A] => F[Unit]) => F[Unit]` is a complex signature
- it doesn't represent any existing concept from the reactive streams spec AFAICT

core/jvm/src/main/scala/fs2/interop/flow/package.scala
armanbilge: Also, I think the name should be `fromSubscribe` as in "from a `subscribe` function". Arguably the name could also be `fromPublisher`, since `Subcriber[A] => Unit` is isomorphic to publisher.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,253 @@
...
+    * @param stream The [[Stream]] to transform.
+    * @return A tuple of the [[Publisher]]
+    *         and a [[DeferredSource]] that will be completed when all the stream elements are published.
+    */
+  def toPublisher[F[_], A](
+      stream: Stream[F, A]
+  )(implicit
+      F: Async[F]
+  ): Resource[F, (Publisher[A], DeferredSource[F, Unit])] =
+    StreamUnicastPublisher(stream)
```
armanbilge, line 204: I actually believe it's unnecessary to expose the `DeferredSource` here. To get access to this information a user can always wrap their `Subscriber` in a `Subscriber` that does something extra for `onComplete` / `onExtra`.

BalmungSan: While that is technically true, IMHO that would force extra burden on our users. Especially since those functions are not effectual.

Additionally, that approach would make it harder to wait for termination.

However, maybe we should use `Either[Throwable, Unit]` for the type of the `Deferred`?

armanbilge: > While that is technically true, IMHO that would force extra burden on our users.

How often will users really need this API? It's an advanced API IMO with rare use-cases. I see how it's useful for the `subscribeStream(...)`, in which case we can use the technique I described in my previous comment.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
BalmungSan: I don't remember why I was sure the `subscribe` bit should not be cancelable.

So, we can guarantee that on the other methods. Thus, we may also guarantee this on this one, or leave that to the discretion of users since this API is more low-level / error prone.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
BalmungSan: > (Publisher[A] => F[Unit]) => F[Unit] is a complex signature

IMHO, is very similar to `async` you pass a function that receives as an argument some kind of callback; thus another function.

> it doesn't represent any existing concept from the reactive streams spec AFAICT

Not directly, yes.
It is an effectual inversion of control of `Publisher.subscribe`.

The idea of this API is to be mostly as useful as `fromSubscriber` but with less surface for error, since the actual `Subscriber` is not exposed, only its capability to `subscribe` to a `Publisher`.
Having said that, one can indeed argue this offer little value when we already have the safe version and the fully flexible version.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
BalmungSan: > It's an advanced API IMO

Right, just like the alternative version of `fromPublisher`, where an user could just create its own `Publisher` and manually call `unsafeRunSync()`.
Yet, you were not really happy about that yesterday :)

I agree this is an advanced API and most users should use `subscribe` instead.
But, I would rather try to make this as ergonomic and safe as possible; e.g. actually canceling subsequent `subscribe` calls rather than leaving them in an undefined state as the original implementation did.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
BalmungSan: > Also, I think the name should be fromSubscribe as in "from a subscribe function"

Well, TBH, that would be `fromOnSubscribe`, not this one. Because this function allows doing way more than just `subscribe`... even if anything else than just subscribe is unsafe; which is why I wasn't too keen to add this.

> Arguably the name could also be fromPublisher, since Subcriber[A] => Unit is isomorphic to publisher

Right, if we remove the other one I think is better to overload; otherwise I personally prefer the three different names.
Although, I wonder if it could lead to type inference issues.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
armanbilge: To clarify: I think `Resource[F, (Publisher[A], DeferredSource[F, Unit])]` where you actually need to use that `Deferred`) is an advanced API. And the API already exposes that information without providing a `DeferredSource`.

Furthermore, if you don't need to use it (i.e. most/all usescases) then it's just plain unergonomic.  `Resource[F, Publisher[A]]` is simple enough.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
armanbilge: > Because this function allows doing way more than just `subscribe`

Not really :) the signature of this function is isomorphic to `Publisher`. Whatever you can do with one, you can do with the other.

> Although, I wonder if it could lead to type inference issues.

In this case it will be fine, because both the arity and types of the function are changing.

BalmungSan: > Not really :) the signature of this function is isomorphic to Publisher

Sure, which is why I do like `fromPublisher`, rather than `fromSubscribe`.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
BalmungSan: > And the API already exposes that information without providing a DeferredSource.

Sure, just like `fromPublisher(publisher: Publisher[A]): Stream[A]` exposes all that is needed.
So, I don't understand how this is different than adding the `fromPublisher(publisher: Subscriber[A] => F[Unit]): Stream[A]`?

> Furthermore, if you don't need to use it (i.e. most/all usescases)

Then you can just use `streamSubscribe` and simpler than a `Resource` that if you leave without ensuring all the elements were consumed will cancel the downstream consumer.

armanbilge: > So, I don't understand how this is different

1. Because there is a performance cost (i.e., it requires `unsafeRunSync()`). In general, it's easier to recover an async effect from an unsafe API than to run an async effect from an unsafe synchronous API.
2. Because there's an existing external usecase in the http4s-jdk-http-client.
3. Because it did not require changing internal implementations, as adding this did.

> Then you can just use `streamSubscribe` and simpler than a `Resource` that if you leave without ensuring all the elements were consumed will cancel the downstream consumer.

What if you actually need a `Publisher`?

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -99,20 +202,69 @@ package object flow {
...
+    * @param subscriber the [[Subscriber]] that will receive the elements of the stream.
+    */
+  def subscribeStream[F[_], A](
+      stream: Stream[F, A],
+      subscriber: Subscriber[A]
+  )(implicit
+      F: Async[F]
+  ): F[Unit] =
+    flow.toPublisher(stream).use { publisher =>
+      F.async_[Unit] { cb =>
```
armanbilge, line 221: This `async_` is uncancelable (or will be, after the bug fix in CE 3.5.0 is release). We should use the `async` variant and return `Some(IO.unit)` so it is cancelable.

core/jvm/src/main/scala/fs2/interop/flow/StreamUnicastPublisher.scala
armanbilge: According to the JavaDoc you should be throwing an `IllegalStateException`.
https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.Publisher.html#subscribe-java.util.concurrent.Flow.Subscriber-

BalmungSan: Not throwing but sending :)

But yes, I didn't realize it should be an `IllegalStateException`, will fix that.
Also, that means there is no need to make this one public.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
BalmungSan: IIRC the subscription process shouldn't be cancelable, but once it has been subscribed then we can cancel that `subscription`; which is done when the `Resource` is closed.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -99,20 +202,69 @@ package object flow {
...
   }
 
   /** Ensures a value is not null. */
   private[flow] def nonNull[A](a: A): Unit =
     if (a == null) {
       throw new NullPointerException()
     }
+
+  private final val RUnit: Either[Nothing, Unit] =
+    Right(())
```
armanbilge, line 269: This is just `Either.unit`. Cats already caches it for us :)

armanbilge:
> the older InputStream/OutputStream stuff

Note that this is a different set of APIs to reactive streams / Flow. We also offer interop with those in fs2-io which we can cannot remove/deprecate, since we rely on them internally.

---

> This feels very optimistic.

I wouldn't say _very_ optimistic, but certainly optimistic. The reactive streams and Java Flow APIs are basically identical except for package names. So it's not a hard migration.

Also as pointed out in https://github.com/typelevel/fs2/pull/3102#issuecomment-1375557179, we are not really dropping support for reactive streams, just requiring the use of a thin adapter. And this wouldn't be until FS2 drops JDK 8 support anyway.

Daenyth:
Gotcha, I misunderstood

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -0,0 +1,256 @@
...
+  /** Creates a [[Stream]] from this [[Subscriber]]. */
+  def stream(subscribe: F[Unit]): Stream[F, A] =
+    subscriber.stream(subscribe)
+}
+
+private[flow] object StreamSubscriber {
+
+  /** Instantiates a new [[StreamSubscriber]] for the given buffer size. */
+  def apply[F[_], A](bufferSize: Int)(implicit F: Async[F]): F[StreamSubscriber[F, A]] =
+    fsm[F, A](bufferSize).map(fsm => new StreamSubscriber(subscriber = fsm))
```
BalmungSan, line 76: Should we throw an exception if `bufferSize` is not positive?

reactive-streams/src/main/scala/fs2/interop/reactivestreams/package.scala
```diff
@@ -48,7 +48,7 @@ import org.reactivestreams._
   *   which provides support for the `java.util.concurrent.Flow` types;
   *   that superseded the `reactive-streams` library.
   *   In case you need to interop with a library that only provides `reactive-streams` types,
-  *   you may use the [[org.reactivestreams.FlowAdapters]].
+  *   you may use `org.reactivestreams.FlowAdapters`: [[https://www.reactive-streams.org/reactive-streams-flow-adapters-1.0.2-javadoc/org/reactivestreams/FlowAdapters.html]]
```
BalmungSan, line 51: I couldn't use a proper link to the class, since it doesn't exist on **Java 8** artifacts.

armanbilge:
Published `3.5.0-7-87bd1e4-SNAPSHOT`.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,257 @@
...
+    * @note The publisher only receives a subscriber when the stream is run.
+    *
+    * @see [[fromPublisher]] for a simpler version that only requires a [[Publisher]].
+    *
+    * @param bufferSize setup the number of elements asked each time from the [[Publisher]].
+    *                   A high number can be useful if the publisher is triggering from IO,
+    *                   like requesting elements from a database.
+    *                   The publisher can use this `bufferSize` to query elements in batch.
+    *                   A high number will also lead to more elements in memory.
+    * @param onSubscribe The effectual function to run during the subscription process,
```
armanbilge, line 84:
```suggestion
    * @param subscribe The effectual function to run during the subscription process,
```

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,257 @@
...
+    *                    receives a [[Subscriber]] that should be used to subscribe to a [[Publisher]].
+    *                    The `subscribe` operation must be called exactly once.
+    */
+  def fromSubscriber[F[_], A](
+      bufferSize: Int
+  )(
+      subscribe: Subscriber[A] => F[Unit]
+  )(implicit
+      F: Async[F]
+  ): Stream[F, A] =
```
armanbilge, line 94: I'm still not sure about this name. We are not creating a `Stream` from a `Subscriber` (i.e. the user is not giving us a `Subscriber`), we are building a `Stream` from a `subscribe` function aka a `Publisher` (i.e. this is what the user is giving us).

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,257 @@
...
+    Stream
+      .eval(StreamSubscriber[F, A](bufferSize))
+      .flatMap { subscriber =>
+        subscriber.stream(
+          F.uncancelable(_ => subscribe(subscriber))
```
armanbilge, line 99: It should be up to the user to make this uncancelable. Since we are not chaining effects in the implementation here, this can be done completely (and more flexibly) in userland.
```suggestion
          subscribe(subscriber)
```

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,257 @@
...
+    *                    the actual `subscribe` operation; as another function receiving the [[Publisher]].
+    *                    The `subscribe` operation must be called exactly once.
+    */
+  def fromOnSubscribe[F[_], A](
+      bufferSize: Int
+  )(
+      onSubscribe: (Publisher[A] => F[Unit]) => F[Unit]
+  )(implicit
+      F: Async[F]
+  ): Stream[F, A] =
```
armanbilge, line 144: Sorry, still 👎 on this method. As I said, the signature is very confusing (as callbacks are) and it's not isomorphic to any existing interface in the Reactive Streams API.

armanbilge (commented):
I was able to migrate the http4s-jdk-http-client 🎉 
- https://github.com/http4s/http4s-jdk-http-client/pull/812

core/jvm/src/main/scala/fs2/interop/flow/package.scala
BalmungSan: I ended up removing it because it was redundant and still unsafe.

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,181 @@
...
+    */
+  def fromPublisher[F[_], A](
+      publisher: Publisher[A],
+      bufferSize: Int
+  )(implicit
+      F: Async[F]
+  ): Stream[F, A] =
+    fromPublisher[F, A](bufferSize) { subscriber =>
+      F.delay(publisher.subscribe(subscriber))
+    }
```
BalmungSan, line 137: Should we use the _"partially applied type trick"_ here?

```scala
// So users can do:
fromPublisher[IO](somePublisher, bufferSize = 16)

// Instead of:
fromPublisher[IO, Int](somePublisher, bufferSize = 16)
```

Since that type is redundant there.

core/jvm/src/main/scala/fs2/interop/flow/syntax.scala
```diff
@@ -0,0 +1,66 @@
+package fs2
+package interop
+package flow
+
+import cats.effect.kernel.{Async, Resource}
+
+import java.util.concurrent.Flow.{Publisher, Subscriber}
+
+object syntax {
```
BalmungSan, line 9: I moved the implicit classes to here from the `package` because:

1. To fix some annoying error with the doctest on `2.12`
2. IMHO, is better not to import both the explicit methods and the syntax when doing `import fs2.inteorp.flow._`
3. To follow the standard of other libraries like **cats** and **cats-effect**

core/jvm/src/main/scala/fs2/interop/flow/package.scala
armanbilge: Sure, we can do it.

core/jvm/src/main/scala/fs2/interop/flow/syntax.scala
```diff
@@ -84,4 +84,17 @@ object syntax {
     def subscribe(subscriber: Subscriber[A])(implicit F: Async[F]): F[Unit] =
       flow.subscribeStream(stream, subscriber)
   }
+
+  private[flow] final class FromPublisherPartiallyApplied[F[_]](private val dummy: Boolean)
```
armanbilge, line 88: Unfortunately this has to be public, otherwise MiMa will not support us here. I've been bitten by this more than once.
```suggestion
  final class FromPublisherPartiallyApplied[F[_]](private val dummy: Boolean)
```

Jasper-M: Also annoying if you can't see the scaladoc for this class while it's actually public API.

reactive-streams/src/main/scala/fs2/interop/reactivestreams/package.scala
```diff
@@ -41,6 +41,14 @@ import org.reactivestreams._
...
   *
   * @see [[http://www.reactive-streams.org/]]
+  *
+  * @deprecated
+  *   The next major version of fs2 will drop these converters.
+  *   Rather, users will be encouraged to use the new [[fs2.interop.flow]] package,
+  *   which provides support for the `java.util.concurrent.Flow` types;
+  *   that superseded the `reactive-streams` library.
+  *   In case you need to interop with a library that only provides `reactive-streams` types,
+  *   you may use `org.reactivestreams.FlowAdapters`: [[https://www.reactive-streams.org/reactive-streams-flow-adapters-1.0.2-javadoc/org/reactivestreams/FlowAdapters.html]]
```
armanbilge, line 51:
```suggestion
  *   you may use [[https://www.reactive-streams.org/reactive-streams-flow-adapters-1.0.2-javadoc/org/reactivestreams/FlowAdapters.html `org.reactivestreams.FlowAdapters`]]
```

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -0,0 +1,174 @@
...
+  )(implicit
+      F: Async[F]
+  ): F[Unit] =
+    StreamSubscription.subscribe(stream, subscriber)
+
+  /** Ensures a value is not null. */
+  private[flow] def nonNull[A](a: A): Unit =
+    if (a == null) {
+      throw new NullPointerException()
+    }
```
armanbilge, line 173: This helper is equivalent to `java.util.Objects.requireNonNull(...)`

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
armanbilge: Sure, I think Cats Effect does this in `std` with `require`. So this value can't be zero?

armanbilge: Luis explained offline that the `bufferSize` is used for `request(...)`, which is required positive by the spec.

armanbilge (approved):
Thanks for all your work, this is such a nice iteration on the reactive-streams module.

I'm going to publish one more snapshot.

armanbilge:
https://github.com/typelevel/fs2/commit/31cf9277a06d899d4d4408b3898b227d0a5a1848 will be a follow-up PR :)

armanbilge:
Looks like there is a [`defaultBufferSize()`](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html#defaultBufferSize--), should we make use of this anywhere?

BalmungSan:
> Looks like there is a [defaultBufferSize()](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html#defaultBufferSize--), should we make use of this anywhere?

Oh, I didn't see that one.
I actually thought of proposing a default for `bufferSize` but I wasn't sure which may be a good value.

IMHO, we should use this one on all public APIs

core/jvm/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -82,12 +82,13 @@ package object flow {
...
     * @param subscribe The effectual function that will be used to initiate the consumption process,
     *                  it receives a [[Subscriber]] that should be used to subscribe to a [[Publisher]].
     *                  The `subscribe` operation must be called exactly once.
     */
   def fromPublisher[F[_], A](
-      bufferSize: Int
+      bufferSize: Int = defaultBufferSize
   )(
       subscribe: Subscriber[A] => F[Unit]
   )(implicit
```
armanbilge, line 94: Even with the default this will require an empty set of parens e.g.
```scala
fromPublisher()(subscriber => ...)
```
not the end of the world though.

BalmungSan: IMHO, that is a good idea, one should not rely on the default, and making it look weird is a good way to motivate folks to use an appropriate value.

armanbilge: If the goal is to discourage use of the default, why have it all then 😅 but fair enough, I don't feel strongly any way about this.

We could also expose `defaultBufferSize` as a method on `flow`, just like in the JDK. So users can use that one if they wish.

BalmungSan: Sorry, I didn't express myself clearly.
My point is mostly that users should be conscious of that `bufferSize` parameter.

armanbilge (approved):

Issue #1678, opened 2019-11-04
Deprecate StepLeg?


mpilquist:
I was looking in to #1624 and was considering improving the error message thrown here: https://github.com/functional-streams-for-scala/fs2/blob/829279df39c6cb779e512061ed6b4cdc8f622bb1/core/shared/src/main/scala/fs2/internal/Algebra.scala#L233

In doing so, I tried to reproduce the original error that led to the creation of `StepLeg` but I couldn't get things to fail. I wrote this test class:

```scala
package fs2

class Proof extends Fs2Spec {
  "proof" - {
    "zip proof" in {
      def brokenZip[F[_], A, B](s1: Stream[F, A], s2: Stream[F, B]): Stream[F, (A, B)] = {
        def go(s1: Stream[F, A], s2: Stream[F, B]): Pull[F, (A, B), Unit] =
          s1.pull.uncons1.flatMap {
            case Some((hd1, tl1)) =>
              s2.pull.uncons1.flatMap {
                case Some((hd2, tl2)) =>
                  Pull.output1((hd1, hd2)) >> go(tl1, tl2)
                case None => Pull.done
              }
            case None => Pull.done
          }
        go(s1, s2).stream
      }
      val s = Stream(0).scope
      brokenZip(s ++ s, s).toList shouldBe List((0, 0))
    }
  }
}
```

The `brokenZip` method implements `zip` by calling `uncons1` on each source stream. The assertion comes from our test suite, where we'd run in to the scope id issue due to interleaving of streams.

This test passes on the current head and fails on v0.10.0. I ran a git bisect and determined that this test has been passing since a9f8a1d437603b6a08f29b019e9891521e886727.

On HEAD, I reimplemented `zipWith_` using `uncons` instead of `stepLeg` and all tests pass.

So, should we deprecate `StepLeg`? Or is there a better example program that demonstrates the need for it?

/cc @AdamChlupacek @pchlupacek

mpilquist:
I haven't succeeded in reproducing an exception with that original test case, but I have succeeded in showing why we need `StepLeg`. Here's two definitions of deterministically merging sorted streams:

```
def sortedMerge[F[_], O: Order](s1: Stream[F, O], s2: Stream[F, O]): Stream[F, O] = {
      def go(h1h: O, s1: Stream[F, O], h2h: O, s2: Stream[F, O]): Pull[F, O, Unit] = {
        if (Order[O].compare(h1h, h2h) <= 0) {
[...]

PR #3110, opened 2023-01-17
Add .interleaveOrdered to combine sorted streams (i.e. "merge-sort")

ivan-klass:
An implementation of "merge-sort" merging for two streams of sorted data.
This might be useful in practice when combining streams from different underlying sources.

[Discord conversation](https://discord.com/channels/632277896739946517/632310980449402880/1063932797145849888)

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1997,11 +1997,21 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       val lChunk = leftLeg.head
       val rChunk = rightLeg.head
       if (lChunk.nonEmpty && rChunk.nonEmpty) { // the only case we need chunk merging and sorting
-        val emitUpTo = order.min(lChunk(lChunk.size - 1), rChunk(rChunk.size - 1))
-        val emitLeftCount = lChunk.indexWhere(order.gt(_, emitUpTo)).getOrElse(lChunk.size)
-        val emitRightCount = rChunk.indexWhere(order.gt(_, emitUpTo)).getOrElse(rChunk.size)
-        val (emitLeft, keepLeft) = lChunk.splitAt(emitLeftCount)
-        val (emitRight, keepRight) = rChunk.splitAt(emitRightCount)
+        val lLast = lChunk(lChunk.size - 1)
+        val rLast = rChunk(rChunk.size - 1)
+        val wholeLeftSide = Order.lteqv(lLast, rLast) // otherwise we can emit whole right
+        val (emitLeft, keepLeft) =
+          if (wholeLeftSide) (lChunk, Chunk.empty)
+          else
+            lChunk.splitAt(
+              lChunk.indexWhere(order.gt(_, rLast)).getOrElse(lChunk.size)
+            )
+        val (emitRight, keepRight) =
+          if (!wholeLeftSide) (rChunk, Chunk.empty)
+          else
+            rChunk.splitAt( // not emitting equal from right side to keep stable sorting
+              rChunk.indexWhere(order.gteq(_, lLast)).getOrElse(rChunk.size)
+            )
         Pull.output(
           Chunk.vector((emitLeft ++ emitRight).toVector.sorted(order))
         ) >> go(leftLeg.setHead(keepLeft), rightLeg.setHead(keepRight))
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1980,14 +1980,14 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     that.mergeHaltL(this)
 
   /** Given two sorted streams emits a single sorted stream, like in merge-sort.
-    * For entries that are considered equal by the Order, left stream element is emmitted first.
+    * For entries that are considered equal by the Order, left stream element is emitted first.
     * Note: both this and another streams MUST BE ORDERED already
     * @example {{{
-    * scala> Stream(1, 2, 5, 6).mergeSorted(Stream(0, 2, 3, 4)).toList
+    * scala> Stream(1, 2, 5, 6).interleaveOrdered(Stream(0, 2, 3, 4)).toList
     * res0: List[Int] = List(0, 1, 2, 2, 3, 4, 5, 6)
     * }}}
     */
-  def mergeSorted[F2[x] >: F[x], O2 >: O: Order](that: Stream[F2, O2]): Stream[F2, O2] = {
+  def interleaveOrdered[F2[x] >: F[x], O2 >: O: Order](that: Stream[F2, O2]): Stream[F2, O2] = {
     val order = Order[O2].toOrdering // collections API needs Ordering, not cats.Order
 
     def go(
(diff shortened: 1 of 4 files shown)


mpilquist:
Thanks! Haven't looked at this PR yet but it reminded me of:
- https://gist.github.com/johnynek/689199b4ac49364e7c94abef996ae59f
- https://gist.github.com/gatorcse/1f92aa7e52a04d9c91511ca79f73e911

mpilquist:
Approach looks good to me (though haven't done a line by line review yet). I lean towards coming up with a different name, as `merge` in fs2 implies concurrency whereas this combinator doesn't do concurrent pulls of the two sources (and I agree it shouldn't). Our deterministic combinators that combine two streams include `zip` and variants, `interleave`. Any ideas?

ivan-klass:
> Approach looks good to me (though haven't done a line by line review yet). I lean towards coming up with a different name, as `merge` in fs2 implies concurrency whereas this combinator doesn't do concurrent pulls of the two sources (and I agree it shouldn't). Our deterministic combinators that combine two streams include `zip` and variants, `interleave`. Any ideas?

Totally agree. While "merge" is associated with concurrency, "zip" usually means a product over original element type, so I think "interleave" is a good verb, what about "interleaveSorted"?
Also, could you please suggest a better suite file to place a test? Or should it be a new one? Thanks

mpilquist:
+1 for `interleaveSorted`

mpilquist (approved):
Looks good to me besides the name bike-shedding. Thanks!

ivan-klass:
@mpilquist @diesalbla I've updated method name and found a better place for test

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1979,6 +1979,57 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+
+    val thisPull = covaryAll[F2, O2].pull
+    val thatPull = that.pull
+
+    (thisPull.stepLeg, thatPull.stepLeg).tupled.flatMap {
+      case (Some(leg1), Some(leg2)) => go(leg1, leg2)
+      case (_, None)                => thisPull.echo
+      case (None, _)                => thatPull.echo
+    }.stream
+  }
```
ivan-klass, line 2031: I've just realised it may seem much more elegant with `.uncons` - what do you think?
```suggestion
    def go(left: Pull[F2, O2, Unit], right: Pull[F2, O2, Unit]): Pull[F2, O2, Unit] =
      left.uncons.flatMap {
        case Some((lChunk, lNext)) =>
          right.uncons.flatMap {
            case Some((rChunk, rNext)) =>
              val emitUpTo = order.min(lChunk(lChunk.size - 1), rChunk(rChunk.size - 1))
              val emitLeftCount = lChunk.indexWhere(order.gt(_, emitUpTo)).getOrElse(lChunk.size)
              val emitRightCount = rChunk.indexWhere(order.gt(_, emitUpTo)).getOrElse(rChunk.size)
              val (emitLeft, keepLeft) = lChunk.splitAt(emitLeftCount)
              val (emitRight, keepRight) = rChunk.splitAt(emitRightCount)
              Pull.output(Chunk.vector((emitLeft ++ emitRight).toVector.sorted(order))) >> go(
                if (keepLeft.isEmpty) lNext else Pull.output(keepLeft) >> lNext,
                if (keepRight.isEmpty) rNext else Pull.output(keepRight) >> rNext
              )
            case None => Pull.output(lChunk) >> lNext
          }
        case None => right
      }
    go(this.underlying, that.underlying).stream
  }
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1979,6 +1979,38 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+    def go(left: Pull[F2, O2, Unit], right: Pull[F2, O2, Unit]): Pull[F2, O2, Unit] =
+      left.uncons.flatMap {
+        case Some((lChunk, lNext)) =>
+          right.uncons.flatMap {
+            case Some((rChunk, rNext)) =>
+              val emitUpTo = order.min(lChunk(lChunk.size - 1), rChunk(rChunk.size - 1))
+              val emitLeftCount = lChunk.indexWhere(order.gt(_, emitUpTo)).getOrElse(lChunk.size)
+              val emitRightCount = rChunk.indexWhere(order.gt(_, emitUpTo)).getOrElse(rChunk.size)
+              val (emitLeft, keepLeft) = lChunk.splitAt(emitLeftCount)
+              val (emitRight, keepRight) = rChunk.splitAt(emitRightCount)
```
diesalbla, line 2002: Perhaps we should add a more elaborate function, one which operates in a single pass through each chunk and runs in linear time, as a single loop with two indices, one for each chunk chunks. 

That can be a matter for a subsequent PR, though. This one looks good enough.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1979,6 +1979,38 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+    val order = Order[O2].toOrdering // collections API needs Ordering, not cats.Order
+
+    def go(left: Pull[F2, O2, Unit], right: Pull[F2, O2, Unit]): Pull[F2, O2, Unit] =
+      left.uncons.flatMap {
+        case Some((lChunk, lNext)) =>
+          right.uncons.flatMap {
+            case Some((rChunk, rNext)) =>
+              val emitUpTo = order.min(lChunk(lChunk.size - 1), rChunk(rChunk.size - 1))
+              val emitLeftCount = lChunk.indexWhere(order.gt(_, emitUpTo)).getOrElse(lChunk.size)
+              val emitRightCount = rChunk.indexWhere(order.gt(_, emitUpTo)).getOrElse(rChunk.size)
```
diesalbla, line 2000: Quick optimisation: we can compare the last element of each chunk to the `emitUpTo`, to decide quickly if we use the whole chunk. This should save a run through one of the chunks.

diesalbla (approved):

mpilquist:
This combinator must be implemented with `stepLeg` and not `uncons`. See https://github.com/typelevel/fs2/issues/1678#issuecomment-549569203 for details as to why.

ivan-klass:
> This combinator must be implemented with `stepLeg` and not `uncons`. See [#1678 (comment)](https://github.com/typelevel/fs2/issues/1678#issuecomment-549569203) for details as to why.

@mpilquist thank you for the explanation, just reverted the last commit

core/shared/src/main/scala/fs2/Stream.scala
ivan-klass: Done, I've just realized that at least one leg head is always emitted

core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
```diff
@@ -29,14 +29,16 @@ import cats.effect.{IO, SyncIO}
...
-import org.scalacheck.Gen
+import org.scalacheck.{Arbitrary, Gen}
 import org.scalacheck.Prop.forAll
 
 import scala.concurrent.duration._
 import scala.concurrent.TimeoutException
 
 class StreamCombinatorsSuite extends Fs2Suite {
 
+  override def scalaCheckInitialSeed = "CGcLMAcUebElb0kzuHuoGp8ZwxhJT9ztcbnWc_4sYHK="
```
ivan-klass, line 40: oops, my bad

Issue #3081, opened 2022-12-15
Scope lookup failure with `hold1`


Jasper-M:
If I use `hold1` my stream terminates with an error:

> java.lang.RuntimeException: Scope lookup failure!
> 
> This is typically caused by uncons-ing from two or more streams in the same Pull.
> To do this safely, use `s.pull.stepLeg` instead of `s.pull.uncons` or a variant
> thereof. See the implementation of `Stream#zipWith_` for an example.

If I switch to `holdOption` the problem goes away.

```scala
val s2 = for {
  sig <- Stream.range(0, 10).covary[IO].metered(50.millis).debug(x => s"SIG: $x").hold1
  res <- Stream.range(0, 10).covary[IO].metered(120.millis).debug(x => s"STREAM: $x").evalMap(x => sig.get.map(x -> _))
} yield res

s2.printlns.onFinalize(IO.println("END")).compile.drain.unsafeRunSync()
```
output:
```
SIG: 0
END
java.lang.RuntimeException: Scope lookup failure!
...
```

Scastie: https://scastie.scala-lang.org/dLpc9VlJRs6DbWu2uUQMGg

armanbilge:
@Jasper-M since you are becoming a regular contributor: would you be interested to work on this? 😁 I guess the solution is to just do what the exception says and replace the `uncons` with `stepLeg`.

Jasper-M:
If I understood why it doesn't work in the first place 😅

armanbilge:
> Actually I think this is a fundamental issue with `uncons`: it seems unsafe to use it to build stream combinators if you do not control the `Stream` you are unconsing.

Ok, this statement was probably a bit overblown 😅 sorry. After playing with it some more, `uncons` is not the issue, since I have the same problem after rewriting with `stepLeg`. The problem seems to be running the tail of a `zip`ped `Stream `.concurrently`.

PR #3112, opened 2023-01-20
Fix `hold1`

armanbilge:
Fixes https://github.com/typelevel/fs2/issues/3081.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1541,15 +1541,20 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     map(Some(_): Option[O2]).hold(None)
 
   /** Like [[hold]] but does not require an initial value. The signal is not emitted until the initial value is emitted from this stream */
-  def hold1[F2[x] >: F[x]: Concurrent, O2 >: O]: Stream[F2, Signal[F2, O2]] =
-    this.pull.stepLeg.flatMap {
-      case Some(leg) =>
-        val hd = leg.head
-        val tl = leg.next
-        val ntl = if (hd.size == 1) tl else Pull.output(hd.drop(1)) >> tl
-        ntl.stream.hold[F2, O2](hd(0)).underlying
+  def hold1[F2[x] >: F[x]: Concurrent, O2 >: O]: Stream[F2, Signal[F2, O2]] = {
+    def go(signal: Deferred[F2, Signal[F2, O2]]) = this.pull.uncons1.flatMap {
+      case Some((o, tail)) =>
+        Pull.eval(SignallingRef.of[F2, O2](o).flatTap(signal.complete(_))).flatMap { ref =>
+          tail.foreach(ref.set(_)).underlying
+        }
+
       case None => Pull.done
     }.streamNoScope
+
+    Stream.eval(Deferred[F2, Signal[F2, O2]]).flatMap { signal =>
+      Stream.eval(signal.get).concurrently(go(signal))
+    }
+  }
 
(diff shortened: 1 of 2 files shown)


Jasper-M (approved):
Still a shame that `uncons` and `concurrently` don't seem to be composable.

mpilquist:
> Still a shame that uncons and concurrently don't seem to be composable.

I wouldn't necessarily conclude that. There's some bug hiding here that broke `hold1` but I'd be surprised if that bug was something fundamental about `uncons` and `concurrently` interactions.

armanbilge:
> I'd be surprised if that bug was something fundamental about `uncons` and `concurrently` interactions.

Yes, it needed the `zip` as well.

I don't _really_ understand how all this scoping works, but it seemed like what was happening was some scope was being opened in the `uncons`, and because the tail was being run elsewhere `concurrently`, it was unable to locate this already open scope. If my understanding is right, then we are lacking some mechanism to thread that opened scope through.

armanbilge:
Wait, maybe there is a way :) let me try something

https://github.com/typelevel/fs2/blob/597102bcad333c43c7fd9a413d1e16fdbe6208eb/core/shared/src/main/scala/fs2/Pull.scala#L460-L467

armanbilge:
No, that didn't work either. be8d3e741e06b9f156e0bc2850262cf342caa7ce

armanbilge:
Yeah, so `extendScopeTo` extended the _lifetime_ of the scope, but it still didn't use that scope to compile the `concurrent` stream.

armanbilge:
I experimented with the idea that the scope of the background stream in `concurrently` should be part of the scope tree of the foreground stream, instead of creating its own root scope. I believe that would solve this issue, without requiring to change the implementation of `hold1`.

I tried two different approaches, either interfacing directly with `Pull.compile` or adding a new API `stream.compileScope(scope)`.

- https://github.com/typelevel/fs2/commit/779a1709fd74db8856cf5c296f81e9eceb0a5367 fixes `hold1`, but breaks `concurrently`
- https://github.com/typelevel/fs2/commit/576acc46495e8d9f6a8954da21465fff68ec1785 didn't fix `hold1`, but didn't break `concurrently`

Once I've mustered up new courage I'll probably take another shot at it.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1541,12 +1541,21 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+      case Some((o, tail)) =>
+        Pull.eval(SignallingRef.of[F2, O2](o).flatTap(signal.complete(_))).flatMap { ref =>
+          tail.foreach(ref.set(_)).underlying
+        }
+
+      case None => Pull.done
     }.streamNoScope
 
+    Stream.eval(Deferred[F2, Signal[F2, O2]]).flatMap { signal =>
+      Stream.eval(signal.get).concurrently(go(signal))
```
Jasper-M, line 1555: This will result in `Stream.never` if the signal never gets completed, e.g. if `this` is empty.

Jasper-M: I'm not sure. An empty stream is also an option. That's what the original implementation did afaict.

armanbilge: Ah I see, we are not concerned with the case `Stream.never.hold1` (which never emits but never ends) but with the case `Stream.empty.hold1`. So there should be an `onFinalize` or something.

armanbilge: Thanks, the fix was even easier than I realized, no need for `onFinalize` :)

PR #3113, opened 2023-01-22
Add limit combinator to Stream

sgjbryan:
# What

Adds a `limit` combinator to `Stream`, which emits the first n elements and raises an error if there are more elements.

# Why

I've implemented this a few times recently, and it seems to be sufficiently general to be worth implementing efficiently as a canoncial combinator. Example use case might be a middleware which enforces size limits on streams.

I'm not sure what the desired approach is to combinator generality - so it's completely fine if the consensus here is that it's not useful enough to warrant being included in the `Stream` API. Also hppy to rework the naming if desired, alternatives might be `limitOrError` or something like `maxSize`.

---

Thanks for opening a PR!

Target the `main` branch if this PR is targeted for the 3.x series and the `series/2.5.x` branch if targeted for the 2.5.x series.

If the CI build fails due to Scalafmt, run `sbt scalafmtAll` and push the changes (generally a good idea to run this prior to opening the PR).

Feel free to ask questions on the fs2 and fs2-dev channels on the [Typelevel Discord server](https://discord.gg/9V8FZTVZ9R).

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1764,8 +1764,9 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       .flatMap {
         case Some(s) =>
           s.pull.uncons.flatMap {
-            case Some(_) => Pull.raiseError(new IllegalStateException)
-            case None    => Pull.done
+            case Some(_) =>
+              Pull.raiseError(new IllegalStateException(s"limit($n) emitted more than $n elements"))
+            case None => Pull.done
           }
         case _ => Pull.done
       }
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1755,6 +1755,22 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       case None    => Pull.output1(fallback)
     }.stream
 
+  /** Emits the first `n` elements of this stream,
+    * raising an IllegalStateException if there are more elements.
+    */
+  def limit[F2[x] >: F[x]](n: Long)(implicit rt: RaiseThrowable[F2]): Stream[F2, O] =
+    this.pull
+      .take(n)
+      .flatMap {
+        case Some(s) =>
+          s.pull.uncons.flatMap {
+            case Some(_) => Pull.raiseError(new IllegalStateException)
+            case None    => Pull.done
+          }
+        case _ => Pull.done
+      }
+      .stream
+
   /** Applies the specified pure function to each input and emits the result.
     *
     * @example {{{
--- a/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
+++ b/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
@@ -1465,6 +1465,49 @@ class StreamCombinatorsSuite extends Fs2Suite {
     }
   }
 
+  group("limit") {
+    test("limit a stream with > n elements, leaving chunks untouched") {
+      val s = Stream(1, 2) ++ Stream(3, 4)
+      s.covary[IO]
+        .limit[IO](5)
+        .chunks
+        .map(_.toList)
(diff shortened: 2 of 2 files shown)


diesalbla:
@sgjbryan Good evening, Sam. Thanks for the PR. As it happens, something like this is also included in the `http4s` library, and it certainly seems a desirable feature, so this should be a nice addition.

The name `limit` is a bit ambiguous, though. How would you feel about `restrainOutputSizeTo(maxSize: Long)`? 

Finally, rather than issuing an `IllegalState` exception, could we perhaps issue a more specific one, such as `StreamExceededSize` or `TooLargeStream`, or similar?

sgjbryan:
Great, thanks for the quick review!

It seems to me that the imperative tone of `restrainOutputSizeTo` isn't really in keeping with the naming conventions of the other combinators - how about `limitLength` or `limitSize`? I'd be quite content with any of `{limit|restrict|constrain}{Length|Size}`.

I would normally agree with raising a more specific exception, but it didn't appear to be the convention here (and I couldn't find any natural place to define a narrower exception). In principle I'd be happy to introduce something like `fs2.error.StreamExceedSizeException extends IllegalStateException` if that would be the preferred approach

sgjbryan:
Just done some digging and I'm guessing [this](https://github.com/http4s/http4s/blob/b164566faaca8616a2ccc4a45ee3d98ba93ba3a9/server/shared/src/main/scala/org/http4s/server/middleware/EntityLimiter.scala#L49-L56) is the equivalent `http4s` implementation - thanks for mentioning that

mpilquist:
I don't think we should add a custom exception type for this. We don't do it for other combinators with similar error cases (e.g. `lastOrError`).

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1755,6 +1755,22 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+  /** Emits the first `n` elements of this stream,
+    * raising an IllegalStateException if there are more elements.
+    */
+  def limit[F2[x] >: F[x]](n: Long)(implicit rt: RaiseThrowable[F2]): Stream[F2, O] =
+    this.pull
+      .take(n)
+      .flatMap {
+        case Some(s) =>
+          s.pull.uncons.flatMap {
+            case Some(_) => Pull.raiseError(new IllegalStateException)
```
mpilquist, line 1767: I suggest `new IllegalStateException(s"limit($n) emitted more than $n elements")` or something similar.

diesalbla:
> It seems to me that the imperative tone of restrainOutputSizeTo isn't really in keeping with the naming conventions of the other combinators - how about limitLength or limitSize? I'd be quite content with any of {limit|restrict|constrain}{Length|Size}.

Well, `take`,  `filter`, `collect`, or `map`, or `zipWith` are also themselves verbs, even though they all have the semantics not of modifying the immutable _description_ (the immutable `Stream` object), but of creating a new immutable objects that wraps around the initial one. So it is not really a break from the rest. :)

sgjbryan:
That's true, I was probably a bit sloppy calling it imperative - I think it's probably the _fluency_ (for want of a better word) that seems a bit out of place to me. That being said, there are examples of both (e.g. combinators like `dropLastIf` and `delayBy` are fluent when read with their arguments, but ones like `interleave` or `concurrently` aren't `interleaveWith` or `concurrentlyWith`).

To me, `limit` feels like more natural language than `restrain` - I think users would be able to find `limitSize` via IDE autocomplete but I personally don't think `restrainOutputSize` is particularly discoverable (other than if a user searches for `size`). In terms of pure discoverability I think `limitOrError` or `takeOrError` is probably most consistent.

I don't have a strong opinion though, so if general opinion is with `restrainOutputSizeTo` then happy to go with that

mpilquist:
`limit` seems fine to me given operations like `chunkLimit`.

core/shared/src/main/scala/fs2/Stream.scala
sgjbryan: Added in https://github.com/typelevel/fs2/pull/3113/commits/98141abd75393d9dc0ad60b9236770e34fd3a5ca

mpilquist (approved):

PR #3120, opened 2023-01-31
Mask `getBasicFileAttributes` errors in `Files[F].walk`

fthomas:
This shows a problem with `Files[F].walk` where it does not walk the whole file tree if files are deleted during the process.  The added tests fails with:
```
==> X fs2.io.file.FilesSuite.walk - can delete files in a nested tree  0.029s munit.ComparisonFailException: fs2/io/shared/src/test/scala/fs2/io/file/FilesSuite.scala:537
536:        .foldMonoid
537:        .assertEquals(25)
538:    }
values are not the same
=> Obtained
5
=> Diff (- obtained, + expected)
-5
+25
```
Only 5 files out of 25 where deleted. I expected that the whole process deletes 25 files because `tempFilesHierarchy` creates 5 child directories with 5 files in each child directory.

If `Files[IO].deleteIfExists(path).as(1)` is replaced with `IO.pure(1)` the test passes.

### DIFF
--- a/io/shared/src/main/scala/fs2/io/file/Files.scala
+++ b/io/shared/src/main/scala/fs2/io/file/Files.scala
@@ -468,27 +468,28 @@ object Files extends FilesCompanionPlatform {
         Stream.emit(start) ++ {
           if (maxDepth == 0) Stream.empty
           else
-            Stream.eval(getBasicFileAttributes(start, followLinks = false)).flatMap { attr =>
+            Stream.eval(getBasicFileAttributes(start, followLinks = false)).mask.flatMap { attr =>
               if (attr.isDirectory)
                 list(start).flatMap { path =>
                   go(path, maxDepth - 1, attr.fileKey.toRight(start) :: ancestry)
                 }.mask
               else if (attr.isSymbolicLink && followLinks)
-                Stream.eval(getBasicFileAttributes(start, followLinks = true)).flatMap { attr =>
-                  val fileKey = attr.fileKey
-                  val isCycle = Traverse[List].existsM(ancestry) {
-                    case Right(ancestorKey) => F.pure(fileKey.contains(ancestorKey))
-                    case Left(ancestorPath) => isSameFile(start, ancestorPath)
-                  }
+                Stream.eval(getBasicFileAttributes(start, followLinks = true)).mask.flatMap {
+                  attr =>
+                    val fileKey = attr.fileKey
+                    val isCycle = Traverse[List].existsM(ancestry) {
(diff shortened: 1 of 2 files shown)


armanbilge:
Thanks for the reproducer! So the issue is partly related to this `.mask`, which swallows the exception and stops iterating over the directory.

https://github.com/typelevel/fs2/blob/134041f59d318a4e5ef98fb4cd6f01e8eac0875e/io/shared/src/main/scala/fs2/io/file/Files.scala#L472-L475

If I take it away I get this.
```
==> X fs2.io.file.FilesSuite.walk - can delete files in a nested tree  0.482s java.nio.file.NoSuchFileException: /tmp/BaseFileSpec9152885937877967676/BaseFileSpec10790971575620310752/BaseFileSpecSub14584697728072684076.tmp
```

armanbilge:
Ah yes, more `mask`ing seems to have solved it 😂

io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -468,27 +468,28 @@ object Files extends FilesCompanionPlatform {
...
+                    }
+
+                    Stream.eval(isCycle).flatMap { isCycle =>
+                      if (!isCycle)
+                        list(start).mask.flatMap { path =>
+                          go(path, maxDepth - 1, attr.fileKey.toRight(start) :: ancestry)
+                        }
+                      else
+                        Stream.raiseError(new FileSystemLoopException(start.toString))
+                    }
```
diesalbla, line 492: I wonder if some of this could be contained within the `F` effect itself, as there is no much use for streaming features: 

```suggestion
                getBasicFileAttributes(start, followLinks = true).handleError(_ => ()).flatMap {
                  attr =>
                    val fileKey = attr.fileKey
                    val isCycle = Traverse[List].existsM(ancestry) {
                      case Right(ancestorKey) => F.pure(fileKey.contains(ancestorKey))
                      case Left(ancestorPath) => isSameFile(start, ancestorPath)
                    }

                    isCycle.flatMap {
                      case false =>
                        list(start).mask.flatMap { path =>
                          go(path, maxDepth - 1, attr.fileKey.toRight(start) :: ancestry)
                        }
                      case true => F.raiseError(new FileSystemLoopException(start.toString))
                    }
```

fthomas: `go` is a `Stream[F, Path]` and you `flatMap` from `F` to it. I don't think this compiles.

PR #3130, opened 2023-02-07
Optimize the **Java** `Flow` interop

BalmungSan:
Follow up of #3102

- [x] Reduce allocations on `StreamSubscription`
- [x] Reduce allocations in the internal buffer of `StreamSubscriber`
- [ ] ~Cross compile~ _(will be done in a separate PR #3160)_


core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -96,75 +107,75 @@ private[flow] final class StreamSubscription[F[_], A] private (
...
+        val canceled = F.async_[Unit] { cb =>
+          cancelToken.set(() => cb(Either.unit))
+        }
+
+        cancelToken -> canceled
+      }
+    ).mapN { case (requests, resume, (cancelToken, canceled)) =>
       new StreamSubscription(
         stream,
         subscriber,
```
diesalbla, line 165: Is there much value in adding a few `F.delay`? The only difference of this overhead is to allow the dispatcher to shift to another task, but would we want that? 

```suggestion
      F.delay { 
        val requests = new AtomicLong(0L)
        val resume = new AtomicReference(defaultCallback)
        val cancelToken = new AtomicReference(defaultCallback)
        val canceled = F.async_[Unit] { cb =>
          cancelToken.set(() => cb(Either.unit))
        }
        new StreamSubscription(
          stream,
          subscriber,
```

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -77,94 +96,96 @@ private[flow] final class StreamSubscription[F[_], A] private (
...
         case Outcome.Succeeded(result) =>
           result.flatMap {
-            case Left(())  => onComplete // Events finished normally.
+            case Left(())  => F.delay(onComplete()) // Events finished normally.
             case Right(()) => F.unit // Events was canceled.
           }
-        case Outcome.Errored(ex) => onError(ex)
-        case Outcome.Canceled()  => cancelMe
+        case Outcome.Errored(ex) => F.delay(onError(ex))
+        case Outcome.Canceled()  => F.delay(onComplete())
```
BalmungSan, line 107: If the `run` is canceled what should we do?
Right now it is behaving as if the `Stream` finished normally... but I am not totally comfortable with that.

armanbilge: `onError(new CancellationException)` ?

BalmungSan: Yeah we can do that, or we could call `cancel`.
My question is what are the right / best semantics?

So, if I have something like:

```scala
val data: Stream[IO, Foo] = ...
val thirdPartySubscriber: Subscriber[Foo] = ...
val killSwitch: F[Unit] = ...

data.subscribe(thirdPartySubscriber).race(killSwitch)
```

And then `killSwitch` is turned on.
The initial gut feeling is that, from the perspective of the `thirdPartySubscriber`, we just finished producing data; e.g. `onComplete`. This makes a lot of sense if `data` is just infinite and the `killSwitch` is the shutdown of the whole application. Nevertheless, if `CancellationException` is made public, this could be documented behavior.
Moreover, what if `killSwitch` is there to signal a failure in the application? We probably want to signal the downstream consumer that something wrong happened; e.g. `onError`
Finally, we may also assume that from the perspective of `subscribe` the `killSwitch` is just another way that _downstream_ has to signal cancellation; e.g. not different from `cancel`.

Thinking about it, I think I either prefer to use a public exception that downstream users may catch, or let users configure the behavior.

armanbilge: I think in that case stream interruption should be used (for graceful closure) instead of cancellation.

BalmungSan (commented):
Some questions:

1. Was just moving the files enough for cross-compiling?
2. For some reason **Scala 3** didn't like my covariant `State` so I had to make it invariant which was a real PITA.
3. I guess the `ClassTag` change is a non-go... but then, what should we do?

armanbilge:
1. No, we should also run the tests to check that it actually works (although not the TCK ones, since those won't cross-compile). When I checked this the current API was using stuff that doesn't cross-compile. Not sure if that's still the case after your changes.

2. 🙃 

3. Why do we need `ClassTag` now? The answer is, don't need it 😂

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -0,0 +1,326 @@
...
+
+        case state =>
+          Failed(new InvalidStateException(operation = "Received subscription", state)) -> run {
+            s.cancel()
+          }
+      }
+
+      case Next(a) => {
+        case WaitingOnUpstream(idx, buffer, cb, s) =>
+          implicit val ct = ClassTag[A](a.getClass)
```
armanbilge, line 134:
```suggestion
          implicit val ct: ClassTag[A] = ClassTag[A](a.getClass)
```

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -39,35 +40,53 @@ import java.util.concurrent.Flow.{Subscription, Subscriber}
...
+            Pull.eval(F.asyncCheckAttempt[Unit] { cb =>
+              F.delay {
+                // If there aren't more pending request,
+                // we will wait until the next one.
+                resume.set(() => cb.apply(Either.unit))
+                // However, before blocking,
+                // we must check if it has been a concurrent request.
+                // In case it was, we abort the wait.
+                if (requests.get() > 0)
+                  Either.unit
```
armanbilge, line 75: If we abort, should we clear `resume` of the callback? At the very least to avoid a memory leak.

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -76,95 +95,93 @@ private[flow] final class StreamSubscription[F[_], A] private (
...
+            // Overflow.
+            Long.MaxValue
+          } else {
+            result
+          }
+        }
 
-        case Some(()) =>
-          F.unit
+        resume.get().apply()
```
armanbilge, line 154: Also a candidate for a `getAndSet` clear.

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -76,95 +95,93 @@ private[flow] final class StreamSubscription[F[_], A] private (
         .compile
```
armanbilge, line 95: The code right above this is:
```scala
      stream
        .through(subscriptionPipe)
        .foreach(x => F.delay(sub.onNext(x)))
```
We can optimize it as:
```scala
      stream
        .through(subscriptionPipe)
        .chunks
        .foreach(chunk => F.delay(chunk.foreach(sub.onNext(_))))
```

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -76,95 +95,93 @@ private[flow] final class StreamSubscription[F[_], A] private (
...
-          if (n == java.lang.Long.MAX_VALUE)
-            requests.offer(StreamSubscription.Request.Infinite)
-          else if (n > 0)
-            requests.offer(StreamSubscription.Request.Finite(n))
-          else
-            onError(
-              ex = new IllegalArgumentException(s"Invalid number of elements [${n}]")
-            )
+  override def request(n: Long): Unit =
+    if (canceled.get() ne null) {
```
armanbilge, line 138: Is this `canceled` check really doing anything useful?

armanbilge: Oh, yes it is, this comment would almost be more helpful here.
https://github.com/typelevel/fs2/blob/1f6bc986c1c2233a53d044c8462269d80380f771/core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala#L124-L129

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala
BalmungSan: @armanbilge uhm not sure in this case, since I need the custom logic to ensure it always stays on `Long.MaxValue`

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala
BalmungSan: Well, that was the original place of the comment, I just polished it a little.
I can move it if you prefer, the thing is that it also affects `cancel`, and well both methods are close together.

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscription.scala
armanbilge: Not sure I understand. It's only possible to use this `resume` callback at most once, so what is the point of keeping it after it is used?

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -38,230 +38,289 @@ import java.util.concurrent.atomic.AtomicReference
...
+          Failed(new InvalidStateException(operation = "Received subscription", state)) -> run {
+            s.cancel()
+          }
+      }
 
-    /** Called when upstream has finished sending records. */
-    def onComplete(): Unit
+      case Next(a) => {
+        case WaitingOnUpstream(idx, buffer, cb, s) =>
+          implicit val ct = ClassTag[A](a.getClass)
```
armanbilge, line 134: I don't think this is safe. What happens if `a` is a subclass of `A`, and the subsequent elements are `A` or some different subclass?

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -38,230 +37,288 @@ import java.util.concurrent.atomic.AtomicReference
...
-            if (newBuffer.size == chunkSize)
-              Idle(s, Chunk.empty) -> (() => r(newBuffer.some.asRight))
-            else
-              WaitingOnUpstream(s, newBuffer, r) -> (() => ())
+        case WaitingOnUpstream(idx, buffer, cb, s) =>
+          Terminal -> run {
+            if (idx == 0) {
+              cb.apply(Right(None))
+            } else {
+              cb.apply(Right(Some(Chunk.array(buffer, offset = 0, length = idx))))
```
armanbilge, line 201: How does the `Stream` terminate in this case? Since it needs a `None` to terminate.

BalmungSan: After the next pull, it receives the `None` :)
I would love to send that `None` right away, but we need another callback for that.

armanbilge (approved):
The diff was horrible, but when viewed standalone the new implementations are quite elegant. Thanks for bearing with me 😇

core/jvm/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -38,230 +37,305 @@ import java.util.concurrent.atomic.AtomicReference
...
-    final def stream(subscribe: F[Unit])(implicit ev: MonadThrow[F]): Stream[F, A] =
-      Stream.bracket(subscribe)(_ => onFinalize) >>
-        Stream
-          .repeatEval(dequeue1)
-          .rethrow
-          .unNoneTerminate
-          .unchunks
-  }
+        case Uninitialized(Some(cb)) =>
+          WaitingOnUpstream(idx = 0, buffer = null, cb, s) -> run {
```
armanbilge, line 122: Actually, why not allocate the array here? Then we can remove the `idx == 0` conditional below in `OnNext`.

armanbilge: Good point. Since there could be significant lag between requesting elements and when they arrive.

PR #3147, opened 2023-02-19
Optimize `Chunk#to{Byte,Char}Buffer`, add note about mutability

armanbilge:
We had some unsuspended uses of `toByteBuffer` which I am fixing in d76bdea6cd9256314ef3df26014f3cedaeeeda72.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -306,7 +306,10 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
       case _ => Chunk.ArraySlice(toArray, 0, size)
     }
 
-  /** Converts this chunk to a `java.nio.ByteBuffer`. */
+  /** Converts this chunk to a `java.nio.ByteBuffer`.
+    * @note that even "read-only" interaction with a `ByteBuffer` may increment its `position`,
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -306,30 +306,38 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
...
+    * @note that even "read-only" interaction with a `ByteBuffer` may increment its `position`,
+    * so this method should be considered as unsafely allocating mutable state.
+    */
   def toByteBuffer[B >: O](implicit ev: B =:= Byte): JByteBuffer =
     this match {
       case c: Chunk.ArraySlice[_] if c.values.isInstanceOf[Array[Byte]] =>
         JByteBuffer.wrap(c.values.asInstanceOf[Array[Byte]], c.offset, c.length)
       case c: Chunk.ByteBuffer =>
-        val b = c.buf.asReadOnlyBuffer
+        val b = c.buf.duplicate // share contents, independent position/limit
```
armanbilge, line 318: `asReadOnlyBuffer` and `duplicate` are similar in that the returned buffer has independent position/limit.

The difference is that a read-only buffer denies access to its underlying `Array`, while a duplicated buffer does not. Being able to get at that array is an important optimization.

PR #3149, opened 2023-02-22
Introduce `Process` API

armanbilge:
Introduces an API for spawning arbitrary subprocesses and piping stdin/stdout/stderr, and implements it on JVM/Native (via [`java.lang.Process`](https://docs.oracle.com/javase/8/docs/api/java/lang/Process.html)) and on Node.js (via [`node:child_process`](https://nodejs.org/api/child_process.html)).

- `fs2.io.process.ProcessBuilder`
- `fs2.io.process.Process[F]`
- `fs2.io.process.ProcessSpawn[F]`

APIs still need some polish, bikeshed, docs, etc., please chime in! I will make a pass of inline comments.

Also published as `3.7-0ec6699-SNAPSHOT`.

### DIFF
--- a/io/js/src/main/scala/fs2/io/internal/SuspendedStream.scala
+++ b/io/js/src/main/scala/fs2/io/internal/SuspendedStream.scala
@@ -26,7 +26,7 @@ package internal
 import cats.effect.kernel.Concurrent
 import cats.effect.kernel.Resource
 import cats.effect.std.Queue
-import cats.effect.std.Semaphore
+import cats.effect.std.Mutex
 import cats.effect.syntax.all._
 
 private[io] trait SuspendedStream[F[_], O] {
@@ -45,15 +45,14 @@ private[io] object SuspendedStream {
       queue <- Queue.synchronous[F, Option[Either[Throwable, Chunk[O]]]].toResource
       _ <- stream.chunks.attempt.enqueueNoneTerminated(queue).compile.drain.background
       suspended <- F.ref(streamFromQueue(queue)).toResource
-      semaphore <- Semaphore[F](1).toResource
+      mutex <- Mutex[F].toResource
     } yield new SuspendedStream[F, O] {
 
-      override def stream: Stream[F, O] = Stream.resource(semaphore.permit).flatMap { _ =>
+      def stream: Stream[F, O] = Stream.resource(mutex.lock) >>
         Stream.eval(suspended.get).flatten.onFinalize(suspended.set(streamFromQueue(queue)))
-      }
 
-      override def getAndUpdate[O2](f: Stream[F, O] => Pull[F, O2, Stream[F, O]]): Stream[F, O2] =
-        Stream.resource(semaphore.permit).flatMap { _ =>
+      def getAndUpdate[O2](f: Stream[F, O] => Pull[F, O2, Stream[F, O]]): Stream[F, O2] =
+        Stream.resource(mutex.lock) >>
           Pull
             .eval(suspended.get)
             .flatMap(f)
@@ -61,7 +60,7 @@ private[io] object SuspendedStream {
               Pull.eval(suspended.set(tail)).void
             }
             .stream
-        }
+
     }
 
   private def streamFromQueue[F[_]: Concurrent, O](
--- a/io/js/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
@@ -44,7 +44,11 @@ private[process] trait ProcessSpawnCompanionPlatform {
               process.args.toJSArray,
               new facade.child_process.SpawnOptions {
                 cwd = process.workingDirectory.fold[js.UndefOr[String]](js.undefined)(_.toString)
-                env = (facade.process.env ++ process.env).toJSDictionary
+                env =
+                  if (process.inheritEnv)
+                    (facade.process.env ++ process.extraEnv).toJSDictionary
+                  else
+                    process.extraEnv.toJSDictionary
               }
             )
 
--- a/io/js/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
@@ -23,18 +23,12 @@ package fs2
 package io
 package process
 
-sealed trait Process[F[_]] {
+import cats.effect.kernel.Resource
+import cats.effect.kernel.Async
+import cats.syntax.all._
 
-  def isAlive: F[Boolean]
-
-  def exitValue: F[Int]
-
-  def stdin: Pipe[F, Byte, Nothing]
-
-  def stdout: Stream[F, Byte]
-
-  def stderr: Stream[F, Byte]
-
+private[process] trait ProcessSpawnCompanionPlatform {
+  implicit def forAsync[F[_]](implicit F: Async[F]): ProcessSpawn[F] = new UnsealedProcessSpawn[F] {
+    def spawn(process: ProcessBuilder): Resource[F, Process[F]] = ???
+  }
 }
-
-private[io] trait UnsealedProcess[F[_]] extends Process[F]
--- a/io/js/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
@@ -26,6 +26,8 @@ package process
 import cats.effect.kernel.Resource
 import cats.effect.kernel.Async
 import cats.syntax.all._
+import fs2.io.internal.MicrotaskExecutor
+import fs2.io.internal.SuspendedStream
 import fs2.io.internal.facade
 
 import scala.scalajs.js
@@ -62,37 +64,47 @@ private[process] trait ProcessSpawnCompanionPlatform {
             }
           }
         }
-        .map { childProcess =>
-          new UnsealedProcess[F] {
+        .flatMap { childProcess =>
+          def suspend(readable: => fs2.io.Readable) =
+            SuspendedStream(
+              Stream.resource(fs2.io.suspendReadableAndRead()(readable)).flatMap(_._2)
+            )
 
-            def isAlive: F[Boolean] = F.delay {
-              (childProcess.exitCode eq null) && (childProcess.signalCode eq null)
-            }
+          (suspend(childProcess.stdout), suspend(childProcess.stderr)).mapN {
+            (suspendedOut, suspendedErr) =>
+              new UnsealedProcess[F] {
 
-            def exitValue: F[Int] = F.asyncCheckAttempt[Int] { cb =>
-              F.delay {
-                (childProcess.exitCode: Any) match {
-                  case i: Int => Right(i)
-                  case _ =>
-                    val f: js.Function1[Any, Unit] = {
-                      case i: Int => cb(Right(i))
-                      case _      => // do nothing
-                    }
-                    childProcess.once("exit", f)
-                    Left(Some(F.delay(childProcess.removeListener("exit", f))))
+                def isAlive: F[Boolean] = F.delay {
+                  (childProcess.exitCode eq null) && (childProcess.signalCode eq null)
                 }
+
+                def exitValue: F[Int] = F.asyncCheckAttempt[Int] { cb =>
+                  F.delay {
+                    (childProcess.exitCode: Any) match {
+                      case i: Int => Right(i)
+                      case _ =>
+                        val f: js.Function1[Any, Unit] = {
+                          case i: Int => cb(Right(i))
+                          case _      => // do nothing
+                        }
+                        childProcess.once("exit", f)
+                        Left(Some(F.delay(childProcess.removeListener("exit", f))))
+                    }
+                  }
+                }
+
+                def stdin: Pipe[F, Byte, Nothing] =
+                  fs2.io.writeWritable(F.delay(childProcess.stdin))
+
+                def stdout: Stream[F, Byte] = suspendedOut.stream
+
+                def stderr: Stream[F, Byte] = suspendedErr.stream
+
               }
-            }
-
-            def stdin: Pipe[F, Byte, Nothing] = fs2.io.writeWritable(F.delay(childProcess.stdin))
-
-            def stdout: Stream[F, Byte] =
-              Stream.resource(fs2.io.suspendReadableAndRead()(childProcess.stdout)).flatMap(_._2)
-
-            def stderr: Stream[F, Byte] =
-              Stream.resource(fs2.io.suspendReadableAndRead()(childProcess.stderr)).flatMap(_._2)
 
           }
         }
+        .evalOn(MicrotaskExecutor) // guarantee that callbacks are setup before yielding to I/O
+
   }
 }
--- a/io/js/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
@@ -26,9 +26,73 @@ package process
 import cats.effect.kernel.Resource
 import cats.effect.kernel.Async
 import cats.syntax.all._
+import fs2.io.internal.facade
+
+import scala.scalajs.js
+import scala.scalajs.js.JSConverters._
 
 private[process] trait ProcessSpawnCompanionPlatform {
   implicit def forAsync[F[_]](implicit F: Async[F]): ProcessSpawn[F] = new UnsealedProcessSpawn[F] {
-    def spawn(process: ProcessBuilder): Resource[F, Process[F]] = ???
+    def spawn(process: ProcessBuilder): Resource[F, Process[F]] =
+      Resource
+        .make {
+          F.async_[facade.child_process.ChildProcess] { cb =>
+            val subprocess = facade.child_process.spawn(
+              process.command,
+              process.args.toJSArray,
+              new facade.child_process.SpawnOptions {
+                cwd = process.workingDirectory.fold[js.UndefOr[String]](js.undefined)(_.toString)
+                env = (facade.process.env ++ process.env).toJSDictionary
+              }
+            )
+
+            subprocess.once("spawn", () => cb(Right(subprocess)))
+            subprocess.once[js.Error]("error", e => cb(Left(js.JavaScriptException(e))))
+          }
+        } { childProcess =>
+          F.asyncCheckAttempt[Unit] { cb =>
+            F.delay {
+              if ((childProcess.exitCode ne null) || (childProcess.signalCode ne null)) {
+                Either.unit
+              } else {
+                childProcess.kill()
+                childProcess.once("exit", () => cb(Either.unit))
+                Left(None)
+              }
+            }
+          }
+        }
+        .map { childProcess =>
+          new UnsealedProcess[F] {
+
+            def isAlive: F[Boolean] = F.delay {
+              (childProcess.exitCode eq null) && (childProcess.signalCode eq null)
+            }
+
+            def exitValue: F[Int] = F.asyncCheckAttempt[Int] { cb =>
+              F.delay {
+                (childProcess.exitCode: Any) match {
+                  case i: Int => Right(i)
+                  case _ =>
+                    val f: js.Function1[Any, Unit] = {
+                      case i: Int => cb(Right(i))
+                      case _      => // do nothing
+                    }
+                    childProcess.once("exit", f)
+                    Left(Some(F.delay(childProcess.removeListener("exit", f))))
+                }
+              }
+            }
+
+            def stdin: Pipe[F, Byte, Nothing] = fs2.io.writeWritable(F.delay(childProcess.stdin))
+
+            def stdout: Stream[F, Byte] =
+              Stream.resource(fs2.io.suspendReadableAndRead()(childProcess.stdout)).flatMap(_._2)
+
+            def stderr: Stream[F, Byte] =
+              Stream.resource(fs2.io.suspendReadableAndRead()(childProcess.stderr)).flatMap(_._2)
+
+          }
+        }
   }
 }
--- a/io/jvm-native/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
@@ -23,15 +23,18 @@ package fs2
 package io
 package process
 
+import cats.effect.kernel.Async
 import cats.effect.kernel.Resource
-import cats.effect.kernel.Sync
+import cats.effect.syntax.all._
 import cats.syntax.all._
 import fs2.io.CollectionCompat._
 
 import java.lang
+import java.io.InputStream
+import java.util.concurrent.atomic.AtomicBoolean
 
 private[process] trait ProcessSpawnCompanionPlatform {
-  implicit def forSync[F[_]](implicit F: Sync[F]): ProcessSpawn[F] = new UnsealedProcessSpawn[F] {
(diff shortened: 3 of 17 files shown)


io/shared/src/main/scala/fs2/io/process/ProcessSpawn.scala
```diff
@@ -0,0 +1,36 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2.io.process
+
+import cats.effect.kernel.Resource
+
+sealed trait ProcessSpawn[F[_]] {
```
armanbilge, line 26: FTR I don't think this is a particularly meaningful constraint. I think it's safe to assume that if there is the `ProcessSpawn` constraint, rockets may very well be launched 😁 

I almost ripped it out. But it does do something important: it allows us to abstract over alternate implementations. For example via [NuProcess](https://github.com/brettwooldridge/NuProcess) on the JVM or io_uring/libuv on Native. So that's actually pretty useful.

io/shared/src/main/scala/fs2/io/process/Process.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def exitValue: F[Int]
+
+  def stdin: Pipe[F, Byte, Nothing]
+
+  def stdout: Stream[F, Byte]
+
+  def stderr: Stream[F, Byte]
+
+}
```
armanbilge, line 38: This API definitely needs some thinking about. Particularly about semantics of `stdin`, `stdout`, and `stderr`. Can you compile these streams at most once, or many times? Do they close the underlying streams when complete? Along these lines, do we want APIs to read/write a `Chunk` directly?

This is all possible I believe but trickier to support. Easiest of course is compile-at-most-once.

io/jvm-native/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
```diff
@@ -0,0 +1,94 @@
...
+            def stdin = writeOutputStream(F.delay(process.getOutputStream()))
+
+            def stdout = readInputStreamCancelably(F.delay(process.getInputStream()))
+
+            def stderr = readInputStreamCancelably(F.delay(process.getErrorStream()))
+
+            // this monstrosity ensures that we can cancel on-going reads from stdout/stderr without hanging
+            // to do so, it must kill the process
+            private[this] def readInputStreamCancelably(is: F[InputStream]) =
+              readInputStreamGeneric(is, F.delay(new Array[Byte](8192)), true) { (is, buf) =>
```
armanbilge, line 79: So this is fun :trollface: should this chunk size here be configurable? Where should that be exposed?

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
```diff
@@ -0,0 +1,67 @@
...
+  def withWorkingDirectory(workingDirectory: Path): ProcessBuilder
+  def withCurrentWorkingDirectory: ProcessBuilder
+
+  final def spawn[F[_]: ProcessSpawn]: Resource[F, Process[F]] =
+    ProcessSpawn[F].spawn(this)
+}
+
+object ProcessBuilder {
+
+  def apply(command: String, args: List[String]): ProcessBuilder =
```
Baccata, line 46: Out of curiosity : why not `def apply(command: String, args: String*)` ? Is there some bin-compat related reason that make it bad ?

io/shared/src/main/scala/fs2/io/process/ProcessSpawn.scala
Baccata: Is it a good idea to make it typeclass though ? I though retrospectively, `Network` being a typeclass was not so great (because there can be several implementations of it). Is there some consensus that it's okay to have a "canonical" implementation that is provided implicitly then ?

io/shared/src/main/scala/fs2/io/process/Process.scala
Baccata: > Easiest of course is compile-at-most-once. 

This matches my personal intuition

io/jvm-native/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
```diff
@@ -0,0 +1,94 @@
...
+                ()
+              },
+              F.unit
+            )
+        }
+        .map { process =>
+          new UnsealedProcess[F] {
+            def isAlive = F.delay(process.isAlive())
+            def exitValue = F.interruptible(process.waitFor())
+            def stdin = writeOutputStream(F.delay(process.getOutputStream()))
```
Baccata, line 70: A problem I've faced several times with the current implementation of `fs2.io.writeOutputStream` is the fact that chunks aren't flushed automatically: see https://github.com/neandertech/jsonrpclib/pull/58/files#r1114125452

I feel terrible for not having raised an issue yet 😞

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
armanbilge: No reason, added 👍

io/shared/src/main/scala/fs2/io/process/ProcessSpawn.scala
armanbilge: This is an excellent point. I'm not sure there is any consensus on this topic yet 😅

The current design choices mirror the rest of the library, but indeed that might not be what we want to do.

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
SystemFw: I would probably use this api in scripts most often, and I'd like a more high-level api that abstracts some common patterns, e.g. passing optional arguments, or arguments k-v pairs, or options of those, etc.
Unfortunately the way you do that is with a bunch of implicit conversions 😬

io/jvm-native/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
armanbilge: Fixed :)

io/shared/src/main/scala/fs2/io/process/Process.scala
SystemFw: I'm also curious about how piping processes looks like

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
armanbilge: > I would probably use this api 

I'm not sure if this is the API you want :) It's kind of like the JDK `Process` stuff vs Scalalib process DSL. The goal here is to provide the API and do the cross-platform stuff.

Another lib makes sense for your high-level API. If it's targeting scripts it can break bincompat more freely than a library can.

io/shared/src/main/scala/fs2/io/process/ProcessSpawn.scala
Baccata: So with that in mind, and if it was me, I'd change the `ProcessBuilder` to have the following overrides : 

```scala
def spawn[F[_]](processSpawn: ProcessSpawn[F]) : Resource[F, Process[F]] 
def spawn[F[_]: Async] : Resource[F, Process[F]]  = spawn(ProcessSpawn.forAsync[F])
def spawnStream[F[_]: Async] : Stream[F, Process[F]]  = spawnStream(ProcessSpawn.forAsync[F])
def spawnStream[F[_]](processSpawn: ProcessSpawn[F]) : Stream[F, Process[F]] = spawnStream(ProcessSpawn.forAsync[F]) 
```

This would reduce the ceremonies around starting subprocesses (which is important for scripting DX), whilst providing the flexibility to switch implementations (in particular for test purposes). As you mentioned, it's reasonably safe to assume that missiles will be launched, and therefore `Async` is a safe constraint to have on the `spawn` method.

ybasket: If we have several implementations, it should probably not be `sealed`? NuProcess and alike sound cool, but I would rather see them as standalone libs implementing this interface than blessed integrations living inside the fs2 repo (dependencies!). So more like `http4s.Client`, just with a reasonable default implementation for free.

io/shared/src/main/scala/fs2/io/process/Process.scala
armanbilge: > I'm also curious about how piping processes looks like

Not a lot of fun right now. I wonder if we added a `Monoid[Process]` for this.

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
Baccata: > Unfortunately the way you do that is with a bunch of implicit conversions 

I think this would be a nice client-side addition to decline, or some higher-level library of the sort. To elaborate, Smithy4s provides an integration with decline OOTB  : https://disneystreaming.github.io/smithy4s/docs/protocols/deriving-cli

With this PR being merged to fs2, I'd be inclined to add the client-side of things to go from high-level method calls to low level args that could be used in the ProcessBuilder, in a way that would match what is done in the decline integration. 

Not trying to advertise, just saying that seems like a job for higher-level libraries. In particular because there is no actual "standard" for things like arguments key-value pairs, etc .

io/shared/src/main/scala/fs2/io/process/ProcessSpawn.scala
armanbilge: > it should probably not be `sealed`

Well, it's an interesting question how to do this :) for example in `fs2.io.uring` I've implemented a bunch of `sealed` stuff and I think it works fine. It discourages casual extension, which theoretically leaves the door open for us to add new methods to the interface.

Of course, if we did add new methods, that would break external implementations. But those external dependencies should never be used in libraries anyway, just in applications.
- https://github.com/armanbilge/fs2-io_uring#versioning-and-compatibility

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
armanbilge: > I'd be inclined to add the client-side of things to go from high-level method calls to low level args that could be used in the ProcessBuilder, in a way that would match what is done in the decline integration.

Wow, I hadn't even thought about this! 🤩

io/shared/src/main/scala/fs2/io/process/Process.scala
Baccata: > Monoid[Process] 

Interesting idea, though what'd happen to the stderr ? :/

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
SystemFw: > Another lib makes sense for your high-level API.

Well, sort of.
I don't strongly disagree, but a couple of counterpoints:
- fs2-io started like this, then over time it became more and more a user-facing lib, I don't really see why this would be that much different
- Any additional lib is a pain for scripts, i.e. you have to remember/google id and version before you can hit the ground running.

armanbilge: > Any additional lib is a pain for scripts

```
//> using toolkit "typelevel"
```

🙂

ChristopherDavenport:
Not content with it living in a library? https://github.com/davenverse/process

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
Baccata: > Any additional lib is a pain for scripts, i.e. you have to remember/google id and version before you can hit the ground running.

@SystemFw can I interest you in a modest tool I've written for that very reason ? IIRC you're a emacs user so here's the CLI version : https://github.com/neandertech/dexsearch

Baccata:
@ChristopherDavenport my personal take is that subprocesses are a fundamental enough building block of applications that having it upstream alongside files / networking makes sense.

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
TimWSpence: > `//> using toolkit "typelevel"`

Is this a thing?!

armanbilge: It will be soon, hopefully :)

armanbilge:
Continuing from https://github.com/typelevel/fs2/pull/3149#discussion_r1114174640.

> So with that in mind, and if it was me, I'd change the `ProcessBuilder` to have the following overrides : 
> 
> ```scala
> def spawn[F[_]: Async] : Resource[F, Process[F]]  = spawn(ProcessSpawn.forAsync[F])
> def spawn[F[_]](processSpawn: ProcessSpawn[F]) : Resource[F, Process[F]] 
> def spawnStream[F[_]: Async] : Stream[F, Process[F]]  = spawnStream(ProcessSpawn.forAsync[F])
> def spawnStream[F[_]](processSpawn: ProcessSpawn[F]) : Stream[F, Process[F]]
> ```
> 
> This would reduce the ceremonies around starting subprocesses (which is important for scripting DX), whilst providing the flexibility to switch implementations (in particular for test purposes). As you mentioned, it's reasonably safe to assume that missiles will be launched, and therefore `Async` is a safe constraint to have on the `spawn` method. 

Thanks for this. This is seems good—I'm almost convinced. However I worry that library authors will just use `Async` instead of asking for `ProcessSpawn[F]` like they are supposed. I wonder if instead we should have:

```scala
def spawn : Resource[IO, Process[IO]]  = spawn(ProcessSpawn.forAsync[IO])
def spawn[F[_]](processSpawn: ProcessSpawn[F]) : Resource[F, Process[F]] 
```

By hard-coding to `IO` it will prevent tagless final library authors from cheating. IMHO we should be doing this for most things actually: provide _builders_ in terms of `Async`, but only an implicit for `IO`. This will force code written in tagless final style to correctly declare _all_ of its constraints, instead of cheatily deriving them from `Async`.

---

Also random bikeshed: rename `ProcessSpawn[F]` -> `Processes[F]`? Like `Files[F]` and `Network[F]`

io/shared/src/main/scala/fs2/io/io.scala
```diff
@@ -118,4 +121,18 @@ package object io extends ioplatform {
...
+    (F.deferred[Unit], F.delay(new AtomicBoolean(false))).flatMapN { (gate, inProgress) =>
+      F.race(
+        gate.get *> F.blocking {
+          inProgress.set(true)
+          try thunk
+          finally inProgress.set(false)
+        },
+        (gate.complete(()) *> F.never[A]).onCancel(F.delay(inProgress.get()).ifM(cancel, F.unit))
+      ).map(_.merge)
+    }
```
armanbilge, line 158: - https://github.com/typelevel/cats-effect/pull/3459

fthomas:
This looks interesting, thanks for working on it @armanbilge! Maybe I can use this in Scala Steward instead of our [homegrown code for reading the output of processes](https://github.com/scala-steward-org/scala-steward/blob/v0.22.0/modules/core/src/main/scala/org/scalasteward/core/io/process.scala).

One requirement we have is that processes never wait for input. We [ensure this by closing stdin immediately after process creation](https://github.com/scala-steward-org/scala-steward/pull/1712). A possible test for this is running `dd count=1`. Without closing stdin, this would wait for input. With closing stdin, it would return immediately.

Another requirement we have is that processes must not inherit the environment variables of Scala Steward itself. This is done by [calling `clear()` on `java.lang.ProcessBuilder#environment`](https://github.com/scala-steward-org/scala-steward/blob/v0.22.0/modules/core/src/main/scala/org/scalasteward/core/io/process.scala#L94)

It would be nice if `fs2.io.process.ProcessBuilder/Process` would support these too.

armanbilge:
@fthomas thanks for the helpful feedback!

> One requirement we have is that processes never wait for input.

In the current design you could write an empty `Stream` to stdin to close it, would that work for you? I need to verify the mechanics of this. That test is very helpful.

> Another requirement we have is that processes must not inherit the environment variables of Scala Steward itself. 

Ha, darn, was hoping I could sweep that under the rug 😂 ok thanks, I'll add support for that 😇

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
```diff
@@ -0,0 +1,75 @@
...
+  def command: String
+  def args: List[String]
+  def inheritEnv: Boolean
+  def extraEnv: Map[String, String]
+  def workingDirectory: Option[Path]
+
+  def withCommand(command: String): ProcessBuilder
+  def withArgs(args: List[String]): ProcessBuilder
+  def withInheritEnv(inherit: Boolean): ProcessBuilder
+  def withExtraEnv(env: Map[String, String]): ProcessBuilder
```
armanbilge, line 38: Bikeshed? :) Probably should add docs, they need more exposition now...

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
armanbilge: ICYMI: https://github.com/typelevel/toolkit

fthomas:
> In the current design you could write an empty `Stream` to stdin to close it, would that work for you?

Yes, that works for me. Thank you for adding a test for that and the option to disable env inheritance!

Btw, Java's `ProceesBuilder` has the option to merge stdout and stderr via [`redirectErrorStream`](https://docs.oracle.com/javase/8/docs/api/java/lang/ProcessBuilder.html#redirectErrorStream-boolean-). I guess with streams, you can achieve the same with `p.stdout merge p.stderr` but it could be much more efficient if `fs2.io.process.ProcessBuilder` would have a similar option.

armanbilge:
> but it could be much more efficient

Yes, I chased it down in OpenJDK and it seems to be using [`dup`](https://man7.org/linux/man-pages/man2/dup.2.html) under the hood. So, yeah 😅 

The annoying thing is that Node.js doesn't support this. As a workaround we can just do the stream merge ourselves.
- https://github.com/nodejs/node-v0.x-archive/issues/5831

More generally, both the JDK and Node.js support various options for these streams.
- https://nodejs.org/api/child_process.html#optionsstdio

For example, inheriting stdin/stdout/stderr from the parent process, or redirecting to a file (or even a `Socket`). So maybe this setting should be modeled with an ADT. I'd also need to check how those options are supposed to interact with `redirectErrorStream`. Might be a good follow-up PR? 😇

Tangential, but FS2's `Stream`-based APIs often miss optimization opportunities like this. For example, [`sendfile`](https://man7.org/linux/man-pages/man2/sendfile.2.html) is an awesome way to upload/download files over a socket. I believe this is exposed in the JDK as [`FileChannel#transferTo`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileChannel.html#transferTo-long-long-java.nio.channels.WritableByteChannel-) and [`FileChannel#transferFrom`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileChannel.html#transferFrom-java.nio.channels.ReadableByteChannel-long-long-). I think it's even [possible to use in Node.js](https://std.in/using-sendfile-with-nodejs/).

io/js/src/main/scala/fs2/io/internal/SuspendedStream.scala
```diff
@@ -26,7 +26,7 @@ package internal
 import cats.effect.kernel.Concurrent
 import cats.effect.kernel.Resource
 import cats.effect.std.Queue
-import cats.effect.std.Semaphore
+import cats.effect.std.Mutex
```
diesalbla, line 29: This switch from `Semaphore` to `Mutex` seems unrelated to main PR. Perhaps separate PR?

Baccata:
Continuing from https://github.com/typelevel/fs2/pull/3149#issuecomment-1444790505 

From my perspective, it does make sense. I think I'd call the monomorphic version `spawnIO`.

fthomas:
> Might be a good follow-up PR?

Makes sense, especially since this is a bigger topic. Thanks for digging into it!

io/js/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
```diff
@@ -0,0 +1,114 @@
...
+              if ((childProcess.exitCode ne null) || (childProcess.signalCode ne null)) {
+                Either.unit
+              } else {
+                childProcess.kill()
+                childProcess.once("exit", () => cb(Either.unit))
+                Left(None)
+              }
+            }
+          }
+        }
```
diesalbla, line 70:
```suggestion
  implicit def forAsync[F[_]](implicit F: Async[F]): ProcessSpawn[F] = new UnsealedProcessSpawn[F] {
    import facade.child_process.ChildProcess
    private def spawnChild(process: ProcessBuilder): F[ChildProcess] = 
      F.async_[facade.child_process.ChildProcess] { cb =>
        val subprocess = facade.child_process.spawn(
          process.command,
          process.args.toJSArray,
          new facade.child_process.SpawnOptions {
            cwd = process.workingDirectory.fold[js.UndefOr[String]](js.undefined)(_.toString)
            env =
              if (process.inheritEnv)
                (facade.process.env ++ process.extraEnv).toJSDictionary
              else
                process.extraEnv.toJSDictionary
          }
        )

        subprocess.once("spawn", () => cb(Right(subprocess)))
        subprocess.once[js.Error]("error", e => cb(Left(js.JavaScriptException(e))))
      }

    private val cleanup: ChildProcess => F[Unit] = 
      (child: ChildProcess) => F.asyncCheckAttempt[Unit] { cb =>
        F.delay {
          if ((child.exitCode ne null) || (child.signalCode ne null)) {
            Either.unit
          } else {
            child.kill()
            child.once("exit", () => cb(Either.unit))
            Left(None)
          }
        }
      }

    def spawn(process: ProcessBuilder): Resource[F, Process[F]] =
      Resource.make(spawnChild(process))(cleanup)
```

io/jvm-native/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
```diff
@@ -0,0 +1,93 @@
...
+          F.delay(process.isAlive())
+            .ifM(
+              F.blocking {
+                process.destroy()
+                process.waitFor()
+                ()
+              },
+              F.unit
+            )
+        }
```
diesalbla, line 63: Could you extract both initializer and finalizer as separate methods, and thus reduce this to `Resource.make(initialise(process))(finaliser).map { process =>` ?

io/jvm-native/src/main/scala/fs2/io/process/ProcessSpawnPlatform.scala
```diff
@@ -0,0 +1,93 @@
...
+              8192
+            )
+
+            def stderr = readInputStreamCancelable(
+              F.delay(process.getErrorStream()),
+              F.blocking(process.destroy()),
+              8192
+            )
+
+          }
```
diesalbla, line 90: Can we extract this as a subclass of `UnsealedProcess`, one which is private to this package? It is often easier to have named classes over anonymous ones.

diesalbla (approved):

io/shared/src/main/scala/fs2/io/process/Process.scala
armanbilge: I guess it would be similar to piping in the shell, only stdout. Also as suggested in https://github.com/typelevel/fs2/pull/3149#issuecomment-1445207210 we could add a feature to redirect stderr to stdout.

On second thought, less sure about `Monoid` (what's the identity process 🤔) but `Semigroup[Process]` seems plausible.

PR #3154, opened 2023-02-27
Stream - reimplement without Pull.loop

diesalbla:
The methods unfoldLoop, unfoldLoopEval, and repeatPull, are written using the `Pull.loop` function. While this is a nice recursion scheme, a lower-order code variant may be more readable and efficient.


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3966,23 +3966,23 @@ object Stream extends StreamLowPriority {
...
     */
-  def unfoldLoop[F[x] <: Pure[x], S, O](s: S)(f: S => (O, Option[S])): Stream[F, O] =
-    Pull
-      .loop[F, O, S] { s =>
-        val (o, sOpt) = f(s)
-        Pull.output1(o) >> Pull.pure(sOpt)
-      }(s)
-      .stream
+  def unfoldLoop[F[x] <: Pure[x], S, O](start: S)(f: S => (O, Option[S])): Stream[F, O] = {
+    def recur(s: S): Pull[F, O, Unit] = f(s) match {
```
mpilquist, line 3970: I'd change all these `recur` to `loop` or `go`. Not sure which is more prevalent in the library but I don't think we use `recur` much of anywhere.

diesalbla: Done.

PR #3160, opened 2023-03-08
WIP: Cross-compile `interop.flow`

BalmungSan:
Follow up of #3130 _(I will rebase this PR after that one is merged so the diff will be just a [bunch of renames](https://github.com/typelevel/fs2/pull/3160/commits/2c791ca405bc720134b6b01e57a8b71e254c336d))_.
This one is blocked since **JavaLib** doesn't provide all the `AtomicReference` methods we use.


armanbilge (approved):
+1 / -1 😎

Issue #2432, opened 2021-06-04
Fix groupWithin test that intermittently fails


mpilquist:
https://github.com/typelevel/fs2/runs/2746360895?check_suite_focus=true#step:6:849

```
==> X fs2.StreamCombinatorsSuite.groupWithin - accumulation and splitting  0.491s munit.ComparisonFailException: /home/runner/work/fs2/fs2/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala:852
850
851:        .toList
851
852:        .assertEquals(expected)
852
853:    }
853
values are not the same
854
=> Obtained
855
List(
856
  List(
857
    1,
858
    2,
859
    3
860
  ),
861
  List(
862
    4,
863
    5,
864
    6,
865
    7
866
  ),
867
  List(
868
    8,
869
    9,
870
    10,
871
    11,
872
    12
873
  ),
874
  List(
875
    13,
876
    14,
877
    15,
878
    16,
879
    17
880
  ),
881
  List(
882
    18,
883
    19,
884
    20,
885
    21,
886
    22
887
  )
888
)
889
=> Diff (- obtained, + expected)
890
     6,
891
-    7
892
+    7,
893
+    8
894
   ),
895
   List(
896
-    8,
897
     9,
898
     11,
899
-    12
900
+    12,
901
+    13
902
   ),
903
   List(
904
-    13,
905
     14,
906
     16,
907
-    17
908
+    17,
909
+    18
910
   ),
911
   List(
912
-    18,
913
     19,
```

Angel-O:
I might have a fix for it 🙏🏾

PR #3162, opened 2023-03-13
groupwithin new implementation

Angel-O:
# Update

New change log here: https://github.com/typelevel/fs2/pull/3162#issuecomment-1477075048

# Old change log

## Summary: groupWithin new implementation 

A new implementation for the combinator that will hopefully:
- be simpler to follow 
- improve performance *
- improve accuracy 

## Changes

- do away with `demand` semaphore: the process is entirely controlled by the the `supply` semaphore
- using `Chunk` instead of `Vector`

## Notes

- I’m leaving both implementations at the moment (old one renamed to `groupWithin1`) so you can compare benchmarks results between the two without switching branches
- I’m not familiar with `jmh` so by “performance improvements” I’m only looking at the result (ops/s). 
- I ran the tests a number of times without experiencing a single failure ~including the `accumulation and splitting` [test](https://github.com/typelevel/fs2/issues/2432) marked as flaky, hence I’m promoting the test to non-flaky status~ (I'm seeing the first failures on this test, less frequently compared to the current implementation, but they're still present, so I'm reverting this)
- I've attached the best result I've got from running the benchmark test (launching sbt with 4G, in both cases). I haven't seen those figures every time.

#### suggested implementation

Screenshot 2023-03-13 at 21 03 39

#### current implementation

Screenshot 2023-03-13 at 21 05 09


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1505,6 +1507,89 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+      chunkSize: Int,
+      timeout: FiniteDuration
+  )(implicit F: Temporal[F2]): Stream[F2, Chunk[O]] = {
+
+    case class State(@uncheckedVariance os: Chunk[O], supplyEnded: Boolean) {
+
+      val size: Long = os.size.toLong
+
+      // checking if it's empty to avoid early termination of the stream if the producer is faster than consumers
+      val streamExhausted: Boolean = supplyEnded && os.isEmpty
```
armanbilge, line 1520: I wonder if these can just be `def`s in which case I expect they will be inlined anyway.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1505,6 +1507,89 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+          val emitLast = state.get.map(Chunk.empty[O] ++ _.os)
+
+          val byTime = F.sleep(timeout) *> state.get.map(_.size)
+
+          val bySize = supply.acquireN(groupSize).as(groupSize)
+
+          def emitChunk(n: Long): F2[Chunk[O]] = state.modify(_.splitAt(n))
+
+          val dequeue: F2[Chunk[O]] =
+            F.race(bySize, byTime).flatMap {
```
armanbilge, line 1571: I'm worried if a race condition where both sides of this race complete can lead to deadlock. Actually, this concern of mine is what inspired this Cats Effect issue.
- https://github.com/typelevel/cats-effect/issues/3456

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1512,7 +1512,7 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       timeout: FiniteDuration
   )(implicit F: Temporal[F2]): Stream[F2, Chunk[O]] = {
 
-    case class State(@uncheckedVariance os: Chunk[O], supplyEnded: Boolean) {
+    case class State(os: Chunk[O @uncheckedVariance], supplyEnded: Boolean) {
```
armanbilge, line 1515: Instead of unchecked variance (which usually indicates possible unsoundness) can we just make it a type parameter?
```suggestion
    case class State[+O](os: Chunk[O], supplyEnded: Boolean) {
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1505,6 +1505,91 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+      Stream.force {
+        for {
+          state <- Ref.of(State[O](os = Chunk.empty, supplyEnded = false))
+          supply <- Semaphore[F2](0)
+        } yield {
+
+          val groupSize = chunkSize.toLong
+
+          val enqueue: F2[Unit] =
+            evalTap(o => state.update(State.add[O](o)) *> supply.release)
```
diesalbla, line 1546: Since you are not using the result after it. 

```suggestion
            foreach(o => state.update(State.add[O](o)) *> supply.release)
```

SystemFw:
Note that both this implementation and the one immediately preceding it have some important test scenarios not covered, e.g. whether errors from the input Stream are correctly propagated. This was coming from free for implementations that use `concurrently`

SystemFw (changes requested):
I'm not sure that removing the `demand` semaphore is valid, you basically lose all backpressure, if downstream starts pulling and then stops for a while, the producer will keep running and fill up the buffer.

Furthermore, this also removes all tracking of upstream errors, which isn't valid (and tbf we should add a test here, which is currently missing).

In other words, the additional complexity of the current implementation seems justified unless we can provide an alternative which is functionally equivalent :)

Angel-O:
@SystemFw yeah that make sense. I understand the role of the demand semaphore now: it wasn't clear to me immediately. I've hooked up VisualVm and realised memory performance is much worse 😓  

I have a question about error handling: what kind of errors can we expect when adding elements to the buffer ?

SystemFw:
> I have a question about error handling: what kind of errors can we expect when adding elements to the buffer ?

ah, that's not what I mean, I'm talking about propagation of upstream failures. Basically if you have 
```scala
val stream2 = stream1.groupWithin(100, 2.seconds)
```
I would expect stream2 to fail if stream1 does, but in this case stream1 becomes just a `.compile.drain.start`, which will not propagate errors to `output` (see how the current implementation tracks the result and does a bunch of work there).

Basically, every combinator has a set of (non-trivial) things it needs to do before we even talk about its specific logic: error propagation, backpressure, chunking, etc. This is why it's generally a good idea to use existing combinators when possible, e.g. `concurrently` instead of `start`. However this can come at a performance cost, and the existing `groupWithin` does a bunch of things manually to optimise.

(As a separate point, today I spoke with @armanbilge a bit about benchmarking the implementation in main not against its immediate predecessor based on `pull.timed`, but against the older implementation which sits in between in terms of high-level usage of combinators and low-level optimisation)

Angel-O:
@SystemFw @diesalbla @armanbilge thanks for the feedback/suggestions. I've made changes to take into account error propagation and memory utilization


### Change log (updated)

- removing ad-hoc logic around error propagation from producer to consumer stream. The logic is provided by the `concurrently` operator semantics in case of error and the supply semaphore in case of cancellation
- using a bounded queue for buffering and back-pressure
- (possibly) better use of supply semaphore (currently we release permits when the consumer fiber on the right succeed, just to re-acquire them shortly after. I might have not understood the reason for this, but perhaps that’s superfluous)
- (possibly) simpler timeout logic
- removes the empty check after the race since the consumer stream is guaranteed not to emit empty chunks 
- adding tests to ensure propagation of failure/cancellation (the existing implementation fails the cancellation test, while the new one passes it. Is that a valid cancellation test? @SystemFw)
- @armanbilge I’ve read the issue attached above (race condition), am I right to think that this implementation is potentially better if the race is `left-biased`, since the timeout logic (on the right) is slightly more convoluted ?
- other small simplifications (ref state encapsulates just a boolean flag, demand semaphore is gone, "short-circuiting" when `timeout == 0`)

After these changes performance is still better (ops/sec up to 100% better especially for large streams, I’m not familiar with `gc` stats, but visual vm graphs look ok to me, better than the first attempt)

I've run the tests a few times, without failures (except for the flaky "accumulation & splitting"): hopefully you will get the same results.

Lastly I’ve followed the discussion on Discord and I’m aware that the long term plan is to move away from semaphores and rely on fs2 data types. Will start looking at Pull et al in the next weeks: in the meantime hopefully these changes are a small step forward. Let me know what you guys think and if there’s anything that still needs fixing.

Screenshot 2023-03-20 at 22 59 39

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1505,6 +1505,72 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+            flushed <- emitChunk
+            // lower supply by {flushed.size} (excluding element already awaited)
+            _ <- supply.tryAcquireN((flushed.size.toLong - awaited).max(0))
+          } yield flushed
+
+          val onTimeout =
+            F.ifM(streamExhausted)(F.pure(Chunk.empty[O]), awaitAndEmitNext)
+
+          val dequeue: F2[Chunk[O]] =
+            F.race(supply.acquireN(groupSize), F.sleep(timeout)).flatMap {
```
armanbilge, line 1561: > I’ve read the issue attached above (race condition), am I right to think that this implementation is potentially better if the race is left-biased, since the timeout logic (on the right) is slightly more convoluted ?

Less chance of deadlock is still not as good as no chance of deadlock 😁 

I think the currently implementation uses a hack, where it always returns the semaphore, and then attempts to re-acquire it. But there's no way to fix this properly without implementing the `Ior`-based `race` as described in that issue.

Angel-O: I was struggling a bit to understand the problem (i.e where the deadlock would occur) until I clicked through a few links from the issue linked above and stumbled across your PR comment [here](https://github.com/typelevel/cats-effect/pull/3233#issue-1439117768)

If I understood your concern correctly the issue is that the semaphore might lose the race, even if it acquired the permits successfully. So in that case we would need to restore them, in order to prevent a deadlock when the `onTimeout` logic tries to acquire a permit again (that simply won't be there if the semaphore acquired them without releasing them)

if that is the case I believe this is not a problem in this implementation because of this logic:

```scala
     val isBufferEmpty: F2[Boolean] =
        buffer.size.map(_ == 0)

      val awaitAndEmitNext: F2[Chunk[O]] = for {
        isEmpty <- isBufferEmpty
        awaited <- supply.acquire.whenA(isEmpty).as(if (isEmpty) 1 else 0)
        flushed <- emitChunk
        // lower supply by {flushed.size} (excluding element already awaited)
        _ <- supply.tryAcquireN((flushed.size.toLong - awaited).max(0))
      } yield flushed
``` 

basically if the buffer contains any element by the time the timeout expires we won't attempt to acquire the permit in a blocking fashion in the for-comprehension. 
So by definition the deadlock cannot occur or in other words we should be able to rule out the deadlock because if the buffer is empty then we will know for a fact that the competing semaphore (the one in the race) hasn't acquired any permit (i.e. it does not need to release them)

armanbilge: Thanks for looking more closely at this! Yes, you've definitely understood the issue.

After looking at your code and thinking about it, I _think_ you are right that it won't deadlock. However, I still suspect that the state can get corrupted. Something like this:

1. The `race` times-out, although the semaphore `acquireN`  also succeeds.
https://github.com/typelevel/fs2/blob/d676eddbce407908b7a6daaa00666b88b968cd0c/core/shared/src/main/scala/fs2/Stream.scala#L1456-L1458
2. We successfully `emitChunk` via `onTimeout` and `awaitAndEmitNext`.
https://github.com/typelevel/fs2/blob/d676eddbce407908b7a6daaa00666b88b968cd0c/core/shared/src/main/scala/fs2/Stream.scala#L1452-L1453
https://github.com/typelevel/fs2/blob/d676eddbce407908b7a6daaa00666b88b968cd0c/core/shared/src/main/scala/fs2/Stream.scala#L1444-L1447
https://github.com/typelevel/fs2/blob/d676eddbce407908b7a6daaa00666b88b968cd0c/core/shared/src/main/scala/fs2/Stream.scala#L1420-L1421
3. This unblocks the `buffer`, which concurrently gets filled with new stuff and releases permits.
https://github.com/typelevel/fs2/blob/d676eddbce407908b7a6daaa00666b88b968cd0c/core/shared/src/main/scala/fs2/Stream.scala#L1441-L1442
4. The `tryAcquireN` in `awaitAndEmitNext` succeeds.
https://github.com/typelevel/fs2/blob/d676eddbce407908b7a6daaa00666b88b968cd0c/core/shared/src/main/scala/fs2/Stream.scala#L1448-L1450
5. At this point, we've consumed `2*chunkSize` permits to emit `chunkSize` elements.

Angel-O:
Closing this, I've discovered a bug that explains the huge performance gain (basically the stream terminates prematurely: not sure why): I'll raise another PR instead adding tests

core/shared/src/main/scala/fs2/Stream.scala
Angel-O: Nice!! Using `TestControl.executeEmbed` is a great idea for this! Will give it a go 👍🏾

PR #3165, opened 2023-03-16
Refactor JVM unixsockets

armanbilge:
I'm working on supporting HTTP/2 over unix sockets in Ember and was experiencing a lot of strange issues e.g. non-deterministic behavior, reads returning empty chunks, hanging.

The bad news was that I was unable to replicate these issues as unit tests 🤔 this is where I started from, focused on just the empty read issue, before I gave up and went for the refactor.

The good news is that I published a snapshot and (spoiler alert) was surprised to discover that this refactor somehow fixed everything 😅 wasn't really expecting that.

This refactor mostly addresses these three issues:

1. Not switching the sockets to non-blocking mode. I suspect that this was what was causing the empty reads bug in Ember. Non-blocking mode is only useful if you are using a selector to poll for events on the socket. Otherwise, you have no way of knowing when there _are_ bytes available on the socket, so you may get empty reads / no-op writes, and be forced to degrade to a busy-loop, which is much worse than doing a blocking read (or write) to begin with.

    Note that when the Cats Effect polling system stuff lands, we will be able to register the JDK unix sockets with the polling system and thus have non-blocking reads/writes, so that's something to look forward to.

2. Carefully wrapping a socket in a resource as soon as it is opened, so that it is guaranteed to be closed in all instances. Attempting to tack on other side-effects as part of the `acquire` phase (e.g., binding or connecting) may lead to a resource leak if those actions fail, because then the socket is never returned, which means it can never be closed, etc.

3. Properly suspending side-effects e.g. interactions with `ByteBuffers`, `blocking`, etc.

### DIFF
--- a/io/jvm/src/main/scala/fs2/io/net/unixsocket/JdkUnixSockets.scala
+++ b/io/jvm/src/main/scala/fs2/io/net/unixsocket/JdkUnixSockets.scala
@@ -21,7 +21,8 @@
 
 package fs2.io.net.unixsocket
 
-import cats.effect.kernel.Async
+import cats.effect.kernel.{Async, Resource}
+import cats.effect.syntax.all._
 import java.net.{StandardProtocolFamily, UnixDomainSocketAddress}
 import java.nio.channels.{ServerSocketChannel, SocketChannel}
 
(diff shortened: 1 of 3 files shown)


io/jvm/src/main/scala/fs2/io/net/unixsocket/JdkUnixSockets.scala
```diff
@@ -35,16 +36,29 @@ object JdkUnixSockets {
...
-    ch
-  }
-
-  protected def openServerChannel(address: UnixSocketAddress) = F.blocking {
-    val serverChannel = ServerSocketChannel.open(StandardProtocolFamily.UNIX)
-    serverChannel.configureBlocking(false)
-    serverChannel.bind(UnixDomainSocketAddress.of(address.path))
-    (F.blocking(serverChannel.accept()), F.blocking(serverChannel.close()))
-  }
+  protected def openChannel(address: UnixSocketAddress) =
```
diesalbla, line 39: Could we add return types?

diesalbla (approved):

PR #3167, opened 2023-03-19
Added writeUtf8 and writeUtf8Lines

TonioGela:
Hello!
That's my first contribution to fs2, so I hope everything is where it should be.
I added two helpers to `Files[F]` that mirror the ones already present for `readUtf8` and `readUtf8Lines`, shamelessly stealing the idea from [this](https://github.com/typelevel/toolkit/issues/15#issuecomment-1464156091) @armanbilge comment.
I haven't found any unit tests for `readUtf8` and `readUtf8Lines` but for their components, and since `writeUtf8` and `writeUtf8Lines` are built on the same ones, I haven't added any.

It may be worth adding these new combinators to fs2 2.5.x and adding an example like `Stream.emit("foo").through(Files[IO].writeUtf8Lines(Path(...)))` to the documentation.

### DIFF
--- a/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
@@ -264,6 +264,8 @@ private[fs2] trait FilesCompanionPlatform {
     override def isSameFile(path1: Path, path2: Path): F[Boolean] =
       F.pure(path1.absolute == path2.absolute)
 
+    override val lineSeparator: String = facade.os.EOL()
+
     override def list(path: Path): Stream[F, Path] =
       Stream
         .bracket(F.fromPromise(F.delay(facade.fs.promises.opendir(path.toString))))(dir =>
--- a/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
@@ -264,7 +264,7 @@ private[fs2] trait FilesCompanionPlatform {
     override def isSameFile(path1: Path, path2: Path): F[Boolean] =
       F.pure(path1.absolute == path2.absolute)
 
-    override val lineSeparator: String = facade.os.EOL
+    override def lineSeparator: String = facade.os.EOL
 
     override def list(path: Path): Stream[F, Path] =
       Stream
--- a/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
@@ -264,7 +264,7 @@ private[fs2] trait FilesCompanionPlatform {
     override def isSameFile(path1: Path, path2: Path): F[Boolean] =
       F.pure(path1.absolute == path2.absolute)
 
-    override val lineSeparator: String = facade.os.EOL()
+    override val lineSeparator: String = facade.os.EOL
 
     override def list(path: Path): Stream[F, Path] =
       Stream
--- a/io/js/src/main/scala/fs2/io/internal/facade/os.scala
+++ b/io/js/src/main/scala/fs2/io/internal/facade/os.scala
@@ -42,6 +42,10 @@ package object os {
   @JSImport("os", "networkInterfaces")
   private[io] def networkInterfaces(): js.Dictionary[js.Array[NetworkInterfaceInfo]] = js.native
 
+  @js.native
+  @JSImport("os", "EOL")
+  private[io] def EOL(): String = js.native
+
 }
 
 package os {
--- a/io/js/src/main/scala/fs2/io/internal/facade/os.scala
+++ b/io/js/src/main/scala/fs2/io/internal/facade/os.scala
@@ -44,7 +44,7 @@ package object os {
 
   @js.native
   @JSImport("os", "EOL")
-  private[io] def EOL(): String = js.native
+  private[io] def EOL: String = js.native
 
 }
 
--- a/io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
@@ -291,6 +291,8 @@ private[file] trait FilesCompanionPlatform {
     def isSameFile(path1: Path, path2: Path): F[Boolean] =
       Sync[F].blocking(JFiles.isSameFile(path1.toNioPath, path2.toNioPath))
 
+    val lineSeparator: String = System.lineSeparator()
+
     def list(path: Path): Stream[F, Path] =
       _runJavaCollectionResource[JStream[JPath]](
         Sync[F].blocking(JFiles.list(path.toNioPath)),
--- a/io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
@@ -291,7 +291,7 @@ private[file] trait FilesCompanionPlatform {
     def isSameFile(path1: Path, path2: Path): F[Boolean] =
       Sync[F].blocking(JFiles.isSameFile(path1.toNioPath, path2.toNioPath))
 
-    val lineSeparator: String = System.lineSeparator()
+    def lineSeparator: String = System.lineSeparator()
 
     def list(path: Path): Stream[F, Path] =
       _runJavaCollectionResource[JStream[JPath]](
--- a/io/shared/src/main/scala/fs2/io/file/Files.scala
+++ b/io/shared/src/main/scala/fs2/io/file/Files.scala
@@ -241,7 +241,7 @@ sealed trait Files[F[_]] extends FilesPlatform[F] {
   def isSameFile(path1: Path, path2: Path): F[Boolean]
 
   /** Returns the line separator for the specific OS */
-  val lineSeparator: String
+  def lineSeparator: String
 
   /** Gets the contents of the specified directory. */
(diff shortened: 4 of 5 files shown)


armanbilge:
> It may be worth adding these new combinators to fs2 2.5.x

FS2 2.5.x is EOL (like CE 2.x series). I can't remember when we last merged a PR there :)

> adding an example like `Stream.emit("foo").through(Files[IO].writeUtf8Lines(Path(...)))` to the documentation.

Sure! There may already be an existing example that can be made less verbose with the new helper :)

io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -573,6 +599,12 @@ object Files extends FilesCompanionPlatform {
             }
           }
     }
+
+    def writeUtf8(path: Path, flags: Flags): Pipe[F, String, Nothing] =
+      _.through(text.utf8.encode).through(writeAll(path, flags))
+
+    def writeUtf8Lines(path: Path, flags: Flags): Pipe[F, String, Nothing] =
+      _.through(text.lines).through(writeUtf8(path, flags))
```
armanbilge, line 607: Out of curiosity can these implementations move into the parent trait, do they require any constraints?

TonioGela: I must admit that I blindly copied the method used for readUtf8 and its sibling. Also, WDYM with constraint here?

TonioGela: > do they require any constraints?
I got it, the `Async`. And no, they don't require it and neither do readUtf8, readUtf8Lines and readCursor apparently. I can move the newly added methods to the trait, what about readUtf8 and readUtf8Lines?

armanbilge: Great, thanks! Sure, I say move them up.

armanbilge: Btw, I'm not sure this does what we want. Doesn't `text.lines` split a stream of `String` into lines, not the other way around? e.g. add `\n` between `String`s?

Might be good to add a test 😇

TonioGela: You're probably right, I'll add a few

io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -409,6 +412,34 @@ sealed trait Files[F[_]] extends FilesPlatform[F] {
...
+    * Use `writeUtf8Lines(path, Flags.Append)` to append to the end
+    * of the file, or pass other flags to further customize behavior.
+    */
+  def writeUtf8Lines(path: Path): Pipe[F, String, Nothing] = writeUtf8(path, Flags.Write)
+
+  /** Writes each string to the specified file as utf8 lines
+    * using the specified flags to open the file.
+    */
+  def writeUtf8Lines(path: Path, flags: Flags): Pipe[F, String, Nothing] = in =>
+    in.through(text.lines).through(writeUtf8(path, flags))
```
diesalbla, line 442: As @armanbilge mentioned, if the goal here is to insert a newline between strings, that would be `in => writeUtf8(path, flags)(text.intersperse("\n")`

armanbilge: Not sure, does that add a newline at the end? Seems like each string should terminate like that, so that ultimately the file terminates in a newline.

Also, does `text.lines` strip newlines? The reason I ask is because ideally these methods would be inverses of each other.

TonioGela: Yep, I'm adding it in a new commit with a couple of unit tests too :)

TonioGela: I cried for almost 1 hour in front of failing unit tests and then I saw it:

```scala
def writeUtf8Lines(path: Path): Pipe[F, String, Nothing] = writeUtf8(path, Flags.Write)
                                                                    ^
```

For what is worth, I think that the logic might take in consideration being on windows probably: `text.lines` implementation takes in consideration `\r` too.

> Not sure, does that add a newline at the end? Seems like each string should terminate like that, so that ultimately the file terminates in a newline.

Can we make it pluggable with a parameter maybe?

> Also, does text.lines strip newlines? The reason I ask is because ideally these methods would be inverses of each other.

It does

TonioGela: > Also, does text.lines strip newlines? The reason I ask is because ideally these methods would be inverses of each other.

Now that I'm thinking about it, is it the case to add roundtrip property based test like 👇 ?

```scala
forAll { xs: List[String] =>
  Stream.emits(xs).through(Files[IO].writeUtf8Lines(path)) ++
      Files[IO].readUtf8Lines(path).compile.toList.assertEquals(xs)
}
```

io/js/src/main/scala/fs2/io/internal/facade/os.scala
```diff
@@ -42,6 +42,10 @@ package object os {
   @JSImport("os", "networkInterfaces")
   private[io] def networkInterfaces(): js.Dictionary[js.Array[NetworkInterfaceInfo]] = js.native
 
+  @js.native
+  @JSImport("os", "EOL")
+  private[io] def EOL(): String = js.native
```
armanbilge, line 47: No parens, since that's now how it's defined in Node.js :)

https://nodejs.org/api/os.html#oseol

```suggestion
  private[io] def EOL: String = js.native
```

io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
```diff
@@ -291,6 +291,8 @@ private[file] trait FilesCompanionPlatform {
     def isSameFile(path1: Path, path2: Path): F[Boolean] =
       Sync[F].blocking(JFiles.isSameFile(path1.toNioPath, path2.toNioPath))
 
+    val lineSeparator: String = System.lineSeparator()
```
armanbilge, line 294: Nit, but I'd use a `def` instead of a `val` since `lineSeparator()` is also a method call.

io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
```diff
@@ -264,6 +264,8 @@ private[fs2] trait FilesCompanionPlatform {
     override def isSameFile(path1: Path, path2: Path): F[Boolean] =
       F.pure(path1.absolute == path2.absolute)
 
+    override val lineSeparator: String = facade.os.EOL()
```
armanbilge: Same here.

io/shared/src/main/scala/fs2/io/file/Files.scala
TonioGela: > For what is worth, I think that the logic might take in consideration being on windows probably: `text.lines` implementation takes in consideration `\r` too.

Now it's handled

io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -409,6 +415,34 @@ sealed trait Files[F[_]] extends FilesPlatform[F] {
...
+    * Use `writeUtf8Lines(path, Flags.Append)` to append to the end
+    * of the file, or pass other flags to further customize behavior.
+    */
+  def writeUtf8Lines(path: Path): Pipe[F, String, Nothing] = writeUtf8Lines(path, Flags.Write)
+
+  /** Writes each string to the specified file as utf8 lines
+    * using the specified flags to open the file.
+    */
+  def writeUtf8Lines(path: Path, flags: Flags): Pipe[F, String, Nothing] = in =>
+    in.intersperse(lineSeparator).through(writeUtf8(path, flags))
```
armanbilge, line 445: I really feel like we should be appending a line-separator to each string, instead of interspersing.

> A sequence of zero or more non- <newline> characters **plus a terminating <newline> character.**

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

TonioGela: `in.flatMap(s => Stream(s, lineSeparator))` would do?

armanbilge: Yes, exactly!

io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -409,6 +415,34 @@ sealed trait Files[F[_]] extends FilesPlatform[F] {
...
+    * Use `writeUtf8(path, Flags.Append)` to append to the end of
+    * the file, or pass other flags to further customize behavior.
+    */
+  def writeUtf8(path: Path): Pipe[F, String, Nothing] = writeUtf8(path, Flags.Write)
+
+  /** Writes to the specified file as an utf8 string using
+    * the specified flags to open the file.
+    */
+  def writeUtf8(path: Path, flags: Flags): Pipe[F, String, Nothing] = in =>
+    in.through(text.utf8.encode).through(writeAll(path, flags))
```
armanbilge, line 431: Should we add this? I'm not entirely sure how `Write` and `Append` flags interact 🤔 

```suggestion
    in.through(text.utf8.encode).through(writeAll(path, flags.addIfAbsent(Flag.Write)))
```

armanbilge: Wait, nevermind, we don't need this. It's already added further down the call-chain :)

armanbilge (approved):
Nice, thanks!

PR #3178, opened 2023-03-23
Fix NPE for macOS unix sockets

armanbilge:
This feels like a Node.js bug.


armanbilge:
Aha, this is not specific to macOS. This is a Node.js 19 issue, which is what brew installs by default. I just replicated on Ubuntu.

armanbilge:
- https://github.com/nodejs/node/issues/47229

PR #3179, opened 2023-03-23
Implicit instances for `IO` only

armanbilge:
Discord thread: https://discord.com/channels/632277896739946517/839257183782436945/1087761519477014644

Continuing from https://github.com/Comcast/ip4s/pull/475.

**tl;dr**: The motivation for removing the implicit implementation for generic `F[_]` is to force tagless final libs to declare _all_ of their constraints (e.g. `Network`) instead of deriving them on-the-spot from e.g. `Async`. Meanwhile, this change will be completely transparent to code written in concrete `IO`.

This is source-breaking but binary-compatible.

Changes were made to:
- `Compression`
- `Network`
- `UnixSockets`
- `Files`, which impacted internal code

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -188,7 +188,8 @@ ThisBuild / mimaBinaryIssueFilters ++= Seq(
   ),
   ProblemFilters.exclude[DirectMissingMethodProblem](
     "fs2.io.net.DatagramSocketGroupCompanionPlatform#AsyncDatagramSocketGroup.this"
-  )
+  ),
+  ProblemFilters.exclude[DirectMissingMethodProblem]("fs2.io.file.Watcher#DefaultWatcher.this")
 )
 
 lazy val root = tlCrossRootProject
--- a/core/jvm-native/src/main/scala/fs2/compression/CompressionPlatform.scala
+++ b/core/jvm-native/src/main/scala/fs2/compression/CompressionPlatform.scala
@@ -118,7 +118,13 @@ private[compression] trait CompressionPlatform[F[_]] { self: Compression[F] =>
 
 }
 
-private[compression] trait CompressionCompanionPlatform {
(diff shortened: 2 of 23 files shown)


rossabaker:
I appreciate and agree with the motivation for this, but it also looks like it might unleash a metric fuckton of busywork.  If I'm reading it right, it's not even going to trigger deprecation warnings, just hard errors.  I am willing to try a snapshot on some apps and see if I'm overestimating it.

I wonder whether a scalafix is possible here?

armanbilge:
Yes, it did create quite a bit of busywork just within FS2.

Not sure about a Scalafix, but an escape hatch would be to add something like `import fs2.all.the.old.implicits._`

armanbilge:
Oh, I was thinking too ambitiously about a Scalafix :)

So the _real_ fix is to lift these constraints up. That seems hard to scalafix.

An easy scalafix workaround is to rewrite `Network[F]` to `Network.forAsync[F]` and `Files[F]` to `Files.forAsync[F]`.

That works fine for direct calls to those typeclasses, but it won't help where they are being passed around implicitly.

rossabaker:
This change strikes me as too aggressive in a minor version of a fundamental library.  I wish they were this way from the beginning, but I dread the fallout in a large org.  As a compromise, could we retain deprecated implicits?  Something so users quickly get a hint, and can proceed with a warning until there's time to do it right?

armanbilge:
> This change strikes me as too aggressive in a minor version of a fundamental library.

I don't disagree!

> As a compromise, could we retain deprecated implicits?

On second thought, yes I think we can. I think if we use the priority tricks so the non-deprecated `IO` implicit takes precedence over the deprecated `Async` implicit, that would avoid the ambiguous implicit trap.

PR #3183, opened 2023-03-25
Fixing interruption behaviour

### LINKED ISSUE #3169: groupWithin: inconsistent behaviour on source termination
### [scastie](https://scastie.scala-lang.org/HAeMqP3BTXSUBGuf2eqPMg)

While playing around with a new implementation of groupWithin I discovered a bug. The combinator doesn't handle upstream termination consistently. See below

```scala
    val finiteStream = Stream
      .range(0, 4)
      .covary[IO]
      .metered(1.second)

    val doesNotTerminate = 
        finiteStream
          .groupWithin(Int.MaxValue, 2.days) 
          .evalTap(IO.println)
          .compile
          .lastOrError.void

    val terminates =
         finiteStream
          .groupWithin(8, 2.days)
          .evalTap(IO.println)
          .compile
          .lastOrError.void
```

I have an idea why this is happening. I'm happy to take a look 💪🏾 

The question I have though is: what's the correct behaviour here ?
 - terminating the stream when upstream terminates 
 - keeping the stream alive

Angel-O:
## Summary

- Fixes #3169
- Adding missing tests around error/interruption propagation
- Adding stress/integrity test

## Changes

increasing supply by `Int.MaxValue + chunkSize` on upstream `finalization` to prevent the supply semaphore from blocking when upstream is interrupted while downstream is waiting for the timeout to expire or to have enough elements

## Notes

- the `stress test: all elements are processed` test is nothing but a copy of the benchmark with and integrity check at the end. While it may seem redundant (i.e. `should never lose any elements` should cover it), in reality it is possible to write an [implementation](https://github.com/typelevel/fs2/pull/3162) that passes that test but fails this one. (~on a side note, that explains the big performance gain of the implementation linked earlier, so~ having this test in place could prevent being misled when looking at benchmark results, while working on a new implementation 🙏🏾 )

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1471,7 +1471,7 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
         }
 
         def endSupply(result: Either[Throwable, Unit]): F2[Unit] =
-          buffer.update(_.copy(endOfSupply = Some(result))) *> supply.releaseN(Int.MaxValue)
+          buffer.update(_.copy(endOfSupply = Some(result))) *> supply.releaseN(Int.MaxValue + outputLong)
 
         def endDemand(result: Either[Throwable, Unit]): F2[Unit] =
           buffer.update(_.copy(endOfDemand = Some(result))) *> demand.releaseN(Int.MaxValue)
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1472,7 +1472,9 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
 
         def endSupply(result: Either[Throwable, Unit]): F2[Unit] =
           buffer.update(_.copy(endOfSupply = Some(result))) *> supply.releaseN(
-            Int.MaxValue + outputLong
+            // enough supply for 2 iterations of the race loop in case of upstream
+            // interruption: so that downstream can terminate immediately
+            outputLong * 2
           )
 
         def endDemand(result: Either[Throwable, Unit]): F2[Unit] =
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1471,7 +1471,9 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
         }
 
         def endSupply(result: Either[Throwable, Unit]): F2[Unit] =
-          buffer.update(_.copy(endOfSupply = Some(result))) *> supply.releaseN(Int.MaxValue + outputLong)
+          buffer.update(_.copy(endOfSupply = Some(result))) *> supply.releaseN(
+            Int.MaxValue + outputLong
+          )
 
         def endDemand(result: Either[Throwable, Unit]): F2[Unit] =
           buffer.update(_.copy(endOfDemand = Some(result))) *> demand.releaseN(Int.MaxValue)
--- a/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
+++ b/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
@@ -859,28 +859,24 @@ class StreamCombinatorsSuite extends Fs2Suite {
       val downstreamTimeout = sourceTimeout + 2.seconds
 
       TestControl
-        .executeEmbed(
-          Ref[IO]
-            .of(0.millis)
-            .flatMap { ref =>
-              val source: Stream[IO, Int] =
-                Stream
-                  .iterate(0)(_ + 1)
-                  .covary[IO]
-                  .meteredStartImmediately(1.second)
-                  .interruptAfter(sourceTimeout)
+        .executeEmbed {
+          val source: Stream[IO, Int] =
+            Stream
+              .iterate(0)(_ + 1)
+              .covary[IO]
+              .meteredStartImmediately(1.second)
+              .interruptAfter(sourceTimeout)
 
-              // large chunkSize and timeout (no emissions expected in the window
-              // specified, unless source ends, due to interruption or
-              // natural termination (i.e runs out of elements)
-              val downstream: Stream[IO, Chunk[Int]] =
-                source.groupWithin(Int.MaxValue, 1.day)
+          // large chunkSize and timeout (no emissions expected in the window
+          // specified, unless source ends, due to interruption or
+          // natural termination (i.e runs out of elements)
+          val downstream: Stream[IO, Chunk[Int]] =
+            source.groupWithin(Int.MaxValue, 1.day)
 
-              downstream.compile.lastOrError
-                .map(_.toList)
-                .timed
-            }
-        )
+          downstream.compile.lastOrError
+            .map(_.toList)
+            .timed
+        }
         .assertEquals(
           // downstream ended immediately (i.e timeLapsed = sourceTimeout)
           // emitting whatever was accumulated at the time of interruption
--- a/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
+++ b/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
@@ -837,17 +837,18 @@ class StreamCombinatorsSuite extends Fs2Suite {
     }
 
     test("upstream failures are propagated downstream") {
+      TestControl.executeEmbed {
+        case object SevenNotAllowed extends NoStackTrace
 
-      case object SevenNotAllowed extends NoStackTrace
+        val source = Stream
+          .iterate(0)(_ + 1)
+          .covary[IO]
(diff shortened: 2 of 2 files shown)


armanbilge:
How does this PR relate to https://github.com/typelevel/fs2/pull/3186? Should this be reviewed first?

Angel-O:
I ended up making all the changes in #3186, since I wanted to verify the correctness of that implementation, but I'm happy for this one to be considered first. I'll sync up the branch, run `sbt prePR` and push again if needed.

core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
```diff
@@ -834,6 +835,77 @@ class StreamCombinatorsSuite extends Fs2Suite {
...
+        case object SevenNotAllowed extends NoStackTrace
+
+        val source = Stream
+          .iterate(0)(_ + 1)
+          .covary[IO]
+          .evalTap(n => IO.raiseError(SevenNotAllowed).whenA(n == 7))
+
+        val downstream = source.groupWithin(100, 2.seconds)
+
+        downstream.intercept[SevenNotAllowed.type]
```
armanbilge, line 850: Should we make an assertion here about what the downstream has / has not received before the error?

core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
```diff
@@ -834,6 +835,77 @@ class StreamCombinatorsSuite extends Fs2Suite {
...
+              // specified, unless source ends, due to interruption or
+              // natural termination (i.e runs out of elements)
+              val downstream: Stream[IO, Chunk[Int]] =
+                source.groupWithin(Int.MaxValue, 1.day)
+
+              downstream.compile.lastOrError
+                .map(_.toList)
+                .timeout(downstreamTimeout)
+                .flatTap(_ => IO.monotonic.flatMap(ref.set))
+                .flatMap(emit => ref.get.map(timeLapsed => (timeLapsed, emit)))
```
armanbilge, line 883:
```suggestion
                .timed
```

core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
```diff
@@ -874,6 +874,7 @@ class StreamCombinatorsSuite extends Fs2Suite {
             source.groupWithin(Int.MaxValue, 1.day)
 
           downstream.compile.lastOrError
+            .timeout(downstreamTimeout)
```
armanbilge, line 877: Why is this timeout necessary? Since its an `executeEmbed` test.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1471,7 +1471,9 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
         }
 
         def endSupply(result: Either[Throwable, Unit]): F2[Unit] =
-          buffer.update(_.copy(endOfSupply = Some(result))) *> supply.releaseN(Int.MaxValue)
+          buffer.update(_.copy(endOfSupply = Some(result))) *> supply.releaseN(
+            Int.MaxValue + outputLong
+          )
```
armanbilge, line 1476: Sorry, dumb question: why is `Int.MaxValue` a "magic number" in this context? I would have thought it's effectively maxing out the semaphore, but if it needs `+ outputLong` to work then I feel like it must have more significance?

Angel-O: Legit question to be fair. Had to think about it again.

Interruption of the upstream fiber (i.e. Outcome.Cancelled) is handled  downstream by doing nothing (permits are never released)
So by increasing the supply to `Int.MaxValue` we are just evening out the negative balance (`Int.MaxValue` is to account for the worst case scenario: at most the `chunkSize` parameter will be equal to `Int.MaxValue`)

```scala
    val waitSupply = supply.acquireN(outputLong).guaranteeCase {
        case Outcome.Succeeded(_) => supply.releaseN(outputLong)
        case _                    => F.unit
     }
```

Now after getting past the "checkpoint"  above we are acquiring `outputLong` permits again

```scala
    acq <- F.race(F.sleep(timeout), waitSupply).flatMap {
      case Left(_)  => onTimeout
      case Right(_) => supply.acquireN(outputLong).as(outputLong)
    }
```

So in order to get past this point we need to release an additional `outputLong` permits and that allows the stream to be unblocked

armanbilge: Thanks for that explanation!

> `Int.MaxValue` is to account for the worst case scenario: at most the `chunkSize` parameter will be equal to `Int.MaxValue`

So could we just use `chunkSize` here, instead of `Int.MaxValue` ?

core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
Angel-O: it's because if the test fails we get a slightly better error message

```java.util.concurrent.TimeoutException: 7500 milliseconds``` which can be easily associated to `downstreamTimeout`

otherwise we get this value on the diff which looks a bit random

```_1 = 86405500000000 nanoseconds,```

but I'm happy to remove it

armanbilge: Got it, that is a nicer error :) thanks!

core/shared/src/main/scala/fs2/Stream.scala
Angel-O: @armanbilge apologies I was wrong, that's not what's happening here. I'm just doing some tests to figure out why we need the additional `outputLong`

armanbilge: Btw, if these implementations details are no longer relevant after your rewrite in the other PR, then let's not get too hung up on this one :)

Angel-O: ok I think I've figured it out (might be useful for the other implementation actually)

basically the problem is that we need enough supply to cover 2 iterations of the race loop. So if we only increase it by `Int.MaxValue` the following will happen

- (current iteration): supply is unblocked
- (next iteration): supply gets stuck (not enough supply because upstream was interrupted)

in instead we increase it by `Int.MaxValue + outputLong`

- (current iteration): supply is unblocked
- (next iteration): supply is not blocked thanks to the additional `outputLong`

So since the chunkSize can be as high as `Int.MaxValue` then the minimum supply to unblock the semaphore should be `Int.MaxValue + outputLong`

armanbilge: > So since the chunkSize can be as high as `Int.MaxValue` then the minimum supply to unblock the semaphore should be `Int.MaxValue + outputLong`

Key word being "can". Wouldn't `chunkSize + outputLong` be sufficient?

He-Pin:
Have you stress the latency of many fs streams with groupWithin？

core/shared/src/main/scala/fs2/Stream.scala
Angel-O: yeah that should work. The test still passes, I'll change it to `outputLong * 2` since chunkSize == outputLong

Angel-O:
> Have you stress the latency of many fs streams with groupWithin？

Hey @He-Pin , may I ask ? what do you mean exactly? Running many streams concurrently and using `groupWithin` on the resulting stream ? i.e. after calling `.parJoinUnbounded` ?

He-Pin:
> > Have you stress the latency of many fs streams with groupWithin？
> 
> Hey @He-Pin , may I ask ? what do you mean exactly? Running many streams concurrently and using `groupWithin` on the resulting stream ? i.e. after calling `.parJoinUnbounded` ?

Eg, usging groupwithin to build a lock-step game server where the latency matters

Angel-O:
> Eg, using groupwithin to build a lock-step game server where the latency matters

Thanks for the clarification. I must admit that unfortunately I've never heard of a lock-step server, so I haven't done any work (or rather wrote any test) with this particular use case in mind. Also keep in mind that I'm fairly new to open source contributions so maybe this is  more of a question for regular maintainers. 

Nonetheless, I'd be interested to find out more: can you point me towards an example or a beginner friendly resource? Thank you 🙏🏾

armanbilge:
@He-Pin you are welcome to contribute some benchmarks for your usecase, that will probably be the best way to answer your question :)

He-Pin:
> > Eg, using groupwithin to build a lock-step game server where the latency matters
> 
> Thanks for the clarification. I must admit that unfortunately I've never heard of a lock-step server, so I haven't done any work (or rather wrote any test) with this particular use case in mind. Also keep in mind that I'm fairly new to open source contributions so maybe this is  more of a question for regular maintainers. 
> 
> Nonetheless, I'd be interested to find out more: can you point me towards an example or a beginner friendly resource? Thank you 🙏🏾 

keep performance in mind, this is the way,mandalorian.

He-Pin:
> @He-Pin you are welcome to contribute some benchmarks for your usecase, that will probably be the best way to answer your question :)

that's true, for my usecase i need to use a Hashed timer instead.

I'm learning more CE code too, will pr later.

armanbilge (approved):
Thanks for chasing this one down! The new tests specifying the expected behavior are great 👍

PR #3185, opened 2023-03-25
fixing assertion in parEval test

Angel-O:
## Summary

According to the test spec the result may not be in the correct order, but the assertion check for exact equality. 

## Changes
- adjusting `parEvalMapUnordered` test to reflect spec, by using assertion that does not check for ordering
- removing no longer used `run` helper method

### DIFF
--- a/core/shared/src/test/scala/fs2/ParEvalMapSuite.scala
+++ b/core/shared/src/test/scala/fs2/ParEvalMapSuite.scala
@@ -66,13 +66,17 @@ class ParEvalMapSuite extends Fs2Suite {
     }
 
     test("may not be preserved in parEvalMapUnordered") {
+      run(_.parEvalMapUnorderedUnbounded(identity)).assertEquals(List(1, 2, 3))
+    }
+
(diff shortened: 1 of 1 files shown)


core/shared/src/test/scala/fs2/ParEvalMapSuite.scala
```diff
@@ -66,17 +66,13 @@ class ParEvalMapSuite extends Fs2Suite {
     }
 
     test("may not be preserved in parEvalMapUnordered") {
-      run(_.parEvalMapUnorderedUnbounded(identity)).assertEquals(List(1, 2, 3))
-    }
-
-    def run(pipe: Pipe[IO, IO[Int], Int]) =
       Stream
         .emits(List(3, 2, 1))
         .map(i => IO.sleep(50.millis * i.toLong).as(i))
```
armanbilge, line 71: Although, based on the test, it looks like a very specific order is expected :)

Instead we should rewrite this to use `executeEmbed` so it won't flake.

armanbilge (commented):
Thanks for tracking these down!

armanbilge (approved):

PR #3186, opened 2023-03-25
Groupwithin improvements

### LINKED ISSUE #3169: groupWithin: inconsistent behaviour on source termination

Angel-O:
## Summary

This PR is a followup from #3162 (thanks all for the feedback), and a possible alternative to that approach in terms of timeout behaviour

### Goals:
- improving timeout behaviour (NEW)
- fixing `interruption` propagation misbehaviour (PREVIOUS PR)
- improving performance (PREVIOUS PR)
- simplifying logic (PREVIOUS PR)

### Timeout behaviour

The current implementation has a small flaw in the timeout logic: when entering the timeout state the stream waits for the first element then tries to calculate how many elements are available using the supply semaphore. It finally uses that figure to decide where to split the buffer.

The problem is that even if the buffer has collected enough elements to make up a chunk, these won’t  always be part of the emitted chunk. I suspect this is what’s happening in the "accumulation and splitting" test, which is more flaky than it should probably be (some level of flakiness is probably acceptable since it’s fundamentally a time based test, but with the current logic it’s very easy to get a 50%, or even higher, failure rate). 

When running the test below (a slightly modified version of the original test that includes an additional `sleep`) the situation is even worse: it’s nearly impossible to get the test passing)

```scala
test("accumulation and splitting 2".only) {
  val t = 200.millis
  val size = 5
  val sleep = Stream.sleep_[IO](2 * t)
  val longSleep = Stream.sleep_[IO](10 * t)

  def chunk(from: Int, size: Int) =
    Stream.range(from, from + size).chunkAll.unchunks

  // this test example is designed to have good coverage of
  // the chunk manipulation logic in groupWithin
  val source =
  chunk(from = 1, size = 3) ++
    sleep ++
    chunk(from = 4, size = 1) ++ longSleep ++
    chunk(from = 5, size = 11) ++
    chunk(from = 16, size = 7)

  val expected = List(
    List(1, 2, 3),
    List(4),
    List(5, 6, 7, 8, 9),
    List(10, 11, 12, 13, 14),
    List(15, 16, 17, 18, 19),
    List(20, 21, 22)
  )

  source.groupWithin(size, t).map(_.toList).assertEmits(expected)
}
```

This PR builds on top of #3162 introducing a new mechanism to improve the accuracy of the timeout logic.
The trick is to wait for a whole chunk to be there, and to achieve that we run a stream within a stream, allowing us to reuse the fs2 api to interact with the stream in terms of `chunks`.

The result is a more accurate behaviour

I've run both "accumulation and splitting" tests a number of times and had zero failures so far

Screenshot 2023-03-25 at 18 38 29

### Performance

benchmarks figures are similar to the other suggested [implementation](https://github.com/typelevel/fs2/pull/3162#issuecomment-1477075048) (I won't attach them again, but let me know if you get different results), better than the existing implementation and up to 100% faster on large streams (ops/sec), with even better `gc` stats, (assuming a lower score is better) I've run them only once though

### Simpler logic 

Like the previous PR this one removes the custom error propagation (relies on `concurrently`) and limits the use of `Semaphore` to what's needed, (`demand` is gone in favour of a bounded queue) piggybacking on the streams api to await a chunk

### Considerations

I haven't yet experienced the issue reported [here](https://github.com/typelevel/fs2/pull/3162#discussion_r1145009634). So I haven't included the change suggested to prevent that edge case. I'd like to understand how that can happen first. Let me know what you think about the suggestion to address it (see previous PR) and if you have a better idea.

### Notes

- may be a fix for #2432 (let me know what's your experience, I haven't seen a failure yet)
- fixes #3169 (unless #3183 gets merged first)

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1404,106 +1404,74 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   def groupWithin[F2[x] >: F[x]](
       chunkSize: Int,
       timeout: FiniteDuration
-  )(implicit F: Temporal[F2]): Stream[F2, Chunk[O]] = {
+  )(implicit F: Temporal[F2]): Stream[F2, Chunk[O]] =
+    if (timeout.toNanos == 0 || chunkSize == 1) chunkN(chunkSize)
+    else
+      Stream.force {
+        for {
+          supply <- Semaphore[F2](0)
+          supplyEnded <- SignallingRef.of[F2, Boolean](false)
+          buffer <- Queue.bounded[F2, O](chunkSize) // buffering and backpressure
+          awaitFirstChunk = Stream.fromQueueUnterminated(buffer, chunkSize).chunkMin(1).head
+        } yield {
+          // note: it can produce an empty chunk
+          val emitChunk: F2[Chunk[O]] = buffer.tryTakeN(Some(chunkSize)).map(Chunk.seq)
 
-    case class JunctionBuffer[T](
-        data: Vector[T],
-        endOfSupply: Option[Either[Throwable, Unit]],
-        endOfDemand: Option[Either[Throwable, Unit]]
-    ) {
-      def splitAt(n: Int): (JunctionBuffer[T], JunctionBuffer[T]) =
-        if (this.data.size >= n) {
-          val (head, tail) = this.data.splitAt(n.toInt)
-          (this.copy(tail), this.copy(head))
-        } else {
-          (this.copy(Vector.empty), this)
-        }
-    }
+          // we need to check the buffer size, rather than the available supply since
+          // the supply is increased at the end so it won't always report the buffer size accurately
+          val isBufferEmpty: F2[Boolean] = buffer.size.map(_ == 0)
 
-    val outputLong = chunkSize.toLong
-    fs2.Stream.force {
-      for {
-        demand <- Semaphore[F2](outputLong)
-        supply <- Semaphore[F2](0L)
-        buffer <- Ref[F2].of(
-          JunctionBuffer[O](Vector.empty[O], endOfSupply = None, endOfDemand = None)
-        )
-      } yield {
-        /* - Buffer: stores items from input to be sent on next output chunk
-         * - Demand Semaphore: to avoid adding too many items to buffer
-         * - Supply: counts filled positions for next output chunk */
-        def enqueue(t: O): F2[Boolean] =
-          for {
-            _ <- demand.acquire
-            buf <- buffer.modify(buf => (buf.copy(buf.data :+ t), buf))
-            _ <- supply.release
-          } yield buf.endOfDemand.isEmpty
+          val streamExhausted: F2[Boolean] = (isBufferEmpty, supplyEnded.get).mapN(_ && _)
 
-        val dequeueNextOutput: F2[Option[Vector[O]]] = {
-          // Trigger: waits until the supply buffer is full (with acquireN)
-          val waitSupply = supply.acquireN(outputLong).guaranteeCase {
-            case Outcome.Succeeded(_) => supply.releaseN(outputLong)
-            case _                    => F.unit
+          // "subscribing" to the buffer waiting for the first chunk. Note: we
+          // can't wait forever: at most we can wait until the supply has ended
+          val awaitChunk: F2[Chunk[O]] = {
+
+            val waitForOne = Stream.eval(supply.acquire).interruptWhen(supplyEnded).compile.drain
+
+            // we need to check again after waiting: (we might have just received the final chunk,
+            // in that case we need to flush any residual elements in the buffer without blocking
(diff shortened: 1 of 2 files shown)


armanbilge:
> avoid deconstructing tuple in for-compr to avoid CI js failure

Sorry, this is my trigger 😆 

This issue is not related to Scala.js. The core Scala language (i.e. syntax) actually works 100% identically for all three platforms (if you ever find a case where it doesn't, that's a legitimate bug and we should fix it :).

Sometimes, a JS CI job will happen to fail first. That doesn't mean it's a JS-specific issue.

/rant 😇

Angel-O:
@armanbilge I had a look, this time it sounds like a Scala.js problem on scala 3 (setting the version to 2 and running the CI command `++ 3.2.2 Test/scalaJSLinkerResult` doesn't cause issues). 

I tried to set the compiler flag `-source:future`, but that brings up a bunch of other unrelated errors/warning, that perhaps deserve a separate PR. 

Also I've found a similar issue [here](https://github.com/zio/zio/issues/6468): sounds like a complier issue from the discussion. Do you think it's worth spending more time on this ?

armanbilge:
>  I had a look, this time it sounds like a Scala.js problem on scala 3

You can't reproduce it with Scala JVM?

Angel-O:
> You can't reproduce it with Scala JVM?

Nope: the command is not available on jvm

armanbilge:
`scalaJSLinkerResult` is just running `compile`, try compiling on JVM :)

Angel-O:
yeah `compile`  works on jvm, (I think it's a subcommand of `test` and I've run these tests a number of times 😆)

armanbilge:
`Test/compile` on Scala 3 JVM works, and `Test/compile` on Scala 3 JS doesn't work? That would be a bug.

Angel-O:
> Test/compile on Scala 3 JVM works, and Test/compile on Scala 3 JS doesn't work? That would be a bug.

I've run the jvm tests on scala 2, let me try with scala 3

armanbilge:
> Hopefully deconstructing vs non-deconstructing the tuple is not a deal breaker.

It's not a problem at all, just a symptom of cross-compiling across Scala 2 and Scala 3. Sorry that I derailed things 😓

Angel-O:
> It's not a problem at all, just a symptom of cross-compiling across Scala 2 and Scala 3. Sorry that I derailed things 😓

No worries: it's just because I've opened a bunch of PRs doing the same thing and it might get confusing. Thanks for explaining though, learned something again 🥇

armanbilge:
Thanks for that update!

> In light of this, I've decided to restore the `Vector` (`Chunk` doesn't seem to improve performance, at least with the existing benchmarks, will test with different benchmark params)

I have a question about this: I understand that `Chunk` didn't improve performance in your benchmarks, but does it degrade performance? The reason I ask is because there is a big difference between e.g. a `Chunk[Byte]` and a `Vector[Byte]`: the chunk can be backed by a primitive array but the vector will be boxed.

Angel-O:
> I have a question about this: I understand that Chunk didn't improve performance in your benchmarks, but does it degrade performance? The reason I ask is because there is a big difference between e.g. a Chunk[Byte] and a Vector[Byte]: the chunk can be backed by a primitive array but the vector will be boxed.

@armanbilge sorry for the late reply.

I see, thanks for explaining. When I ran the benchmark last time using `s2.Chunk` I had noticed a similar behaviour (memory-wise) observed with the `cats Queue` (visual vm stats/graphs were similar). 
In terms of `ops/sec` instead I didn't notice any particular difference, but I was running the existing benchmarks using `Int` values.

I'm wondering if the difference between `Chunk[Byte]` vs `Vector[Byte]` calls for a separate method, or maybe better a separate extended api targeting specifically `Streams[F, Byte]`

armanbilge:
> I'm wondering if the difference between `Chunk[Byte]` vs `Vector[Byte]` calls for a separate method, or maybe better a separate extended api specifically targeting `Streams[F, Byte]`

It doesn't just apply to `Byte`, it applies to all primitive types. The specific needs of FS2 are precisely why `Chunk` exists and is used in APIs instead of e.g. `Vector`. Here's a presentation by Michael discussing some of this :)
https://www.youtube.com/watch?v=wOybldcyMLs

PR #3189, opened 2023-03-26
Fix unintended unchunking by adding override map

### LINKED ISSUE #3187: Unintended unchunking when using Monad instance of Stream
When using implementations of cats' type classes instances `fs2.Stream#monadErrorInstance` or `fs2.StreamLowPriority#monadInstance` unchunking will occur as a side effect.

```scala
def testCase(pipe: Pipe[fs2.Pure, Int, (Int, Int)]) = (Stream(1,2,3) ++ Stream(4,5,6)) .through(pipe).chunks.toList

testCase(_.map(e => e -> (e + 1)))
// List(Chunk((1,2), (2,3), (3,4)), Chunk((4,5), (5,6), (6,7)))

testCase(_.fproduct(_ + 1))
// List(Chunk((1,2)), Chunk((2,3)), Chunk((3,4)), Chunk((4,5)), Chunk((5,6)), Chunk((6,7)))
```

This happens because the default implementation of `map` in `Monad` is `flatMap(fa)(a => pure(f(a)))`. I believe that overriding that default implementation in mentioned instances using:
```scala
override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = fa.map(f)
```
will fix that. I think such unchunking may decrease the performance. By a lot?

yyy1000:
It wants to fix #3187
It's my first time to PR typelevel! If there's something I need to improve or change, I'm willing to do that!

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3461,6 +3461,7 @@ object Stream extends StreamLowPriority {
     */
   def fromOption[F[_]]: PartiallyAppliedFromOption[F] =
     new PartiallyAppliedFromOption(dummy = true)
+
   private[fs2] final class PartiallyAppliedFromEither[F[_]](
       private val dummy: Boolean
   ) extends AnyVal {
@@ -5236,10 +5237,7 @@ object Stream extends StreamLowPriority {
       def handleErrorWith[A](s: Stream[F, A])(h: Throwable => Stream[F, A]) =
         s.handleErrorWith(h)
       def raiseError[A](t: Throwable) = Stream.raiseError[F](t)
-      override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = {
-        println("error map")
-        fa.map(f)
-      }
+      override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = fa.map(f)
     }
 
   /** `Monoid` instance for `Stream`. */
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3089,6 +3089,21 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   override def toString: String = "Stream(..)"
 }
 
+class StreamMonad[F[_]] extends Monad[Stream[F, *]] {
+  override def pure[A](x: A): Stream[F, A] = Stream.emit(x)
+
+  override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = fa.map(f)
+
+  override def flatMap[A, B](fa: Stream[F, A])(f: A => Stream[F, B]): Stream[F, B] =
+    fa.flatMap(f)
+
+  override def tailRecM[A, B](a: A)(f: A => Stream[F, Either[A, B]]): Stream[F, B] =
+    f(a).flatMap {
+      case Left(a)  => tailRecM(a)(f)
+      case Right(b) => Stream(b)
+    }
+}
+
 object Stream extends StreamLowPriority {
 
   /** Creates a pure stream that emits the supplied values. To convert to an effectful stream, use `covary`. */
@@ -3446,7 +3461,6 @@ object Stream extends StreamLowPriority {
     */
   def fromOption[F[_]]: PartiallyAppliedFromOption[F] =
     new PartiallyAppliedFromOption(dummy = true)
-
   private[fs2] final class PartiallyAppliedFromEither[F[_]](
       private val dummy: Boolean
   ) extends AnyVal {
@@ -5217,18 +5231,15 @@ object Stream extends StreamLowPriority {
   implicit def monadErrorInstance[F[_]](implicit
       ev: ApplicativeError[F, Throwable]
   ): MonadError[Stream[F, *], Throwable] =
-    new MonadError[Stream[F, *], Throwable] {
-      def pure[A](a: A) = Stream(a)
+    new StreamMonad[F] with MonadError[Stream[F, *], Throwable] {
+      override def pure[A](a: A) = Stream(a)
       def handleErrorWith[A](s: Stream[F, A])(h: Throwable => Stream[F, A]) =
         s.handleErrorWith(h)
       def raiseError[A](t: Throwable) = Stream.raiseError[F](t)
-      override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = fa.map(f)
-      def flatMap[A, B](s: Stream[F, A])(f: A => Stream[F, B]) = s.flatMap(f)
-      def tailRecM[A, B](a: A)(f: A => Stream[F, Either[A, B]]) =
-        f(a).flatMap {
-          case Left(a)  => tailRecM(a)(f)
-          case Right(b) => Stream(b)
-        }
+      override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = {
+        println("error map")
+        fa.map(f)
+      }
     }
 
   /** `Monoid` instance for `Stream`. */
@@ -5362,21 +5373,8 @@ object Stream extends StreamLowPriority {
 }
 
 private[fs2] trait StreamLowPriority {
-  implicit def monadInstance[F[_]]: Monad[Stream[F, *]] =
-    new Monad[Stream[F, *]] {
-      override def pure[A](x: A): Stream[F, A] = Stream.emit(x)
-
+  implicit def monadInstance[F[_]]: StreamMonad[F] =
+    new StreamMonad[F] {
       override def unit: Stream[F, Unit] = Stream.unit
-
-      override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = fa.map(f)
-
-      override def flatMap[A, B](fa: Stream[F, A])(f: A => Stream[F, B]): Stream[F, B] =
-        fa.flatMap(f)
-
-      override def tailRecM[A, B](a: A)(f: A => Stream[F, Either[A, B]]): Stream[F, B] =
-        f(a).flatMap {
-          case Left(a)  => tailRecM(a)(f)
-          case Right(b) => Stream(b)
-        }
     }
 }
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -5222,6 +5222,7 @@ object Stream extends StreamLowPriority {
       def handleErrorWith[A](s: Stream[F, A])(h: Throwable => Stream[F, A]) =
         s.handleErrorWith(h)
       def raiseError[A](t: Throwable) = Stream.raiseError[F](t)
+      override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = fa.map(f)
       def flatMap[A, B](s: Stream[F, A])(f: A => Stream[F, B]) = s.flatMap(f)
       def tailRecM[A, B](a: A)(f: A => Stream[F, Either[A, B]]) =
         f(a).flatMap {
@@ -5366,6 +5367,8 @@ private[fs2] trait StreamLowPriority {
       override def pure[A](x: A): Stream[F, A] = Stream.emit(x)
 
       override def unit: Stream[F, Unit] = Stream.unit
+
+      override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = fa.map(f)
 
       override def flatMap[A, B](fa: Stream[F, A])(f: A => Stream[F, B]): Stream[F, B] =
         fa.flatMap(f)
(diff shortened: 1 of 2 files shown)


mpilquist:
Thanks! Can you run `set prePR` and commit the result?

core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -1024,6 +1024,11 @@ class StreamSuite extends Fs2Suite {
     assert(compileErrors("Stream.eval(IO(1)).through(p)").nonEmpty)
   }
 
+  test("Unintended unchunking when using Monad instance of Stream") {
```
bwiercinski, line 1027: you have only tested StreamLowPriority instance. I think that duplicating this test and adding `.covary[IO]` before `.through` (and `.compile` before `.toList`) will be sufficient :)

bwiercinski (changes requested):

core/shared/src/test/scala/fs2/StreamSuite.scala
yyy1000: Thanks! I will do that.

yyy1000:
Hi, I run prePR and add the new test. :)

core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -1024,6 +1024,18 @@ class StreamSuite extends Fs2Suite {
...
+    def generateTuple(pipe: Pipe[fs2.Pure, Int, (Int, Int)]) =
+      (Stream(1, 2, 3) ++ Stream(4, 5, 6)).through(pipe).chunks.toList
+    assertEquals(generateTuple(_.map(e => e -> (e + 1))), generateTuple(_.fproduct(_ + 1)))
+  }
+
+  test("Unintended unchunking when using Stream#monadErrorInstance") {
+    def generateTuple(pipe: Pipe[fs2.Pure, Int, (Int, Int)]) =
+      (Stream(1, 2, 3) ++ Stream(4, 5, 6)).covary[Pure].through(pipe).chunks.compile.toList
+    assertEquals(generateTuple(_.map(e => e -> (e + 1))), generateTuple(_.fproduct(_ + 1)))
+  }
```
armanbilge, line 1037: sorry if I missed it, is this using `MonadError` instance? Since it also appears to be a `Pure` stream :)

yyy1000: Sorry, I don't know it will not use the `MonadError` instance before.😂
I changed it to `IO` and locally printed some logs, and found it will use the `MonadError`.
As for the assertEquals, I found that two 'IO' values can't be compared. So I use unsafeRunSync(). Is it right?

armanbilge: No problem! :) Instead of using `unsafeRunSync()`, you can do something like:

```scala 
val lhs = generateTuple(_.map(e => e -> (e + 1)))
val rhs = generateTuple(_.fproduct(_ + 1))
(lhs, rhs).mapN(assertEquals(_, _))
```

yyy1000: Yeah! This looks more functional.
I changed to that. Thanks!

armanbilge:
Re-triggering CI

core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -1024,6 +1024,20 @@ class StreamSuite extends Fs2Suite {
...
+
+  test("Unintended unchunking when using Stream#monadErrorInstance") {
+    def generateTuple(pipe: Pipe[IO, Int, (Int, Int)]) =
+      (Stream(1, 2, 3) ++ Stream(4, 5, 6)).covary[IO].through(pipe).chunks.compile.toList
+    val lhs = generateTuple(_.map(e => e -> (e + 1)))
+    val rhs = generateTuple(_.fproduct(_ + 1))
+    (lhs, rhs).mapN(assertEquals(_, _))
+  }
+
```
diesalbla, line 1040: These tests seem to be needlessly complicated, with the use of different literal values, the tuple, the increment, etc. 

In a "test-driven-development" fashion, the process is to 1. write the smallest possible tests that fails 2. change code, 3. check that test passes. It seems to me that would be achieved, for the problem, with a test core that did the following: 

```Scala
def countChunks(str: Stream[Nothing, _]): Int = 
  str.chunks.toList.length

val source = Stream(0) ++ Stream(0, 0)
val mapped = source.map(_ => 1)
assertEquals(countChunks(mapped), 2)
```

yyy1000: Thanks for your review!
It's helpful, and I will make a change for this:)

yyy1000: Hi, sorry if I made some mistakes.
I run a local test and found that only calling the 'map' method like `source.map(_ => 1)` can't call the method I changed. :)
I changed my code to below to make it simpler, is it OK?
```scala
  test("Unintended unchunking when using StreamLowPriority#monadInstance") {
    def countChunks(): Int =
      (Stream(0) ++ Stream(0, 0)).fproduct(_ + 1).chunks.toList.length
    assertEquals(countChunks(), 2)
  }

  test("Unintended unchunking when using Stream#monadErrorInstance") {
    def countChunks() =
      (Stream(0) ++ Stream(0, 0)).covary[IO].fproduct(_ + 1).chunks.compile.toList.map(_.length)
    countChunks().assertEquals(2)
  }
```

armanbilge: Actually, we can do even better.
```
Stream.monadInstance.map(source)(_ => 1)
// etc.
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -5222,6 +5222,7 @@ object Stream extends StreamLowPriority {
       def handleErrorWith[A](s: Stream[F, A])(h: Throwable => Stream[F, A]) =
         s.handleErrorWith(h)
       def raiseError[A](t: Throwable) = Stream.raiseError[F](t)
+      override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = fa.map(f)
```
armanbilge, line 5225: It's unfortunate that this fix has to be duplicated. Perhaps we should extract a:
```scala
class StreamMonad[F] extends Monad[Stream[F, *]] {
  ...
}
```
Then the code can be shared.

core/shared/src/test/scala/fs2/StreamSuite.scala
yyy1000: Sorry, but I encountered a problem. :(
`Stream.monadInstance.map(source)` works well, but for `Stream.monadErrorInstance.map(str)`, my IDE shows 'No implicits found for parameter ev: ApplicativeError[F_ , Throwable]', what's that? 😂

armanbilge: Thanks for trying that! If you make the change I suggested in https://github.com/typelevel/fs2/pull/3189#discussion_r1154558125, then we only need one test.

yyy1000: OK, but can it cover the 'MonadErrorInstance'? 🤔

armanbilge: Yes, because my proposal is that `Monad` and `MonadError` implementations should share the same code. Then we don't need to write the same code twice, fix the same bugs twice, and add the same tests twice :)

core/shared/src/main/scala/fs2/Stream.scala
yyy1000: Now my understanding is:
After extracting a class like
```scala
class StreamMonad[F] extends Monad[Stream[F, *]] {
  ...
}
```
the original code
```scala
private[fs2] trait StreamLowPriority {
  implicit def monadInstance[F[_]]: Monad[Stream[F, *]] =
    new Monad[Stream[F, *]] {
      ...
}
```
will become like below?
```scala
private[fs2] trait StreamLowPriority {
  implicit def monadInstance[F[_]]: StreamMonad[Stream[F, *]] =
    new StreamMonad[Stream[F, *]] {
      ...
    }
}
```

armanbilge:
```scala
implicit def monadErrorInstance[F[_]]: MonadError[Stream[F, *], Throwable] =
new StreamMonad[F] with MonadError[Stream[F, *], Throwable] {
  // inherit monad methods, implement monad error methods
}

Comment (yyy1000) [reply core/shared/src/main/scala/fs2/Stream.scala:5225]: Thanks! It's very helpful.😊

**Comment by yyy1000** [comment]
I try to extract the class and please review. :)

File: core/shared/src/main/scala/fs2/Stream.scala (hunk 4 above)
Comment (diesalbla) [reply core/shared/src/main/scala/fs2/Stream.scala:5240]: @armanbilge IMHIO it is always good practice to give explicit names to long classes, just as it is good practice to name long lambdas. @yyy1000 Just remember to keep it private to the `Stream` companion object.

File: core/shared/src/main/scala/fs2/Stream.scala [hunk 5]
```diff
@@ -3089,6 +3089,21 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   override def toString: String = "Stream(..)"
 }
 
+class StreamMonad[F[_]] extends Monad[Stream[F, *]] {
```
diesalbla, line 3092: Could we place this inside the `Stream` companion object and mark it as `private`?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -5361,19 +5371,8 @@ object Stream extends StreamLowPriority {
 }
 
 private[fs2] trait StreamLowPriority {
-  implicit def monadInstance[F[_]]: Monad[Stream[F, *]] =
-    new Monad[Stream[F, *]] {
-      override def pure[A](x: A): Stream[F, A] = Stream.emit(x)
-
+  implicit def monadInstance[F[_]]: StreamMonad[F] =
+    new StreamMonad[F] {
```
diesalbla, line 5375: Since `StreamMonad` is not an abstract class, you can omit the body (the `{` / `}` block) here.

diesalbla (approved):

core/shared/src/main/scala/fs2/Stream.scala
yyy1000: OK, I will do it the next day :), (it's a little late in my place)!

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3089,6 +3089,21 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+  override def map[A, B](fa: Stream[F, A])(f: A => B): Stream[F, B] = fa.map(f)
+
+  override def flatMap[A, B](fa: Stream[F, A])(f: A => Stream[F, B]): Stream[F, B] =
+    fa.flatMap(f)
+
+  override def tailRecM[A, B](a: A)(f: A => Stream[F, Either[A, B]]): Stream[F, B] =
+    f(a).flatMap {
+      case Left(a)  => tailRecM(a)(f)
+      case Right(b) => Stream(b)
+    }
```
armanbilge, line 3104: If you do `extends StackSafeMonad[Stream[F, *]]` instead of `Monad` then we don't need this method implementation :)

yyy1000: OK!

core/shared/src/main/scala/fs2/Stream.scala
yyy1000: Thanks! :)
If I do this, I will move the `unit` method to the `StreamMonad`. And `monadErrorInstance` will also use that method.
But it don't use the override unit method before.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -5217,17 +5228,11 @@ object Stream extends StreamLowPriority {
   implicit def monadErrorInstance[F[_]](implicit
       ev: ApplicativeError[F, Throwable]
   ): MonadError[Stream[F, *], Throwable] =
-    new MonadError[Stream[F, *], Throwable] {
-      def pure[A](a: A) = Stream(a)
+    new StreamMonad[F] with MonadError[Stream[F, *], Throwable] {
+      override def pure[A](a: A) = Stream(a)
```
armanbilge, line 5232: The inherited implementation is fine :)
```suggestion
```

core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -1024,6 +1024,13 @@ class StreamSuite extends Fs2Suite {
     assert(compileErrors("Stream.eval(IO(1)).through(p)").nonEmpty)
   }
 
+  test("Unintended unchunking") {
```
armanbilge, line 1027:
```suggestion
  test("monad instance overrides map and preserves chunks") {
```

armanbilge (approved):
Thanks for all your work on this!

PR #3190, opened 2023-03-26
Chunk - Optimise Flatten implementation.

diesalbla:
The current flatten performs one Chunk.Queue insertion per source chunk, which means it creates a new "Chunk.Queue" object and one "Queue" object for each one. Instead, we can avoid those objects and use a Queue- builder.

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/ChunkFlatten.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/ChunkFlatten.scala
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2
+
+import org.openjdk.jmh.annotations.{Benchmark, Param, Scope, Setup, State}
+
+import cats.syntax.all._
+
+@State(Scope.Thread)
+class ChunkFlatten {
+  @Param(Array("8", "32", "128"))
+  var numberChunks: Int = _
+
+  @Param(Array("1", "2", "10", "50"))
+  var chunkSize: Int = _
+
+  case class Obj(dummy: Boolean)
+  object Obj {
+    def create: Obj = Obj(true)
+  }
+
+  var chunkSeq: Seq[Chunk[Obj]] = _
+
+  @Setup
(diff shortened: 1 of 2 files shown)


armanbilge:
We can make a similar optimization here?
https://github.com/typelevel/fs2/blob/f5a4d11dc8fed5b7713a34e02a7f60f69de1c235/core/shared/src/main/scala/fs2/Chunk.scala#L1173-L1174

mpilquist:
I'm not convinced this one is worth the gains. Can we benchmark it?

mpilquist:
Yeah, I'm concerned we're inlining implementations here without proof that the allocations matter. I don't want to put too much faith in to things like escape analysis, but also want to ensure any code duplication / inlining is justified with benchmarks showing the improvements actually matter in real use cases.

diesalbla:
As an update, I have moved the "inlined" implementation to a `def build` factory method within the `Queue` companion object, so the constructor can be kept private as it is.

diesalbla:
I have pushed a file with a benchmark for flatten, with different parameters of number of chunks to flatten and size of each chunk. Running it with the command 
```
Jmh/run -i 3 -wi 3 -f1 -t1 .*ChunkFlatten.* -prof gc -jvmArgs -XX:MaxInlineLevel=18
```

Focusing on the `gc.alloc.rate.norm` of each case, it obtains the following results for `main`: 

```
:·gc.alloc.rate.norm            1               8  thrpt    3       688.000 ±       0.001    B/op
:·gc.alloc.rate.norm            1              32  thrpt    3      2608.000 ±       0.001    B/op
:·gc.alloc.rate.norm            1             128  thrpt    3     10288.000 ±       0.001    B/op
:·gc.alloc.rate.norm            2               8  thrpt    3       688.000 ±       0.001    B/op
:·gc.alloc.rate.norm            2              32  thrpt    3      2608.000 ±       0.004    B/op
:·gc.alloc.rate.norm            2             128  thrpt    3     10288.000 ±       0.005    B/op
:·gc.alloc.rate.norm           10               8  thrpt    3       688.000 ±       0.001    B/op
:·gc.alloc.rate.norm           10              32  thrpt    3      2560.000 ±       0.002    B/op
:·gc.alloc.rate.norm           10             128  thrpt    3     10288.001 ±       0.016    B/op
:·gc.alloc.rate.norm           50               8  thrpt    3       688.000 ±       0.001    B/op
:·gc.alloc.rate.norm           50              32  thrpt    3      2608.000 ±       0.001    B/op
:·gc.alloc.rate.norm           50             128  thrpt    3     10288.001 ±       0.016    B/op

```

And these are the results with the changes in this PR: 

```
:·gc.alloc.rate.norm            1               8  thrpt    3       360.000 ±       0.001    B/op
:·gc.alloc.rate.norm            1              32  thrpt    3       936.000 ±       0.001    B/op
:·gc.alloc.rate.norm            1             128  thrpt    3      3240.001 ±       0.012    B/op
:·gc.alloc.rate.norm            2               8  thrpt    3       192.000 ±       0.001    B/op
:·gc.alloc.rate.norm            2              32  thrpt    3       936.000 ±       0.003    B/op
:·gc.alloc.rate.norm            2             128  thrpt    3      3240.000 ±       0.014    B/op
:·gc.alloc.rate.norm           10               8  thrpt    3       192.000 ±       0.001    B/op
:·gc.alloc.rate.norm           10              32  thrpt    3       936.000 ±       0.002    B/op
:·gc.alloc.rate.norm           10             128  thrpt    3      3240.000 ±       0.009    B/op
:·gc.alloc.rate.norm           50               8  thrpt    3       360.000 ±       0.001    B/op
:·gc.alloc.rate.norm           50              32  thrpt    3       936.000 ±       0.003    B/op
:·gc.alloc.rate.norm           50             128  thrpt    3      3240.000 ±       0.001    B/op
```

in all cases the total number of allocations is reduced by up to a third or more than half.

diesalbla:
@mpilquist Let me know if there are any further benchmarks that can be run.

mpilquist:
Benchmark looks good, though I'm still a bit concerned about chasing allocation rate improvements like this. What type of application would benefit from this change?

diesalbla:
> Benchmark looks good, though I'm still a bit concerned about chasing allocation rate improvements like this. What type of application would benefit from this change?

Flattening chunks could be useful in the context of using `unconsN` or `unconsMin` or `groupWithin`, although it is currently not used. However, I do not have any measurements (memory profiling) from any production system, that may allow us to see what other major allocation hotspots there are. Any actual measurements on the field would be helpful.

diesalbla:
@mpilquist What is your verdict on this PR? Merge? Close? Worthy improvement? Would rather seek a stronger optimisation?

mpilquist:
I'm good with merging. Can we inline `Chunk.build` in to the `else` block inside of `flatten`?

diesalbla:
> I'm good with merging. Can we inline Chunk.build in to the else block inside of flatten?

No, because the constructor of `Chunk.Queue`, which the `Chunk.build` code uses, is private.

PR #3191, opened 2023-03-26
Chunk - Add optimised traverse_ for StackSafeMonad.

diesalbla:
If our Applicative F is actually an instance of `StackSafeMonad`, we can avoid the default implementation of traverse_, which is based on a `foldRight` method of Foldable and has to allocate a set of `Eval` objects. Instead, we can just loop on the monad itself.

_Note_ Given that in practice Streams are mostly used with IO-like datatypes which are stack-safe, this should have some impact.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -28,7 +28,7 @@ import scala.reflect.ClassTag
 import scodec.bits.{BitVector, ByteVector}
 import java.nio.{Buffer => JBuffer, ByteBuffer => JByteBuffer, CharBuffer => JCharBuffer}
 
-import cats.{Alternative, Applicative, Eq, Eval, Monad, Monoid, Traverse, TraverseFilter}
+import cats._
 import cats.data.{Chain, NonEmptyList}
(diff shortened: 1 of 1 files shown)


armanbilge:
Linking to:
- https://github.com/typelevel/cats/issues/4408

Actually, if we made this optimization in Cats, we probably wouldn't need to repeat it here?

diesalbla:
> Actually, if we made this optimization in Cats, we probably wouldn't need to repeat it here?

As it happens I was taking a look at the `cats` implementation, but the `Foldable` type class seems to provide little. With `Chunk` we can optimise it right away.

armanbilge:
> Note that the ticket refers to the `traverse` method, not the procedural `traverse_`

Yes indeed, but we can make the described optimizations for both `traverse` and `traverse_` upstream in Cats.

diesalbla:
> Yes indeed, but we can make the described optimizations for both traverse and traverse_ upstream in Cats.

It may take a while for those to get released, if and when they are done (the link is to an issue). This PR is ready as of now.

armanbilge:
Sorry, I didn't mean it as a blocker here 🙂 rather, to further justify landing these changes in Cats.

PR #3196, opened 2023-04-01
Stream - microptimise EvalTap

diesalbla:
Expand the implementation of evalMap inside evalTap. Instead of making the `as` call in the underlying effect `F`, we can just use the Pull `>>` sequencing operator, to ignore the result of running the action.


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1041,8 +1041,11 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * Not as powerful as `observe` since not all pipes can be represented by `O => F[O2]`, but much faster.
     * Alias for `evalMap(o => f(o).as(o))`.
     */
-  def evalTap[F2[x] >: F[x]: Functor, O2](f: O => F2[O2]): Stream[F2, O] =
-    evalMap(o => f(o).as(o))
+  def evalTap[F2[x] >: F[x], O2](f: O => F2[O2])(implicit F: Functor[F2]): Stream[F2, O] = {
```
armanbilge, line 1044:
```suggestion
    @deprecated("Use overload without functor", "3.7.0")
    private[fs2] def evalTap[F2[x] >: F[x], O2](f: O => F2[O2], F: Functor[F2]): Stream[F2, O] =
      evalTap(f)

    def evalTap[F2[x] >: F[x], O2](f: O => F2[O2]): Stream[F2, O] = {
```

armanbilge (approved):
Nice, always happy to drop a constraint.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1041,8 +1041,14 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * Not as powerful as `observe` since not all pipes can be represented by `O => F[O2]`, but much faster.
     * Alias for `evalMap(o => f(o).as(o))`.
     */
-  def evalTap[F2[x] >: F[x]: Functor, O2](f: O => F2[O2]): Stream[F2, O] =
-    evalMap(o => f(o).as(o))
+  @deprecated("Use overload without functor", "3.7.0")
+  private[fs2] def evalTap[F2[x] >: F[x], O2](f: O => F2[O2], F: Functor[F2]): Stream[F2, O] =
+    evalTap(f)
+
+  def evalTap[F2[x] >: F[x], O2](f: O => F2[O2]): Stream[F2, O] = {
```
armanbilge, line 1048: Oh, my edit caused the scaladoc to become separated from the public method.

diesalbla: fixes.

armanbilge (approved):

PR #3197, opened 2023-04-01
Stream - expand evalMapChunk, evalTapChunk.

diesalbla:
In `evalTapChunk`, instead of `traverse` we can use `traverse_` (which discards outputs), and just keep emitting same input chunk, without allocating a new entire output chunk.

### DIFF
--- a/core/shared/src/main/scala/fs2/Pull.scala
+++ b/core/shared/src/main/scala/fs2/Pull.scala
@@ -353,6 +353,7 @@ object Pull extends PullLowPriority {
         case _               => Uncons(self)
       }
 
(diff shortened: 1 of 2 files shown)


mpilquist:
Like the other PR, we need benchmarks showing that the complexity is worth the cost.

diesalbla:
> Like the other PR, we need benchmarks showing that the complexity is worth the cost.

TBH I do not think that, for the `evalMapChunk`, it would make much difference. On the other hand,  for `evalTapChunk`, the difference between building a new `Chunk` (with the `traverse`) and one `.as` operation per item, vs a `traverse_` and emitting existing output, that could be more significant.

PR #3198, opened 2023-04-03
Specialize `Chunk#toArraySlice`

armanbilge:
1. Specializes `Chunk#toArraySlice` for `ByteVector` and `{Byte,Char}Buffer` to avoid array copies.
2. Fix class cast exceptions in `Chunk.{Byte,Char}Buffer#toArray`.
3. Avoid converting buffers to read-only, since this prevents access to underlying array.
4. Use `toArraySlice` to implement `to{Byte,Char}Buffer`

### DIFF
--- a/core/shared/src/test/scala/fs2/ChunkSuite.scala
+++ b/core/shared/src/test/scala/fs2/ChunkSuite.scala
@@ -27,7 +27,10 @@ import cats.kernel.laws.discipline.EqTests
 import cats.laws.discipline.{AlternativeTests, MonadTests, TraverseFilterTests, TraverseTests}
 import org.scalacheck.{Arbitrary, Cogen, Gen, Test}
 import org.scalacheck.Prop.forAll
+import scodec.bits.ByteVector
 
+import java.nio.ByteBuffer
+import java.nio.CharBuffer
 import scala.reflect.ClassTag
 
 class ChunkSuite extends Fs2Suite {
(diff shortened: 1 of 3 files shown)


core/shared/src/test/scala/fs2/ChunkSuite.scala
```diff
@@ -225,13 +228,46 @@ class ChunkSuite extends Fs2Suite {
...
-  test("ArraySlice does not copy when chunk is already an ArraySlice instance") {
+  test("toArraySlice does not copy when chunk is already an ArraySlice instance") {
     val chunk: Chunk[Int] = Chunk.ArraySlice(Array(0))
     assert(chunk eq chunk.toArraySlice)
     val chunk2: Chunk[Any] = Chunk.ArraySlice(Array(new Object))
     assert(chunk2 eq chunk2.toArraySlice)
   }
 
+  test("toArraySlice does not copy when chunk is an array-backed bytevector") {
+    val arr = Array[Byte](0)
```
diesalbla, line 239: The singleton Array may be a degenerate case. Could we use something like `"Hello World".getBytes()`?

core/shared/src/test/scala/fs2/ChunkSuite.scala
```diff
@@ -225,13 +228,46 @@ class ChunkSuite extends Fs2Suite {
...
+    val chunk: Chunk[Byte] = Chunk.byteVector(ByteVector.view(arr))
+    assert(chunk.toArraySlice.values eq arr)
+  }
+
+  test("ByteVectorChunk#toArraySlice does not throw class cast exception") {
+    val chunk: Chunk[Byte] = Chunk.byteVector(ByteVector.view(Array[Byte](0)))
```
diesalbla, line 245: Same.

diesalbla (approved):

armanbilge:
Actually, maybe we shouldn't be comparing `ClassTag`s for equality at all.

armanbilge:
@diesalbla I'd appreciate it if you didn't squash-merge my PRs 😕 I worked carefully to split this PR into a series of easy-to-review commits.

PR #3205, opened 2023-04-12
fix: chunk get together when random rechunk

### LINKED ISSUE #3195: rechunkRandomly sometimes emits everything in a single chunk
I have five 1-element chunks and a maxFactor of 2.0, but with certain seeds `rechunkRandomly` emits all elements in a single chunk. The number of chunks in the input stream doesn't matter. If you have 10000 1-element chunks, you end up with a single 10000-element chunk.

```scala
val fiveChunks = Stream(1) ++ Stream(2) ++ Stream(3) ++ Stream(4) ++ Stream(5)

fiveChunks.rechunkRandomlyWithSeed(0.1, 2.0)(5).chunks.toList == List(Chunk(1,2,3,4,5))
```
https://scastie.scala-lang.org/TqE3oVzwQ2asDEdW8rf3uA

Other example of a wrong result:
```scala
scala> (Stream(1,1) ++ Stream(2,2) ++ Stream(3,3) ++ Stream(4,4) ++ Stream(5,5)).rechunkRandomlyWithSeed(0.1, 2.0)(9).chunks.toList
val res16: List[fs2.Chunk[Int]] = List(Chunk(1), Chunk(1, 2, 2, 3, 3, 4, 4, 5, 5))
```

yyy1000:
This wants to fix #3195 

Consider the case: if (size < acc.size) 
```scala
val (out, rem) = acc.splitAt(size - 1)
Pull.output(out) >> go(rem ++ hd, -1, tl)
```
The code will continue to append the 'hd' to 'rem', which may be larger than the 'size', to the next 'go' method.
And this will lead to the 'tl' being consumed totally without restricting the size of 'Chunk'.
:)

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2399,7 +2399,7 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     Stream.suspend {
       assert(maxFactor >= minFactor, "maxFactor should be greater or equal to minFactor")
       val random = new scala.util.Random(seed)
-      def factor: Double = random.between(minFactor, maxFactor + 1e-4)
+      def factor: Double = Math.abs(random.nextInt()) % (maxFactor - minFactor) + minFactor + 1e-4
 
       def nextSize(sourceSize: Int): Int = (factor * sourceSize).toInt
 


diesalbla (approved):

core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -1031,6 +1031,13 @@ class StreamSuite extends Fs2Suite {
     assertEquals(countChunks(source), 2)
   }
 
+  test("rechunkRandomly sometimes emits everything in a single chunk") {
+    val fiveChunks = Stream(1) ++ Stream(2) ++ Stream(3) ++ Stream(4) ++ Stream(5)
+
+    val source = fiveChunks.rechunkRandomlyWithSeed(0.1, 2.0)(5).chunks.toList
+    assert(source.forall(_.size <= 2), "Some Chunks have larger size than maxFactor")
```
diesalbla, line 1038: To clarify: what `rechunkRandomly` does is to allow the output to have chunks of size up to `chunkSize * maxFactor`, so if the input were chunks of size 2, it should allow output chunks of size up to 4.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2413,7 +2413,7 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
               Pull.output(acc) >> go(hd, size, tl)
             else {
               val (out, rem) = acc.splitAt(size - 1)
-              Pull.output(out) >> go(rem ++ hd, -1, tl)
+              Pull.output(out) >> go(rem, -1, Pull.output(hd) >> tl)
```
Jasper-M, line 2416: Is `Pull.output(hd) >> tl` equivalent to `s`?

core/shared/src/test/scala/fs2/StreamSuite.scala
yyy1000: Oh, I ignore that in the comment. Thanks for that.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2413,7 +2413,7 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
               Pull.output(acc) >> go(hd, size, tl)
```
diesalbla, line 2413: I think this is another error. The `size` should ideally be the desired size of the next output chunk, but that is the one being emited in `acc`. Instead should be -1, so that the next iteration recomputes a new size based on the first chunk of `tl`.

core/shared/src/main/scala/fs2/Stream.scala
yyy1000: Yes. I should change it to `s`. :) Thanks for that.

core/shared/src/main/scala/fs2/Stream.scala
yyy1000: Yes! I realize it now.

yyy1000:
Uh, it seems that it will lead the time to be longer and fail the pipeline. :(

core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -1031,6 +1031,13 @@ class StreamSuite extends Fs2Suite {
     assertEquals(countChunks(source), 2)
   }
 
+  test("rechunkRandomly sometimes emits everything in a single chunk") {
+    val fiveChunks = Stream(1) ++ Stream(2) ++ Stream(3) ++ Stream(4) ++ Stream(5)
+
+    val source = fiveChunks.rechunkRandomlyWithSeed(0.1, 2.0)(5).chunks.toList
```
Jasper-M, line 1037: Also the way `factor` and `nextSize` are implemented now, I don't think `factor` can ever reach `maxFactor`. So for this example `nextSize` can only be 0 or 1.

yyy1000:
> Thanks! I  replicated it locally but didn't discover the details. I should check it carefully. :)

PR #3208, opened 2023-04-14
Replace recursive `Pull` with `foreach` in `writeWritable`

armanbilge:
Not sure why this was a `Pull`.

### DIFF
--- a/io/js/src/main/scala/fs2/io/ioplatform.scala
+++ b/io/js/src/main/scala/fs2/io/ioplatform.scala
@@ -144,22 +144,16 @@ private[fs2] trait ioplatform {
       Stream
         .eval(writable)
         .flatMap { writable =>
-          def go(
-              s: Stream[F, Byte]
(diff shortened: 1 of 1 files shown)


io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -144,22 +144,16 @@ private[fs2] trait ioplatform {
...
+          val writes = in.chunks.foreach { chunk =>
+            F.async[Unit] { cb =>
+              F.delay {
+                writable.write(
+                  chunk.toUint8Array,
+                  e => cb(e.filterNot(_ == null).toLeft(()).leftMap(js.JavaScriptException))
+                )
+                Some(F.delay(writable.destroy()))
+              }
+            }
```
diesalbla, line 156: Could we extract this lambda as a "writeChunk" function?

diesalbla (approved):

PR #3210, opened 2023-04-15
FS2-IO for JavaScript: refactor ioPlatform

diesalbla:
Refactorings on the WriteWritable function

Proposed in recent PR.

### DIFF
--- a/io/js/src/main/scala/fs2/io/ioplatform.scala
+++ b/io/js/src/main/scala/fs2/io/ioplatform.scala
@@ -140,36 +140,31 @@ private[fs2] trait ioplatform {
       writable: F[Writable],
       endAfterUse: Boolean = true
   )(implicit F: Async[F]): Pipe[F, Byte, Nothing] =
-    in =>
+    (in: Stream[F, Byte]) =>
       Stream
         .eval(writable)
         .flatMap { writable =>
-          val writes = in.chunks.foreach { chunk =>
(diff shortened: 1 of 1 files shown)


io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -140,36 +140,31 @@ private[fs2] trait ioplatform {
...
       endAfterUse: Boolean = true
   )(implicit F: Async[F]): Pipe[F, Byte, Nothing] =
-    in =>
+    (in: Stream[F, Byte]) =>
       Stream
         .eval(writable)
         .flatMap { writable =>
-          val writes = in.chunks.foreach { chunk =>
+          def toEither(e: js.UndefOr[js.Error]): Either[Throwable, Unit] =
+            e.filterNot(_ == null).toLeft(()).leftMap(js.JavaScriptException)
```
armanbilge, line 148: Oh, this is a good idea 👍  I think I repeated this many times. I wonder if we should move this to a `JSErrorOps` (or merge it with `ThrowableOps`) and then use it in all the other places too.

https://github.com/typelevel/fs2/blob/f1fa2d927efb52146561560a3055a1b9cdf306ae/io/js/src/main/scala/fs2/io/internal/ThrowableOps.scala#L26

diesalbla: Perhaps other PR? I saw it in this one, so seemed useful.

PR #3215, opened 2023-04-26
Add `TLSSocket.applicationProtocolOption` and better docs for `TLSSocket.applicationProtocol`

arturaz:
See https://discord.com/channels/632277896739946517/632310980449402880/1100649913223745597

### DIFF
--- a/io/native/src/main/scala/fs2/io/net/tls/S2nConnection.scala
+++ b/io/native/src/main/scala/fs2/io/net/tls/S2nConnection.scala
@@ -51,7 +51,12 @@ private[tls] trait S2nConnection[F[_]] {
 
   def shutdown: F[Unit]
 
-  def applicationProtocol: F[String]
+  /**
+   * Returns [[None]] if there is no protocol being negotiated.
+   *
+   * @see [[https://aws.github.io/s2n-tls/doxygen/s2n_8h.html#ae53faa26669e258afff875d45140f14e]]
+   */
+  def applicationProtocol: F[Option[String]]
 
   def session: F[SSLSession]
 
@@ -199,7 +204,7 @@ private[tls] object S2nConnection {
           .void
 
       def applicationProtocol =
-        F.delay(guard(s2n_get_application_protocol(conn))).map(fromCString(_))
+        F.delay(Option(s2n_get_application_protocol(conn)).map(fromCString(_)))
 
       def session = F.delay {
         val len = guard(s2n_connection_get_session_length(conn))
--- a/io/native/src/main/scala/fs2/io/net/tls/S2nConnection.scala
+++ b/io/native/src/main/scala/fs2/io/net/tls/S2nConnection.scala
@@ -51,11 +51,10 @@ private[tls] trait S2nConnection[F[_]] {
 
   def shutdown: F[Unit]
 
-  /**
-   * Returns [[None]] if there is no protocol being negotiated.
-   *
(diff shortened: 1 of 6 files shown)


io/shared/src/main/scala/fs2/io/net/tls/TLSSocket.scala
```diff
@@ -38,9 +38,19 @@ sealed trait TLSSocket[F[_]] extends Socket[F] with TLSSocketPlatform[F] {
   def session: F[SSLSession]
 
   /** Provides access to the current application protocol that has been negotiated.
+    *
+    * Raises a [[NoSuchElementException]] if it has not yet been determined if application protocols might
+    * be used for this connection. See [[applicationProtocolOption]] for a safer option.
+    *
+    * @see https://discord.com/channels/632277896739946517/632310980449402880/1100649913223745597
     */
   def applicationProtocol: F[String]
```
armanbilge, line 47: I think we can even deprecate this API to encourage the other one :)

Not sure about linking to Discord. We should record anything important either in discussion on this PR or in code / comments themselves.

io/shared/src/main/scala/fs2/io/net/tls/TLSSocket.scala
```diff
@@ -38,9 +38,19 @@ sealed trait TLSSocket[F[_]] extends Socket[F] with TLSSocketPlatform[F] {
...
 
+  /** Provides access to the current application protocol that has been negotiated.
+    *
+    * Returns [[None]] if it has not yet been determined if application protocols might be used for this connection.
```
armanbilge, line 51: Maybe we should include the bit about the empty `String`? Actually, this _almost_ deserves its own ADT but I'm not sure if it's worth it 🤔 or, maybe that case should be mapped to `None` as well? I'm not sure if there is a meaningful distinction.
> null if it has not yet been determined if application protocols might be used for this connection, an empty String if application protocols values will not be used, or a non-empty application protocol String if a value was successfully negotiated.

io/native/src/main/scala/fs2/io/net/tls/S2nConnection.scala
```diff
@@ -51,7 +51,11 @@ private[tls] trait S2nConnection[F[_]] {
 
   def shutdown: F[Unit]
 
-  def applicationProtocol: F[String]
+  /** Returns [[None]] if there is no protocol being negotiated.
+    *
+    * @see [[https://aws.github.io/s2n-tls/doxygen/s2n_8h.html#ae53faa26669e258afff875d45140f14e]]
```
armanbilge, line 56: The doc says:
> In the event of no protocol being negotiated, NULL is returned.

Is this actually equivalent to the empty string `""` case on the JVM? I'm not sure. On the JVM `None` is if it's not yet determined, but `""` is if it's not being used (not being negotiated?).

armanbilge (commented):
If it's not too hard would be good to add some tests for this. It might be straightforward to call `applicationProtocol` in our existing tests and check its value.

io/native/src/main/scala/fs2/io/net/tls/S2nConnection.scala
arturaz: I am even less qualified to answer this question, not familiar with TLS enough :)

io/shared/src/main/scala/fs2/io/net/tls/TLSSocket.scala
arturaz: In my experience linking to Discord gives the perspective on back-and-forth that we had until we arrived at this. It's not super important, but it might give some insight into the context for a person looking at it.

PR #3218, opened 2023-05-07
Add String to Char and Char to String conversion

susuro:
The methods in the `text` package operate on `String`s. This is the optimal choice for most of the use cases. In some scenarios, however, a stream of `Char`acters may be required (at least I needed it). Although the conversion is a one-liner, it is not apparent how to do it at first (at least it wasn't to me). I think these simple conversion methods may be of use to others and nicely fit to the purpose of `text` package.
I'm not fully sure if the proposed naming is correct - maybe `stringToChar` is a preferred one?
As a comment to the reviewer, both methods may be implemented at least one more way:
```Scala
  def string2char[F[_]]: Pipe[F, String, Char] = _.map(s => Chunk.array(s.toCharArray)).flatMap(Stream.chunk)
  def char2string[F[_]]: Pipe[F, Char, String] = _.chunks.map(c => new String(c.toArray))
```
They do not differ in performance from the proposed implementation.

### DIFF
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -22,6 +22,7 @@
 package fs2
 
 import cats.ApplicativeThrow
+import cats.syntax.foldable._
 import java.nio.{Buffer, ByteBuffer, CharBuffer}
 import java.nio.charset.{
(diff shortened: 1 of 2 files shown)


mpilquist:
Thanks!

PR #3229, opened 2023-06-02
Add support for mergePreferred

shagoon:
This PR adds support for [mergePreferred](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/mergePreferred.html) like akka provides.

Internally two bounded CE Queues are used, mainly because they provide `tryTake` for peeking the queue of the preferred stream. If no data is available on the preferred stream, a race is started on both queues. The fibre of the looser will be reused when retrieving the next element.

Testing is kind of hard because race() is not deterministic. Since tests are missing atm, I created this as a draft PR.


shagoon:
I adjusted the test of the behaviour of `mergePreferred` to work without any tweaks of timing (i.e. no `delay`, `metered` or `TestControl.executeEmbed`, etc.). It tolerates some (currently 2%) of the elements of the resulting stream to originate from the non preferred stream. This can happen, if the preferred stream did not yet offer an element to its queue, resulting in a `race()` on both queues, which the non preferred queue might win.

PR #3240, opened 2023-06-14
I/O polling

armanbilge:
This PR is published as FS2 [v3.12.0-RC1](https://github.com/typelevel/fs2/releases/tag/v3.12.0-RC1).

Upgrades to Cats Effect ~~`3.6-e9aeb8c`~~ v3.6.0-RC1 which includes the new I/O polling system.

- https://github.com/typelevel/cats-effect/pull/3332

The following APIs can be implemented with I/O polling. I've done some in this PR, the rest might end up in follow-up PRs.

#### JVM `Selector`-based I/O
- [x] `SocketGroup` and `Socket`
- [ ] `DatagramSocketGroup` and `DatagramSocket`
- [ ] `UnixSockets`

#### Native file descriptor polling-based I/O
- [x]  `SocketGroup` and `Socket`
- [ ] `DatagramSocketGroup` and `DatagramSocket`
- [x]  `UnixSockets`
- [ ] `Processes` and `Process`
- [x] `fs2.io.{stdin,stdout,stderr}`


armanbilge:
Published as `3.8-1580d81` to Maven central.

yilinwei:
Just writing this down here before I forget,

I saw this error (which looks eerily like we're using an `at` before checking), I'm not certain why it's `Unknown Source` unfortunately.

```java.lang.ArrayIndexOutOfBoundsException: 0
	at java.lang.StackTrace$.currentStackTrace$$anonfun$1(Unknown Source)
	at java.lang.StackTrace$$$Lambda$2.applyVoid(Unknown Source)
	at apply @ fs2.io.net.FdPollingSocket.go$1$$anonfun$2(Unknown Source)
	at apply @ fs2.io.net.FdPollingSocket.go$1(Unknown Source)
	at flatMap @ fs2.io.net.FdPollingSocket.go$1(Unknown Source)
	at pollReadRec @ fs2.io.net.FdPollingSocket.read$$anonfun$1(Unknown Source)
	at pollReadRec @ fs2.io.net.FdPollingSocket.read$$anonfun$1(Unknown Source)```

**Comment by armanbilge** [comment]
> I'm not certain why it's `Unknown Source` unfortunately.

All Scala Native stack traces are like this currently. They don't include any source/line information, but they will soon.

**Comment by armanbilge** [comment]
> I saw this error (which looks eerily like we're using an `at` before checking)

Fixed in 500e5453230383ac95e2aa183301eae7e86af278.

**Comment by armanbilge** [comment]
Published `3.10-4b5f50b` to Maven central.

**Comment by armanbilge** [comment]
Published `3.10-365636d` to Maven central.

**Comment by armanbilge** [comment]
This PR is now published as a release candidate.
https://github.com/typelevel/fs2/releases/tag/v3.12.0-RC1

File: build.sbt [hunk 1]
```diff
@@ -299,9 +299,9 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
     libraryDependencies ++= Seq(
       "org.scodec" %%% "scodec-bits" % "1.1.38",
       "org.typelevel" %%% "cats-core" % "2.11.0",
-      "org.typelevel" %%% "cats-effect" % "3.5.7",
-      "org.typelevel" %%% "cats-effect-laws" % "3.5.7" % Test,
-      "org.typelevel" %%% "cats-effect-testkit" % "3.5.7" % Test,
+      "org.typelevel" %%% "cats-effect" % "3.6.0-RC1",
```
mpilquist, line 302: Bump to RC2 before merging?

io/native/src/main/scala/fs2/io/internal/NativeUtil.scala
```diff
@@ -0,0 +1,86 @@
...
+      val e = errno
+      if (mask(e)) rtn
+      else throw errnoToThrowable(e)
+    } else
+      rtn
+  }
+
+  @alwaysinline def errnoToThrowable(e: CInt): Throwable = {
+    val msg = fromCString(strerror(e))
+    if (e == EADDRINUSE /* || e == EADDRNOTAVAIL */ )
```
mpilquist, line 74: Left over?

io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala
```diff
@@ -221,7 +225,7 @@ class SocketSuite extends Fs2IoSuite with SocketSuitePlatform {
         }
     }
 
-    test("read after timed out read not allowed on JVM or Native") {
+    test("read after timed out read not allowed on JVM or Native".ignore) {
```
mpilquist, line 228: Is this not passing?

mpilquist (approved):

io/shared/src/test/scala/fs2/io/net/tcp/SocketSuite.scala
armanbilge: Ah, the new polling implementation actually allows this behavior. A read can timeout without corrupting the socket because we are only canceling the `select`. The old implementation using the JDK's `AsynchronousSocketChannel` doesn't support cancelable reads.

Technically the test is still relevant to the old implementations (which I guess we aren't testing anymore).

mpilquist: Ah nice, that always felt like an annoying restriction. Can we flip this test around to assert this works as expected now then?

armanbilge: Cool, good idea!

io/native/src/main/scala/fs2/io/internal/NativeUtil.scala
armanbilge: > I'll uncomment it and see if anything breaks.

Ahh, it's because this was missing in SN 0.4.x but added in 0.5.x. We can uncomment it soon :)

PR #3242, opened 2023-06-19
Refactor merge implementation

zainab-ali:
To give some context, @djspiewak  and I were looking through `merge` and found the implementation difficult to understand, then tried to rewrite it to use new cats-effect primitives.

I'm not sure that the code I've settled on is actually any better than the old implementation, but think it's worth getting a second opinion on. 

The main differences are:
 - The use of `SingallingRef` over two `Deferred`s for managing the upstream state.
 - The use of `evalMap(guard.aquire)` instead of an explicit pull for pulling from upstream.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1936,12 +1936,10 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   )(implicit F: Concurrent[F2]): Stream[F2, O2] =
     Stream.force {
       // `State` describes the state of an upstream stream (`this` and `that` are both upstream streams)
-      // None            : the stream is has not yet terminated
+      // None            : the stream has not yet terminated
       // Some(Left(t))   : the stream terminated with an error
       // Some(Right(())) : the stream terminated successfully
       type State = Option[Either[Throwable, Unit]]
-      // `bothStates` keeps track of the state of `this` and `that` stream so we can terminate the
-      // downstream when both upstreams terminate.
       for {
         // `bothStates` keeps track of the state of `this` and `that` stream
         // so we can terminate downstream when both upstreams terminate.
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1933,63 +1933,78 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     */
   def merge[F2[x] >: F[x], O2 >: O](
       that: Stream[F2, O2]
-  )(implicit F: Concurrent[F2]): Stream[F2, O2] = {
-    val fstream: F2[Stream[F2, O2]] = for {
-      interrupt <- F.deferred[Unit]
-      resultL <- F.deferred[Either[Throwable, Unit]]
-      resultR <- F.deferred[Either[Throwable, Unit]]
-      otherSideDone <- F.ref[Boolean](false)
-      resultChan <- Channel.unbounded[F2, Stream[F2, O2]]
-    } yield {
+  )(implicit F: Concurrent[F2]): Stream[F2, O2] =
+    Stream.force {
+      // `State` describes the state of an upstream stream (`this` and `that` are both upstream streams)
+      // None            : the stream is has not yet terminated
+      // Some(Left(t))   : the stream terminated with an error
+      // Some(Right(())) : the stream terminated successfully
+      type State = Option[Either[Throwable, Unit]]
+      // `bothStates` keeps track of the state of `this` and `that` stream so we can terminate the
+      // downstream when both upstreams terminate.
+      SignallingRef.of[F2, (State, State)]((None, None)).flatMap { bothStates =>
+        // `output` is used to send chunks from upstreams to downstream. It sends streams, not chunks,
+        // as each chunk has a finalizer associated with it.
+        Channel.synchronous[F2, Stream[F2, O2]].flatMap { output =>
+          // `stopDef` is used to interrupt the streams in the case that an upstream errors, or
+          // in the case that the downstream terminates.
+          Deferred[F2, Unit].map { stopDef =>
+            val signalStop: F2[Unit] = stopDef.complete(()).void
+            val stop: F2[Either[Throwable, Unit]] = stopDef.get.as(Right(()))
+            def complete(result: Either[Throwable, Unit]): F2[Unit] =
+              bothStates.update {
+                case (None, None)  => (Some(result), None)
+                case (other, None) => (other, Some(result))
+                case _             => sys.error("impossible")
+              }
+            val bothStopped: PartialFunction[(State, State), Either[Throwable, Unit]] = {
+              case (Some(r1), Some(r2)) => CompositeFailure.fromResults(r1, r2)
+            }
+            def run(s: Stream[F2, O2]): F2[Unit] =
+              // `guard` ensures we do not pull another chunk until the previous one has been consumed downstream.
+              Semaphore[F2](1).flatMap { guard =>
+                Stream
+                  .exec(guard.acquire)
+                  .append(s)
+                  .chunks
+                  .evalMap(chk =>
+                    output.send(
+                      Stream.chunk(chk).onFinalize(guard.release)
+                    )
+                  )
+                  .evalTap(_ => guard.acquire)
+                  // Stop when the other upstream has errored or the downstream has completed.
+                  .interruptWhen(stop)
+                  .compile
+                  .drain
+                  .attempt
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1933,63 +1933,78 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+                Stream
+                  .exec(guard.acquire)
+                  .append(s)
+                  .chunks
+                  .evalMap(chk =>
+                    output.send(
+                      Stream.chunk(chk).onFinalize(guard.release)
+                    )
+                  )
+                  .evalTap(_ => guard.acquire)
```
diesalbla, line 1975:
```suggestion
                def sendChunk(ch: Chunk[O2]): F2[Unit] = {
                  val outStr = Stream.chunk(chk).onFinalize(guard.release)
                  output.send(outStr) >> guard.acquire
                }
                
                (Stream.exec(guard.acquire) ++ s.chunks.foreach(sendChunk))
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1933,63 +1933,78 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+              .compile
+              .drain
+              .guarantee(output.close.void)
+
+            // There is no need to clean up these fibers. If the downstream is cancelled,
+            // both streams will stop gracefully and the fibers will complete.
+            val setup: F2[Fiber[F2, Throwable, Unit]] =
+              run(this).start >> run(that).start >> waitForBoth.start
+            Stream.bracket(setup)(waitForBoth =>
+              signalStop >> waitForBoth.joinWithUnit
```
diesalbla, line 2002: Avoid name collision.
```suggestion
            Stream.bracket(setup)(wfb => signalStop >> wfb.joinWithUnit)
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1933,63 +1933,78 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+                  .append(s)
+                  .chunks
+                  .evalMap(chk =>
+                    output.send(
+                      Stream.chunk(chk).onFinalize(guard.release)
+                    )
+                  )
+                  .evalTap(_ => guard.acquire)
+                  // Stop when the other upstream has errored or the downstream has completed.
+                  .interruptWhen(stop)
```
diesalbla, line 1977: Would it be worth pointing out that, because the initial `guard.acquire` is at the  beginning of the stream, the `stop` can interrupt it as well?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1933,63 +1933,78 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+      type State = Option[Either[Throwable, Unit]]
+      // `bothStates` keeps track of the state of `this` and `that` stream so we can terminate the
+      // downstream when both upstreams terminate.
+      SignallingRef.of[F2, (State, State)]((None, None)).flatMap { bothStates =>
+        // `output` is used to send chunks from upstreams to downstream. It sends streams, not chunks,
+        // as each chunk has a finalizer associated with it.
+        Channel.synchronous[F2, Stream[F2, O2]].flatMap { output =>
+          // `stopDef` is used to interrupt the streams in the case that an upstream errors, or
+          // in the case that the downstream terminates.
+          Deferred[F2, Unit].map { stopDef =>
```
diesalbla, line 1951: As a matter of syntax, is there any concern about the use of `for` comprehensions? We have used them to reduce code inlining. 
```suggestion
      for {
        // `bothStates` keeps track of the state of `this` and `that` stream 
        // so we can terminate downstream when both upstreams terminate.      
        bothStates <- SignallingRef.of[F2, (State, State)]((None, None))
        // `output` is used to send chunks from upstreams to downstream. 
        // It sends streams, not chunks, to tie each chunk with a finalizer
        output <- Channel.synchronous[F2, Stream[F2, O2]]
        // `stopDef` is used to interrupt the upstreams if a) any of the
        // upstreams raises an error, or b) the downstream terminates.
        stopDef <- Deferred[F2, Unit]
      } yield {
```

zainab-ali: I've no problems with `for` comprehensions whatsoever :+1:

zainab-ali:
The most recent builds failed due to failures in the following tests:
 - `StreamCombinatorsSuite - awakeEvery - basic` on MacOS
 - `UnixSocketsSuite.echoes - jdk` on MacOS

Neither of these failures are related to the `merge` implementation. I assume these tests are flaky and should be marked as such, or corrected?

armanbilge:
> or corrected

most time-based tests can be run on the Cats Effect test runtime with `executeEmbed`, this should fix the flakiness

https://github.com/typelevel/fs2/blob/d7f1b71a8d7badb8a7ef0a8a5468807f9e384358/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala#L42-L45

zainab-ali:
I've used `TestControl` for most viable cases in the `StreamCombinatorsSuite`. I've also added a pretty printer for the scalacheck failures so we can see the elements of a stream on test failure.

Looking at the size of the changeset, it might be worth pulling it out into a different PR - if so, let me know.

core/shared/src/test/scala/fs2/HashSuite.scala
```diff
@@ -55,7 +55,7 @@ class HashSuite extends Fs2Suite with HashSuitePlatform with TestPlatform {
   }
 
   test("empty input") {
-    Stream.empty[IO].through(sha1).compile.count.assertEquals(20L)
+    Stream.empty.covary[IO].through(sha1).compile.count.assertEquals(20L)
```
zainab-ali, line 58: On adding the pretty printer implicit conversion, `Stream.empty` resolved to the `val empty` instead of `def empty[A]` on the monoidK instance. I'm at a loss as to why.

io/js/src/test/scala/fs2/io/IoPlatformSuite.scala
```diff
@@ -108,7 +108,8 @@ class IoPlatformSuite extends Fs2Suite {
     }
 
     Stream
-      .empty[IO]
+      .empty
+      .covary[IO]
```
armanbilge, line 112: Out of curiosity, why this change?

armanbilge: Oh, sorry, I missed https://github.com/typelevel/fs2/pull/3242#discussion_r1241253824. Thanks!

zainab-ali: There's a [compile error in the build](https://github.com/typelevel/fs2/actions/runs/5371040241/jobs/9743532919#step:10:44) (locally too), which I'm still figuring out:
```
[error] /home/runner/work/fs2/fs2/io/js/src/test/scala/fs2/io/IoPlatformSuite.scala:111:13: value empty of type fs2.Stream[fs2.Pure,Nothing] does not take type parameters.
[error]       .empty[IO]
[error]             ^
[error] one error found
```
It's caused by the implicit added for the pretty printer.

I _think_ it's interfering with the resolution of the `monoidKInstance`, that provides the `Stream.empty[IO]` function:
```
  implicit def monoidKInstance[F[_]]: MonoidK[Stream[F, *]] =
    new MonoidK[Stream[F, *]] {
      def empty[A]: Stream[F, A] = Stream.empty
      def combineK[A](x: Stream[F, A], y: Stream[F, A]): Stream[F, A] = x ++ y
    }
```

But I don't know why this would be the case.

zainab-ali: I think I know why: 

`Stream.empty[IO]` uses an `apply`, so is desugared to `Stream.empty.apply[IO]`:
 
The `apply[F[_]]` function is added by `PureOps`:

```scala
  implicit final class PureOps[O](private val self: Stream[Pure, O]) extends AnyVal {
    def apply[F[_]]: Stream[F, O] = covary
}
```
This conflicts with [the `apply` function in the scalacheck pretty printer](https://github.com/typelevel/scalacheck/blob/6df98d611ae6c2cf584596ac18628494c6dfac79/core/shared/src/main/scala/org/scalacheck/util/Pretty.scala#L19).

core/shared/src/test/scala/fs2/HashSuite.scala
zainab-ali: See [the comment below](https://github.com/typelevel/fs2/pull/3242#discussion_r1241270269)

PR #3248, opened 2023-07-03
Stream - Refactor Parjoins

diesalbla:
The four parJoin variants, namely bounded/unbounded and ordered/unordered, delegate to `parEvalMapActionImpl` passing it a `initFork` function. These `initFork` action can be reduced to two cases, depending on whether the parJoin has to preserve order or not. Thus, we can replace the function argument with a boolean flag to `parEvalMapActionImpl`, and write the `initFork` cases inside the latter.


diesalbla:
Thanks @mpilquist

PR #3256, opened 2023-07-11
Fix `TLSSocket` addresses on JS

armanbilge:
h/t @etspaceman

### DIFF
--- a/io/js/src/test/scala/fs2/io/net/tls/TLSSocketSuite.scala
+++ b/io/js/src/test/scala/fs2/io/net/tls/TLSSocketSuite.scala
@@ -360,5 +360,46 @@ class TLSSocketSuite extends TLSSuite {
         .to(Chunk)
         .intercept[SSLException]
     }
+
+    test("get local and remote address") {
+      val setup = for {
+        tlsContext <- Resource.eval(testTlsContext(true))
+        addressAndConnections <- Network[IO].serverResource(Some(ip"127.0.0.1"))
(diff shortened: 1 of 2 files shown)


etspaceman (approved):

io/js/src/test/scala/fs2/io/net/tls/TLSSocketSuite.scala
```diff
@@ -360,5 +360,46 @@ class TLSSocketSuite extends TLSSuite {
...
+              socket.localAddress.product(socket.remoteAddress)
+            }
+
+          serverSocketAddresses.parZip(clientSocketAddresses).map {
+            case ((serverLocal, serverRemote), (clientLocal, clientRemote)) =>
+              assertEquals(clientRemote, serverLocal)
+              assertEquals(clientLocal, serverRemote)
+          }
+        }
+        .compile
```
diesalbla, line 401: I wonder if this test can be written in `Resource`, down from `Stream`?

diesalbla (approved):

io/js/src/test/scala/fs2/io/net/tls/TLSSocketSuite.scala
armanbilge: `serverSocketAddresses` is a `Stream` because `server` is a `Stream` of sockets. So we need a `.compile` somewhere, although it could be sooner rather than later.

PR #3257, opened 2023-07-11
Fix TLS session access on JS

armanbilge:
We had a test, but it wasn't working properly for server-side TLS sockets. So I fixed that.

In the end fussing with `'session'` events was more complicated and didn't seem to work for server sockets anyway. Even though the docs warn against it for TLS v1.3 `getSession()` seems to work fine 🤔 

> Note: `getSession()` works only for TLSv1.2 and below. For TLSv1.3, applications must use the `'session'` event

https://nodejs.org/api/tls.html#tlssocketgetsession

Well, let me double-check that before we merge 😅 

Another one h/t @etspaceman

### DIFF
--- a/io/js/src/main/scala/fs2/io/internal/facade/tls.scala
+++ b/io/js/src/main/scala/fs2/io/internal/facade/tls.scala
@@ -188,6 +188,8 @@ package tls {
 
     def ssl: SSL = js.native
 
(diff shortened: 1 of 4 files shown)


armanbilge:
> Well, let me double-check that before we merge 😅

I feel satisfied this is working correctly. Not the first time Node.js TLS docs have been wrong ...

etspaceman (approved):

PR #3261, opened 2023-07-22
Scope - Avoid lambda and `fold`, use a match

diesalbla:
In a Flight recording from a Http4S milli-benchmark with Gatling and a server) we observed lot of allocations of a lambda located there. 

My _guess_ is be that, by having the `self.id`, the compiler needs to build a new lambda every time, which it cannot remove later.

### DIFF
--- a/core/shared/src/main/scala/fs2/internal/Scope.scala
+++ b/core/shared/src/main/scala/fs2/internal/Scope.scala
@@ -265,7 +265,11 @@ private[fs2] final class Scope[F[_]] private (
           resultChildren <- traverseError[Scope[F]](previous.children, _.close(ec))
           resultResources <- traverseError[ScopedResource[F]](previous.resources, _.release(ec))
           _ <- self.interruptible.map(_.cancelParent).getOrElse(F.unit)
-          _ <- self.parent.fold(F.unit)(_.releaseChildScope(self.id))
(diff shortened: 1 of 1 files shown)


diesalbla:
> Did you try this with http4s to see if it avoided the allocation?

How would the use of `http4s` avoid this allocation? To be clear, these was spotted in one of several flight-recordings I made after this message https://discord.com/channels/632277896739946517/839254871646404638/1131233596959641673 using the application https://github.com/kamilkloch/websocket-benchmark

mpilquist:
I'm asking if, after making this change, we see the allocation go away in the http4s benchmark.

diesalbla:
Ok. The benchmark in question is not a micro-benchmark run with JMH. It is a whole-server application being hit by a client running a Gatling client. It was the monitoring of the server JVM, using the Azul Flight Recorder, where I started to look into the memory allocation profiles, one of which showed this lambda. 

This "benchmark" does include a warmup period, but even without it it is possible to do the recordings only at the second run.

mpilquist:
Yep, understood -- just want to know if this PR does in fact make a material difference.

He-Pin:
I think if it's an allocation  then the compiler or the jvm needs some optimization in this case. The id is passed as an argument.

PR #3264, opened 2023-07-27
Add variant of `Stream.fromQueueUnterminated` for Monad.

kamilkloch:
Repo with the benchmark: https://github.com/kamilkloch/fs2-tests

Alternative implementation of `Stream.fromQueueUnterminated`, replacing
```scala
  private def fromQueueNoneTerminatedSingletons_[F[_], A](
      take: F[Option[A]],
      tryTake: F[Option[Option[A]]],
      limit: Int
  ): Stream[F, A] = {
    def await: Stream[F, A] =
      Stream.eval(take).flatMap {
        case None => Stream.empty
        case Some(c) =>
          val builder = collection.mutable.Buffer.newBuilder[A]
          builder += c
          pump(1, builder)
      }
    def pump(
        currSize: Int,
        acc: collection.mutable.Builder[A, collection.mutable.Buffer[A]]
    ): Stream[F, A] =
      if (currSize == limit) Stream.emits(acc.result()) ++ await
      else
        Stream.eval(tryTake).flatMap {
          case None => Stream.emits(acc.result()) ++ await
          case Some(Some(c)) =>
            acc += c
            pump(currSize + 1, acc)
          case Some(None) => Stream.emits(acc.result())
        }
    await
  }
```

with 
```scala
  def fromUnterminatedQueue[A](q: QueueSource[IO, A]): Stream[IO, A] = {
    /** First, try non-blocking batch dequeue.
     *   Only if the result is an empty list, semantically block and get exactly one element. 
     */
    val asf = q.tryTakeN(None).flatMap {
      case Nil => q.take.map(_ :: Nil)
      case as => IO.pure(as)
    }

    Stream.evalSeq(asf).repeat
  }
```

## Test setup

Producer feeds `cats.effect.std.Queue` with 100k timestamps, consumer ingests the queue as an `fs2.Stream` and runs it for effects.
100 concurrent instances of (producer, consumer) pairs. 

## Results

`Stream.fromQueueUnterminated` is almost 80% slower (4s vs 7s, and this only after a long warmup), 
puts more pressure on ZGC (despite pre-allocating and pre-touching 32G of heap).

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3529,16 +3529,37 @@ object Stream extends StreamLowPriority {
   )(implicit F: Functor[F]): Stream[F, A] =
     F match {
       case f0: Monad[F] =>
-        val someLimit = Some(limit)
+        if (limit > 1) {
 
-        /** use non-blocking tryTakeN, which is possibly more performant than n * take */
-        Stream
-          .eval(queue.tryTakeN(someLimit)(f0))
-          .flatMap {
-            case Nil => Stream.eval(queue.take)
-            case as  => Stream.emits(as)
+          /** use non-blocking tryTakeN, which is possibly more performant than n * take */
+
+          val someLimit = Some(limit)
+          val someLimitLess1 = Some(limit - 1)
+
+          /** First, try non-blocking batch dequeue.
+            * Only if the result is an empty list, semantically block to get one element,
+            * then attempt 2nd tryTakeN to get any other elements that are immediately available.
+            */
+          val asf = f0.flatMap(queue.tryTakeN(someLimit)(f0)) {
+            case Nil => f0.flatMap(queue.take)(h => queue.tryTakeN(someLimitLess1)(f0).map(h :: _))
+            case as  => f0.pure(as)
           }
-          .repeat
+
+          Stream.evalSeq(asf).repeat
+
+        } else {
+          val someOne = Some(1)
+
+          /** First, try non-blocking batch dequeue.
+            * Only if the result is an empty list, semantically block to get exactly one element.
+            */
+          val asf = f0.flatMap(queue.tryTakeN(someOne)(f0)) {
+            case Nil => queue.take
+            case as  => f0.pure(as.head)
+          }
+
+          Stream.eval(asf).repeat
+        }
 
       case _ =>
         fromQueueNoneTerminatedSingletons_[F, A](
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3528,24 +3528,17 @@ object Stream extends StreamLowPriority {
       limit: Int = Int.MaxValue
   )(implicit F: Functor[F]): Stream[F, A] =
     F match {
-      case f0: Async[F] =>
-        /** specialized case for BoundedAsyncQueue / UnboundedAsyncQueue,
-          * which both have an efficient implementation of tryTakeN.
-          * See implementation of [[Queue.bounded]] / [[Queue.bounded]]
-          */
-
-        implicit val F: Async[F] = f0
+      case f0: Monad[F] =>
         val someLimit = Some(limit)
 
-        /** First, try non-blocking batch dequeue.
-          *  Only if the result is an empty list, semantically block and get exactly one element.
-          */
-        val asf = queue.tryTakeN(someLimit).flatMap {
-          case Nil => queue.take.map(_ :: Nil)
-          case as  => F.pure(as)
-        }
-
-        Stream.evalSeq(asf).repeat
+        /** use non-blocking tryTakeN, which is possibly more performant than n * take */
+        Stream
+          .eval(queue.tryTakeN(someLimit)(f0))
+          .flatMap {
+            case Nil => Stream.eval(queue.take)
+            case as  => Stream.emits(as)
+          }
+          .repeat
 
       case _ =>
         fromQueueNoneTerminatedSingletons_[F, A](
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3538,8 +3538,8 @@ object Stream extends StreamLowPriority {
         val someLimit = Some(limit)
 
         /** First, try non-blocking batch dequeue.
-           *  Only if the result is an empty list, semantically block and get exactly one element. 
-           */
+          *  Only if the result is an empty list, semantically block and get exactly one element.
+          */
         val asf = queue.tryTakeN(someLimit).flatMap {
           case Nil => queue.take.map(_ :: Nil)
           case as  => F.pure(as)
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3529,6 +3529,10 @@ object Stream extends StreamLowPriority {
   )(implicit F: Functor[F]): Stream[F, A] = {
     F match {
       case f0: Async[F] =>
+        /** specialized case for BoundedAsyncQueue / UnboundedAsyncQueue,
+         * which both have an efficient implementation of tryTakeN.
+         * See implementation of [[Queue.bounded]] / [[Queue.bounded]]
+         */
         Stream.evalSeq(queue.tryTakeN(None)(f0)).repeat
 
       case _ =>
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -3526,14 +3526,26 @@ object Stream extends StreamLowPriority {
   def fromQueueUnterminated[F[_], A](
       queue: QueueSource[F, A],
       limit: Int = Int.MaxValue
-  )(implicit F: Functor[F]): Stream[F, A] = {
+  )(implicit F: Functor[F]): Stream[F, A] =
     F match {
       case f0: Async[F] =>
         /** specialized case for BoundedAsyncQueue / UnboundedAsyncQueue,
-         * which both have an efficient implementation of tryTakeN.
-         * See implementation of [[Queue.bounded]] / [[Queue.bounded]]
-         */
-        Stream.evalSeq(queue.tryTakeN(None)(f0)).repeat
+          * which both have an efficient implementation of tryTakeN.
+          * See implementation of [[Queue.bounded]] / [[Queue.bounded]]
+          */
+
+        implicit val F: Async[F] = f0
+        val someLimit = Some(limit)
+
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3523,15 +3523,22 @@ object Stream extends StreamLowPriority {
...
-  ): Stream[F, A] =
-    fromQueueNoneTerminatedSingletons_[F, A](
-      queue.take.map(a => Some(a)),
-      queue.tryTake.map(_.map(a => Some(a))),
-      limit
-    )
+  )(implicit F: Functor[F]): Stream[F, A] = {
+    F match {
+      case f0: Async[F] =>
+        Stream.evalSeq(queue.tryTakeN(None)(f0)).repeat
```
armanbilge, line 3532: Isn't this busy-polling? Or, I miss the part where it blocks if the queue is empty.

diesalbla: I guess the `takeN` adds the semantics of "if queue is empty block the taker fiber until there are any elements. 

Not sure what you mean by "no chance of success". The problem is, this method is public and therefore probably used by applications, which may be relying on the blocking semantics of this method.

However, it is still possible to add this or other elements, _as a different method_.

kamilkloch: Busy polling is now hopefully removed.

diesalbla (approved):

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3523,15 +3523,37 @@ object Stream extends StreamLowPriority {
...
       limit: Int = Int.MaxValue
-  ): Stream[F, A] =
-    fromQueueNoneTerminatedSingletons_[F, A](
-      queue.take.map(a => Some(a)),
-      queue.tryTake.map(_.map(a => Some(a))),
-      limit
-    )
+  )(implicit F: Functor[F]): Stream[F, A] =
+    F match {
+      case f0: Async[F] =>
```
armanbilge, line 3531: We don't need `Async` here. In fact, I propose we replace this entire implementation here with:

```scala
Stream.eval(queue.tryTakeN(someLimit)).flatMap {
  case Nil => Stream.eval(queue.take)
  case as => Stream.emits(as)
}.repeat
```

armanbilge (changes requested):

mpilquist:
Doesn't this have a change in semantics when the resulting stream is terminated early? E.g., consider case of `tryTakeN` returning 10 elements when used with `.take(1)`. The 9 unconsumed elements are lost.

armanbilge:
> Doesn't this have a change in semantics when the resulting stream is terminated early?

No, I don't think so. I believe the existing implementation has the exact same semantic. We can add a test to confirm this.

https://github.com/typelevel/fs2/blob/14ecd858b89b29b438dcea73671bc24c4e314573/core/shared/src/main/scala/fs2/Stream.scala#L3598-L3625

core/shared/src/main/scala/fs2/Stream.scala
kamilkloch: I believe this will drain the queue indefinitely, as `someLimit` will not reflect the already pulled elements.

armanbilge: How is the existing logic different?
https://github.com/typelevel/fs2/blob/8cf9273bf9f99f32f7989724ca4564a6b8f3ac03/core/shared/src/main/scala/fs2/Stream.scala#L3540-L3548

kamilkloch: > We don't need `Async` here. In fact, I propose we replace this entire implementation here with:

Not really, `tryTakeN` expects at least a `Monad`:
```scala
def tryTakeN(maxN: Option[Int])(implicit F: Monad[F])
```

armanbilge: Ok, so we need `Monad`, not `Async` :)

kamilkloch: > How is the existing logic different?

My bad, same same. 

The rationale behind `Async` was that only a bounded async queue sports an optimized `tryTakeN`. I like your proposition, (as it simplifies the code substantially), just to make sure I might benchmark it against current solution, which uses `acc: collection.mutable.Builder[A, collection.mutable.Buffer[A]]`, as opposed to a `List`.

armanbilge: > The rationale behind `Async` was that only a bounded async queue sports an optimized `tryTakeN`

True, but checking if `F` is `Async` actually doesn't tell us if the `Queue` is async or not, we don't really have a good way to check that.

I think it is better to use the most optimal method in the `Queue` API. The onus should be on the `Queue` implementation to be fast.

kamilkloch: I updated the code. 

If the signature 
```scala
  def fromQueueUnterminated[F[_], A](
      queue: QueueSource[F, A],
      limit: Int = Int.MaxValue
  )(implicit F: Functor[F])
```
must not change to 
```scala
  def fromQueueUnterminated[F[_], A](
      queue: QueueSource[F, A],
      limit: Int = Int.MaxValue
  )(implicit F: Monad[F])
```
(source/bin compatibility?), the old implementation has to stay and we need to pattern-match on `F`?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3523,15 +3523,30 @@ object Stream extends StreamLowPriority {
...
       limit: Int = Int.MaxValue
-  ): Stream[F, A] =
-    fromQueueNoneTerminatedSingletons_[F, A](
-      queue.take.map(a => Some(a)),
-      queue.tryTake.map(_.map(a => Some(a))),
-      limit
-    )
+  )(implicit F: Functor[F]): Stream[F, A] =
+    F match {
+      case f0: Monad[F] =>
```
armanbilge, line 3531: Yes, you were right: we do need to keep the pattern match. I forgot that `tryTakeN` requires `Monad[F]` because we added that method after-the-fact ...

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -3523,15 +3523,30 @@ object Stream extends StreamLowPriority {
...
+        val someLimit = Some(limit)
+
+        /** use non-blocking tryTakeN, which is possibly more performant than n * take */
+        Stream
+          .eval(queue.tryTakeN(someLimit)(f0))
+          .flatMap {
+            case Nil => Stream.eval(queue.take)
+            case as  => Stream.emits(as)
+          }
+          .repeat
```
armanbilge, line 3541: Sorry, now that I realize we do need `Monad[F]` for `tryTakeN`, actually it's more efficient to use `flatMap` on `F[_]` than `Stream`. So basically what you originally proposed, sorry.

We also need to add one more feature to preserve the original semantics: after a `take` we should attempt to `tryTakeN` to get any other elements that are immediately available.

```suggestion
        /** use non-blocking tryTakeN, which is possibly more performant than n * take */
        if (limit > 1) {
          val someLimit = Some(limit)
          val someLimitLess1 = Some(limit - 1)
          Stream.evalSeq {
            queue.tryTakeN(someLimit)(f0).flatMap {
              case Nil => queue.take.map2(queue.tryTakeN(someLimitLess1)(f0)))(_ :: _)
              case as => F.pure(as)
            }
          }.repeat
        } else Stream.repeatEval(queue.take)
```

armanbilge (commented):
Sorry, another round of revisions. Thanks for opening this PR, we'd totally missed this optimization.

core/shared/src/main/scala/fs2/Stream.scala
kamilkloch: It weren't the problem if we used a conrete `IO` instead of TF, which would arguably make the signatures more comprehensible without any loss of power for 99.9% of use-cases ;) But that is a personal opinion, not relevant here :)

core/shared/src/main/scala/fs2/Stream.scala
kamilkloch: `map2` goes through an extra step of tupling, please review the direct version with `flatMap` + `map`.
Also, we can still do `tryTakeN` even for `n == 1`.

armanbilge: > `map2` goes through an extra step of tupling

where is the tupling?

> Also, we can still do `tryTakeN` even for `n == 1`.

What is the point of `tryTakeN` for `n == 1`? It's just adding an extra and unnecessary step. When you `take` you will either get the element right away or have to wait. What is `tryTakeN` bringing to the table here?

kamilkloch: > > `map2` goes through an extra step of tupling
> 
> where is the tupling?

```scala
  def map2[A, B, Z](fa: F[A], fb: F[B])(f: (A, B) => Z): F[Z] =
    map(product(fa, fb))(f.tupled)
```

> 
> > Also, we can still do `tryTakeN` even for `n == 1`.
> 
> What is the point of `tryTakeN` for `n == 1`? It's just adding an extra and unnecessary step. When you `take` you will either get the element right away or have to wait. What is `tryTakeN` bringing to the table here?

I looked at the internals of a generic `take` vs `tryTakeN` for the async queue and got the notion that the latter might be actually more performant. If I am wrong, we shall revert to your proposition.

armanbilge: 1. That's not the final implementation used. Please see the override in `FlatMap` which does not tuple.

https://github.com/typelevel/cats/blob/7e04753b20a21f6f0e35eb2e6db1a0fb4ae512d2/core/src/main/scala/cats/FlatMap.scala#L113-L114

2. Hmm, I'm getting that notion as well. But I'm not sure why it's that way and I think we can fix it in Cats Effect.

armanbilge: > I think we can fix it in Cats Effect.

Nope, it's needed for correctness. I tried moving the `uncancelable` inside the `catch` but that allows an element to be lost on cancelation.

https://github.com/typelevel/cats-effect/blob/25c948b065de0c7c77c9655162cdeb093f0b2b3b/std/shared/src/main/scala/cats/effect/std/Queue.scala#L641-L643

Ok, so for the record that `uncancelable` in `take` is adding some allocation overhead over `tryTakeN`. However I still think the simpler implementation in terms of `take` is better here. At the end of the day limiting the chunk size `n == 1` is sub-optimal and I would expect the per-element overhead of `Stream` machinery to dwarf the `uncancelable`. I'd be surprised if a benchmark measures a difference.

kamilkloch: Please see the revised version.

armanbilge (approved):
Thank you for iterating!

PR #3265, opened 2023-07-28
Add `Stream.filterNot`

### LINKED ISSUE #3224: Stream.filterNot does not exist but still referenced from Scaladocs
There is a couple of references to `filterNot` in Scaladocs:
https://github.com/typelevel/fs2/blob/2efbedeb8316bf7b83f01964733089df7a883291/core/shared/src/main/scala/fs2/Stream.scala#L1112C31-L1123

But there's no the `filterNot` function per se.

I think it would be fair either to add `filterNot` method or to rephrase the docs and hence to avoid any confusion about it.
Not sure which is better though.

Also, it was mentioned here:
https://github.com/typelevel/fs2/pull/1635#issuecomment-538380721

kiendang:
Close #3224.

Ran into this today. I'm not too familiar with the internal implementation of fs2 though so not sure if this is the most optimized way to do it.

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -120,6 +120,9 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
     Chunk.array(b.result()).asInstanceOf[Chunk[O]]
   }
 
+  /** Returns a chunk that has only the elements that do not satisfy the supplied predicate. */
+  def filterNot(p: O => Boolean): Chunk[O] = filter(p.andThen(!_))
+
   /** Returns the first element for which the predicate returns true or `None` if no elements satisfy the predicate. */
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -120,6 +120,9 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
     Chunk.array(b.result()).asInstanceOf[Chunk[O]]
   }
 
+  /** Returns a chunk that has only the elements that do not satisfy the supplied predicate. */
+  def filterNot(p: O => Boolean): Chunk[O] = filter(p.andThen(!_))
```
satorg, line 124: Thank you taking on this! A minor note, wouldn't it be simpler and neater?
```suggestion
  def filterNot(p: O => Boolean): Chunk[O] = filter(!p(_))
```

PR #3266, opened 2023-07-29
Hold microptimisation: write all items in chunk.

diesalbla:
Use a chunk `traverse_` rather than an `evalMap`, since the stream is discarded drained.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1541,7 +1541,8 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       initial: O2
   ): Stream[F2, Signal[F2, O2]] =
     Stream.eval(SignallingRef.of[F2, O2](initial)).flatMap { sig =>
-      Stream(sig).concurrently(evalMap(sig.set))
+      def setAll(ch: Chunk[O]): Pull[F2, Nothing, Unit] = Pull.eval(ch.traverse_(sig.set))
+      Stream(sig).concurrently(underlying.unconsFlatMap(setAll).streamNoScope)
     }
 
   /** Like [[hold]] but does not require an initial value, and hence all output elements are wrapped in `Some`. */


Jasper-M:
This has probably been discussed before, but is there any point in not just setting the signal to the last element of the chunk and discarding the rest?

diesalbla:
@Jasper-M It was discussed before, I think it is somewhere in Discord...

Jasper-M:
I thought so too, but can't find it anymore. Also still can't see the point 😶

diesalbla:
@SystemFw I seem to recall we discussed the semantics of `hold` at some point, and that it mattered that _all_ the updates be recorded to the signal, even if they were so close to each other than only the last one should be expected to last for longer.

armanbilge:
Discord conversation is here.

https://discord.com/channels/632277896739946517/839257183782436945/1071509620415271093

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1541,7 +1541,8 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
       initial: O2
   ): Stream[F2, Signal[F2, O2]] =
     Stream.eval(SignallingRef.of[F2, O2](initial)).flatMap { sig =>
-      Stream(sig).concurrently(evalMap(sig.set))
+      def setAll(ch: Chunk[O]): Pull[F2, Nothing, Unit] = Pull.eval(ch.traverse_(sig.set))
+      Stream(sig).concurrently(underlying.unconsFlatMap(setAll).streamNoScope)
```
armanbilge, line 1545: I'd understand if we replaced `evalMap(sig.set)` with `foreach(sig.set)`.

Otherwise, aren't we essentially re-implementing `foreach` here?

diesalbla: No. If you look at the code of `foreach`, you can see that it wraps each execution of the effect function `f` with a `Pull.eval`, whereas here we are using a `traverse_` on the entire chunk.

armanbilge: So why can't we make this optimization to `foreach` ?

diesalbla: Oh, you mean why not do the `traverse_` directly in the body of `foreach`? I have tried a few times, but it does not work. There are some problems with the interruption scope depending on the actions.


However this code, where we only do a `sig.set`, should not have that problem.

PR #3269, opened 2023-08-02
Handle `AggregateError`s on Node.js

armanbilge:
Node.js 20 appears to have started throwing [`AggregateError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError)s in some situations (similar to FS2's `CompositeFailure`) e.g. see https://github.com/typelevel/skunk/pull/935#issue-1833175626. This confounds the `js.Error => java.io.IOException` translation. So now we specially handle `AggregateError` as part of that translation.

### DIFF
--- a/io/js/src/main/scala/fs2/io/internal/facade/AggregateError.scala
+++ b/io/js/src/main/scala/fs2/io/internal/facade/AggregateError.scala
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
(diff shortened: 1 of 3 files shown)


io/js/src/main/scala/fs2/io/internal/facade/AggregateError.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package fs2.io.internal.facade
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSGlobal
+
+@JSGlobal
+@js.native
+private[io] class AggregateError extends js.Error {
```
armanbilge, line 29: Temporary:
- https://github.com/scala-js/scala-js/issues/4888#issuecomment-1662693830

PR #3273, opened 2023-08-13
fix comment about followLinks

nikiforo:
The fixed comment contradicts with the [comment](https://github.com/typelevel/fs2/compare/main...nikiforo:fix-comment?expand=1#diff-9ec1ae212cbd3462c65322379f67496c8c4274971e5f2e4fe3033c42a69e2cf0R175) above it.
Also, all other comments about methods with explicit `followLinks` attribute in `Files.scala` are used without negation.
Moreover, in sources `followLinks` means `stat`, which follows link.


armanbilge (approved):
thanks, good catch!

PR #3274, opened 2023-08-13
Chunk improvements

### LINKED ISSUE #3271: Provide a asSeq facade for Chunk
Sometimes when interoperating all you want is to be able to pass a `Chunk` to a method that expects a `Seq` without having to actually copy data. The main argument is that such a method will iterate the sequence once to produce something else.

On top of this, it may also be good to provide an `asJava` facade as well, returning `ju.List`.
Technically speaking, `asSeq` would allow us to also call the `asJava` the stdlib already provides, but that would be a facade over a facade; not the end of the world but it may be a good tiny optimization.

BalmungSan:
Fixes #3271

-----

## Improvements
  - [x] `Chunk.asJava`.
  - [x] `Chunk.asSeq`.
  - [x] More `Chunk` operations like `contains` & `exists` _(also overridden on the `Foldable` instance)_.
  - [x] Add `Chunk.javaList` to create chunks based on a **Java** `ju.List` .
  - [x] Add `Chunk.from` to accept any stdlib collection, and deprecate specific factories.
  - [x] Improve `Chunk.Queue` `:+`, `+:`, and `++` methods.

----

As always, I look forward to your feedback. Especially related to naming, scaladoc, and tests.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,3 +1,3 @@
--J-Xms4g
--J-Xmx8g
--J-XX:MaxMetaspaceSize=1024m
+-J-Xms2g
+-J-Xmx4g
+-J-XX:MaxMetaspaceSize=512m
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,3 +1,3 @@
--J-Xms2g
--J-Xmx4g
--J-XX:MaxMetaspaceSize=512m
+-J-Xms4g
+-J-Xmx8g
+-J-XX:MaxMetaspaceSize=1024m
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -1,3 +1,24 @@
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
 package fs2
 
 import scala.collection.immutable.ArraySeq
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -21,7 +21,7 @@
 
 package fs2
 
-import scala.collection.{Factory, SeqFactory}
+import scala.collection.{Iterable => GIterable, Factory, SeqFactory}
 import scala.collection.immutable.ArraySeq
 import scala.reflect.ClassTag
 
@@ -81,18 +81,26 @@ private[fs2] trait ChunkAsSeq213And3Compat[+O] {
 private[fs2] trait ChunkCompanion213And3Compat {
   self: Chunk.type =>
 
-  protected def platformIterable[O](i: Iterable[O]): Option[Chunk[O]] =
+  protected def platformFrom[O](i: GIterable[O]): Option[Chunk[O]] =
     i match {
-      case a: ArraySeq[O]   => Some(arraySeq(a))
-      case w: ChunkAsSeq[O] => Some(w.chunk)
-      case _                => None
+      case arraySeq: ArraySeq[O] =>
+        val arr = arraySeq.unsafeArray.asInstanceOf[Array[O]]
+        Some(array(arr)(ClassTag[O](arr.getClass.getComponentType)))
+
+      case w: ChunkAsSeq[O] =>
+        Some(w.chunk)
+
+      case _ =>
+        None
     }
 
   /** Creates a chunk backed by an immutable `ArraySeq`. */
-  def arraySeq[O](arraySeq: ArraySeq[O]): Chunk[O] = {
-    val arr = arraySeq.unsafeArray.asInstanceOf[Array[O]]
-    array(arr)(ClassTag[O](arr.getClass.getComponentType))
-  }
+  @deprecated(
+    "Use the `from` general factory instead",
+    "3.8.1"
+  )
+  def arraySeq[O](arraySeq: ArraySeq[O]): Chunk[O] =
+    from(arraySeq)
 
   /** Creates a chunk from a `scala.collection.IterableOnce`. */
   def iterableOnce[O](i: IterableOnce[O]): Chunk[O] =
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -120,7 +120,7 @@ private[fs2] final class ChunkAsSeq[+O](
     chunk.toVector
 
   override def toString: String =
-    chunk.toString
+    chunk.iterator.mkString("ChunkAsSeq(", ", ", ")")
 
   override def zipWithIndex: Seq[(O, Int)] =
     new ChunkAsSeq(chunk.zipWithIndex)
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -70,6 +70,9 @@ private[fs2] final class ChunkAsSeq[+O](
   override def isEmpty: Boolean =
     chunk.isEmpty
 
+  override def reverseIterator: Iterator[O] =
+    chunk.reverseIterator
+
   override def foreach[U](f: O => U): Unit =
     chunk.foreach { o => f(o); () }
 
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -57,13 +57,13 @@ private[fs2] trait ChunkAsSeq213And3Compat[+O] {
     math.min(len, xs.length - start)
   }
 
-  override def map[O2](f: O => O2): Seq[O2] =
+  override def map[O2](f: O => O2): IndexedSeq[O2] =
     new ChunkAsSeq(chunk.map(f))
 
-  override def zipWithIndex: Seq[(O, Int)] =
+  override def zipWithIndex: IndexedSeq[(O, Int)] =
     new ChunkAsSeq(chunk.zipWithIndex)
 
-  override def tapEach[U](f: O => U): Seq[O] = {
+  override def tapEach[U](f: O => U): IndexedSeq[O] = {
     chunk.foreach { o => f(o); () }
     this
   }
@@ -71,10 +71,10 @@ private[fs2] trait ChunkAsSeq213And3Compat[+O] {
   override def to[C1](factory: Factory[O, C1]): C1 =
     chunk.to(factory)
 
-  override val iterableFactory: SeqFactory[Seq] =
+  override val iterableFactory: SeqFactory[IndexedSeq] =
     ArraySeq.untagged
 
-  override val empty: Seq[O] =
+  override val empty: IndexedSeq[O] =
     new ChunkAsSeq(Chunk.empty)
 }
 
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -46,6 +46,7 @@ private[fs2] trait Chunk213And3Compat[+O] { self: Chunk[O] =>
   /** Views this Chunk as a Scala immutable Seq.
     * Contrary to all methods that start with _"to"_ (e.g. {{toVector}}, {{toArray}}),
     * this method does not copy data.
+    * As such, this method is mostly intended for `foreach` kind of interop.
     */
   def asSeq: Seq[O] =
     new ChunkAsSeq(this)
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -87,9 +87,6 @@ private[fs2] trait ChunkCompanion213And3Compat {
         val arr = arraySeq.unsafeArray.asInstanceOf[Array[O]]
         Some(array(arr)(ClassTag[O](arr.getClass.getComponentType)))
 
-      case w: ChunkAsSeq[O] =>
-        Some(w.chunk)
-
       case _ =>
         None
     }
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -51,7 +51,9 @@ private[fs2] trait Chunk213And3Compat[+O] { self: Chunk[O] =>
     new ChunkAsSeq(this)
 }
 
-private[fs2] final class ChunkAsSeq[+O](chunk: Chunk[O]) extends immutable.Seq[O] {
+private[fs2] final class ChunkAsSeq[+O](chunk: Chunk[O])
+    extends immutable.Seq[O]
+    with Serializable {
   override def iterator: Iterator[O] =
     chunk.iterator
 
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -74,7 +74,7 @@ private[fs2] trait ChunkAsSeq213And3Compat[+O] {
   override val iterableFactory: SeqFactory[IndexedSeq] =
     ArraySeq.untagged
 
-  override val empty: IndexedSeq[O] =
+  override lazy val empty: IndexedSeq[O] =
     new ChunkAsSeq(Chunk.empty)
 }
 
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -51,7 +51,10 @@ private[fs2] trait Chunk213And3Compat[+O] { self: Chunk[O] =>
     new ChunkAsSeq(this)
 }
 
-private[fs2] final class ChunkAsSeq[+O](chunk: Chunk[O]) extends Seq[O] with Serializable {
+private[fs2] final class ChunkAsSeq[+O](
+    private val chunk: Chunk[O]
+) extends Seq[O]
+    with Serializable {
   override def iterator: Iterator[O] =
     chunk.iterator
 
@@ -61,6 +64,12 @@ private[fs2] final class ChunkAsSeq[+O](chunk: Chunk[O]) extends Seq[O] with Ser
   override def length: Int =
     chunk.size
 
+  override def knownSize: Int =
+    chunk.size
+
+  override def isEmpty: Boolean =
+    chunk.isEmpty
+
   override def foreach[U](f: O => U): Unit =
     chunk.foreach { o => f(o); () }
 
@@ -69,10 +78,7 @@ private[fs2] final class ChunkAsSeq[+O](chunk: Chunk[O]) extends Seq[O] with Ser
     this
   }
 
-  override def isEmpty: Boolean =
-    chunk.isEmpty
-
-  override def copyToArray[B >: O](xs: Array[B], start: Int, len: Int): Int = {
+  override def copyToArray[O2 >: O](xs: Array[O2], start: Int, len: Int): Int = {
     chunk.take(len).copyToArray(xs, start)
     math.min(len, xs.length - start)
   }
@@ -100,7 +106,7 @@ private[fs2] final class ChunkAsSeq[+O](chunk: Chunk[O]) extends Seq[O] with Ser
   override def to[C1](factory: Factory[O, C1]): C1 =
     chunk.to(factory)
 
-  override def toArray[B >: O: ClassTag]: Array[B] =
+  override def toArray[O2 >: O: ClassTag]: Array[O2] =
     chunk.toArray
 
   override def toList: List[O] =
@@ -114,6 +120,15 @@ private[fs2] final class ChunkAsSeq[+O](chunk: Chunk[O]) extends Seq[O] with Ser
 
   override def zipWithIndex: Seq[(O, Int)] =
     new ChunkAsSeq(chunk.zipWithIndex)
+
+  override def equals(that: Any): Boolean =
+    that match {
+      case thatChunkWrapper: ChunkAsSeq[_] =>
+        chunk == thatChunkWrapper.chunk
+
+      case _ =>
+        false
+    }
 }
 
 private[fs2] trait ChunkCompanion213And3Compat { self: Chunk.type =>
--- a/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
+++ b/core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
@@ -49,7 +49,19 @@ private[fs2] trait Chunk213And3Compat[+O] { self: Chunk[O] =>
     * As such, this method is mostly intended for `foreach` kind of interop.
     */
   def asSeq: Seq[O] =
-    new ChunkAsSeq(this)
+    this match {
+      case indexedSeqChunk: Chunk.IndexedSeqChunk[_] =>
+        indexedSeqChunk.s match {
+          case seq: Seq[O] =>
+            seq
+
+          case _ =>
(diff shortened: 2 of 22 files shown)


core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
```diff
@@ -21,9 +21,28 @@ private[fs2] trait Chunk213And3Compat[+O] { self: Chunk[O] =>
...
+
+  /** Views this Chunk as a Scala immutable Seq.
+    * Contrary to all methods that start with _"to"_ (e.g. {{toVector}}, {{toArray}}),
+    * this method does not copy data.
+    */
+  def asSeq: Seq[O] =
+    new ChunkAsSeq(this)
+}
+
+private[fs2] final class ChunkAsSeq[+O](chunk: Chunk[O]) extends immutable.Seq[O] {
```
BalmungSan, line 33: Those three are the only abstract methods on `Seq` thus, the only ones we need to override.
However, I wonder if we should override other ones for performance's sake, things like `copyToArray` ring my bell, but not sure what you all think.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
```diff
@@ -21,9 +21,28 @@ private[fs2] trait Chunk213And3Compat[+O] { self: Chunk[O] =>
     }
     buf.result()
   }
+
+  /** Views this Chunk as a Scala immutable Seq.
+    * Contrary to all methods that start with _"to"_ (e.g. {{toVector}}, {{toArray}}),
+    * this method does not copy data.
+    */
```
BalmungSan, line 49: I think, not totally sure, that my **Scaladoc** formatting is wrong here.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
armanbilge: Yep, I'm not very familiar with Scala collections API but `copyToArray` (and maybe `foreach`) seem useful.

satorg: `knownSize` by default returns -1, but I guess that at least for a compacted `Chunk` it would be really helpful to return its real size.

satorg: Just to clarify: would it make sense to inherit it from `immutable.IndexedSeq` instead of just `Seq`? I wonder because according to the docs, `Chunk` is
> Immutable, strict, finite sequence of values that supports _efficient index-based random access_ of elements

So to me it looks like the `IndexedSeq` matches the idea of "efficient index-based random access" pretty well.

armanbilge: Not sure.

https://github.com/typelevel/fs2/blob/3c93811b7f8d367eb23b02456cd6b18dbaec9962/core/shared/src/main/scala/fs2/Stream.scala#L4468-L4470

satorg: I see. But in that case the doc's header looks quite misleading. I feel it would be better to re-phrase it to emphasize that it may or may not support efficient random access, depending on the underlying collection it is build of.

satorg: One more method to consider: `iterableFactory`. The point is that `Seq` is not a view-like class. Therefore, by default it tries to re-create itself in every single transformation or filtering method. E.g. `map`, `flatMap`, `filter`, etc. That means, in particular, that if `iterableFactory` is not overridden, then `map` called on such a collection will convert it into a default representation (which is `List` for `Seq`). Not sure if it is the behavior that users would expect. So perhaps it would be better to create a compacted version of `Chunk` in such cases or something like that.

BalmungSan: @satorg IMHO, the whole point of this method is to provide a copy-free view of `Chunk` as a `Seq`, whose main use case is precisely a `foreach` kind of interop.
Thus, while we do try to avoid pointless indirections for methods like `isEmpty` or `to*`, any real use of this collection will consume it and produce a real `Seq` _(like `List`)_.

We may be more explicit on the **Scaladoc** if you think that would be useful.

BalmungSan: However, what about using `ArraySeq` or `Vector` as the default `iterableFactory` which would be more akin to `Chunk` than `List`?
cc @satorg @armanbilge

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -88,6 +88,10 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
     Chunk.array(b.result()).asInstanceOf[Chunk[O2]]
   }
 
+  /** Returns true if the Chunk contains the given element. */
+  def contains[O2 >: O](elem: O2): Boolean =
```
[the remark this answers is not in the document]
armanbilge: Yes, we should definitely use an `Eq` based version. Generally the method signatures should match the syntax from `cats.Foldable`.

.sbtopts
```diff
@@ -1,3 +1,3 @@
--J-Xms2g
--J-Xmx4g
--J-XX:MaxMetaspaceSize=512m
+-J-Xms4g
+-J-Xmx8g
+-J-XX:MaxMetaspaceSize=1024m
```
armanbilge, line 3: GitHub runners only have 7 GB of RAM, and we need to leave memory for forked JVMs / Node.js / Native processes.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
BalmungSan: @satorg @armanbilge done, both the **Scaladoc** adjustment and the _override_ of `iterableFactory`.
Please, let me know what you both think.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
```diff
@@ -0,0 +1,162 @@
...
+      case a: ArraySeq[O]   => Some(arraySeq(a))
+      case w: ChunkAsSeq[O] => Some(w.chunk)
+      case _                => None
+    }
+
+  /** Creates a chunk backed by an immutable `ArraySeq`. */
+  def arraySeq[O](arraySeq: ArraySeq[O]): Chunk[O] = {
+    val arr = arraySeq.unsafeArray.asInstanceOf[Array[O]]
+    array(arr)(ClassTag[O](arr.getClass.getComponentType))
+  }
```
satorg, line 157: I see that you copied it from the previous implementation to consolidate it in one file. However, perhaps such an implementation does not make a lot of sense anymore since there's `indexedSeq` method already and `ArraySeq` inherits to `IndexedSeq`. So I guess you could simply redirect this method to `indexedSeq` instead.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
```diff
@@ -0,0 +1,162 @@
...
+    buf.result()
+  }
+
+  /** Views this Chunk as a Scala immutable Seq.
+    * Contrary to all methods that start with _"to"_ (e.g. {{toVector}}, {{toArray}}),
+    * this method does not copy data.
+    * As such, this method is mostly intended for `foreach` kind of interop.
+    */
+  def asSeq: Seq[O] =
+    new ChunkAsSeq(this)
```
satorg, line 52: Perhaps it is worth some optimization:
As I can see, there is already
```
private final class IndexedSeqChunk[O](s: GIndexedSeq[O]) extends Chunk[O]
```
So if `this` is `IndexedSeqChunk` is would be better to return the underlying `s` from here.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
armanbilge: Hmm, I think it still makes sense. This is getting at the underlying `Array` which is much more optimized to work with e.g. see `ArraySlice`.

satorg: Maybe.. But in that case I feel it would make sense to add such an optimization to the `indexedSeq` itself, i.e. if a seq is an `ArraySeq`, then make an `ArraySlice` out of it.

Then redirect `arraySeq` to `indexedSeq` :) 

But it seems to be out of scope for this PR anyway.

BalmungSan: @satorg `indexedSeq` is AFAIK generic to all **Scala** versions, so it can check for `arraySeq`

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
BalmungSan: @satorg @armanbilge I tried to do this, but `GIndexedSeq` is `collection.IndexedSeq` whereas we are using `collection.immutable.Seq`

Personally, I find it weird and wrong that `Chunk` is using the general trait, since that allows mutable collections, meaning the data may change under the hood.

satorg: yeah me too... I guess it might be due to an oversight....

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -619,6 +663,44 @@ object Chunk
...
   }
 
+  /** Creates a chunk from a mutable `java.util.List`. */
+  def javaList[O](javaList: ju.List[O]): Chunk[O] =
+    javaList match {
+      case chunkAsJavaList: ChunkAsJavaList[O] =>
+        chunkAsJavaList.chunk
+
+      case randomAccess: ju.RandomAccess =>
+        new JavaListChunk(randomAccess)
```
satorg, line 673: Just to double check: `java.util.List` is a mutable thing by design. Is it ok to wrap it without making a copy?

BalmungSan: @satorg I had the same doubt and spent a while looking if there was any way to know if a `ju.List` was **unmodifiable**, even using _reflection_. But no, **Java** things that is an implementation detail nobody would ever care for e.e 😡 

Anyways, ranting aside, it seems most `Chunk` factories wrap mutable things without copying. So I guess it is fine per the philosophy of the library.
I assume the rationale is mostly performance and let users take care of it.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
```diff
@@ -120,7 +120,7 @@ private[fs2] final class ChunkAsSeq[+O](
     chunk.toVector
 
   override def toString: String =
-    chunk.toString
+    chunk.iterator.mkString("ChunkAsSeq(", ", ", ")")
```
mpilquist: I don't think the toString should leak the implementation class. For example:

```scala
scala> val c = Chunk(1, 2, 3)
val c: fs2.Chunk[Int] = Chunk(1, 2, 3)

scala> val c2 = c ++ c
val c2: fs2.Chunk[Int] = Chunk(1, 2, 3, 1, 2, 3)

scala> val first4 = c2.take(4)
val first4: fs2.Chunk[Int] = Chunk(1, 2, 3, 1)

scala> (c.getClass, c2.getClass, first4.getClass)
val res3: (Class[? <: fs2.Chunk[Int]], Class[? <: fs2.Chunk[Int]],
  Class[? <: fs2.Chunk[Int]]
) = (class fs2.Chunk$ArraySlice,class fs2.Chunk$Queue,class fs2.Chunk$Queue)
```

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -619,6 +663,44 @@ object Chunk
...
+
+      case _ =>
+        val size = javaList.size
+        val arr = new Array[Object](size).asInstanceOf[Array[O with Object]]
+        javaList.toArray(arr)
+        new ArraySlice(arr, 0, size)(ClassTag.Object.asInstanceOf[ClassTag[O with Object]])
```
satorg, line 679: If it is ok to make a `Chunk` that is backed by a mutable collection (as we discussed above), then is it really necessary to make a copy here? I mean, since `Chunk` does not guarantee efficient random access:
https://github.com/typelevel/fs2/blob/5dbe235e5a4e763251f94c3880a75ba20915edbb/core/shared/src/main/scala/fs2/Chunk.scala#L1029

So is it that necessary to enforce it here then?

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
mpilquist: Take a look at the `platformIterable` method, which is used in the implementation of `iterable`, which in turn `seq` delegates to. We could unwind this a bit, deprecating the public facing methods that simply delegate to the type checking uber method.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -525,6 +527,14 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
 
   override def toString: String =
     iterator.mkString("Chunk(", ", ", ")")
+
+  /** Views this Chunk as a Java unmodifiable List.
+    * Contrary to all methods that start with _"to"_ (e.g. {{toVector}}, {{toArray}}),
+    * this method does not copy data.
+    * As such, this method is mostly intended for `foreach` kind of interop.
+    */
+  def asJava: ju.List[O @uncheckedVariance] =
```
mpilquist, line 556: What's the motivation for `asJava`? Versus using built-in jdk converters like:

```scala
scala> import scala.jdk.CollectionConverters._

scala> c.iterator.asJava
val res5: java.util.Iterator[Int] = scala.collection.convert.JavaCollectionWrappers$IteratorWrapper@7a2b9987
```

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: Yeah, we intentionally don't do defensive copies for performance reasons. See [this talk](https://mpilquist.github.io/blog/2021/11/26/fs2/) for some details.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
BalmungSan: @mpilquist Well then what should we use then?
The only two other options are `Seq` and `Chunk`, and I don't think either is a good option.
+ `Chunk`: Since `Chunk` is not a `Seq`, then doing `aSeq.toString` printing `Chunk(...)` feels misleading.
+ `Seq`: AFAIK _(I could be wrong in this)_ there is no a single type in the stdlib printing `Seq(...)` 

Also, I just realized that if we don't override then we would have got the same result: https://github.com/scala/scala/blob/v2.13.11/src/library/scala/collection/Iterable.scala#L78

core/shared/src/main/scala/fs2/Chunk.scala
BalmungSan: @satorg uhm, that is a good point, not sure then.
@armanbilge @mpilquist should we just always use `JavaListAsChunk`?

core/shared/src/main/scala/fs2/Chunk.scala
BalmungSan: @mpilquist copy-pasting from the original issue:
> On top of this, it may also be good to provide an `asJava` facade as well, returning `ju.List`.
> Technically speaking, `asSeq` would allow us to also call the `asJava` the stdlib already provides, but that would be a facade over a facade; not the end of the world but it may be a good tiny optimization.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
BalmungSan: @mpilquist so you are suggesting that we deprecate the `arraySeq` factories and rather let `iterable` call the optimized impl?
I don't have an issue with that, should we also deprecate `indexedSeq` and `seq` as well and rather just have a single `iterable` factory?

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
mpilquist: Disregard, as a toString for the `Seq` subtype, this is fine. I was thinking it was a new chunk subtype when I commented.

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: I would keep a copy in this case. Chunk.Queue is the only case we've relaxed O(1) indexed access. I'd like to keep it the only case.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
mpilquist: We could, yeah. The various specific constructors, like `arraySeq` and `indexedSeq` and `vector` all wrap direct construction of a corresponding `Chunk` subtype. Whereas `iterable` does type casing to pick the most efficient subtype. Then `seq` breaks the first rule and delegates to `iterable`.

Moving forward, one option is deprecating all of those and creating `Chunk.collection` (or `fromCollection` or maybe even just `from`) which uses the current implementation of `iterable`.

core/shared/src/main/scala/fs2/Chunk.scala
mpilquist: Hm, is this a common need? I don't think I've ever wanted Chunk <=> j.u.List conversion.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -1298,5 +1380,125 @@ object Chunk
...
+      val result = chunk.apply(i)
+      i += 1
+      result
+    }
+  }
+
+  override def toArray: Array[Object] =
+    chunk.toArray[Any].asInstanceOf[Array[Object]]
+
+  override def toArray[T](a: Array[T with Object]): Array[T with Object] = {
```
mpilquist, line 1496: Check this with primitive based arrays. I'm concerned you'll hit a bunch class cast exceptions due to primitive arrays vs object arrays.

core/shared/src/main/scala/fs2/Chunk.scala
BalmungSan: @mpilquist so, to be completely honest, I am currently needing such conversion which was the main driver for this PR.
However, in the original conversation I had with Arman on **Discord**, before creating the issue, we both concluded that `asSeq` was the most important piece since it would allow `asJava` and more.
Yet, I couldn't shake the idea that having `asJava` directly would be best to avoid the double facade.

mpilquist: OK, I don't feel strongly about it so I'm fine with keeping it. Note we dropped some rarely used chunk subtypes in the past -- the various non-byte buffers from java.nio.

BalmungSan: @mpilquist I am okay with removing the `JavaListAsChunk` and rather just copy-to-array always if you prefer.
However, I do think the `javaList` factory is useful even if just for the sake of being able to `unwrap` the facade.

mpilquist: I'm fine either way. Your call.

BalmungSan: @mpilquist for now I will leave it then, since the code is already there and in the spirit of _"avoid defensive copies"_.
But, will add a comment mentioning that if in the future data show this is rarely used it can be removed.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -88,6 +90,10 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
     Chunk.array(b.result()).asInstanceOf[Chunk[O2]]
   }
 
+  /** Returns true if the Chunk contains the given element. */
+  def contains[O2 >: O: Eq](elem: O2): Boolean =
+    iterator.exists(o => Eq[O2].eqv(o, elem))
```
satorg, line 95: I will be calling the Eq's summoner for every `o`. Even though it should not have a huge impact, I feel it would be better to avoid it:
```suggestion
  def contains[O2 >: O](elem: O2)(implicit ev: Eq[O2]): Boolean =
    iterator.exists(o => ev.eqv(o, elem))
```

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -1298,5 +1380,125 @@ object Chunk
...
+  override def toVector: Vector[O] =
+    chunk.toVector
+
+  override def toString: String =
+    chunk.iterator.mkString("ChunkAsSeq(", ", ", ")")
+
+  override def equals(that: Any): Boolean =
+    that match {
+      case thatChunkWrapper: ChunkAsSeq[_] =>
+        chunk == thatChunkWrapper.chunk
```
filipwiech, line 1459: Not sure, but maybe it would be a good idea to override the `hashCode` implementation as well (so that it only takes the `chunk` into account), since we have this custom `equals`. It might be that it's already the case, but I feel like it would still make the code less error-prone in case of any future changes. :slightly_smiling_face:

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -1298,5 +1380,125 @@ object Chunk
...
+
+      case _ =>
+        false
```
filipwiech, line 1462: The `Chunk#asSeq` returns an `IndexedSeq`: is it expected that it will not be treated as equal to other similar sequences, including the `IndexedSeqChunk` contents that can also be returned as the result of the `asSeq` call? The `ChunkAsJavaList` used by the `Chunk#asJava` doesn't seem to have such limitation either. Sorry if I misunderstood anything. :slightly_smiling_face:

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -1066,11 +1186,39 @@ object Chunk
 
     /** Prepends a chunk to the start of this chunk queue. */
     def +:[O2 >: O](c: Chunk[O2]): Queue[O2] =
-      if (c.isEmpty) this else new Queue(c +: chunks, c.size + size)
+      if (c.isEmpty) this
+      else
+        c match {
```
satorg, line 1191:
```suggestion
      if (c.isEmpty) this
      else if (this.isEmpty) c
      else
        c match {
```
and the same recommendation for `:+` down below.

satorg: I think that `++` can simply call to this method (i.e. `+:`) because they both seem doing the same thing, only the return type is different (`++` returns `Chunk` whereas `+:` returns `Queue`).

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -1298,5 +1446,137 @@ object Chunk
...
+
+  override def toVector: Vector[O] =
+    chunk.toVector
+
+  override def toString: String =
+    chunk.iterator.mkString("ChunkAsSeq(", ", ", ")")
+
+  override def hashCode: Int = {
+    import util.hashing.MurmurHash3
+    var h = MurmurHash3.stringHash("ChunkAsSeq")
```
filipwiech, line 1563: Not sure if we should mix the class name (`ChunkAsSeq`) into the hash code, based on the `equals` implementation. For example, if the `ChunkAsSeq` instance is equal to the `GSeq` (e.g. coming from the `IndexedSeqChunk`), then they should have the same hash code values. At least that's my understanding, hope I haven't mixed anything. Maybe computing it only from the `chunk` contents would be more appropriate. :slightly_smiling_face:

core/shared/src/main/scala/fs2/Chunk.scala
BalmungSan: @satorg I made the changes in a slightly different way, in order to minimize the amount of allocations.

core/shared/src/main/scala/fs2/Chunk.scala
BalmungSan: Man, the collections stdlib are a nightmare to understand.
Thanks for the heads up, just fixed it _(turned out, we shouldn't have need to override this in the first place)_.

core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -1298,5 +1485,133 @@ object Chunk
...
+    MurmurHash3.indexedSeqHash(this, seed = MurmurHash3.seqSeed)
+  }
+
+  override def equals(that: Any): Boolean =
+    that match {
+      case thatChunkWrapper: ChunkAsSeq[_] =>
+        chunk == thatChunkWrapper.chunk
+
+      case seq: GSeq[_] =>
+        chunk.iterator.sameElements(seq.iterator)
```
filipwiech, line 1572:
```suggestion
        chunk.size == seq.size && chunk.iterator.sameElements(seq.iterator)
```
Might be a worthwhile improvement (`Chunk#equals` does something similar). :slightly_smiling_face:

BalmungSan: @filipwiech since that is a general `Seq` calling `size` may be expensive.

filipwiech: Oh, good point! :+1:

filipwiech:
For what it's worth the changes look good to me based on my limited knowledge, so take it with a grain of salt. :wink: The only thing missing from my point of view are unit tests, would be great to cover the newly introduced functionality:
* New `Chunk` methods: `contains`, `count`, `exists`.
* Some basic `asSeq` usage.
* Some basic `asJava` usage.
* New `Chunk.from` factory.
* New `Chunk.javaList` factory.

core/shared/src/main/scala-2.13+/fs2/Chunk213And3Compat.scala
BalmungSan: @mpilquist I just noticed that `Chunk.empty` breaks this rule, and rather prints `empty` instead of `Chunk()`
Is that intentional?

BalmungSan: @mpilquist would you like me to remove that as well on this PR?
Or are we afraid that would break some user code?

mpilquist: Sure

core/shared/src/main/scala/fs2/Chunk.scala
BalmungSan: @mpilquist it seems that API will always box, even when called from **Java** and as such, I think it is working as expected.
I added a couple of property base testing which seem to confirm that is the case.

But, not sure if you have some special case we should cover as a traditional hand-made unit test.

BalmungSan:
@mpilquist @armanbilge I think the PR is ready for review, sorry for taking too long to add the tests.
Let me know if you have more ideas for things to test.

PS: Not sure how to fix the **ScalaJS** link error, that seems like an oversight in **Javalib**?

mpilquist (approved):

Issue #3276, opened 2023-08-15
rechunkRandomlyWithSeed fails to rechunk in case of large chunks


domaspoliakas:
The following fails to rechunk in accordance to the contract:

```scala
import cats.effect._
import fs2._

Stream
  .emits(List.tabulate(5000)(_ => 1))
  .covary[SyncIO]
  .rechunkRandomlyWithSeed(0.01, 0.1)(1L)
  .chunks
  .head
  .compile
  .toList
  .unsafeRunSync()
  .head
  .size // prints 5000
  ```
  
This is because my stream is just one huge chunk. The first recursive case hits [this](https://github.com/typelevel/fs2/blob/main/core/shared/src/main/scala/fs2/Stream.scala#L2390-L2391) as the acc is empty, and then the second recursion hits [this](https://github.com/typelevel/fs2/blob/main/core/shared/src/main/scala/fs2/Stream.scala#L2398-L2399) as there is nothing else in the stream, immediately emitting everything instead of continuing to chunk until the size is adhered to.

PR #3277, opened 2023-08-15
fix rechunkRandomly with large ending chunk

domaspoliakas:
Fixes https://github.com/typelevel/fs2/issues/3276

Not sure if I put the test in the right place, let me know if you'd like to move it!

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2378,44 +2378,45 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   ): Stream[F2, O] =
     Stream.suspend {
       assert(maxFactor >= minFactor, "maxFactor should be greater or equal to minFactor")
-      val random = new scala.util.Random(seed)
-      def factor: Double = Math.abs(random.nextInt()) % (maxFactor - minFactor) + minFactor
-
-      def nextSize(sourceSize: Int): Int = (factor * sourceSize).toInt
-
-      def go(
-          acc: Chunk[O],
-          sizeOpt: Int,
-          lastChunkSize: Int,
-          s: Pull[F2, O, Unit]
-      ): Pull[F2, O, Unit] = {
-
-        val size = if (sizeOpt > 0) sizeOpt else nextSize(lastChunkSize)
-
-        if (acc.size < size)
-          s.uncons.flatMap {
-            case None => Pull.output(acc)
-            case Some((hd, tl)) =>
-              go(acc ++ hd, size, hd.size, tl)
-          }
-        else if (acc.size == size)
-          Pull.output(acc) >>
-            s.uncons.flatMap {
-              case None => Pull.done
-              case Some((hd, tl)) =>
-                go(hd, size, hd.size, tl)
-            }
-        else {
-          val (out, rem) = acc.splitAt(size - 1)
-          Pull.output(out) >> go(rem, -1, lastChunkSize, s)
-
-        }
-
-      }
 
       underlying.uncons.flatMap {
         case None => Pull.done
         case Some((hd, tl)) =>
+          val random = new scala.util.Random(seed)
+          def factor: Double = Math.abs(random.nextInt()) % (maxFactor - minFactor) + minFactor
+
+          def nextSize(sourceSize: Int): Int = (factor * sourceSize).toInt
+
+          def go(
+              acc: Chunk[O],
+              sizeOpt: Int,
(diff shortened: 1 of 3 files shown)


core/shared/src/test/scala/fs2/StreamSuite.scala
```diff
@@ -1042,4 +1042,13 @@ class StreamSuite extends Fs2Suite {
     }
   }
 
+  test("rechunkRandomlyWithSeed should correclty rechunk big chunks at the end of a stream") {
```
armanbilge, line 1045: Looks like we keep those tests here.

https://github.com/typelevel/fs2/blob/5dbe235e5a4e763251f94c3880a75ba20915edbb/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala#L1317

core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
```diff
@@ -1354,6 +1354,16 @@ class StreamCombinatorsSuite extends Fs2Suite {
         }
       }
     }
+
+    test("rechunkRandomlyWithSeed should correclty rechunk big chunks at the end of a stream") {
```
armanbilge, line 1358:
```suggestion
    test("correctly rechunk big chunks at the end of a stream") {
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2383,23 +2383,41 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+          val (out, rem) = acc.splitAt(size - 1)
+          Pull.output(out) >> go(rem, -1, lastChunkSize, s)
+
         }
 
-      go(Chunk.empty, -1, underlying).stream
+      }
+
+      underlying.uncons.flatMap {
+        case None => Pull.done
```
filipwiech, line 2417: Idea, since the new implementation is slightly more complex anyway, maybe we could avoid creation of the `random` object (`val random = new scala.util.Random(seed)`) in this empty case, since it's then not used? Or maybe it's not worth it, not sure. :slightly_smiling_face:

domaspoliakas: We could, but I'm not sure if it's worth it. It'd only have any effect if the random rechunking was called repeatedly on empty streams, which seems like an extraordinarily niche scenario.

Anyway, I'm happy to do it if told otherwise, but I'll save the CI electricity for now

core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala
```diff
@@ -1354,6 +1354,16 @@ class StreamCombinatorsSuite extends Fs2Suite {
...
     }
+
+    test("correctly rechunk big chunks at the end of a stream") {
+      val chunks = Stream
+        .chunk(Chunk.seq(List.fill(5000)(1)))
+        .rechunkRandomlyWithSeed(0.01, 0.1)(1L)
+        .chunks
+        .compile
+        .toList
+      assert(chunks.forall(_.size < 500))
```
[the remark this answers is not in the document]
domaspoliakas: You're right, it should be inclusive

domaspoliakas:
CI seems to have broken on some unrelated part of code

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2378,28 +2378,47 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
 
-      def go(acc: Chunk[O], sizeOpt: Int, s: Pull[F2, O, Unit]): Pull[F2, O, Unit] =
-        s.uncons.flatMap {
-          case Some((hd, tl)) =>
-            val size = if (sizeOpt > 0) sizeOpt else nextSize(hd.size)
             if (acc.size < size)
-              go(acc ++ hd, size, tl)
+              s.uncons.flatMap {
+                case None => Pull.output(acc)
+                case Some((hd, tl)) =>
```
filipwiech, line 2402: Nit: while reading the newest iteration of the code I had to do a double take, because there is now shading here (`hd` and `tl` used in the `go` function hide the variables with the same names from the higher scope of `underlying.uncons.flatMap`). Maybe it would be nice to avoid this for the sake of clarity. :slightly_smiling_face:

filipwiech (approved):
LGTM. :+1:

Jasper-M:
#3195 also seems to have progressed after this PR. You no longer get 1 very large `Chunk`.

Issue #3279, opened 2023-08-22
Add `Channel#sendClose`


armanbilge:
And also maybe `Channel#trySendClose`. Naming subject to bikeshed e.g. `sendAndClose`? 

The idea is to be able to atomically `send` a final element to the `Channel` and immediately `close` it. With `ch.send(...) *> ch.close` there is a possibility that another concurrent `send` can sneak between.

Motivation: in https://github.com/http4s/http4s/pull/7196 we want to `close` the `Channel` immediately upon sending a WebSocket CloseFrame.

@tothpeti has kindly volunteered to work on this :)

tothpeti:
I would like to work on this one :)

PR #3280, opened 2023-08-22
Add `Channel#closeWithElement`

tothpeti:
This PR adds the requested functionality: https://github.com/typelevel/fs2/issues/3279

- Add `closeWithElement` in `Channel`  
- Add test case

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/Channel.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Channel.scala
@@ -110,8 +110,7 @@ sealed trait Channel[F[_], A] {
     */
   def close: F[Either[Channel.Closed, Unit]]
 
-  /** Sends an element through this channel, and closes it right after.
-    * This operation is atomic so it is guaranteed that this will be the last element sent to the channel.
+  /** Gracefully closes this channel with a final element. This method will never block.
     *
     * No-op if the channel is closed, see [[close]] for further info.
     */
@@ -176,7 +175,7 @@ object Channel {
                       State(values, size, None, (a, producer) :: producers, close),
                       signalClosure.whenA(close) *>
                         notifyStream(waiting).as(rightUnit) <*
-                        waitOnBound(producer, poll).whenA(!close)
+                        waitOnBound(producer, poll).unlessA(close)
                     )
               }
             }
--- a/core/shared/src/main/scala/fs2/concurrent/Channel.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Channel.scala
@@ -176,7 +176,7 @@ object Channel {
                       State(values, size, None, (a, producer) :: producers, close),
                       signalClosure.whenA(close) *>
                         notifyStream(waiting).as(rightUnit) <*
-                        waitOnBound(producer, poll)
+                        waitOnBound(producer, poll).whenA(!close)
                     )
               }
             }
--- a/core/shared/src/main/scala/fs2/concurrent/Channel.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Channel.scala
@@ -176,7 +176,7 @@ object Channel {
                       State(values, size, None, (a, producer) :: producers, close),
                       signalClosure.whenA(close) *>
                         notifyStream(waiting).as(rightUnit) <*
-                        waitOnBound(producer, poll).whenA(!close)
+                        waitOnBound(producer, poll)
                     )
               }
             }
--- a/core/shared/src/main/scala/fs2/concurrent/Channel.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Channel.scala
@@ -111,6 +111,7 @@ sealed trait Channel[F[_], A] {
   def close: F[Either[Channel.Closed, Unit]]
 
   /** Sends an element through this channel, and closes it right after.
+    * This operation is atomic so it is guaranteed that this will be the last element sent to the channel.
     *
     * No-op if the channel is closed, see [[close]] for further info.
     */
--- a/core/shared/src/main/scala/fs2/concurrent/Channel.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Channel.scala
@@ -167,12 +167,12 @@ object Channel {
                 case State(values, size, waiting, producers, closed @ false) =>
                   if (size < capacity)
                     (
-                      State(a :: values, size + 1, None, producers, false),
+                      State(a :: values, size + 1, None, producers, close),
                       signalClosure.whenA(close) *> notifyStream(waiting).as(rightUnit)
                     )
                   else
                     (
-                      State(values, size, None, (a, producer) :: producers, false),
+                      State(values, size, None, (a, producer) :: producers, close),
                       signalClosure.whenA(close) *>
                         notifyStream(waiting).as(rightUnit) <*
                         waitOnBound(producer, poll)
--- a/core/shared/src/main/scala/fs2/concurrent/Channel.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Channel.scala
@@ -110,6 +110,12 @@ sealed trait Channel[F[_], A] {
     */
   def close: F[Either[Channel.Closed, Unit]]
 
+  /** Sends an element through this channel, and closes it right after.
+    *
+    * No-op if the channel is closed, see [[close]] for further info.
+    */
+  def sendAndClose(a: A): F[Either[Channel.Closed, Unit]]
+
   /** Returns true if this channel is closed */
   def isClosed: F[Boolean]
 
@@ -186,6 +192,28 @@ object Channel {
                 )
               else
                 (s, rightFalse.pure[F])
+          }
+
+        def sendAndClose(a: A) =
+          F.deferred[Unit].flatMap { producer =>
+            state.flatModifyFull { case (_, state) =>
+              state match {
+                case s @ State(_, _, _, _, closed @ true) =>
+                  (s, Channel.closed[Unit].pure[F])
+
+                case State(values, size, waiting, producers, closed @ false) =>
+                  if (size < capacity)
+                    (
+                      State(a :: values, size + 1, None, producers, false),
+                      notifyStream(waiting).as(rightUnit)
+                    )
+                  else
+                    (
+                      State(values, size, None, (a, producer) :: producers, true),
+                      notifyStream(waiting).as(rightUnit) <* signalClosure
+                    )
+              }
+            }
           }
 
         def close =
--- a/core/shared/src/main/scala/fs2/concurrent/Channel.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Channel.scala
@@ -157,7 +157,7 @@ object Channel {
             .drain
         }
 
-        def send(a: A) =
+        def sendImpl(a: A, close: Boolean) =
           F.deferred[Unit].flatMap { producer =>
             state.flatModifyFull { case (poll, state) =>
               state match {
@@ -170,6 +170,11 @@ object Channel {
                       State(a :: values, size + 1, None, producers, false),
                       notifyStream(waiting).as(rightUnit)
                     )
+                  else if (close)
+                    (
+                      State(values, size, None, producers, true),
+                      notifyStream(waiting).as(rightUnit) <* signalClosure
+                    )
                   else
                     (
                       State(values, size, None, (a, producer) :: producers, false),
@@ -178,6 +183,10 @@ object Channel {
               }
             }
           }
+
+        def send(a: A) = sendImpl(a, false)
+
+        def sendAndClose(a: A) = sendImpl(a, true)
 
         def trySend(a: A) =
           state.flatModify {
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -188,6 +194,28 @@ object Channel {
                 (s, rightFalse.pure[F])
           }
 
+        def sendAndClose(a: A) =
```
armanbilge, line 197: This is a copy-paste of the logic from `send` right? Instead of duplicating it, can we have:

```scala
def send(a: A) = sendImpl(a, false)
def sendAndClose(a: A) = sendImpl(a, true)
def sendImpl(a: A, close: Boolean) = ???
```

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
```diff
@@ -251,4 +251,18 @@ class ChannelSuite extends Fs2Suite {
 
     TestControl.executeEmbed(test)
   }
+
+  test("sendAndClose closes right after sending the last element") {
```
armanbilge, line 255: Maybe I missed it but I don't think this test exercises the race condition that rules out an implementation like `send(...) *> close`.

I was thinking we can write a test like:
```scala
ch.send(0).both(ch.sendAndClose(1)).product(ch.stream.compile.toList)
```
The result of that should be one of:
```scala
((Right(()),Right(())),List(0, 1))

((Left(Closed),Right(())),List(1))
```

Any other result should be impossible I believe. The trivial `send(...) *> close` would fail that test.

armanbilge (commented):
More name ideas: `sendThenClose` or `sendLast`.

tothpeti:
> More name ideas: `sendThenClose` or `sendLast`.

I think sendAndClose would be pretty clear for the user.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -164,6 +170,11 @@ object Channel {
...
                       notifyStream(waiting).as(rightUnit)
                     )
+                  else if (close)
+                    (
+                      State(values, size, None, producers, true),
+                      notifyStream(waiting).as(rightUnit) <* signalClosure
+                    )
                   else
                     (
                       State(values, size, None, (a, producer) :: producers, false),
```
armanbilge, line 180: Hmm, this logic doesn't seem right to me. Whether `size < capacity` or not is indepndent of `close`. For example, if `size < capacity` then you take the first branch and are not closing the channel, which is incorrect.

I think it should be something like this.

```scala
                  if (size < capacity)
                    (
                      State(a :: values, size + 1, None, producers, close),
                      signalClosure.whenA(close) *> notifyStream(waiting).as(rightUnit)
                    )
                  else
                    (
                      State(values, size, None, (a, producer) :: producers, close),
                      signalClosure.whenA(close) *> notifyStream(waiting).as(rightUnit) <* waitOnBound(producer, poll)
                    )
```

tothpeti: Yes, indeed. This makes sense! Thanks!

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
```diff
@@ -251,4 +251,28 @@ class ChannelSuite extends Fs2Suite {
...
+    }
+
+    result.assertEquals(true)
+  }
+
+  test("racing send and sendAndClose should work in bounded case") {
+    val test = Channel.bounded[IO, Int](2).flatMap { ch =>
+      ch.send(0).both(ch.sendAndClose(1))
+    }
+    test.assertEquals((Right(()), Right(())))
```
filipwiech, line 267: I could be wrong, but I think remark from @armanbilge (https://github.com/typelevel/fs2/pull/3280#discussion_r1301707331) makes a lot of sense, i.e. maybe we should cover both possible cases and assert which elements were actually sent in the Channel (for example the current assertion would pass even if `send` would still succeed *after* `sendAndClose`). :slightly_smiling_face:

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
```diff
@@ -251,4 +251,28 @@ class ChannelSuite extends Fs2Suite {
 
     TestControl.executeEmbed(test)
   }
+
+  test("sendAndClose closes right after sending the last element") {
+    val result = Channel.bounded[IO, Int](1).flatMap { ch =>
+      ch.sendAndClose(0) *> ch.isClosed
```
filipwiech, line 257: Would it make sense to have this test for the unbounded Channel case as well, similarly to the "racing" scenarios? :slightly_smiling_face: 
Also, would it be a good idea to cover both impacted branches in the Channel implementation (`size < capacity` and `else`)? :slightly_smiling_face:

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
```diff
@@ -251,4 +251,89 @@ class ChannelSuite extends Fs2Suite {
...
+
+  test("sendAndClose closes right after sending the last element (unbounded case)") {
+    val result = Channel.unbounded[IO, Int].flatMap { ch =>
+      ch.sendAndClose(0) *> ch.isClosed
+    }
+
+    result.assertEquals(true)
+  }
+
+  test("racing send and sendAndClose should work in bounded(1) case") {
```
armanbilge, line 271: Instead of repeating this code four times, can we extract it to a function and test it with different `Channel`s?

Also, because the test is non-deterministic we should run it multiple times with `replicateA_(100)` or something.

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
```diff
@@ -251,4 +251,89 @@ class ChannelSuite extends Fs2Suite {
 
     TestControl.executeEmbed(test)
   }
+
+  test("sendAndClose closes right after sending the last element (bounded case)") {
```
armanbilge, line 255: Actually, same thing with this test. We can write it as a function and then test it with different channels without repeating.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -176,7 +176,7 @@ object Channel {
                       State(values, size, None, (a, producer) :: producers, close),
                       signalClosure.whenA(close) *>
                         notifyStream(waiting).as(rightUnit) <*
-                        waitOnBound(producer, poll)
+                        waitOnBound(producer, poll).whenA(!close)
```
armanbilge, line 179: Hmm, I'm not sure this change makes sense. `send` is supposed to block if there is no capacity in the `Channel`, and this is where the blocking happens. Semantically I would expect `sendAndClose(...)` to behave like `send(...) *> close` except atomically.

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
```diff
@@ -278,9 +278,17 @@ class ChannelSuite extends Fs2Suite {
...
-    val channel = Channel.synchronous[IO, Int]
+    val test = Channel
+      .synchronous[IO, Int]
+      .flatMap { ch =>
+        ch.stream
+          .concurrently(Stream.emit(0).covary[IO].evalMap(ch.sendAndClose))
+          .compile
+          .drain *> ch.isClosed
+      }
+      .assertEquals(true)
```
armanbilge, line 289: Nice! Here's a little refactoring of the same idea: draining the channel concurrently in the background.

```suggestion
    val test = Channel.synchronous[IO, Int].flatMap { ch =>
      ch.stream.compile.drain.background.surround {
        ch.sendAndClose(0) *> ch.isClosed.assert
      }
```

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
```diff
@@ -278,9 +278,17 @@ class ChannelSuite extends Fs2Suite {
...
 
-    checkIfSendAndCloseClosing(channel).parReplicateA_(if (isJVM) 10000 else 1)
+    test.parReplicateA_(if (isJVM) 10000 else 1)
```
armanbilge, line 291: Actually this particular test I don't think we need to `replicateA`: there isn't a race condition because the test has one possible outcome. This is unlike the other test where there is a race, and depending on the winner there are multiple outcomes that we have to check for.

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
```diff
@@ -251,4 +251,88 @@ class ChannelSuite extends Fs2Suite {
...
+        case obtained @ ((Left(Channel.Closed), Right(())), List(1)) =>
+          assertEquals(obtained, expectedThirdCase)
+        case e => fail(s"An unknown test result: $e")
+      }
+  }
+
+  test("racing send and sendAndClose should work in bounded(1) case") {
+    val channel = Channel.bounded[IO, Int](1)
+
+    racingSendOperations(channel).parReplicateA_(if (isJVM) 10000 else 1)
```
armanbilge, line 317: Can we use `replicateA_` instead of `parReplicateA_`? I'm less worried about running the tests in parallel to each other and more interested about the fibers _within_ the test running parallel to each other. If we overwhelm the scheduler with 10000 of this test running at the same time then that is less likely to happen.

If it's slow we probably don't need this many replicates. There's only two possible outcomes: honestly maybe even a 100 replicates would suffice.

Oh, also we can probably move this to the `racingSendOperations` method so we don't have to repeat it for each test :)

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
tothpeti: Oh okay, I will refactor it. Thanks!

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
tothpeti: Yes, sure, a moment and I will refactor it

SystemFw:
I've been trying to figure out the correct semantics for this operation, which I think depends on expected use case, because as it stands, sending is _not_ atomic with closing with this implementation. 
In particular cancelling a `sendAndClose` blocked on a bound means the channel is closed without having the element sent included in the channel's graceful closure.

SystemFw:
If I understand the use case correctly, I think the correct behaviour for this operation is to be called `closeWithElement`, and bypass the bound on a bounded channel without ever blocking, but enqueueing the final element alongside setting `closed =true`.

armanbilge:
Thanks. That sounds right to me, @tothpeti wdyt?

tothpeti:
> Thanks. That sounds right to me, @tothpeti wdyt?

Sounds good to me, as well.

core/shared/src/main/scala/fs2/concurrent/Channel.scala
```diff
@@ -115,7 +115,7 @@ sealed trait Channel[F[_], A] {
     *
     * No-op if the channel is closed, see [[close]] for further info.
     */
-  def sendAndClose(a: A): F[Either[Channel.Closed, Unit]]
+  def closeWithElement(a: A): F[Either[Channel.Closed, Unit]]
```
armanbilge, line 118: I guess we should update the scaladoc, something like "Gracefully closes this channel with a final element. This method will never block."

core/shared/src/test/scala/fs2/concurrent/ChannelSuite.scala
```diff
@@ -251,4 +251,79 @@ class ChannelSuite extends Fs2Suite {
...
+
+    channel
+      .flatMap { ch =>
+        ch.send(0).both(ch.closeWithElement(1)).parProduct(ch.stream.compile.toList)
+      }
+      .map {
+        case obtained @ ((Right(()), Right(())), List(0, 1)) =>
+          assertEquals(obtained, expectedFirstCase)
+        case obtained @ ((Right(()), Left(Channel.Closed)), List(1)) =>
+          assertEquals(obtained, expectedSecondCase)
```
armanbilge, line 297: Hmm, I don't think this case should ever happen? Is it happening?

tothpeti: It is happening once or twice during 1000 runs, but it is pretty rare.

armanbilge: I don't see how that can happen if everything is working correctly. If the `closeWithElement` returns that the `Channel` is already closed, then who is closing it? Seems like a bug 😕

tothpeti: To be honest, I left this match case there. When we iterated over multiple possible solutions. So, probably this will not happen. I will try it without this match case soon.

tothpeti: Yep, I tried with 100000 runs, and that match case never occurred. So, it can be safely removed. 👍

PR #3290, opened 2023-08-30
Fix for decoding a PES packet with a header, when the DTS (Decoding Timestamp) is unavailable.

tomiwu:
With the fix, when the DTS is absent and only the PTS is available, the decoding of PTS is now working. The starting bits are '0010', conveying the presence of PTS information.


mpilquist:
Awesome find, thanks! Want to run `scalafmtAll` to fix the formatting issue?

tomiwu:
Formatting fixed ;)

PR #3291, opened 2023-08-30
Localize `@nowarn212` to method parameters

armanbilge:
Tested locally and it seems to be a workaround for https://github.com/lampepfl/dotty/issues/18487.


armanbilge:
> Seems like skunk 1.0.0-M1 has same issue in [openssl object](https://github.com/typelevel/skunk/blob/main/modules/core/native/src/main/scala/net/message/openssl.scala)

Are you able to get the scaladoc problem with the openssl object? I thought that one is internal.

Agree it's a problem for core package object. I think I have an idea for that.

ahjohannessen:
@armanbilge I got the problem with the core package object. So I just searched for usage of nowarn, cannot tell if the scaladoc problem is with the openssl object.

PR #3312, opened 2023-09-25
Fix temporary file name generation on Node.js

armanbilge:
h/t @TonioGela in https://github.com/typelevel/toolkit/pull/105

### DIFF
--- a/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
@@ -123,8 +123,8 @@ private[fs2] trait FilesCompanionPlatform {
         permissions: Option[Permissions]
     ): F[Path] =
       for {
-        dir <- createTempDirectory(dir, prefix, permissions)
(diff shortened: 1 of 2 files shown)


io/js/src/main/scala/fs2/io/file/FilesPlatform.scala
```diff
@@ -123,8 +123,8 @@ private[fs2] trait FilesCompanionPlatform {
         permissions: Option[Permissions]
     ): F[Path] =
       for {
-        dir <- createTempDirectory(dir, prefix, permissions)
-        path = dir / Option(suffix).filter(_.nonEmpty).getOrElse(".tmp")
+        dir <- createTempDirectory(dir, "", permissions)
+        path = dir / (prefix + suffix)
```
TonioGela, line 127: Won't, in this way, all files have the same name? (Despite having different paths)

armanbilge: Yes, but I don't think that's a problem, since the full path is unique.

TonioGela (approved):

Issue #3106, opened 2023-01-10
readInputStream(in, chunkSize) allocates a full chunkSize for each read operation


jakobmerrild:
When attempting to upgrade from version `3.2.7` to `3.2.8` we encountered an `OutOfMemoryError` in one of our tests.

Unfortunately we haven't been able to reproduce the error using a minimal example. An attempt can be seen here: https://scastie.scala-lang.org/NflW2ZCIRzSnw7YJ8p5HCQ

Edit: Scastie has been updated to reflect the findings in the comments that the issue likely arises from calling `fs2.io.readInputStream` with a `chunkSize` that is (much) larger than the number of bytes read from the underlying stream when its `read(buffer, offset, length)` method is invoked.

Edit: The below is not really relevant for the issue, but kept for historic purposes

---
Our code takes an `java.io.InputStream` and uploads it to AWS S3 using the multipart upload functionality from [their SDK](https://github.com/aws/aws-sdk-java/blob/1.12.362/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3.java)

Our test generates a random kilobyte of data and then uploads varying sizes of repeated kilobytes to S3.
While obviously not exactly the same, the [Scastie](https://scastie.scala-lang.org/N6LPpKEeQAGyvXxYxHgnew) example reproduces the `fs2` parts of our implementation as faithfully as possible.

As mentioned the example doesn't actually exhibit the same behavior as our tests. See below for a screenshot of a heapdump captured by setting `javaOptions +=  "-XX:+HeapDumpOnOutOfMemoryError"`
![image](https://user-images.githubusercontent.com/2920412/211547429-599f3ea4-e21f-41ec-904d-6e1c593bff36.png)

It looks like the [accB field of the OuterRun class](https://github.com/typelevel/fs2/blob/v3.2.8/core/shared/src/main/scala/fs2/Pull.scala#L1245) for whatever reason builds up a 500MB `scodec.bits.ByteVector.Buffer`

It might be of interest that our tests don't fail if the test data is only kilobytes large (10-13 KB).

mpilquist:
The stack trace in the screenshot seems to be referring to a stream of bytes that's compiled to an `F[ByteVector]` via `src.compile.to(ByteVector)`. Do you have any code like that in the original code base?

kalanzai:
[...]

PR #3318, opened 2023-10-11
Fix 'fs.io.readInputStreamGeneric' overallocation of underlying buffers

seigert:
Current implementation of `fs2.io.readInputStream` allocates new array of `chunkSize` size for every invocation of `InputStream#read(..)`. The problem is that `read` spec does not require `InputStream` implementations to write provided buffer fully or until stream exhaustion.

This leads to situation where if `readInputStream` chunk size is big (megabytes) and underlying input stream innner 'chunks' are small (bytes or kilobytes), returned `Stream[F, Byte]` are very 'sparse' where in every chunk only small amount of allocated capacity is actually used.

It could be shown easily by combining `fs2.io.toInputStream` with `fs2.io.readInputStream`.

This PR fixes this by reusing leftovers of allocated `Array[Byte]` for consequent 'InputStream' reads until either buffer is fully written or stream is exhausted. Like current implementation `fs2.Stream` chunks are published with each successful `InputStream#read` invocation but may share underlying array buffer.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -236,6 +236,10 @@ ThisBuild / mimaBinaryIssueFilters ++= Seq(
   ),
   ProblemFilters.exclude[InheritedNewAbstractMethodProblem](
     "fs2.io.file.Files.openSeekableByteChannel"
+  ),
+  // package-private method: #3318
+  ProblemFilters.exclude[IncompatibleMethTypeProblem](
+    "fs2.io.package.readInputStreamGeneric"
   )
 )
 
--- a/io/jvm/src/main/scala/fs2/io/ioplatform.scala
+++ b/io/jvm/src/main/scala/fs2/io/ioplatform.scala
@@ -43,10 +43,10 @@ private[fs2] trait ioplatform extends iojvmnative {
       F.pure(System.in),
       F.delay(new Array[Byte](bufSize)),
       false
-    ) { (is, buf) =>
+    ) { (is, buf, off) =>
       F.async[Int] { cb =>
         F.delay {
-          val task: Runnable = () => cb(Right(is.read(buf)))
+          val task: Runnable = () => cb(Right(is.read(buf, off, buf.length - off)))
           stdinExecutor.submit(task)
         }.map { fut =>
           Some(F.delay {
--- a/io/jvm/src/test/scala/fs2/io/IoPlatformSuite.scala
+++ b/io/jvm/src/test/scala/fs2/io/IoPlatformSuite.scala
@@ -40,6 +40,42 @@ class IoPlatformSuite extends Fs2Suite {
   // This suite runs for a long time, this avoids timeouts in CI.
   override def munitIOTimeout: Duration = 2.minutes
 
+  group("readInputStream") {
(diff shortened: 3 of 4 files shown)


armanbilge (commented):
~Thanks for the PR! Unfortunately I do not think this is the right way to solve this problem. The concern is what if there is actually not enough data available on the `InputStream` at a particular moment, but it is not yet closed. E.g. consider interacting with an external process over stdin/stdout.~

~Have you seen this issue?~
- https://github.com/typelevel/fs2/issues/3106

~I proposed a couple different ideas to solve it in https://github.com/typelevel/fs2/issues/3106#issuecomment-1378887690.~

armanbilge:
> chunks ... but may share underlying array buffer.

I guess my only concern with this approach is that a single `Chunk` that is retained for a long time may still have a surprisingly large memory footprint due to a large backing array. I'm not sure how realistic this is in practice :)

To avoid that would require allocating an appropriately sized array and copying into that, as suggested in https://github.com/typelevel/fs2/issues/3106#issuecomment-1378887690.

seigert:
@armanbilge, thnal for your input!

The issue I'm trying to resolve is that we have network data in size of megabytes and, naturally, we assume that buffer of 1MiB would be good enough. Only problem is that underlying `InputStream#read` decided to write data in 1KiB chunks. And suddenly per every mere megabyte of data we have whooping 1 Gigabyte of allocated RAM that only 0.1% populated. 

What's worse, it is really had to diagnose: at first everything works, maybe using more memory than expecting (it's JVM after all), then suddenly there is OOM error, then you spend some time to find a leak, except there's no leaks, and only finally you find that memory is full of `fs2.Chunk.ArraySlice` of 1MiB each with only 1 first kilo written. 

To fix this currently we need to put custom `Pull` or something like `.chunkN(chunkSize).map(_.compact).unchunks` after _every_ `readInputStream`. :(

> I guess my only concern with this approach is that a single `Chunk` that is retained for a long time may still have a surprisingly large memory footprint due to a large backing array. I'm not sure how realistic this is in practice :)
> 
> To avoid that would require allocating an appropriately sized array and copying into that, as suggested in [#3106 (comment)](https://github.com/typelevel/fs2/issues/3106#issuecomment-1378887690).

I think situation of 'short data streams' if not less frequent then at least much more detectable than issue above. 

To replicate issue above you will need 1024 unconsumed streams of 1KiB each and I would say that it is more question if wrongly guessing median data size. 

Also, use of copy introduces some questions:

- What should we do in `unsafeReadInputStream`? Copy, write into buffer from zero index each time, write in circular pattern?
- What should be the ratio when we copy? 1:1000, 1:10, 1:2?
- With every copy we introduce additional allocation and pointer indirection. Yes, currently we allocate new `ArraySlice` too but at least backing array is the same. Maybe we could even optimise chunk `++` to detect concst of two nearest slices with common backing array (maybe `.compact` already does that?). 

I would argue that this implementation tries its best to translate user intention: 'write data from stream into continuous memory regions of `chunkSize` bytes'.

mpilquist:
I like this approach and I'm curious if we should do something similar for TCP sockets.

PR #3322, opened 2023-10-20
Implement Node.js stream `destroy` w/ `interruptWhen`

armanbilge:
A Node.js stream can be `destroy`ed with or without an error.

- https://nodejs.org/api/stream.html#writabledestroyerror
- https://nodejs.org/api/stream.html#readabledestroyerror

Previously when converting an `fs2.Stream` to a Node.js stream, destruction-with-no-error was handled by raising a special `StreamDestroyedException`. Instead, we should handle it as ordinary interruption via `interruptWhen`.

h/t @kubukoz for reporting


kubukoz:
@armanbilge can we have a release of this, please?

PR #3336, opened 2023-11-15
Remove JS TLS socket event listeners on finalise

mikeshepherd:
This fixes an issue we see when running this in AWS lambdas via Scala.js and feral.

The issue is that the lambdas can fail with `java.lang.IllegalStateException: dispatcher already shutdown`. This is coming from the event handler for an `error` event on a TLS socket. As this happens in a node error handler there is no way for our code to catch this exception.

For this to happen this error event must be processed by the callback after the finalizers for the `TLSSocket` Resource have run, as that is where the dispatchers will be shutdown.

I've extended the existing helpers for adding event listeners to provide Resource versions for `once` events and changed the upgrade thunk a `Resource` so these helpers can be used to add, and remove, the event listeners

### DIFF
--- a/io/js/src/main/scala/fs2/io/ioplatform.scala
+++ b/io/js/src/main/scala/fs2/io/ioplatform.scala
@@ -80,17 +80,17 @@ private[fs2] trait ioplatform {
       error <- F.deferred[Throwable].toResource
       readableResource = for {
         readable <- thunk
-        _ <- Resource.makeCase(F.unit) {
-          case (_, Resource.ExitCase.Succeeded) =>
+        _ <- Resource.onFinalizeCase[F] {
+          case Resource.ExitCase.Succeeded =>
             F.delay {
               if (!readable.readableEnded & destroyIfNotEnded)
                 readable.destroy()
             }
-          case (_, Resource.ExitCase.Errored(_)) =>
+          case Resource.ExitCase.Errored(_) =>
             // tempting, but don't propagate the error!
             // that would trigger a unhandled Node.js error that circumvents FS2/CE error channels
             F.delay(readable.destroy())
-          case (_, Resource.ExitCase.Canceled) =>
+          case Resource.ExitCase.Canceled =>
             if (destroyIfCanceled)
               F.delay(readable.destroy())
             else
--- a/io/js/src/main/scala/fs2/io/ioplatform.scala
+++ b/io/js/src/main/scala/fs2/io/ioplatform.scala
@@ -59,22 +59,38 @@ private[fs2] trait ioplatform {
       destroyIfNotEnded: Boolean = true,
       destroyIfCanceled: Boolean = true
   )(thunk: => R)(implicit F: Async[F]): Resource[F, (R, Stream[F, Byte])] =
+    suspendResourceReadableAndRead(destroyIfNotEnded, destroyIfCanceled)(
+      Resource.eval(F.delay(thunk))
+    )
+
+  /** Suspends the creation of a `Readable` and a `Stream` that reads all bytes from that `Readable`.
+    *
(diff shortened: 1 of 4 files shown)


mikeshepherd:
I've realised there's probably an issue with this approach, and the reason that the thunk wasn't a resource in the first place, which is the explanation in ioplatform about running on the MicrotaskExecutor. I think there should be a reasonable fix for this, it'll just be a bit bigger. Hang tight while I sort it

armanbilge:
> I've realised there's probably an issue with this approach, and the reason that the thunk wasn't a resource in the first place, which is the explanation in ioplatform about running on the MicrotaskExecutor.

Yes, exactly!! Good catch. I'll mark as a draft for now.

Thanks for the PR :)

io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -59,22 +59,38 @@ private[fs2] trait ioplatform {
...
+  )(thunk: Resource[F, R])(implicit F: Async[F]): Resource[F, (R, Stream[F, Byte])] =
     (for {
       dispatcher <- Dispatcher.sequential[F]
       channel <- Channel.unbounded[F, Unit].toResource
       error <- F.deferred[Throwable].toResource
       readableResource = for {
-        readable <- Resource.makeCase(F.delay(thunk)) {
-          case (readable, Resource.ExitCase.Succeeded) =>
+        readable <- thunk
+        _ <- Resource.makeCase(F.unit) {
```
mikeshepherd, line 83: This feels like a weird construction to me and that I must be missing a better way of achieving the same thing

armanbilge: I think this is `Resource.onFinalizeCase`?

armanbilge:
Would you be open to trying a more involved refactoring? I was never crazy about how this was working, and the problems seem to be compounding.

So here's what I propose: we introduce a new lifecycle concept of `Resource[F, EventEmitterScope]` with helpers like
```scala
def unsafeRegisterOneTimeListener[F[_], E](eventName: String, dispatcher: Dispatcher[F], scope: EventEmitterScope)(
    listener: E => F[Unit]
): Unit = /* register listener and add cleanup task to scope */
```

This way we can register multiple listeners in a single `delay` block and they can add their cleanup action to the `EventEmitterScope`. The `scope` would guarantee that the listeners get cleaned up without requiring that registration itself be a `Resource`. Then we would no longer need the `evalOn(MicrotaskExecutor)` hack.

io/js/src/main/scala/fs2/io/ioplatform.scala
mikeshepherd: Yeah that feels neater, thanks!

mikeshepherd:
>Would you be open to trying a more involved refactoring?

Yes and no.
I'm very happy to do that, though I might have to reach out separately to work out the finer details with you, but I'd like to get this merged in as well, if you are happy with it. Without this change there is the very real possibility that we'll switch our lambdas from JS to JVM, and then I'll have a lot less scope to work on a larger change.

Would you accept a promise that it'll be the very next thing I do if we get this change merged so it unblocks our lambdas sooner?

gvolpe:
@armanbilge I can act as witness to that promise if that's any guarantee 😅 

We work together with Mike and would like to continue using JS lambdas, but are close to losing the battle with this new issue.

armanbilge:
> if you are happy with it

Sorry, I appreciate your work, but I don't want to merge/release this PR as-is. I understand you don't have time to work on the larger refactoring I proposed, so I did it in https://github.com/typelevel/fs2/pull/3338.

mikeshepherd:
Totally understood. Thanks for getting your proposal out so quickly as well. Gonna close this out in favour of that

Issue #3138, opened 2023-02-12
API to directly open TLS connections, instead of upgrading an existing socket


armanbilge:
**Background**: currently to get a `TLSSocket`, you first create a non-encrypted `Socket`, then upgrade it via `TLSContext` to get a `TLSSocket`. This proposal is for new `TLSContext` APIs to directly create client/server `TLSSocket`s.

**Motivation**:

- On Native, s2n-tls is capable of interfacing directly with the underlying socket's file descriptor. This should have higher performance and is probably easier to implement than redirecting it through an FS2 `Socket`.

  Even more compelling, s2n-tls also supports using [Kernel TLS](https://docs.kernel.org/networking/tls.html), where all the encryption/decryption happens in kernel-space. This minimizes data copies.

- On JS, the upgrade is implemented by a costly Node.js Socket <-> FS2 Socket <-> Node.js Socket <-> Node.js TLS Socket <-> FS2 TLS Socket conversion. In Node.js it is possible to directly open TLS connections/servers directly, and thus skip a few layers of conversion.

I'm not aware of any motivation to do this on the JVM, at least with the JDK. Probably a hypothetical Netty-based `TLSContext` could take advantage of this.

**Questions**:
1. Could e.g. Ember and Skunk take advantage of such an API, instead of manually upgrading?
2. Alternatively, could we make these optimizations within the upgrade mechanism, without exposing a new API? Seems hard, since in theory the user could continue doing things with the raw `Socket` even after upgrading it. If we assume that won't happen, then maybe it can work.

armanbilge:
> On Native, s2n-tls is capable of interfacing directly with the underlying socket's file descriptor. This should have higher performance

There is a significant caveat to this: s2n-tls does not (and cannot without redesign) use io_uring for dispatching the reads/writes. So it might not have higher performance compared to our io_uring-powered sockets.

There's still a chance for kTLS to be compatible with io_uring, depending on the implementation strategy taken.
- https://github.com/aws/s2n-tls/issues/773

rahulrangers:
@armanbilge can you assign me to this issue as now s2n-tls has added support for kTLS.

armanbilge:
> s2n-tls has added support for kTLS.

@rahulrangers That's correct, however that support is unstable.

https://github.com/aws/s2n-tls/blob/785ed148882ecadb82aea18cc607b1badb3e6291/api/unstable/ktls.h#L27-L28

PR #3341, opened 2023-11-20
Upgrade a JS socket to TLS directly if possible

mikeshepherd:
This change alters how an upgrade to a TLS Socket is performed on JS runtimes.

Currently there is a lot of machinery to back and forth between native JS and FS2 constructs in order to perform the upgrade. 
On the client side it currently goes "JS Socket <-> FS2 Socket <-> JS Duplex <-> JS TLS Socket <-> FS2 Socket".

Since JS can perform this upgrade directly we can let it do so when we have access to an underlying JS socket.
This results in "JS Socket <-> JS TLS Socket <-> FS2 Socket" instead, with the associated removal of several layers of Queues and Channels involved in the conversions. Note that the non TLS FS2 socket remains as is, it just is no longer involved in the TLS upgrade.

This is related to #3138 as it performs the optimization described there in limited situations that we can access without changing the API.

Will also address one of the issues on #3338 related to use of closed dispatchers in the duplex stream, as we no longer use it.

### DIFF
--- a/io/js/src/main/scala/fs2/io/net/SocketPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/net/SocketPlatform.scala
@@ -53,7 +53,7 @@ private[net] trait SocketCompanionPlatform {
         }
       }
 
-  private[net] class AsyncSocket[F[_]](
+  private[net] case class AsyncSocket[F[_]](
       sock: facade.net.Socket,
       readStream: SuspendedStream[F, Byte]
   )(implicit F: Async[F])
--- a/io/js/src/main/scala/fs2/io/net/tls/TLSSocketPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/net/tls/TLSSocketPlatform.scala
@@ -38,16 +38,29 @@ private[tls] trait TLSSocketCompanionPlatform { self: TLSSocket.type =>
 
   private[tls] def forAsync[F[_]](
       socket: Socket[F],
-      upgrade: fs2.io.Duplex => facade.tls.TLSSocket
+      upgrade: fs2.io.Duplex => facade.tls.TLSSocket,
+      clientMode: Boolean
   )(implicit F: Async[F]): Resource[F, TLSSocket[F]] =
     for {
-      duplexOut <- mkDuplex(socket.reads)
-      (duplex, out) = duplexOut
-      _ <- out.through(socket.writes).compile.drain.background
-      tlsSockReadable <- suspendReadableAndRead(
-        destroyIfNotEnded = false,
-        destroyIfCanceled = false
-      )(upgrade(duplex))
+      tlsSockReadable <- socket match {
+        case Socket.AsyncSocket(sock, _) if clientMode =>
+          for {
+            tlsSockReadable <- suspendReadableAndRead(
+              destroyIfNotEnded = false,
+              destroyIfCanceled = false
+            )(upgrade(sock))
+          } yield tlsSockReadable
+        case _ =>
+          for {
+            duplexOut <- mkDuplex(socket.reads)
+            (duplex, out) = duplexOut
+            _ <- out.through(socket.writes).compile.drain.background
+            tlsSockReadable <- suspendReadableAndRead(
+              destroyIfNotEnded = false,
+              destroyIfCanceled = false
+            )(upgrade(duplex))
+          } yield tlsSockReadable
+      }
       (tlsSock, readable) = tlsSockReadable
       readStream <- SuspendedStream(readable)
     } yield new AsyncTLSSocket(
--- a/io/js/src/main/scala/fs2/io/net/tls/TLSContextPlatform.scala
+++ b/io/js/src/main/scala/fs2/io/net/tls/TLSContextPlatform.scala
@@ -89,7 +89,8 @@ private[tls] trait TLSContextCompanionPlatform { self: TLSContext.type =>
                             )
                         )
                         tlsSock
-                      }
+                      },
+                      clientMode = clientMode
                     )
                     .evalTap(_ => handshake.get.rethrow)
                 }
@@ -128,7 +129,8 @@ private[tls] trait TLSContextCompanionPlatform { self: TLSContext.type =>
                             )
                         )
                         tlsSock
-                      }
+                      },
+                      clientMode = clientMode
                     )
                     .evalTap(_ => verifyError.get.rethrow)
                 }


io/js/src/main/scala/fs2/io/net/tls/TLSSocketPlatform.scala
```diff
@@ -38,16 +38,29 @@ private[tls] trait TLSSocketCompanionPlatform { self: TLSSocket.type =>
...
     for {
-      duplexOut <- mkDuplex(socket.reads)
-      (duplex, out) = duplexOut
-      _ <- out.through(socket.writes).compile.drain.background
-      tlsSockReadable <- suspendReadableAndRead(
-        destroyIfNotEnded = false,
-        destroyIfCanceled = false
-      )(upgrade(duplex))
+      tlsSockReadable <- socket match {
+        case Socket.AsyncSocket(sock, _) if clientMode =>
```
mikeshepherd, line 46: In theory the same logic should work for server sockets as well, but it results in a hang. I think this is connected to the pause or readable state of the socket vs duplex but I can't figure out how to unpause the socket at the right moment to allow things to flow.

io/js/src/main/scala/fs2/io/net/tls/TLSSocketPlatform.scala
```diff
@@ -37,6 +37,18 @@ private[tls] trait TLSSocketPlatform[F[_]]
 private[tls] trait TLSSocketCompanionPlatform { self: TLSSocket.type =>
 
   private[tls] def forAsync[F[_]](
```
armanbilge, line 39: Should we rename this `forAsyncServer`?

armanbilge: Or couldn't we just pass `clientMode` in `forAsync` so we don't need separate methods? I don't really care, these are just nitpicks :)

armanbilge: Oh I see, it was bincompat. But this is all private anyway. We can just add the exclusion.

armanbilge: > Since it was fairly easy to avoid I did

Yeah, this is good practice so thanks for doing that. But in this case I think its cleaner like this.

io/js/src/main/scala/fs2/io/net/SocketPlatform.scala
```diff
@@ -53,7 +53,7 @@ private[net] trait SocketCompanionPlatform {
         }
       }
 
-  private[net] class AsyncSocket[F[_]](
+  private[net] case class AsyncSocket[F[_]](
       sock: facade.net.Socket,
       readStream: SuspendedStream[F, Byte]
```
armanbilge, line 58: I've been thinking about this. Do we know what happens to this `SuspendedStream` after the `sock` is directly upgraded?

What I mean is that this `readStream` has already established listeners on the `sock`. So if events are firing on those listeners as `sock` is being used, but nobody is consuming from the `SuspendedStream`, then I am concerned that this is actually a memory leak.

mikeshepherd: Good question.
As far as I can tell the listeners are all on control events, rather than data, i.e. they trigger changes of behaviour rather than push any data anywhere.

The read loop boils down to stream pull -> readable.read(), so if the stream isn't being consumed then `read() ` is never being called and there is no data to leak.

This also means that we're relying on the Javascript runtime to propagate any error or close events, which I think is a reasonable assumption, but we could add additional listeners for them if we feel that it's needed

armanbilge: > As far as I can tell the listeners are all on control events, rather than data, i.e. they trigger changes of behaviour rather than push any data anywhere.

Right, sure. But those triggers still invoke callbacks, that typically put things in `Queue`s. If they are not consumed, they accumulate and leak.

mikeshepherd: I don't think we're going to leak any data anywhere, but I guess we we'll end up with a channel full of `()`.

The only `Queue` that I can see in the process is `synchronous` so wont accumulate anything.

It seems like it ought to be possible for that channel not to be a channel since it's only acting as an indicator that something can be read or not (that is we need to communicate that something can be read, but duplicate events prior to a read are meaningless)

armanbilge: > but I guess we will end up with a channel full of `()`.

Yes, sorry, this is what I meant by "memory leak". Not specifically a data leak, as in bytes of data.

> It seems like it ought to be possible for that channel not to be a channel since it's only acting as an indicator that something can be read or not (that is we need to communicate that something can be read, but duplicate events prior to a read are meaningless)

This is a great point. I wonder if we can replace it with a `Signal`.

armanbilge: > The only `Queue` that I can see in the process is `synchronous` so wont accumulate anything.

Ah, this is a misconception. Unless whoever is pushing to the `synchronous` queue respect backpressure and stops trying to push, the attempts to push will queue up unboundedly.

mikeshepherd: Yes sorry, I actually thought that we wouldn't leak anything (including `()`). Had to reread the Node docs to correct myself about how the readable event actually works (I thought it was only fired once, rather than repeatedly).

I'm happy to switch over to a `Signal` here but I wonder if it's worth it with #3348 now which should solve the issue as well. Don't know if you've got any preference for handling related PRs like these?


> Unless whoever is pushing to the synchronous queue respect backpressure

I admit I'd assumed that the internals of fs2 would get that right 😉

armanbilge: > Had to reread the Node docs to correct myself about how the readable event actually works (I thought it was only fired once, rather than repeatedly).

It is _very_ confusing. While working on https://github.com/typelevel/fs2/pull/3348 I discovered that the `readable` event may be fired multiple times in a row (honestly it seemed a bit buggy) without `read()`s in between, which is why we needed a counter for readable events and not simply a boolean toggle ...

I think if you merge https://github.com/typelevel/fs2/pull/3348 into your PR, then my concerns about memory leaks would be addressed. But I need to think about this more. The "dangling" listeners are difficult to reason about.

Issue #3343, opened 2023-11-24
Fetching permissions through getPosixFileAttributes always fails with None.get


dantb:
This is JVM-specific - the bug is in `FilePlatform`.

E.g.

`Files[IO].getPosixFileAttributes(p).map(_.permissions)` 

Fix incoming, just adding an issue for the trail :)

Issue #3344, opened 2023-11-24
Fetched file attributes fail equality check


dantb:
For instance this assertion fails, which can make it confusing to troubleshoot real differences
```scala
val attr = Files[IO].getBasicFileAttributes(p)
(attr, attr).mapN(assertEquals(_, _))
```

Fix incoming.

mpilquist:
Fixed in #3345

PR #3345, opened 2023-11-24
Fix fetching permissions through getPosixFileAttributes and add equality

dantb:
Fixes https://github.com/typelevel/fs2/issues/3343 and https://github.com/typelevel/fs2/issues/3344


io/shared/src/main/scala/fs2/io/file/FileAttributes.scala
```diff
@@ -42,6 +42,20 @@ sealed trait BasicFileAttributes {
   def lastAccessTime: FiniteDuration
   def lastModifiedTime: FiniteDuration
   def size: Long
+
+  override def equals(that: Any): Boolean = that match {
```
mpilquist, line 46: Override `hashCode` and `toString` here too?

io/shared/src/main/scala/fs2/io/file/FileAttributes.scala
```diff
@@ -54,6 +68,11 @@ object BasicFileAttributes {
 // the owner/group operations JVM only.
 sealed trait PosixFileAttributes extends BasicFileAttributes {
   def permissions: PosixPermissions
+
+  final override def equals(that: Any): Boolean = that match {
```
mpilquist, line 72: `hashCode` and `toString` as well?

mpilquist:
Thanks! The mima warning can be safely added to exclusion list.

PR #3346, opened 2023-11-24
Move `j.u.c.Flow` interop methods onto Stream

armanbilge:
So that it Just Works™️  without the annoying `import fs2.interop.flow.syntax._`, which is now deprecated. cc @BalmungSan


core/shared/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -192,9 +192,8 @@ private[flow] object StreamSubscription {
 
   def subscribe[F[_], A](stream: Stream[F, A], subscriber: Subscriber[A])(implicit
       F: Async[F]
-  ): F[Unit] =
-    apply(stream, subscriber).flatMap { subscription =>
-      F.delay(subscriber.onSubscribe(subscription)) >>
-        subscription.run
+  ): Stream[F, Nothing] =
+    Stream.eval(apply(stream, subscriber)).flatMap { subscription =>
+      Stream.eval(F.delay(subscriber.onSubscribe(subscription))) >> subscription.run
```
BalmungSan, line 197: This should be `exec`, shouldn't it?

armanbilge: If it's exec, then `subscription.run` will never evaluate.

The two options are:
```scala
Stream.eval(...) >> subscription.run
Stream.exec(...) ++ subscription.run
```
The former felt more appropriate to me, it establishes the monadic dependence. Not sure what the subtle differences are.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4415,6 +4481,24 @@ object Stream extends StreamLowPriority {
...
+      * @note This [[Publisher]] can be reused for multiple [[Subscribers]],
+      *       each [[Subscription]] will re-run the [[Stream]] from the beginning.
+      *
+      * @see [[toPublisher]] for a safe version that returns a [[Stream]].
+      */
+    def unsafeToPublisher()(implicit
+        runtime: IORuntime
+    ): Publisher[A] =
+      interop.flow.unsafeToPublisher(self)
+  }
```
armanbilge, line 4500: This is the one I'm most interested in. It is the `Stream` analog of `io.unsafeToFuture()`.

The motivation to facilitate easy interop with other libraries e.g. https://github.com/raquo/Airstream/pull/114:
```scala
EventStream.fromFlowPublisher(fs2Stream.unsafeToPublisher())
```

BalmungSan:
A couple of random thoughts against this:

1. This is interop with a different _"library"_ so it makes sense to be on the `interop` package.
2. Is an elimination form, and folks think `Stream => IO` is the way, plus this one produces an _"impure"_ thing.
3. It requires **Java 9+**.
4. We are already used to syntax imports even if we hate them.

A couple of random thoughts in favor of this:

1. I really dislike syntax imports; this is one of the reasons I prefer to use `IO` over `F`.
2. API discoverability.
3. These may be very useful / common for folks doing interop with other libraries.
4. **Monix** had these as part of their `Obersavle` API.

armanbilge:
I think those are really good points, which is why I asked Luis to post them here. I think a reasonable compromise could be to keep `Stream#unsafeToPublisher` and `Stream.fromPublisher` since those are the easiest to use and cover the most important use-cases. The `interop.flow` package can continue to be used for everything else.

PR #3347, opened 2023-11-25
Fix subte bug in the interop.flow/StreamSubscription

BalmungSan:
It is unlikely anyone was affected by the bug, since they would have needed to have a `Stream` with more than `Long.MaxValue` elements for it to manifest.
Still, it is better to do the right thing.

c.c. @armanbilge


core/shared/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -65,6 +65,7 @@ private[flow] final class StreamSubscription[F[_], A] private (
       def go(s: Stream[F, A]): Pull[F, A, Unit] =
         Pull.eval(F.delay(requests.get())).flatMap { n =>
```
armanbilge, line 66: If we just do this, could we avoid the new loop condition?
```suggestion
        Pull.eval(F.delay(requests.getAndSet(0))).flatMap { n =>
```

armanbilge (approved):
Nice. bug fix + optimization 😘

PR #3348, opened 2023-11-26
Don't use `Dispatcher` for `suspendReadableAndRead` and TLS initialization on Node.js

armanbilge:
Fixes errors reported in https://github.com/typelevel/fs2/pull/3336 and https://github.com/typelevel/fs2/pull/3338.

By introducing some custom concurrent data structures, we can completely avoid `Dispatcher`, the indirection and performance cost, and the `dispatcher already shutdown` errors that have been plaguing us.

### DIFF
--- a/io/js/src/main/scala/fs2/io/ioplatform.scala
+++ b/io/js/src/main/scala/fs2/io/ioplatform.scala
@@ -60,22 +60,23 @@ private[fs2] trait ioplatform {
   )(thunk: => R)(implicit F: Async[F]): Resource[F, (R, Stream[F, Byte])] = {
 
     final class Listener {
-      private[this] var readableCounter = 0
+      private[this] var readable = false
       private[this] var error: Either[Throwable, Boolean] = null
       private[this] var ended = false
       private[this] var callback: Either[Throwable, Boolean] => Unit = null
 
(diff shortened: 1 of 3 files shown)


armanbilge:
@mikeshepherd @gvolpe Ok, here's another attempt 😅 I published `3.9.3-27-7874843-SNAPSHOT`.

mikeshepherd:
> @mikeshepherd @gvolpe Ok, here's another attempt 😅 I published `3.9.3-27-7874843-SNAPSHOT`.

Thanks @armanbilge, we'll give this a try and get back to you. But it looks good to me reading it :+1:

io/js/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -60,22 +60,23 @@ private[fs2] trait ioplatform {
   )(thunk: => R)(implicit F: Async[F]): Resource[F, (R, Stream[F, Byte])] = {
 
     final class Listener {
-      private[this] var readableCounter = 0
+      private[this] var readable = false
```
armanbilge, line 63: I could have sworn this didn't work initially 🤔 but seems fine now.

PR #3350, opened 2023-11-28
Update evalTapChunk

kamilkloch:
Avoid calling `evalMapChunk`.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1052,8 +1052,10 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
 
   /** Alias for `evalMapChunk(o => f(o).as(o))`.
     */
-  def evalTapChunk[F2[x] >: F[x]: Applicative, O2](f: O => F2[O2]): Stream[F2, O] =
-    evalMapChunk(o => f(o).as(o))
+  def evalTapChunk[F2[x] >: F[x]: Applicative, O2](f: O => F2[O2]): Stream[F2, O] = {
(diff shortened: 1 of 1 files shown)


peterneyens:
This looks similar to @diesalbla's https://github.com/typelevel/fs2/pull/3197.

kamilkloch:
> This looks similar to @diesalbla's #3197.

We have a benchmark here :)

diesalbla (approved):

diesalbla:
It's similar, but @kamilkloch commit is more focused.

diesalbla:
@kamilkloch I remembered that the `evalMap` could not be directly written to an `unconsFlatMap`. There are some differences between the `flatMap` and the `unconsFlatMap`, related to how the latter treats interruption. 

In the PR I had, the code I was using was this: 

```
    underlying.unconsFlatMap(Pull.output1).scoped.flatMapOutput(onChunk).streamNoScope
```

This is more complicated, but needed to preserve that semantics. Could you try using that line on your PR and see if the benchmarks still show it worth it, please?

kamilkloch:
> ```
>     underlying.unconsFlatMap(Pull.output1).scoped.flatMapOutput(onChunk).streamNoScope
> ```
Does it compile? I get stuck at `.scoped`.

armanbilge:
> There are some differences between the `flatMap` and the `unconsFlatMap`, related to how the latter treats interruption.

Can we add a test for `evalTapChunk` covering this case? It's clearly subtle and easy to make a regression.

kamilkloch:
> Can we add a test for `evalTapChunk` covering this case? It's clearly subtle and easy to make a regression.

By all means, but I am afraid I would need some rope here (scopes are still an urchartered territory for me). @diesalbla, do you have and idea for a test?

PR #3354, opened 2023-11-30
FIX: Handle root path correctly

### LINKED ISSUE #3353: Root `Path` is incorrectly handled in ScalaJS
fs2 version: `3.9.3`

Reproduction example:
```scala
import fs2.io.file.Path

val p = Path("/")
println(p.normalize.absolute)
```
Outputs current working directory instead of `/` when run in JS.

Happens because of this code:
https://github.com/typelevel/fs2/blob/53f458d6ac4044df2b8e2d4310b92a1fb8953d80/io/js/src/main/scala/fs2/io/file/Path.scala#L102-L105

InversionSpaces:
Drop trailing separators in `Path` creation only if it is not the only thing in the path.

Fixes #3353

### DIFF
--- a/io/js/src/main/scala/fs2/io/file/Path.scala
+++ b/io/js/src/main/scala/fs2/io/file/Path.scala
@@ -100,8 +100,13 @@ object Path extends PathCompanionApi {
   private[file] val sep = facade.path.sep
 
   private[file] def dropTrailingSep(path: String): String = {
+    /*
+     * NOTE: It seems like scala js does not rewrite this to a loop?
+     * Call stack limit is reached if there are too many separators.
+     */
     @tailrec
     def drop(view: IndexedSeqView[Char]): IndexedSeqView[Char] =
+      // Drop separator only if there is something else left
       if (view.endsWith(sep) && view.length > sep.length)
         drop(view.dropRight(sep.length))
       else view
@@ -109,6 +114,10 @@ object Path extends PathCompanionApi {
     drop(path.view).mkString
   }
 
+  /** This method drops trailing separators to match
+    * `java.nio.file.Path.get` behaviour.
+    * But root ("/") is untouched.
+    */
   def apply(path: String): Path =
     new Path(dropTrailingSep(path))
 }
--- a/io/js/src/main/scala/fs2/io/file/Path.scala
+++ b/io/js/src/main/scala/fs2/io/file/Path.scala
@@ -26,6 +26,7 @@ package file
 import fs2.io.internal.facade
 
 import scala.annotation.tailrec
+import scala.collection.IndexedSeqView
 
 final case class Path private[file] (override val toString: String) extends PathApi {
 
@@ -98,9 +99,16 @@ final case class Path private[file] (override val toString: String) extends Path
 object Path extends PathCompanionApi {
   private[file] val sep = facade.path.sep
 
+  private[file] def dropTrailingSep(path: String): String = {
+    @tailrec
+    def drop(view: IndexedSeqView[Char]): IndexedSeqView[Char] =
+      if (view.endsWith(sep) && view.length > sep.length)
+        drop(view.dropRight(sep.length))
+      else view
+
+    drop(path.view).mkString
+  }
+
   def apply(path: String): Path =
-    if (path.endsWith(sep))
-      new Path(path.dropRight(sep.length))
-    else
-      new Path(path)
+    new Path(dropTrailingSep(path))
 }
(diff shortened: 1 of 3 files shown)


io/js/src/main/scala/fs2/io/file/Path.scala
```diff
@@ -98,9 +98,21 @@ final case class Path private[file] (override val toString: String) extends Path
...
+  /*
+   * NOTE: It seems like scala js does not rewrite this to a loop?
+   * Call stack limit is reached if there are too many separators.
+   */
+  @tailrec
+  private[file] def dropTrailingSep(path: String): String =
+    // Drop separator only if there is something else left
+    if (path.endsWith(sep) && path.length > sep.length)
+      dropTrailingSep(path.dropRight(sep.length))
+    else path
```
armanbilge, line 110: It does actually 🤔  Here's what it's compiled to:

```js
$c_Lfs2_io_file_Path$.prototype.dropTrailingSep__T__T = (function(path) {
  while (true) {
    if ($f_T__endsWith__T__Z($n(path), this.Lfs2_io_file_Path$__f_sep)) {
      var this$1 = $n(path);
      var this$2 = $n(this.Lfs2_io_file_Path$__f_sep);
      var $$x2 = (this$1.length > this$2.length)
    } else {
      var $$x2 = false
    };
    if ($$x2) {
      var $$x1 = $m_sc_StringOps$();
      var x = path;
      var this$4 = $n(this.Lfs2_io_file_Path$__f_sep);
      path = $n($$x1).dropRight$extension__T__I__T(x, this$4.length)
    } else {
      break
    }
  };
  return path
});
```

armanbilge: This passes for me. How were you getting the failure?
```scala
assertEquals(Path("/".repeat(Int.MaxValue/8)), Path("/"))
```

io/shared/src/test/scala/fs2/io/file/PathSuite.scala
```diff
@@ -44,6 +44,20 @@ class PathSuite extends Fs2IoSuite {
   test("construction") {
     assertEquals(Path("foo/bar"), Path("foo") / "bar")
     assertEquals(Path("/foo/bar"), Path("/foo") / "bar")
+    forAll((sepLength: Int) =>
+      if (sepLength >= 0 && sepLength <= 100)
```
armanbilge, line 48: These `if` guards are probably skipping most/all of the generated values. Instead you should do something like this.
```suggestion
    forAll(Gen.choose(0, 100))((sepLength: Int) =>
```

io/js/src/main/scala/fs2/io/file/Path.scala
```diff
@@ -98,9 +98,21 @@ final case class Path private[file] (override val toString: String) extends Path
...
   private[file] val sep = facade.path.sep
 
+  /*
+   * NOTE: It seems like scala js does not rewrite this to a loop?
+   * Call stack limit is reached if there are too many separators.
+   */
+  @tailrec
+  private[file] def dropTrailingSep(path: String): String =
+    // Drop separator only if there is something else left
+    if (path.endsWith(sep) && path.length > sep.length)
```
armanbilge, line 108: I don't like to Officially:tm: support Windows 🤮 but its path can look like `C:\` which I think would get re-written to `C:`. I have no idea if that's good or bad 😅

BalmungSan: Doing some tests on **Powershell**

```
PS C:\Users\PC> cd c:
PS C:\Users\PC> cd C:
PS C:\Users\PC> cd C:\
PS C:\>
```

io/js/src/main/scala/fs2/io/file/Path.scala
InversionSpaces: Strange, I wrote this comment when I used `IndexedSeqView` instead of `String`.

io/js/src/main/scala/fs2/io/file/Path.scala
InversionSpaces: Used node API exclusively

io/shared/src/test/scala/fs2/io/file/PathSuite.scala
```diff
@@ -38,12 +38,31 @@ class PathSuite extends Fs2IoSuite {
...
     assertEquals(Path("/foo/bar"), Path("/foo") / "bar")
+
+    val sepLengthGen = Gen.choose(0, 1024)
+
+    forAll(sepLengthGen)((sepLength: Int) =>
+      assertEquals(
+        Path("/".repeat(sepLength)).toString,
+        "/"
+      )
+    )
```
armanbilge, line 59: Wait a minute, is this test passing even when `sepLength = 0`?

InversionSpaces: No, it should not. After investigation I found out that scalacheck is mishandled in this suite. Tried to fix that

armanbilge (approved):
thank you!

PR #3360, opened 2023-12-13
Call Subscriber.onSubscribe(StreamSubscription) before retruning from StreamPublisher.subscribe(Subscriber)

### LINKED ISSUE #3359: Error using reactive streams toStreamBuffered with newer versions of the AWS SDK
I have a method which uses the S3 transfer manager to return a publisher:
```scala
  def download(bucket: String, key: String): IO[Publisher[ByteBuffer]] = {
    val getObjectRequest = GetObjectRequest.builder.key(key).bucket(bucket).build
    val downloadRequest = DownloadRequest.builder
      .getObjectRequest(getObjectRequest)
      .responseTransformer(AsyncResponseTransformer.toPublisher[GetObjectResponse])
      .build()
    val completableFuture = transferManager
      .download(downloadRequest)
      .completionFuture()
    IO.fromCompletableFuture(IO(completableFuture)).map(_.result())
  }
```

I'm then calling this and calling `toStreamBuffered`
```scala
s3.download(downloadBucket, downloadBucketKey)
      .flatMap(
        _.toStreamBuffered[IO](10 * 1024)
```

With recent (I'm not 100% sure which one) versions of the AWS SDK, I'm now getting the following error:
```
Exception in thread "Thread-71" java.lang.NullPointerException: Cannot invoke "org.reactivestreams.Subscription.request(long)" because "this.subscription" is null
	at software.amazon.awssdk.utils.async.ByteBufferStoringSubscriber.maybeRequestMore(ByteBufferStoringSubscriber.java:205)
	at software.amazon.awssdk.utils.async.ByteBufferStoringSubscriber.addBufferedDataAmount(ByteBufferStoringSubscriber.java:200)
	at software.amazon.awssdk.utils.async.ByteBufferStoringSubscriber.transferTo(ByteBufferStoringSubscriber.java:102)
	at software.amazon.awssdk.services.s3.internal.crt.S3CrtRequestBodyStreamAdapter.sendRequestBody(S3CrtRequestBodyStreamAdapter.java:47)

```

SDK version 2.20.69 is working but 2.21.42 is not. The only recent change in the SDK I can see is https://github.com/aws/aws-sdk-java-v2/pull/4308
If I revert this change locally and run my code again, it's working so it looks like that change to the AWS SDK is at least partly responsible. 

Do you know what is causing this error or if there are any workarounds I can use?

Thanks

BalmungSan:
Fixes #3359 ?

c.c. @armanbilge as always I couldn't resit and made some renames along the way, sorry 😅

### DIFF
--- a/core/shared/src/main/scala/fs2/interop/flow/StreamSubscription.scala
+++ b/core/shared/src/main/scala/fs2/interop/flow/StreamSubscription.scala
@@ -173,10 +173,10 @@ private[flow] final class StreamSubscription[F[_], A] private (
 private[flow] object StreamSubscription {
   private final val Sentinel = () => ()
 
-  // SIDE-EFFECTING!
-  // We need this constructor to don't be wrapped on F,
-  // Because some consumers expect we call Subscriber.onSubscribe(StreamSubscription)
-  // Before returning from StreamPublisher.subscribe(Subscriber)
+  // UNSAFE + SIDE-EFFECTING!
+  // We cannot wrap this constructor in F[_].
+  // Some consumers expect we call Subscriber.onSubscribe(StreamSubscription)
+  //   before returning from StreamPublisher.subscribe(Subscriber).
   // See https://github.com/typelevel/fs2/issues/3359
   def apply[F[_], A](
       stream: Stream[F, A],
--- a/core/shared/src/main/scala/fs2/interop/flow/StreamSubscription.scala
+++ b/core/shared/src/main/scala/fs2/interop/flow/StreamSubscription.scala
@@ -39,7 +39,7 @@ import java.util.concurrent.atomic.{AtomicLong, AtomicReference}
   */
 private[flow] final class StreamSubscription[F[_], A] private (
     stream: Stream[F, A],
-    sub: Subscriber[A],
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -173,28 +173,37 @@ private[flow] final class StreamSubscription[F[_], A] private (
 private[flow] object StreamSubscription {
   private final val Sentinel = () => ()
 
-  // Mostly for testing purposes.
-  def apply[F[_], A](stream: Stream[F, A], subscriber: Subscriber[A])(implicit
+  // SIDE-EFFECTING!
+  // We need this constructor to don't be wrapped on F,
+  // Because some consumers expect we call Subscriber.onSubscribe(StreamSubscription)
+  // Before returning from StreamPublisher.subscribe(Subscriber)
+  // See https://github.com/typelevel/fs2/issues/3359
```
BalmungSan, line 180: Feedback or better wording is welcome :)

armanbilge (approved):

PR #3363, opened 2023-12-14
Add `List[Stream]#parJoinUnbounded`.

kamilkloch:
`List[Stream]#parJoinUnbounded` is `7x` faster compared to `Stream[Stream]#parJoinUnbounded`:
```
[info] Benchmark                                  Mode  Cnt    Score    Error  Units
[info] ConcurrentBenchmark.listParJoinUnbounded  thrpt    5  489.179 ± 35.120  ops/s
[info] ConcurrentBenchmark.parJoinUnbounded      thrpt    5   66.381 ±  4.879  ops/s
```

```scala
  @Benchmark
  def parJoinUnbounded(): Int = {
    val each = Stream
      .range(0, 1000)
      .map(i => Stream.eval(IO.pure(i)))
    each.parJoinUnbounded.compile.last.unsafeRunSync().get
  }

  @Benchmark
  def listParJoinUnbounded(): Int = {
    val each = List
      .range(0, 1000)
      .map(i => Stream.eval(IO.pure(i)))
    each.parJoinUnbounded.compile.last.unsafeRunSync().get
  }
```

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4214,9 +4214,10 @@ object Stream extends StreamLowPriority {
       * See [[NestedStreamOps.parJoinUnbounded]] for a strictly more powerful (albeit slower) variant
       * capable of merging a stream of streams.
       */
-    def parJoinUnbounded(implicit F: Concurrent[F]): Stream[F, O] =
-      if (xs.sizeIs <= 1) xs.headOption.getOrElse(Stream.empty)
-      else {
+    def parJoinUnbounded(implicit F: Concurrent[F]): Stream[F, O] = xs match {
+      case Nil         => Stream.empty
+      case head :: Nil => head
+      case _ =>
         Stream.eval(Channel.bounded[F, Chunk[O]](64)).flatMap { c =>
           val outcomes = xs
             .parTraverse_(_.chunks.foreach(x => c.send(x).void).compile.drain)
@@ -4225,7 +4226,7 @@ object Stream extends StreamLowPriority {
           Stream
             .bracket(F.start(outcomes))(f => f.cancel >> f.joinWithUnit) >> c.stream.unchunks
         }
-      }
+    }
   }
 
   /** Provides syntax for streams of streams. */
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4196,6 +4196,18 @@ object Stream extends StreamLowPriority {
     }
   }
 
+  /** Provides syntax for list of streams. */
+  implicit final class ListStreamOps[F[_], O](private val xs: List[Stream[F, O]]) extends AnyVal {
+
+    def parJoinUnbounded(implicit F: Concurrent[F]): Stream[F, O] =
+      Stream.eval(Channel.bounded[F, Chunk[O]](64)).flatMap { c =>
+        val producers = xs.parTraverse_(_.chunks.foreach(x => c.send(x).void).compile.drain)
+        c.stream
+          .concurrently(Stream.exec(producers *> c.close.void))
+          .unchunks
+      }
+  }
+
   /** Provides syntax for streams of streams. */
   implicit final class NestedStreamOps[F[_], O](private val outer: Stream[F, Stream[F, O]])
       extends AnyVal {
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4215,7 +4215,7 @@ object Stream extends StreamLowPriority {
       * capable of merging a stream of streams.
       */
     def parJoinUnbounded(implicit F: Concurrent[F]): Stream[F, O] =
-      if (xs.size <= 1) xs.headOption.getOrElse(Stream.empty)
+      if (xs.sizeIs <= 1) xs.headOption.getOrElse(Stream.empty)
       else {
         Stream.eval((Channel.bounded[F, Chunk[O]](64), F.deferred[Unit]).tupled).flatMap {
           case (c, stopPublishers) =>
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -4200,11 +4200,19 @@ object Stream extends StreamLowPriority {
   implicit final class ListStreamOps[F[_], O](private val xs: List[Stream[F, O]]) extends AnyVal {
 
     def parJoinUnbounded(implicit F: Concurrent[F]): Stream[F, O] =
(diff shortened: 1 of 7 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4196,6 +4196,41 @@ object Stream extends StreamLowPriority {
...
+      * concurrently with other stream computations.
+      *
+      * Finalizers on the output stream are run after the output stream has finished
+      * (i.e., all open inner streams have finished).
+      *
+      * See [[NestedStreamOps.parJoinUnbounded]] for a strictly more powerful (albeit slower) variant
+      * capable of merging a stream of streams.
+      */
+    def parJoinUnbounded(implicit F: Concurrent[F]): Stream[F, O] =
+      if (xs.size <= 1) xs.headOption.getOrElse(Stream.empty)
```
armanbilge, line 4218: Calculating the size of a `List` is `O(n)`.

```suggestion
      if (xs.sizeIs <= 1) xs.headOption.getOrElse(Stream.empty)
```

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4196,6 +4196,41 @@ object Stream extends StreamLowPriority {
...
+      else {
+        Stream.eval((Channel.bounded[F, Chunk[O]](64), F.deferred[Unit]).tupled).flatMap {
+          case (c, stopPublishers) =>
+            val outcomes = xs
+              .parTraverse_(_.chunks.foreach(x => c.send(x).void).compile.drain)
+              .guarantee(c.close.void)
+
+            Stream
+              .bracket(F.start(outcomes.race(stopPublishers.get).void))(fiber =>
+                stopPublishers.complete(()) >> fiber.joinWithUnit
```
armanbilge, line 4228: I'm confused what the point of `stopPublishers` is. Can't we just directly do `fiber.cancel`?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4196,6 +4196,41 @@ object Stream extends StreamLowPriority {
...
+      * Finalizers on the output stream are run after the output stream has finished
+      * (i.e., all open inner streams have finished).
+      *
+      * See [[NestedStreamOps.parJoinUnbounded]] for a strictly more powerful (albeit slower) variant
+      * capable of merging a stream of streams.
+      */
+    def parJoinUnbounded(implicit F: Concurrent[F]): Stream[F, O] =
+      if (xs.size <= 1) xs.headOption.getOrElse(Stream.empty)
+      else {
+        Stream.eval((Channel.bounded[F, Chunk[O]](64), F.deferred[Unit]).tupled).flatMap {
```
armanbilge, line 4220: Is this 64 arbitrary?

kamilkloch: It is. `Stream#parJoin` uses `Channel.synchronous`, `Channel.bounded` gives s speed bump, and I could not find a test procedure which distincts the two.

core/shared/src/main/scala/fs2/Stream.scala
kamilkloch: I believe you are right, thank you!

core/shared/src/main/scala/fs2/Stream.scala
kamilkloch: I tried the snippet and it never fails. With the removed `.timeoutTo` the results are as follows:

```scala
      IO.ref(0).flatMap { ref =>
        Stream(
          Stream.repeatEval(ref.getAndUpdate(_ + 1).void),
          Stream.empty,
        ).parJoinUnbounded.head.compile.drain >> ref.get.assertEquals(1)
      }
```
fails with `1 != 2`,
```scala
      IO.ref(0).flatMap { ref =>
        List(
          Stream.repeatEval(ref.getAndUpdate(_ + 1).void),
          Stream.empty,
        ).parJoinUnbounded.head.compile.drain >> ref.get.assertEquals(1)
      }
```
fails with `1 != 5`.

armanbilge: > I tried the snippet and it never fails. With the removed `.timeoutTo`

Oh sorry, you're right, the `timeoutTo` doesn't make sense :)

In any case, it seems like it should be `assertEquals(2)` then. Does your `List` version _always_ complete with `5`? Seems like a race condition so I expect you to get inconsistent values.

kamilkloch: Yes, 5 is arbitrary. If  it must return 2, we shall revert to Channel.synchronous. And add proper tests :)

kamilkloch: @armanbilge I switched to `Channel.synchronous` and added the tests. Still, the results are flaky - `ref.get` returns `1` or `2` (even for `Stream#parJoinUnbounded`). Both tests assert `ref.get <= 2`.

kamilkloch: @armanbilge what do you think in general?

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -4196,6 +4196,37 @@ object Stream extends StreamLowPriority {
     }
   }
 
+  /** Provides syntax for list of streams. */
+  implicit final class ListStreamOps[F[_], O](private val xs: List[Stream[F, O]]) extends AnyVal {
+
+    /** Nondeterministically merges a (static) list of streams in to a single output stream.
```
armanbilge, line 4202: Question: should
```scala
List(a, b, c).parJoinUnbounded
```
be semantically equivalent to
```scala
a.merge(b).merge(c)
```
?

If not, what's the difference?

If so, then I think the implementation should look more like `merge`.

<details>

https://github.com/typelevel/fs2/blob/f6702f0de74202af7272874ce3a18d0741f8cca7/core/shared/src/main/scala/fs2/Stream.scala#L1904-L2005

</details>

Also, in that case I wonder if `parJoinUnbounded` is the right name. What if we exposed this API like:
```scala
object Stream {
  def merge[F[_], A](streams: Stream[F, A]*): Stream[F, A] = ???
}
```

or `mergeAll` or `mergeMany` or something?

kamilkloch: That is a good question, I would go back one step and ask if
```scala
Stream(a, b).parJoinUnbounded
```
is semantically equivalent to
```scala
a.merge(b)
```
?
Citing the `merge` scaladoc:
```
Note that even when this is equivalent to `Stream(this, that).parJoinUnbounded`, this implementation is little more efficient
```
Looking at the implementation, I am not sure, if it is equivalent. `merge` adds some fairness with regard to chunks? (Also, as for the performance, simple benchmarks hint that the opposite is actually true - `merge` is slower compared to `parJoinUnbounded`).

PR #3364, opened 2023-12-20
Avoid double evaluation of pattern matchers in Chunk.collect

flipp5b:
`fs2.Chunk#collect` evaluates pattern matchers and guards of a passed partial function twice because it calls `pf.isDefinedAt` and then `pf.apply`:
```scala
// ...
foreach(o => if (pf.isDefinedAt(o)) b += pf(o))
// ...
```
Scala collections use `pf.applyOrElse`/`pf.runWith` (depending on scala version) instead to avoid double evaluation. This PR implements the same approach in a `Chunk`.

It should be noted that this may change observable behavior for those who use a partial function with side-effecting matchers/guards. But hey, side-effecting matchers/guards is considered a bad practice anyway, and I doubt that there's an fs2 user relying on that double evaluation "feature."

### DIFF
--- a/core/shared/src/main/scala/fs2/Chunk.scala
+++ b/core/shared/src/main/scala/fs2/Chunk.scala
@@ -87,7 +87,8 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
   def collect[O2](pf: PartialFunction[O, O2]): Chunk[O2] = {
     val b = makeArrayBuilder[Any]
     b.sizeHint(size)
-    foreach(o => if (pf.isDefinedAt(o)) b += pf(o))
+    val f = pf.runWith(b += _)
+    foreach { o => f(o); () }
     Chunk.array(b.result()).asInstanceOf[Chunk[O2]]
   }
 
--- a/benchmark/src/main/scala/fs2/benchmark/ChunkBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/ChunkBenchmark.scala
@@ -46,4 +46,14 @@ class ChunkBenchmark {
     ints.filter(_ % 3 == 0)
     ()
   }
+
+  private object OddStringExtractor {
+    def unapply(i: Int): Option[String] = if (i % 2 != 0) Some(i.toString) else None
+  }
+
+  @Benchmark
+  def collect(): Unit = {
+    ints.collect { case OddStringExtractor(s) => s }
+    ()
+  }
 }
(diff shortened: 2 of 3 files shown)


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -87,7 +87,8 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
   def collect[O2](pf: PartialFunction[O, O2]): Chunk[O2] = {
     val b = makeArrayBuilder[Any]
     b.sizeHint(size)
-    foreach(o => if (pf.isDefinedAt(o)) b += pf(o))
+    val f = pf.runWith(b += _)
+    foreach { o => f(o); () }
```
flipp5b, line 91: Well, this looks rather ugly. `pf.runWith` returns `O => Boolean` so we need to transform it to `O => Unit` to avoid warnings. This could be rewritten as follows (at the cost of extra lambda call):
```scala
foreach(pf.runWith(b += _).andThen(_ => ()))
```
@armanbilge, which one would you prefer?

armanbilge: Ugly but performant is the right choice 😅

flipp5b:
Here is some benchmark results.

### 1. Master
```scala
foreach(o => if (pf.isDefinedAt(o)) b += pf(o))
```
```
Benchmark               (chunkSize)   Mode  Cnt        Score       Error  Units
ChunkBenchmark.collect           16  thrpt   25  2828318.069 ± 72556.887  ops/s
ChunkBenchmark.collect          256  thrpt   25   191183.476 ±   581.751  ops/s
ChunkBenchmark.collect         4096  thrpt   25    11461.704 ±   504.370  ops/s
```

### 2. This PR
```scala
    val f = pf.runWith(b += _)
    foreach { o => f(o); () }
```
```
Benchmark               (chunkSize)   Mode  Cnt        Score       Error  Units
ChunkBenchmark.collect           16  thrpt   25  3373807.015 ± 25730.144  ops/s
ChunkBenchmark.collect          256  thrpt   25   257101.963 ±  7552.952  ops/s
ChunkBenchmark.collect         4096  thrpt   25    15417.548 ±   333.196  ops/s
```

### 3. This PR (prettified)
```scala
foreach(pf.runWith(b += _).andThen(_ => ()))
```
```
Benchmark               (chunkSize)   Mode  Cnt        Score       Error  Units
ChunkBenchmark.collect           16  thrpt   25  3420711.048 ± 90333.134  ops/s
ChunkBenchmark.collect          256  thrpt   25   242847.544 ±  4885.911  ops/s
ChunkBenchmark.collect         4096  thrpt   25    14369.581 ±   442.678  ops/s
```

Option 2 (this PR) shows slightly better performance than the other two options.

armanbilge (approved):
Thank you for the very thorough PR, this is very nice work 🙏

flipp5b:
@armanbilge Thank you for the prompt and kind feedback :)

PR #3367, opened 2023-12-31
enforce new wildcard syntax

xuwei-k:
- Scala 3 report warnings since 3.4 https://github.com/lampepfl/dotty/pull/18813
- Scala 2.12 and 2.13 support new syntax without `-Xsource:3` option https://github.com/scala/scala/pull/10005


He-Pin:
@xuwei-k hello, would you like to do the same for pekko, thank you very much.

PR #3374, opened 2024-01-17
feat: add keepAlive method in Stream.

fmeriaux:
Hello, 

Here is a modest contribution to add a keepAlive on streams, very useful when handling TCP connections with proxies that have timeouts.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -765,13 +765,13 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     */
   def keepAlive[F2[x] >: F[x]: Temporal, O2 >: O](
       maxIdle: FiniteDuration,
-      heartbeat: O2
+      heartbeat: () => O2
   ): Stream[F2, O2] =
     covaryAll[F2, O2].pull.timed { timedPull =>
       def go(timedPull: Pull.Timed[F2, O2]): Pull[F2, O2, Unit] =
         timedPull.timeout(maxIdle) >> timedPull.uncons.flatMap {
           case Some((Right(chunks), next)) => Pull.output(chunks) >> go(next)
-          case Some((Left(_), chunks))     => Pull.output1(heartbeat) >> go(chunks)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -760,6 +760,24 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   def delayBy[F2[x] >: F[x]: Temporal](d: FiniteDuration): Stream[F2, O] =
     Stream.sleep_[F2](d) ++ this
 
+  /** Ensure that the stream always emits elements by defining a maxIdle duration.
+    * In other words, the stream will emit an element when it hasn't emitted any since the maximum time specified.
+    */
+  def keepAlive[F2[x] >: F[x]: Temporal, O2 >: O](
+      maxIdle: FiniteDuration,
+      heartbeat: O2
```
He-Pin, line 768: Better to be a function.

fmeriaux: You suggest, update to : `heartbeat: => O2` ?

fmeriaux:
Do you have any idea why I have undefined entries in my test? 

```
=> Obtained
List(
  1,
  2,
  0,
  undefined,
  3,
  4,
  0,
  0,
  undefined,
  5
)
=> Diff (- obtained, + expected)
   0,
-  undefined,
   3,
   0,
-  undefined,
   5
```

core/shared/src/main/scala/fs2/Stream.scala
He-Pin: Yes, a user may provide different heart beat messages, iirc, pekko akka stream is act like that.

fmeriaux: I thought about it but couldn't figure out how to make a Pull.output from a F effect.

mpilquist: `Pull.eval(o2).flatMap(o => Pull.output1(o))`

mpilquist:
Thanks!

PR #3379, opened 2024-01-19
feat: Add Chunk#collectWhile.

He-Pin:
Motivation:
Add a collectWhile method to chunk


core/shared/src/main/scala/fs2/Chunk.scala
```diff
@@ -87,8 +87,31 @@ abstract class Chunk[+O] extends Serializable with ChunkPlatform[O] with ChunkRu
   def collect[O2](pf: PartialFunction[O, O2]): Chunk[O2] = {
     val b = makeArrayBuilder[Any]
     b.sizeHint(size)
-    val f = pf.runWith(b += _)
-    foreach { o => f(o); () }
+    var i = 0
+    while (i < size) {
+      import fs2.Chunk.NotApplied
+      pf.applyOrElse(apply(i), NotApplied) match {
```
He-Pin, line 93: this should be faster than pf.runWith.

PR #3383, opened 2024-02-04
Improve performance of Files.walk on the JVM

### LINKED ISSUE #3329: Bad performance of fs2.io.file.Files[IO].walk()
I am migrating a tool to use fs2 (v3.9.2). 
The first step is just to collect about 26000 files in a directory tree for further processing.

Here are the timings of this first step for several solutions:

- `java.nio.file.Files#walkFileTree`: 0.5s 
- `DeprecatedFilesApi.walk`: 0.7s
- `fs2.io.file.Files#walk`: 4s

mpilquist:
Fixes #3329.

For small walks, the overhead of the fs2/ce machinery dominates. For large walks, fs2's performance is within ~25% or so of the jvm's performance. For example, using @djspiewak's scenario with `MaxDepth = 7', I get:

```
fs2 took: 6600 ms
nio took: 5291 ms
```

### DIFF
--- a/io/jvm/src/test/scala/fs2/io/file/WalkBenchmark.scala
+++ b/io/jvm/src/test/scala/fs2/io/file/WalkBenchmark.scala
@@ -1,0 +1,83 @@
+
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2
+package io
+package file
+
+import cats.effect.IO
+import java.io.File
+import scala.concurrent.duration.*
+
+class WalkBenchmark extends Fs2IoSuite {
+
+  private var target: Path = _
+
+  override def beforeAll() = {
+		super.beforeAll()
+    val file = File.createTempFile("fs2-benchmarks-", "-walk")
+    file.delete()
+    file.mkdir()
+    target = Path(file.toString)
+
+    val MaxDepth = 7
+    val Names = 'A'.to('E').toList.map(_.toString)
+
+    def loop(cwd: File, depth: Int): Unit = {
+      if (depth < MaxDepth) {
+        Names foreach { name =>
+          val sub = new File(cwd, name)
+          sub.mkdir()
+          loop(sub, depth + 1)
+        }
+      } else if (depth == MaxDepth) {
+        Names foreach { name =>
+          val sub = new File(cwd, name)
+          sub.createNewFile()
+          loop(sub, depth + 1)
+        }
+      }
+    }
+
+    loop(file, 0)
+  }
+
+  def time[A](f: => A): FiniteDuration = {
+    val start = System.nanoTime()
+    val _ = f
+    (System.nanoTime() - start).nanos
+  }
+
+
+	test("Files.walk has similar performance to java.nio.file.Files.walk") {
+    val fs2Time = time(Files[IO]
+      .walk(target)
+      .compile
+      .count
+      .unsafeRunSync())
+    val nioTime = time(java.nio.file.Files.walk(target.toNioPath).count())
+    val epsilon = nioTime.toNanos * 1.5
+    println(s"fs2 took: ${fs2Time.toMillis} ms")
+    println(s"nio took: ${nioTime.toMillis} ms")
+    assert((fs2Time - nioTime).toNanos.abs < epsilon, s"fs2 time: $fs2Time, nio time: $nioTime, diff: ${fs2Time - nioTime}") 
+  }
+}


io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
```diff
@@ -389,6 +391,54 @@ private[file] trait FilesCompanionPlatform {
...
+            JFiles.walkFileTree(
+              start.toNioPath,
+              if (followLinks) Set(FileVisitOption.FOLLOW_LINKS).asJava else Set.empty.asJava,
+              maxDepth,
+              new SimpleFileVisitor[JPath] {
+                private def enqueue(path: JPath): FileVisitResult = {
+                  bldr += Path.fromNioPath(path)
+                  size += 1
+                  if (size >= limit) {
+                    val result = dispatcher.unsafeRunSync(channel.send(Chunk.from(bldr.result())))
```
djspiewak, line 411: I really wish there were a way to suspend the visitation and continue it later. That would allow us to avoid the `unsafeRunSync` here and use `unsafeRunAndForget` instead, likely bouncing out of the `interruptible` once every *n* enqueues and passing through a `Stream#append` in order to preserve backpressure.

Is `walkFileTree` meaningfully faster than just doing the traversal by hand?

io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
```diff
@@ -389,6 +391,54 @@ private[file] trait FilesCompanionPlatform {
         .resource(Resource.fromAutoCloseable(javaCollection))
         .flatMap(ds => Stream.fromBlockingIterator[F](collectionIterator(ds), pathStreamChunkSize))
 
+    override def walk(start: Path, maxDepth: Int, followLinks: Boolean): Stream[F, Path] =
+      Stream.resource(Dispatcher.sequential[F]).flatMap { dispatcher =>
+        Stream.eval(Channel.bounded[F, Chunk[Path]](10)).flatMap { channel =>
```
djspiewak, line 396: Btw @armanbilge one thing that occurs to me is that our fancy new unsafe queue thing isn't going to help very much if someone's using `Channel`.

djspiewak (commented):
I'm assuming the reason the overhead begins to converge in larger traversals is because there's a single big `eval`.

io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
mpilquist: Even eagerly collecting everything is only 5% faster than the channel based solution (using the 4096 limit):

```scala
def walkEager(start: Path, maxDepth: Int, followLinks: Boolean): Stream[F, Path] = {
    val doWalk = Sync[F].interruptibleMany {
      val bldr = Vector.newBuilder[Path]
      JFiles.walkFileTree(
        start.toNioPath,
        if (followLinks) Set(FileVisitOption.FOLLOW_LINKS).asJava else Set.empty.asJava,
        maxDepth,
        new SimpleFileVisitor[JPath] {
          private def enqueue(path: JPath): FileVisitResult = {
            bldr += Path.fromNioPath(path)
            FileVisitResult.CONTINUE
          }

          override def visitFile(file: JPath, attrs: JBasicFileAttributes): FileVisitResult =
            enqueue(file)

          override def visitFileFailed(file: JPath, t: IOException): FileVisitResult =
            FileVisitResult.CONTINUE

          override def preVisitDirectory(dir: JPath, attrs: JBasicFileAttributes): FileVisitResult =
            enqueue(dir)

          override def postVisitDirectory(dir: JPath, t: IOException): FileVisitResult =
            FileVisitResult.CONTINUE
        }
      )
      Chunk.from(bldr.result())
    }
    Stream.eval(doWalk).flatMap(Stream.chunk)
  }
```

djspiewak: Wow, that's wild honestly. Have to ponder that. It's nice that we can just be lazy about our thread blocking though, since it simplifies this stuff.

io/jvm/src/main/scala/fs2/io/file/AsyncFilesPlatform.scala
```diff
@@ -50,7 +50,7 @@ private[file] trait AsyncFilesPlatform[F[_]] { self: Files.UnsealedFiles[F] =>
       chunkSize: Int
   ): Stream[F, Path] =
     Stream.resource(Dispatcher.sequential[F]).flatMap { dispatcher =>
-      Stream.eval(Channel.bounded[F, Chunk[Path]](10)).flatMap { channel =>
+      Stream.eval(Channel.bounded[F, Chunk[Path]](2)).flatMap { channel =>
```
djspiewak, line 53: I mean, why not go the next step and do `synchronous` instead?

io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
mpilquist: @djspiewak BTW, there's a bunch of performance hackery in the JDK's file walking that's not (directly) available to us if we implement our own walk. For example, [`Path` can cache file attributes](https://github.com/JetBrains/jdk8u_jdk/blob/94318f9185757cc33d2b8d527d36be26ac6b7582/src/share/classes/java/nio/file/FileTreeWalker.java#L205-L213) avoiding some filesystem calls.

mpilquist: Just to close this out, I tried this prototype:

```scala

  override def walk(
      start: Path,
      maxDepth: Int,
      followLinks: Boolean,
      chunkSize: Int
  ): Stream[F, Path] =
    walkJustInTime(start, maxDepth, followLinks, chunkSize)
    // if (chunkSize == Int.MaxValue) walkEager(start, maxDepth, followLinks)
    // else walkLazy(start, maxDepth, followLinks, chunkSize)

  private def walkJustInTime(
      start: Path,
      maxDepth: Int,
      followLinks: Boolean,
      chunkSize: Int
  ): Stream[F, Path] = {

    def loop(acc: Vector[Path], toWalk: Vector[Path]): Stream[F, Path] = {
      if (toWalk.isEmpty) {
        Stream.chunk(Chunk.from(acc))
      } else {
        val path = toWalk.head

        val (toEmit, newAcc) =
          if (acc.size + 1 >= chunkSize)
            (Chunk.from(acc :+ path), Vector.empty)
          else (Chunk.empty, acc :+ path)

        val list = Sync[F].interruptibleMany {
          val npath = path.toNioPath
          if (JFiles.isDirectory(npath)) {
            val listing = JFiles.list(npath)
            try listing.iterator.asScala.map(Path.fromNioPath).toVector
            finally listing.close()
          }
          else Vector.empty
        }

        Stream.chunk(toEmit) ++ Stream.eval(list).flatMap(descendants => loop(newAcc, toWalk.drop(1) ++ descendants))
      }
    }

    loop(Vector.empty, Vector(start))
  }
```

Using `MaxDepth = 7`, I got these results:

```
fs2 took: 16070 ms
fs2 eager took: 13935 ms
nio took: 6356 ms
```

Whereas the implementation in this PR results in:

```
fs2 took: 8000 ms
fs2 eager took: 5975 ms
nio took: 6858 ms
```

mpilquist: Here's a better prototype that does file attribute reading at the time of directory listing.

```scala
  private def walkJustInTime(
      start: Path,
      maxDepth: Int,
      followLinks: Boolean,
      chunkSize: Int
  ): Stream[F, Path] = {

    def loop(acc: Vector[Path], toWalk: Vector[(Path, JBasicFileAttributes)]): Stream[F, Path] = {
      if (toWalk.isEmpty) {
        Stream.chunk(Chunk.from(acc))
      } else {
        val (path, attr) = toWalk.head

        val (toEmit, newAcc) =
          if (acc.size + 1 >= chunkSize)
            (Chunk.from(acc :+ path), Vector.empty)
          else (Chunk.empty, acc :+ path)

        if (attr.isDirectory) {
          val list = Sync[F].interruptibleMany {
            val npath = path.toNioPath
            if (JFiles.isDirectory(npath)) {
              val listing = JFiles.list(npath)
              try listing.iterator.asScala.map(p => 
                (Path.fromNioPath(p), JFiles.readAttributes(p, classOf[JBasicFileAttributes]))).toVector
              finally listing.close()
            }
            else Vector.empty
          }

          Stream.chunk(toEmit) ++ Stream.eval(list).flatMap(descendants => loop(newAcc, toWalk.drop(1) ++ descendants))
        } else Stream.chunk(toEmit) ++ loop(newAcc, toWalk.drop(1))
      }
    }

    Stream.eval(Sync[F].interruptibleMany {
      start -> JFiles.readAttributes(start.toNioPath, classOf[JBasicFileAttributes])
    }).flatMap { s => loop(Vector.empty, Vector(s)) }
  }
```

Performs better but still doesn't beat the `walkFileTree` solution:

```
fs2 took: 10847 ms
fs2 eager took: 8734 ms
nio took: 6633 ms
```

mpilquist: Alright, maybe we should switch to a version based on this:

```scala
  private def walkJustInTime(
      start: Path,
      maxDepth: Int,
      followLinks: Boolean,
      chunkSize: Int
  ): Stream[F, Path] = {

    def loop(toWalk0: Queue[(Path, JBasicFileAttributes)]): Stream[F, Path] = {
      val partialWalk = Sync[F].interruptibleMany {
        var acc = Vector.empty[Path]
        var toWalk = toWalk0

        while (acc.size < chunkSize && toWalk.nonEmpty) {
          val (path, attr) = toWalk.head
          toWalk = toWalk.drop(1)
          acc = acc :+ path
          if (attr.isDirectory) {
            val listing = JFiles.list(path.toNioPath)
            try {
              val descendants = listing.iterator.asScala.map(p => 
                (Path.fromNioPath(p), JFiles.readAttributes(p, classOf[JBasicFileAttributes]))).toVector
              toWalk = toWalk ++ descendants
            }
            finally listing.close()              
          }
            
        }

        Stream.chunk(Chunk.from(acc)) ++ (if (toWalk.isEmpty) Stream.empty else loop(toWalk))
      }
      Stream.eval(partialWalk).flatten
   }

    Stream.eval(Sync[F].interruptibleMany {
      start -> JFiles.readAttributes(start.toNioPath, classOf[JBasicFileAttributes])
    }).flatMap(s => loop(Queue(s)))
  }
```

```
fs2 took: 9312 ms
fs2 eager took: 8538 ms
nio took: 7769 ms
```

djspiewak: So basically what we're trying to figure out is whether it's worth eating 9% overhead to avoid blocking a thread which is already getting blocked by filesystem I/O? My *guess* is that it's not worth it but I shall ponder a bit.

mpilquist: Pushed a new version:

```
fs2 took: 8131 ms
fs2 eager took: 5950 ms
nio took: 7346 ms
```

I'd like to add some tests for symbolic link following & max depth limits (we don't have any now). Then this PR should be good.

mpilquist:
It appears `Files.walkFileTree` on Scala Native doesn't throw `FileSystemLoopException`. We could just skip the `walkEager` optimization on Scala Native for now I guess but then we're back to more platform specific traits.

mpilquist:
Okay this is ready for final review. Here's how we netted out performance wise:

```
fs2 took: 7574 ms
fs2 eager took: 5809 ms
nio took: 6956 ms
```

io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
```diff
@@ -389,6 +391,140 @@ private[file] trait FilesCompanionPlatform {
         .resource(Resource.fromAutoCloseable(javaCollection))
         .flatMap(ds => Stream.fromBlockingIterator[F](collectionIterator(ds), pathStreamChunkSize))
 
+    protected def walkEager(start: Path, options: WalkOptions): Stream[F, Path] = {
+      val doWalk = Sync[F].interruptibleMany {
```
djspiewak, line 395: Does it really need the `Many`?

io/jvm-native/src/main/scala/fs2/io/file/FilesPlatform.scala
```diff
@@ -389,6 +391,140 @@ private[file] trait FilesCompanionPlatform {
...
+        options: WalkOptions
+    ): Stream[F, Path] = {
+      import scala.collection.immutable.Queue
+
+      def loop(toWalk0: Queue[WalkEntry]): Stream[F, Path] = {
+        val partialWalk = Sync[F].interruptibleMany {
+          var acc = Vector.empty[Path]
+          var toWalk = toWalk0
+
+          while (acc.size < options.chunkSize && toWalk.nonEmpty) {
```
djspiewak, line 450: May be worth checking `Thread.interrupted()`

Issue #3384, opened 2024-02-06
reactive-streams interop: handled errors treated as unhandled and printed to the console 


fredfp:
When using `unsafeRunAndForget`, unhandled errors are printed to the console.

When using reactive-streams interop, stream errors end up printed to the console even if they are handled.

mpilquist:
Fixed by #3385

PR #3385, opened 2024-02-06
Do not let handled errors get printed to the console

fredfp:
fix for https://github.com/typelevel/fs2/issues/3384

### DIFF
--- a/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscription.scala
+++ b/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscription.scala
@@ -87,6 +87,7 @@ private[reactivestreams] final class StreamSubscription[F[_], A] private (
         case Outcome.Canceled()  => cancelMe
       }
       .void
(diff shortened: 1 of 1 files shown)


reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscription.scala
```diff
@@ -87,6 +87,7 @@ private[reactivestreams] final class StreamSubscription[F[_], A] private (
         case Outcome.Canceled()  => cancelMe
       }
       .void
+      .handleError(_ => ()) // errors handled above and passed to subscriber
```
armanbilge, line 90:
```suggestion
      .voidError // errors handled above and passed to subscriber
```

fredfp: Oh, I initially tried replacing `.void` by `.voidError`, thinking that `.voidError` was an extra layer on top of `.void`. You made me realise it's not the case, thank you.

armanbilge (approved):

PR #3387, opened 2024-02-10
Optmize `interop.flow.StreamSubscriber.onNext`

BalmungSan:
Since the `reactive-streams` spec mentions that all calls must be done _"serially"_ these changes optimize the tight loop of multiple consequential `onNext` calls.

-----

### Before the optimization

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| FlowInteropBenchmark.fastPublisher | 1000 |   1024 | thrpt | 20 | 25.938 | ± 0.306 | ops/s |
| FlowInteropBenchmark.fastPublisher | 1000 |   5120 | thrpt | 20 |  6.519 | ± 0.015 | ops/s |
| FlowInteropBenchmark.fastPublisher | 1000 |  10240 | thrpt | 20 |  3.201 | ± 0.168 | ops/s |
| FlowInteropBenchmark.fastPublisher | 1000 |  51200 | thrpt | 20 |  0.680 | ± 0.013 | ops/s |
| FlowInteropBenchmark.fastPublisher | 1000 | 512000 | thrpt | 20 |  0.062 | ± 0.001 | ops/s |

### After the optimization

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| FlowInteropBenchmark.fastPublisher | 1000 |   1024 | thrpt | 20 | 63.846 | ± 0.094 | ops/s |
| FlowInteropBenchmark.fastPublisher | 1000 |   5120 | thrpt | 20 | 22.314 | ± 0.107 | ops/s |
| FlowInteropBenchmark.fastPublisher | 1000 |  10240 | thrpt | 20 | 12.662 | ± 0.028 | ops/s |
| FlowInteropBenchmark.fastPublisher | 1000 |  51200 | thrpt | 20 |  2.736 | ± 0.012 | ops/s |
| FlowInteropBenchmark.fastPublisher | 1000 | 512000 | thrpt | 20 |  0.275 | ± 0.002 | ops/s |

-----

## Analysis

For only `2` chunks of `512` elements the improvements were almost `2.5` times better.
For `20` chunks, the improvements were almost `4` times better.
But the improvements for `1000` chunks were just `4.4` times better.

It seems then that in a fully sequential and CPU-bound `Publisher`, the new `Subscriber` is roughly 4 times more efficient. As long as the chunk size is adequate, and there are enough chunks to flatten the overhead of the general machinery.
Overall, I would say this optimization looks great, even if realistic use cases won't get such a big increase, it is likely the effects will be noticeable.

-----

> Lies, Damn Lies, and Benchmarks

### DIFF
--- a/benchmark/src/main/scala/fs2/benchmark/FlowInteropBenchmark.scala
+++ b/benchmark/src/main/scala/fs2/benchmark/FlowInteropBenchmark.scala
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2013 Functional Streams for Scala
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2
+package benchmark
+
+import cats.effect.IO
+import cats.effect.unsafe.implicits.global
+
+import org.openjdk.jmh.annotations.{
+  Benchmark,
+  BenchmarkMode,
+  Mode,
+  OutputTimeUnit,
+  Param,
+  Scope,
+  State
+}
+
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.Flow.{Publisher, Subscriber, Subscription}
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class FlowInteropBenchmark {
+  @Param(Array("1024", "5120", "10240", "51200", "512000"))
+  var totalElements: Long = _
+
+  @Param(Array("1000"))
+  var iterations: Int = _
+
+  @Benchmark
+  def fastPublisher(): Unit = {
+    def publisher =
+      new Publisher[Unit] {
+        override final def subscribe(subscriber: Subscriber[? >: Unit]): Unit =
+          subscriber.onSubscribe(
+            new Subscription {
+              var i: Long = 0
+              @volatile var canceled: Boolean = false
+
+              // Sequential fast Publisher.
+              override final def request(n: Long): Unit = {
+                val elementsToProduce = math.min(i + n, totalElements)
+
+                while (i < elementsToProduce) {
+                  subscriber.onNext(())
+                  i += 1
+                }
+
+                if (i == totalElements || canceled) {
+                  subscriber.onComplete()
+                }
+              }
+
+              override final def cancel(): Unit =
+                canceled = true
+            }
+          )
+      }
+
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/fs2/interop/flow/StreamSubscription.scala
```diff
@@ -109,6 +109,7 @@ private[flow] final class StreamSubscription[F[_], A] private (
             // if we were externally canceled, this is handled below
             F.unit
         }
+        .mask
```
BalmungSan, line 112: Related to #3385

benchmark/src/main/scala/fs2/benchmark/FlowInteropBenchmark.scala
```diff
@@ -0,0 +1,95 @@
...
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class FlowInteropBenchmark {
+  @Param(Array("1024", "5120", "10240"))
+  var totalElements: Int = _
+
+  @Param(Array("1000"))
+  var iterations: Int = _
```
BalmungSan, line 51: Are these parameters okay?

benchmark/src/main/scala/fs2/benchmark/FlowInteropBenchmark.scala
```diff
@@ -0,0 +1,95 @@
...
+                  }
+
+                  if (i == totalElements || canceled) {
+                    subscriber.onComplete()
+                  }
+                }(global.compute)
+
+                // Discarding the Future so it runs in the background.
+                ()
+              }
```
BalmungSan, line 79: @armanbilge is this close to the idea that you had in mind originally?

benchmark/src/main/scala/fs2/benchmark/FlowInteropBenchmark.scala
```diff
@@ -0,0 +1,95 @@
...
+                canceled = true
+            }
+          )
+      }
+
+    val stream =
+      interop.flow.fromPublisher[IO](publisher, chunkSize = 512)
+
+    val program =
+      stream.compile.toVector
```
BalmungSan, line 91: I was not sure what was the best way to collect the results.
My other two options were doing something like `compile.fold(0)(_ + _)` or a `.foreach { _ => IO.unit }.compile.drain`

BalmungSan: @armanbilge I guess I was worried that something like that could be optimized and thus we would not see if the production of elements was actually faster or not.
Same about publishing `Ints` over `Units`, I kind of wanted to ensure we actually published and consumed something.

But, I guess that given `Stream` could produce any side-effects, there can't be any optimization that avoid actually producing the `Chunks` even if their values are going to be discarded.
Thus, yeah probably pushing `Units` and doing `compile.drain` is the best idea.

benchmark/src/main/scala/fs2/benchmark/FlowInteropBenchmark.scala
```diff
@@ -0,0 +1,95 @@
...
+                  var j = 0
+                  while ((j < n) && (i < totalElements) && !canceled) {
+                    subscriber.onNext(())
+                    i += 1
+                    j += 1
+                  }
+
+                  if (i == totalElements || canceled) {
+                    subscriber.onComplete()
+                  }
```
armanbilge, line 74: We are doing multiple volatile reads/writes of `i` in this loop. I think we can just copy it into a local variable and then copy it back out.

```suggestion
                  var i = this.i
                  while ((j < n) && ((i < totalElements) && !canceled) {
                    subscriber.onNext(())
                    i += 1
                    j += 1
                  }

                  if (i == totalElements || canceled) {
                    subscriber.onComplete()
                  }
                  
                  this.i = i
```

mzuehlke: Maybe make this lore obvious and call the local variable "locali" and not rely on shadowing

BalmungSan: After more discussion in **Discord** we decided to go with a sequential publisher for simplicity.

benchmark/src/main/scala/fs2/benchmark/FlowInteropBenchmark.scala
```diff
@@ -0,0 +1,90 @@
...
+          subscriber.onSubscribe(
+            new Subscription {
+              var i: Int = 0
+              @volatile var canceled: Boolean = false
+
+              // Sequential fast Publisher.
+              override final def request(n: Long): Unit = {
+                var j = 0
+
+                while ((j < n) && (i < totalElements)) {
```
armanbilge, line 65: performance nit 😝 

instead of two comparisons in the loop condition we can just calculate `min(i + n, totalElements)` up-front.

armanbilge (approved):

PR #3394, opened 2024-02-20
Remove walkEager optimization

mpilquist:
Drops the specialized `walkEager` optimization as `walkJustInTime` is just as efficient in the case where `chunkSize = Int.MaxValue`. This lets us get rid of the specialized platform traits (`AsyncFilesPlatform`).

Merge after #3390


SystemFw (approved):
nice!

Issue #1958, opened 2020-07-16
[Documentation] Add a reference page / cookbook of common concurrency operations


Daenyth:
It should cover when you'd use and how to use:
- parEvalMap/Unordered
- parJoin
- concurrently
- merge*

There's the page on Queue/Topic/etc but there's not a great single page on the operations directly on `Stream`. The Guide section on concurrency is very short and doesn't cover some of the most common operations (parEvalMap especially)

Hombre-x:
I can help a little bit with that! 👀

onsah:
Hi, is this finished or are there still things to do?

Irfan762:
@Daenyth I’d like to work on this issue. Could you please assign it to me?

PR #3437, opened 2024-05-17
📝 Add more examples / uses to the Concurrency section in the documentation guide of Fs2

Hombre-x:
Helps with  #1958 

What's included: 

- Examples for both `parEvalMap` and `parEvalMapUnordered`
- Reasoning about the three `merge` variants; it's about when they will halt or not depending on the provided streams
- Added a `concurrently` example using a simple producer-consumer environment (Had to use Scala 2 syntax with implicits, I didn't know how to make mdoc work with Scala 3 at the same time 🤔).
- Changed the previous example to make it suitable for `parJoin`

### DIFF
--- a/site/guide.md
+++ b/site/guide.md
@@ -509,10 +509,10 @@ import cats.effect.unsafe.implicits.global
 import fs2.io.file.{Path, Files}
 
 val paths = List(
-  Path("file1.txt"),
-  Path("file2.txt"),
-  Path("file3.txt"),
-).map(Path("path/to/files/") / _)
+  Path("sample_file_part1.txt"),
+  Path("sample_file_part2.txt"),
+  Path("sample_file_part3.txt"),
+).map(Path("testdata") / _)
 
 def loadFile(path: Path): IO[String] = 
   Files[IO].readUtf8(path).compile.string
@@ -520,7 +520,7 @@ def loadFile(path: Path): IO[String] =
 Stream.emits(paths)
   .parEvalMap[IO, String](3)(loadFile(_))   // Loads files into memory
   .reduce(_ + _)                            // Combines the content of the files into single one in order
-  .through(Files[IO].writeUtf8(Path("path/to/output.txt")))
+  .through(Files[IO].writeUtf8(Path("testdata/sample_file_output.txt")))
   .compile
   .drain
   .unsafeRunSync()
--- a/site/guide.md
+++ b/site/guide.md
@@ -439,6 +439,66 @@ Stream(1,2,3).merge(Stream.eval(IO { Thread.sleep(200); 4 })).compile.toVector.u
 
 The `merge` function supports concurrency. FS2 has a number of other useful concurrency functions like `concurrently` (runs another stream concurrently and discards its output), `interruptWhen` (halts if the left branch produces `true`), `either` (like `merge` but returns an `Either`), `mergeHaltBoth` (halts if either branch halts), and others.
 
+The `parEvalMap` function allows you to evaluate effects in parallel and emit the results in order on up to `maxConcurrent` fibers at the same time, similar to the `parTraverseN` method that you would normally use in standard library collections:
+
+``` scala mdoc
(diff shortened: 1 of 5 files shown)


site/guide.md
```diff
@@ -437,17 +437,146 @@ import cats.effect.unsafe.implicits.global
...
+infinite.mergeHaltL(finite)     // Doesn't terminate
+
+finite.mergeHaltR(infinite)     // Doesn't terminate
+infinite.mergeHaltR(finite)     // Terminates
+
+finite.mergeHaltBoth(infinite)  // Terminates
+infinite.mergeHaltBoth(finite)  // Also terminates
+```
+
+Since it is quite common to terminate one stream as soon as the other is finished (as in a producer-consumer environment), there are optimisations over the `merge` variants. `concurrently` is one of them, as it will terminate the resulting stream when the stream on the left halts. The stream on the right will also terminate at that point, discarding its values in the meantime (similar to `finite.mergeHaltL(infinite.drain)`):
```
mpilquist, line 459:
```suggestion
Since it is quite common to terminate one stream as soon as the other is finished (as in a producer-consumer environment), there are optimizations over the `merge` variants. `concurrently` is one of them, as it will terminate the resulting stream when the stream on the left halts. The stream on the right will also terminate at that point, discarding its values in the meantime (similar to `finite.mergeHaltL(infinite.drain)`):
```

site/guide.md
```diff
@@ -437,17 +437,146 @@ import cats.effect.unsafe.implicits.global
...
+
+```scala mdoc
+import cats.effect.IO
+import cats.effect.std.{Queue, Random}
+import cats.effect.unsafe.implicits.global
```
mpilquist, line 464: Don't need all these imports again here -- can definitely remove `cats.effect.IO` and `cats.effect.unsafe.implicits.global` for instance.

site/guide.md
```diff
@@ -437,17 +437,146 @@ import cats.effect.unsafe.implicits.global
...
+concurrentlyDemo.compile.drain.unsafeRunSync()
+```
+
+In the example above, the `consumer` stream will terminate if an element produced by the `producer` takes more than 750 milliseconds.
+
+
+The `parEvalMap` function allows you to evaluate effects in parallel and emit the results in order on up to `maxConcurrent` fibers at the same time, similar to the `parTraverseN` method that you would normally use in standard library collections:
+
+``` scala mdoc
+import cats.effect.IO
```
mpilquist, line 495: Drop imports here and in each example below

site/guide.md
```diff
@@ -437,17 +437,146 @@ import cats.effect.unsafe.implicits.global
...
+import cats.effect.unsafe.implicits.global
+import fs2.io.file.{Path, Files}
+
+val paths = List(
+  Path("sample_file_part1.txt"),
+  Path("sample_file_part2.txt"),
+  Path("sample_file_part3.txt"),
+).map(Path("testdata") / _)
+
+def loadFile(path: Path): IO[String] =
```
mpilquist, line 515: I don't love this example for a couple of reasons:
- concurrency not really adding any benefit given all actions are against the file system
- unnecessarily parses the file contents to a string
- unnecessarily loads full file contents in memory
- if those downsides aren't important, then the same thing could be done entirely with `IO` and without `Stream`

Here's a full streaming version of this file concat routine that maintains a streaming nature and avoids parsing:

```scala
def concat(inPaths: Stream[IO, Path], outPath: Path): Stream[IO, Nothing] =
  inPaths.flatMap(p => Files[IO].readAll(p)).through(Files[IO].writeAll(outPath))
```

Hombre-x: You are right, I did It just as an excuse to maintain the order of the original stream 🤔 let me thing of a better example!

Hombre-x:
I sometimes forget that GitHub don't usually notifies the commits in PRs, so changes are done @mpilquist!

mpilquist:
Thanks!

PR #3447, opened 2024-06-14
Fix README link

bwignall:
The [README currently](https://github.com/typelevel/fs2/blob/0377f21532f3486c54da62d38430d9657d052960/README.md) contains "There are Scaladoc API documentations for [the core library][core-api], which defines and..."; note that a link has not been rendered to "[core-api]".

Looking through [git blame](https://github.com/typelevel/fs2/blame/main/README.md), "core-api" previously existed ([relevant commit](https://github.com/typelevel/fs2/commit/fa9cd8cb963d7a106ae1b69530d6b2ad7538e94f)) and pointed to a link on sonatype.org. The reference "[core-api]" changed to "[api]" in [this commit](https://github.com/typelevel/fs2/commit/15c790e498175e65c4d43dd8396d8fcee85bd847) but one reference to "core-api" was not updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Quick links:
 
 ### <a id="docs"></a>Documentation and getting help ###
 
-* There are Scaladoc API documentations for [the core library][core-api], which defines and implements the core types for streams and pulls, as well as the type aliases for pipes and sinks. [The `io` library][io-api] provides FS2 bindings for NIO-based file I/O and TCP/UDP networking.
+* There are Scaladoc API documentations for [the core library][api], which defines and implements the core types for streams and pulls, as well as the type aliases for pipes and sinks. [The `io` library][io-api] provides FS2 bindings for NIO-based file I/O and TCP/UDP networking.
 * [The official guide](https://fs2.io/#/guide) is a good starting point for learning more about the library.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,7 +33,7 @@ Quick links:
 
 ### <a id="docs"></a>Documentation and getting help ###
 
-* There are Scaladoc API documentations for [the core library][core-api], which defines and implements the core types for streams and pulls, as well as the type aliases for pipes and sinks. [The `io` library][io-api] provides FS2 bindings for NIO-based file I/O and TCP/UDP networking.
+* There are Scaladoc API documentations for [the core library][api], which defines and implements the core types for streams and pulls, as well as the type aliases for pipes and sinks. [The `io` library][io-api] provides FS2 bindings for NIO-based file I/O and TCP/UDP networking.
```
armanbilge, line 36: Actually the API docs are no longer split into core/io, it's all under the `[api]` link now. I suspect the second link is also broken.

bwignall: I have removed reference to that split. I made the phrasing of this bullet point similar to the one after it. Please let me know if you have other preferences/suggestions on wording.

bwignall: As there remained no references to `[io-api]` or to `[rx-api]`, I removed the Sonatype links to both.

armanbilge (approved):
thank you so much!

PR #3449, opened 2024-06-14
Add `interop.flow.pipeToProcessor` & `interop.flow.processorToPipe`

BalmungSan:
Convenient helpers.

Roadmap:
- [x] safe version.
- [x] **Scaladoc**.
- [x] `unsafe` version.
- [x] Unit test.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -5564,6 +5564,26 @@ object Stream extends StreamLowPriority {
       interop.flow.pipeToProcessor(pipe = self, chunkSize)
   }
 
+  /** Provides operations on IO pipes for syntactic convenience. */
+  implicit final class IOPipeOps[I, O](private val self: Pipe[IO, I, O]) extends AnyVal {
+
+    /** Creates a [[Processor]] from this [[Pipe]].
+      *
+      * You are required to manually subscribe this [[Processor]] to an upstream [[Publisher]], and have at least one downstream [[Subscriber]] subscribe to the [[Consumer]].
+      *
+      * @param chunkSize setup the number of elements asked each time from the upstream [[Publisher]].
+      *                  A high number may be useful if the publisher is triggering from IO,
+      *                  like requesting elements from a database.
+      *                  A high number will also lead to more elements in memory.
+      */
+    def unsafeToProcessor(
+        chunkSize: Int
+    )(implicit
+        runtime: IORuntime
+    ): Processor[I, O] =
+      interop.flow.unsafePipeToProcessor(pipe = self, chunkSize)
+  }
+
   /** Provides operations on pure pipes for syntactic convenience. */
   implicit final class PurePipeOps[I, O](private val self: Pipe[Pure, I, O]) extends AnyVal {
 
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2853,7 +2853,7 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * @param subscriber the [[Subscriber]] that will receive the elements of the stream.
     */
   def subscribe[F2[x] >: F[x]: Async, O2 >: O](subscriber: Subscriber[O2]): Stream[F2, Nothing] =
-    interop.flow.subscribeAsStream[F2, O2](this, subscriber)
+    interop.flow.StreamSubscription.subscribe[F2, O2](this, subscriber)
 
   /** Emits all elements of the input except the first one.
     *
@@ -3001,7 +3001,7 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
 
   /** @see [[toPublisher]] */
   def toPublisherResource[F2[x] >: F[x]: Async, O2 >: O]: Resource[F2, Publisher[O2]] =
-    interop.flow.toPublisher(this)
+    interop.flow.StreamPublisher(this)
 
   /** Translates effect type from `F` to `G` using the supplied `FunctionK`.
     */
@@ -3911,7 +3911,7 @@ object Stream extends StreamLowPriority {
     *                  either the `Chunk` is filled or the publisher finishes.
     */
   def fromPublisher[F[_]]: interop.flow.syntax.FromPublisherPartiallyApplied[F] =
-    interop.flow.fromPublisher
+    new interop.flow.syntax.FromPublisherPartiallyApplied(dummy = true)
 
   /** Like `emits`, but works for any G that has a `Foldable` instance.
     */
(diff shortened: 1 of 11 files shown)


core/shared/src/main/scala/fs2/interop/flow/StreamProcessor.scala
```diff
@@ -0,0 +1,44 @@
...
+private[flow] object StreamProcessor {
+  def fromPipe[F[_], I, O](
+      pipe: Pipe[F, I, O],
+      chunkSize: Int
+  )(implicit
+      F: Async[F]
+  ): Resource[F, StreamProcessor[F, I, O]] =
+    for {
+      streamSubscriber <- Resource.eval(StreamSubscriber[F, I](chunkSize))
+      inputStream = streamSubscriber.stream(subscribe = F.unit)
```
BalmungSan, line 37: We don't have access to the upstream `Publisher` yet, so we can't do the `subscribe` here. So we do nothing and wait for it.

core/shared/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -199,6 +198,51 @@ package object flow {
...
+    *
+    * @see [[unsafePipeToProcessor]] for an unsafe version that returns a plain [[Processor]].
+    *
+    * @param pipe The [[Pipe]] which represents the [[Processor]] logic.
+    * @param chunkSize setup the number of elements asked each time from the upstream [[Publisher]].
+    *                  A high number may be useful if the publisher is triggering from IO,
+    *                  like requesting elements from a database.
+    *                  A high number will also lead to more elements in memory.
+    */
+  def pipeToProcessor[F[_], I, O](
```
armanbilge, line 218: We can add these as syntax on `Pipe` too :)

https://github.com/typelevel/fs2/blob/0377f21532f3486c54da62d38430d9657d052960/core/shared/src/main/scala/fs2/Stream.scala#L5494

core/shared/src/main/scala/fs2/interop/flow/StreamProcessor.scala
armanbilge: > So we do nothing and wait for it.

I'm not entirely sure I understand the waiting part. Can/should we use a `Deferred` here?

BalmungSan: No sorry, my point is that `subscribe` should be something like `IO(publisher.subscribe(streamSubscriber))` as you can see in the implementation of `syntax.fromPublisher`.
Here, however, we don't have a `Publisher` yet, so we don't control when we are published.

Basically, this is very similar to the `fromPublisher` overload that accepts a `Subscriber => F[Unit]`, just that rather than receiving the lambda, we are just returning a raw `Processor / Subscriber`.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -5521,6 +5521,27 @@ object Stream extends StreamLowPriority {
...
+      * but waiting for all active ones to finish consuming.
+      * Canceling the [[Resource.use]] means gracefully shutting down all active subscriptions.
+      * Thus, no more elements will be published.
+      *
+      * @param chunkSize setup the number of elements asked each time from the upstream [[Publisher]].
+      *                  A high number may be useful if the publisher is triggering from IO,
+      *                  like requesting elements from a database.
+      *                  A high number will also lead to more elements in memory.
+      */
+    def toProcessor(
```
armanbilge, line 5539: Can we also add `unsafeToProcessor` on a `PipeIOOps` syntax class

core/shared/src/main/scala/fs2/interop/flow/package.scala
armanbilge, line 218: Actually I forgot, we (me 😅 ) deprecated this entire package. So actually we don't need to add these here at all, or if we decide to they can be private so as not to increase the API surface? Idk, not a huge deal.

core/shared/src/main/scala/fs2/interop/flow/StreamProcessor.scala
armanbilge: Thanks, I understand now.

I think what confused me was that this is the first time we are effectively exposing `StreamSubscriber`. Our existing `{to,from}Publisher` APIs have kept it hidden.

So I do feel weird about exposing it now. But I guess nothing here is unreasonable 🤔

core/shared/src/main/scala/fs2/interop/flow/package.scala
BalmungSan: Wait, why? This `package object` is not / should not be deprecated!
This is the API we are exposing.

core/shared/src/test/scala/fs2/interop/flow/ProcessorPipeSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+
+final class ProcessorPipeSpec extends Fs2Suite {
+  test("should process upstream input and propagate results to downstream") {
+    forAllF(Arbitrary.arbitrary[Seq[Int]], Gen.posNum[Int]) { (ints, bufferSize) =>
+      // Since creating a Flow.Processor is very complex,
+      // we will reuse our Pipe => Processor logic.
+      val processor = unsafePipeToProcessor[Int, Int](
+        pipe = stream => stream.map(_ * 1),
+        chunkSize = bufferSize
+      )
```
BalmungSan, line 38: Not sure if there is a better way to test this.
Also, not sure if there is something else we may test.

core/shared/src/main/scala/fs2/interop/flow/package.scala
```diff
@@ -243,6 +243,24 @@ package object flow {
...
+    * Each invocation of the pipe will create and manage its own internal [[Publisher]] and [[Subscriber]],
+    * and use them to subscribe to and from the [[Processor]] respectively.
+    *
+    * @param [[processor]] the [[Processor]] that represents the [[Pipe]] logic.
+    * @param chunkSize setup the number of elements asked each time from the upstream [[Publisher]].
+    *                  A high number may be useful if the publisher is triggering from IO,
+    *                  like requesting elements from a database.
+    *                  A high number will also lead to more elements in memory.
+    */
+  def processorToPipe[F[_]]: syntax.FromProcessorPartiallyApplied[F] =
```
armanbilge, line 261: We can add this as a syntax on `Pipe` companion object, `fromProcessor`.

core/shared/src/main/scala/fs2/interop/flow/package.scala
armanbilge: > but what about the other side of the syntax? Also, what about methods that are not syntax?

It should be possible to implement everything as either implicit syntax (not requiring imports) or as methods on the relevant companion objects. I think syntax requiring imports is more of an anti-pattern.

armanbilge:
CI passed this time, but one of the tests seems flaky. Possible race condition?

```
==> X fs2.interop.flow.StreamProcessorSpec.should process upstream input and propagate results to downstream  30.006s java.util.concurrent.TimeoutException: test timed out after 30 seconds
```

https://github.com/BalmungSan/fs2/actions/runs/12014747489/job/33491269929#step:11:1615

BalmungSan:
@armanbilge I have been re-running the test for a while now, and it always succeeds in a couple of seconds.
Do you have any idea how I could try to force the race condition? Or maybe it was just something weird? It may as well be the case that the merge with `main` fixed it?
I also have checked the code and everything looks correct as far as I can tell.

armanbilge:
> It may as well be the case that the merge with `main` fixed it?

I see the exact same failure in your latest merge commit 😕 it must be something legit. I will try to reproduce ...

PR #3454, opened 2024-07-04
Replace fs2.hash with fs2.hashing.Hashing[F]

mpilquist:
The `fs2.hash` object provides the ability to compute the hash of a stream of bytes (i.e. `Stream[F, Byte]`). The hash computation is modeled as a `Pipe[F, Byte, Byte]`, where pulling on the output stream results in all the source bytes getting hashed and then a final chunk being emitted that contains the hash of the seen bytes.

This API is not expressive enough for common use cases though. Consider the case where a `Stream[F, Byte]` needs to be written to a file (or uploaded to an S3 bucket, or sent to a socket, etc) and the hash also needs to be stored. The current hashing API makes this awkward -- either requiring heavy duty machinery like `broadcastThrough` or requiring the stream to be processed twice.

This PR deprecates the `fs2.hash` object and replaces it with the new `fs2.hashing` package. The entry point to the new package is the `Hashing[F]` capability trait, allowing the creation of `Hash[F]` objects as well as providing various convenience methods.

The hashing package contains:
- `HashAlgorithm` enumeration -- e.g., `HashAlgorithm.SHA256` and `HashAlgorithm.Named("MD2")`
- `Hash[F]` mutable object -- allows incremental computation of hashes with a specific algorithm. `Hash` defines the following operations:
  - Low level hashing operations: `addChunk(c: Chunk[Byte]): F[Unit]` and `computeAndReset: F[Chunk[Byte]]`
  - `update: Pipe[F, Byte, Byte]` - updates the hash with the chunks pulled through the pipe
  - `observe(source: Stream[F, Byte], sink: Pipe[F, Byte, Nothing]): Stream[F, Byte]` - returns a stream that outputs the hash of the source bytes after they've been consumed by the supplied sink
  - `hash: Pipe[F, Byte, Byte]` - outputs the hash of the source bytes
  - `verify(expected: Chunk[Byte])(implicit F: RaiseThrowable[F]): Pipe[F, Byte, Byte]` - pipe that outputs the source bytes but raises a `HashVerificationException` when the source terminates if the hash of seen bytes doesn't match the expected hash
- `Hashing[F]` capability trait -- allows creation of `Hash[F]` objects

The `Hashing[F]` trait returns `Hash[F]` objects as resources (i.e. `Resource[F, Hash[F]]`) because (on some platforms) they have to be released after computations are complete.

With this new API, writing the contents of a source to a file and then subsequently writing a hash to a separate file, while processing the source just once, can be accomplished like so:

```scala
def writeWithHash[F[_]: Files: Hashing: MonadCancelThrow](path: Path): Pipe[F, Byte, Nothing] =
  source =>
     Stream.resource(Hashing[F].sha256).flatMap { h =>
       h.observe(source, Files[F].writeAll(path)).through(Files[F].writeAll(Path(s"$path.sha256")))
     }
```

The `Hashing` object also contains utility functions for hashing a pure stream and a chunk.

```scala
val h1 = Hashing.hashChunk(HashAlgorithm.SHA256, Chunk.array("The quick brown fox".getBytes))
val h2 = Hashing.hashPureStream(HashAlgorithm.SHA256, Stream.chunk(Chunk.array("The quick brown fox".getBytes)))
```


core/shared/src/main/scala/fs2/hashing/Hashing.scala
```diff
@@ -0,0 +1,52 @@
...
+import cats.effect.Sync
+
+/** Capability trait that provides hashing.
+  *
+  * The [[create]] method returns an action that instantiates a fresh `Hash` object.
+  * `Hash` is a mutable object that supports incremental computation of hashes. A `Hash`
+  * instance should be created for each hash you want to compute.
+  */
+trait Hashing[F[_]] {
+  def create(algorithm: String): F[Hash[F]]
```
armanbilge, line 34: Ideally this should return a `Resource` so that we can do cleanup or back it by a pool.

```suggestion
  def create(algorithm: String): Resource[F, Hash[F]]
```

mpilquist: Hm, [do we really need object pooling](https://stackoverflow.com/questions/13913075/to-pool-or-not-to-pool-java-crypto-service-providers)?

armanbilge: Maybe, maybe not. We still need a `Resource` anyway for Scala Native, and it opens the door to pooling if someone decides that's what they want to do.

mpilquist: Oh gotcha, hadn’t gotten as far as native integration yet. Makes sense that we’ll need to release something.

core/shared/src/main/scala/fs2/hashing/Hashing.scala
```diff
@@ -0,0 +1,52 @@
...
+  def sha512: F[Hash[F]] = create("SHA-512")
+
+  def hashWith(hash: F[Hash[F]]): Pipe[F, Byte, Byte] =
+    source => Stream.eval(hash).flatMap(h => h.hash(source))
+}
+
+object Hashing {
+  implicit def apply[F[_]](implicit F: Hashing[F]): F.type = F
+
+  implicit def forSync[F[_]: Sync]: Hashing[F] = new Hashing[F] {
```
armanbilge, line 48: We should follow the same practice of only offering implicits for `LiftIO` if the constraint is `Sync` or higher. https://github.com/typelevel/fs2/pull/3179, https://github.com/typelevel/fs2/pull/3192

core/shared/src/main/scala/fs2/hashing/Hashing.scala
```diff
@@ -0,0 +1,52 @@
...
+  def sha256: F[Hash[F]] = create("SHA-256")
+  def sha384: F[Hash[F]] = create("SHA-384")
+  def sha512: F[Hash[F]] = create("SHA-512")
+
+  def hashWith(hash: F[Hash[F]]): Pipe[F, Byte, Byte] =
+    source => Stream.eval(hash).flatMap(h => h.hash(source))
+}
+
+object Hashing {
+  implicit def apply[F[_]](implicit F: Hashing[F]): F.type = F
```
armanbilge, line 46:
```suggestion
  def apply[F[_]](implicit F: Hashing[F]): F.type = F
```

core/shared/src/main/scala/fs2/hashing/HashAlgorithm.scala
```diff
@@ -0,0 +1,34 @@
...
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2
+package hashing
+
+sealed trait HashAlgorithm
```
armanbilge, line 25:
```suggestion
sealed abstract class HashAlgorithm
```

core/shared/src/main/scala/fs2/hashing/HashAlgorithm.scala
```diff
@@ -0,0 +1,34 @@
...
+
+object HashAlgorithm {
+  case object MD5 extends HashAlgorithm
+  case object SHA1 extends HashAlgorithm
+  case object SHA256 extends HashAlgorithm
+  case object SHA384 extends HashAlgorithm
+  case object SHA512 extends HashAlgorithm
+  case class Named(name: String) extends HashAlgorithm
```
armanbilge, line 33:
```suggestion
  final case class Named(name: String) extends HashAlgorithm
```

core/js/src/main/scala/fs2/hashing/HashCompanionPlatform.scala
```diff
@@ -0,0 +1,83 @@
...
+      private var h = newHash()
+
+      def addChunk(bytes: Chunk[Byte]): F[Unit] =
+        Sync[F].delay(unsafeAddChunk(bytes.toArraySlice))
+
+      def computeAndReset: F[Chunk[Byte]] =
+        Sync[F].delay(unsafeComputeAndReset())
+
+      def unsafeAddChunk(slice: Chunk.ArraySlice[Byte]): Unit =
+        h.update(slice.toUint8Array)
```
armanbilge, line 50: `ArraySlice` is the optimal representation on JVM/Native but not on JS where `Uint8Array`s are primitive and not `Array`s. Forcing the API to go through `ArraySlice` can introduce up to two unnecessary copies.

mpilquist:
@armanbilge Thanks, addressed comments

PR #3465, opened 2024-08-19
Update Hashing to support HMACs, update names in fs2.hashing to be clearer, introduce Digest type

mpilquist:
This PR adds a few features:
- support for computing HMACs
- SHA3 and various other modern codecs

This PR makes a number of name changes including:
- replacing `addChunk` with `update`
- replacing `computeAndReset` with `hash`
- replacing `hash` with `drain`
- renaming `Hash[F]` to `Hasher[F]`
- renaming `Hashing.create` to `Hashing.hasher`

Additionally:
- `fs2.hashing.Hash` type was introduced, wrapping a `Chunk[Byte]` and providing a constant time equality check
- pipes were refactored to use `Hash` as the output type when appropriate


core/shared/src/main/scala/fs2/hashing/Hashing.scala
```diff
@@ -26,50 +26,90 @@ import cats.effect.{IO, LiftIO, Resource, Sync, SyncIO}
...
   */
 sealed trait Hashing[F[_]] {
 
   /** Creates a new hash using the specified hashing algorithm. */
-  def create(algorithm: HashAlgorithm): Resource[F, Hash[F]]
+  def hash(algorithm: HashAlgorithm): Resource[F, Hash[F]]
 
   /** Creates a new MD-5 hash. */
-  def md5: Resource[F, Hash[F]] = create(HashAlgorithm.MD5)
+  def md5: Resource[F, Hash[F]] = hash(HashAlgorithm.MD5)
```
mpilquist, line 53: Undecided if we should have all these aliases. Options:
- No aliases -- use `Hashing[F].hash(HashAlgorithm.SHA256)` instead
- Aliases for most popular hashes only -- SHA1, SHA256, SHA3_256?
- Aliases for all built-in hashes (that's current state of this PR)

We don't have aliases for HMACs (though we could) -- e.g., `def hmacSha256(key: Chunk[Byte]) = hmac(hash(HashAlgorithm.SHA256), key)`

Thoughts @armanbilge?

armanbilge: > * Aliases for all built-in hashes

It's not really possible to have this in general. By "built-in" we mean hashes that we have a `case` for. Because what's actually built-in on the runtime system can be totally variable depending on the build of the JDK/Node.js/OpenSSL. In truth, any of these methods could raise an error. Maybe no aliases would be more honest to that, although I can see how that's annoying from a usability perspective.

core/shared/src/main/scala/fs2/hashing/HashAlgorithm.scala
```diff
@@ -27,8 +27,30 @@ sealed abstract class HashAlgorithm
...
   case object SHA256 extends HashAlgorithm
   case object SHA384 extends HashAlgorithm
   case object SHA512 extends HashAlgorithm
+  case object SHA512_224 extends HashAlgorithm
+  case object SHA512_256 extends HashAlgorithm
+  case object SHA3_224 extends HashAlgorithm
+  case object SHA3_256 extends HashAlgorithm
+  case object SHA3_384 extends HashAlgorithm
+  case object SHA3_512 extends HashAlgorithm
   final case class Named(name: String) extends HashAlgorithm
```
armanbilge, line 40: If `HashAlgorithm` is `sealed` then we technically can't add new `case`s in the future. There are various annoying workarounds, we could unseal it but make the ctor private for example ... or just ignore it and break compatibility in this relatively minor way.

core/shared/src/main/scala/fs2/hashing/Hashing.scala
mpilquist: Added some docs to make the possibility of failure clear and removed all aliases.

core/shared/src/main/scala/fs2/hashing/HashAlgorithm.scala
mpilquist: Thanks, I made it unsealed with private constructor.

Issue #2866, opened 2022-04-02
Documentation: Add information about `Channel`


davesmith00000:
@armanbilge Has pointed me in the direction of `Channel` as the primitive type I may be after.

Channel certainly appears in the API reference, but it might be nice to briefly mention it in the main docs - possibly next to `Topic`?

armanbilge:
When this graduates to docs, here's a popular recipe!

```scala
  val prog =
    Channel
      .bounded[IO, String](10)
      .flatMap { ch =>
        val producer = ch.send("first")
        val consumer = ch.stream.evalMap(IO.println).compile.drain

        def gracefulClosure(f: FiberIO[Unit]) =
          ch.close >> f.join.void.timeoutTo(5.seconds, f.cancel)

        producer >>
        Resource.make(consumer.start)(gracefulClosure).use_
    }
```

PR #3475, opened 2024-09-14
Adding Channel to documentation

hkateu:
Hi, 
I added documentation for Channel based on [issue2866](https://github.com/typelevel/fs2/issues/2866). 

I also think that for new users to easily follow the documentation, the concurrency primitives, we can do the following: 

- Change the order  in the docs to: `Signal` , `Channel`, `Topic` then `Queue`. Going from a `Channel` to a `Topic` seems logical since its a  many-to-one vs many to many pub-sub relationship.
- Show that Topics are used for a many-to-many pub-sub scenario, since it's not explicitly mentioned in the docs.

Any feedback is welcome.


testdata/tom_sawyer.txt
mzuehlke, line 1: why did you add this file ?

hkateu: I have no idea how that got generated, but I've taken it out. Submitting changes.

site/concurrency-primitives.md
```diff
@@ -3,12 +3,31 @@
...
+`Channel` implements a publish-subscribe pattern and is particularly useful where we have multiple `publishers` and a single `subscriber`. In the following example, we have two streams, `pub1` and `pub2` publishing the strings `"Hello"` and `"World"` every `1` and `2` seconds respectively. Additionally, we have `sub`, a subscriber that consumes and prints each element. The three streams are then run in parallel and interrupted after `6` seconds.
+
+```scala mdoc:silent
+import cats.effect._
+import fs2.Stream
+import scala.concurrent.duration._
+import cats.effect.unsafe.implicits.global
+import fs2.concurrent.Channel
+
+Channel.unbounded[IO, String].flatMap{channel =>
```
mpilquist, line 23:
```suggestion
Channel.unbounded[IO, String].flatMap { channel =>
```

PR #3480, opened 2024-09-26
Fix typo in `exists` scaladoc

vladislavsheludchenkov:
See this scastie for reference: https://scastie.scala-lang.org/KKWe9upeRR6im3yZDfpDaQ


mzuehlke:
Do you mind adding this as a testcase, if there isn't one already ?

vladislavsheludchenkov:
@mzuehlke I believe [this](https://github.com/typelevel/fs2/blob/eea0c2542efc0d6e6587fd90f6079fc627104cee/core/shared/src/test/scala/fs2/StreamCombinatorsSuite.scala#L553) test covers it, as generator produces empty Stream sometimes. It's not explicit, but the outcome is the same: it compares it to List behavior, and `List.empty.exists` is always false

mzuehlke:
Makes sense 👍

diesalbla (approved):

PR #3489, opened 2024-10-08
Fix `writeUtf8Lines` evaluates leading effects in the input stream twice

### LINKED ISSUE #3488: `writeUtf8Lines` evaluates leading effects in the input stream twice
## Version
Scala - 3.5.1
fs2-core - 3.11.0
fs2-io - 3.11.0

## Minimized code
```scala
import fs2.Stream
import fs2.io.file.Files
import fs2.io.file.Path

import cats.effect.IO

@main def main: Unit = {
  val program = Stream
    .eval(IO { println("side-effect 1"); "out1" })
    .append(Stream.eval(IO { println("side-effect 2"); "out2" }))
    .through(Files.forIO.writeUtf8Lines(Path("text.txt")))
    .compile
    .drain

  program.unsafeRunSync()
}
```

## Console output
```
side-effect 1
side-effect 1
side-effect 2
```

## Expected result
`side-effect 1` is output only once.

## Actual result
`side-effect 1` is output twice.

ChenCMD:
Fixes #3488.

There was a problem with `writeUtf8Lines` executing `in.pull.uncons` and evaluating leading effects, but discarding the result and using `in`.

(Sorry if it is difficult to read due to the use of translation software.)

### DIFF
--- a/io/shared/src/main/scala/fs2/io/file/Files.scala
+++ b/io/shared/src/main/scala/fs2/io/file/Files.scala
@@ -464,8 +464,13 @@ sealed trait Files[F[_]] extends FilesPlatform[F] {
   def writeUtf8Lines(path: Path, flags: Flags): Pipe[F, String, Nothing] = in =>
     in.pull.uncons
       .flatMap {
-        case Some(_) =>
-          in.intersperse(lineSeparator).append(Stream[F, String](lineSeparator)).underlying
+        case Some((next, rest)) =>
+          Stream
+            .chunk(next)
+            .append(rest)
+            .intersperse(lineSeparator)
+            .append(Stream[F, String](lineSeparator))
+            .underlying
         case None => Pull.done
       }
       .stream
--- a/io/shared/src/test/scala/fs2/io/file/FilesSuite.scala
+++ b/io/shared/src/test/scala/fs2/io/file/FilesSuite.scala
@@ -23,7 +23,7 @@ package fs2
 package io
 package file
 
-import cats.effect.{IO, Resource}
+import cats.effect.{IO, Resource, Ref}
 import cats.kernel.Order
 import cats.syntax.all._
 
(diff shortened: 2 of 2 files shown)


io/shared/src/main/scala/fs2/io/file/Files.scala
```diff
@@ -464,8 +464,13 @@ sealed trait Files[F[_]] extends FilesPlatform[F] {
   def writeUtf8Lines(path: Path, flags: Flags): Pipe[F, String, Nothing] = in =>
     in.pull.uncons
       .flatMap {
-        case Some(_) =>
-          in.intersperse(lineSeparator).append(Stream[F, String](lineSeparator)).underlying
+        case Some((next, rest)) =>
```
mpilquist, line 467: Can we just remove the `in.pull.uncons.flatMap` entirely?

ChenCMD: If we keep the “don't add lineSeparator if the stream is empty” specification, I don't think we can remove `in.pull.uncons.flatMap` since we cannot determine if the stream is empty until we evaluate the first element.

mpilquist: Right, ok... can we test a bit with scopes? I worry, perhaps unnecessarily, that the proposed fix might mess with resource finalization -- i.e. some resources might get finalized early.

ChenCMD: It looks like any resource that was to be released at the end of the input stream also gets released at the end of the `uncons`ed-and-`flatMap`ed stream.

I was chatting with @kory33 about this potential corner case. His opinion was that `uncons` is not specified enough, and that we may want to have additional test-cases on `uncons` to ensure such early finalizations do not occur.

## Code
```scala
val program = Stream
  .bracketWeak(IO { println("Resource 1 acquired"); "1" })(_ => IO.println("Resource 1 released"))
  .append(
    Stream
      .bracketWeak(IO { println("Resource 2 acquired"); "2" })(_ => IO.println("Resource 2 released"))
      .append(Stream.emit("3"))
      .append(Stream.exec(IO.println(s"test 1")))
      .scope
  )
  .append(Stream.exec(IO.println(s"test 2")))
  .through(Files.forIO.writeUtf8Lines(Path("test.txt")))
  .compile
  .drain
program.unsafeRunSync()
```

## Output
```
Resource 1 acquired
Resource 2 acquired
test 1
Resource 2 released
test 2
Resource 1 released
```

PR #3492, opened 2024-10-11
Change repeatN to support n=0

L7R7:
While putting together a couple of simple Stream example, I was surprised that `repeatN(0)` fails with an error instead of completing with an empty Stream. I would have expected it to work as `repeat.take(0)` would do.

I'm opening this PR to get some feedback about this proposal. I'm happy to discuss possible consequences of this change.

Apart from avoiding surprises or confusion on the user side, I don't have any real use cases in mind where this would be necessary to have

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2574,9 +2574,9 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * }}}
     */
   def repeatN(n: Long): Stream[F, O] = {
-    require(n > 0, "n must be > 0") // same behaviour as sliding
-    if (n > 1) this ++ repeatN(n - 1)
-    else this
+    require(n >= 0, "n must be >= 0")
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2574,8 +2574,9 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     * }}}
     */
   def repeatN(n: Long): Stream[F, O] = {
-    require(n > 0, "n must be > 0") // same behaviour as sliding
-    if (n > 1) this ++ repeatN(n - 1)
+    require(n >= 0, "n must be >= 0")
+    if (n > 0) this ++ repeatN(n - 1)
+    else if (n == 0) Stream.empty
     else this
```
ybasket, line 2580: As we know from the `require`, the only case not covered in the `if(n > 0)` is exactly `0`, so maybe simplify to
```suggestion
    else Stream.empty
```

L7R7: Oh yes, absolutely! I'll amend my commit

Issue #3485, opened 2024-10-01
prefetch hangs on a canceled stream


enlait:
fs2 version 3.10.2

this code does not terminate:
```scala
Stream.eval(IO.canceled).prefetch.compile.drain
```

expected behavior: prefetch should not change execution semantics, at least not so much as to prevent termination

ValdemarGr:
Since `concurrently` doesn't guarantee that the foreground stream is closed when the background reaches termination, then I think the channel should be closed once the `this` stream exits.
```diff
  def prefetchN[F2[x] >: F[x]: Concurrent](
      n: Int
  ): Stream[F2, O] =
    Stream.eval(Channel.bounded[F2, Chunk[O]](n)).flatMap { chan =>
      chan.stream.unchunks.concurrently {
-        chunks.through(chan.sendAll)
+        chunks.through(chan.sendAll).onFinalize(chan.close)
      }
    }
```

[Here](
https://github.com/typelevel/fs2/blob/6352df279e1a69f9672fcd5f1e1ccf9dfd26ba63/core/shared/src/main/scala/fs2/Stream.scala#L2437-L2444).

ValdemarGr:
Further note, this will cause the foreground stream to exit with `Success` which might not be correct. I think `Canceled` should be the right outcome in that case.

```diff
  def prefetchN[F2[x] >: F[x]: Concurrent](
      n: Int
  ): Stream[F2, O] =
-   Stream.eval(Channel.bounded[F2, Chunk[O]](n)).flatMap { chan =>
+   Stream.eval(Channel.bounded[F2, Either[ExitCase, Chunk[O]]](n)).flatMap { chan =>
      chan.stream.unchunks.concurrently {
-        chunks.through(chan.sendAll)
+        chunks.map(_.asRight[ExitCase]).through(chan.sendAll).onFinalizeCase(chan.send(_) >> chan.close)
      }
    }.flatMap{
      case Right(x) => Stream.chunk(x)
      case Left(ec) => Stream.exec(ec.toOutcome[F].embed(Concurrent[F].canceled))
    }
```

Maybe `prefetch` can be expressed via one of the other combinators that run two streams concurrently which also guarantees communication of cancellation and errors if any of the concurrent streams exits (`merge`? but it doesn't look like it handles cancellation properly).

PR #3505, opened 2024-11-21
Prevent deadlock when publisher is canceled

### LINKED ISSUE #3502: Stream hangs on error with broadcastThrough and readOutputStream
fs2 version: 3.11.0
scala version: 3.3.4

[Scastie](https://scastie.scala-lang.org/eDo0xTsNQJmG5UTC88Xl8g)

The following code snippet throws a `java.util.NoSuchElementException: None.get`, which means that the stream is hanging.

I expect it to throw `java.lang.Exception: boom` instead.

```scala
import cats.effect.IO

fs2.Stream
  .emit[IO, Byte](1)
  .evalTap(_ => IO.raiseError(new Exception("boom")))
  .broadcastThrough(in =>
    fs2.io.readOutputStream(1) { out =>
      in.evalMap(b => IO.interruptible(out.write(b))).compile.drain
    }
  )
  .compile
  .drain
  .unsafeRunTimed(2.second)
  .get
```

dimitriho:
Also fixes #3502

### DIFF
--- a/core/shared/src/main/scala/fs2/concurrent/Topic.scala
+++ b/core/shared/src/main/scala/fs2/concurrent/Topic.scala
@@ -23,6 +23,7 @@ package fs2
 package concurrent
 
 import cats.effect._
+import cats.effect.Resource.ExitCase
 import cats.syntax.all._
 import scala.collection.immutable.LongMap
 
@@ -220,7 +221,8 @@ object Topic {
         }
 
         def publish: Pipe[F, A, Nothing] = { in =>
-          in.onFinalize(close.void)
+          in
+            .onFinalizeCase(closeWithExitCase(_).void)
             .evalMap(publish1)
             .takeWhile(_.isRight)
             .drain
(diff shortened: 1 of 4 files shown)


BalmungSan (approved):

core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -208,7 +208,7 @@ object Topic {
         }
 
         def publish: Pipe[F, A, Nothing] = { in =>
-          (in ++ Stream.exec(close.void))
+          in.onFinalize(close.void)
```
mpilquist, line 211: Do we need the same fix in [Channel](https://github.com/typelevel/fs2/blob/abe9b03e205d05d41ca2950e0d9f5db7c77a26cf/core/shared/src/main/scala/fs2/concurrent/Channel.scala#L154)

dimitriho: I made the change, it should also fix #3485. But as @ValdemarGr [pointed out](https://github.com/typelevel/fs2/issues/3485#issuecomment-2388220538), the exit case might not be correct.

BalmungSan (approved):

core/shared/src/main/scala/fs2/concurrent/Topic.scala
dimitriho: Here is an attempt to handle cancelation and error in Channel: e1d7b0925. What do you think?

mpilquist: Looks good to me!

dimitriho: I added this commit. Regarding Topic, I don't know how to implement the same behavior without exposing `raiseError` and `canceled` to Channel.

dimitriho: I ended up adding `raiseError` and `cancel` to `Channel`, but now I'm facing mima issues. Is it acceptable to add new methods?

Issue #3474, opened 2024-09-13
`Pull#lease`, `Pull.bracketCase` and `Pull.extendScopeTo` are not cancellation safe.


ValdemarGr:
`Pull.extendScopeTo` [here](https://scastie.scala-lang.org/vcAPQVhjRZCraHj2QsBpcQ)
A safer variant could be:
```scala
object Pull {
  ...
  def leaseScope[F[_]: MonadThrow]: Pull[F, Nothing, Resource[F, Unit]] = 
    Pull.getScope[F].map(s => Resource.make(s.lease)(_.cancel.rethrow))
}
```
<details>
<summary>
The code that can produce the error.
</summary>

```scala
import cats.effect.*
import cats.effect.unsafe.implicits.global
import fs2.*
import scala.concurrent.duration.*
import cats.implicits.*
import cats.*

object Main extends IOApp.Simple {
  def run = IO.deferred[Unit].flatMap { d =>
    IO.ref(0).flatMap { ref =>
      val s =
        Stream
          .resource(Resource.make(ref.update(_ + 1))(_ => ref.update(_ - 1)))
          .pull
          .uncons1
          .flatMap {
            case None => Pull.done
            case Some((hd, tl)) =>
              Pull.output1(hd) >>
                Pull.extendScopeTo(Stream.unit.covary[IO]).flatMap {
                  killLease =>
                    Pull.eval(d.complete(()).void >> IO.never[Unit]) >>
                      Pull.eval(killLease.compile.drain) >>
                      tl.pull.echo
                }
          }
          .streamNoScope

      IO.race(s.compile.drain, d.get) >> ref.get.flatMap(IO.println(_))
    }
  }
}
```

</details>

`Pull.bracketCase` can be cancelled after the `acquire` and during `use`.
I don't know what uses `Pull.bracketCase` has, other than `Pull#lease`, but I think it should at least be documented that it shouldn't be used for resources, since `release` is not guaranteed.

`Pull#lease` is subject to the issues of `bracketCase` and can cause resources on the scope to never be released.
I think it could be implemented in a safe way using the new `leaseScope` introduced above.
```scala
sealed abstract class Pull[...] {
  ...
  // binary incomat with `lease`
  def leaseSafe(implicit F: MonadCancel[F, _]): Pull[...] =
    Pull.leaseScope.flatMap(r => Stream.resourceWeak(r).pull.echo >> this)
}
```

PR #3512, opened 2024-12-26
Fix #3076 parEvalMap resource scoping

### LINKED ISSUE #3076: parEvalMap* runs resource finaliser before usage
Found by: @lavrov 

Example:
```scala

import cats.effect.*
import fs2.*

  def run =
    Stream
      .parEvalMap(2)(id =>
      )
      .compile
      .drain
```

Expected logs:
```
acquire
use
release
```

Actual logs:
```
acquire
release
use
```

reardonj:
Updates parEvalMap*,  broadcastThrough and prefetch to extend the resource scope past the channel/topic used to implement concurrency for these operators.

In all cases the solution is the same:

1. use `underlying.uncons.flatMap` to get into a Pull context with access to the source stream
2. setup the new foreground stream with Pull.extendScopeTo to pull the scope across to the new stream

~~I don't see an obvious way to extract this to something general. The extra finalization in the background stream for `parEvalMapUnorderedUnbounded` means I can't just make a `concurrently` variant that takes the background stream, since the foreground needs to consume the transformed stream, while only taking the original stream's scope.~~

Edit: I have abstracted this solution to a new `extendScopeThrough` method, which works as `through` except propagating the scope to the new stream. It appears we should be doing this for any stream combinator where the resulting stream is not directly derived from the current stream (e.g. any combinator that uses an internal buffer, channel, topic, or such).

I also had to fix the cancallation safety of `extendScopeTo` (see #3474). The `StreamSuite` "resource safety test 4" started failing after my changes, presumably because the scope started propagating far enough to actually get hit by cancellation.

There is at least `conflateChunks` broken the same way but I'd like to validate this solution is correct before I continue trying to chase down all the places the fix should be applied.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -238,19 +238,19 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     */
   def broadcastThrough[F2[x] >: F[x]: Concurrent, O2](pipes: Pipe[F2, O, O2]*): Stream[F2, O2] = {
     assert(pipes.nonEmpty, s"pipes should not be empty")
-    underlying.uncons.flatMap {
-      case Some((hd, tl)) =>
+    extendScopeThrough { source =>
+      Stream.force {
         for {
           // topic: contains the chunk that the pipes are processing at one point.
           // until and unless all pipes are finished with it, won't move to next one
-          topic <- Pull.eval(Topic[F2, Chunk[O]])
+          topic <- Topic[F2, Chunk[O]]
           // Coordination: neither the producer nor any consumer starts
           // until and unless all consumers are subscribed to topic.
-          allReady <- Pull.eval(CountDownLatch[F2](pipes.length))
+          allReady <- CountDownLatch[F2](pipes.length)
+        } yield {
+          val checkIn = allReady.release >> allReady.await
 
-          checkIn = allReady.release >> allReady.await
-
-          dump = (pipe: Pipe[F2, O, O2]) =>
+          def dump(pipe: Pipe[F2, O, O2]): Stream[F2, O2] =
             Stream.resource(topic.subscribeAwait(1)).flatMap { sub =>
               // Wait until all pipes are ready before consuming.
               // Crucial: checkin is not passed to the pipe,
@@ -258,19 +258,13 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
               Stream.exec(checkIn) ++ pipe(sub.unchunks)
             }
 
-          dumpAll: Stream[F2, O2] <-
-            Pull.extendScopeTo(Stream(pipes: _*).map(dump).parJoinUnbounded)
-
-          chunksStream = Stream.chunk(hd).append(tl.stream).chunks
-
+          val dumpAll: Stream[F2, O2] = Stream(pipes: _*).map(dump).parJoinUnbounded
           // Wait until all pipes are checked in before pulling
-          pump = Stream.exec(allReady.await) ++ topic.publish(chunksStream)
-
-          _ <- dumpAll.concurrently(pump).underlying
-        } yield ()
-
-      case None => Pull.done
-    }.stream
+          val pump = Stream.exec(allReady.await) ++ topic.publish(source.chunks)
+          dumpAll.concurrently(pump)
+        }
+      }
+    }
   }
 
   /** Behaves like the identity function, but requests `n` elements at a time from the input.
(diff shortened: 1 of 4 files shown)


reardonj:
Looking at some other concurrency issues, I wonder if this should be using stepLeg.

core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -2921,6 +2925,23 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
...
+  /** Transforms this stream, explicitly extending the current scope through the given pipe.
+    *
+    * Use this when implementing a pipe where the resulting stream is not directly constructed from
+    * the source stream, e.g. when sending the source stream through a Channel and returning the
+    * channel's stream.
+    */
+  def extendScopeThrough[F2[x] >: F[x], O2](
+      f: Stream[F, O] => Stream[F2, O2]
+  )(implicit F: MonadError[F2, Throwable]): Stream[F2, O2] =
+    this.pull.peek
```
yurique, line 2937: Is it guaranteed to be safe (in the context of scopes) to use `.peek` before `Pull.extendScopeTo`?

reardonj: I'm assuming scopes work properly for simple streams. 🤞 Do you have a specific scenario in mind?

The fundamental issue is that I need to get ahold of a scope before we go adding more finalizers to `source` to avoid deadlocking in parEvalMap and I don't see a way to do that without `uncons`. `peek` is just a convenience helper.

yurique: > Do you have a specific scenario in mind?

No, not really, it's just me being paranoid :)

I don't see a way to do it without `uncons` either, I was thinking about swapping the order of those `Pull`s:
```scala
def extendScopeThrough[F2[x] >: F[x], O2](
      f: Stream[F, O] => Stream[F2, O2]
  )(implicit F: MonadError[F2, Throwable]): Stream[F2, O2] =
    Pull
      .extendScopeTo(this.covary[F2])
      .flatMap { stream =>
        stream.pull.peek.flatMap {
          case Some((_, tl)) => f(tl).underlying // <---------------
          case None          => f(Stream.empty).underlying
        }
      }
      .stream
```

but the problem is `tl` is not a `Stream[F, O]` (it's a `Stream[F2, O]`), and `.covary[F2]` at the beginning has to be there because we only have `MonadError` for `F2`.

reardonj: Just make it `f: Stream[F2, O] => Stream[F2, O2]`. Requires type annotations at the call sites, but otherwise, this version also passes the tests.

Unless we can find an observable difference I'd rather keep the current version to avoid the extra annotations.

yurique: I agree :)

reardonj:
This solution is inadequate. It only handles scopes open at the first pull.

yurique:
@reardonj 

wouldn't the `Scope.lease` (that is used in the `Pull.extendScopeTo`) handle the subsequent (child) scopes?

> Leases the resources of this scope until the returned lease is cancelled.
>    Note that this leases all resources in this scope, resources in all parent scopes (up to root)
>    and resources of all child scopes.

(unless, of course, I'm mis-understanding what the above means)

yurique:
I tried some ridiculous things like the following:
```scala
  fs2.Stream.unit
    .covary[IO]
    .flatMap { _ =>
      fs2.Stream.bracket {
        IO.println("making resource 1").as("TempFile")
      } { res =>
        IO.println(s"!! releasing resource 1: $res")
      }
    }
    .flatMap { _ =>
      fs2.Stream.bracket {
        IO.println("making resource 2").as("TempFile")
      } { res =>
        IO.println(s"!! releasing resource 2: $res")
      }
    }
    .parEvalMap(2) { _ =>
      IO.println("eval")
    }
    .flatMap { _ =>
      fs2.Stream.bracket {
        IO.println("making resource 3").as("TempFile")
      } { res =>
        IO.println(s"!! releasing resource 3: $res")
      }
    }
    .parEvalMap(2) { _ =>
      IO.println("eval 2")
    }
    .flatMap { _ =>
      fs2.Stream.bracket {
        IO.println("making resource 4").as("TempFile")
      } { res =>
        IO.println(s"!! releasing resource 4: $res")
      }
    }
    .parEvalMap(2) { _ =>
      IO.println("eval 3")
    }
    .compile
    .drain
    .unsafeRunSync()
```

and it works like a clock

```
making resource 1
making resource 2
eval
making resource 3
eval 2
making resource 4
eval 3
!! releasing resource 4: TempFile
!! releasing resource 3: TempFile
!! releasing resource 2: TempFile
!! releasing resource 1: TempFile
```

reardonj:
@yurique , try a stream that creates multiple scopes. I just pushed the test that I broke based on the `Network[IO].server` conversation:

```scala
      Stream(1, 2, 3, 4, 5, 6)
        .flatMap(i => Stream.bracket(Deferred[IO, Int])(_.complete(i).void)) // 1
        .parEvalMap(2)(d => IO.sleep(1.second) >> d.complete(0))
        .evalMap(completed => IO.raiseWhen(!completed)(new RuntimeException("already completed")))
        .timeout(5.seconds)
        .compile
        .last
        .assertEquals(Some(()))
```

As I understand, each element passing through `// 1` opens another scope, but only the first scope is actually propagated

Issue #3182, opened 2023-03-24
Non-blocking backend for `Process` APIs


armanbilge:
There are two interesting candidates for this:
- https://github.com/brettwooldridge/NuProcess
- https://github.com/jnr/jnr-process

Probably NuProcess is the better established of the two. jnr-process seems to expose a lower-level API. Suffice to say, both need a closer look, particularly at the threading model.

Similar to `UnixSockets`, we can add one of these as an optional dependency, and use reflection to try and detect it at runtime, otherwise falling back to the blocking JDK APIs.

rossabaker:
Eek!  Optional dependencies are a linker disaster waiting to happen.  Even with fallbacks, the end user has to maintain a fragile dependency relationship instead of the standard transitivity and eviction warnings.  [Quoth the Maven](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) (emphasis mine):

> **Optional dependencies are used when it's not possible (for whatever reason) to split a project into sub-modules**. The idea is that some of the dependencies are only used for certain features in the project and will not be needed if that feature isn't used. **Ideally, such a feature would be split into a sub-module that depends on the core functionality project.** This new subproject would have only non-optional dependencies, since you'd need them all if you decided to use the subproject's functionality.
>
> However, since the project cannot be split up (again, for whatever reason), these dependencies are declared optional. If a user wants to use functionality related to an optional dependency, they have to redeclare that optional dependency in their own project. This is not the clearest way to handle this situation, but both **optional dependencies and dependency exclusions are stop-gap solutions**.

I would strongly recommend a module for this sort of thing.

rossabaker:
Some comparisons to other places we do classpath discovery:

1. SLF4J bindings, JDBC drivers: these implement a specified API, and the implementations are mostly called only called via that API, so the diamond risk is relatively low.  Also, it wasn't our decision.
2. UnixDomainSocket: JNR is its own API with its own dependencies, so carries somewhat higher risk.  It looks to have been stable the past few years, but a breaking change will be a sad day for FS2.  A saving grace is that Java >= 17 adoption is asymptotically approaching 100%.
3. These process libraries: similar risks to UnixDomainSocket, but without Java 17 adoption relieving the problem over time.

All this magic potentially gets a better implicit on the algebra's companion object, but I don't think it's worth the classpath landmine.

mpilquist:
Are the benefits of a non-blocking process backend worth the maintenance costs?

[...]

PR #3539, opened 2025-03-06
run `Process#waitFor` on virtual thread if available

onsah:
Run `Process.waitFor` in virtual thread if possible to solve https://github.com/typelevel/fs2/issues/3182 

I am not sure if this is the correct approach but I wanted to share my progress to get fast feedback.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,7 @@ ThisBuild / crossScalaVersions := Seq("2.12.20", Scala213, "3.3.5")
 ThisBuild / tlVersionIntroduced := Map("3" -> "3.0.3")
 
 ThisBuild / githubWorkflowOSes := Seq("ubuntu-latest")
-ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("21"))
+ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("17"), JavaSpec.temurin("21"))
 ThisBuild / githubWorkflowBuildPreamble ++= nativeBrewInstallWorkflowSteps.value
 ThisBuild / nativeBrewInstallCond := Some("matrix.project == 'rootNative'")
 
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,7 @@ ThisBuild / crossScalaVersions := Seq("2.12.20", Scala213, "3.3.5")
 ThisBuild / tlVersionIntroduced := Map("3" -> "3.0.3")
 
 ThisBuild / githubWorkflowOSes := Seq("ubuntu-latest")
-ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("17"))
+ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("21"))
 ThisBuild / githubWorkflowBuildPreamble ++= nativeBrewInstallWorkflowSteps.value
 ThisBuild / nativeBrewInstallCond := Some("matrix.project == 'rootNative'")
 
--- a/io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
@@ -36,7 +36,8 @@ import scala.concurrent.ExecutionContext
 private[process] trait ProcessesCompanionPlatform {
   def forAsync[F[_]](implicit F: Async[F]): Processes[F] = new UnsealedProcesses[F] {
 
-    private def javaVersion: Int = System.getProperty("java.version").stripPrefix("1.").takeWhile(_.isDigit).toInt
+    private def javaVersion: Int =
+      System.getProperty("java.version").stripPrefix("1.").takeWhile(_.isDigit).toInt
 
     def spawn(process: ProcessBuilder): Resource[F, Process[F]] =
       Resource
@@ -73,7 +74,7 @@ private[process] trait ProcessesCompanionPlatform {
                     .asInstanceOf[ExecutorService]
 
                   val vtEC = ExecutionContext.fromExecutor(virtualThreadExecutor)
-                    
+
                   F.evalOn(f, vtEC)
                 } else {
                   f
--- a/io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
@@ -25,35 +25,13 @@ package process
 
 import cats.effect.kernel.Async
 import cats.effect.kernel.Resource
-import cats.syntax.all._
-import fs2.io.CollectionCompat._
+import cats.syntax.all.*
+import fs2.io.CollectionCompat.*
 
 import java.lang
-import java.util.concurrent.Executors
-import java.util.concurrent.ExecutorService
-import scala.concurrent.ExecutionContext
 
 private[process] trait ProcessesCompanionPlatform {
   def forAsync[F[_]](implicit F: Async[F]): Processes[F] = new UnsealedProcesses[F] {
-
-    private lazy val vtEC: ExecutionContext = {
-      val virtualThreadExecutor = classOf[Executors]
-        .getDeclaredMethod("newVirtualThreadPerTaskExecutor")
-        .invoke(null)
-        .asInstanceOf[ExecutorService]
-
-      ExecutionContext.fromExecutor(virtualThreadExecutor)
-    }
-
-    private val javaVersion: Int =
-      System.getProperty("java.version").stripPrefix("1.").takeWhile(_.isDigit).toInt
-
-    private def evOnVirtualThreadECIfPossible[A](f: => F[A]): F[A] =
-      if (javaVersion >= 21) {
-        F.evalOn(f, vtEC)
-      } else {
-        f
-      }
 
     def spawn(process: ProcessBuilder): Resource[F, Process[F]] =
       Resource
@@ -76,7 +54,7 @@ private[process] trait ProcessesCompanionPlatform {
         } { process =>
           F.delay(process.isAlive())
             .ifM(
-              evOnVirtualThreadECIfPossible(
+              evalOnVirtualThreadIfAvailable(
                 F.blocking {
                   process.destroy()
                   process.waitFor()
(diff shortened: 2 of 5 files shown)


armanbilge (commented):
Nice work, thanks for opening the PR, this is the right approach!

io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
```diff
@@ -29,9 +29,16 @@ import cats.syntax.all._
...
 
 import java.lang
+import java.util.concurrent.Executors
+import java.util.concurrent.ExecutorService
+import scala.concurrent.ExecutionContext
 
 private[process] trait ProcessesCompanionPlatform {
   def forAsync[F[_]](implicit F: Async[F]): Processes[F] = new UnsealedProcesses[F] {
+
+    private def javaVersion: Int =
```
onsah, line 39: Is there a centralized place for such things?

io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
```diff
@@ -53,10 +60,25 @@ private[process] trait ProcessesCompanionPlatform {
...
+                  ()
+                }
+                // Run in virtual thread if possible
+                if (javaVersion >= 21) {
+                  val virtualThreadExecutor = classOf[Executors]
+                    .getDeclaredMethod("newVirtualThreadPerTaskExecutor")
+                    .invoke(null)
+                    .asInstanceOf[ExecutorService]
+
+                  val vtEC = ExecutionContext.fromExecutor(virtualThreadExecutor)
```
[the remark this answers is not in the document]
onsah: That's a good idea! Should it be a `lazy val` so it's not created if `spawn` is never used?

io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
armanbilge: Hmm, maybe you could put this stuff here?
https://github.com/typelevel/fs2/blob/3b56691a0830580ea113ef391c2fc169a6f17b97/core/jvm/src/main/scala/fs2/internal/ThreadFactories.scala#L31

onsah: Is this a place to move `javaVersion`? I wasn't explicit but that I was asking specifically for `javaVersion`.

onsah:
@armanbilge I think native test suites crash due to not being able to find `getDeclaredMethod` https://github.com/typelevel/fs2/actions/runs/13701427909/job/38315972895?pr=3539 How can we solve it?

io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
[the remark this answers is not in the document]
onsah: Updated

io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
```diff
@@ -53,10 +60,25 @@ private[process] trait ProcessesCompanionPlatform {
...
+
+                  F.evalOn(f, vtEC)
+                } else {
+                  f
```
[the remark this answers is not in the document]
onsah: Done!

io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
armanbilge: It's a place where we can move the `lazy val` for the virtual thread executor, not just `javaVersion`.

armanbilge:
> I think native test suites crash due to not being able to find `getDeclaredMethod`

That's correct, reflection is not supported on Scala Native.

---

> How can we solve it?

We should move this code out of `jvm-native` sources into `jvm`, since it only works on the JVM.

armanbilge:
> this means we need two `ProcessesCompanionPlatform` then. Is it correct?

Well, that's one possible way ... but I think we should just move all this stuff into a different file, that's already split for the 3 platforms (JVM/JS/Native).

For example, we can add a method to `ioplatform`:
```scala
def evalOnVirtualThreadIfAvailable[F[_], A](fa: F[A]): F[A]
```

On the JVM, we can move it onto the VT executor if on JDK >= 21. On Native, we can do nothing.

io/jvm/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -144,4 +146,25 @@ private[fs2] trait ioplatform extends iojvmnative {
...
+
+      ExecutionContext.fromExecutor(virtualThreadExecutor).some
+    } else {
+      None
+    }
+
+  }
+
+  def evalOnVirtualThreadIfAvailable[F[_]: Async, A](fa: F[A]): F[A] =
+    vtExecutor.fold(fa)(ec => fa.evalOn(ec))
```
onsah, line 168: AFAIK accessing a `lazy val` is costly for the first initialization but not worse than a regular if check afterwards.

io/jvm/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -144,4 +146,25 @@ private[fs2] trait ioplatform extends iojvmnative {
     }
   }
 
+  private lazy val vtExecutor: Option[ExecutionContext] = {
```
armanbilge, line 149: Since this is private code, instead of `Option` we can just use `null`, that will be a bit more efficient than the `fold`.

io/jvm/src/main/scala/fs2/io/ioplatform.scala
armanbilge: > AFAIK accessing a `lazy val` is costly for the first initialization

That's correct.

> not worse than a regular if check afterwards.

Well, not quite. Because `lazy val`s are synchronized for thread-safe initialization, reading one is equivalent to reading a `@volatile val`, which can be more expensive than a non-volatile. But I don't think we need to worry too much about it, because working with processes already involves lots of synchronization and memory barriers.

---

If we really cared, instead of using `lazy val` we could use a `var` with our own non-thread-safe initialization scheme. In this case, we don't care if we create the VT executor more than once during initialization, it's not a big deal.

io/jvm/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -144,4 +146,25 @@ private[fs2] trait ioplatform extends iojvmnative {
...
+        .asInstanceOf[ExecutorService]
+
+      ExecutionContext.fromExecutor(virtualThreadExecutor).some
+    } else {
+      None
+    }
+
+  }
+
+  def evalOnVirtualThreadIfAvailable[F[_]: Async, A](fa: F[A]): F[A] =
```
armanbilge, line 167: Oh also, let's keep this private to FS2.
```suggestion
  private[io] def evalOnVirtualThreadIfAvailable[F[_]: Async, A](fa: F[A]): F[A] =
```

armanbilge (commented):
Nice work! One last change, we should make sure we test the Virtual Threads in CI using Java 21 (run `sbt githubWorkflowGenerate` to update the CI workflow):

https://github.com/typelevel/fs2/blob/29d687631ab5954b74856be12019d1d720e4d908/build.sbt#L18

build.sbt
```diff
@@ -15,7 +15,7 @@ ThisBuild / crossScalaVersions := Seq("2.12.20", Scala213, "3.3.5")
 ThisBuild / tlVersionIntroduced := Map("3" -> "3.0.3")
 
 ThisBuild / githubWorkflowOSes := Seq("ubuntu-latest")
-ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("17"))
+ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("21"))
```
armanbilge, line 18: Sorry, I should have clarified: let's _add_ JDK 21 to the CI, but let's also still test on JDK 17.

io/native/src/main/scala/fs2/io/ioplatform.scala
```diff
@@ -32,4 +32,7 @@ private[fs2] trait ioplatform extends iojvmnative {
   def stdinUtf8[F[_]](bufSize: Int)(implicit F: Sync[F]): Stream[F, String] =
     stdin(bufSize).through(text.utf8.decode)
 
+  // Scala-native doesn't support virtual threads
+  def evalOnVirtualThreadIfAvailable[F[_], A](fa: F[A]): F[A] = fa
```
armanbilge, line 36:
```suggestion
  private[io] def evalOnVirtualThreadIfAvailable[F[_], A](fa: F[A]): F[A] = fa
```

onsah:
@armanbilge I run `Test/dependencyTree` on the scalafix project and I see that `scalatest` version is `3.2.19`. I searched for `FunSuiteLike` in docs but it doesn't exist https://www.scalatest.org/scaladoc/3.2.19/org/index.html?search=FunSuiteLike. There is a similar trait called `AnyFunSuiteLike`. Should we change to that or drop the dependency version of scalafix?

armanbilge:
@onsah good investigation! Yes, I think we should switch to `AnyFunSuiteLike`.

onsah:
@armanbilge seems like there is a bug in one of the scalafix rules that causes a test to fail. Seems like irrelevant to the changes on this PR. Is it a general issue?

armanbilge:
@onsah the problem is that now that we added JDK 21 to the CI, the scalafix had not been updated in a while and was not working.

I tried a bit too hard to fix it in your PR (updating versions, fixing the test suite, but now we still needother changes) ... I'm sorry, we should have just resolved in a separate PR 😅 

In anycase, it's not a "bug" per se, just some changes we need to make as part of the version update.

armanbilge:
- https://github.com/typelevel/fs2/pull/3544

onsah:
Thank you! I checked the scalafix PR. As soon as it's merged I can rebase this PR.

armanbilge:
there you go, it's merged :)

armanbilge (approved):
Nice work, this looks good to me!

armanbilge:
@onsah I'll give a little time in case anybody else wants to review, otherwise I will merge in the next day or two :)

Btw, I just realized, we can also use virtual threads for the JDK unix sockets implementation:
https://github.com/typelevel/fs2/blob/1d67ef4740b9aa87a39858e42e6e99b55f1a46fe/io/jvm/src/main/scala/fs2/io/net/unixsocket/JdkUnixSockets.scala#L41

onsah:
@armanbilge I will check this in the meantime, if it's already merged I can prepare a separate PR.

PR #3543, opened 2025-03-08
Added output configurations

Dralt03:
Exposed a Configuration for the user to select how they want to deal with sterr and stdout for streams in [io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala](https://github.com/typelevel/fs2/blob/main/io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala)

### DIFF
--- a/io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
+++ b/io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
@@ -1,24 +1,3 @@
-/*
- * Copyright (c) 2013 Functional Streams for Scala
- *
- * Permission is hereby granted, free of charge, to any person obtaining a copy of
- * this software and associated documentation files (the "Software"), to deal in
- * the Software without restriction, including without limitation the rights to
- * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
- * the Software, and to permit persons to whom the Software is furnished to do so,
- * subject to the following conditions:
- *
- * The above copyright notice and this permission notice shall be included in all
- * copies or substantial portions of the Software.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
- * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
- * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
- * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
- * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- */
-
 package fs2.io
 package process
 
@@ -49,6 +28,9 @@ sealed abstract class ProcessBuilder private {
     */
   def workingDirectory: Option[Path]
 
+  /** Configures how stdout and stderr should be handled. */
+  def outputMode: ProcessOutputMode
+
   /** @see [[command]] */
   def withCommand(command: String): ProcessBuilder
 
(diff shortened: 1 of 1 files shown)


io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
```diff
@@ -67,17 +70,29 @@ sealed abstract class ProcessBuilder private {
...
+
   /** Starts the process and returns a handle for interacting with it.
     * Closing the resource will kill the process if it has not already terminated.
     */
   final def spawn[F[_]: Processes]: Resource[F, Process[F]] =
     Processes[F].spawn(this)
 }
 
+sealed trait ProcessOutputMode
+object ProcessOutputMode {
```
armanbilge, line 84: These look like the right set of options! We need to support `stdin`, `stdout`, and `stderr` to be configured separately.

armanbilge (commented):
Nice start! Do you think you can also make the necessary changes to the `Processes` implementations?

io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
```diff
@@ -37,60 +38,81 @@ private[process] trait ProcessesCompanionPlatform {
...
+                builder.redirectOutput(Redirect.to(path.toNioPath.toFile))
+              case StreamOutputMode.Pipe =>
             }
 
-            builder.start()
-          }
-        } { process =>
-          F.delay(process.isAlive())
-            .ifM(
-              evalOnVirtualThreadIfAvailable(
```
armanbilge, line 57: Deleted by mistake?

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
```diff
@@ -67,17 +70,34 @@ sealed abstract class ProcessBuilder private {
...
     Processes[F].spawn(this)
 }
 
+sealed trait StreamOutputMode
+object StreamOutputMode {
+  case object Pipe extends StreamOutputMode       
+  case object Inherit extends StreamOutputMode   
+  case object Ignore extends StreamOutputMode    
+  case class FileOutput(path: Path) extends StreamOutputMode 
+}
```
armanbilge, line 89:
```suggestion
sealed abstract class StreamRedirect
object StreamRedirect {
  case object Pipe extends StreamRedirect       
  case object Inherit extends StreamRedirect   
  case object Discard extends StreamRedirect    
  final case class File(path: Path) extends StreamRedirect 
}
```

io/js/src/main/scala/fs2/io/process/ProcessesPlatform.scala
```diff
@@ -36,19 +36,47 @@ private[process] trait ProcessesCompanionPlatform {
...
         F.async_[(Process[F], F[Unit])] { cb =>
-          val childProcess = facade.child_process.spawn(
-            process.command,
-            process.args.toJSArray,
-            new facade.child_process.SpawnOptions {
-              cwd = process.workingDirectory.fold[js.UndefOr[String]](js.undefined)(_.toString)
-              env =
+          
+
+          val spawnOptions = js.Dynamic.literal {
```
armanbilge, line 41: Instead of using `js.Dynamic` can you update the `SpawnOptions` face to add the additional fields?

io/shared/src/main/scala/fs2/io/process/ProcessBuilder.scala
```diff
@@ -67,17 +70,34 @@ sealed abstract class ProcessBuilder private {
   /** @see [[workingDirectory]] */
   def withCurrentWorkingDirectory: ProcessBuilder
 
+  /** @see [[outputMode]] */
+  def withOutputConfig(outputConfig: ProcessOutputConfig): ProcessBuilder
```
armanbilge, line 74: I think separating it out into three config methods for stdin/stdout/stderr (like the JDK process builder) will make the API easier to use.

io/jvm-native/src/main/scala/fs2/io/process/ProcessesPlatform.scala
Dralt03: Must've been deleted while solving merge conflicts adding them back

armanbilge (commented):
Thanks for your work on this PR! Tests currently are not passing on JVM or JS. (Native is broken for reasons not related to you.)

Dralt03:
@armanbilge The tests for rootJS are currently not passing for the merged option. Do you have an idea on how we can proceed with that. I'm thinking about setting the stdio to `["pipi", "pipe", 1]` to show if it should be merged but I'm not sure if that would work so I need some help with that

armanbilge:
I think Node.js might not support that feature, so we can just implement it with FS2 `Stream#merge`.

Issue #3542, opened 2025-03-07
`fs2.interop.flow.PublisherToSubscriberSpec.should cancel upstream if downstream completes` flake


armanbilge:
3b56691a0830580ea113ef391c2fc169a6f17b97

ubuntu-latest, 3, temurin@17, rootJVM

Unfortunately, ScalaCheck minimized the buffer size argument to zero, which is invalid, so I think we lost the original error ...

```
==> X fs2.interop.flow.PublisherToSubscriberSpec.should cancel upstream if downstream completes  0.565s munit.FailException: Failing seed: 82KDw0NDjWAXJUVVYTrYStmnTfKrZEE9oA8wbVXJAuG=
You can reproduce this failure by adding the following override to your suite:

  override def scalaCheckInitialSeed = "82KDw0NDjWAXJUVVYTrYStmnTfKrZEE9oA8wbVXJAuG="

Exception raised on property evaluation.
> ARG_0: List(2147483647)
> ARG_1: List(923319774, 2147483647, -140306582, 2147483647, -2147483648, 1239258450, -154613030)
> ARG_2: 0
> ARG_2_ORIGINAL: 41
> Exception: java.lang.IllegalArgumentException: requirement failed: The buffer size MUST be positive
    at munit.Assertions.fail(Assertions.scala:256)
    at munit.Assertions.fail$(Assertions.scala:16)
    at munit.FunSuite.fail(FunSuite.scala:11)
    at munit.ScalaCheckEffectSuite.munit$ScalaCheckEffectSuite$$parseTestResult(ScalaCheckEffectSuite.scala:80)
    at munit.ScalaCheckEffectSuite.checkPropF$$anonfun$2(ScalaCheckEffectSuite.scala:68)
    at apply @ munit.CatsEffectAssertions.assertIO$$anonfun$1(CatsEffectAssertions.scala:52)
    at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
    at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
    at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
    at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
    at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
    at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
[...]

Issue #3541, opened 2025-03-07
`fs2.interop.flow.PublisherToSubscriberSpec.should have the same output as input` timeout


armanbilge:
773de68f672c4e736bf642e9c30b3d915a8fea82

ubuntu-latest, 3, temurin@17, rootJVM

```
==> X fs2.interop.flow.PublisherToSubscriberSpec.should have the same output as input  30.004s java.util.concurrent.TimeoutException: test timed out after 30 seconds
    at munit.CatsEffectSuite.munit$CatsEffectSuite$$anon$1$$_$_$$anonfun$2(CatsEffectSuite.scala:78)
    at scala.Option.fold(Option.scala:263)
    at munit.CatsEffectSuite$$anon$1.applyOrElse(CatsEffectSuite.scala:79)
    at scala.PartialFunction$Lifted.apply(PartialFunction.scala:338)
    at scala.PartialFunction$Lifted.apply(PartialFunction.scala:334)
    at munit.ValueTransforms$ValueTransform.apply(ValueTransforms.scala:14)
    at munit.ValueTransforms.$anonfun$2(ValueTransforms.scala:28)
    at scala.collection.Iterator$$anon$9.next(Iterator.scala:584)
    at scala.collection.IterableOnceOps.collectFirst(IterableOnce.scala:1256)
    at scala.collection.IterableOnceOps.collectFirst$(IterableOnce.scala:1248)
    at scala.collection.AbstractIterator.collectFirst(Iterator.scala:1303)
    at munit.ValueTransforms.$anonfun$1(ValueTransforms.scala:29)
    at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:467)
    at unsafeRunSync @ fs2.interop.packageDoctest$$anon$1.$init$$$anonfun$2$$anonfun$1(packageDoctest.scala:36)
    at race$extension @ fs2.Stream.forkOnElem$1$$anonfun$1$$anonfun$1$$anonfun$1(Stream.scala:2364)
    at timeoutTo @ munit.CatsEffectSuite.munit$CatsEffectSuite$$anon$1$$_$_$$anonfun$2(CatsEffectSuite.scala:78)
```

https://github.com/typelevel/fs2/actions/runs/13727001151/job/38395670558?pr=3539#step:11:2940

cc @BalmungSan

PR #3545, opened 2025-03-10
Be more deffensive with the `StreamSubscriber` internal state

BalmungSan:
I think this fixes the flaky **flow** interop tests: #3541 #3542 
Related to #3449 

I think the main change is ensuring we reset the `StreamSubscriber` state BEFORE completing the callback with the `Chunk`.
But, I also made sure we reset the state in any place where we complete a callback, even with exceptions, just in case.

c.c. @armanbilge

### DIFF
--- a/core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
+++ b/core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
@@ -66,9 +66,9 @@ private[flow] final class StreamSubscriber[F[_], A] private (
     * since they are always done on the effect run after the state update took place.
     * Meaning this should be correct if the Producer is well-behaved.
     */
-  private var inOnNextLoop: Boolean = _
+  private var inOnNextLoop: Boolean = false
   private var buffer: Array[Any] = null
-  private var index: Int = _
+  private var index: Int = 0
 
   /** Receives the next record from the upstream reactive-streams system. */
   override final def onNext(a: A): Unit = {
@@ -165,9 +165,9 @@ private[flow] final class StreamSubscriber[F[_], A] private (
               // We do the updates here,
               // to ensure they happen after we have secured the state.
               inOnNextLoop = true
-              index = 1
               buffer = new Array(chunkSize)
               buffer(0) = a
+              index = 1
             }
           }
 
@@ -188,9 +188,10 @@ private[flow] final class StreamSubscriber[F[_], A] private (
           Idle(s) -> run {
             // We do the updates here,
             // to ensure they happen after we have secured the state.
-            cb.apply(Right(Some(Chunk.array(buffer))))
+            val chunk = Chunk.array(buffer)
             inOnNextLoop = false
             buffer = null
+            cb.apply(Right(Some(chunk)))
           }
 
         case state =>
@@ -205,20 +206,20 @@ private[flow] final class StreamSubscriber[F[_], A] private (
       case Error(ex) => {
         case Uninitialized(Some(cb)) =>
           Terminal -> run {
-            cb.apply(Left(ex))
             // We do the updates here,
             // to ensure they happen after we have secured the state.
             inOnNextLoop = false
             buffer = null
+            cb.apply(Left(ex))
           }
 
         case WaitingOnUpstream(cb, _) =>
           Terminal -> run {
-            cb.apply(Left(ex))
             // We do the updates here,
             // to ensure they happen after we have secured the state.
             inOnNextLoop = false
             buffer = null
+            cb.apply(Left(ex))
           }
 
         case _ =>
@@ -228,6 +229,10 @@ private[flow] final class StreamSubscriber[F[_], A] private (
       case Complete(canceled) => {
         case Uninitialized(Some(cb)) =>
           Terminal -> run {
+            // We do the updates here,
+            // to ensure they happen after we have secured the state.
+            inOnNextLoop = false
+            buffer = null
             cb.apply(Right(None))
           }
 
@@ -240,17 +245,22 @@ private[flow] final class StreamSubscriber[F[_], A] private (
 
         case WaitingOnUpstream(cb, s) =>
           Terminal -> run {
+            // We do the updates here,
+            // to ensure they happen after we have secured the state.
             if (canceled) {
               s.cancel()
-              cb.apply(Right(None))
-            } else if (index == 0) {
-              cb.apply(Right(None))
-            } else {
-              cb.apply(Right(Some(Chunk.array(buffer, offset = 0, length = index))))
-              // We do the updates here,
-              // to ensure they happen after we have secured the state.
               inOnNextLoop = false
               buffer = null
+              cb.apply(Right(None))
+            } else if (index == 0) {
+              inOnNextLoop = false
+              buffer = null
+              cb.apply(Right(None))
+            } else {
+              val chunk = Chunk.array(buffer, offset = 0, length = index)
+              inOnNextLoop = false
+              buffer = null
+              cb.apply(Right(Some(chunk)))
             }
           }
 
@@ -267,15 +277,19 @@ private[flow] final class StreamSubscriber[F[_], A] private (
 
         case Idle(s) =>
           WaitingOnUpstream(cb, s) -> run {
-            s.request(chunkSize.toLong)
             // We do the updates here,
             // to ensure they happen after we have secured the state.
             inOnNextLoop = false
-            index = 0
+            buffer = null
+            s.request(chunkSize.toLong)
           }
 
         case state @ Uninitialized(Some(otherCB)) =>
           Terminal -> run {
+            // We do the updates here,
+            // to ensure they happen after we have secured the state.
+            inOnNextLoop = false
+            buffer = null
             val ex = Left(new InvalidStateException(operation = "Received request", state))
             otherCB.apply(ex)
             cb.apply(ex)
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -66,9 +66,9 @@ private[flow] final class StreamSubscriber[F[_], A] private (
     * since they are always done on the effect run after the state update took place.
     * Meaning this should be correct if the Producer is well-behaved.
     */
-  private var inOnNextLoop: Boolean = _
+  private var inOnNextLoop: Boolean = false
   private var buffer: Array[Any] = null
-  private var index: Int = _
+  private var index: Int = 0
```
BalmungSan, line 71: AFAIK, these are equivalent, but just in case, I made sure `inOnNextLoop` starts as `false`, the `index` shouldn't matter but just for consistency; I also thought of using `-1` but that feels confusing.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -188,9 +188,10 @@ private[flow] final class StreamSubscriber[F[_], A] private (
           Idle(s) -> run {
             // We do the updates here,
             // to ensure they happen after we have secured the state.
-            cb.apply(Right(Some(Chunk.array(buffer))))
+            val chunk = Chunk.array(buffer)
             inOnNextLoop = false
             buffer = null
+            cb.apply(Right(Some(chunk)))
```
BalmungSan, line 194: Reset the state BEFORE publishing the `Exception`.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -205,20 +206,20 @@ private[flow] final class StreamSubscriber[F[_], A] private (
       case Error(ex) => {
         case Uninitialized(Some(cb)) =>
           Terminal -> run {
-            cb.apply(Left(ex))
             // We do the updates here,
             // to ensure they happen after we have secured the state.
             inOnNextLoop = false
             buffer = null
+            cb.apply(Left(ex))
```
BalmungSan, line 213: Reset the state BEFORE publishing the `Exception`.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -205,20 +206,20 @@ private[flow] final class StreamSubscriber[F[_], A] private (
...
           }
 
         case WaitingOnUpstream(cb, _) =>
           Terminal -> run {
-            cb.apply(Left(ex))
             // We do the updates here,
             // to ensure they happen after we have secured the state.
             inOnNextLoop = false
             buffer = null
+            cb.apply(Left(ex))
```
BalmungSan, line 222: Reset the state BEFORE publishing the `Exception`.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -228,6 +229,10 @@ private[flow] final class StreamSubscriber[F[_], A] private (
       case Complete(canceled) => {
         case Uninitialized(Some(cb)) =>
           Terminal -> run {
+            // We do the updates here,
+            // to ensure they happen after we have secured the state.
+            inOnNextLoop = false
+            buffer = null
```
BalmungSan, line 235: In theory, not needed anymore since we are cancelled.
But, at minimum, this could lead to the `buffer` data to be GC early.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -240,17 +245,22 @@ private[flow] final class StreamSubscriber[F[_], A] private (
...
+              buffer = null
               cb.apply(Right(None))
             } else {
-              cb.apply(Right(Some(Chunk.array(buffer, offset = 0, length = index))))
-              // We do the updates here,
-              // to ensure they happen after we have secured the state.
+              val chunk = Chunk.array(buffer, offset = 0, length = index)
               inOnNextLoop = false
               buffer = null
+              cb.apply(Right(Some(chunk)))
```
BalmungSan, line 263: Make sure we reset the state BEFORE anything else.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -267,39 +277,51 @@ private[flow] final class StreamSubscriber[F[_], A] private (
 
         case Idle(s) =>
           WaitingOnUpstream(cb, s) -> run {
-            s.request(chunkSize.toLong)
             // We do the updates here,
             // to ensure they happen after we have secured the state.
             inOnNextLoop = false
-            index = 0
+            buffer = null
```
BalmungSan, line 283: In theory redundant, but just in case, we make sure we start in the initial state.
The `index` however felt wrong setting it to `0` here since it doesn't matter, the `index` is only used when we are in the `onNext` loop. Thus, I made sure that the use of that value only happens in interactions directly related to that loop.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -267,39 +277,51 @@ private[flow] final class StreamSubscriber[F[_], A] private (
...
+            s.request(chunkSize.toLong)
           }
 
         case state @ Uninitialized(Some(otherCB)) =>
           Terminal -> run {
+            // We do the updates here,
+            // to ensure they happen after we have secured the state.
+            inOnNextLoop = false
+            buffer = null
```
BalmungSan, line 292: Reset the state BEFORE publishing the `Exception`.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -267,39 +277,51 @@ private[flow] final class StreamSubscriber[F[_], A] private (
...
-            otherCB.apply(ex)
-            cb.apply(ex)
             // We do the updates here,
             // to ensure they happen after we have secured the state.
             inOnNextLoop = false
             buffer = null
+            s.cancel()
+            val ex = Left(new InvalidStateException(operation = "Received request", state))
+            otherCB.apply(ex)
+            cb.apply(ex)
```
BalmungSan, line 307: Reset the state BEFORE publishing the `Exception`.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -267,39 +277,51 @@ private[flow] final class StreamSubscriber[F[_], A] private (
...
+            buffer = null
             cb.apply(Left(ex))
           }
 
         case Terminal =>
           Terminal -> run {
+            // We do the updates here,
+            // to ensure they happen after we have secured the state.
+            inOnNextLoop = false
+            buffer = null
```
BalmungSan, line 324: In theory fully redundant, but just in case.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -206,19 +209,15 @@ private[flow] final class StreamSubscriber[F[_], A] private (
         case Uninitialized(Some(cb)) =>
           Terminal -> run {
             cb.apply(Left(ex))
-            // We do the updates here,
-            // to ensure they happen after we have secured the state.
-            inOnNextLoop = false
-            buffer = null
```
BalmungSan, line 212: Redundant.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -240,17 +239,21 @@ private[flow] final class StreamSubscriber[F[_], A] private (
...
+              inOnNextLoop = false
               cb.apply(Right(None))
             } else {
-              cb.apply(Right(Some(Chunk.array(buffer, offset = 0, length = index))))
-              // We do the updates here,
-              // to ensure they happen after we have secured the state.
+              val chunk = Chunk.array(buffer, offset = 0, length = index)
               inOnNextLoop = false
               buffer = null
+              cb.apply(Right(Some(chunk)))
```
BalmungSan, line 256: Reset the state BEFORE publishing the `Chunk` _(or lack of)_.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -240,17 +239,21 @@ private[flow] final class StreamSubscriber[F[_], A] private (
...
           Terminal -> run {
+            // We do the updates here,
+            // to ensure they happen after we have secured the state.
             if (canceled) {
               s.cancel()
+              inOnNextLoop = false
+              buffer = null
               cb.apply(Right(None))
-            } else if (index == 0) {
+            } else if (buffer eq null) {
```
BalmungSan, line 249: Rather than relying on the `index`, we simply check if the `buffer` hasn't been initialized.

core/shared/src/main/scala/fs2/interop/flow/StreamSubscriber.scala
```diff
@@ -268,10 +271,6 @@ private[flow] final class StreamSubscriber[F[_], A] private (
         case Idle(s) =>
           WaitingOnUpstream(cb, s) -> run {
             s.request(chunkSize.toLong)
-            // We do the updates here,
-            // to ensure they happen after we have secured the state.
-            inOnNextLoop = false
-            index = 0
```
BalmungSan, line 274: Redundant.
The initial state is this, and finishing each loop leaves the state like this as well.

PR #3546, opened 2025-03-11
Implement `{Signal,SignallingRef}#mapK`

rahulrangers:
Previously, calling .mapK on a SignallingRef[F, A] would yield a plain Ref[G, A], causing the extra signaling behavior (such as discrete, continuous) to be lost. This PR introduces a new helper class, TransformedSignallingRef, which implements mapK for SignallingRef so that all signal-specific operations are lifted to the new effect type. This change ensures that after a transformation, the resulting reference remains a full SignallingRef[G, A] with its intended functionality.


rahulrangers:
@armanbilge Could you please review this and let me know if there are any mistakes that need correction?

core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -341,7 +349,30 @@ object SignallingRef {
...
+    override def modify[B](f: A => (A, B)): G[B] = trans(underlying.modify(f))
+    override def tryModifyState[B](state: State[A, B]): G[Option[B]] =
+      trans(underlying.tryModifyState(state))
+    override def modifyState[B](state: State[A, B]): G[B] = trans(underlying.modifyState(state))
+    override def access: G[(A, A => G[Boolean])] =
+      G.compose[(A, *)].compose[A => *].map(trans(underlying.access))(trans(_))
+
+    // --- Signal-specific methods
+    override def discrete: Stream[G, A] = underlying.discrete.translate(trans)
+    override def continuous: Stream[G, A] = underlying.continuous.translate(trans)
```
armanbilge, line 374: Can you also add an override for `changes`? I also wish we could override `getAndDiscreteUpdates` and `waitUntil`, but I don't think it's possible 😕

rahulrangers: added changes

core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -341,7 +349,38 @@ object SignallingRef {
...
+    override def changes(implicit eqA: Eq[A]): Signal[G, A] =
+      new Signal[G, A] {
+        def discrete = TransformedSignallingRef.this.discrete.changes
+        def continuous = TransformedSignallingRef.this.continuous
+        def get = TransformedSignallingRef.this.get
+      }
+    override def waitUntil(p: A => Boolean)(implicit G: Concurrent[G]): G[Unit] =
+      TransformedSignallingRef.this.discrete.forall(a => !p(a)).compile.drain
```
armanbilge, line 382: Thanks for adding these. Unfortunately, this is just duplicating the default implementations. We already inherit these default implementations, so we don't need to duplicate it.

Instead, we should override these with implementations that delegate to `underlying`. The reason we want to do this, is because `underlying` may have overridden the default implementation with an optimized implementation. We want to prefer to use the optimized implementation whenever possible.

rahulrangers:
I'm encountering an issue while implementing the `waitUntil` method in the TransformedSignallingRef class. Specifically, the compiler raises an error indicating that it cannot find an implicit value for the parameter F: cats.effect.kernel.Concurrent[F].

core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -341,7 +349,36 @@ object SignallingRef {
...
+
+    // --- Signal-specific methods
+    override def discrete: Stream[G, A] = underlying.discrete.translate(trans)
+    override def continuous: Stream[G, A] = underlying.continuous.translate(trans)
+    override def changes(implicit eqA: Eq[A]): Signal[G, A] =
+      new Signal[G, A] {
+        def discrete = underlying.changes.discrete.translate(trans)
+        def continuous = underlying.changes.continuous.translate(trans)
+        def get: G[A] = trans(underlying.changes.get)
+      }
```
armanbilge, line 380: This is re-implementing `mapK` 😁

rahulrangers: Oh! I see this re implementing mapK for the most of the part but  `changes` primarily changes `discrete` and `getAndDiscreteUpdates` functions and in the above i am taking the `discrete` function from changes so it is not totally re-implementing mapK, can you guide on how to override  `changes` function correctly @armanbilge 😅.

armanbilge: Just use `mapK`!

rahulrangers: Yes I thought of using mapk in the first place but mapK is implemented  for abstract class `SignallingRef` not for trait `Signal`

rahulrangers: I have implemented `mapK` for `Signal` and used to override `changes` . Can you please review @armanbilge

core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -162,6 +170,15 @@ object Signal extends SignalInstances {
...
     }
 
+  final private class TransformedSignal[F[_], G[_], A](
+      underlying: Signal[F, A],
+      trans: FunctionK[F, G]
+  ) extends Signal[G, A] {
+    override def get: G[A] = trans(underlying.get)
+    override def discrete: Stream[G, A] = underlying.discrete.translate(trans)
+    override def continuous: Stream[G, A] = underlying.continuous.translate(trans)
+  }
```
armanbilge, line 180: Thanks for adding this! We should override `changes` too.

rahulrangers: added changes function

core/shared/src/main/scala/fs2/concurrent/Signal.scala
```diff
@@ -196,7 +214,12 @@ object Signal extends SignalInstances {
   * function, in the presence of `discrete`, can return `false` and
   * need looping even without any other writers.
   */
-abstract class SignallingRef[F[_], A] extends Ref[F, A] with Signal[F, A]
+abstract class SignallingRef[F[_], A] extends Ref[F, A] with Signal[F, A] {
+  def mapK[G[_]](
+      f: FunctionK[F, G]
+  )(implicit F: Concurrent[F], G: Functor[G], dummy: DummyImplicit): SignallingRef[G, A] =
```
armanbilge, line 220: I would rather that we don't collect any constraints for `F`, see https://github.com/typelevel/cats-effect/pull/4115#issuecomment-2414839583 for more info.

I know this means we won't be able to override `waitUntil`, but we'll just have to live with that.

rahulrangers: removed Constraints from F

armanbilge (commented):
Great, thanks, this looks good!

One final thing, can we add a test? Mostly I'm interested in a test to check that `SignallingRef#mapK` returns a `SignallingRef` instead of a `Signal` ... because we have two methods with the same name, I'm not entirely sure how they get resolved.

rahulrangers:
>I'm note entirely sure how they get resolved.

I think it is resolved as SignallingRef's mapK has an extra implicit parameter making it different from Signal’s mapK.

core/shared/src/test/scala/fs2/concurrent/SignalSuite.scala
```diff
@@ -320,6 +321,19 @@ class SignalSuite extends Fs2Suite {
     TestControl.executeEmbed(prog).assertEquals(expected)
   }
 
+  test("SignallingRef.mapK() returns a SignallingRef") {
```
armanbilge, line 324:
```suggestion
  test("SignallingRef#mapK returns a SignallingRef") {
```

core/shared/src/test/scala/fs2/concurrent/SignalSuite.scala
```diff
@@ -320,6 +321,19 @@ class SignalSuite extends Fs2Suite {
...
+    for {
+      s <- SignallingRef[IO, Int](0)
+      nt = new FunctionK[IO, IO] {
+        def apply[A](fa: IO[A]): IO[A] = fa
+      }
+      transformed = s.mapK(nt)
```
armanbilge, line 330: Let's also check that the type infers correctly at compile-time.

```suggestion
      transformed: SignallingRef[IO, Int] = s.mapK(nt)
```

armanbilge (commented):
It does seem to be working!

> I think it is resolved as SignallingRef's mapK has an extra implicit parameter making it different from Signal’s mapK.

That's right, they are different. I am just not sure how the compiler decides which method to choose when it's ambiguous.

PR #3547, opened 2025-03-12
fix `flake.nix`

onsah:
Fix `flake.nix`. It wasn't working before.

After changes, I can open a nix shell as expected.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,7 +18,7 @@
           imports = [ typelevel-nix.typelevelShell ];
           name = "fs2-shell";
           typelevelShell = {
-            jdk.package = pkgs.jdk23;
+            jdk.package = pkgs.jdk17;
             nodejs.enable = true;
             native.enable = true;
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -10,15 +10,15 @@
...
+          overlays = [ typelevel-nix.overlays.default ];
         };
       in
       {
         devShell = pkgs.devshell.mkShell {
           imports = [ typelevel-nix.typelevelShell ];
           name = "fs2-shell";
           typelevelShell = {
-            jdk.package = pkgs.jdk17;
+            jdk.package = pkgs.jdk23;
```
armanbilge, line 21: Is the JDK bump necessary? We use JDK 17 in CI.

armanbilge:
Since that's not specific to FS2 (most projects use scalafmt) you can propose to add it to the upstream typelevel-nix flake.

https://github.com/typelevel/typelevel-nix

PR #3548, opened 2025-03-12
Use virtual threads when possible for JDK `UnixSocket`s

onsah:
Part of #3182


flake.nix
```diff
@@ -10,7 +10,7 @@
       let
         pkgs = import nixpkgs {
           inherit system;
-          overlays = [ typelevel-nix.overlay ];
+          overlays = [ typelevel-nix.overlays.default ];
```
onsah, line 13: This change comes from https://github.com/typelevel/fs2/pull/3547. I will rebase when it get's merged.

io/jvm/src/main/scala/fs2/io/net/unixsocket/JdkUnixSockets.scala
```diff
@@ -42,27 +44,28 @@ private[unixsocket] class JdkUnixSocketsImpl[F[_]: Files](implicit F: Async[F])
     extends UnixSockets.AsyncUnixSockets[F] {
   protected def openChannel(address: UnixSocketAddress) =
     Resource
-      .make(F.blocking(SocketChannel.open(StandardProtocolFamily.UNIX)))(ch =>
-        F.blocking(ch.close())
+      .make(F.blocking(SocketChannel.open(StandardProtocolFamily.UNIX)).pipe(evalOnVirtualThreadIfAvailable(_)))(ch =>
```
onsah, line 47: I think it is more readable when using `pipe`. Using regular function application it obscures the actual code.

io/jvm/src/main/scala/fs2/io/net/unixsocket/JdkUnixSockets.scala
```diff
@@ -42,27 +43,32 @@ private[unixsocket] class JdkUnixSocketsImpl[F[_]: Files](implicit F: Async[F])
     extends UnixSockets.AsyncUnixSockets[F] {
   protected def openChannel(address: UnixSocketAddress) =
     Resource
-      .make(F.blocking(SocketChannel.open(StandardProtocolFamily.UNIX)))(ch =>
-        F.blocking(ch.close())
-      )
+      .make(
+        F.blocking(SocketChannel.open(StandardProtocolFamily.UNIX)).evalOnVirtualThreadIfAvailable()
```
onsah, line 47: Using `evalOnVirtualThreadIfAvailable` as a member function is more readable in my opinion. It doesn't get in the way on the actual code.

onsah: Fair enough

io/jvm/src/main/scala/fs2/io/net/unixsocket/JdkUnixSockets.scala
```diff
@@ -41,28 +42,32 @@ object JdkUnixSockets {
...
-      )
-      .evalTap { ch =>
-        F.blocking(ch.connect(UnixDomainSocketAddress.of(address.path)))
-          .cancelable(F.blocking(ch.close()))
-      }
+    evalOnVirtualThreadIfAvailable(
+      Resource
+        .make(
+          F.blocking(SocketChannel.open(StandardProtocolFamily.UNIX))
+        )(ch => evalOnVirtualThreadIfAvailable(F.blocking(ch.close())))
```
armanbilge, line 49: Because we are wrapping the entire `Resource` in `evalOnVirtualThreadIfAvailable`, we don't need to wrap the individual effects.

onsah: Right, I think I missed it out when I was changing from individual wrappings to wrapping the whole `Resource`.

armanbilge:
We should also use virtual threads for reading/writing in the socket implementaiton.

https://github.com/typelevel/fs2/blob/c381fd84a1ffa61ca6e1bd28c76488101c8c430a/io/jvm/src/main/scala/fs2/io/net/unixsocket/UnixSocketsPlatform.scala#L96

io/jvm/src/main/scala/fs2/io/net/unixsocket/UnixSocketsPlatform.scala
```diff
@@ -101,12 +102,15 @@ private[unixsocket] trait UnixSocketsCompanionPlatform {
...
       def go(buff: ByteBuffer): F[Unit] =
-        F.blocking(ch.write(buff)).cancelable(close) *>
-          F.delay(buff.remaining <= 0).ifM(F.unit, go(buff))
+        evalOnVirtualThreadIfAvailable(
+          F.blocking(ch.write(buff)).cancelable(close) *>
+            F.delay(buff.remaining <= 0).ifM(F.unit, go(buff))
+        )
 
       writeMutex.lock.surround {
         F.delay(bytes.toByteBuffer).flatMap(go)
```
armanbilge, line 116: Instead of doing the `eval` for every iteration, can just apply it to the first invocation of `go`.

armanbilge (approved):

Issue #3171, opened 2023-03-22
Support reading from `ReadCursor` / writing to `WriteCursor` via `Socket`


armanbilge:
There are some really efficient APIs for this:
- https://man7.org/linux/man-pages/man2/sendfile.2.html
- https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileChannel.html#transferTo-long-long-java.nio.channels.WritableByteChannel-
- https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileChannel.html#transferFrom-java.nio.channels.ReadableByteChannel-long-long-
- https://std.in/using-sendfile-with-nodejs/

I'm fairly confident we can add a default implementation to `Socket` just using the `Stream` APIs. If so, then this opens the door for more specialized overrides on the actual implementation.

rahulrangers:
@armanbilge If this issue is still relevant ,can you assign this issue to me ?

rahulrangers:
I am thinking that in this issue
1) we need to create two files SocketReadCursor and SocketWriteCursor which is used to read from the socket and write to socket just like the cursors in file handling.
2) use these to override the implementation in platforms-specific files using sendfile syscall where we send the data directly from the file in disk to kernel which increases efficiency.
But I am not understanding the idea of read and writecursors through sockets as these are not persistence or static as files , so idea of sequentially reading using offset seems not right to me? 
Am i going in right direction <@845013677816021083>

PR #3551, opened 2025-03-18
Implement `Socket#sendFile`

rahulrangers:
Added efficient transfer of a file to the network for jvm platform using `transferTo` and `transferFrom` methods of filechannel.
part of #3171

### DIFF
--- a/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
@@ -67,7 +67,7 @@ private[file] trait FileHandlePlatform[F[_]] {
 
 }
 
-class SyncFileHandle[F[_]](val chan: FileChannel)(implicit F: Sync[F]) extends FileHandle[F] {
+private[io] final class SyncFileHandle[F[_]](val chan: FileChannel)(implicit F: Sync[F]) extends FileHandle[F] {
   type Lock = FileLock
 
   override def force(metaData: Boolean): F[Unit] =
--- a/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
@@ -28,6 +28,8 @@ import java.nio.channels.{FileChannel, FileLock, SeekableByteChannel}
 import java.nio.file.{OpenOption, Path => JPath}
 
 import cats.effect.kernel.{Async, Resource, Sync}
+import java.nio.channels.WritableByteChannel
+import java.nio.channels.ReadableByteChannel
 
 private[file] trait FileHandlePlatform[F[_]] {
 
@@ -64,6 +66,10 @@ private[file] trait FileHandlePlatform[F[_]] {
     * @param lock the lock object which represents the locked file or region.
     */
   def unlock(lock: Lock): F[Unit]
+
+  def transferTo(position: Long, count: Long, target: WritableByteChannel): F[Long]
+
+  def transferFrom(src: ReadableByteChannel, position: Long, count: Long): F[Long]
 
 }
 
@@ -118,6 +124,12 @@ private[file] trait FileHandleCompanionPlatform {
 
       override def write(bytes: Chunk[Byte], offset: Long): F[Int] =
         F.blocking(chan.write(bytes.toByteBuffer, offset))
+
+      override def transferTo(position: Long, count: Long, target: WritableByteChannel): F[Long] =
+        F.blocking(chan.transferTo(position, count, target))
+
+      override def transferFrom(src: ReadableByteChannel, position: Long, count: Long): F[Long] =
+        F.blocking(chan.transferFrom(src, position, count))
     }
 
   /** Creates a `FileHandle[F]` from a `java.nio.channels.SeekableByteChannel`. Because a `SeekableByteChannel` doesn't provide all the functionalities required by `FileHandle` some features like locking will be unavailable. */
@@ -173,5 +185,10 @@ private[file] trait FileHandleCompanionPlatform {
             chan.write(bytes.toByteBuffer)
           }
         }
+      override def transferTo(position: Long, count: Long, target: WritableByteChannel): F[Long] =
+        F.raiseError(unsupportedOperationException)
+
+      override def transferFrom(src: ReadableByteChannel, position: Long, count: Long): F[Long] =
+        F.raiseError(unsupportedOperationException)
     }
 }
--- a/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
@@ -64,7 +64,7 @@ private[file] trait FileHandlePlatform[F[_]] {
     * @param lock the lock object which represents the locked file or region.
     */
   def unlock(lock: Lock): F[Unit]
-  
+
 }
 
 class SyncFileHandle[F[_]](val chan: FileChannel)(implicit F: Sync[F]) extends FileHandle[F] {
--- a/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
@@ -118,7 +118,6 @@ private[file] trait FileHandleCompanionPlatform {
 
       override def write(bytes: Chunk[Byte], offset: Long): F[Int] =
         F.blocking(chan.write(bytes.toByteBuffer, offset))
-
     }
 
   /** Creates a `FileHandle[F]` from a `java.nio.channels.SeekableByteChannel`. Because a `SeekableByteChannel` doesn't provide all the functionalities required by `FileHandle` some features like locking will be unavailable. */
--- a/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
@@ -28,6 +28,8 @@ import java.nio.channels.{FileChannel, FileLock, SeekableByteChannel}
 import java.nio.file.{OpenOption, Path => JPath}
 
 import cats.effect.kernel.{Async, Resource, Sync}
+import java.nio.channels.WritableByteChannel
+import java.nio.channels.ReadableByteChannel
 
 private[file] trait FileHandlePlatform[F[_]] {
 
@@ -64,6 +66,10 @@ private[file] trait FileHandlePlatform[F[_]] {
     * @param lock the lock object which represents the locked file or region.
     */
   def unlock(lock: Lock): F[Unit]
+
+  def transferTo(position: Long, count: Long, target: WritableByteChannel): F[Long]
+
+  def transferFrom(src: ReadableByteChannel, position: Long, count: Long): F[Long]
 
 }
 
@@ -118,6 +124,13 @@ private[file] trait FileHandleCompanionPlatform {
 
       override def write(bytes: Chunk[Byte], offset: Long): F[Int] =
         F.blocking(chan.write(bytes.toByteBuffer, offset))
+
+      override def transferTo(position: Long, count: Long, target: WritableByteChannel): F[Long] =
+        F.blocking(chan.transferTo(position, count, target))
+
+      override def transferFrom(src: ReadableByteChannel, position: Long, count: Long): F[Long] =
+        F.blocking(chan.transferFrom(src, position, count))
+    
     }
 
   /** Creates a `FileHandle[F]` from a `java.nio.channels.SeekableByteChannel`. Because a `SeekableByteChannel` doesn't provide all the functionalities required by `FileHandle` some features like locking will be unavailable. */
@@ -173,5 +186,10 @@ private[file] trait FileHandleCompanionPlatform {
             chan.write(bytes.toByteBuffer)
           }
         }
+        override def transferTo(position: Long, count: Long, target: WritableByteChannel): F[Long] =
+        F.raiseError(unsupportedOperationException)
+
+        override def transferFrom(src: ReadableByteChannel, position: Long, count: Long): F[Long] =
+        F.raiseError(unsupportedOperationException)
     }
 }
--- a/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
+++ b/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
@@ -28,7 +28,6 @@ import java.nio.channels.{FileChannel, FileLock, SeekableByteChannel}
 import java.nio.file.{OpenOption, Path => JPath}
 
 import cats.effect.kernel.{Async, Resource, Sync}
-import java.nio.channels.WritableByteChannel
 import java.nio.channels.ReadableByteChannel
 
 private[file] trait FileHandlePlatform[F[_]] {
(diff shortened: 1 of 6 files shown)


rahulrangers:
@armanbilge the tests are failing as i have not implemented it in other platforms , I kindly request you to tell me whether i am going in the right direction. If Yes, I would try to implement it in other platforms.

armanbilge:
Thanks for starting this PR! I think we should expose the `sendfile` method on the shared `Socket` interface for JVM/JS/Native and take a `FileHandle` as an argument. I am hoping that we can provide a default implementation just using `Stream` APIs, this is essential for adding this API compatibly.

Then, we can do specialized implementations for each platform that use the optimized routines.

rahulrangers:
@armanbilge implemented `sendfile`  on the shared Socket interface ,can you please review it.
 and I've added two new methods (`transferTo` and `transferFrom`) to the FileHandlePlatform trait, but I'm getting binary compatibility errors. The MiMa tool suggests adding filters to ignore these issues. Is it appropriate to use these filters in this case, or should I handle this differently?

io/shared/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -69,6 +71,29 @@ trait Socket[F[_]] {
...
+
+  /** Reads a file in chunks and writes it to a socket.
+    * Streams the file contents in chunks of the specified size and sends them over the socket.
+    * The stream terminates when the entire file has been read and written.
+    *
+    * @param file the file handle to read from
+    * @param chunkSize the size of each chunk to read
+    * @param F an implicit `Async` instance for effectful computations
+    */
+  def sendfile(file: FileHandle[F], chunkSize: Int)(implicit F: Async[F]): F[Unit] =
```
armanbilge, line 83: Thanks, this is looking much better!

1. Let's also take `offset` and `count` parameters.
2. I think it should be possible to implement this without `Async`
3. Let's also add a method for the other direction, socket -> file.

rahulrangers: @armanbilge file.read and socket.read are returning F[Option[Chunk[Byte]]] but, without Async[F] I was not having access to .map, which is necessary for sequence and transform effectful computations.

armanbilge: > having access to .map, which is necessary for sequence and transform effectful computations.

If you use `Stream.eval` you can lift an `F` into `Stream`, and then you have access to `map`. I think with some clever rewriting, it is possible to avoid calling methods directly on `F`.

rahulrangers: @armanbilge used `stream.eval` for mapping and created go function to perform recursion instead `unfoldeval` as using `stream.eval` inside the `unfoldeval` returning nested stream which is causing issue

io/shared/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -69,6 +71,69 @@ trait Socket[F[_]] {
...
+    * @param count the maximum number of bytes to transfer
+    * @param chunkSize the size of each chunk to read
+    * @param F an implicit `Concurrent` instance for effectful computations
+    */
+  def sendfile(
+      file: FileHandle[F],
+      offset: Long,
+      count: Long,
+      chunkSize: Int
+  )(implicit F: Concurrent[F]): F[Unit] = {
```
armanbilge, line 90: Then we don't need a constraint at all :)
```suggestion
  ): Stream[F, Nothing] = {
```

io/shared/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -69,6 +71,69 @@ trait Socket[F[_]] {
...
+
+    def go(currOffset: Long, remaining: Long): Stream[F, Byte] =
+      if (remaining > 0)
+        Stream.eval(file.read(chunkSize.min(remaining.toInt), currOffset)).flatMap {
```
armanbilge, line 94: Are you sure that `remaining.toInt` is safe? What if `remaining` is bigger than an `Int`?

rahulrangers: Oh! sorry overlooked it 😅 changed now

io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
```diff
@@ -118,6 +118,7 @@ private[file] trait FileHandleCompanionPlatform {
 
       override def write(bytes: Chunk[Byte], offset: Long): F[Int] =
         F.blocking(chan.write(bytes.toByteBuffer, offset))
+
```
armanbilge, line 121: Spurious change

io/shared/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -69,6 +70,65 @@ trait Socket[F[_]] {
...
+    * @param offset the starting position in the file
+    * @param count the maximum number of bytes to transfer
+    * @param chunkSize the size of each chunk to read
+    */
+  def sendfile(
+      file: FileHandle[F],
+      offset: Long,
+      count: Long,
+      chunkSize: Int
+  ): Stream[F, Unit] = {
```
armanbilge, line 88: Great, these APIs are looking good! The next step is to add platform-specific overrides that can use more efficient implementations.

armanbilge (commented):
Thanks for your work on this! You can also start adding some tests.

io/shared/src/main/scala/fs2/io/net/Socket.scala
rahulrangers: I will make separate PR's for it

armanbilge: Actually, do you mind doing it in this PR? Before we add this API, I want to make sure we can actually implement the optimized versions.

armanbilge, line 88: `Stream[F, Unit]` is a stream of 0 to infinite `()`s. `Stream[F, Nothing]` is definitively an empty stream.

```suggestion
  ): Stream[F, Nothing] = {
```

rahulrangers:
@armanbilge I have tried to override `sendfile` and `recvfile` in different platforms and got some issues.

1) For JVM I got binary compatibility issue due to the addition of `transferTo` and `transferFrom` functions.
2) For js I think it is still not stable to override and the `sendfile` (in nodejs)  needs socket's file descriptor but the net.Socket does not provide file descriptor .   https://github.com/nodejs/node/issues/8297
3) For now in JVM I was able to override these functions in `unixsocketsplatform` but not in `socketplatform` as `transferto` and '`transferfrom` function need a `WriteByteChannel` like `SocketChannel` but `socketplatform` is having `AsynchronousSocketChannel`

armanbilge:
1. Let's not make changes to the `FileHandle` API.
2. I think you should be able to add an override to `SelectingSocket`.
3. I agree, I don't think we should override for JS.

rahulrangers:
> 1. Let's not make changes to the `FileHandle` API.
>
@armanbilge I think to include `transferto` and `transferfrom` methods, we need to change the `FileHandle` API as it has `FileChannel` through which we can add the above methods. Can please give me any other idea to get these methods without changing 'FileHandle' API.

armanbilge:
> Can you please give me any other idea to get these methods

We can make this into a named class `SyncFileHandle` (instead of anonymous), then we can pattern-match in `SelectingSocket` to get access to the underlying `FileChannel`. Let me know if that makes sense!

https://github.com/typelevel/fs2/blob/3ff885f2c96a9762696ca731541ab429394c6031/io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala#L86-L89

rahulrangers:
@armanbilge While trying to do pattern matching I realized that I cannot import `SyncFilehandle` as it is in `jvm-native` but we are doing the pattern-matching in `JVM`

armanbilge:
> I cannot import `SyncFilehandle` as it is implemented in `jvm-native` but we are doing the pattern-matching in `JVM`

Anything in `jvm-native` is available to both JVM and Native. Are you getting a compiler error? Can you share it?

io/jvm-native/src/main/scala/fs2/io/file/FileHandlePlatform.scala
```diff
@@ -67,58 +67,60 @@ private[file] trait FileHandlePlatform[F[_]] {
 
 }
 
-private[file] trait FileHandleCompanionPlatform {
-  @deprecated("Use Files[F].open", "3.0.0")
-  def fromPath[F[_]: Async](path: JPath, flags: Seq[OpenOption]): Resource[F, FileHandle[F]] =
-    Files.forAsync[F].open(path, flags)
+class SyncFileHandle[F[_]](val chan: FileChannel)(implicit F: Sync[F]) extends FileHandle[F] {
```
armanbilge, line 70:
```suggestion
private[io] final class SyncFileHandle[F[_]](val chan: FileChannel)(implicit F: Sync[F]) extends FileHandle[F] {
```

io/jvm/src/main/scala/fs2/io/net/SelectingSocket.scala
```diff
@@ -77,7 +79,69 @@ private final class SelectingSocket[F[_]: LiftIO] private (
...
+  override def sendfile(
+      file: FileHandle[F],
+      offset: Long,
+      count: Long,
+      chunkSize: Int
+  ): Stream[F, Nothing] = file match {
+    case syncFileHandle: SyncFileHandle[F] =>
+      val fileChannel = syncFileHandle.chan
 
+      def go(currOffset: Long, remaining: Long): Stream[F, Unit] =
```
armanbilge, line 91: This is looking good! Couple suggestions:

1. We can acquire the `writeMutex` once, before calling `go` (see for example `write` method above). This will be more efficient, and will also guarantee that no one else can write until we are done sending the file.

2. It's okay to write this method in terms of `F` and convert to `Stream` at the end. Since we are inside of `SelectingSocket`, we now have access to `Async[F]`. This wasn't the case for the `Socket` API.

rahulrangers: I have made the change to acquire `writeMutex` once before calling `go` and similarly done with `readMutex` in recvfile.

io/jvm/src/main/scala/fs2/io/net/SelectingSocket.scala
```diff
@@ -77,7 +79,61 @@ private final class SelectingSocket[F[_]: LiftIO] private (
...
+      chunkSize: Int
+  ): Stream[F, Nothing] = file match {
+    case syncFileHandle: SyncFileHandle[F] =>
+      val fileChannel = syncFileHandle.chan
 
+      def go(currOffset: Long, remaining: Long): F[Unit] =
+        if (remaining <= 0) F.unit
+        else {
+          val toTransfer = remaining.min(chunkSize.toLong)
+          F.blocking(fileChannel.transferTo(currOffset, toTransfer, ch)).flatMap { written =>
```
armanbilge, line 95: We don't need to use `chunkSize` here, we should leave it to the underlying syscall to chunk/buffer things appropriately.

```suggestion
          F.blocking(fileChannel.transferTo(currOffset, remaining, ch)).flatMap { written =>
```

io/shared/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -69,6 +70,65 @@ trait Socket[F[_]] {
   /** Writes the supplied stream of bytes to this socket via `write` semantics.
     */
   def writes: Pipe[F, Byte, Nothing]
+
+  /** Reads a file in chunks and writes it to a socket.
+    * Streams the file contents in chunks of the specified size and sends them over the socket.
+    * The stream terminates when the entire file has reached end of file or the specified count is reached.
```
armanbilge, line 76: Same here, we don't need to mention the chunks, they are just an implementation detail of the default implementation. The point of the API is to transfer the file contents to the socket. The mechanism may or may not rely on chunking.

rahulrangers: Removed chunks from the comment

io/jvm/src/main/scala/fs2/io/net/SelectingSocket.scala
rahulrangers: @armanbilge made the required change.

io/jvm/src/main/scala/fs2/io/net/SelectingSocket.scala
```diff
@@ -77,7 +79,59 @@ private final class SelectingSocket[F[_]: LiftIO] private (
...
+      count: Long,
+      chunkSize: Int
+  ): Stream[F, Nothing] = file match {
+    case syncFileHandle: SyncFileHandle[F] =>
+      val fileChannel = syncFileHandle.chan
+
+      def go(currOffset: Long, remaining: Long): F[Unit] =
+        if (remaining <= 0) F.unit
+        else {
+          F.blocking(fileChannel.transferFrom(ch, currOffset, remaining)).flatMap { readBytes =>
```
armanbilge, line 121: I have been trying to figure out how the JDK implements `transferFrom`.

I thought that it uses `sendfile` for both `transferTo` and `transferFrom`, but I've been reading the [`sendfile` manpage](https://man7.org/linux/man-pages/man2/sendfile.2.html) and I think I might be mistaken.

Specifically, it says:
> The in_fd argument must correspond to a file which supports `mmap`-like operations (i.e., it cannot be a socket).

So it sounds like `sendfile` doesn't work for this case. I wonder how it is working.

Unfortunately, if there isn't an optimized implementation we can use for this method, then in my opinion there isn't much point to add it to the API.

armanbilge: Thanks for investigating!

> But I think `recvfile` should not have any problem as it is using `transferto` , which used `sendfile` syscall in its implementation

Hmm, did you mean the other way around? The code above is the implementation of `recvfile`, which is using `transferFrom`, not `transferTo`.

rahulrangers: @armanbilge oh sorry! I mean `sendfile` shall I keep it's implementation and remove remaining?

armanbilge: Yes, let's keep `sendfile` (you can rename to `sendFile` to better match our naming convention). Let's remove `recvfile` for now. Thanks!

rahulrangers:
@armanbilge  removed `recvfile` implementation please review my latest commit .

io/jvm/src/main/scala/fs2/io/net/SelectingSocket.scala
```diff
@@ -77,7 +79,32 @@ private final class SelectingSocket[F[_]: LiftIO] private (
...
+          F.blocking(fileChannel.transferTo(currOffset, remaining, ch)).flatMap { written =>
+            if (written > 0) {
+              go(currOffset + written, remaining - written)
+            } else {
+              selector.select(ch, OP_WRITE).to *> go(currOffset, remaining)
+            }
+          }
+        }
+
+      Stream.eval(writeMutex.lock.surround(go(offset, count))).drain
```
armanbilge, line 103:
```suggestion
      Stream.exec(writeMutex.lock.surround(go(offset, count)))
```

io/jvm/src/main/scala/fs2/io/net/unixsocket/UnixSocketsPlatform.scala
```diff
@@ -134,5 +135,30 @@ private[unixsocket] trait UnixSocketsCompanionPlatform {
...
+          else {
+            F.blocking(fileChannel.transferTo(currOffset, remaining, ch)).flatMap { written =>
+              if (written == 0) F.unit
+              else {
+                go(currOffset + written, remaining - written)
+              }
+            }
+          }
+
+        Stream.eval(writeMutex.lock.surround(go(offset, count))).drain
```
armanbilge, line 158:
```suggestion
        Stream.exec(writeMutex.lock.surround(go(offset, count)))
```

armanbilge (approved):
nice work!

PR #3563, opened 2025-04-08
Unify TCP/UDP and UnixSocket APIs

### LINKED ISSUE #3560: Improve unix sockets support
While adding support for [connecting to Postgres via unix sockets to Skunk](https://github.com/typelevel/skunk/pull/1192), I've come up with a few improvements we should make to fs2:

- Add support for socket options to unix sockets
- Support reading socket options from socket
- Unify client connection API

### Add support for socket options to unix sockets

Unix sockets support socket options but the `client` and `server` methods on `UnixSockets` don't currently accept an options list. Both methods should be updated to take a new param `options: List[SocketOption] = Nil` like the equivalent operations in `SocketGroup`.

### Support reading socket options from socket

There's currently no way to programmatically get the value of a specific socket option. This is a significant gap for unix sockets where the `SO_PEERCRED` option is set by the kernel and may be queried by the application. We should add a `getOption` method to `Socket` and implement for all socket types and platforms.

### Unify client connection API

Currently, if an application wants to support connecting via both TCP and unix sockets, the app has to call `SocketGroup#client` for TCP, supplying a host and port, and `UnixSockets#client` for unix domain sockets, supplying a path. The former is typically accessed by the `Network[F]` capability while the latter via `UnixSockets[F]` capability.

Instead, we should:
- roll `UnixSockets` in to `Network`, throwing an unsupported operation on platforms without support
- consider creation of a unified client api like `Network[F].clientFromSpec(SocketClientSpec.tcp(hostAndPort))` where `SocketClientSpec` is a builder that supports full configuration of both TCP and unix sockets.

mpilquist:
Fixes #3560 

This PR revamps `fs2.io.net` in a number of significant ways:
- Unifies support for IP and Unix sockets in to a single set of operations -- `connect` for clients and `bind` for servers, both taking the new `com.comcast.ip4s.GenSocketAddress` type, which is a supertype of `SocketAddress[Host]` and `UnixSocketAddress`
- Resultantly adds support for socket options on unix sockets
- Deprecates `SocketGroup` - there has never really been a great reason to manage explicit thread pools / async socket groups & the API was JVM only. With the new polling system, there's even less of a reason to ever use the explicitly managed socket groups.
- Deprecates `UnixSockets` as there's no utility provided that's not also provided by `Network` now
- Deprecates `client`, `server`, `serverResource` in favor of `connect`, `bindAndAccept`, `bind`
- Introduces `ServerSocket`, which provides the bound address of the server socket, access to options, and a stream of client sockets.
- Introduces `SocketInfo` as a supertype of `Socket` and `ServerSocket`.
- Introduces `address: GenSocketAddress` on `SocketInfo` and `peerAddress: GenSocketAddress` on `Socket`, replacing `localAddress` and `remoteAddress` (each of type `F[SocketAddress[IpAddress]]`.
- Deprecating `isOpen` on `Socket` which is a vestige from when there was a `close` method.

Datagram sockets have been similarly revamped:
- `openDatagramSocket` has been replaced with `bindDatagramSocket`, taking a `GenSocketAddress`
- As of this PR, datagrams over unix sockets are only supported on the JVM, via jnr-unixsocket (JEP-380 doesn't support datagrams). Node.js doesn't have built-in support for datagrams over unix sockets, though we could add via a dependency. Datagram sockets, in general, are not currently implemented on native.

Updates to downstream projects:
- https://github.com/http4s/http4s/pull/7675/
- https://github.com/typelevel/fs2-chat/commit/c4f264324b2c61d876b9d064dd61fd01c5c603ba


io/shared/src/main/scala/fs2/io/net/Network.scala
```diff
@@ -51,6 +51,8 @@ sealed trait Network[F[_]]
     with SocketGroup[F]
     with DatagramSocketGroup[F] {
 
+  def tcp: TcpBuilder.NeedAddress[F]
```
mpilquist, line 54: Current sketch of an API uses a builder pattern for creating TCP clients & servers. Not sure if this is a good idea or not.

There are three bits of information needed to open a TCP socket:
1. address of type `GenSocketAddress`
2. zero or more `SocketOption`s
3. whether to connect or bind

The API `SocketGroup` and hence `Network` currently encode this as two operations -- `client` and `server` -- each which take an address and a list of options defaulted to `Nil`. We can't overload these operations to take a `GenSocketAddress` as doing so introduces ambiguous overloads and conflicts with default parameters.

So we have two options API-wise: (1) `connect`/`bind` operations (2) builder pattern.

The `connect`/`bind` approach avoids the overload issues by simply using fresh names -- we'd deprecate `client` and `server`. These operations would both take a `GenSocketAddress` and `List[SocketOption]`. Unix domain specific server params like `deleteIfExists` and `deleteOnClose` would have to be encoded as proprietary socket options that don't get passed down to underlying socket.

The builder pattern gives us more flexibility. We can provide various convenience constructors for different address types (like in this WIP PR) and we can support arbitrary additional configuration options without tunneling through `SocketOption` -- e.g. `.withServerUnixSocketDeleteIfExists`. OTOH, the builder pattern feels a bit overkill here.

Thoughts?

io/shared/src/main/scala/fs2/io/net/Bind.scala
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * The socket can be inspected, e.g. for its bound port, via the `socketInfo` method.
+ * Note some platforms do not support getting and setting socket options on server sockets
+ * so take care when using `socketInfo`.
+ *
+ * Client sockets can be accepted by pulling on the `accept` stream. A concurrent server
+ * that is limited to `maxAccept` concurrent clients is accomplished by
+ * `b.accept.map(handleClientSocket).parJoin(maxAccept)`.
+ */
+sealed trait Bind[F[_]] {
```
armanbilge, line 36: Naming is hard ... why not `Bound` or `BoundServer` or `BoundSocket` ?

io/shared/src/main/scala/fs2/io/net/SocketInfo.scala
```diff
@@ -0,0 +1,43 @@
...
+package net
+
+import com.comcast.ip4s.{GenSocketAddress, IpAddress, SocketAddress}
+
+trait SocketInfo[F[_]] {
+
+  /** Asks for the local address of the socket. */
+  def localAddressGen: F[GenSocketAddress]
+
+  def localAddress: F[SocketAddress[IpAddress]]
```
armanbilge, line 33: Having both of these feels strange.

io/shared/src/main/scala/fs2/io/net/Socket.scala
```diff
@@ -23,12 +23,12 @@ package fs2
...
 package net
 
-import com.comcast.ip4s.{IpAddress, SocketAddress}
+import com.comcast.ip4s.{GenSocketAddress, IpAddress, SocketAddress}
 import fs2.io.file.FileHandle
 
 /** Provides the ability to read/write from a TCP socket in the effect `F`.
   */
-trait Socket[F[_]] {
+trait Socket[F[_]] extends SocketInfo[F] {
```
armanbilge, line 31: I think this is a breaking change, because `SocketInfo` brings abstract methods.

io/shared/src/main/scala/fs2/io/net/Bind.scala
armanbilge: I guess the JDK just calls this a `ServerSocket` and it would directly extend `SocketInfo`

io/shared/src/main/scala/fs2/io/net/SocketInfo.scala
```diff
@@ -0,0 +1,43 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package fs2
+package io
+package net
+
+import com.comcast.ip4s.{GenSocketAddress, IpAddress, SocketAddress}
+
+trait SocketInfo[F[_]] {
```
armanbilge, line 28: Is it valuable to have this as a public-facing abstraction? I guess the JDK calls this `NetworkChannel`, which is implemented by `SocketChannel`, `ServerSocketChannel`, and `DatagramChannel`.

io/shared/src/main/scala/fs2/io/net/SocketInfo.scala
mpilquist: Yeah, needed for binary compatibility unfortunately

io/shared/src/main/scala/fs2/io/net/Bind.scala
mpilquist: Yeah good idea

io/shared/src/main/scala/fs2/io/net/SocketInfo.scala
mpilquist: Right, we can push it back down to Socket

mpilquist:
Need to lazily load socket options that were added in Java 9+. Here's an example failure with 3.13.0-M2 and http4s tests running on Java 8:
```
java.lang.NoSuchFieldError: SO_REUSEPORT
	at fs2.io.net.SocketOptionCompanionPlatform.$init$(SocketOptionPlatform.scala:75)
	at fs2.io.net.SocketOption$.<clinit>(SocketOption.scala:36)
	at org.http4s.ember.server.internal.ServerHelpers$.$anonfun$unixSocketServer$1(ServerHelpers.scala:150)
```

mpilquist:
UnixDatagramSuite on Linux is failing due to https://github.com/jnr/jnr-unixsocket/pull/107. Will exclude it for now.

PR #3565, opened 2025-04-22
Add ldbc to the ecosystem

takapi327:
Hi!

We have created a database client library with Scala's own connector to connect to MySQL.

https://github.com/takapi327/ldbc

Connectors are created using fs2.
It would be awesome to have it mentioned on the fs2 ecosystem page 🙂


mpilquist:
Happy to, thanks for opening the PR!

PR #3569, opened 2025-05-08
clarify chunkN doc about allowFewer

samspills:
Just clarifying in the docs that using `chunkN` with `allowFewer=true` will only emit the smaller chunk when it is the last chunk in the stream *and* the stream is completed (`chunkN` builds on `unconsMin` which only returns a partial chunk once it hits the end of the stream)

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -446,7 +446,7 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   /** Outputs chunks of size `n`.
     *
     * Chunks from the source stream are split as necessary.
-    * If `allowFewer` is true, the last chunk that is emitted, upon completion of the stream, may have less than `n` elements.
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -446,7 +446,7 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
   /** Outputs chunks of size `n`.
     *
     * Chunks from the source stream are split as necessary.
-    * If `allowFewer` is true, the last chunk that is emitted may have less than `n` elements.
+    * If `allowFewer` is true, the last chunk that is emitted, upon completion of the stream, may have less than `n` elements.
```
diesalbla, line 449: A small mistake that was already there.

```suggestion
    * If `allowFewer` is true, the last chunk that is emitted, upon completion of the stream, may have fewer than `n` elements.
```

diesalbla (approved):

core/shared/src/main/scala/fs2/Stream.scala
samspills: Ah yes, thank you!

PR #3575, opened 2025-06-04
Added Processes to IONative

rahulrangers:
Implemented ProcessBuilder-based process spawning for Scala Native using posix_spawn, with support for stdin, stdout, and stderr piping. Integrated pidfd_open and fileDescriptorPoller to asynchronously and non-blockingly wait for process termination. Falls back to waitpid when pidfd_open is unavailable.


armanbilge (commented):
Thanks, nice work! A few quick thoughts:

1. Let's try to get CI green. The native bindings you introduced may need `@nowarn` annotations (there should be other examples in the codebase).
2. I think we can unguard some `ProcessSuite` tests, at least on Linux.
3. We may want to keep the `java.lang.Process` implementation in shared sources, and use it as a fallback on Native for non Linux/macOS (i.e. Windows).

antoniojimeneznieto:
I checked the timeout exception in the env inheritance test. I think the issue comes in how the envMap/envp is being constructed. Interestingly, adding a dummy variable when the Map is empty seems to solve the problem.

rahulrangers:
@antoniojimeneznieto  Yep, I tried using dummy variable, and it actually resolved the issue.
However, using dummy variable doesn't result in ```countEnv(false) == 0``` as it does on the JVM platform. Instead, we get ```countEnv(true) > countEnv(false)```, which still causes the test to pass.
So I’m unsure if this is the correct fix, since the behavior isn’t exactly the same.

Also, I’m not fully understanding why the EOF signal isn’t being triggered in ```readFd```, which is preventing the loop from exiting. Currently, both test cases failing in CI is due to this issue.

antoniojimeneznieto:
Yes! this isn't a fix, since it may introduce other issues or conflicts. It was more of a remark than a proper solution, sorry for the confusion.

armanbilge (commented):
I am sorry it took me so long to give this the review it deserves! Overall this looks very good and so happy to seem so many tests enabled in the test suite :) great job with that

I do have a quite a bit of feedback but it is mostly polish. Also, a general advice is to use more of the helpers like `guard_` and `errnoToThrowable` in [`NativeUtil`](https://github.com/rahulrangers/fs2/blob/d410dee4273b081c12b0287d064f4e92ed3d2baf/io/native/src/main/scala/fs2/io/internal/NativeUtil.scala#L36).

PR #3578, opened 2025-06-17
Fix minor guide explanation

matsluni:
This should be `F`, right?

### DIFF
--- a/site/guide.md
+++ b/site/guide.md
@@ -284,7 +284,7 @@ Otherwise, we return a function which processes the next chunk in the stream. Th
 
 Sometimes, `scanChunksOpt` isn't powerful enough to express the stream transformation. Regardless of how complex the job, the `fs2.Pull` type can usually express it.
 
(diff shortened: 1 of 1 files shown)


mpilquist:
This is trying to say that there's a `Monad[[x] =>> Pull[F, O, x]]`. "Forming a monad in R" is shorthand for "the hole in the type constructor is in the spot occupied by `R`. We should just reword this to avoid confusion.

matsluni:
@mpilquist Thanks for the clarification. I have to let that sink in first :-)

matsluni:
@mpilquist Ok, now I think understand what you were saying. 

I was also reading the guide further and it became also clear to me how the relationship between Monad and `Pull[F[_], O, R]` is meant.

I don't know on the spot how to improve the documentation. Do you have something particular in mind?

PS: Is there some glossar (I don't know if this is the right term) somewhere, which may explains different ways to explain monad or category sayings, e.g. "needs to have an instance of Monad" or "forms a monad in R"?

PPS: (Sorry, for the noise around this)

PR #3582, opened 2025-07-13
Added `DatagramSocket` for Native with Polling

rahulrangers:
### Add Datagram Socket Support with Polling

This PR adds support for datagram sockets with polling for native. The key changes include:

- **`FdPollingDatagramSocket.scala`**: Defines an interface for datagram sockets.
- **`FdPollingIpDatagramSocketsProvider.scala`**: Creates and binds IP-based datagram sockets.
- **`IpMulticast.scala`**: Provides types for low-level multicast support (`ip_mreq`, `ip_mreq_source`, etc.).
- **`SocketHelpers.scala`**:
  - Adds `join`, `drop`,`block` and `unblock` helpers for multicast`.
  - Adds new `SocketOptions` for datagramSockets.
  - Adds `setIpOption` and `getIpOption`

### DIFF
--- a/io/native/src/main/scala/fs2/io/internal/SocketHelpers.scala
+++ b/io/native/src/main/scala/fs2/io/internal/SocketHelpers.scala
@@ -39,6 +39,10 @@ import scala.scalanative.meta.LinktimeInfo
 import scala.scalanative.posix.arpa.inet._
 import scala.scalanative.posix.errno.ENOPROTOOPT
 import scala.scalanative.posix.netinet.in.IPPROTO_TCP
+import scala.scalanative.posix.netinet.in.IPPROTO_IP
+import scala.scalanative.posix.netinet.in.IPPROTO_IPV6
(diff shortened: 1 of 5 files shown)


io/shared/src/test/scala/fs2/io/net/DatagramSocketSuite.scala
```diff
@@ -0,0 +1,97 @@
...
+package io
+package net
+
+import cats.effect.IO
+import scala.concurrent.duration._
+import com.comcast.ip4s._
+
+import cats.effect.kernel.Resource
+
+class DatagramSocketSuite extends Fs2Suite {
```
mpilquist, line 32: Can we move the existing `UdpSuite` to the shared directory now?

rahulrangers: Yes, I think we can move `UdpSuite` to the shared directory, but the test cases are currently failing on macOS for native platform. I'll fix those issues, and then move it.

PR #3585, opened 2025-07-28
Add metrics to fs2.io.net

mpilquist:
Sketch for now... idea is to hook this up to otel4s via `Meter[F].batchCallback`

cc @iRevive


iRevive:
That would be neat!

PR #3610, opened 2025-09-16
Stream merge blocking downstream work.

### LINKED ISSUE #3598: Some streams merged with `merge` stop emitting when downstream does certain actions
Hello, I came across a bug in an application of mine where we stop seeing updates on one stream when multiple streams are merged with `mergeHaltBoth`. I've come up with a fairly minimal reproduction of the issue. Our application will sometimes inspect the first element of the stream to determine how to process the remaining elements of the stream. We make use of the `splitHead` function used in this example.

Here is the code which demonstrates this issue:
```scala
package bug

import cats.effect.{IO, IOApp}
import fs2.{Chunk, Stream}

object FS2Bug extends IOApp.Simple {

  sealed trait Data

  case class Item(value: Int) extends Data
  case object Tick1 extends Data
  case object Tick2 extends Data

  override def run: IO[Unit] = {
    val source = Stream.emits(1 to 100).evalMap(i => IO(Item(i)).delayBy(2.seconds))

    val timer = fs2.Stream.awakeEvery[IO](1.second).evalMap(_ => IO(()).as(Tick1))
    val timer2 = fs2.Stream.awakeEvery[IO](1.second).evalMap(_ => IO(()).as(Tick2))

    val sources = timer2.mergeHaltBoth(source.mergeHaltBoth(timer))

    val writer = splitHead(sources)
      .flatMap { case (head, tail) =>
        splitHead(tail).flatMap { case (head2, tail) =>
          Stream.emit(head) ++ Stream.emit(head2) ++ tail
        }.parEvalMap(3) { i =>
          IO(println(s"Writing ${i.toString}"))
        }
      }

    writer.compile.drain
  }

  def splitHead[F[_], O](in: fs2.Stream[F, O]): fs2.Stream[F, (O, fs2.Stream[F, O])] =
    in.pull.uncons1
      .flatMap {
        case Some((head, tail)) => fs2.Pull.output(Chunk((head, tail)))
        case None => fs2.Pull.done
      }
      .stream
}

```

The program merges 3 streams together.  It splits the head of the stream from the remaining elements, before merging it back together. The `source` stream will terminate after 100 elements are emitted, the timer streams never terminate. 

When run, this program generates output like:
```
Writing Tick2
Writing Tick1
Writing Item(1)
Writing Tick1
Writing Tick1
Writing Tick1
Writing Item(2)
Writing Tick1
Writing Tick1
Writing Item(3)
Writing Tick1
Writing Tick1
Writing Item(4)
Writing Tick1
Writing Tick1
Writing Item(5)
Writing Tick1
Writing Tick1
Writing Item(6)
Writing Tick1
Writing Tick1
```

Note how `Tick2` from the `timer2` stream is only printed once. This should happen at the same frequency as the `Tick1` outputs.

This affects fs2 versions:
- 3.12.0
- 2.5.10

I'll take a look at getting a PR up with a test to highlight the issue. Thanks

AdamChlupacek:
Stream merge does not block until scope is closed, but only until the chunk is read from the output channel.

Merge was waiting for the resulting chunk to be fully processed, this introduced a scope (resource) to the stream, which then if leased for further parallel processing would cause the merge to be waiting forever (until the parallel processing finished).

Here I am changing that we are only guarding the production of values from merge. Instead of guarding the whole processing of the chunk, we are only guarding the chunk up to the point where it is read from the output of the merge. 

This also fixes #3598. 

This is now aligned with documentation of merge where `Stream(this, that).parJoinUnbounded` == `this merge that`

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -2000,6 +2000,107 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     Stream.force(fstream)
   }
 
+  /** Implementation of [[merge]], however allows specifying how to combine the output stream.
+    * This can be used to control how chunks are emitted downstream. See [[mergeAndAwaitDownstream]] for example.
+    *
+    * @param f The function that combines the output stream and a finalizer for the chunk.
+    *          This way we can controll when to pull pull next chunk from upstream.
+    */
+  private def merge_[F2[x] >: F[x], O2 >: O](
+      that: Stream[F2, O2]
+  )(
+      f: (Stream[F2, O2], F2[Unit]) => Stream[F2, O2]
+  )(implicit F: Concurrent[F2]): Stream[F2, O2] =
+    Stream.force {
+      // `State` describes the state of an upstream stream (`this` and `that` are both upstream streams)
+      // None            : the stream has not yet terminated
+      // Some(Left(t))   : the stream terminated with an error
+      // Some(Right(())) : the stream terminated successfully
+      type State = Option[Either[Throwable, Unit]]
+      for {
+        // `bothStates` keeps track of the state of `this` and `that` stream
+        // so we can terminate downstream when both upstreams terminate.
+        bothStates <- SignallingRef.of[F2, (State, State)]((None, None))
+        // `output` is used to send chunks from upstreams to downstream.
+        // It sends streams, not chunks, to tie each chunk with a finalizer
+        output <- Channel.synchronous[F2, Stream[F2, O2]]
+        // `stopDef` is used to interrupt the upstreams if a) any of the
+        // upstreams raises an error, or b) the downstream terminates.
+        stopDef <- Deferred[F2, Unit]
+      } yield {
+        val signalStop: F2[Unit] = stopDef.complete(()).void
+        val stop: F2[Either[Throwable, Unit]] = stopDef.get.as(Right(()))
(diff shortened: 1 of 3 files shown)


AdamChlupacek:
The failing tests: 

- fs2.StreamMergeSuite.merge not emit ahead 
I think that this test is actually testing the wrong behaviour. It is explicitly stated in documentation of merge that there can be up to 2 chunks pulled (1 from each side), while the downstream data is being consumed.

This test assumes lock step, but we cannot assure this in the first 2 chunks.

If we extend this test we can assure that we do not pul more than 1 chunk any longer. Allowing the first 2 values to be the same and then having the values increment would prove this case. 

- fs2.TimedPullsSuite.After the first uncons, timeouts start immediately
`timed` on stream is actually run using `merge`, in this test we could not actually ever produce a second element ever, since  the first chunk received via `uncons` which acquired the resource in merge and hence could never actually produce a value again. This is the same issue as in #3598.

I will apply fixes to these test if we agree that this change to merge is desired.

mpilquist:
Looks good to me, good find!

AdamChlupacek:
Thank you for the approval, I have fixed the tests.

However in the test: `fs2.StreamMergeSuite.merge not emit ahead` cannot be fixed as is. As the expected behaviour of merge i that we will read 1 chunk ahead. As such I changed the test that we will not read more than 1 chunk ahead. This test was introduced as part of fix to https://github.com/typelevel/fs2/issues/1987, not sure if some people rely on this to happen. Especially since its in conflict with the documentation of `merge`

Not sure about the failed action, seems unrelated.

PR #3618, opened 2025-09-28
fix: close soundness holes from Scala 3.8.0

hamzaremmal:
A soundness hole was fixed in Scala 3.8.0 that allowed this code to compile before. In this PR, we allow the code to compile again by making sure it is well typed. The fix here was straightforward since a `Terminal[U]` can only be a `Pull[F2, O2, Unit]` only if `U =:= Unit`.

See https://github.com/scala/scala3/issues/24096


mpilquist:
Thank you!

PR #3665, opened 2026-01-17
fix: ensure Topic#publish1 returns accurate result during concurrent closure

guptapratykshh:
this pr ensures Topic#publish1 accurately returns Left(Closed) when delivery fails due to concurrent topic closure by checking individual subscriber channel results and also guarantees that Right(()) response from publish1 consistently signifies successful delivery to all currently registered subscribers.

Resolves #3644


core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -164,8 +164,21 @@ object Topic {
...
+                      case Right(_) => Topic.rightUnit.pure[F]
+                      case Left(_)  =>
+                        // Channel send failed, check if topic was closed
+                        state.get.map {
+                          case State.Closed()     => Topic.closed
+                          case State.Active(_, _) => Topic.rightUnit
+                        }
+                    }
+                }
+              }
```
TomasMikula: Minor thing: I _think_ constructing the effect in a right-associative way will lead to slightly faster execution for an `IO`-like `F`.

TomasMikula:
@guptapratykshh So you went full on to satisfy the test as is (as opposed to relaxing the test as suggested above).

My main concern with this implementation is that
 - `publish1` is now uncancelable (via `Ref`'s `flatModify`)
 - `close`, which was already uncancelable, now waits for publishers to finish.

These might be breaking changes for existing users.

core/shared/src/main/scala/fs2/concurrent/Topic.scala
```diff
@@ -164,8 +168,21 @@ object Topic {
             case State.Closed() =>
               Topic.closed.pure[F]
             case State.Active(subs, _) =>
-              foreach(subs)(_.send(a).void)
-                .as(Topic.rightUnit)
+              subs.foldLeft(Topic.rightUnit.pure[F]: F[Either[Topic.Closed, Unit]]) {
```
TomasMikula: This might be a tiny bit simpler using `foldLeftM`. What do you think?

TomasMikula (approved):
Looks good to me, except the one (non-essential) nitpick.

core/shared/src/main/scala/fs2/concurrent/Topic.scala
guptapratykshh: i have refactored `publish1` to use `foldLeftM`

TomasMikula (approved):
Looks good to me.

guptapratykshh:
@mpilquist any updates on this PR

mpilquist:
Apologies on the delay on this one. JS build has a test failure. If we can get that addressed, I'll merge. Thanks!

Issue #3607, opened 2025-09-09
Unexpected behavior of onFinalize behavior


wookievx:
This issue revolves around onFinalize (which would probably translate to the bracket operation) not behaving like expected. During work on fs2-kafka library with the help of colleagues we discovered that partitionMapStream calls `onFinalize` too early causing incorrect rebalance behavior (I am raising the issue here because we pin-pointed it directly to the onFinalize invocation). The reproduction can be found on this branch in a test (To run it just call: `sbt testOnly fs2.kafka.KafkaConsumerSpec -- -z "Graceful"`, it might not happen every time, but it is pretty consistent.): https://github.com/wookievx/fs2-kafka/pull/2.

Here is a fragment of logs, distilled to contain relevant points that showcases that it indeed violates the behavior I would expect from it: that upon stream interruption all already executing evalMaps on the stream are grecefuly cancelled (so uncancallable blocks are allowed to finish) before calling onFinalize callback.

```
Consumer1 started processing key-18
2025-09-09 12:57:44,695 | INFO | o.a.k.c.c.i.ConsumerCoordinator - [Consumer clientId=consumer-rebalance-test-group-2, groupId=rebalance-test-group] Request joining group due to: group is already rebalancing
2025-09-09 12:57:44,695 | INFO | o.a.k.c.c.i.ConsumerRebalanceListenerInvoker - [Consumer clientId=consumer-rebalance-test-group-2, groupId=rebalance-test-group] Revoke previously assigned partitions topic-f508a0db-13b1-4e09-86fd-242539ee67a2-0, topic-f508a0db-13b1-4e09-86fd-242539ee67a2-1, topic-f508a0db-13b1-4e09-86fd-242539ee67a2-2, topic-f508a0db-13b1-4e09-86fd-242539ee67a2-3
Finished stream for topic-f508a0db-13b1-4e09-86fd-242539ee67a2-0 with id e8c0050b-8a6b-4e61-9f86-f7d70e8054a3
Finished stream for topic-f508a0db-13b1-4e09-86fd-242539ee67a2-1 with id c50d9d83-3ef3-46de-a643-778b483cf8e5
Finished stream for topic-f508a0db-13b1-4e09-86fd-242539ee67a2-2 with id 2d509097-9a34-4f08-ad25-3eb523e9adac
Finished stream for topic-f508a0db-13b1-4e09-86fd-242539ee67a2-3 with id 18165c64-476f-4fd7-bd8a-f7dc38e4d1e2
[...]

PR #3676, opened 2026-02-07
Document scope limitations of concurrent Stream combinators

### LINKED ISSUE #3076: parEvalMap* runs resource finaliser before usage

reardonj:
- add general documentation to Stream warning about scopes not extending through operations that use `Concurrent`
- add specific documentation for the couple of methods that are safe

I've taken a stab at documenting the current behavior instead of trying to fix #3076 given how intractable fixing that bug has been.

I ran variations of this code with various combinators to see what the full scope of the problem is:

```scala
Deferred[IO, Unit]
      .flatMap { d =>
        Stream
          .bracket(IO.unit)(_ => d.complete(()).void)
          .prefetch
          .evalMap(_ => IO.sleep(1.second) >> d.complete(()))
          .timeout(5.seconds)
          .compile
          .last
      }
```

Unfortunately, scopes don't seem to propagate for basically any combinator requiring a `Concurrent` instance. More specifically, these are generally passing stream elements through channels to a concurrently running stream, and the scopes of the original stream aren't held open until the background stream completes.

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -92,6 +92,15 @@ import java.util.concurrent.Flow.{Processor, Publisher, Subscriber}
   * but their stream counterparts return an (either empty or singleton) stream.
   * Other methods, like `zipWithPrevious`, have a more complicated but still pure translation to list methods.
   *
+  * == Scopes and Concurrent Effects ==
+  *
+  * Stream combinators which require `F` to have a `Concurrent` instance generally do not propagate open scopes. Finalizers
+  * may be run concurrently with subsequent operations in the stream. This means that these combinators should not be used
+  * with streams that must hold resources (eg. a network socket).
+  *
+  * The `parJoin` family of methods is designed to allow the elements of streams of resources to be used safely concurrently.
+  * Beware however that they still do not propagate the resource scope past the `parJoin` combinator.
+  *
   * == Type-Class instances and laws of the Stream Operations ==
   *
   * Laws (using infix syntax):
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -92,6 +92,15 @@ import java.util.concurrent.Flow.{Processor, Publisher, Subscriber}
...
   * Other methods, like `zipWithPrevious`, have a more complicated but still pure translation to list methods.
   *
+  * == Scopes and Concurrent Effects ==
+  *
+  * Stream combinators which require `F` to have a `Concurrent` instance generally do not propagate open scopes. Finalizers
+  * may be run concurrently with subsequent operations in the stream. This means that these combinators should not be used
+  * with streams that must hold resources (eg. a network socket).
+  *
+  * The `parJoin` family of methods is designed to allow the elements of streams of resources to be used safely concurrently.
+  * Beware however that they still do not propagate the resource scope past the `parJoin` combinator.
```
SystemFw: What's an example of this that is harmful?

reardonj: As it happens, fs2-kafka uses parJoinUnbounded and hits this problem: https://github.com/typelevel/fs2/issues/3607#issuecomment-5025260729

PR #3696, opened 2026-03-13
Wrote 1 test cases, modified writeAll function to stop unwanted file …

ankitkumarrain:
Bug Number in issue section - 3667

Write all function in the main(Files.Scala) used stream.resource (writeCursor(...)) which opens and create the during resource aquisation, before any chunk are pulled from the input stream. I added two test cases 👍 
 1) Empty stream does not create file.
 2) Error stream does not create file .


ankitkumarrain:
hey @armanbilge , Please review this pull request , i have solved a bug that stops creation of unwanted file and also wrote tests for empty stream and error stream. Thank You !!!

io/shared/src/test/scala/fs2/io/file/FilesSuite.scala
```diff
@@ -200,6 +200,32 @@ class FilesSuite extends Fs2Suite with BaseFileSuite {
         .foldMonoid
         .assertEquals("")
     }
+    test("empty stream does not create file") {
```
mpilquist: Writing an empty stream to a file **should** create an empty file.

ankitkumarrain: hey @mpilquist Should i remove this test ?

ankitkumarrain:
Hey @mpilquist I have removed the empty stream test . Wanted a review again .

PR #3708, opened 2026-04-05
Modified documentation to FilterWithPrevious

ankitkumarrain:
Modified Documentation in Stream.scala . Mainly 2 docs i.e evamMapChunk and FilterWithPrevious .

### DIFF
--- a/core/shared/src/main/scala/fs2/Stream.scala
+++ b/core/shared/src/main/scala/fs2/Stream.scala
@@ -1040,17 +1040,18 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     underlying.flatMapOutput(evalOut).streamNoScope
   }
 
-  /** Like `evalMap`, but operates on chunks for performance. This means this operator
-    * is not lazy on every single element, rather on the chunks.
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/fs2/Stream.scala
```diff
@@ -1040,17 +1040,18 @@ final class Stream[+F[_], +O] private[fs2] (private[fs2] val underlying: Pull[F,
     underlying.flatMapOutput(evalOut).streamNoScope
   }
 
-  /** Like `evalMap`, but operates on chunks for performance. This means this operator
-    * is not lazy on every single element, rather on the chunks.
+  /** Like `evalMap`, but operates on chunks for performance. Evaluates `f` for all elements
+    * within a chunk using the Applicative instance, which allows parallel evaluation if supported by `F`.
```
mpilquist: Recommend removing "using the Applicative instance, which allows ...". Feels out of place here.

ankitkumarrain: Hey @mpilquist, thank you for your suggestions. I’ve updated the changes accordingly. Could you please take another look?

Issue #3697, opened 2026-03-16
Stream.merge hangs when one branch is empty + repeat, unresponsive to interruptAfter and IO.cancel


TalkingFoxMid:
### Description:

When merging two streams where one branch is Stream.empty.repeat (an infinitely repeating empty stream), the resulting stream hangs indefinitely. Neither interruptAfter nor explicit fiber cancellation via IO.cancel can terminate it.

### Expected behavior: 

The merged stream should complete (or be interruptible) once the non-empty branch is exhausted and the interrupt/cancel signal is received.

### Actual behavior: 

The stream hangs forever. Both interruptAfter and fiber.cancel have no effect.

### The code:

```
object Main extends IOApp.Simple {

  override def run: IO[Unit] =
    Supervisor.apply[IO].use { s =>
      for {
        fibra <-
          Stream.empty.repeat
            .merge(
              Stream.emits[IO, Int](List(4, 5, 6))
            )
          .interruptAfter(1.seconds)
          .compile
          .drain
          .start
        _ <- IO.sleep(1.seconds)
        _ <- fibra.cancel
      } yield ()
    }
}
```

LoveChauhan-18:
Hi, I would like to work on this issue. Please assign it to me!

PR #3713, opened 2026-04-09
Added f.cede for async boundry on empty stream

ankitkumarrain:
Issue Number -#3697

I modified documentation of the merge combinator. I added f.cede for async boundry and filter empty chunks to avoid unnecessary sendchunk calls.


ankitkumarrain:
Hey @armanbilge and @mpilquist Please review this code , when you have time .

armanbilge (changes requested):
Thanks for the PR. Please add a test related to the linked issue.

ankitkumarrain:
Hi @armanbilge I have added test for the linked issue with a .ignore mark to skip test. Because it actaully did not made a fix to the issue.
After a deep digging in , Stream.merge.repeat hangs because stream.merge.repeat internally creates an infinite chain of bind (Succeeded(()),k) nodes. The viewl function on pull.scala is a @tailrac loop that handles bind (treminal,count) by immediately recursing, It has no IO boundry in between . This pointed me the fiber running in the stream.empty.repeat is struck in a pure JVM loop forever.
Well, About my changes i have added filter(_.non Empty) change in merge combinator that avoids unnecessary sendChunk calls for empty chunks in the merge_ path.
I was looking for preffered approach for the fix in the right direction . Can you please help me ?

PR #3714, opened 2026-04-11
Support fs2.io.readOutputStream on Scala Native

lolgab:


io/jvm-native/src/test/scala/fs2/io/IoPlatformSuite.scala
```diff
@@ -275,7 +275,8 @@ class IoPlatformSuite extends Fs2Suite {
       bar.assertEquals("bar")
     }
     test("classloader") {
-      val size = readClassLoaderResource[IO]("fs2/io/foo", 8192).as(1L).compile.foldMonoid
+      val resourcePath = if (isNative) "/fs2/io/foo" else "fs2/io/foo"
```
armanbilge, line 278: This difference surprised me. Does the absolute path not work on the JVM?

lolgab: Haven't investigated, but I think it's a bug in Scala Native, but since it's test code I preferred to add a `if` here and eventually when gets fixed upstream you'll see a failing test and it's trivial to remove the `if`.

lolgab:
Unfortunately the tests are hitting timeout on Scala Native, there might be something wrong with the Scala Native implementation (?).

Did you see [this](https://github.com/lolgab/fs2/actions/runs/24302873915/job/70989534254) before @armanbilge?

io/jvm-native/src/test/scala/fs2/io/IoPlatformSuite.scala
armanbilge: > when gets fixed upstream you'll see a failing test

Maybe! But also it could be that both are valid, in which case there wouldn't be a failed test.

armanbilge:
> Unfortunately the tests are hitting timeout on Scala Native, there might be something wrong with the Scala Native implementation (?).

I don't think there is anything wrong per-se, I think it may just actually be timing out since it involves a very large stream and may be running a bit slowly in Scala Native. Does it pass locally if you increase the timeout?

https://github.com/typelevel/fs2/blob/3ec3a250ec825db8c7879b065ad4c562f059c726/io/jvm-native/src/test/scala/fs2/io/IoPlatformSuite.scala#L223-L226

io/jvm-native/src/test/scala/fs2/io/IoPlatformSuite.scala
armanbilge: You're right. I just checked, and the leading `/` doesn't work on the JVM. So it's a bug :)

armanbilge:
> Does it pass locally if you increase the timeout?

Seems to pass for me. So we just need to increase the timeout for CI.

lolgab:
Thank you @armanbilge, increasing the timeout indeed did fix it!

armanbilge (approved):
thanks!

io/jvm-native/src/test/scala/fs2/io/IoPlatformSuite.scala
lolgab: Just opened a PR on Scala Native to fix the bug: https://github.com/scala-native/scala-native/pull/4901

lolgab: Fixed in #3728

PR #3721, opened 2026-05-05
Add doc for installation of s2n

Quafadas:
I bumped into this whilst attempting a http4s native project - https://github.com/http4s/http4s/issues/7826, so doing my best to be a good citizen.

### DIFF
--- a/site/getstarted/install.md
+++ b/site/getstarted/install.md
@@ -28,7 +28,7 @@ libraryDependencies += "co.fs2" %%% "fs2-scodec" % "@VERSION@"
 
 // Node.js only, and requires module support to be enabled
 libraryDependencies += "co.fs2" %%% "fs2-io" % "@VERSION@"
-scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule)) 
+scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule))
 ```
 
 The fs2-core as well as fs2-io and fs2-scodec libraries are also supported on Scala Native:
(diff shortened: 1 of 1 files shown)


Quafadas:
Note: on a second, closer reading of the install docs, there is a hint about this already on the install page.

```
// TLS support requires https://github.com/aws/s2n-tls
libraryDependencies += "co.fs2" %%% "fs2-io" % "3.13.0"
```
So this PR may be superfluous - although I would still make the case for an anchor that can be externally referenced from other projects, e.g. http4s.

site/getstarted/install.md
```diff
@@ -48,3 +48,36 @@ If upgrading from the 2.x series, see the [release notes for 3.0.0](https://gith
...
+By way of an example on mac OS;
+
+```sh
+brew install s2n
+```
+And then, we must tell our build tool, where to find this library;
+
+```scala
+nativeLinkingOptions ++= Seq(
+  "-L/opt/homebrew/Cellar/s2n/1.7.2"
```
mpilquist: Curious why this step is necessary - I've never done it and instead have just used `brew install s2n`.

site/getstarted/install.md
```diff
@@ -48,3 +48,36 @@ If upgrading from the 2.x series, see the [release notes for 3.0.0](https://gith
 If upgrading from the 1.x series, see the [release notes for 2.0.0](https://github.com/typelevel/fs2/releases/tag/v2.0.0) for help with upgrading.
 
 There are [detailed migration guides](https://github.com/typelevel/fs2/blob/main/docs/) for migrating from older versions.
+
+### Native
+
+For scala native, fs2 has a dependancy on [s2n](https://github.com/aws/s2n-tls). To be able to link a scala-native fs2 based application, s2n will need to be made available to the linker. If you are observing a linker error in the logs similar to the following on native (commonly, if fs2 is a transitive dependency):
```
mpilquist: Note openssl is also required.

site/getstarted/install.md
Quafadas: Huh. The honest answer, is that I can't substantiate it.

My recollection is that "brew install s2n" didn't "just work" after a brew install with mill. 

I found that setting in the docs of both mill and set, and apparently it worked afterwards. However, if I now comment it out of the build config, it does indeed work - suggestive that this step isn't necessary as you say. 

As I can't replay that first time experience, I'm inclined to believe I had made some other mistake on the journey and mistakenly diagnosed this as being a necessary step. I'll remove it from the PR.

Quafadas: I have the hypothesis, that if I did the brew install in a separate terminal, perhaps mill's path was stale, making this a temporary workaround for that single terminal session.

site/getstarted/install.md
Quafadas: thanks. I had that in the same redundant setting, oddly enough...

PR #3553, opened 2025-03-22
fix(io): handle relative paths in Watcher by converting to absolute p…

SanjayUG:
# Fix: Handle Relative Paths in Watcher by Converting to Absolute Paths

## Which issue does this PR close?

- Closes #3472

## Rationale for this change

This update fixes a NullPointerException (NPE) that occurs when watching files with relative paths by converting them to absolute paths before registration. This ensures that the file watcher can handle relative paths correctly.

## What changes are included in this PR?

- **fix(io):** Handle relative paths in Watcher by converting to absolute paths
  - Fixes NPE when watching files with relative paths by converting them to absolute paths before registration.

## Are these changes tested?

Yes, the changes are tested to ensure that the Watcher can handle relative paths without causing a NullPointerException.

## Are there any user-facing changes?

No, there are no user-facing changes as this update is related to internal path handling improvements.

## Are there any breaking changes to public APIs?

No, there are no breaking changes to public APIs.

## Additional Information

Thanks for opening a PR!

If the CI build fails due to Scalafmt, run `sbt scalafmtAll` and push the changes (generally a good idea to run this prior to opening the PR).

It's also helpful to run `sbt prePR`, which runs tests, generates docs, and performs various checks.

Feel free to ask questions on the fs2 and fs2-dev channels on the [Typelevel Discord server](https://discord.gg/9V8FZTVZ9R).

PR #3750, opened 2026-07-20
fix: avoid NPE when a relative path is provided to Watcher::watch

i10416:
Before this commit, `Watcher::watch` throws NPE when a path is relative path without a parent.
This commit updates watchFile methods to defensively convert a path into absolute path and adds a test to confirm it won't throw NPE.

It supersedes https://github.com/typelevel/fs2/pull/3553 by SanjayUG


i10416:
@mpilquist This is not urgent, but would you mind taking a look at this PR if you have a time?

PR #3751, opened 2026-07-20
feat(protocols): ntp packet

i10416:
This PR adds NTP packet codec to the protocols module.
https://datatracker.ietf.org/doc/html/rfc5905


mpilquist:
Happy to merge this when ready

i10416:
Thanks for the feedback! I was on the fence about adding this feature, but let's move forward with this.

PR #3756, opened 2026-08-13
Fix linesLimited bypass when line completes in one chunk

### LINKED ISSUE #3725: text.linesLimited lets through arbitrarily long lines when the line and its terminator land in the same chunk
## Version

Scala 3.8.3, fs2-core 3.13.0 (also reproduces on 3.12.0; verified against current `main`).

## Minimized code

```scala
//> using scala 3.8.3
//> using dep co.fs2::fs2-core:3.13.0

import fs2._

@main def repro091(): Unit =
  // 1000 chars, limit is 10
  val longLine = "x" * 1000 + "\n"

  val singleChunk = Stream.emit(longLine).covary[Fallible]
    .through(text.linesLimited(10)).toList
  println(s"Single chunk: $singleChunk")

  val multiChunk = Stream.emits(longLine.toList.map(_.toString)).covary[Fallible]
    .through(text.linesLimited(10)).toList
  println(s"Multi  chunk: $multiChunk")
```

## Console output

```
Single chunk: Right(List(xxxxxx...1000 chars))
Multi  chunk: Left(LineTooLongException: ... limit 10)
```

## Expected result

`text.linesLimited(maxLineLength)` raises `LineTooLongException` whenever any line exceeds `maxLineLength`, regardless of how the input is chunked.

## Actual result

When a complete line including its terminator arrives inside one chunk, the line is processed entirely inside `fillBuffers` and pushed straight into `linesBuffer` (the completed-lines accumulator). The `maxLineLength` check at line 556 of `text.scala` only inspects `stringBuilder.length` — the *pending* line — which is 0 after the line has been flushed. Lines already in `linesBuffer` aren't checked against the limit, so any line completed within a chunk passes through.

## Source

https://github.com/typelevel/fs2/blob/b2d1f5917bc571b782b86aaf4ea15a72650441ef/core/shared/src/main/scala/fs2/text.scala#L484-L567

`fillBuffers` (lines 484-527) processes input character by character. On a newline the completed line is pushed to `linesBuffer` (lines 509 / 512) and `stringBuilder` is cleared (line 513). After `fillBuffers` returns, the check at lines 555-559 only looks at `stringBuilder.length`:

```scala
maxLineLength match {
  case Some((max, raiseThrowable)) if stringBuilder.length > max =>
    Pull.raiseError[F](new LineTooLongException(stringBuilder.length, max))(using raiseThrowable)
  case _ =>
    Pull.output(Chunk.from(linesBuffer)) >> go(...)
}
```

Two ways to fix:

- Check the length of each line as it's added to `linesBuffer` inside `fillBuffers` and raise immediately, or
- Check all lines in `linesBuffer` before outputting them.

The single existing test in `TextSuite.scala` line 311 only covers a no-newline single-line input — which always sits in `stringBuilder` and is properly checked — so this gap was hidden.

Happy to follow up with a PR adding both a fix and a regression test that exercises both chunkings.

arimu1:
## Summary

- `text.linesLimited` only checked `stringBuilder.length` after processing each chunk, so lines flushed into `linesBuffer` inside `fillBuffers` (when a line and its terminator arrive in the same chunk) bypassed the limit.
- Check completed lines in `linesBuffer` as well as the pending builder before emitting output.
- Add regression test covering single-chunk vs per-character chunking (repro from #3725).

Fixes #3725

## Test plan

- [x] `sbt "coreJVM/testOnly fs2.TextSuite"` — 92 tests passed (JDK 17)
- [x] New test: `linesLimited rejects long lines completed in a single chunk`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/core/shared/src/main/scala/fs2/text.scala
+++ b/core/shared/src/main/scala/fs2/text.scala
@@ -553,11 +553,25 @@ object text {
           }
 
           maxLineLength match {
-            case Some((max, raiseThrowable)) if stringBuilder.length > max =>
-              Pull.raiseError[F](
-                new LineTooLongException(stringBuilder.length, max)
-              )(using raiseThrowable)
-            case _ =>
+            case Some((max, raiseThrowable)) =>
+              val tooLongLength =
+                (linesBuffer.iterator.map(_.length) ++ Iterator(stringBuilder.length))
+                  .filter(_ > max)
+                  .reduceOption(_ max _)
+              tooLongLength match {
+                case Some(length) =>
+                  Pull.raiseError[F](
+                    new LineTooLongException(length, max)
+                  )(using raiseThrowable)
+                case None =>
+                  Pull.output(Chunk.from(linesBuffer)) >> go(
+                    stream,
+                    stringBuilder,
+                    ignoreFirstCharNewLine,
+                    first = false
+                  )
+              }
+            case None =>
               Pull.output(Chunk.from(linesBuffer)) >> go(
                 stream,
                 stringBuilder,
--- a/core/shared/src/test/scala/fs2/TextSuite.scala
+++ b/core/shared/src/test/scala/fs2/TextSuite.scala
@@ -324,6 +324,15 @@ class TextSuite extends Fs2Suite {
         )
       }
     }
+
+    test("linesLimited rejects long lines completed in a single chunk") {
+      val longLine = "x" * 1000 + "\n"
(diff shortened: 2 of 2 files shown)


arimu1:
Note for maintainers/CI watchers: the red `root*` / Native jobs on the opening push look like GitHub Actions infra (Homebrew `unknown install step: run` / bootstrap), not product assertions from this change. Local TextSuite was green when opened. Happy to re-run when the matrix is healthy.

mpilquist:
Thanks!

Issue #3762, opened 2026-09-03
fs2-io is not closing ports when a DatagramChannel is used


markehammons:
On fs2 3.13.0. I was testing usage of DatagramChannel with some code of mine in an integration test and noted that subsequent tests/suites would fail with the error `java.net.BindException: Address already in use`. 

Attached is a minimal test case that shows the failure. It's not sufficient to merely create the DatagramChannel, it has to be used in a read/write. I believe the Selector registration may be at fault, it may not be being closed:

https://stackoverflow.com/questions/12977164/datagramchannel-close-keeps-port-open-on-windows

https://scastie.scala-lang.org/GD6QuczvSRitJaMBrLyZPw

reardonj:
Note: Setting the options `SocketOption.reusePort(true), SocketOption.reuseAddress(true)` is not effective because options are not being set until after bind: https://github.com/typelevel/fs2/blob/00814bf90f135a132f49c737e5a59491c5a2a0eb/io/jvm/src/main/scala/fs2/io/net/SelectingIpDatagramSocketsProvider.scala#L49-L50

I'm going to see if I can get a fix together today.

reardonj:
PR to fix this is out in #3764 , but even then, testing on some platforms still failed when trying to immediately reuse a port until I include the `SocketOption.reuseAddress(true)` option. This is probably necessary to be safe on all platforms. I don't think there is a good way for us to ever know when the port is actually available again (besides using it, which would be a bit circular).

Currently, using the options doesn't help because they aren't set early enough. The PR fixes this.

PR #3764, opened 2026-09-03
Fix SelectingIpDatagramSocket option and close handing

reardonj:
Fixes two issues with SelectingIpDatagramSockets from #3762:

- **Options not applied before binding**: Some UDP socket options (eg. SO_REUSEPORT) must be set before the socket is bound to a port. `SelectingIpDatagramSocketsProvider` was waiting until after bind to set options, making these options ineffective. All options can be set before binding, so we just need to move up the setOptions before the bind.
- **No wait for selector channels to be deregistered when cleaning up**: Duplicated the wait logic from SelectingIpSocketsProvider for UDP as well, so we wait for the socket to be unregistered, as the port will not actually be available for reuse until after this. I also simpified the logic in both places to use `whileM_` instead of an explicit tailrec loop. Note: It may still be unavailable at the OS level, so it is best to also use the SO_REUSEADDR option (`SocketOption.reuseAddress(true)`) if you really want to close and reopen a UDP socket immediately.


io/jvm/src/main/scala/fs2/io/net/SelectingIpDatagramSocketsProvider.scala
```diff
@@ -40,14 +42,18 @@ private final class SelectingIpDatagramSocketsProvider[F[_]](selector: Selector)
...
+        F.delay(ch.close()) >> F.whileM_(F.delay(ch.isRegistered()))(F.sleep(2.millis))
+      }
       .evalMap { ch =>
         address.host.resolve[F].flatMap { addr =>
           val jAddr = new InetSocketAddress(addr.toInetAddress, address.port.value)
           F.delay {
             ch.configureBlocking(false)
-            ch.bind(jAddr)
             options.foreach(opt => ch.setOption(opt.key, opt.value))
+            ch.bind(jAddr)
```
mpilquist, line 56: Do we need a similar change for JS and Native?

io/jvm/src/test/scala/fs2/io/net/udp/UdpSuitePlatform.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import cats.effect.IO
+
+import scala.concurrent.duration._
+
+trait UdpSuitePlatform { self: UdpSuite =>
+  val concurrentBindOptionsPlatform =
+    List(SocketOption.reuseAddress(true), SocketOption.reusePort(true))
+
+  test("Network allows reuse of port immediately") {
```
mpilquist, line 35: Looking at commit history makes me think this doesn't pass on JS/Native?

reardonj: Native doesn't support UDP (which was surprising, but explains why the UdpSuite is in js-jvm) and JS doesn't support socket options. Realistically, I don't think I can really properly test the cleanup code since there is no guarantee the port is actually available immediately without the socket option, which would also let us reuse the port before closing it.

io/jvm/src/main/scala/fs2/io/net/SelectingIpDatagramSocketsProvider.scala
reardonj: JS doesn't support socket options, and native doesn't support UDP :/

mpilquist: Node supports `reusePort` and `reuseAddr` though doesn't look like we support it in our JS binding. Forgot we never added UDP support to native!

